PR #2409, opened 2019-11-06
Fix parsing of temporary message

nickvergessen:


nickvergessen:
If you have a chat with a scrollbar, the menu open/close button is jumping in and out.
At the same time the button is also jumping up and down


Bildschirmfoto von 2019-11-06 10-44-50
Bildschirmfoto von 2019-11-06 10-44-57

PR #2425, opened 2019-11-07
Feature vuejs add ability to add participants

ma12-co:


src/components/Sidebar/ParticipantsList/Participant/Participant.vue
```diff
@@ -71,29 +69,47 @@ export default {
 	},
 
 	props: {
-		userId: {
-			type: String,
-			required: true,
+		participant: {
+			type: Object,
+			required: true
```
nickvergessen:
```suggestion
			required: true,
```

src/components/Sidebar/ParticipantsList/Participant/Participant.vue
```diff
@@ -71,29 +69,47 @@ export default {
...
 		},
-		displayName: {
-			type: String,
-			required: true,
+	},
+
+	computed: {
+		/**
+		 * If the Participant component is used as to display a search result, it will
+		 * return true. We use dis not to display actions on the searched contacts and
```
nickvergessen:
```suggestion
		 * return true. We use this not to display actions on the searched contacts and 
```

PR #2458, opened 2019-11-13
Guest support

nickvergessen:


src/store/actorStore.js
```diff
@@ -0,0 +1,91 @@
+/**
+ * @copyright Copyright (c) 2019 Marco Ambrosini <marcoambrosini@pm.me>
+ *
+ * @author Marco Ambrosini <marcoambrosini@pm.me>
```
ma12-co: Joas :)

src/store/actorStore.js
```diff
@@ -0,0 +1,91 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
```
ma12-co: It would be nice to have some comments here to say what this store is about. We should maybe do it for all the stores?

nickvergessen: Done

PR #2460, opened 2019-11-13
Fix the avatar of deleted users

nickvergessen:
1. As user1 create a conversation with user2
2. As user2 write something in the chat
3. Delete user2
4. As user1 check the chat

### Before
`>_` Bot avatar

### After
`X` Guest style avatar


nickvergessen:
Oh, now I read your comment. Hmm could also do that. not sure how the 2 backgrounds will work out, but feel free to give it a try

PR #2464, opened 2019-11-14
Prevent lookForNewMessage call infinite loop

### LINKED ISSUE #2462: Conversation with no messages infinite lookfornewmessages call loop.
1) open the network panel
2) Create a new conversation

ma12-co:
fix #2462

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -351,18 +351,17 @@ export default {
 		 * @returns {string} The last known message id.
 		 */
 		getLastKnownMessageId() {
-			for (let i = 1; i < this.messagesList.length; i++) {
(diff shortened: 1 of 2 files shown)


src/components/MessagesList/MessagesList.vue
```diff
@@ -351,18 +350,16 @@ export default {
...
-					}
-				}
+			// Reverse a copy of the messages array
+			const reversedMessages = this.messagesList.slice().reverse()
+			// Get the id of the last non-temporary message
+			for (const message of reversedMessages) {
+				const id = message.id.toString()
+				if (!id.startsWith('temp-')) {
+					return id
+				} else return '0'
```
nickvergessen: This should be illegal...
Always use `{}`

Also you return here always on the first item. So if there is a temporary message you return with the 0 instead of looping to the next message.

PR #2469, opened 2019-11-15
minor updates

gorn:
more done in transiflex.

I did not found the major glitch - name of the module is tranlated like "Mluvte" somewhere (it appears under the Talk icon in top nav menu), which is nonsense.


nickvergessen:
Sorry, translations have to happen on transifex.
They are afterwards synced daily to our repository.

gorn:
Thanks, I have made more extensive patches at transiflex, however I am unable to find the source string used in the navigation bar (which currently reads  "Společenské" as translation of "Social"

PR #2478, opened 2019-11-20
Start with the call service

nickvergessen:
- [x] Current participant info is only available when you visited the participants tab:
  - This does not work for guests at all
  - The information for the current user should be taken from the conversation endpoint to populate the current user at least
- [x] The Conversation info is not updated instantly but just on the next 30s pull, so the current user can trigger multiple joins
- [ ] The audio/video flag need to be set based on a store (last used settings + start without video in big calls) - Maybe in a follow up

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -167,8 +167,12 @@ export default {
 		})
 
 		if (getCurrentUser()) {
+			console.debug('Setting current user')
 			this.$store.dispatch('setCurrentUser', getCurrentUser())
+		} else {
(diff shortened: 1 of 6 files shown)


nickvergessen:
First step of calling is done, still no WebRTC and no signaling. but you join and leave and it triggers the system message.

nickvergessen:
> So this is marked as checked but in my case I can still trigger multiple joins. The button says 'Start call' fore some seconds after the first click.

Works fine here, because I now read the state of the participant which is updated via the store :thinking: 

But lets continue for now.

PR #2486, opened 2019-11-22
Add chat in sidebar when in call

nickvergessen:
- [ ] ~~Currently it seems that conditional `AppSidebarTab` are not working when they are disabled in the beginning and enabled afterwards...~~ Moved to https://github.com/nextcloud/nextcloud-vue/issues/747

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -25,7 +25,8 @@
 		<AppContent>
 			<router-view />
 		</AppContent>
-		<RightSidebar />
+		<RightSidebar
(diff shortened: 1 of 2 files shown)


ma12-co:
> Moved to nextcloud/nextcloud-vue#747

So should we close this?

nickvergessen:
Only the todo was moved. This here needs merging and once the vue library was fixed to be able to handle this case we make another PR to use what we should. For now the tab is always registered and a placeholder message is shown when the tab should jot be displayed

Issue #2414, opened 2019-11-06
Flow event and action


blizzz:
We would like to equip Talk for Nc 18 with following events and actions:

## Events

* [ ] call has finished (provide conversation name, start and end date, participants, conversation message in this time frame)

## Actions (Operations)

* [x] write a message into a Talk conversation
* [ ] create a Talk notification (so probably like ^ but one on one)

The actions should be independent of the triggering event, nevertheless should support/consider some context (e.g. the event itself like file modified, file path/name for Files, or also tag when in combination with a tagging event).

An Action (Operation in tech terms) implementation to look at could be https://github.com/nextcloud/workflow_script/pull/22

As discussed @danxuliu  :) Don't hesitate to reach out for any questions, starter points, API stuff, also missing bits.

blizzz:
@danxuliu had the wrong mention before

nickvergessen:
Closing?

blizzz:
We still might get a call has finished (and perhaps started) event in (just not for 18/this week). Even if with less information. Just flipping the milestone?

marcelklehr:
An event for new chat messages would be nice, too.

nickvergessen:
There are events available for other apps. A floe action for talk messages doesn't make a lot of sense to me.

What do you want to do with the event?

marcelklehr:
I was thinking of automatically collecting links posted into a chat room in the bookmarks app, as an example. I've seen the existing events and they're great, but I think there are many instances like this one, where you don't want to impose that integration on all users, but rather have it optional via flow.

marcelklehr:
In general I think exposing all events in flow would make a lot of sense as they exist already.

nickvergessen:
Well there are a lot of events. Talk alone has around 20. And also not all of them should be accessible to all users, but e.g. moderators or no one. Also woth a flow about messages, you could create problems as there a flows which create messages and so on.

marcelklehr:
I agree, maybe not all events should be available. Shooting yourself in the foot will inevitably be a problem with all apps that expose both operations and events. It might be worth thinking of a general solution for this.

blizzz:
IIRC comments posted trigger an events via Dispatcher. @marcelklehr / Bookmarks could hook in there. I guess the question is mostly about configurability: whether to pick up links generally, or only from selected channels, perhaps giving pre-defined tags (not systemtags). With this in mind, flows might make sense, but it is not the only possible route to go.

[...]

PR #2487, opened 2019-11-22
add post to conversation flow operator

blizzz:
for #2414 

* [x] respect mode
* [x] get the message text from outside (needs https://github.com/nextcloud/server/pull/18134)
* [x] get the actual scope user (needs https://github.com/nextcloud/server/pull/18134)
* [x] some cleanups/consolidations
* [x] requires https://github.com/nextcloud/server/pull/17834 and patched HookConnector for files events, I'll shoot a PR

fyi @juliushaertl @nickvergessen

### DIFF
--- a/lib/Flow/Operation.php
+++ b/lib/Flow/Operation.php
@@ -151,7 +151,9 @@ class Operation implements IOperation {
 			case self::MESSAGE_MODES['ROOM_MENTION']:
 				return '@all ';
 			case self::MESSAGE_MODES['SELF_MENTION']:
-				return '@"' . $participant->getUser() . '" ';
+				$hasWhitespace = strpos($participant->getUser(), ' ') !== false;
+				$enclosure = $hasWhitespace ? '"' : '';
+				return '@' . $enclosure . $participant->getUser() . $enclosure . ' ';
 			case self::MESSAGE_MODES['NO_MENTION']:
 			default:
 				return '';
--- a/lib/Flow/Operation.php
+++ b/lib/Flow/Operation.php
@@ -113,39 +113,31 @@ class Operation implements IOperation {
 		$this->validateOperationConfig($mode, $token);
 	}
 
-	/**
-	 * Is being called by the workflow engine when an event was triggered that
-	 * is configured for this operation. An evaluation whether the event
-	 * qualifies for this operation to run has still to be done by the
-	 * implementor by calling the RuleMatchers getMatchingOperations method
-	 * and evaluating the results.
-	 *
-	 * If the implementor is an IComplexOperation, this method will not be
-	 * called automatically. It can be used or left as no-op by the implementor.
-	 *
-	 * @since 18.0.0
-	 */
 	public function onEvent(string $eventName, Event $event, IRuleMatcher $ruleMatcher): void {
 		$flows = $ruleMatcher->getMatchingOperations(self::class, false);
 		foreach ($flows as $flow) {
 			try {
 				list($mode, $token) = $this->parseOperationConfig($flow['operation']);
 				$this->validateOperationConfig($mode, $token);
+
+				$room = $this->getRoom($token);
+				$participant = $this->getParticipant($room);
+				$this->chatManager->sendMessage(
+					$room,
+					$participant,
+					'bots',
+					$participant->getUser(),
+					$this->prepareMention($mode, $participant) . 'MESSAGE TODO',
+					new \DateTime(),
+					null
+				);
 			} catch(UnexpectedValueException $e) {
 				continue;
+			} catch (ParticipantNotFoundException $e) {
+				continue;
+			} catch (RoomNotFoundException $e) {
+				continue;
 			}
-
-			$room = $this->getRoom($token);
-			$participant = $this->getParticipant($room);
-			$this->chatManager->sendMessage(
-				$room,
-				$participant,
-				'bots',
-				$participant->getUser(),
-				$this->prepareMention($mode, $participant) . 'MESSAGE TODO',
-				new \DateTime(),
-				null
-			);
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


lib/Flow/Operation.php
```diff
@@ -0,0 +1,218 @@
...
+		foreach ($flows as $flow) {
+			try {
+				list($mode, $token) = $this->parseOperationConfig($flow['operation']);
+				$this->validateOperationConfig($mode, $token);
+			} catch(UnexpectedValueException $e) {
+				continue;
+			}
+
+			$room = $this->getRoom($token);
+			$participant = $this->getParticipant($room);
```
nickvergessen: Throwing here is okay?

src/PostToConversation.vue
```diff
@@ -0,0 +1,103 @@
...
+	</div>
+</template>
+
+<script>
+import Multiselect from '@nextcloud/vue/dist/Components/Multiselect'
+import axios from '@nextcloud/axios'
+
+const conversationModeOptions = [
+	{
+		id: 1,
```
nickvergessen: Magic numbers

See src/constants.js for examples

src/PostToConversation.vue
```diff
@@ -0,0 +1,103 @@
...
+	},
+	data() {
+		return {
+			modeOptions: conversationModeOptions,
+			roomOptions: [],
+		}
+	},
+	computed: {
+		currentRoom() {
+			console.debug('room ' + this.value)
```
nickvergessen: debug stuff

src/PostToConversation.vue
```diff
@@ -0,0 +1,103 @@
...
+		},
+	},
+	beforeMount() {
+		this.fetchRooms()
+	},
+	methods: {
+		fetchRooms() {
+			axios.get(OC.linkToOCS('/apps/spreed/api/v1', 2) + 'room').then((response) => {
+				this.roomOptions = response.data.ocs.data.filter(function(room) {
+					return room.readOnly === 0
```
nickvergessen: magic number

src/PostToConversation.vue
```diff
@@ -0,0 +1,103 @@
...
+			}
+			console.debug('sel mode ' + selectedMode)
+			return newValue
+		},
+	},
+	beforeMount() {
+		this.fetchRooms()
+	},
+	methods: {
+		fetchRooms() {
```
nickvergessen: There is a RoomSelector.vue available for the projects stuff. Would it make sense to reuse?

src/flow.js
```diff
@@ -0,0 +1,8 @@
+import PostToConversation from './PostToConversation'
```
nickvergessen:
```suggestion
import FlowPostToConversation from './views/FlowPostToConversation'
```
Can this be moved to`src/views/FlowPostToConversation.vue`?

lib/Flow/Operation.php
```diff
@@ -0,0 +1,218 @@
...
+		}
+
+		$mode = (int)($opConfig['m'] ?? 0);
+		$token = trim((string)($opConfig['t'] ?? ''));
+
+		return [$mode, $token];
+	}
+
+	protected function validateOperationConfig(int $mode, string $token): void {
+		if(($mode < 1 || $mode > 3)) {
```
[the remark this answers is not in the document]
blizzz: nope, but moved to a constant locally, didn't push yet

lib/Flow/Operation.php
```diff
@@ -0,0 +1,218 @@
...
+			$room = $this->getRoom($token);
+		} catch (RoomNotFoundException $e) {
+			throw new UnexpectedValueException('Room not found', $e->getCode(), $e);
+		}
+
+		if($mode === 3) {
+			try {
+				$participant = $this->getParticipant($room);
+				if (!$participant->hasModeratorPermissions(false)) {
+					throw new UnexpectedValueException('Not allowed to mention room');
```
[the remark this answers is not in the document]
blizzz: We agreed that we do not want a person be able to spam others with mentions. Now I had the idea that some trusted individuals it might be allowed to trigger mentions to  a crowd, and in Talk they would be identified by Moderators. If it does not make sense, we can kill it. Or extend/relax it later.

src/PostToConversation.vue
blizzz: How to you check room.readOnly in other places? Like a bool?

src/PostToConversation.vue
blizzz: I'll have a look, thx for the pointer

src/flow.js
blizzz: sure

src/PostToConversation.vue
nickvergessen:
```suggestion
					return room.readOnly === CONVERSATION.STATE.READ_WRITE
```

after you did this right after `<script>`:
```
import { CONVERSATION } from '../../constants'
```

lib/Flow/Operation.php
blizzz: This was still to add.

src/PostToConversation.vue
blizzz: Only with refactoring the RoomSelector as it brings a hard model with it. I'd refrain of it for now.

lib/Flow/Operation.php
```diff
@@ -0,0 +1,244 @@
...
+
+				$entity = $ruleMatcher->getEntity();
+
+				$message = $eventName;
+				if($entity instanceof IDisplayText) {
+					$message = $entity->getDisplayText(3);
+					if($message === '') {
+						continue;
+					}
+				}
```
juliushaertl: Adding a link if IUrl is implemented would be nice as well

blizzz: Yeah, makes sense. Then, only when some other text is already given, since a single url alone is too strange in a conversation

PR #2499, opened 2019-11-26
Feature/new group conversation

ma12-co:
## new New-group-conversation feature

previously we had to create group conversations with the drop-down options in the left sidebar and then add the users manually one by one in the right sidebar. Now we're going to have this '3 page modal' with navigation. All the requests are made once the user has finished setting things up and all at once (everything responds immediatyely as there's no lag between steps).

#### Todo: 

- [x] Restructure and rename to prepare for new components;
- [x] Add trigger action and modal components;
- [x] Create Modal content component;
- [x] Add a form component that allows to set the conversation name; 
- [x] Add navigation;
- [x] Add SearchBox and search results to modal content;
- [x] Add ability to select contacts and groups to be added;
- [x] Add toggle to make the group public on the first page;
- [x] Compute a front-tail whitespace trimmed version of the title and use that to whether check if the 
bad input hint has to be displayed or not;

- [x] Do all the requests;
- [x] Add third page to display loading while requests are being made and a:
* Final confirmation with share link (only if checkbox checked in 1st page);
* Errors while creating the conv if any;
* Otherwise just close the modal and -->
- [x] Push route to the new conv  when created;
- [x] Remove old way of creating groups;
- [ ] Add docs :books: 
- [ ] Make it pretty :nail_care: 
- [ ] Add transitions :tada: 

#### Improvements: 
- [ ] Modal size: fixed on desktop and fullscreen on mobile;
- [ ] Check in the store if the conversation name is already  used an warn the user if so;
- [x] Make the checkbox actually check-able;
- [x] Add checkmarks to the right of selected participants;
- [ ] Selected Participant bg color: use the same as active appnavigationitem  
- [ ] Tell user what the errors are about;
- [x] fix navigation to the bottom (the participants list is pushing down the navigation)
- [ ] display proper icons for groups and circles
- [x] component naming, folder structure

Peek 2019-12-09 12-54


ma12-co:
There's still quite a bit to refine, but the big chunk of it is done and I'd like to hear some early feedback :)

jancborchardt:
Nice! :)

Early feedback, just looking at the screenshots:
- The "Please wait" slide is a bit off. "Please wait" is never nice feedback :D Can we just fake create the conversation and throw any errors if they appear?
- Same for the "All set" – this is unnecessary, simply show the conversation

src/components/LeftSidebar/NewGroupConversation/NewGroupConversation.vue
```diff
@@ -0,0 +1,288 @@
...
+			console.log(event)
+			this.checked = event
+		},
+
+		handleClickForward() {
+			if (this.page === 0) {
+				if (this.conversationName !== '') {
+					this.page = 1
+					this.hint = ''
+				} else if (this.conversationName === '') {
```
nickvergessen:
```suggestion
				} else {
```

src/components/LeftSidebar/NewGroupConversation/SetConversationType/SetConversationType.vue
```diff
@@ -0,0 +1,77 @@
...
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+-->
+
+<template>
+	<div class="conversation-type">
+		<input
+			type="checkbox"
+			:checked="value"
+			@change="handleInput" />
+		<label class="conversation-type__label">{{ t('spreed', 'Make ') }}<strong>{{ conversationDisplayName }}</strong>{{ t('spreed', ' public') }}</label>
```
nickvergessen: This breaks translations. Assign the following to a computed and use that:

```
 t('spreed', 'Make {name} public', { name: this.conversationDisplayName)
```

src/components/LeftSidebar/NewGroupConversation/SetConversationType/SetConversationType.vue
```diff
@@ -0,0 +1,77 @@
...
+export default {
+	name: 'SetConversationType',
+	props: {
+		value: {
+			type: Boolean,
+			required: true
+		},
+		conversationName: {
+			type: String,
+			default: t('spreed', 'your conversation'),
```
nickvergessen: Uppercase start

src/components/LeftSidebar/NewGroupConversation/SetContacts/SetContacts.vue
```diff
@@ -0,0 +1,126 @@
...
+	<div class="set-contacts">
+		<input
+			v-model="searchText"
+			class="set-contacts__input"
+			type="text"
+			autofocus
+			:placeholder="t('spreed', `Search Participants`)"
+			@input="handleInput">
+		<Caption
+			:title="t('spreed', `Add participants to ${conversationName}`)" />
```
nickvergessen: See https://github.com/nextcloud/spreed/pull/2499/files#r355394368
This will never translate

src/components/LeftSidebar/NewGroupConversation/SetConversationType/SetConversationType.vue
nickvergessen: Also use a full string for the empty name and do the switch in the computed:

```js
if (this.conversationName === '') {
    return  t('spreed', 'Make your conversation public')
}
return  t('spreed', 'Make {name} public', { name: this.conversationName })

**Comment by ma12-co** [comment]
>The "Please wait" slide is a bit off. "Please wait" is never nice feedback :D Can we just fake create the conversation and throw any errors if they appear?

@jancborchardt  all the API calls are going to be made at the same time, and there could be many of them so the chances of having a bit of waiting time are high. I know it's a bit of a pain, but on the other hand immediately closing the dial and creating a fake conversation might be super confusing if something goes wrong with the requests. 

>Same for the "All set" – this is unnecessary, simply show the conversation

I introduced this 'view' because I'd like to display it the conversation link and a quick copy button in case you select the option of making the conversation public. If you don't select that option, you just see it for a split second while the UI underneath loads the new conversation. I don't think it's strictly necessary in this case but to me it's a nice touch :) 

ps: removed screenshots and added gif to showcase this better!

**Comment by jancborchardt** [comment]
> I introduced this 'view' because I'd like to display it the conversation link and a quick copy button in case you select the option of making the conversation public. If you don't select that option, you just see it for a split second while the UI underneath loads the new conversation. I don't think it's strictly necessary in this case but to me it's a nice touch :)

It seems a bit off and broken when it’s only showing for a split second and you can’t even read it. :) Just get rid of it, much simpler, and others don’t have this either.


-------

- [ ] From the GIF – I’m sure you noticed this yourself – the fields need to be autofocused when going to the next step. :)
- [ ] The checkbox needs to use the proper style
- [ ] Instead of "Next" the button label would be better with "Add participants" so people know what’s next. On the participants screen, the label would then be "Create conversation".
- [ ] The buttons seem to be moved with the content?
- [ ] Test with more people in the list – when there are more, it should cut off the last entry half, so people know there’s more. Also a fade from transparent to white behind the buttons, just like in the "Move or copy" dialog
- [ ] Round the corners of the dialog a bit with `var(--border-radius)` :)

**Comment by ma12-co** [comment]
Call with @jenniferpiperek : 

- [ ] 'Add participants to a looks like a button ---> use 'Select participants'
- [ ] show it only once u type something
- [ ] Tooltip on the plus button,
- [ ] Password protect for public convs? 
- [ ] 'Choose' or 'select' participants in the first page button

File: src/components/LeftSidebar/NewGroupConversation/SetConversationType/SetConversationType.vue (hunk 2 above)
Comment (ma12-co) [reply src/components/LeftSidebar/NewGroupConversation/SetConversationType/SetConversationType.vue]: Can I have this bold as it is right now @nickvergessen ?

File: src/components/LeftSidebar/NewGroupConversation/SetConversationType/SetConversationType.vue (hunk 3 above)
Comment (ma12-co) [reply src/components/LeftSidebar/NewGroupConversation/SetConversationType/SetConversationType.vue]: What should be uppercase??

File: src/components/LeftSidebar/NewGroupConversation/SetContacts/SetContacts.vue (hunk 4 above)
Comment (ma12-co) [reply src/components/LeftSidebar/NewGroupConversation/SetContacts/SetContacts.vue]: This has been dropped after a design discussion with Jenni

**Comment by karlitschek** [comment]
Very cool!

**Comment by karlitschek** [comment]
Hell freezes over. @jancborchardt likes modal dialogs ;-)

**Comment by jancborchardt** [comment]
> Hell freezes over. @jancborchardt likes modal dialogs ;-)

Never said I like it ;) Just taking a few steps back – user research will always point out if it works or not.

File: src/components/LeftSidebar/NewGroupConversation/SetConversationType/SetConversationType.vue (hunk 2 above)
Comment (nickvergessen) [reply src/components/LeftSidebar/NewGroupConversation/SetConversationType/SetConversationType.vue]:
```js
if (this.conversationName === '') {
    return  t('spreed', 'Make your conversation public')
}
return  t('spreed', 'Make {name} public', { name: '<strong>' + escapeHTML(this.conversationName) + '</strong>' }, undefined, { escape: false })
```

src/components/LeftSidebar/NewGroupConversation/SetConversationType/SetConversationType.vue
nickvergessen:
```suggestion
			default: t('spreed', 'Your conversation'),
```

src/components/LeftSidebar/NewGroupConversation/SetConversationType/SetConversationType.vue
```diff
@@ -0,0 +1,77 @@
...
+		},
+	},
+	computed: {
+		conversationDisplayName() {
+			if (this.conversationName === '') {
+				return t('spreed', 'your conversation')
```
nickvergessen:
```suggestion
				return t('spreed', 'Your conversation')
```

src/components/LeftSidebar/NewGroupConversation/SetConversationType/SetConversationType.vue
ma12-co: Oh that is cool!

src/components/LeftSidebar/NewGroupConversation/SetContacts/SetContacts.vue
```diff
@@ -0,0 +1,126 @@
...
+-->
+
+<template>
+	<div class="set-contacts">
+		<input
+			v-model="searchText"
+			class="set-contacts__input"
+			type="text"
+			autofocus
+			:placeholder="t('spreed', `Search Participants`)"
```
nickvergessen:
```suggestion
			:placeholder="t('spreed', 'Search Participants')"
```

src/components/LeftSidebar/NewGroupConversation/SetConversationType/SetConversationType.vue
ma12-co: Nope, that is a placeholder inline text

src/components/LeftSidebar/NewGroupConversation/Confirmation/Confirmation.vue
```diff
@@ -0,0 +1,88 @@
...
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+-->
+
+<template>
+	<div class="confirmation">
+		<template v-if="isLoading && !error">
+			<template v-if="!success">
+				<div class="icon-loading confirmation__icon" />
+				<p class="confirmation__warning">
+					{{ t('spreed', 'Please wait') }}
```
nickvergessen: Maybe the following is better
```suggestion
					{{ t('spreed', 'Creating your conversation, please wait') }}
```
Or maybe even without the "please wait"

src/components/LeftSidebar/NewGroupConversation/SetConversationType/SetConversationType.vue
skjnldsv: You will need to use v-html though

src/components/LeftSidebar/NewGroupConversation/SetConversationType/SetConversationType.vue
nickvergessen: I knew it ;) , and it was an indirect hint that this will not work with translations in all languages. If it is not a set name, dont use a it inline, but use a full string

ma12-co:
Call with Jan: 

1)  
- [ ] style as regular input;
- [ ] conversation name only as a placeholder;
- [ ] Disable button instead of displaying warning if no name;
- [ ] 'Allow guests to join via link';
- [ ] remove text;

2)
- [ ] style as regular input;
- [ ] search icon;
- [ ] empty content view: 
icon-contacts 
line below: search participants 
- [ ] inputs capitalise only first word
- [ ] Conv name should stay the same
position relative on main container new-group-conversation
position abs on navigation and search top 0 bottom 0

PR #2500, opened 2019-11-26
Properly focus of input fields

ma12-co:

### DIFF
--- a/src/components/LeftSidebar/NewConversation/SearchBox.vue
+++ b/src/components/LeftSidebar/NewConversation/SearchBox.vue
@@ -24,26 +24,20 @@
 		class="app-navigation-search"
 		@submit.prevent="handleSubmit">
 		<input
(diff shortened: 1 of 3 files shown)


src/components/LeftSidebar/NewConversation/SearchBox.vue
```diff
@@ -27,8 +27,8 @@
 			v-model="localValue"
 			class="app-navigation-search__input"
 			type="text"
-			:autofocus="autoFocused"
-			:placeHolder="placeholderText">
+			:placeHolder="placeholderText"
+			v-focus>
```
nickvergessen: Actually this should not be focused, but the NewMessageForm (at least when it is displayed)

ma12-co: Yep, that's the reason I started this pr

PR #2502, opened 2019-11-27
Typed events

nickvergessen:
Reduces the log spam with newest 18 due to old Event dispatcher

- [x] Manager events
- [x] Room events
    - [x] *UserDisconnectRoom
- [x] ChatManager events
- [x] GuestManager events
- [x] Message parsing event
- [x] Command execution event
- [x] SignalingController event
- [x] RoomController event
- [ ] ~Room Share provider event~ - Can not be done because share events are also still GenericEvents


docs/events.md
```diff
@@ -0,0 +1,147 @@
+# PHP Events
+
+## Conversation related events
+
+### Created conversation
+
+* Event class: `OCA\Talk\Events\RoomEvent`
+* Event name: `OCA\Talk\Room::createdRoom`
```
ChristophWurst: IMO it should be `roomCreated`. That sounds more like a description of an event. `createdRoom` is the action.

docs/events.md
```diff
@@ -0,0 +1,147 @@
...
+* Pre-event name: `OCA\Talk\Room::preDeleteRoom`
+* Post-event name: `OCA\Talk\Room::postDeleteRoom`
+
+## Participant related events
+
+### Add participants
+
+* Event class: `OCA\Talk\Events\AddParticipantsEvent`
+* Pre-event name: `OCA\Talk\Room::preAddUsers`
+* Post-event name: `OCA\Talk\Room::postAddUsers`
```
ChristophWurst: For consistency with legacy hooks and events I'm fine with *pre* and *post* but we could also name these something like *BeforeUserAdded* and *AfterUserAdded*.

lib/Chat/ChatManager.php
```diff
@@ -154,11 +153,8 @@ public function sendMessage(Room $chat, Participant $participant, string $actorT
 			$comment->setParentId($replyTo->getId());
 		}
 
-		$this->dispatcher->dispatch(self::class . '::preSendMessage', new GenericEvent($chat, [
-			'comment' => $comment,
-			'room' => $chat,
-			'participant' => $participant,
-		]));
+		$event = new ChatParticipantEvent($chat, $comment, $participant);
+		$this->dispatcher->dispatch(self::class . '::preSendMessage', $event);
```
ChristophWurst: To prevent typos and make refactoring easier you could add a static method to the event classes and use it like

```php
$this->dispatcher->dispatch(ChatParticipantEvent::preSendMessageEventName(), $event);
```

(function instead of constant so you can use ``self::class``)

ChristophWurst: Oh you can actually use a constant. `public const EVENT_NAME = self::class . '...';` works just fine :)

docs/events.md
nickvergessen: Makes sense if that is the new guide line.
Since we change the event class anyway it will break with all stuff.
But at least you knew it broke, but well so far the events were not publicly announced anyway

ChristophWurst: It was more of a spontaneous idea but after a few days I still like it. I'll try to find some time to write down some guidelines for the events in our documentation.

nickvergessen:
Updated

PR #2514, opened 2019-12-03
Fix sharing integration tests and loading shares

nickvergessen:
* User and group shares are not accepted before assuming they work
* The accept state was not set on the object after loading a share from the database


nickvergessen:
Nice, finally working again (apart from timeouts)

PR #2518, opened 2019-12-05
Add support for circles

nickvergessen:
1. Install circles: https://github.com/nextcloud/circles/pull/361
2. Create a circle
3. Try to create a conversation with a circle
4. Try to add a circle to an existing conversation

### DIFF
--- a/lib/Controller/RoomController.php
+++ b/lib/Controller/RoomController.php
@@ -26,6 +26,8 @@ declare(strict_types=1);
 
 namespace OCA\Talk\Controller;
 
+use OCA\Circles\Api\v1\Circles;
+use OCA\Circles\Model\Member;
 use OCA\Talk\Chat\ChatManager;
 use OCA\Talk\Chat\MessageParser;
 use OCA\Talk\Events\UserEvent;
@@ -39,6 +41,7 @@ use OCA\Talk\Participant;
 use OCA\Talk\Room;
 use OCA\Talk\TalkSession;
 use OCA\Talk\Webinary;
+use OCP\App\IAppManager;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\Utility\ITimeFactory;
@@ -57,6 +60,8 @@ class RoomController extends AEnvironmentAwareController {
 
 	/** @var string|null */
 	private $userId;
+	/** @var IAppManager */
+	private $appManager;
 	/** @var TalkSession */
 	private $session;
 	/** @var IUserManager */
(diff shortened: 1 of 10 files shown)


nickvergessen:
Rebased due to conflict in RoomController (constant for #2502 was added in same place as a member vairable)

nickvergessen:
>  Circles support is coming. you just request them additionally on the endpoint when asking for results:
https://github.com/nextcloud/spreed/pull/2518/files#diff-fced531870aa624bf64ba668bff611c5R58

>  and then a request similar to group conversations is sent, just including `&source=circles in the url
https://github.com/nextcloud/spreed/pull/2518/files#diff-fced531870aa624bf64ba668bff611c5R85

>  in the end we just add the users of the circle to the conversation, like we do with groups
no membership tracking, etc

lib/Controller/RoomController.php
```diff
@@ -659,6 +724,40 @@ public function addParticipantToRoom(string $newParticipant, string $source = 'u
...
+			try {
+				$circle = Circles::detailsCircle($newParticipant);
+			} catch (\Exception $e) {
+				return new DataResponse([], Http::STATUS_NOT_FOUND);
+			}
+
+			$participants = [];
+			foreach ($circle->getMembers() as $member) {
+				/** @var Member $member */
+				if ($member->getType() === Member::TYPE_USER || $member->getUserId() === '') {
```
daita: I would add a check on the Level: $member->isLevel(MEMBER::LEVEL_MEMBER) as in current state, non-yet members (invited or requesting for access) will be invited to the room.

lib/Controller/RoomController.php
```diff
@@ -456,6 +467,60 @@ protected function createGroupRoom(string $targetGroupName): DataResponse {
...
+		$room = $this->manager->createGroupRoom($circle->getName());
+		$room->addUsers([
+			'userId' => $currentUser->getUID(),
+			'participantType' => Participant::OWNER,
+		]);
+
+		$participants = [];
+		foreach ($circle->getMembers() as $member) {
+			/** @var Member $member */
+			if ($member->getType() === Member::TYPE_USER || $member->getUserId() === '') {
```
daita: I would add a check on the Level: $member->isLevel(MEMBER::LEVEL_MEMBER) as in current state, non-yet members (invited or requesting for access) will be invited to the room.
@daita

lib/Controller/RoomController.php
nickvergessen: Invites would be fine I guess. requests is bad of course :D

daita: in that case, have a look to ->getStatus() that should returns a string related to https://github.com/nextcloud/circles/blob/12934e8c5c886401231fb5808022891824250bc7/lib/Model/BaseMember.php#L41-L46 :]

nickvergessen:
@daita another review?

daita:
Code looks good to me

PR #2545, opened 2019-12-10
Prevent unload when in a call

### LINKED ISSUE #1175: Prevent accidentally leaving call
## Steps to reproduce
1. Be in a call
2. Forget that clicking anywhere else (other conversation in the list, other app in header) will drop the call
3. Click on another conversation to read messages. ;)

### Expected behaviour
You should continue to be in the call.

### Actual behaviour
You are dropped from the call and need to rejoin.

## There are several ways we could mitigate this problem:

1. Automatic: As soon as you are in a call, clicking on anything which would drop the call or switch the view (which leads to dropping) could automatically open in a new tab. This is the most automatic way, but also unexpected.
2. Confirmation: As soon as you are in a call, clicking on anything which would drop the call or switch the view (which leads to dropping) could ask people if they want to leave the call. This is quite annoying as popups are rarely read. When you click "OK" out of habit and then the call drops it’s even more annoying.
3. Keeping call active in background: This would be possible somewhat when staying in the Talk app, but is definitely more difficult when you navigate in a way which forces a full refresh. Future stuff.

I’d say that measure 1 (opening these call-dropping navigation elements in a new tab during an active call) is the best short-term solution. Then in the future when at some point we have 3 implemented, we can drop that.

What do you think @nickvergessen @Ivansss?

nickvergessen:
Fix #1175

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11591,6 +11591,11 @@
 			"resolved": "https://registry.npmjs.org/vue-observe-visibility/-/vue-observe-visibility-0.4.6.tgz",
 			"integrity": "sha512-xo0CEVdkjSjhJoDdLSvoZoQrw/H2BlzB5jrCBKGZNXN2zdZgMuZ9BKrxXDjNP2AxlcCoKc8OahI3F3r3JGLv2Q=="
 		},
+		"vue-prevent-unload": {
+			"version": "0.2.3",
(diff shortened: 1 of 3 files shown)


jancborchardt:
Which wording does the popover use, the default of vue-prevent-unload? That is a bit generic, so if possible it would be better to use something like this:

> **Leave call?**
> If you close the page, you will leave the call as well.
> [ Stay on page ]   [ Leave page ]

nickvergessen:
Sorry, but all browsers removed custom text support (apart from IE11)

PR #2550, opened 2019-12-13
Bugfix/noissue/fix left sidebar layout

ma12-co:
based on  #2499


nickvergessen:
Lets merge it like is and fix the create-conversation dialog in a follow up? @ma12-co

PR #2593, opened 2019-12-17
add to Flow category

blizzz:
since #2487  is merged, Talk should also be enlisted in the new [Flow category](https://apps.nextcloud.com/categories/workflow) on the appstore.

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -34,6 +34,7 @@ And in the works for the [coming versions](https://github.com/nextcloud/spreed/m
 
 	<category>multimedia</category>
 	<category>social</category>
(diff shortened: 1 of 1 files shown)


jancborchardt:
I thought the "Flow" category is for actual specific Flows which people wrote, like the "Explore" list of If This Then That: https://ifttt.com/discover

So if an app supports Flow, their specific flows would be listed, but not just the app itself which is not so informative.

cc @karlitschek @rullzer @juliushaertl for clarification.

rullzer:
@jancborchardt we have no way to do that. Our appstore can only list apps. So either talk shows up there or not.

PR #2606, opened 2019-12-19
Disable tooltip and menu in the participant component

### LINKED ISSUE #2568: Three dots item in sidebar seems always selected
## Steps to reproduce
1. Install Spreed master
2. Open a conversation
3. Take a look at the sidebar - three dots item seems to always be selected

### Expected behaviour
It shouldn't look selected if it's not!

### LINKED ISSUE #2575: Avatar menu on a participant in sidebar menu is broken
## Steps to reproduce 
1. Install Spreed master
2. Create a conversation
3. Open sidebar and click on a three dot menu of another participant
4. See that the menu is broken

### Expected behaviour
The menu should not be broken.

Browser: Latest Firefox on MacOS

Screenshot 2019-12-14 at 22 48 15

ma12-co:
fix #2575

No need for a menu or tooltip in the `Participant` component's avatar.

### DIFF
--- a/src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
+++ b/src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
@@ -30,7 +30,9 @@
 		<div class="participant-row__avatar-wrapper">
 			<Avatar
 				:user="computedId"
-				:display-name="computedName" />
+				:display-name="computedName"
+				:disableMenu="true"
+				:disableTooltip="true" />
 		</div>
 
 		<span class="participant-row__user-name">{{ computedName }}</span>


ma12-co:
Starting a new conversation Is the only thing that's there now and since for participants we have actions on the right, I think every action should go there. This menu should be the last resort imo, as it's only visible on hover and it's invisible on mobile.

ma12-co:
for reference: 

Screenshot from 2019-12-19 12-36-51

skjnldsv:
I agree with Daniel. As a UX pov, users are getting more and more used to managing contacts with this menu, removing it is counter intuitive for the user I think.

ma12-co:
> I agree with Daniel. As a UX pov, users are getting more and more used to managing contacts with this menu, removing it is counter intuitive for the user I think.

People are more used to managing contacts with a visible menu than with an invisible one :)  
I get your point and I would agree if we wouldn't have space for the actions on the right, but we happen to have them anyway. Imo it's just confusing to have 2 actions menus for the same element: the participant. That's why I would go with 1 menu only, and the one that's always visible is a clear winner for me..

Plus, avatars in the `Participant` component ant it's clickable area are only 32x32 px, unlike in the conversations list.

skjnldsv:
> Imo it's just confusing to have 2 actions menus for the same element: the participant. That's why I would go with 1 menu only, and the one that's always visible is a clear winner for me..

yes, that is a fair point.
But that require far more work for you to implement your own contacts menu action handler in the right menu than just showing both. And I'm not sure y'all got the time for it :see_no_evil: 

I think it would make sense to have a vue component for this. Something like a `ContactActions`. That can be also used by the Avatar component (let's not waste time)

But in the meantime, keep it simple :)

ma12-co:
We could keep both menus and move all talk-related actions, like 'talk with ...' to the always visible menu?

nickvergessen:
@ma12-co The "Talk with " needs to stay in that menu as it is the same in every user avatar. Also note that the `...` menu of the participant list is only available for moderators.

Also since you asked about it:
If you only see Talk in the avatar popover, add an email address to all your users and you see 1 more option ;)

nickvergessen:
So closing this.

PR #2624, opened 2019-12-20
Set the correct placeholder avatar

nickvergessen:
Before the connection is established the avatar still shows the `a` for me.
After a connection is established it correctly shows `H`for the displayname now.

Any idea where the first wrong avatar is loaded @danxuliu ?


nickvergessen:
This was still developing with an unsolved comment 😅

ma12-co:
:see_no_evil:

PR #2649, opened 2019-12-29
Fix message placeholder, add @-mention hint

jancborchardt:
Please review @ma12-co @nextcloud/designers-talk :) 

More helpful message, also pointing out the possibility of @-mention to people who don’t know about this.


src/components/NewMessageForm/AdvancedInput/AdvancedInput.vue
```diff
@@ -38,7 +38,7 @@ export default {
 		 */
 		placeholderText: {
 			type: String,
-			default: 'New Message ...',
+			default: 'Write message, @ to mention someone …',
```
nickvergessen: Död...

`t('spreed', '....')`

PR #2653, opened 2019-12-30
Update Talk action to Nextcloud blue

jancborchardt:
As discussed @juliushaertl 

Not tested just quickly changed. Nextcloud Talk should of course have the Nextcloud color (not themed), not a random "tomato" color. ;)

### DIFF
--- a/src/flow.js
+++ b/src/flow.js
@@ -24,7 +24,7 @@ import FlowPostToConversation from './views/FlowPostToConversation'
 
 window.OCA.WorkflowEngine.registerOperator({
 	id: 'OCA\\Talk\\Flow\\Operation',
-	color: 'tomato',
+	color: '#0082c9',
 	operation: '',
 	options: FlowPostToConversation,
 })


blizzz:
if all actions have the same color, colors are not helpful anymore to identify the (configured) flows. we can drop this all along.

jancborchardt:
> if all actions have the same color, colors are not helpful anymore to identify the (configured) flows. we can drop this all along.

They don’t have all the same color. The reason why I changed this around is because Talk was red and PDF was blue – that’s just switched.

blizzz:
> DCO though!

we can ignore it in this less-than-one-liner

jancborchardt:
> > if all actions have the same color, colors are not helpful anymore to identify the (configured) flows. we can drop this all along.
> 
> They don’t have all the same color. The reason why I changed this around is because Talk was red and PDF was blue – that’s just switched.

Btw @blizzz, could you make sure the Flow action for PDF uses the red color we also use for the PDF filetype icon? It’s `#dc5047` – thanks!

PR #2654, opened 2019-12-31
Remove bottom border for search result category captions

jancborchardt:
Before, the separators actually made it look like the headings belong to the elements _above_ them, which is of course not the case. Simple whitespace separation is enough here, so we can remove the borders. :)

Also looks much more like Nextcloud style, light and not so boxy. Please review @nextcloud/designers-talk 

search results before search results after

### DIFF
--- a/src/components/Caption.vue
+++ b/src/components/Caption.vue
@@ -47,7 +47,6 @@ export default {
 		box-shadow: none !important;
 		user-select: none;
 		pointer-events: none;
-		border-bottom: 1px solid var(--color-border-dark);
 		padding-left: 10px;
 
 		&:not(:first-child) {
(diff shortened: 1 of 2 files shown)


nickvergessen:
> "looks good to me!"

Not to me. I much more prefer having the bottom border.
But well

jancborchardt:
> Not to me. I much more prefer having the bottom border.

What’s the reason for that? If it’s separation, then we should have more whitespace above the search result category captions. But not a separation between the caption and the contents of it. :)

nickvergessen:
For me the color + underline makes it pretty clear that this is another level of message and what actually should have been fixed is the left/right padding on the border so it does not go over the full width of the navigation anymore, which is a regression from some weeks ago (see issue about missing padding of "No search results")

PR #2673, opened 2020-01-06
Small fixes on the ConversationIcon design

skjnldsv:
- Fix alignment and proper sizing of the elements
- Reduced star size for better antialiasing

| Before | After |
|:---------:|:------:|
|Kazam_screenshot_00001|Kazam_screenshot_00000|

### DIFF
--- a/src/components/ConversationIcon.vue
+++ b/src/components/ConversationIcon.vue
@@ -25,7 +25,7 @@
 			class="avatar icon"
 			:class="iconClass" />
 		<Avatar v-else
-			:size="40"
+			:size="44"
 			:user="item.name"
 			:display-name="item.displayName"
 			class="conversation-icon__avatar" />
@@ -94,11 +94,6 @@ export default {
 	width: 44px;
 	height: 44px;
 
-	&__avatar {
-		// we request 40px avatars, but
-		// conversation icons are 44px
-		margin: 2px;
-	}
 	.icon:not(.icon-favorite) {
 		width: 44px;
 		height: 44px;
--- a/src/components/ConversationIcon.vue
+++ b/src/components/ConversationIcon.vue
@@ -27,7 +27,8 @@
 		<Avatar v-else
 			:size="40"
 			:user="item.name"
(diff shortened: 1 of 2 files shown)


nickvergessen:
`3. to review` or `Draft PR`, you can not have both!?

src/components/ConversationIcon.vue
```diff
@@ -90,41 +91,45 @@ export default {
...
-		background-size: 44px;
-	}
-	&.icon-public,
-	&.icon-contacts,
-	&.icon-password,
-	&.icon-file,
-	&.icon-mail {
-		background-size: 20px;
+	&__avatar {
+		// we request 40px avatars, but
```
nickvergessen: we can also make everything 40 instead. I think this was not an intentional mismatch

skjnldsv: Well, both are good to me, I really have no preference, but you need to choose :)

skjnldsv:
> `3. to review` or `Draft PR`, you can not have both!?

Wrong button, my bad

src/components/ConversationIcon.vue
nickvergessen: Let's go for 40 then

skjnldsv: Well, minimum clickeable area on nextcloud is 44px. SO I'd rather have everything 44.
What do you prefer @jancborchardt ?

skjnldsv: But visually it looks cleaner with 40px

| 40 | 44 |
|:---------:|:------:|
|dev skjnldsv com_call_s934fu82|dev skjnldsv com_call_s934fu82 (1)|

ma12-co: Maybe a 44 by 44 clickable square wrapping a 40px avatar?

skjnldsv: The avatar component doesn't support it, let's not over complicate things ^^

ma12-co: Then I'd also prefer 44px!

PR #2695, opened 2020-01-08
Make all text bold for unread conversations

### LINKED ISSUE #2657: Bolding for unread messages
Currently the only way to identify unread messages is the little counter in the list. That’s not very distinguished and can be overlooked.
vertical issues

As other messengers are doing as well (and email apps too, including Nextcloud Mail), conversations with unread messages should be displayed with bold text in the navigation.

This is a regression in some ways because the previous counter was more obvious. cc @nextcloud/designers-talk

nickvergessen:
Fix #2657 

Bildschirmfoto von 2020-01-08 12-05-53


ma12-co:
oops, i double-reviewed this :sweat_smile:

Issue #2677, opened 2020-01-06
email invitations no longer working


wiswedel:
## Steps to reproduce
1. Create a public conversation
2. Head to the _Participants_ tab
3. Enter an email address and hit enter

### Expected behaviour
The email recipient gets invited to the public link via email.

### Actual behaviour
Nothing happens even on hitting enter (step 3)

PR #2698, opened 2020-01-08
Bugfix/noid/combine empty sources

### LINKED ISSUE #2574: "No search results" is missing padding
## Steps to reproduce
1. Install Spreed master
2. Search for "admin"
3. See that "No search results" misses padding

### Expected behaviour
All entries, even "No search results" should have padding.

Screenshot 2019-12-14 at 22 44 14

nickvergessen:
Before | After with some matches | After with no matches
---|---|---
Bildschirmfoto von 2020-01-08 12-10-27 | Bildschirmfoto von 2020-01-08 16-30-01 | Bildschirmfoto von 2020-01-08 16-29-29

cc @nextcloud/designers-talk 
Should we use "Other sources" instead of listing them?
The problem is the number of sources is getting bigger and bigger, currently we have:
* Contacts (local users)
* Groups
* Emails
* Circles

We could also use "Other sources" or something like that. But then it's not clear which values are checked.

Pre-requirement for #2677 

Fix #2574

### DIFF
--- a/src/components/Hint.vue
+++ b/src/components/Hint.vue
@@ -47,5 +47,6 @@ export default {
 		box-shadow: none !important;
 		user-select: none;
 		pointer-events: none;
+		padding-left: 10px;
 	}
 </style>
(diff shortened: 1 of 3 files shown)


nickvergessen:
> As per grouping the sources: I don't think it's needed at all.

So "Other sources" or just keep the current listing of matching types which have no results?

ma12-co:
> So "Other sources" or just keep the current listing of matching types which have no results?

Keep the current listing, sorry for the ambiguity

PR #2737, opened 2020-01-14
Fix shares done via talk

### LINKED ISSUE #2733: Wrong message when sharing folders
### Steps to reproduce
1) We experienced this in our cloud (external signaling), we cannot reproduce with internal signaling (might or might not be related to this).
2) Share a folder via the talk file-sharing button

### Expected behavior
You should see the folder in your chat

### Actual behavior
* You get a message that says "Shared a file which is no longer available" as if the share was unsuccessful
* Upon reloading the share is available in the chat;

nickvergessen:
Fix #2733

> ### Re-setup the filesystem in case a share message failed
> 
> This happens when your request is already waiting for new chat messages.
> The filesystem is already set up, so the share will not be discovered
> unless the mount points are initialised freshly

### DIFF
--- a/lib/Chat/Parser/SystemMessage.php
+++ b/lib/Chat/Parser/SystemMessage.php
@@ -269,6 +269,16 @@ class SystemMessage {
 				$userFolder = $this->rootFolder->getUserFolder($participant->getUser());
 				if ($userFolder instanceof Node) {
 					$userNodes = $userFolder->getById($node->getId());
+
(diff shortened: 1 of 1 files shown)


nickvergessen:
@icewind1991 do you have any idea how we could do this better? Like is it possible to force a refresh of a single mount point or something like this?

nickvergessen:
Ready for review as a temporary solution for the release.

PR #2770, opened 2020-01-16
Add copy link button

### LINKED ISSUE #2540: Not possible to copy share link for existing shared conversations
Display copy link actionbutton under share link in the sidebar's actions

ma12-co:
Fix #2540

### DIFF
--- a/src/components/RightSidebar/RightSidebar.vue
+++ b/src/components/RightSidebar/RightSidebar.vue
@@ -49,6 +49,13 @@
 				@change="toggleGuests">
 				{{ t('spreed', 'Share link') }}
 			</ActionCheckbox>
+			<ActionButton
+				v-clipboard:copy="linkToConversation"
+				icon="icon-clippy"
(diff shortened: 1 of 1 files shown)


nickvergessen:
Good enough for now, but in the ideal world, it would be immediately visible?

src/components/RightSidebar/RightSidebar.vue
```diff
@@ -313,6 +321,11 @@ export default {
 		isPasswordProtected() {
 			return this.$store.getters.conversations[this.token].hasPassword
 		},
+		linkToConversation() {
+			if (this.token !== '') {
+				return window.location.protocol + '//' + window.location.host + generateUrl('/call/' + this.token)
+			} else return ''
```
nickvergessen: ...

nickvergessen:
```suggestion
			} else {
				return ''
			}
```

PR #2783, opened 2020-01-17
Fix redirect on "Talk to …" links

### LINKED ISSUE #2777: "Talk to …" option in contacs menu doesnt open conversation
It only creates it

nickvergessen:
Fix #2777

1. Open the contacts menu in the top
2. Click on the "Talk to …" option for a user

### Expected
Conversation opens

### Actually
You are redirected to `index.php/apps/spreed?token=…` which is not recognized by the router and therefor not opening the conversation


nickvergessen:
That is all a problem in the server. We can only specify the text, the icon and the link to redirect to (which was fixed)

PR #2815, opened 2020-01-21
Prevent escaping 😇 play with fire instead 🔥

nickvergessen:
See https://github.com/nextcloud/server/pull/18741


ma12-co:
What is this?

nickvergessen:
See the linked server issue, it would be a hack to fix html in the last message in the conversation list. But it should be fixed in the server, it's just a proof of concept.

jospoortvliet:
> What is this?

It is one of the best PR titles I've seen in a while, that is what it is 😂😂😂

PR #2824, opened 2020-01-22
Set no rewrite for generateUrl, in case URL contains index.php despite configured rewrite

Ivansss:
Signed-off-by: Ivan Sein <ivan@nextcloud.com>

### DIFF
--- a/src/router/router.js
+++ b/src/router/router.js
@@ -22,18 +22,24 @@
 
 import Vue from 'vue'
 import Router from 'vue-router'
-import { generateUrl } from '@nextcloud/router'
+import { getRootUrl, generateUrl } from '@nextcloud/router'
 import MainView from '../views/MainView.vue'
 import NotFoundView from '../views/NotFoundView.vue'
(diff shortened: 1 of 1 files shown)


src/router/router.js
```diff
@@ -22,18 +22,24 @@
...
+import { getRootUrl, generateUrl } from '@nextcloud/router'
 import MainView from '../views/MainView.vue'
 import NotFoundView from '../views/NotFoundView.vue'
 import WelcomeView from '../views/WelcomeView.vue'
 
 Vue.use(Router)
 
+const webRootWithIndexPHP = getRootUrl() + '/index.php'
+const doesURLContainIndexPHP = window.location.pathname.startsWith(webRootWithIndexPHP)
+const base = generateUrl('/', {}, {
```
georgehrke: shouldn't this contain '/apps/spreed' or at least '/s'?

nickvergessen: We use index.php/apps/spreed and index.php/call/...

So it needs to be empty

PR #2856, opened 2020-01-28
Pressing enter again re-sends last message

### LINKED ISSUE #2853: Pressing enter again after sending a message sends it again
## Steps to reproduce
1. Write a message
2. Hit enter so it gets sent
3. In the current room or a different one press enter again in the empty text field

### Expected behaviour
Nothing

### Actual behaviour
The last message gets sent again

nickvergessen:
Fix #2853

### DIFF
--- a/src/components/NewMessageForm/AdvancedInput/AdvancedInput.vue
+++ b/src/components/NewMessageForm/AdvancedInput/AdvancedInput.vue
@@ -181,7 +181,9 @@ export default {
 	},
 	watch: {
 		text(text) {
-			this.$emit('update:contentEditable', this.$refs.contentEditable.cloneNode(true))
+			this.$nextTick(() => {
+				this.$emit('update:contentEditable', this.$refs.contentEditable.cloneNode(true))
+			})
 
 			this.$emit('update:value', text)
 			this.$emit('input', text)
--- a/src/components/NewMessageForm/NewMessageForm.vue
+++ b/src/components/NewMessageForm/NewMessageForm.vue
@@ -220,6 +220,7 @@ export default {
 				const temporaryMessage = this.createTemporaryMessage()
 				this.$store.dispatch('addTemporaryMessage', temporaryMessage)
 				this.text = ''
+				this.parsedText = ''
 				// Scrolls the message list to the last added message
 				this.$nextTick(function() {
 					document.querySelector('.scroller').scrollTop = document.querySelector('.scroller').scrollHeight


src/components/NewMessageForm/AdvancedInput/AdvancedInput.vue
```diff
@@ -181,7 +181,9 @@ export default {
 	},
 	watch: {
 		text(text) {
-			this.$emit('update:contentEditable', this.$refs.contentEditable.cloneNode(true))
+			this.$nextTick(() => {
+				this.$emit('update:contentEditable', this.$refs.contentEditable.cloneNode(true))
```
juliushaertl: From reading the code it seems that the contentNode is used to make sure that the DOM element passed though to the parent component is not changed after the event has been emitted.

You could avoid the whole node cloning by moving the parsing logic from https://github.com/nextcloud/spreed/blob/e69740b372e17eec4541337b47baa262a5766510/src/components/NewMessageForm/NewMessageForm.vue#L100-L143 to the AdvancedInput component and just emit the plain text of the message with the update event.

juliushaertl: The reason why this only works with nextTick is that the text is not reset properly when sending the message in the AdvancedInput component. Therefore a update:contentEditable event i emitted with the old data before it gets reset by the NewMessageForm.

The following patch should fix the issue without using $nextTick:

```diff
+++ b/src/components/NewMessageForm/AdvancedInput/AdvancedInput.vue
@@ -267,6 +267,7 @@ export default {
                        // TODO: add support for CTRL+ENTER new line
                        if (!(event.shiftKey)) {
                                event.preventDefault()
+                               this.text = ''
                                this.$emit('submit', event)
                        }
                },
diff --git a/src/components/NewMessageForm/NewMessageForm.vue b/src/components/NewMessageForm/NewMessageForm.vue
index 0b95eebf..7b6b69e8 100644
--- a/src/components/NewMessageForm/NewMessageForm.vue
+++ b/src/components/NewMessageForm/NewMessageForm.vue
@@ -220,6 +220,7 @@ export default {
                                const temporaryMessage = this.createTemporaryMessage()
                                this.$store.dispatch('addTemporaryMessage', temporaryMessage)
                                this.text = ''
+                               this.parsedText = ''
                                // Scrolls the message list to the last added message
                                this.$nextTick(function() {
                                        document.querySelector('.scroller').scrollTop = document.querySelector('.scroller').scrollHeight

```

juliushaertl: Hm nevermind, it seems I'm missing something there with the above patch. :see_no_evil:

PR #2857, opened 2020-01-28
Improve new group conversation feature

### LINKED ISSUE #2546: Refine new_group_conversation feature
Complete unchecked todos here #2499

---

- [x] Add docs :books:
- [x] Make it pretty :nail_care:
- [x] Selected Participant bg color: use the same as active appnavigationitem

Jan: 
- [x] Test with more people in the list – when there are more, it should cut off the last entry half, so people know there’s more. Also a fade from transparent to white behind the buttons, just like in the "Move or copy" dialog
- [x] search icon in searchbox;

The dialog should no only show an emptycontent view and force you to search, but:

- [x] Show the first ~50 contacts/groups/etc. in the list so people see immediate feedback that content exists and in the ideal case can directly choose the people to start a conversation with.
- [x] While it loads these initial contacts, show a loading spinner and text below saying "Loading contacts …" (compare to the Contacts menu in the top right)
- [x] On the bottom of this list of 50, show an emptycontent element saying "Search for more contacts" which on tap focuses the search field.

This should be the same cross-platform, on the web, Android and iOS.

ma12-co:
fix #2546

### DIFF
--- a/src/components/LeftSidebar/NewGroupConversation/SetContacts/SetContacts.vue
+++ b/src/components/LeftSidebar/NewGroupConversation/SetContacts/SetContacts.vue
@@ -41,7 +41,8 @@
 				:no-results="noResults"
 				:items="searchResults"
 				:display-search-hint="!contactsLoading"
-				@updateSelectedParticipants="handleUpdateSelectedParticipants" />
+				@updateSelectedParticipants="handleUpdateSelectedParticipants"
+				@clickSearchHint="focusInput" />
 		</template>
 	</div>
 </template>
@@ -79,7 +80,7 @@ export default {
 
 	async mounted() {
 		// Focus the input field of the current component.
-		this.$refs.setContacts.focus()
+		this.focusInput()
 		// Perform a search with an empty string
 		await this.fetchSearchResults()
 		// Once the contacts are fetched, remove the spinner.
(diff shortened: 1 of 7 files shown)


src/components/LeftSidebar/NewGroupConversation/SetContacts/SetContacts.vue
```diff
@@ -73,20 +70,20 @@ export default {
 		return {
 			searchText: '',
 			searchResults: [],
-			contactsLoading: false,
+			// The loading state is true when the component is initialised as we perform a searc for 'contacts'
```
nickvergessen:
```suggestion
			// The loading state is true when the component is initialised as we perform a search for 'contacts'
```

nickvergessen:
Is this supposed to fix https://github.com/nextcloud/spreed/issues/2546 ?

nickvergessen:
Not sure if this all belongs here or should be fixed/mentioned elsewhere

- [ ] Password input is plain text instead of password
- [ ] Navigating back and forth or re-searching shows selected participants as unselected
- [ ] Pressing enter in the conversation name field does not continue to next screen

ma12-co:
> Is this supposed to fix #2546 ?

yes, that was it

nickvergessen:
How would you like placeholder items instead of a loading circle with "Loading contacts" :)

ma12-co:
> How would you like placeholder items instead of a loading circle with "Loading contacts" :)

In this case I like the spinner more, and here I'm also following Jan's spec. Let's wait for when he's back before changing it?

nickvergessen:
1. Create a conversation with password
2. Create a new conversation:
    * Enter name
    * Allow guests
    * Password is automatically ticked (not reset to default state from before the first creation)

Apart from that ready to merge

PR #2859, opened 2020-01-28
Avoid jump when avatar gets sticky

juliushaertl:
The sticky avatar currently jumps a bit since it already has enough spacing though the margin that is applied. Therefore setting the top to 0px makes sure that it is not jumping around when scrolling but just staying attached to the top.

Really neat design detail btw. :heart:

### DIFF
--- a/src/components/MessagesList/MessagesGroup/MessagesGroup.vue
+++ b/src/components/MessagesList/MessagesGroup/MessagesGroup.vue
@@ -181,7 +181,7 @@ export default {
 	flex-direction: column;
 	&__avatar {
 		position: sticky;
-		top: 44px;
+		top: 0;
 		height: 52px;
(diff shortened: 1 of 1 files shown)


src/components/MessagesList/MessagesGroup/MessagesGroup.vue
```diff
@@ -181,7 +181,7 @@ export default {
 	flex-direction: column;
 	&__avatar {
 		position: sticky;
-		top: 44px;
+		top: 0px;
```
nickvergessen:
```suggestion
		top: 0;
```

juliushaertl: Ah sure, fixed.

PR #2870, opened 2020-01-30
Improve chat loading

nickvergessen:
Fixed errors:
* Chat was initialised twice, the later request would cancel the first one and therefor slowing the loading down a lot
* Chat headers were ignored. So in case someone used 200 commands in a row, the chat would load infinitly for all other users


nickvergessen:
Backport was done in https://github.com/nextcloud/spreed/pull/2902

PR #2876, opened 2020-01-31
Receive shares in Talk/ folder

### LINKED ISSUE #1811: Improve the way we handle shared files
At the moment every file shared in a room is a normal file share for all room members. This means all files from all rooms will be stored in my Nextcloud root. If you are in multiple rooms and people start sharing some pictures this quickly become a huge mess in the users root folder.

Also the user can easily destroy their chat history by deleting/unsharing a file shared in a chat by accident.

image

I don't have a a good solution :disappointed: 

- We could share the files from rooms into a hidden folder. But sometimes it can be a nice feature if a file shared to a room ends up on your desktop. IMHO it highly depends on the file type. 

  - If people share some funny pictures in a room I probably don't want to download all of them to my desktop (but I also don't want to destroy the whole chat history by deleting/unsharing them). 
  - If a coworker shares a document in the chat on which we both want to work, then having it as a normal file in my cloud is a welcoming feature

- To have at least some structure we could introduce a "Talk" folder to which we share chat files. This would also allow users to exclude it from syncing for example

I'm not entirely convinced of any of the solution but I also don't like the current state... Maybe others have good ideas?

### LINKED ISSUE #3041: Shared files are saved in the root folder for new users
## Steps to reproduce
1. Create a new user (don't login with this user);
2. Share files with this user in a talk conversation;

### Expected behavior
The shared files are saved in the /Talk folder

### Actual behavior
The shared files are saved in the new user's root folder. (The talk folder doesn't exist yet)

nickvergessen:
Fix #1811 
Fix #3041 

- [x] Requires https://github.com/nextcloud/server/pull/19321
- [x] Based on #2983

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -49,6 +49,7 @@ use OCA\Talk\PublicShareAuth\Listener as PublicShareAuthListener;
 use OCA\Talk\PublicShareAuth\TemplateLoader as PublicShareAuthTemplateLoader;
 use OCA\Talk\Room;
 use OCA\Talk\Settings\Personal;
+use OCA\Talk\Share\Listener as ShareListener;
 use OCA\Talk\Share\RoomShareProvider;
 use OCA\Talk\Signaling\Listener as SignalingListener;
 use OCP\AppFramework\App;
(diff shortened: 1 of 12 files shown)


nickvergessen:
One time I would like to see the sharing integration tests pass here

nickvergessen:
I don't know whats wrong with those failing two tests.
They work fine isolated and running the full file and running the full folder,
but they fail 100% of the time on drone.

hugleo:
I think the Talk folder should be created only if needed. Even if I delete this folder it will automatically be recreated, even without any files inside.

PR #2879, opened 2020-01-31
Bugfix/noid/fix chatview layout glitches

ma12-co:
### Layout

How to test:
1) Reduce the width of the screen to 600px or so;
2) Write a long word in the new message form so that the word is take more than 1 line long;

Before this pr: 
* The newmessageform layout breaks (gets wider)
* When posting the message, the messageslist layout brakes as well in the same way;

After this pr: 
* The words are correctly broken down

### Quote: 
Now the quote component limits the amount of rows in the quote via CSS

How to test:

1) Open a conversation with long quotes (more than 2 lines);
2) Resize the window (make it narrower);

Before this pr: 
* The amount of text displayed doesn't change and the quote row number increase.

After this pr: 
* The amount of text displayed is less and the row number is the same (2)

### DIFF
--- a/src/components/Quote.vue
+++ b/src/components/Quote.vue
@@ -31,7 +31,7 @@ components.
 				<h6>{{ getDisplayName }}</h6>
 			</div>
 			<div class="quote__main__text">
-				<p>{{ shortenQuotedMessage }}</p>
+				<p>{{ simpleQuotedMessage }}</p>
 			</div>
 		</div>
 		<div v-if="isNewMessageFormQuote" class="quote__main__right">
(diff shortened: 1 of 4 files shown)


src/components/Quote.vue
```diff
@@ -145,49 +145,6 @@ export default {
 			return subtitle
 		},
 
-		shortenQuotedMessage() {
-			let cutAtChar = 200
```
nickvergessen: Wouldn't it still make sense to do something like this in case of very long messages? There is not much use in putting 64k chars into the html when only 2 lines are rendered

ma12-co: I'm not sure... on the other side also executing all that logic for each every quote message (most of them are not going to be that long) could also slow things down.. What do you think?

nickvergessen: Well we can kill the logic now and just substring(0, 250) or something

nickvergessen:
- [ ] Breaks mention bubbles

Bildschirmfoto von 2020-01-31 16-40-35

nickvergessen:
Works for plain messages, mentions have the issue shown above

src/components/Quote.vue
ma12-co: done

src/components/Quote.vue
```diff
@@ -144,50 +144,17 @@ export default {
...
-						// since we cut in the middle of the word.
-						cuttingAfterWord = false
-					}
-				}
+		// Shorten the message to 250 characters and append three dots to the end of the
+		// string. This is needed because on very wide screens, if the 250 characters
+		// fit, the css rules won't ellipsize the text-overflow.
+		shortenedQuoteMessage() {
+			if (this.simpleQuotedMessage.length >= 250) {
+				return this.simpleQuotedMessage.substring(0, 250) + '...'
```
skjnldsv: Please use the ellipsis character instead (for Jan)

skjnldsv:
```suggestion
				return this.simpleQuotedMessage.substring(0, 250) + '…'
```

src/components/Quote.vue
```diff
@@ -221,10 +187,15 @@ export default {
 		}
 		&__text {
 			color: var(--color-text-light);
+			white-space: pre-wrap;
+			word-break: break-word;
 			& p {
-				max-width: 662px;
 				text-overflow: ellipsis;
 				overflow: hidden;
+				// Allow 2 lines max and ellipsize the overflow;
```
skjnldsv: clamping is experimental, be aware of the browsers talk claim to support (no idea, just a warning :wink: )

src/components/Quote.vue
ma12-co: OK! I didn't know there was a difference, thanks :)

src/components/Quote.vue
ma12-co: The latest version of every browser but ie supports it. We recommend the latest version in the talk page. For ie, the substring() method above is a nice fallback!

PR #2880, opened 2020-01-31
Cancel getparticipants requests once changing conversation

ma12-co:
Avoid duplicating getparticipants requests when quickly switching between conversations.

### DIFF
--- a/src/components/RightSidebar/Participants/ParticipantsTab.vue
+++ b/src/components/RightSidebar/Participants/ParticipantsTab.vue
@@ -91,6 +91,7 @@ import isInLobby from '../../../mixins/isInLobby'
 import { loadState } from '@nextcloud/initial-state'
 import SHA1 from 'crypto-js/sha1'
 import Hex from 'crypto-js/enc-hex'
+import CancelableRequest from '../../../utils/cancelableRequest'
 
 export default {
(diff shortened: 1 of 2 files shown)


nickvergessen:
Maybe when it's not the first visit in the conversation we could even delay this request a bit? It's okay to see an outdated but mostlikely unchanged page for some seconds, if everything is a lot quicker in response

ma12-co:
mmm not sure, you could have visited it already in the same tab but some time ago?

nickvergessen:
Yeah, but the next refresh (currently happening every 30s) will fix it, sounds good enough?
Or we use a timeout and refresh it manually once after ~5s, but not before the chat is loaded. that seems more important?

ma12-co:
> sounds good enough

it doesn't to me.. It can be misleading. Imagine you expect someone to be there and you don't see it when going back to the conv. Then you try to add him and you get some kind of error (the participant is already in the conversation??) :P

> Or we use a timeout and refresh it manually once after ~5s, but not before the chat is loaded. that seems more important?

If we do this I think we should keep the list in a loading state until refreshed, for the same reason described above

ma12-co:
So what do we do here @nickvergessen??

PR #2895, opened 2020-02-04
Immediately update last message preview in conversation list

ma12-co:
### How to test: 
1) Open a conversation;
2) Write some messages;
3) Look at the message preview in `Conversation` component;

### Before this pr:
The preview message is not updated until the whole `ConversationList` refreshes (which happens every 30 seconds).

### After this pr: 
The conversation updates immediately.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -349,6 +349,7 @@ export default {
 
 				// this.$store.dispatch('purgeConversationsStore')
 				this.$store.dispatch('addConversation', response.data.ocs.data)
+				this.$store.dispatch('markConversationRead', token)
 
 				/**
 				 * Emits a global event that is used in App.vue to update the page title once the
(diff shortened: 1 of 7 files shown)


nickvergessen:
So, :+1: for updating the message right away. The thing I dislike about this is that it should happen in the same way with the unread message bubble and so on. But yeah, I guess we can do it like this for now.

nickvergessen:
Another error I just noticed:

1. Navigate to a conversation and copy the link
2. Navigate away
3. As another user write a message
4. Paste link from 1. again

### Expected
Last message is last message

### Actually
It flickers:
1. Last message (as per conversation list)
2. Last read message (as it's the last unread item)
3. Last message (once the chat loaded the new messages)

nickvergessen:
The error from above also means chats you visited once will never be updated until you visited them actively

PR #2899, opened 2020-02-07
Also scroll on inactive tabs

nickvergessen:
It's more what people expect than having to see that the window
is scrollable and not at the bottom. We might revert this, when the
"Unread messages" marker was implemented, but I don't think we should


mario:
Looks decent :)

PR #2934, opened 2020-02-10
Fix sidebars integrations

### LINKED ISSUE #2933: Chat in sidebar broken
Tested with Talk 8.0.3 on Nextcloud 18.

Open the right sidebar ("Details" view) and go to the chat tab. 

Expected: see the chat for the file.

Result: The user only sees an empty tab.

image

I couldn't test it until now because the whole chat in the sidebar is broken. But please also double check that the video calling functionality works in the sidebar when a document (office and/or markdown) is opened.

nickvergessen:
Fix #2933

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## 8.0.4 – 2020-02-11
+### Fixed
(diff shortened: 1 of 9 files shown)


Ivansss:
Public share and video verification works.
But I still get `Missing required prop: "id"` in Files sidebar.

PR #2942, opened 2020-02-11
Send the notifications after joining the call

nickvergessen:
This prevents issues when sending takes long, e.g. with a lot of users

### DIFF
--- a/lib/Notification/Listener.php
+++ b/lib/Notification/Listener.php
@@ -44,6 +44,9 @@ class Listener {
 	/** @var ILogger */
 	protected $logger;
 
+	/** @var bool */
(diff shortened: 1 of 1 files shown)


nickvergessen:
Change:
1. In the before we now only check whether we should send the call notifications
2. The actual sending and clearing of old notifications is done in the after event

Before this patch the android app could be tricked into displaying a "You missed a call in ..." notification, when the connections to the push proxy took long enough after the user received their notification already. Can be tested by adding a `sleep(5);` after https://github.com/nextcloud/notifications/blob/ae6a18c31d73cefd3eb68b6ca98a44432b391e02/lib/Push.php#L170

mario:
Awesome.

PR #2962, opened 2020-02-17
Make sure the secondary view has an id

juliushaertl:
The secondary view registration requires its views to have an id property set https://github.com/nextcloud/server/blob/b476659ac888618fdb01c1316db073237c9222d3/apps/files/src/services/Sidebar.js#L69

To reproduce enable the talk app and try to access the system tags in the files sidebar.

Related fix for system tags https://github.com/nextcloud/server/pull/19511


nickvergessen:
> To reproduce enable the talk app and try to access the system tags in the files sidebar.

Ah, I already wondered why tags are not working....

So backport to stable18 is needed?

PR #2976, opened 2020-02-18
Print sharing error on correct level

### LINKED ISSUE #2900: Don't fail silently when you're not allowed to re-share a file
## Steps to reproduce
1. User A creates a folder and shares it with B (re-sharing is disabled globally in the settings)
2. A and B talk together and B tries to add a file of the shared folder to the chat (with the Paperclip)

### Expected behaviour
Either the file should be added to the chat (but I see this may be problematic permission-wise), or it should at least result in an error-message and not failing silently.

### Actual behaviour
Nothing happens

## Browser
**Operating system:** Windows

**Browser name:** Chrome

**Browser version:** 79.0.3945.130 

### Browser log
image

image

## Server configuration
Talk 8.0.1
NC 18.0.0

nickvergessen:
Fix #2900

### DIFF
--- a/src/components/NewMessageForm/NewMessageForm.vue
+++ b/src/components/NewMessageForm/NewMessageForm.vue
@@ -190,25 +190,35 @@ export default {
 
 		async handleFileShare() {
 			picker.pick()
-				.then(path => {
+				.then(async(path) => {
 					console.debug(`path ${path} selected for sharing`)
 					if (!path.startsWith('/')) {
 						throw new Error(t('files', 'Invalid path selected'))
 					}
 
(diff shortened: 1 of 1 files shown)


nickvergessen:
Error message is better without the `/<user>/files` prefix, that is fixed in: https://github.com/nextcloud/server/pull/19528

nickvergessen:
> but shouldn't one be always allowed to reshare the file with the file owner in the chat?

Well arguable, but gets difficult already with group conversation and so on. So for now if you cant share, you have to copy paste an internal link yourself.

PR #2891, opened 2020-02-03
File upload

ma12-co:
Signed-off-by: Marco Ambrosini <marcoambrosini@pm.me>

PR #2983, opened 2020-02-19
Allow to select an attachment folder

nickvergessen:
- [ ] Option to create a folder is currently missing. Bug in server: https://github.com/nextcloud/server/pull/19539
- [ ] AppNavigationSettings close when interacting with the filepicker: https://github.com/nextcloud/nextcloud-vue/issues/862

### DIFF
--- a/docs/capabilities.md
+++ b/docs/capabilities.md
@@ -42,3 +42,8 @@ title: Capabilities
 ## 8.0
 * `chat-replies` - Normal chat messages can now be replied to. Check the `isReplyable` parameter on the message object.
 * `circles-support` - Conversations can be created for circles and all circle members can be added to existing conversations
+
+
+## 9.0
+* `config => attachments => allowed` - Whether the user can upload files into a chat
+* `config => attachments => folder` - User defined folder where items should be uploaded to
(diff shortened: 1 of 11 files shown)


src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -115,6 +137,8 @@ export default {
 			searchResultsCircles: [],
 			contactsLoading: false,
 			isCirclesEnabled: loadState('talk', 'circles_enabled'),
+			attachmentFolder: loadState('talk', 'attachment_folder'),
```
nickvergessen: I guess in the end this needs to go into a store, otherwise the file upload from #2891 would still have the old value.

docs/capabilities.md
```diff
@@ -42,3 +42,8 @@ title: Capabilities
 ## 8.0
 * `chat-replies` - Normal chat messages can now be replied to. Check the `isReplyable` parameter on the message object.
 * `circles-support` - Conversations can be created for circles and all circle members can be added to existing conversations
+
+
+## 9.0
+* `config => attachments => allowed` - Whether the user can upload files into a chat
+* `config => attachments => folder` - User defined folder where items should be uploaded to
```
nickvergessen: @mario @Ivansss when you do upload, this is the folder where you should upload to.

src/components/LeftSidebar/LeftSidebar.vue
nickvergessen: @ma12-co where should we store this? New store? or can we put it in an existing one?

nickvergessen:
Another chunk read @ma12-co lets get this in

PR #3000, opened 2020-02-25
Rename duplicate files when uploading

### LINKED ISSUE #2987: Uploading a file with the same name overwrites the old one
## Steps to reproduce
1. Upload `Readme.md` in conversation A
2. Upload another different `Readme.md` in conversation B
3. Realise the `Readme.md` in conversation A is updated.

### Expected behaviour
In B a new file `Readme (2).md` should be uploaded

### Actual behaviour
Update of the existing file

### Solution
The android app sends a HEAD request to the webdav endpoint to check if the file already exists. If it does, `(2)` is added before the last `.` in the filename (or at the end in case there is no dot). Then send another head request. In case of another match, search for ` \(\d+\)` before the last dot, and increase the number by 1. Repeat this until the file files not exist.

ma12-co:
fix #2987 

### DIFF
--- a/src/store/fileUploadStore.js
+++ b/src/store/fileUploadStore.js
@@ -127,19 +127,19 @@ const actions = {
 			// Mark file as uploading to prevent a second function call to start a
 			// second upload for the same file
 			commit('markFileAsUploading', { uploadId, index })
-			// Get the current user id
-			const userId = getters.getUserId()
 			// currentFile to be uploaded
 			const currentFile = state.uploads[uploadId].files[index].file
-			// Destination path on the server
-			const path = '/files/' + userId + getters.getAttachmentFolder() + '/' + currentFile.name
-			const uniquePath = await findUniquePath(client, path)
+			// userRoot path
+			const userRoot = '/files/' + getters.getUserId()
+			// Candidate rest of the path
+			const path = getters.getAttachmentFolder() + '/' + currentFile.name
+			// Get a unique relative path based on the previous path variable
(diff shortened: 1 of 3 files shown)


src/store/fileUploadStore.js
```diff
@@ -108,44 +112,36 @@ const mutations = {
...
 			// Destination path on the server
 			const path = '/files/' + userId + getters.getAttachmentFolder() + '/' + currentFile.name
+			const uniquePath = await findUniquePath(client, path)
 			try {
 				// Upload the file
-				await client.putFileContents(path, currentFile)
+				await client.putFileContents(uniquePath, currentFile)
+				// Compute the sharePath by removing the first part of the upload path
+				const partOfPathToRemove = '/files/' + userId
+				const sharePath = uniquePath.substring(partOfPathToRemove.length, uniquePath.length)
```
nickvergessen: make your life easier by not passing `'/files/' + userId` to `findUniquePath` in first place?

ma12-co: This has to be done, whether it's in findUniquePath or outside of it. So no easier life I think :)

nickvergessen: I ment something like https://gist.github.com/nickvergessen/e88c72d68fbc5d457a2c145c9e880954

nickvergessen:
For me the sharing throws a 404, because the path is missing from the sent parameters:
```json
{"shareType":10,"shareWith":"vnia5iw6"}
```

PR #3005, opened 2020-02-25
Prevent resharing of the same file

ma12-co:
### How to test

1) Create a conversation with no previously shared files 
2) Upload a file;
3) Upload a second file;

### before this pr: 
When sharing the second file, the first file is re-shared before the second.

### after this pr: 
When sharing the second file, only the second file is shared.

### DIFF
--- a/src/components/NewMessageForm/NewMessageForm.vue
+++ b/src/components/NewMessageForm/NewMessageForm.vue
@@ -285,7 +285,18 @@ export default {
 			for (const index in shareableFiles) {
 				// The pat of the file to share to the conversation
 				const path = this.$store.getters.getAttachmentFolder() + '/' + shareableFiles[index].name
-				shareFile(path, token)
+				try {
+					// Mark current file as sharing
+					this.$store.dispatch('markFileAsSharing', { token, index })
+				} catch {
+					continue
+				}
+				try {
+					await shareFile(path, token)
+					this.$store.dispatch('markFileAsShared', { token, index })
+				} catch (exception) {
+					console.debug('An error happened when triying to share your file: ', exception)
+				}
 			}
 		},
 	},
--- a/src/store/fileUploadStore.js
+++ b/src/store/fileUploadStore.js
@@ -169,6 +169,7 @@ const actions = {
 	 * Mark a file as shared
 	 * @param {object} context default store context;
 	 * @param {object} param1 conversation token and original file index
+	 * @throws {Error} when the item is already being shared by another async call
 	 */
 	markFileAsSharing({ commit, state }, { token, index }) {
 		if (state.files[token][index].status !== 'successUpload') {
--- a/src/store/fileUploadStore.js
+++ b/src/store/fileUploadStore.js
@@ -39,15 +39,17 @@ const getters = {
 	// Returns all the files that have been successfully uploaded
 	getShareableFiles: (state) => (token) => {
 		if (state.files[token]) {
-			const shareableFiles = []
+			const shareableFiles = {}
 			for (const index in state.files[token]) {
 				const currentFile = state.files[token][index]
 				if (currentFile.status === 'successUpload') {
-					shareableFiles.push(currentFile.file)
+					shareableFiles[index] = (currentFile.file)
 				}
 			}
 			return shareableFiles
-		} return undefined
+		} else {
+			return {}
+		}
 	},
 
 	getAttachmentFolder: (state) => () => {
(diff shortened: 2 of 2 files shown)


src/store/fileUploadStore.js
```diff
@@ -84,6 +86,11 @@ const mutations = {
 		state.files[token][index].status = 'uploading'
 	},
 
+	// Marks a given file as shared
+	markFileasShared(state, { token, index }) {
```
nickvergessen: *As

src/store/fileUploadStore.js
```diff
@@ -152,6 +159,16 @@ const actions = {
 	setAttachmentFolder(context, attachmentFolder) {
 		context.commit('setAttachmentFolder', attachmentFolder)
 	},
+
+	/**
+	 * Mark a file as shared
+	 * @param {object} context default store context;
+	 * @param {object} param1 conversation token and original file index
+	 */
+	markFileasShared(context, { token, index }) {
```
nickvergessen: *As

src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -285,7 +285,9 @@ export default {
 			for (const index in shareableFiles) {
 				// The pat of the file to share to the conversation
 				const path = this.$store.getters.getAttachmentFolder() + '/' + shareableFiles[index].name
-				shareFile(path, token)
+				await shareFile(path, token)
+				// Mark current file as shared
+				this.$store.dispatch('markFileasShared', { token, index })
```
nickvergessen: Concurrency problem. Need to update it to sharing before doing the await similar as with uploading

ma12-co:
> Unrelated to this PR, but I find it a bit weird, that when you select a lot of files, they are first all uploaded, then all shared.
It would feel nicer if each file is asynchronously uploaded and shared synchronously after its upload instead of all uploads?thinking

Probably for now, but we would have to go back to the code we have now once we introduce the intermediate view with previews and editing i guess

src/store/fileUploadStore.js
```diff
@@ -39,15 +39,17 @@ const getters = {
 	// Returns all the files that have been successfully uploaded
 	getShareableFiles: (state) => (token) => {
 		if (state.files[token]) {
-			const shareableFiles = []
+			const shareableFiles = {}
 			for (const index in state.files[token]) {
 				const currentFile = state.files[token][index]
 				if (currentFile.status === 'successUpload') {
-					shareableFiles.push(currentFile.file)
+					shareableFiles[index] = (currentFile.file)
```
nickvergessen: change the state in here, otherwise a concurrent run gets it too

src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -285,7 +285,9 @@ export default {
 			for (const index in shareableFiles) {
 				// The pat of the file to share to the conversation
 				const path = this.$store.getters.getAttachmentFolder() + '/' + shareableFiles[index].name
-				shareFile(path, token)
+				// Mark current file as shared
+				this.$store.dispatch('markFileasShared', { token, index })
```
nickvergessen:
```suggestion
				this.$store.dispatch('markFileAsShared', { token, index })
```

src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -285,7 +285,9 @@ export default {
 			for (const index in shareableFiles) {
 				// The pat of the file to share to the conversation
 				const path = this.$store.getters.getAttachmentFolder() + '/' + shareableFiles[index].name
-				shareFile(path, token)
+				// Mark current file as shared
+				this.$store.dispatch('markFileAsShared', { token, index })
```
nickvergessen: this has to happen for all files before the first `await shareFile`
Or you have to check the state within the loop (if it even updates), to make sure no other execution is sharing it between you sharing item1 and item2.

src/store/fileUploadStore.js
```diff
@@ -152,6 +164,28 @@ const actions = {
...
 	},
+
+	/**
+	 * Mark a file as shared
+	 * @param {object} context default store context;
+	 * @param {object} param1 conversation token and original file index
+	 */
+	markFileAsSharing({ commit, state }, { token, index }) {
+		if (state.files[token][index].status !== 'successUpload') {
+			throw new Error('Item is already being shared')
```
nickvergessen: document `@throws` or is this not common in JS?

ma12-co: I don't get the question https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw

PR #3014, opened 2020-02-26
Get rid of underscore, jQuery and the console.log statements

### LINKED ISSUE #2562: Get rid of jquery and underscore in signaling and webrtc
The Signaling and WebRTC code are still using:

jQuery:
* ajax()
* find() - when handling the 403 dialog to ask for the conversation password
* Deferred()
* each()

Underscore:
* keys
* extend
* defer

nickvergessen:
- [x] console.log in signaling
- [x] console.log in webrtc
- [x] $.* in signaling
- [x] $() in signaling
- [x] $.* in webrtc
- [x] _.* in signaling
- [x] Replace OC.Notifications with `@nextcloud/dialogs`

Fix #2562

### DIFF
--- a/src/utils/webrtc/webrtc.js
+++ b/src/utils/webrtc/webrtc.js
@@ -776,7 +776,7 @@ export default function initWebRTC(signaling, _callParticipantCollection) {
 		}
 	})
 
-	webrtc.on('localMediaStarted', function(configuration) {
+	webrtc.on('localMediaStarted', function(/* configuration */) {
 		console.info('localMediaStarted')
 
 		clearLocalStreamRequestedTimeoutAndHideNotification()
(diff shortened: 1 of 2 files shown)


src/utils/webrtc/webrtc.js
```diff
@@ -669,15 +666,15 @@ export default function initWebRTC(signaling, _callParticipantCollection) {
 				checkPeerMedia(peer, video, 'video').then(function() {
 					clearInterval(peer.check_video_interval)
 					peer.check_video_interval = null
-				})
+				}).catch()
```
ChristophWurst: may I suggest `.catch(console.error.bind(this))`? Then it will still log the error

nickvergessen: The error is already logged inside checkPeerMedia

ChristophWurst: Still, the `then` callback could throw as well …

nickvergessen: but then we can also jsut get rid of the catch, right?
Because if clearInterval throws....

ChristophWurst: Depends. Some browsers will still show you the error then. Others will just swallow it. Or complain that you're not handling the error.

IIRC `.catch(console.error.log(this))` is the easiest way to "handle" it when you don't care or think it's unlikely to ever get triggered.

ChristophWurst: See my remark as a hint for a "best practice". You're free to ignore it :)

PR #3024, opened 2020-02-28
Only update the named session if it is a guest

nickvergessen:
Before all guests were updated to moderators in the database


nickvergessen:
This was actually a regression from 81db771748c2a35c0d953d608c9996a3cdf02102
So it only affects Talk 8 and backports to 7 and 6 are not needed

Issue #2994, opened 2020-02-24
Drag and drop file upload


ma12-co:
Add ability to upload and share files to a chat by dragging them in the chat.

nickvergessen:
Btw, I would make it similar to gihub and make the dropzone the new message part only.

ma12-co:
> Btw, I would make it similar to gihub and make the dropzone the new message part only.

Why not the whole chatview? I think it's more user friendly

nickvergessen:
Well as long as there is good enough hover feedback, that should be good enough.
Having the full chat view show a hover state when you drop a file looks kind of weird (check files app)

szaimen:
> Why not the whole chatview? I think it's more user friendly

+1 for that. E.g. whatsapp web is also allowing that.

PR #3045, opened 2020-03-03
Add ability to upload files by dropping them in the chat

ma12-co:
Fix: #2994

Peek 2020-03-04 09-40

### DIFF
--- a/src/components/ChatView.vue
+++ b/src/components/ChatView.vue
@@ -31,11 +31,11 @@
 				<div class="drop-hint">
 					<div
 						class="drop-hint__icon"
-						:class="{'icon-upload' : !isGuest, 'icon-error' : isGuest}" />
-					<h3
+						:class="{'icon-upload' : !isGuest, 'icon-user' : isGuest}" />
+					<h2
 						class="drop-hint__text">
 						{{ dropHintText }}
-					</h3>
+					</h2>
 				</div>
 			</div>
 		</transition>
@@ -127,6 +127,11 @@ export default {
 
 .drop-hint {
 	margin: auto;
+	&__icon {
+		background-size: 48px;
+		height: 48px;
+		margin-bottom: 16px;
+	}
 }
 
 .slide {
--- a/src/components/ChatView.vue
+++ b/src/components/ChatView.vue
@@ -23,7 +23,7 @@
 		class="chatView"
 		@dragover.prevent="isDraggingOver = true"
 		@dragleave.prevent="isDraggingOver = false"
-		@drop.prevent="processFiles">
+		@drop.prevent="handleDropFiles">
 		<transition name="slide" mode="out-in">
 			<div
 				v-show="isDraggingOver"
@@ -50,7 +50,7 @@
 <script>
 import MessagesList from './MessagesList/MessagesList'
 import NewMessageForm from './NewMessageForm/NewMessageForm'
-import { shareFile } from '../services/filesSharingServices'
+import { processFiles } from '../utils/fileUpload'
 
 export default {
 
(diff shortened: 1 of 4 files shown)


src/components/ChatView.vue
```diff
@@ -45,15 +68,107 @@ export default {
...
+			isDraggingOver: false,
+		}
+	},
+
+	methods: {
+		/**
+		 * Uploads and shares the selected files
+		 * @param {object} event the file input event object
+		 */
+		async processFiles(event) {
```
nickvergessen: Maybe this can be deduplicated from NewMessageForm. It looks rather complex and is almost the same content?
```diff
--- ChatView.vue
+++ NewMessageForm.vue
@@ -1,14 +1,14 @@
 
 		async processFiles(event) {
-			// restore non dragover state
-			this.isDraggingOver = false
 			// Store the token in a variable to prevent changes when changing conversation
 			// when the upload is still running
 			const token = this.token
 			// Create a unique id for the upload operation
 			const uploadId = new Date().getTime()
 			// The selected files array coming from the input
-			const files = Object.values(event.dataTransfer.files)
+			const files = Object.values(event.target.files)
+			// Clear the input for the next uploads
+			this.$refs.fileUploadInput.value = ''
 			// Process these files in the store
 			await this.$store.dispatch('uploadFiles', { uploadId, token, files })
 			// Get the files that have successfully been uploaded from the store
```

src/components/ChatView.vue
```diff
@@ -19,15 +19,38 @@
...
 
 <template>
-	<div class="chatView">
-		<MessagesList :token="token" />
-		<NewMessageForm />
+	<div
+		class="chatView"
+		@dragover.prevent="isDraggingOver = true"
+		@dragleave.prevent="isDraggingOver = false"
+		@drop.prevent="processFiles">
```
nickvergessen: Hover state shows for guests and triggers a basic auth login mask by the browser afterwards.
Either we show "You need to be logged in to upload files" in the drop zone, or the drop zone should not show up at all.

src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -274,35 +276,12 @@ export default {
...
-		 * @param {object} event the file input event object
-		 */
-		async processFiles(event) {
-			// Store the token in a variable to prevent changes when changing conversation
-			// when the upload is still running
-			const token = this.token
+		handleFileInput(event) {
+			const files = Object.values(event.target.files)
 			// Create a unique id for the upload operation
 			const uploadId = new Date().getTime()
```
nickvergessen: Since you always create it in the same way, handle it inside `processFiles` ?

ma12-co: I'd like to have the ID at hand within the components for future features

ma12-co: It's the only way we can access the file and do things with them e.g. keep track of upload status

src/components/ChatView.vue
```diff
@@ -19,15 +19,40 @@
...
+		@dragleave.prevent="isDraggingOver = false"
+		@drop.prevent="handleDropFiles">
+		<transition name="slide" mode="out-in">
+			<div
+				v-show="isDraggingOver"
+				class="dragover">
+				<div class="drop-hint">
+					<div
+						class="drop-hint__icon"
+						:class="{'icon-upload' : !isGuest, 'icon-error' : isGuest}" />
```
nickvergessen: I really dislike `icon-error`, I think I would keep the upload icon instead.

Also (similar to the empty content message) I think the icons should be 64*64px
Speaking of empty content view, maybe you can use that code layout too (h2, bigger icon)?

https://github.com/nextcloud/spreed/blob/fa890665b2c5f50ac8189ad2c7d72b8edfef64ad/src/views/NotFoundView.vue#L2-L8

nickvergessen: I really dislike `icon-error`, I think I would keep the `icon-user` or `icon-upload` instead.

ma12-co: 64 looked gigantic, I increased to 48

ma12-co:
> It is important to see this to verify that you actually drop into the correct conversation.

Can we display the conversation name in the hint instead @jancborchardt? 

I think that this, together with a bit of transparency would solve the problem you expose.

ma12-co:
@jancborchardt how about this: 

Peek 2020-03-06 12-13

src/store/conversationsStore.js
```diff
@@ -45,6 +45,13 @@ const state = {
 const getters = {
 	conversations: state => state.conversations,
 	conversationsList: state => Object.values(state.conversations),
+	/**
+	 * Get a conversation providing it's token
+	 * @param {object} state state object
+	 * @returns {function} The callback function
+	 * @returns {object} The conversation object
+	 */
+	conversation: state => token => state.conversations[token],
```
nickvergessen: You make a follow up PR to use this new method everywhere? :)

Issue #3048, opened 2020-03-04
Problem with conversations renaming


ma12-co:
## Steps to reproduce
1. Click on rename conversation;
2. Write a new name in the input field but don't hit enter;

### Expected behavior
The name doesn't change;

### Actual behavior
After some seconds, the name changes back to the original conversation's name;

PR #3051, opened 2020-03-05
Don't interrupt the renaming process when conversations are refreshed

ma12-co:
fix: #3048


nickvergessen:
Now the name input is empty when you press the edit option.
Makes it quite complicated if you just want to fix a typo or something?
Can't we copy the current value?

ma12-co:
>  Can't we copy the current value?

Yes we can, please re review!

Issue #3047, opened 2020-03-04
Using conversations' actions while searching is broken


ma12-co:
## Steps to reproduce
1. Search a conversation;
2. Open the conversation's 3 dot menu while searching;

### Expected behavior
The 3 dot menu opens  and the user is able to do actions;

### Actual behavior
The 3 dot menu opens but the search operation ends, moving the conversation that the user was acting on further down the list;

PR #3052, opened 2020-03-05
Don't interrupt search operation when clicking on a conversation

ma12-co:
fix: #3047 

I think this is the best solution for now.


nickvergessen:
Works, although I think this is somewhat unexpected. To me it seems that the Actions button is missing a `.prevent` on it's click?

PR #3073, opened 2020-03-10
Display contact sources in new group conversation creation search

ma12-co:
Screenshot from 2020-03-12 13-38-16


nickvergessen:
Some remarks:

1. I think the `Search for more contacts` text + icon should only show up when the search term is empty. It looks very confusing at the bottom otherwise:

> Bildschirmfoto von 2020-03-12 16-14-42


2. I think empty headers should not be displayed, or they should be grouped and show a "No search results" message like in the left sidebar

PR #3077, opened 2020-03-11
Handle the new reference id for snappier chat feeling

nickvergessen:
- [x] Requires https://github.com/nextcloud/server/pull/19890
- [x] Handle reference id in the backend (allowing to set it and expose it)
- [x] Handle reference id in the UI when sending a message
- [x] Expand integration and unit tests

### To test
Apply the following patch to simulate a lot of notifications being sent:
```diff
diff --git a/lib/Controller/ChatController.php b/lib/Controller/ChatController.php
index d0bf74ee..ef984dd5 100644
--- a/lib/Controller/ChatController.php
+++ b/lib/Controller/ChatController.php
@@ -191,6 +191,8 @@ class ChatController extends AEnvironmentAwareController {
                if ($parentMessage instanceof Message) {
                        $data['parent'] = $parentMessage->toArray();
                }
+
+               sleep(10);
                return new DataResponse($data, Http::STATUS_CREATED);
        }
 

```

### Before
Peek 2020-03-11 13-28

### After
Peek 2020-03-11 13-30

### DIFF
--- a/docs/capabilities.md
+++ b/docs/capabilities.md
@@ -50,3 +50,4 @@ title: Capabilities
 * `config => conversations => can-create` - Whether the user can create public and group conversations, if not only one-to-one conversations are allowed
 * `force-mute` - "forceMute" signaling messages can be sent to mute other participants.
 * `conversation-v2` - The conversations API v2 is less load heavy and should be used by clients when available. Check the difference in the [Conversation API documentation](conversation.md).
+* `chat-reference-id` - an optional referenceId can be sent with a chat message to be able to identify it in parallel get requests to earlier fade out a temporary message
--- a/docs/chat.md
+++ b/docs/chat.md
@@ -44,6 +44,7 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
         `systemMessage` | string | empty for normal chat message or the type of the system message (untranslated)
         `messageType` | string | Currently known types are `comment`, `system` and `command`
         `isReplyable` | bool | True if the user can post a reply to this message (only available with `chat-replies` capability)
+        `referenceId` | string | A reference string that was given while posting the message to be able to identify a sent message again (only available with `chat-reference-id` capability)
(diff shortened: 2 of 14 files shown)


mario:
Can I get a clarification on this PR so I can do this on the mobile as well? :))

Thank youuu!

nickvergessen:
> Can I get a clarification on this PR so I can do this on the mobile as well? :))
> 
> Thank youuu!

Doc changes are there, are they good enough?

nickvergessen:
Server PR is in, so ready for review

mario:
> > Can I get a clarification on this PR so I can do this on the mobile as well? :))
> > Thank youuu!
> 
> Doc changes are there, are they good enough?

Yup.

nickvergessen:
Rebased after server PR merge so integration tests pass

nickvergessen:
Did you reinstall your nextcloud or run the database command to add optional columns?

mario:
I did. On the mobile apps I receive the referenceId just fine, the issue exists only on the web.

nickvergessen:
Works fine here, maybe you didn't recompile the frontend, so it would not set the reference id from the web and therefor also get none back?

mario:
> Works fine here, maybe you didn't recompile the frontend, so it would not set the reference id from the web and therefor also get none back?

That would be weird, but I won't say no - weirder things happened in the past. I think it does set the reference id but it uses one it receives from the POST of the message rather than the one from getting messages.

nickvergessen:
Rebased on latest master so it is testable with latest master of server.

Please try again.

> I think it does set the reference id but it uses one it receives from the POST of the message rather than the one from getting messages.

I don't get what you mean by this. We use both results basically. The result (all message data) of the POST and the GET are written into the message store based on ID. So whatever comes later just "updates" the entry. And since both return the same data (unchanged) it is also fine like this.

ma12-co:
Drone not happy @nickvergessen

nickvergessen:
Drone failures are the known timeout and the one sometimes failing sharing test, so it's fine

PR #3095, opened 2020-03-17
Limit creating public and group conversations to a group

### LINKED ISSUE #2285: Add permissions to groups to create new discussions
It would be great to be able to define which groups can create new discussions (by default to everyone). This would allow an admin group to create a set of chatrooms and share them among groups. Users in groups would see their allowed rooms, talk inside these but would not be able to create new ones. This would help fighting against unwanted discussions.

### LINKED ISSUE #2295: LImit conversation
Hello,

Is it possible to limit conversations in groups?

I'd like to make 1 group per projects and assign group to user (customer/partner) .

Thanks

nickvergessen:
Fix #2285 
Fix #2295 

- [x] Add a setting
- [x] Restrict creation in the UI
- [x] Add some integration tests
- [x] Publish option via capabilities for mobile apps

### DIFF
--- a/docs/capabilities.md
+++ b/docs/capabilities.md
@@ -47,4 +47,5 @@ title: Capabilities
 ## 9.0
 * `config => attachments => allowed` - Whether the user can upload files into a chat
 * `config => attachments => folder` - User defined folder where items should be uploaded to
+* `config => conversations => can-create` - Whether the user can create public and group conversations, if not only one-to-one conversations are allowed
 * `force-mute` - "forceMute" signaling messages can be sent to mute other participants.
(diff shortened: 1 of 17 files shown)


nickvergessen:
@Ivansss @mario how much effort is it for you to also implement the config from the capability (only allowing one-to-one conversations when it is disabled)

Ivansss:
> how much effort is it for you to also implement the config from the capability (only allowing one-to-one conversations when it is disabled)

Shouldn't be too much.

mario:
Done from my side. Will be released as part of 9.0.

PR #3096, opened 2020-03-17
Kickstart unit tests

ma12-co:


tests/unit/example.spec.js
```diff
@@ -0,0 +1,12 @@
...
+
+describe('HelloWorld.vue', () => {
+  it('renders props.msg when passed', () => {
+    const msg = 'new message'
+    const wrapper = shallowMount(HelloWorld, {
+      propsData: { msg }
+    })
+    expect(wrapper.text()).toMatch(msg)
+  })
+})
```
nickvergessen: can we have them in tests/js/ or something?

PR #1926, opened 2019-06-28
⏳ Lobby

nickvergessen:
- [x] Database and API
- [x] Add UI option to set the lobby value
- [x] Restrict the API to prevent lobbied useres from chatting, signaling, ....
- [x] Keep left sidebar for normal users => use empty content message
- [x] System messages
- [x] ~~Quicker kill when turn on lobby again~~ => #1928 

Fix #382

PR #3133, opened 2020-03-23
Add ability to set guest name in lobby

ma12-co:

### DIFF
--- a/src/components/LobbyScreen.vue
+++ b/src/components/LobbyScreen.vue
@@ -25,7 +25,7 @@
 			<h2>{{ currentConversationName }}</h2>
 			<p>{{ message }}</p>
 		</div>
-		<SetGuestUsername />
+		<SetGuestUsername v-if="currentUserIsGuest" />
 	</div>
 </template>
 
@@ -67,7 +67,10 @@ export default {
 
 			return message
 		},
-
+		// Determines whether the current user is a guest user
+		currentUserIsGuest() {
+			return !this.$store.getters.getUserId()
+		},
 	},
 
 }
--- a/src/components/LobbyScreen.vue
+++ b/src/components/LobbyScreen.vue
@@ -19,19 +19,27 @@
   -->
 
 <template>
-	<div class="lobby emptycontent">
-		<div class="icon icon-lobby" />
-		<h2>{{ currentConversationName }}</h2>
-		<p>{{ message }}</p>
+	<div class="lobby">
+		<div class="lobby emptycontent">
+			<div class="icon icon-lobby" />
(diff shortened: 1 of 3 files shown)


src/components/LobbyScreen.vue
```diff
@@ -19,20 +19,28 @@
...
-		<div class="icon icon-lobby" />
-		<h2>{{ currentConversationName }}</h2>
-		<p>{{ message }}</p>
+	<div class="lobby">
+		<div class="lobby emptycontent">
+			<div class="icon icon-lobby" />
+			<h2>{{ currentConversationName }}</h2>
+			<p>{{ message }}</p>
+		</div>
+		<SetGuestUsername />
```
nickvergessen: Missing check for guest?
Also normal users end up in the lobby

ma12-co: Check for guest is in the parent component!

ma12-co: feature/noid/show-selected-participants-in-new-group-creation

ma12-co: Sorry you're right :)

jancborchardt:
This was btw in the spec for the lobby ;)

> - You can set your name already before you join the meeting.
> Talk lobby

https://github.com/nextcloud/spreed/pull/1926#issuecomment-506962798

ma12-co:
Wow, that looks so different than what we have now... I just copied the same component that we use in the sidebar for now, we did it together in January's contributor week @jancborchardt. Don't you think that it's better than the explicit input field??  

Peek 2020-03-27 11-19

jancborchardt:
@ma12-co the idea was that _if_ the name is the default "Guest", you probably want to change it and we can directly show the input field. It wouldn’t need to show as input field if you already set a name. My only suggestions for that:
- We should call it "Your name" instead of "Display name" as that’s quite technical
- The input field could replace the current name and the edit button in place, especially

> Wow, that looks so different from what we have now...

Yeah – it’s rather bland at the moment. I’d still love for the mockup and spec to be followed a bit more:
- It looks much much better. We don’t necessarily need to have the dark background, but this standard blue header and the white content just looks so basic.
- Gives you feedback as to who you will be talking to – just like e.g. Hangouts does before joining
- Shows your own camera in case it’s a video call – also just like Hangouts
- Unclear what "waiting in the lobby" means. "Waiting for moderators to start the call" is clear and you know who to talk to if it does not start.
- Some other points in the spec as well – give it a read at https://github.com/nextcloud/spreed/pull/1926#issuecomment-506962798

ma12-co:
Ok created an issue from your comment

PR #1284, opened 2018-11-08
Prefer h264

mario:
This is my humble attempt at fixing #1223.

PR #1171, opened 2018-08-30
First try to port the JS port

MorrisJobke:
@LukasReschke wants to look at this.

PR #3153, opened 2020-03-25
Prefer H264 codec when it is available

### LINKED ISSUE #1223: Prefer hardware based video encoding
A first attempt was started in #1171 

This should be looked into again, after we moved away from SimpleWebRTC (see #1063 ).
Using the current hack from #1171 (which just does it in php instead of a proper js method) already cuts the cpu usage by 25-50%.

nickvergessen:
Previous attempts:
* #1284
* #1171

### To enable it

```
./occ config:app:set spreed prefer_h264 --value yes
```

You should see the following message as debug in your browsers console:
```
'Preferring hardward codec H.264 as per global configuration'
```

Fix #1223

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -44,6 +44,7 @@ use OCP\AppFramework\Http\Template\PublicTemplateResponse;
 use OCP\EventDispatcher\IEventDispatcher;
 use OCP\Files\IRootFolder;
 use OCP\Files\NotPermittedException;
+use OCP\IConfig;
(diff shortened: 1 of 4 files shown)


nickvergessen:
Let's get this also in and backport, now that it has a config option to be enabled for testing @danxuliu

szaimen:
@nickvergessen why shouldn't this setting be the default configuration?

nickvergessen:
Because we need to do more tests in bigger scale before telling browsers "what you think is best is actually not the best"

PR #3157, opened 2020-03-25
Also bail out when the path is actually null

nickvergessen:
Apparently this can happen when a external mount was shared
that is later not available anymore

Ref https://help.nextcloud.com/t/nextcloud-talk-kills-nextcloud-after-update-to-v18-0-3/75257


nickvergessen:
> Is the bug in Talk, or is it in the external storage?

Well I just saw the report in the forum, so dunno and dont care too much. a null path can not be split

> Anyway, if the bug is in isAccessibleResult it was copied from server, so this should be applied there as well.

Right, okay

Issue #3154, opened 2020-03-25
End of a call is too heavy for the server with O(n²) requests


nickvergessen:
## Steps to reproduce
1. Start a call with 20 people
2. Everytime a participant changes (inCall flag on leaves the call), every participant updates the participant list

### Expected behaviour
The requests should be "bundled"

### Actual behaviour
n users request the participant list each n times, resulting in n² requests

nickvergessen:
Additionally on leaving the conversation a chat message is created, meaning everyone is constantly restarting the "wait for new chat messages" request

nickvergessen:
The update of the participant list is now debounced by 2 seconds, so that should help already a lot.

nickvergessen:
Ref https://github.com/nextcloud/spreed/issues/3377#issuecomment-863850037

PR #3165, opened 2020-03-26
Don't cancel an on going fetch of participants via signaling

nickvergessen:
Otherwise it could be that loading participants never finishes
when your server is not too responsive, as people join and quit
or ping faster than your requests load, it will always cancel
the previous attempt and start a new one

### Patch to help testing
```diff
diff --git a/lib/Controller/RoomController.php b/lib/Controller/RoomController.php
index 1ddfc5e4..ee5c3317 100644
--- a/lib/Controller/RoomController.php
+++ b/lib/Controller/RoomController.php
@@ -689,6 +689,7 @@ class RoomController extends AEnvironmentAwareController {
                        ]);
                }
 
+               sleep(5);
                return new DataResponse($results);
        }
 
diff --git a/lib/Controller/SignalingController.php b/lib/Controller/SignalingController.php
index eef6baec..a05bc95a 100644
--- a/lib/Controller/SignalingController.php
+++ b/lib/Controller/SignalingController.php
@@ -164,7 +164,7 @@ class SignalingController extends OCSController {
                }
 
                $data = [];
-               $seconds = self::PULL_MESSAGES_TIMEOUT;
+               $seconds = 3;//self::PULL_MESSAGES_TIMEOUT;
 
                try {
                        $sessionId = $this->session->getSessionForRoom($token);

```

### DIFF
--- a/src/components/RightSidebar/Participants/ParticipantsTab.vue
+++ b/src/components/RightSidebar/Participants/ParticipantsTab.vue
@@ -94,6 +94,7 @@ export default {
 			 * Stores the cancel function for cancelableGetParticipants
 			 */
 			cancelGetParticipants: () => {},
+			fetchingParticipants: false,
 		}
 	},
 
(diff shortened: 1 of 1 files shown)


nickvergessen:
This helps a bit with #3154  but it does not solve it completly

PR #3190, opened 2020-03-29
Improve reply feature's user experience

ma12-co:
- [x] Provide hover feedback for the message that is going to be the object of the reply;
- [x] Align the reply button with the bottom of the Message component (now it's aligned to the top and it can be very confusing at times);
- [ ] <del> Add on-click tooltip for quick reply by right-clicking from anywhere on the message. Since we now have only one action we could make the cursor a pointer over a replyable message and provide a tooltip with a reply button. This way the user doesn't have to reach over to the reply button on the right.</del>

Peek 2020-03-29 19-57

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -27,6 +27,7 @@ the main body of the message as well as a quote.
 <template>
 	<div
 		class="message"
+		:class="{'hover': showActions && !isSystemMessage}"
 		@mouseover="showActions=true"
 		@mouseleave="showActions=false">
 		<div v-if="isFirstMessage && showAuthor" class="message__author">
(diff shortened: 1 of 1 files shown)


nickvergessen:
Puh, please dont add an on click handler. Making it close to impossible to simply copy text with the mouse?

ma12-co:
> Puh, please dont add an on click handler. Making it close to impossible to simply copy text with the mouse?

Good point, maybe right click? :)

skjnldsv:
> Good point, maybe right click? :)

With the only exception being the files right click, we do not use right click event as this is not mobile-friendly nor discoverable :)

ma12-co:
> we do not use right click event as this is not mobile-friendly nor discoverable

Wasn't suggesting to replace the current functionality with it, just as an added thing :)

ma12-co:
Abort right click after talk call!

jancborchardt:
Right-click can be an _additional_ thing, see [Handle right-click properly so it’s a shortcut for action menus](https://github.com/nextcloud/server/issues/13164).

However, that’s only an addition, if we ever do it. We need to make sure that everything is fully usable without right-click and thus should first focus on that. :)

PR #3201, opened 2020-03-30
Start with the last camera/microphone active state

### LINKED ISSUE #2485: Ability to start call with camera & mic on

nickvergessen:
same behaviour as with the old (17) UI.
It's on by default on the first visit and afterwards remembers the state (on a per conversation level as discussed with @jancborchardt ).

Once https://github.com/nextcloud/nextcloud-browser-storage is there, we shall use that, for now we access it directly like in the old days.

Fix #2485

### DIFF
--- a/src/utils/webrtc/models/LocalMediaModel.js
+++ b/src/utils/webrtc/models/LocalMediaModel.js
@@ -19,6 +19,8 @@
  *
  */
 
+import store from '../../../store/index.js'
+
 export default function LocalMediaModel() {
 
 	this.attributes = {
@@ -32,6 +34,7 @@ export default function LocalMediaModel() {
 		videoAvailable: false,
 		videoEnabled: false,
 		localScreen: null,
+		token: '',
 	}
 
 	this._handlers = []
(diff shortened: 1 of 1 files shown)


jancborchardt:
It currently uses browser local storage so is limited to the web interface. Would we follow this for mobile as well, as currently it always starts with microphone active? cc @mario @Ivansss

apiraino:
thanks @nickvergessen :-)

Issue #3194, opened 2020-03-30
Can’t add /wiki command


Thovi98:
## Steps to reproduce
occ talk:command:add-samples

### Expected behaviour
Add all samples commands. 

### Actual behaviour
PHP Parse error:  syntax error, unexpected ',', expecting ']' in /var/www/nextcloud/apps/spreed/sample-commands/wikipedia.php on line 52
An error occured while setting up the wiki command

**Operating system:** 
Linux Yunohost 3.7, Nextcloud 18.2

## Talk app

**Talk app version:** 
8.0.5

## Server configuration

**Operating system**: Linux Yunohost 3.7

**Web server:** Nginx

**Database:** MySQL

**PHP version:** 7.3

**Nextcloud Version:** 18.2

**List of activated apps:**

<details>

```

Enabled:
  - accessibility: 1.4.0
  - activity: 2.11.0
  - admin_audit: 1.8.0
  - apporder: 0.9.0
  - audioplayer: 2.9.0
  - calendar: 2.0.2
  - camerarawpreviews: 0.7.3
  - cloud_federation_api: 1.1.0
  - comments: 1.8.0
  - contacts: 3.2.0
  - cookbook: 0.6.3
  - dav: 1.14.0
  - deck: 0.8.0
  - federatedfilesharing: 1.8.0
  - federation: 1.8.0
  - files: 1.13.1
  - files_accesscontrol: 1.8.1
  - files_external: 1.9.0
  - files_pdfviewer: 1.7.0
  - files_rightclick: 0.15.2
  - files_sharing: 1.10.1
  - files_trashbin: 1.8.0
  - files_versions: 1.11.0
  - files_videoplayer: 1.7.0
  - firstrunwizard: 2.7.0
  - logreader: 2.3.0
  - lookup_server_connector: 1.6.0
  - maps: 0.1.6
  - news: 14.1.3
  - nextbackup: 19.12.1
  - nextcloud_announcements: 1.7.0
  - notifications: 2.6.0
  - oauth2: 1.6.0
  - occweb: 0.0.7
  - ocsms: 2.1.7
  - onlyoffice: 4.1.4
  - password_policy: 1.8.0
  - phonetrack: 0.6.2
  - photos: 1.0.0
  - polls: 1.3.0
  - privacy: 1.2.0
  - provisioning_api: 1.8.0
  - quota_warning: 1.7.0
  - recommendations: 0.6.0
  - serverinfo: 1.8.0
  - settings: 1.0.0
[...]

PR #3218, opened 2020-03-31
Resolving the syntax error through #3194

Dids-Himanshu:


nickvergessen:
Actually the problem is Wikimedia broke the api:
https://phabricator.wikimedia.org/T241437

nickvergessen:
I updated it in #3223 to use a new api endpoint

nickvergessen:
Thanks again for your help and sorry we didn't take your PR again :cat: 

Also it seems your other changes are still in here.
You need to reset your branch to the nextcloud master branch before you work on something. Also it's best to create a new branch for everything you are changing, so we can pick individual changes easily.

PR #3233, opened 2020-04-02
Compile some more libs to make IE load

nickvergessen:
Got further until:
> Assignment to read-only properties is not permitted in strict mode.

from line
https://github.com/nextcloud/spreed/blob/3f56ae160638ef37e1040b1f16e544ddddca559a/src/utils/webrtc/simplewebrtc/getscreenmedia.js#L26

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -42,7 +42,7 @@ module.exports = {
 			{
 				test: /\.js$/,
 				loader: 'babel-loader',
-				exclude: /node_modules/,
+				exclude: /node_modules(?!(\/|\\)(@juliushaertl\/vue-richtext|fast-xml-parser|hot-patcher|nextcloud-vue-collections|webdav)(\/|\\))/,
 			},
 			{
 				/**
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,0 +1,12 @@
+module.exports = {
+	presets: [
+		[
+			'@babel/preset-env',
+			{
+				corejs: 3,
+				useBuiltIns: 'usage',
+				modules: false,
+			},
+		],
+	],
+}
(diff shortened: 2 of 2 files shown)


nickvergessen:
Works in IE11 now, lets get this in

webpack.common.js
```diff
@@ -42,7 +42,7 @@ module.exports = {
 			{
 				test: /\.js$/,
 				loader: 'babel-loader',
-				exclude: /node_modules/,
+				exclude: /node_modules(?!(\/|\\)(@juliushaertl\/vue-richtext|fast-xml-parser|hot-patcher|nextcloud-vue-collections|webdav)(\/|\\))/,
```
skjnldsv, line 45: what, you use webdav & hotpatcher here?

nickvergessen: webdav for the file upload
hotpatcher I dont know, it was complained by IE and when I added it here it worked.

skjnldsv: Ah! I was not aware we had file upload yet ^^

skjnldsv: Sorry I was afraid you used the viewer config file ;)

nickvergessen:
```
$ npm ls hot-patcher
talk@9.0.0 /home/nickv/Nextcloud/19/server/appsbabies/spreed
└─┬ webdav@3.2.0
  └── hot-patcher@0.5.0 
```

would it be enough to add `webdav` ?

skjnldsv: No, there is no nested dependencies :)

nickvergessen:
Backport in https://github.com/nextcloud/spreed/pull/3240

PR #3240, opened 2020-04-02
[stable18] Compile some more libs to make IE load

nickvergessen:
backport of #3233

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -47,7 +47,7 @@ module.exports = {
 			{
 				test: /\.js$/,
 				loader: 'babel-loader',
-				exclude: /node_modules/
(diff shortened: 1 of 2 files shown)


webpack.common.js
```diff
@@ -47,7 +47,7 @@ module.exports = {
 			{
 				test: /\.js$/,
 				loader: 'babel-loader',
-				exclude: /node_modules/
+				exclude: /node_modules(?!(\/|\\)(@juliushaertl\/vue-richtext|nextcloud-vue-collections)(\/|\\))/
```
nickvergessen, line 50: Removed things which are not there in stable18

nickvergessen:
Quickly checked this with IE11 on my server, works

PR #3253, opened 2020-04-03
Chat layout improvements

ma12-co:
Various chat improvements

* Increase font size to 15px for better readability;
* Slightly decrease the message width, now it allows ~90 chars per line, winch is still well above the recommendations for readability, but to me it feels like a good compromise. Before it allowed ~110 characters per line;
* Aligned the reply button with the date. Now the date disappears on hover in "1 line" messages and it's shown on taller messages (see gif);

Peek 2020-04-03 12-33


jancborchardt:
Very cool! :)


> Increase font size to 15px for better readability;

Do we want to go slightly larger only in this app? It looks a bit strange, no? If we do go for bigger, then let’s do recommended size of 16px directly, and also do it for every appcontent?

-----------------

Also one issue: Check the left-alignment of the messages you can hover vs the "Write message …" input field and the system messages. The normal messages are a bit shifted to the right cause of the padding caused by the hover-effect, they need to be moved to the left again (possibly using the same negative margin as is used for top and bottom).

ma12-co:
> Also one issue

This is already fixed, check gif

> we want to go slightly larger only in this app? It looks a bit strange, no? If we do go for bigger, then let’s do recommended size of 16px directly, and also do it for every appcontent?

I don't think we should bring all app content to 16px, font sizes should be depend on the purpose of the application. E.g. it wouldn't make much sense to have 16px in files and deck (we tried 15 in the past and it looked so odd we had to dial it back to 14). On the other hand it's very desirable to have 16px in apps with long paragraphs like notes and text (notes is 16px already, text is 14px). 
Talk is somewhere in the middle so I would advise to try 15px in talk first and then maybe up it to 16px in the future.

skjnldsv:
> I don't think we should bring all app content to 16px, font sizes should be depend on the purpose of the application. E.g. it wouldn't make much sense to have 16px in files and deck (we tried 15 in the past and it looked so odd we had to dial it back to 14). On the other hand it's very desirable to have 16px in apps with long paragraphs like notes and text (notes is 16px already, text is 14px).
> Talk is somewhere in the middle so I would advise to try 15px in talk first and then maybe up it to 16px in the future.

I would also be careful on having a per-app font change, looks super weird when using nextcloud :/

ma12-co:
> I would also be careful on having a per-app font change, looks super weird when using nextcloud :/

To clarify I was talking about the AppContent only. The rest of the interface (navigation, sidebar, header ... ) would stay consistent. 
I see your concern @skjnldsv  but check out notes for example, the body of the note works very well at 16px and bringing it down to 14px for the sake of consistency would be an error imo. I think that as long as these decision are informed by the purpose of each app we should be good :)

jancborchardt:
> This is already fixed, check gif

I did, it still looks a few pixels off? :)

-------

Also agree with @skjnldsv that it looks strange to have just slightly different font sizes across apps. The difference between 14 to 16 in Notes is ok as it is a proper contrast. But 14 and 15 gives you this weird feeling of not knowing if it is actually different, and basically makes us look like we wanted to make it the same and it’s just a bug?

ma12-co:
> The difference between 14 to 16 in Notes is ok as it is a proper contrast. 

Lol, the first thing I thought opening notes was the exact opposite: I thought that that was too much contrast xD. In any case we don't need to create a hierarchy between navigation and content, so font-size contrast is not that important here imo. It's more about striking a balance between readability and how much content can fit the screen in the AppContent.
Anyway I'm fine with bringing talk app to 16px, but I would do this only for talk, not for all apps.

skjnldsv:
> Anyway I'm fine with bringing talk app to 16px, but I would do this only for talk, not for all apps.

I would actually leave it like that. Even 16 still looks weird compared to other apps and headers... etc

ma12-co:
> I would actually leave it like that.

Meaning 15 or 16?

skjnldsv:
> Meaning 15 or 14?

default server

ma12-co:
:disappointed:

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -293,6 +297,11 @@ export default {
 			}.bind(this))
 			return richParameters
 		},
+
+		// Determines wether the date has to be displayed or not
```
nickvergessen, line 301:
```suggestion
		// Determines whether the date has to be displayed or not
```

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -370,15 +386,18 @@ export default {
 			justify-self: flex-start;
 			justify-content:  space-between;
 			position: relative;
-			flex: 0 0 $message-utils-width;
 			display: flex;
 			color: var(--color-text-maxcontrast);
-			font-size: 13px;
+			font-size: 14px;
```
nickvergessen, line 391: Do you mean?
```suggestion
			font-size: $chat-font-size;
```

nickvergessen:
Needs a manual backport and a fix to the temporary message width

PR #3271, opened 2020-04-06
Add RoomV2 api which does not add the full user list to the api reponse

nickvergessen:
- [x] Add a v2 api version:
     - [x] without participant list
     - [x] without guest count
     - [x] with canDeleteConversation
     - [x]  canLeaveConversation
- [x] Use it in the web UI
- [x] Update docs for Room API
- [x] Expose capability
- [ ] Make mobile apps use it depending on the capability

Just as a POC for now, but I think this is how we should do such changes. It allows a smooth transition in most cases without breaking the old clients immediatelly.
This also allows backporting such stuff to potentially reduce heavy load in special setups where e.g. mobile apps are not needed.


nickvergessen:
Please review @danxuliu

PR #3286, opened 2020-04-07
Remove unused template parameters

nickvergessen:
The template does not use those variables, so let's remove them for now:
https://github.com/nextcloud/spreed/blob/2f154e5b3fedcd5c3ccde84f9b692725063acf00/templates/index.php

If they are ever used again it will use initial state.


nickvergessen:
Backport in https://github.com/nextcloud/spreed/pull/3291

PR #3306, opened 2020-04-09
Show previews of files in quotes

### LINKED ISSUE #3283: Missing preview in quoted messages
1) User "A" sends a funny image
2) User "B" quotes the message and writes a reply after a while
3) User "A" is confused because he got a reply along with a cryptic file name
4) ???

Expected result: User A should see the preview of an image (original message) to understand the context of the message.

nickvergessen:
Bildschirmfoto von 2020-04-09 10-51-46

Fix #3283

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
@@ -27,11 +27,11 @@
 		rel="noopener noreferrer"
 		@click="showPreview">
 		<img v-if="!isLoading && !failed"
-			class="preview"
+			:class="previewSizeClass"
 			alt=""
 			:src="previewUrl">
 		<img v-if="!isLoading && failed"
-			class="preview"
+			:class="previewSizeClass"
 			alt=""
 			:src="defaultIconUrl">
 		<span v-if="isLoading"
@@ -74,6 +74,10 @@ export default {
 			type: String,
 			default: 'no',
 		},
+		previewSize: {
+			type: Number,
+			default: 128,
+		},
 	},
 	data() {
 		return {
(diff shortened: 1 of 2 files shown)


nickvergessen:
### Follow up proposal

> Bildschirmfoto von 2020-04-09 10-51-46

so my suggestions would be:
1. Show the name next to the file instead of bellow (to save horizontal space)
2. Show a hint to enlarge by clicking the thumbnail when the viewer is there

Maybe we can even reduce the default preview size to 64 when the viewer is supported, as the 128px preview is useless in most cases anyway and you will enlarge it

nickvergessen:
### Follow up proposal

> Bildschirmfoto von 2020-04-09 10-51-46

so my suggestions would be:
1. Show the name next to the file instead of bellow (to save horizontal space)
2. Show a hint to enlarge by clicking the thumbnail when the viewer is there

Maybe we can even reduce the default preview size to 64 when the viewer is supported, as the 128px preview is useless in most cases anyway and you will enlarge it

cc @nextcloud/designers-talk

ma12-co:
A couple of comments (already discussed with @nickvergessen ):
- I would not make the image in the quote previewable, clicking anywhere in the quote should bring you to the original message with the bigger image. There you can click and preview attachments. 
- I'd avoid the click hint, let's be clear about the clickability with UI and feedbacks

mario:
Nice, thanks.

jancborchardt:
> A couple of comments (already discussed with @nickvergessen ):
> 
> * I would not make the image in the quote previewable, clicking anywhere in the quote should bring you to the original message with the bigger image. There you can click and preview attachments.
> 
> * I'd avoid the click hint, let's be clear about the clickability with UI and feedbacks

Yep, totally agree on that.

On the other points:

> Show the name next to the file instead of bellow (to save horizontal space)

I’d say let’s not show the filename at all, it is useless in most cases. With Viewer support it will be shown in there anyway if people want to see it.

> Maybe we can even reduce the default preview size to 64 when the viewer is supported, as the 128px preview is useless in most cases anyway and you will enlarge it

If you are talking about general image display and not the quoting (where smaller size is ok): Images should be big, and most should not require opening them with the Viewer. Check out the other chat apps, and they show images so that mostly you don’t even have to enlarge it (e.g. Signal: stickers are 200*200px, and images always have a width of 300 with height accordingly).
I’d be careful with the assumption "you will enlarge it" because it is another annoying step, and the Viewer also has some additional load time as well.

nickvergessen:
> and most should not require opening them with the Viewer.

Bad news, our preview api cuts everything in squares

> > Show the name next to the file instead of bellow (to save horizontal space)
> 
> I’d say let’s not show the filename at all, it is useless in most cases. With Viewer support it will be shown in there anyway if people want to see it.

Well that's only for images I guess. When we don't have a preview, but only the mimetype icon and files that don't show good in thumbnails I think the name still makes sense?

But okay, then let's merge it with this tiny improvement and discuss all the other things in a separate issue.

mario:
@nickvergessen so are we hiding the filename if there's a preview available? Cause I'd want this to happen on mobile as well then.

nickvergessen:
We dont change anything for now, just add the preview in quotes and then discuss further steps

mario:
Awesome.

Issue #2196, opened 2019-09-18
📝 Note to self - Overview


jancborchardt:
For quickly sharing links or notes between your devices, a "Note to self" entry in the conversations is cool.

Signal for example has this. Technically it used to be the case when you sent a message to your own number, and now it's directly called "Note to self".

For us it could be a default conversation called "Note to self ✅", and as picture it could use the ["pencil" icon](https://pictogrammers.com/library/mdi/icon/pencil/) in white on the blue gradient background, similar like we have it for the "Talk updates ✅" conversation.

The issue of quickly exchanging text content between devices or desktop and mobile was also brought up by @tobiasKaminsky @jakobsack and this is a cool solution for that. :)

- [x] API https://github.com/nextcloud/spreed/issues/10404
- [x] Frontend https://github.com/nextcloud/spreed/issues/10405
- [x] Android https://github.com/nextcloud/talk-android/issues/786
- [x] iOS https://github.com/nextcloud/talk-ios/issues/300

nickvergessen:
Yeah basically a designated group conversation with just you.

marcusquinn:
Rocket Chat also does this. It's both useful and nice to be able to *Talk* to oneself and get a sensible answer 😁

marcusquinn:
Just realised, you can only do this with the `+` sign for a new conversation, and only adding yourself.

However, you can't search your own name in the "Search conversations or users" input, which is what most users have assumed should be the way to do this.

@mario Chance we can make your own name return in the search results and initiate this 1-2-1 conversation with oneself that way too?

nickvergessen:
Mario is not with the project anymore.
The thing is we would like to have more than a normal conversation with ourselves.

Yiannis128:
This would be a nice quality of life feature.

jancborchardt:
Added specification details on the top:

> For us it could be a default conversation called "Note to self ✅", and as picture it could use the ["pencil" icon](https://pictogrammers.com/library/mdi/icon/pencil/) in white on the blue gradient background, similar like we have it for the "Talk updates ✅" conversation.

jancborchardt:
@nickvergessen yes on the first 3 sounds good, yep.

And for the last one the special handling as with Talk updates sounds good. Manual restore is a good question, I think it's probably fine to not have that first, otherwise we are going into overengineering territory.

nickvergessen:
@jancborchardt should there be a predefined chat messages be posted in the room with some kind of explanation text?
[...]

PR #3320, opened 2020-04-13
Notes

mario:
Fixes https://github.com/nextcloud/spreed/issues/2196

Early feedback appreciated.


mario:
I'm not sure why the tests here are failing so any hint would be welcome. Maybe they're failing in the master as well? Dunno :)

nickvergessen:
Can you rebase on latest master? then the icon should be working I guess?

mario:
Done.

mario:
> Nice :)
> 
> Only one thing, I see that the notes conversation is favourite by default, I wouldn't do that..

Only the first time. If you delete the conversation, once it's re-created it won't be favorite. I copied this part of the experience from some of the leading messengers in the market.

nickvergessen:
> Only the first time. If you delete the conversation, once it's re-created it won't be favorite. 

The thing is a refresh of the room list will automatically readd it, so you can get rid of it for like 29 seconds at best. So we could check the user config instead also for the creation and simply never recreate the conversation instead.

The other thing I would change is using yourself as an author of the sample messages. It has the high advantage, that you can 1. reply to the messages to test things out to begin with and 2. there is less special handling needed in the mobile clients (one less author type to be aware of).

We could also use your own avatar instead? at least the blue = is a bit heavy for me. Would require theming maybe? For the changelog I think its okayish as it is the talk app icon

mario:
> > Only the first time. If you delete the conversation, once it's re-created it won't be favorite.
> 
> The thing is a refresh of the room list will automatically readd it, so you can get rid of it for like 29 seconds at best. So we could check the user config instead also for the creation and simply never recreate the conversation instead.

Yes, but it will not be favorite anymore. That's what I was pointing at.

> 
> The other thing I would change is using yourself as an author of the sample messages. It has the high advantage, that you can 1. reply to the messages to test things out to begin with and 2. there is less special handling needed in the mobile clients (one less author type to be aware of).
> 

I guess so, but this makes more sense to me - an explanation of how to use it. I'm not emotionally attached to the current ways though if there are more people who think we should change it.

> We could also use your own avatar instead? at least the blue = is a bit heavy for me. Would require theming maybe? For the changelog I think its okayish as it is the talk app icon

That would be scary :D Anyway, the "blue" should ideally be gradient like our real icon has (and same is true for the changelog icon).

mario:
As for not recreating it on room refresh: that would mean we need complicated logic in sharing from other apps to always show this conversation since it might not exist in the first place.

lib/Chat/SpecialRoom/Manager.php
```diff
@@ -72,13 +72,13 @@ public function userHasNewChangelog(string $userId): bool {
 	public function createNotesIfNeeded(string $userId): void {
 		$room = $this->roomManager->getSpecialRoom($userId, Room::NOTES_CONVERSATION);
 
-		if ($this->getNotesForUser($userId) === (int) 0) {
+		if ($this->getNotesForUser($userId) === 0) {
```
mario, line 75: I think this will fail. It will think 0 is a string afaik. I originally had == which worked just fine but then lint complained.

nickvergessen:
> As for not recreating it on room refresh: that would mean we need complicated logic in sharing from other apps to always show this conversation since it might not exist in the first place, and a way to create it automagically.

Well you deleted it, so it's gone and of course no longer available as a sharing option. If you want it again, create a new conversation and invite no one. ¯\_(ツ)_/¯ sounds good enough to me.

mario:
> > As for not recreating it on room refresh: that would mean we need complicated logic in sharing from other apps to always show this conversation since it might not exist in the first place, and a way to create it automagically.
> 
> Well you deleted it, so it's gone and of course no longer available as a sharing option. If you want it again, create a new conversation and invite no one. ¯_(ツ)_/¯ sounds good enough to me.

That sounds very, very odd :-/

lib/Chat/SpecialRoom/Manager.php
nickvergessen: Lint complains because `==` is illegal nowerdays :speak_no_evil: 

```php
public function getNotesForUser(string $userId): int {
```

it will always return an integer (you defined it like that correctly) and so you need to compare it to a integer `0` is always an integer, `'0'` would be a string.

nickvergessen:
> That sounds very, very odd :-/

Deleting something and have it popup a split second later with HPB and <30s without is more odd

lib/Chat/SpecialRoom/Manager.php
mario: When I had it that way the comparison result was always false :-/ Maybe it was just my mistake.

ma12-co:
> Deleting something and have it popup a split second later with HPB and <30s without is more odd

Let's not go for the least odd :p

A nice behavior here would be: 

* Provide the notes_to_self conversation by default;
* If user deletes it never show it again;
* Provide a checkbox in the settings at the bottom of the conversations list to toggle the notes to self element back;

It would also be nice to tell the user that there's a toggle in the settings to bring the notes back upon deletion.

nickvergessen:
Should be ready for review now

nickvergessen:
This is currently so much special code being added duplicating the hack we did for the changelog.
I would propose to do it for 20 properly instead.

Thovi98:
> We never use the "my" form of referring to things (we also try to avoid the "your" and only use it when not avoidable). This is why in the issue it’s specifically called "Note to self". Let’s use that wording.

Hello @jancborchardt ! Just for information, may I ask the reason behind this ? :)

jancborchardt:
> > We never use the "my" form of referring to things (we also try to avoid the "your" and only use it when not avoidable). This is why in the issue it’s specifically called "Note to self". Let’s use that wording.
> 
> Hello @jancborchardt ! Just for information, may I ask the reason behind this ? :)

There’s a lot of discussion and writing about using "your *" or "my *" in interfaces, and this is simply our convention. :)

PR #3321, opened 2020-04-14
Fix broken conversation icons and CSS

### LINKED ISSUE #3319: Missing Talk Updates  icon
Screenshot 2020-04-13 at 21 00 07

nickvergessen:
Fix #3319

### DIFF
--- a/css/icons.scss
+++ b/css/icons.scss
@@ -9,7 +9,7 @@
 @include icon-black-white('lobby', 'spreed', 1);
 @include icon-black-white('text', 'filetypes', 1, true);
 
-.app-Talk,
+.app-talk,
 .talk-modal,
 #talk-sidebar,
 #call-container,


css/icons.scss
```diff
@@ -9,7 +9,7 @@
 @include icon-black-white('lobby', 'spreed', 1);
 @include icon-black-white('text', 'filetypes', 1, true);
 
-.app-Talk,
+.app-talk,
```
skjnldsv, line 12: ahahahha!
That's why you never use capital letters in class/id

nickvergessen: Well, then enforce this on your property?
https://github.com/nextcloud/spreed/commit/72edd115cccf18494ce54d08cf3cc801141586d7
*cough*

It works well. But of course if you just change the template and not the css it doesn't match anymore?

PR #3329, opened 2020-04-14
Move to PSR2 coding style as all the other code does

nickvergessen:
Let's wait with this until right before stable19 is branched of, and then:
1. Rebase the PR and drop the commit "Auto fix PHP CS"
2. Run `composer run cs:fix`
3. Commit the changes
4. Force push and merge here

Ref https://github.com/nextcloud/coding-standard/#installation


composer.json
```diff
@@ -0,0 +1,9 @@
+{
+	"scripts": {
+		"cs:check": "php-cs-fixer fix --dry-run",
```
nickvergessen, line 3:
```suggestion
		"cs:check": "php-cs-fixer fix --dry-run --diff",
```
https://github.com/nextcloud/server/pull/20502

nickvergessen:
Rebased and ran the script, so stable19 is also in a fixed state when we branch it off

PR #3330, opened 2020-04-14
Call view revamp

ma12-co:
#### This pr:
- [x] Make grid auto resize (columns and rows number) based on available space and videos dimensions;
- [x] Add ability to toggle between grid and promoted view;
- [x] Add grid navigation;
- [x] Add ability to set a cap;
- [x] Add developer mode;
- [x] Add docs;
- [x] Rebuild grid when navigation is toggled;
- [x] Rebuild grid when sidebar is toggled;
- [x] Add tooltip hint to switch to promoted view when there's a overflow of videos;
- [x] Modify video components so that they adapt to the grid;
- [x] Flip only own video
- [x] Navigation icon color;
- [x] Fade transition between chatview and callview;
- [x] Don't show controls unless hovered;
- [x] Show name on hover as well;
- [x] When hint is dismissed, don't re-display the warning on resize;
- [x] Hint typo "don'd -> don't"
- [x] Bacgrounds
- [x] Missing promoted view icon
- [x] Highlight speakers
- [ ] Disable video button does not change to "disabled"
- [x] Always show the name if no video
- [x] Guest background color is not gray when they choose a name 
- [x] Darken a bit all backgrounds for people with no video to mimick the difference that cards with
pictures have
- [x] Crop the video only in one axis
- [x] Full-screen button is broken 
- [x] Show all the names and all the controls on mousemove for some seconds
- [x] Screen sharing is broken;
- [ ] Bring the grid to the promoted view; 
- [ ] Allow to navigate through pages in the promoted view;

#### Follow-up: 
- [x] Figure out what's going on in screenshot 1 https://github.com/nextcloud/spreed/pull/3330#issuecomment-619094426
- [ ] Show icons and names always if videos are big enough;
- [ ] Full support for grid navigation;
    - [ ] Rebuild grid when navigating into last page;
    - [ ] Rebuild grid when navigating away from last page;
    - [ ] Add tooltip on navigation buttons saying e.g. "Show next 10 videos";
    - [ ] Properly style navigation buttons and dots;
    - [ ] Make dots clickable and not a mere reference for pages;
- [ ] Complete target aspect ratio feature;
- [ ] Make dots clickable and not a mere reference for pages;


src/components/GridView/GridView.vue
```diff
@@ -0,0 +1,187 @@
...
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+-->
+
+<template>
+	<div
+		class="grid"
+		:style="{ gridTemplateColumns: `repeat(${columns},  minmax(${minWidth}px, 1fr))`, gridTemplateRows: `repeat(${rows}, minmax(${minHeight}px, 1fr)))`}">
```
nickvergessen, line 25: move to computed? :P

ma12-co: sure!

nickvergessen:
Only your own video was flipped and should be flipped

juliushaertl:
Two things i noticed during our test call today:
- The cropping seems to take a smaller part of the video than necessary, I had a quick look at the css and it has width/height both set to auto. when setting the width to 100% it was fine for a portrait video. Not sure if there was some discussion about this?
- Maybe sorting people with video first would also be nice to have your eyes rest a bit more in one area of the screen if e.g. half of the people joined have video and the other half not.

src/components/CallView/CallView.vue
```diff
@@ -10,612 +12,54 @@
...
-				:shared-data="sharedDatas[callParticipantModel.attributes.peerId]" />
-		</div>
+	<div class="call-view">
+		<transition name="fade">
+			<GridView v-if="isGrid"
+				v-bind="$attrs"
+				:dev-mode="false"
+				:dummies="2" />
+		</transition>
+		<transition name="fade">
```
skjnldsv, line 30: Why not in the same transiton with v-if/v-else? :thinking:

ma12-co:
> The cropping seems to take a smaller part of the video than necessary, I had a quick look at the css and it has width/height both set to auto. when setting the width to 100% it was fine for a portrait video. Not sure if there was some discussion about this?

Thanks for reporting Julius, I'll check this out. 

>  Maybe sorting people with video first would also be nice to have your eyes rest a bit more in one area of the screen if e.g. half of the people joined have video and the other half not.

Interesting, maybe we could even display the participants without video as smaller elements in the ui? This is further down the road though

ma12-co:
> The cropping seems to take a smaller part of the video than necessary, I had a quick look at the css and it has width/height both set to auto. when setting the width to 100% it was fine for a portrait video. Not sure if there was some discussion about this?

@juliushaertl  I checked this and it seems fine to me. Basically it's the least zoomed in possible while filling the whole frame. This is done with `min-width: 100%` and `min-height: 100%`

src/components/CallView/shared/VideoBackground.vue
```diff
@@ -0,0 +1,85 @@
...
+			type: String,
+			default: '?',
+		},
+	},
+
+	computed: {
+		backgroundColor() {
+			// If the prop is empty. We're not checking for the default value
+			// because the user's displayName might be '?'
+			if (!this.$options.propsData.displayName) {
```
skjnldsv, line 46: I'm not sure I understand this correctly.
Why put a default to '?' and not use it in the end?

skjnldsv: Also, should be `this.$props`, not `this.$options.propsData` :wink:

src/components/CallView/shared/VideoBackground.vue
```diff
@@ -0,0 +1,85 @@
...
+				return `var(--color-text-maxcontrast)`
+			} else {
+				const color = this.displayName.toRgb()
+				return `rgba(${color.r}, ${color.g}, ${color.b}, 1)`
```
skjnldsv, line 50:
```suggestion
				return `rgb(${color.r}, ${color.g}, ${color.b})`
```

src/components/CallView/shared/VideoBackground.vue
ma12-co: this.$options.propsData tells me if the prop  has been passed in. So I can differentiate between my default "?" (gray bkg) and a passed in "?" displayname (toRgb color)

ma12-co: Anyway in the end I think I will just show the avatar all the time

skjnldsv: Ah right, I see the difference.
Not sure if it's a reactive property though.

nickvergessen:
So there seem to be some weird angles where the calculation causes "unexpected" layouts:
Bildschirmfoto von 2020-04-24 17-24-37
Bildschirmfoto von 2020-04-24 17-25-13
1322-620-max

ma12-co:
>So there seem to be some weird angles where the calculation causes "unexpected" layouts:

Only layout in screenshot 1 is unexpected @nickvergessen

nickvergessen:
So in the last screenshot, when you open the chat, it jumps to 2*2
So I wonder if we should prefer that?

ma12-co:
> So in the last screenshot, when you open the chat, it jumps to 2*2

Yes because if the size of the whole grid changes (upon sidebar open/close,  navigation open/close, window resize) the grid is rebuilt to give the best possible aspect ratios and sizes given the new grid dimensions. Otherwise in some cases if we don't rebuild we could end up with again too stretched elements.

> Well we could also just have the black at the left?

This would be tricky, the grid slots are populated from top left to bottom right. We could look into it but I think it might be a lot of work...

src/components/CallView/shared/VideoBackground.vue
```diff
@@ -23,35 +23,60 @@
...
 			}
-
 		},
+		backgroundImage() {
+			return `/avatar/${this.user}/500`
+		},
+	},
+
+	async beforeMount() {
+		const response = await axios.get(generateUrl(`avatar/${this.user}/1`))
```
skjnldsv, line 74: Is this component used for guests too?
If so there is a different endpoint :)

src/components/CallView/shared/VideoBackground.vue
```diff
@@ -64,6 +89,22 @@ export default {
 	top: 0;
 	height: 100%;
 	width: 100%;
+	&__picture {
+		filter: blur(20px);
```
skjnldsv, line 93: @Ivansss what is the blur intensity you're using on iOS?
I remember checking this last year at Stuttgart with you & Jan :)

Ivansss: On iOS you need to set the inputRadius of the filer.
We finally set it to:
`CGFloat inputRadius = 8.0f;`
But I don't know how to convert that into blur pixels :S

mario: I use 5.0f. Should I set it to 8?

skjnldsv: I think we should be consistent yes :)

skjnldsv: > But I don't know how to convert that into blur pixels :S

Hum, good question. Both are gaussian blur. Is this a percentage for you @Ivansss, @mario ?

skjnldsv: For css: "blur: Applies a Gaussian blur to the input image. The value of ‘radius’ defines the value of the standard deviation to the Gaussian function"

skjnldsv: https://developer.apple.com/documentation/coreimage/cigaussianblur?language=objc

According to the docs, the radius is also in px, but you're passing it as float, so we should use 8px as well I assume :)

skjnldsv: Can we compare on the same image so we make sure this is the same? Can you post a screenshot @Ivansss ?

ma12-co: 8px is way too low here. We have overlay text and controls and the beckground could become very distracting. I don't see the need of matching the mobile radius here

src/components/CallView/shared/VideoBackground.vue
ma12-co: No it's not @skjnldsv. @nickvergessen , should we start to use the proper guest endpoint instead of making all of them gray?

src/components/CallView/shared/VideoBackground.vue
Ivansss: This is how it looks on iOS:
IMG_0281

Ivansss: This has an `inputGradient = 8.0` and a black top layer with `alpha = 0.6`

skjnldsv: > 8px is way too low here

Agreed @ma12-co, I guess this doesn't translate properly.
Can you tweak things a bit so you're reaching the similar design for iOS ?
Maybe it's factor 10 ? 80px? Only god knows! :grin:

skjnldsv:
Is this really `3. to review` ?

nickvergessen:
Fixed fullscreen target:  The content ID was changed in
https://github.com/nextcloud/nextcloud-vue/commit/33f4af39f14648429650db6ca3319786188ec3a6

src/components/CallView/GridView/GridView.vue
```diff
@@ -0,0 +1,601 @@
...
+.previous {
+	left: 20px;
+}
+
+.pages-indicator {
+	position: absolute;
+	right: 50%;
+	top: 4px;
+	display: flex;
+	background-color: var(--color-bakground-hover);
```
nickvergessen, line 583:
```suggestion
	background-color: var(--color-background-hover);
```
??

src/components/CallView/PromotedView/PromotedView.vue
```diff
@@ -0,0 +1,474 @@
...
+	 * screen it is not too bad and it is better than compressing even further
+	 * the shared screen. */
+	overflow: visible;
+}
+
+::v-deep video {
+	z-index: 0;
+	max-height: 100%;
+	/* default filter for slightly better look */
+	-webkit-filter: contrast(1.1) saturate(1.1) sepia(.1);
```
nickvergessen, line 255: nooooooooooooo

Please apply #3413

ma12-co: xD

src/views/MainView.vue
```diff
@@ -94,17 +111,20 @@ export default {
...
+			// FIXME reset the settings so we can check it later on if loading is finished
 		},
 	},
 	*/
 }
 </script>
 
 <style lang="scss" scoped>
-.mainView {
+@import '../assets/variables.scss';
```
nickvergessen, line 123: not used?

src/components/CallView/PromotedView/PromotedView.vue
```diff
@@ -0,0 +1,477 @@
...
+	z-index: 1;
+}
+.videoContainer.promoted ::v-deep video,
+.participants-2 .videoContainer:not(.videoView) ::v-deep video {
+	position: absolute;
+	width: initial;
+	height: 100%;
+	left: 50%;
+	top: 50%;
+	transform: translateY(-50%) translateX(-50%);
```
skjnldsv, line 354: You should center with `object-cover`, not `transform`

src/components/CallView/shared/LocalVideo.vue
```diff
@@ -175,7 +193,38 @@ export default {
...
+
+	/* Setting width & height to auto prevents the browser from stretching or squishing the video */
+	width: auto;
+	height: auto;
+
+	/* Center the video and flip horizontally */
+	position: absolute;
+	top: 50%;
+	left: 50%;
+	transform: translate(-50%,-50%) rotateY(180deg) !important;
```
skjnldsv, line 223: Same as the other comment I guess, object-cover :)

src/App.vue
```diff
@@ -432,4 +432,9 @@ export default {
 		}
 	}
 }
+
+#app-content-vue {
```
skjnldsv, line 436: Or use tha class directly
`.app-content`

src/components/CallView/PromotedView/PromotedView.vue
ma12-co: Form what I read in mdn this is unrelated

src/App.vue
```diff
@@ -422,7 +422,7 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-#content {
+#content-vue {
```
skjnldsv, line 425: Same, classes might be cleaner

src/components/CallView/PromotedView/PromotedView.vue
skjnldsv: That is not what I mean, good habit in css is to avoid absolute to the max you can. Additionally, trasnfrom is a browser rendering function that is expensive. Centering with the dedicated object-fit function is what you need to do.

skjnldsv: Please not though that it doesn't work with ie, which is fine, you will just see black bars. But it's ie and we deprecate for nc20
https://caniuse.com/#feat=object-fit

ma12-co: Fit modifies height and width of the element, there I only need to center it into it's container so this doesn't seem to be an alternative.. Height and width of the video are computed manually depending on the video aspect ratio and the parent aspect ratio.

ma12-co: This is how the computation is done: https://github.com/nextcloud/spreed/blob/df4570615b87792b240557e60e1cb0e6013d306d/src/components/CallView/shared/video.js#L64

src/components/CallView/shared/VideoBackground.vue
```diff
@@ -72,10 +72,15 @@ export default {
 	},
 
 	async beforeMount() {
-		const response = await axios.get(generateUrl(`avatar/${this.user}/1`))
-		if (response.headers[`x-nc-iscustomavatar`] === '1') {
-			this.hasPicture = true
+		try {
+			const response = await axios.get(generateUrl(`avatar/${this.user}/1`))
```
skjnldsv, line 76: why `/1` ?
Why not get the full image directly for the cache?

ma12-co: How should I do this? Can I use that image in the template instead of the url?

nickvergessen: Well reusing is one thing, but if you query /128 it might read the data from local cache?

ma12-co: I'm just checking if it has a picture or not here, that's why I was trying to keep it as small as possible. But if I can use this image in the template somehow, It would make sense to have it bigger

skjnldsv: Yep, cache is working.
Just fetch directly the image you need.
(if you try with dev tools opened, it might fetch it twice because of the cache disabled option)

PR #3335, opened 2020-04-16
Don't mark the window/tab unread for own messages

### LINKED ISSUE #3256: Own messages should not update the `*` flag for the tab
## Steps to reproduce
1. Write a chat message
2. Quickly switch to another tab
3. After a refresh of the room (list) the tab is marked "unread" with the `*` prefix

### Expected behaviour
Own messages need to be excluded from the list

### Actual behaviour
Own messages mark the talk tab unread

nickvergessen:
Otherwise when you write a message and immediately navigate away,
the next refresh of the room list will always tell you that messages
have been posted.

### Steps

1. Create a conversation
2. Post a chat message and switch to another tab before the room list is refreshed

### Before

Tab is marked with `*` after the room list is refreshed

### After

Tab is not marked.

Fix #3256

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -122,7 +122,19 @@ export default {
 
 			const lastMessage = {}
 			conversationList.forEach(conversation => {
-				lastMessage[conversation.token] = 0 + (conversation.lastMessage && conversation.lastMessage.id ? conversation.lastMessage.id : 0)
+				lastMessage[conversation.token] = 0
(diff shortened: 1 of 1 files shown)


nickvergessen:
Added one more commit that should also fix that case.

nickvergessen:
> With the new commit, why is the special handling for own messages still needed? 

The special handling makes sure your own messages never triggered it, even when sent on another device e.g.

PR #3338, opened 2020-04-16
Die gracefully when user joins again in another tab/device

nickvergessen:
### Steps
1. In Tab 1 open a conversation as User A and join the call
2. In Window 2 open the same conversation as User B and join the call
3. In Tab 2 or Device 2 open the conversation as User 1 and leave it again
4. :boom: 

- [ ] Based on #3591
- [x] `addParticipantOnce` used a wrong check to see if a user already exists.
- [x] Internal signaling stops the signaling query and is just silent from there on in the old tab
- [ ] External signaling is in limbo state and people joining later will get a different set of participants compared to users that were online all the time

### DIFF
--- a/docs/participant.md
+++ b/docs/participant.md
@@ -109,18 +109,28 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
     field | type | Description
     ------|------|------------
     `password` | string | Optional: Password is only required for users which are of type `4` or `5` and only when the conversation has `hasPassword` set to true.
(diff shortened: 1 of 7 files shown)


nickvergessen:
### Internal signaling

> Bildschirmfoto von 2020-04-16 16-26-32

btw the `OC.dialogs.confirmDestructive` could really use a blurred/overlayed background as nothing should be done until it has been dealt with.

nickvergessen:
Now merged into #3591

PR #3348, opened 2020-04-18
Make leave-call button red

ma12-co:
Peek 2020-04-18 14-41


nickvergessen:
This will make the dark mode icon color issue even more difficult to solve 😬

ma12-co:
> This will make the dark mode icon color issue even more difficult to solve

Worth the hassle imo: every "leave call" button in the world is red, so here we have the opportunity to make our UI clearer by sticking to the pattern

PR #3357, opened 2020-04-20
Add `promoted-view` icon

ma12-co:
Screenshot from 2020-04-21 12-20-08

### DIFF
--- a/css/icons.scss
+++ b/css/icons.scss
@@ -68,6 +68,9 @@
 		&.icon-screen-off {
 			background-image: url(icon-color-path('screen-off', 'actions', 'fff', 1, true));
 		}
+		&.icon-promoted-view{
+			background-image: url(icon-color-path('promoted-view', 'actions', 'fff', 1, true));
+		} 
 	}
 
 	.icon-favorite {
@@ -114,3 +117,4 @@
 		background-image: url(icon-color-path('close', 'actions', 'fff', 1, true));
 	}
 }
+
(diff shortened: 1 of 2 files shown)


css/icons.scss
```diff
@@ -8,6 +8,8 @@
 @include icon-black-white('emoji-smile', 'spreed', 1);
 @include icon-black-white('lobby', 'spreed', 1);
 @include icon-black-white('text', 'filetypes', 1, true);
+@include icon-black-white('promoted-view', 'actions', 1, true);
+@include icon-black-white('promoted-view-white', 'actions', 1, true);
```
skjnldsv, line 12: This scss function generate white icons, you do not need both ?
If it's always on top of the main call dark background, shouldn't you be using the white version only?

nickvergessen: Actually this should be in the `.forced-white {` block couple of lines down, as it will be always white (independent from dark mode) like the video/audio controls.

img/promoted-view.svg
```diff
@@ -0,0 +1,6 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
nickvergessen, line 1: Please run:
```
mv img/promoted-view.svg img/promoted-view.svg.opttmp
scour --create-groups --enable-id-stripping --enable-comment-stripping --shorten-ids --remove-metadata --strip-xml-prolog --no-line-breaks  -i img/promoted-view.svg.opttmp -o img/promoted-view.svg
rm img/promoted-view.svg.opttmp
```

img/promoted-view-white.svg
```diff
@@ -0,0 +1,6 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
nickvergessen, line 1: should not be needed, as our svg controller takes care of coloring

PR #3370, opened 2020-04-21
Only show the single file for now, so people are not scared all their…

nickvergessen:
… root items are shared


nickvergessen:
Regression: Breaks the Text app.

Issue #2913, opened 2020-02-07
duplicate avatars


r2evans:
### Steps to reproduce
1. Connect, start a conversation
2. Network glitch, see "Network Error", then reconnect
3. See "Current participant found", see a duplicate grayed-out avatar (current user)

### Expected behaviour
No dupes

### Actual behaviour
Dupe

## Server configuration detail

**Operating system:** Linux 4.4.0-171-generic #200-Ubuntu SMP Tue Dec 3 11:04:55 UTC 2019 x86_64

**Webserver:** Apache/2.4.38 (Debian) (apache2handler)

**Database:** pgsql PostgreSQL 11.6 (Debian 11.6-1.pgdg90+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 6.3.0-18+deb9u1) 6.3.0 20170516, 64-bit

**PHP version:** 

7.3.14
Modules loaded: Core, date, libxml, openssl, pcre, sqlite3, zlib, ctype, curl, dom, fileinfo, filter, ftp, hash, iconv, json, mbstring, SPL, PDO, session, posix, Reflection, standard, SimpleXML, pdo_sqlite, Phar, tokenizer, xml, xmlreader, xmlwriter, mysqlnd, apache2handler, apcu, exif, gd, gmp, imagick, intl, ldap, memcached, pcntl, pdo_mysql, pdo_pgsql, redis, sodium, zip, Zend OPcache

**Nextcloud version:** 18.0.0 - 18.0.0.10

**Updated from an older Nextcloud/ownCloud or fresh install:** 

**Where did you install Nextcloud from:** docker, 18.0.0-apache (b81821f761f0)

<details><summary>Signing status</summary>

Array
(
)

</details>

<details><summary>List of activated apps</summary>

```
Enabled:
 - accessibility: 1.4.0
 - activity: 2.11.0
 - admin_audit: 1.8.0
 - calendar: 2.0.1
 - cloud_federation_api: 1.1.0
 - comments: 1.8.0
 - contacts: 3.1.8
 - dav: 1.14.0
 - federatedfilesharing: 1.8.0
 - federation: 1.8.0
 - files: 1.13.1
 - files_external: 1.9.0
 - files_pdfviewer: 1.7.0
 - files_rightclick: 0.15.2
 - files_sharing: 1.10.1
 - files_trashbin: 1.8.0
 - files_versions: 1.11.0
 - files_videoplayer: 1.7.0
 - impersonate: 1.5.0
[...]

PR #3382, opened 2020-04-22
User participants are identified by userId in the participant store, …

nickvergessen:
…not by participant

Split from https://github.com/nextcloud/spreed/pull/3338 for faster merging

Not sure where `participant` comes from. I guess a copy paste fail.
userId is used in the rooms getParticipant:
https://github.com/nextcloud/spreed/blob/445756dea3356a3f2d39ad00ad581ebb4c86bda2/lib/Controller/RoomController.php#L851
And the calls getPeers:
https://github.com/nextcloud/spreed/blob/06a045ed1e82facb477caedd918200cdd4c34ea9/lib/Controller/CallController.php#L66

The old Room API also had no `participant` key:
https://github.com/nextcloud/spreed/blob/445756dea3356a3f2d39ad00ad581ebb4c86bda2/lib/Controller/RoomController.php#L338

This should explain and solve some of the "duplicate avatar" issues like #2913

### DIFF
--- a/src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
+++ b/src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
@@ -224,7 +224,7 @@ export default {
 				}
 			} else {
 				data = {
-					participant: this.computedId,
+					userId: this.computedId,
 				}
 			}
 			return data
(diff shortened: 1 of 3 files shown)


nickvergessen:
Fixed it

nickvergessen:
Also checked the others and the only other endpoint with "participant" was already getting it from the userId: https://github.com/nextcloud/spreed/blob/2c6e88cfd52a79d3fb8ba657dd5940cdfbd21be9/src/services/participantsService.js#L104-L108

PR #3398, opened 2020-04-27
Make callbutton green when there's an ongoing call

ma12-co:
Add green state to the call button for when there's an ongoing call. This is another visual aid for knowing that there's an ongoing call in the conversation.
The button states would be: 

Screenshot from 2020-04-27 09-41-03

Screenshot from 2020-04-27 09-40-51

Screenshot from 2020-04-27 09-41-15

 cc @nextcloud/designers


src/components/TopBar/CallButton.vue
```diff
@@ -151,6 +152,11 @@ export default {
 			return this.conversation.readOnly === CONVERSATION.STATE.READ_WRITE
 				&& this.participant.inCall !== PARTICIPANT.CALL_FLAG.DISCONNECTED
 		},
+		callButtonStyle() {
+			if (this.conversation.hasCall && !this.isBlockedByLobby) {
+				return { backgroundColor: `#5FDA34`, border: `1px solid #5FDA34` }
```
skjnldsv, line 157: `var(--color-success)`

src/components/TopBar/CallButton.vue
```diff
@@ -151,6 +152,11 @@ export default {
...
+			} else return ''
```
skjnldsv, line 158: No inline else :see_no_evil:

ma12-co: Eslint rule possible?

src/components/TopBar/CallButton.vue
ma12-co: @skjnldsv  can we make var(--color-sucecss) a bit more vibrant? I think it's too dull, at least for this purpose...
Screenshot from 2020-04-27 10-14-22

src/components/TopBar/CallButton.vue
```diff
@@ -29,6 +29,7 @@
 			html: true
 		}"
 		:disabled="startCallButtonDisabled || loading || blockCalls"
+		:style="callButtonStyle"
```
nickvergessen, line 32: use a class instead?

src/components/TopBar/CallButton.vue
nickvergessen: Yes please

src/components/TopBar/CallButton.vue
skjnldsv: Sure! I think it's best to keep the same green on between our apps, but we can definitely improve the default green here! :+1:

ma12-co: Ok!

ma12-co:
So as discussed yesterday in the server pr, I'm adding the colors here as it's likely the only place they'll be used.

src/components/TopBar/CallButton.vue
jancborchardt: > @skjnldsv can we make var(--color-sucecss) a bit more vibrant? I think it's too dull, at least for this use case...
> Screenshot from 2020-04-27 10-14-22

I was about to write that we need to make the green darker for better contrast with the text, and the --color-success does the job quite well. :) We can always tweak, but agree with keeping it consistent.

src/components/TopBar/CallButton.vue
```diff
@@ -191,6 +192,16 @@ export default {
 	}
 }
 
+.success {
+	color: white;
+	background-color: #35CC00;
+	border: 1px solid #35CC00;
+	&:hover {
```
jancborchardt, line 199: Don’t forget the `&:focus`! ;)

src/components/TopBar/CallButton.vue
```diff
@@ -29,7 +29,8 @@
 			html: true
 		}"
 		:disabled="startCallButtonDisabled || loading || blockCalls"
-		class="top-bar__button primary"
+		class="top-bar__button"
+		:class="{'primary': !(conversation.hasCall && !isBlockedByLobby),'success': conversation.hasCall && !isBlockedByLobby}"
```
nickvergessen, line 33: Computed please?

src/components/TopBar/CallButton.vue
```diff
@@ -191,6 +192,16 @@ export default {
 	}
 }
 
+.success {
+	color: white;
+	background-color: #35CC00;
```
nickvergessen, line 197: Can we use the variables.scss for this? So in case they are reused its the same, i'd like to not use any non variable colors

src/components/TopBar/CallButton.vue
nickvergessen: Should this be kept with something like `darken($color-success, 7%) !default` ?
Or drop it, when we move to `var(--color-success)` @jancborchardt

nickvergessen:
Now also works in darkmode.

- [ ] Should there be a `darken($color-success, 7%)` on the success color? Marco had it added, but neither the error nor the primary one have this, so either it is added similarly to all, or to none I think.

ma12-co:
Let's brighten it please, since we already used the dark one :pray:

PR #3426, opened 2020-04-29
Executor /help bugfixe

GlobeNico:
There is a bug when showing /help with alias command, since I presume an alias is just another way to write the command. I decide to skip the alias: when showing help
I also added the command name in the output to show the user what comand to input.

### DIFF
--- a/lib/Chat/Command/Executor.php
+++ b/lib/Chat/Command/Executor.php
@@ -132,9 +132,9 @@ class Executor {
 						$tempHelp = substr($tempHelp, 0, strpos($tempHelp, "\n"));
 					}
 				}
-				$helps[] = $command->getCommand() .' -> ' . $tempHelp;
+				$helps[] = $tempHelp;
 			} else {
-				$helps[] = $command->getCommand() .' -> ' . $response;
+				$helps[] = $response;
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


lib/Chat/Command/Executor.php
```diff
@@ -132,9 +132,9 @@ protected function execHelp(Room $room, IComment $message, string $arguments): s
 						$tempHelp = substr($tempHelp, 0, strpos($tempHelp, "\n"));
 					}
 				}
-				$helps[] = $tempHelp;
+				$helps[] = $command->getCommand() .' -> ' . $tempHelp;
```
nickvergessen, line 135: Not sure about this. Currently all the samples print `/<sample>` already:
https://github.com/nextcloud/spreed/blob/master/sample-commands/hackernews.php#L36
```
/wiki - A simple command to find wikipedia articles for a term
/hackernews - A simple command to list the Top 5 top, new or best stories
An error occurred while running the command. Please ask an administrator to check the logs.
```

So with your change it is duplicated:
```
wiki -> /wiki - A simple command to find wikipedia articles for a term
hackernews -> /hackernews - A simple command to list the Top 5 top, new or best stories
```

nickvergessen:
Thanks for patch, I fixed the coding style and reverted the prefixing.

jancborchardt:
Hi @GlobeNico! We have a Talk development chat for all the contributors where we can invite you with a guest account. Let us know if you would like to be added. :)

PR #3433, opened 2020-04-30
[stable18] Improve call view

ma12-co:


nickvergessen:
Feedback: Scrollbar is rather small and therefor it is not very easy for inexperienced users to find it and scroll along. So buttons ala "Next" and "Previous" at the end and beginning of the section would be prefered. Not sure how doable that it?

PR #3447, opened 2020-05-04
Fix ambiguous id in query of getRoomForParticipant

nickvergessen:
cc @apircalabu want to review this?

### DIFF
--- a/lib/Manager.php
+++ b/lib/Manager.php
@@ -251,7 +251,7 @@ class Manager {
 		$query = $this->db->getQueryBuilder();
 		$query->select('*')
 			->from('talk_rooms', 'r')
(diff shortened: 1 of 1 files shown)


lib/Manager.php
```diff
@@ -251,7 +251,7 @@ public function getRoomForParticipant(int $roomId, ?string $participant): Room {
 		$query = $this->db->getQueryBuilder();
 		$query->select('*')
 			->from('talk_rooms', 'r')
-			->where($query->expr()->eq('id', $query->createNamedParameter($roomId, IQueryBuilder::PARAM_INT)));
+			->where($query->expr()->eq('r.id', $query->createNamedParameter($roomId, IQueryBuilder::PARAM_INT)));
```
apircalabu, line 254: Fixed, the SQL error is now gone.

PR #3456, opened 2020-05-05
Correctly catch errors that happen when fetching conversations

nickvergessen:
Currently spaming on c.nc.c


nickvergessen:
ah, needs #3373 before

PR #3457, opened 2020-05-05
Chat not loading when network was interrupted once

nickvergessen:


kesselb:
Screenshot from 2020-05-05 17-49-33

Looks good to me. Probably the more robust exception handling is also required for other places.

nickvergessen:
Yeah that is in #3457

PR #3465, opened 2020-05-05
Add talk:room:* CLI commands

PhrozenByte:
~~This is **WORK IN PROGRESS**.~~ It's neither fully tested nor do unit tests exist yet (will follow after some feedback).

Feedback is highly appreciated :heart: 

* Create new rooms with `talk:room:create`
* Delete rooms with `talk:room:delete`
* Add users to rooms with `talk:room:add`
* Remove users from rooms with `talk:room:remove`
* Promote users to moderators with `talk:room:promote`
* Demote users with `talk:room:demote`
* Update room settings with `talk:room:update`
* List rooms of a user with `talk:room:list`

The `talk:room:*` CLI commands don't support one-to-one rooms on purpose - couldn't think of a use case why one would need to create those using CLI commands.

---

* [ ] Write PHP unit tests; *Postponed*
* [x] Add `--circle` and `--group` options to `talk:room:add` to add all users of a circle resp. group to a room
* [ ] Add command completion, see https://github.com/nextcloud/server/blob/cb057829f72c70e819f456edfadbb29d72dba832/core/Command/Base.php#L172-L191; *Postponed*
* [x] Open question: Decide what to do with the `TRoomCommand` trait; *Decision:* Keep the trait for now
* [x] Open question: Keep `talk:room:list` command or remove it due to privacy concerns? *Decision:* Remove the command
* [x] Open question: Rename commands to `talk:conversation:*`? *Decision:* Keep the commands as `talk:room:*`


nickvergessen:
So while on a technical level the name room is used, we try to refer to them as conversations in the UI. Should the commands be called talk:conversation:* ? (just thinking loud here)

lib/Command/Room/Update.php
```diff
@@ -0,0 +1,137 @@
...
+			)->addOption(
+				'readonly',
+				null,
+				InputOption::VALUE_REQUIRED,
+				'Modifies the room to be a room with read-only (value 1) or read-write (value 0) access'
+			)->addOption(
+				'password',
+				null,
+				InputOption::VALUE_REQUIRED,
+				'Sets a new password for the room; pass a empty value to remove password protection'
```
nickvergessen, line 85:
```suggestion
				'Sets a new password for the room; pass an empty value to remove password protection'
```

lib/Command/Room/Update.php
```diff
@@ -0,0 +1,137 @@
...
+			)->addOption(
+				'public',
+				null,
+				InputOption::VALUE_REQUIRED,
+				'Modifies the room to be a public room (value 1) or private room (value 0)'
+			)->addOption(
+				'readonly',
+				null,
+				InputOption::VALUE_REQUIRED,
+				'Modifies the room to be a room with read-only (value 1) or read-write (value 0) access'
```
nickvergessen, line 80:
```suggestion
				'Modifies the room to be read-only (value 1) or read-write (value 0) access'
```

lib/Command/Room/Update.php
```diff
@@ -0,0 +1,137 @@
...
+		try {
+			$room = $this->manager->getRoomByToken($token);
+		} catch (RoomNotFoundException $e) {
+			$output->writeln('<error>Room not found.</error>');
+			return 1;
+		}
+
+		try {
+			if ($name !== null) {
+				$this->setRoomName($room, $name);
```
nickvergessen, line 115: This works by accident for one-to-one conversations at the moment.

I would say that we should check the type after each `getRoomByToken` explicitly

lib/Command/Room/Remove.php
```diff
@@ -0,0 +1,94 @@
...
+use OCA\Talk\Manager;
+use OCP\IConfig;
+use OCP\IUserManager;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class Remove extends Base {
+	use TRoomCommand;
```
nickvergessen, line 39: Instead of a trait adding all the functions in all the commands, just add them to the commands

lib/Command/Room/Remove.php
```diff
@@ -0,0 +1,94 @@
...
+	/** @var IUserManager */
+	public $userManager;
+
+	/** @var Manager */
+	public $manager;
+
+	public function __construct(IConfig $config, IUserManager $userManager, Manager $manager) {
+		parent::__construct();
+
+		$this->config = $config;
```
nickvergessen, line 53: this is only used in the constructor

lib/Command/Room/TRoomCommand.php
```diff
@@ -0,0 +1,222 @@
...
+		foreach ($userIds as $userId) {
+			try {
+				$participants[] = $room->getParticipant($userId);
+			} catch (ParticipantNotFoundException $e) {
+				throw new InvalidArgumentException(sprintf("User '%s' is no participant.", $userId));
+			}
+		}
+
+		foreach ($participants as $participant) {
+			$room->setParticipantType($participant, Participant::MODERATOR);
```
nickvergessen, line 194: You need to check that the user was a normal user before. It should not allow "demoting" owners to moderators

lib/Command/Room/Promote.php
```diff
@@ -0,0 +1,94 @@
...
+	/** @var IUserManager */
+	public $userManager;
+
+	/** @var Manager */
+	public $manager;
+
+	public function __construct(IConfig $config, IUserManager $userManager, Manager $manager) {
+		parent::__construct();
+
+		$this->config = $config;
```
nickvergessen, line 53: not used

lib/Command/Room/ListCommand.php
```diff
@@ -0,0 +1,111 @@
...
+use OC\Core\Command\Base;
+use OCA\Talk\Manager;
+use OCA\Talk\Room;
+use OCP\IConfig;
+use Symfony\Component\Console\Helper\Table;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class ListCommand extends Base {
```
nickvergessen, line 36: I'm unsure about this command. I feel it exposes the user's privacy too much/easily. I know that the admin could also check this in the database, but it feels wrong to have it available like this.

lib/Command/Room/Demote.php
```diff
@@ -0,0 +1,94 @@
...
+	/** @var IUserManager */
+	public $userManager;
+
+	/** @var Manager */
+	public $manager;
+
+	public function __construct(IConfig $config, IUserManager $userManager, Manager $manager) {
+		parent::__construct();
+
+		$this->config = $config;
```
nickvergessen, line 53: not used

lib/Command/Room/Delete.php
```diff
@@ -0,0 +1,75 @@
...
+	/** @var IConfig */
+	private $config;
+
+	/** @var Manager */
+	public $manager;
+
+	public function __construct(IConfig $config, Manager $manager) {
+		parent::__construct();
+
+		$this->config = $config;
```
nickvergessen, line 45: not used

lib/Command/Room/Create.php
```diff
@@ -0,0 +1,124 @@
...
+	/** @var IUserManager */
+	public $userManager;
+
+	/** @var Manager */
+	public $manager;
+
+	public function __construct(IConfig $config, IUserManager $userManager, Manager $manager) {
+		parent::__construct();
+
+		$this->config = $config;
```
nickvergessen, line 52: not used

lib/Command/Room/Add.php
```diff
@@ -0,0 +1,94 @@
...
+	/** @var IUserManager */
+	public $userManager;
+
+	/** @var Manager */
+	public $manager;
+
+	public function __construct(IConfig $config, IUserManager $userManager, Manager $manager) {
+		parent::__construct();
+
+		$this->config = $config;
```
nickvergessen, line 53: not used

nickvergessen:
Another thought. On the API you can add users, groups and circles to a conversation. Maybe we should allow that too here, check https://github.com/nextcloud/spreed/blob/445756dea3356a3f2d39ad00ad581ebb4c86bda2/lib/Controller/RoomController.php#L881 for comparison.

lib/Command/Room/Remove.php
PhrozenByte: I was feeling like having it all in one place is a better solution - some of the methods are used multiple times (leading to duplicate code), some are not. Usually I'd create a service, but I don't think it's sensible to create a separate service for a command.

Most of the code in the trait shouldn't exist at all. I don't know the reason why the manager just returns `false` on errors instead of throwing exceptions with proper messages, or why it is an error to change something to its current value. So the best solution IMO would be to refactor the manager accordingly. However, I'm definitely not enough into the code base yet to estimate the work this requires - and it's kinda out of the scope of this PR.

So there's IMO no "perfect" solution right now. I'll choose whatever solution you prefer (trait, duplicate functions, separate service, refactor manager, …) - just let me know :+1:

lib/Command/Room/ListCommand.php
PhrozenByte: Since all commands require a room token, I was feeling like there must be a way to retrieve a list of the available room tokens via CLI. However, to be honest, without a list of all participants it can get quite hard to find the right room anyway (since room names don't have to be unique)... On the other hand, it just lists the bare minimum - and no participants. So one could argue that it isn't showing sensitive data anyway. So, as you can see, I'm not sure either :laughing:

nickvergessen: Well showing the name can be very intrusive. E.g. if management has a conversation: "Can we fire sysadmin1". But then again they could also check it in the database. but it's not so easy, so :/

lib/Command/Room/Remove.php
nickvergessen: Ah, the Create conversation uses them too.

And yeah the manager just grew and was never organized.

PhrozenByte:
> So while on a technical level the name room is used, we try to refer to them as conversations in the UI. Should the commands be called talk:conversation:* ? (just thinking loud here)

Yeah, I was thinking about that, too. Depends on whether we want CLI commands to be closer to the UI or closer to the code. Personally I still call group conversations "rooms" in everyday use, but that's just personal preference. Maybe calling it `talk:room:*` outlines that it works with group calls (and public calls as sub-type of group calls) only. Anyway, I'll leave this up to you guys, just let me know if you want me to rename them to `talk:conversation:*` :+1: 

> Another thought. On the API you can add users, groups and circles to a conversation.

Good idea :+1: Will implement this later (added a ToDo list to the top comment).

> Bonus points to get autocomplete for the options and arguments

Good idea :+1: Will look into this :smiley:

nickvergessen:
We are actually coming closer to the feature freeze (it was supposed to be last week), but I'd like to get this in. When do you think you could work on this?

PhrozenByte:
> We are actually coming closer to the feature freeze (it was supposed to be last week), but I'd like to get this in. When do you think you could work on this?

Depends. I've just added the `--group` and `--circle` options as well as the `--owner` option to set a room owner. We might postpone the command completion feature.

As discussed above, there are some open questions. I'd leave everything as-is, but as I said, I'll leave the final decision up to you. So just tell me which solution you prefer:

* Open question: Decide what to do with the `TRoomCommand` trait (keep trait, create a service, move methods to command classes and use duplicated code)
* Open question: Keep `talk:room:list` command or remove it due to privacy concerns?
* Open question: Rename commands to `talk:conversation:*`?

So the only missing thing are unit tests. When is the feature freeze going to happen? This weekend? Sunday? Saturday? Today? I might find some time on Sunday.

nickvergessen:
I would say:

> Open question: Decide what to do with the TRoomCommand trait (keep trait, create a service, move methods to command classes and use duplicated code)

Keep it for now and we check back once a service is there.

> Open question: Keep talk:room:list command or remove it due to privacy concerns?

I would remove it

> Open question: Rename commands to talk:conversation:*?

Flip a coin :P

We will most likely do the freeze today, but we can also merge this without tests and add them afterwards that is fine by me.

PhrozenByte:
Ok, I've just removed the `talk:room:list` command. So I'd say it's ready for "intermediate" merging now. I'll open another pull request for unit tests and command completion in the next few days.

lib/Command/Room/ListCommand.php
PhrozenByte: Removed, see 0c57a68

lib/Command/Room/Add.php
```diff
@@ -0,0 +1,110 @@
...
+
+	protected function configure(): void {
+		$this
+			->setName('talk:room:add')
+			->setDescription('Adds users to a room')
+			->addArgument(
+				'token',
+				InputArgument::REQUIRED,
+				'Token of the room to add users to'
+			)->addArgument(
```
nickvergessen, line 63: lets make user also an option

lib/Command/Room/Create.php
```diff
@@ -0,0 +1,142 @@
...
+
+	protected function configure(): void {
+		$this
+			->setName('talk:room:create')
+			->setDescription('Create a new room')
+			->addArgument(
+				'name',
+				InputArgument::REQUIRED,
+				'The name of the room to create'
+			)->addArgument(
```
nickvergessen, line 61: lets make user also an option

lib/Command/Room/Create.php
```diff
@@ -0,0 +1,142 @@
...
+		try {
+			$this->setRoomReadOnly($room, $readonly);
+			$this->setRoomPassword($room, $password);
+
+			$this->addRoomParticipants($room, $users);
+			$this->addRoomParticipantsByGroup($room, $groups);
+			$this->addRoomParticipantsByCircle($room, $circles);
+			$this->addRoomModerators($room, $moderators);
+
+			$this->setRoomOwner($room, $owner);
```
nickvergessen, line 131:
```suggestion
			if ($owner !== null) {
				$this->setRoomOwner($room, $owner);
			}
```

lib/Command/Room/Create.php
```diff
@@ -0,0 +1,142 @@
...
+		if (!$this->validateRoomName($name)) {
+			$output->writeln("<error>Invalid room name.</error>");
+			return 1;
+		}
+
+		$room = $public ? $this->manager->createPublicRoom($name): $this->manager->createGroupRoom($name);
+
+		try {
+			$this->setRoomReadOnly($room, $readonly);
+			$this->setRoomPassword($room, $password);
```
nickvergessen, line 124:
```suggestion
			if ($password !== null) {
				$this->setRoomPassword($room, $password);
			}
```

PhrozenByte: Yeah, missing unit tests :man_facepalming:

PhrozenByte:
Done :rocket: 

And https://github.com/nextcloud/spreed/pull/3465#discussion_r422168314 was an great example why one needs unit tests :laughing:

nickvergessen:
Thanks a lot for the PR, looking forward to the follow up :)

mentalinc:
Would be great to add a command to "end" calls (en mass or per user)...
Seems to be a bug in the android app when the back button is pressed instead of the red hang up the call doesn't end. 
Which causes issues trying to have a call with them in the future as the server thinks they are on the call (maybe for hundreds of hours)

PR #3484, opened 2020-05-08
Fix fullscreen having a black bar on top

jancborchardt:
The room freed by the header vanishing was not filled up:
fullscreen black bar

Fixed:
Fullscreen fixed

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -434,6 +434,15 @@ export default {
 			color: #FFFFFF;
 		}
 	}
+
+	// Fix fullscreen black bar on top
+	&:fullscreen {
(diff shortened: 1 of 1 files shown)


src/App.vue
```diff
@@ -434,6 +434,15 @@ export default {
 			color: #FFFFFF;
 		}
 	}
+
+	// Fix fullscreen black bar on top
+	&:fullscreen {
+		padding-top: 0;
+
+		::v-deep .app-sidebar {
+			height: 100vh;
```
nickvergessen, line 443: Missing `!important`

PR #3500, opened 2020-05-11
Bugfix/noid/grid recalculation

nickvergessen:


nickvergessen:
Could add this additionally:
```
diff --git a/src/components/CallView/GridView/GridView.vue b/src/components/CallView/GridView/GridView.vue
index 5468a89a..bcc3db19 100644
--- a/src/components/CallView/GridView/GridView.vue
+++ b/src/components/CallView/GridView/GridView.vue
@@ -400,7 +400,7 @@ export default {
                        // currently if the user is not on the 'first page', upon resize the
                        // current position in the videos array is lost (first element
                        // in the grid goes back to be first video)
-                       debounce(this.makeGrid(), 200)
+                       debounce(this.makeGrid(true), 200)
                        if (this.hasPagination) {
                                this.setNumberOfPages()
                                // Set the current page to 0
@@ -411,7 +411,7 @@ export default {
 
                // Find the right size if the grid in rows and columns (we already know
                // the size in px).
-               makeGrid() {
+               makeGrid(isResizing) {
                        this.gridWidth = this.$refs.grid.clientWidth
                        this.gridHeight = this.$refs.grid.clientHeight
                        // prevent making grid if no videos
@@ -419,6 +419,12 @@ export default {
                        if (this.videos.length === 0) {
                                return
                        }
+
+                       if (!isResizing && this.videos.length < this.slots) {
+                               // No need to recalculate as the number of slots is still enough
+                               return
+                       }
+
                        console.debug('makeGrid')
                        // We start by assigning the max possible value to our rows and columns
                        // variables. These variables are kept in the data and represent how the

```

so e.g. when you are on a 3*4 grid, with 9 others, and number 10+11 join we don't recalculate, as it would still fit (would need to add one more check so it only "return"s when the number of videos is increasing

PR #3505, opened 2020-05-12
Improve talk:room:* CLI commands

PhrozenByte:
This is a follow-up to #3465 and **WORK IN PROGRESS**.

**ToDo:**

* [x] Write PHP unit tests
* [x] Add command completion; see https://github.com/nextcloud/server/blob/cb057829f72c70e819f456edfadbb29d72dba832/core/Command/Base.php#L172-L191
* [ ] Open question: Decide what to do with the `TRoomCommand` trait; see https://github.com/nextcloud/spreed/pull/3465#pullrequestreview-406344263


PhrozenByte:
Not really much to see here yet :smile: Currently working on the unit tests, a task that is way more complex than I expected at first...

nickvergessen:
Just for the record. I just merged code-style checking which is part of Nextcloud 19 with #3329 
So maybe make sure you rebase on master early to avoid conflicts.

PhrozenByte:
@nickvergessen That's not really PSR-2, just another custom standard slightly closer to PSR-2, right? For example PSR-2 requires to have the opening brackets of class and method definitions to be on the next line :confused: 

Anyway, PHPUnit tests are ready to rumble! :rocket:

PHPUnit tests are lacking tests for the `--circle` options. I'm not sure how to test them... The API isn't very friendly to unit testing (like using static methods). Any ideas?

Bash autocompletion is still work in progress. I need some advice resp. help here... [`TRoomCommand::completeCircleValues()`](https://github.com/nextcloud/spreed/pull/3505/files#diff-1937e6690fc3cf818ba427511fd182deR366-R369) (for completing `--circle` options) and [`TRoomCommand::completeParticipantValues()`](https://github.com/nextcloud/spreed/pull/3505/files#diff-1937e6690fc3cf818ba427511fd182deR371-R374) (for completing participant options and arguments, like `--owner`) are missing.

1. `TRoomCommand::completeCircleValues()`: I'm not very familiar with circles (I don't use them at all). Is there an API like for users/groups to perform autocompletion for circles? I couln't find one.

2. `TRoomCommand::completeParticipantValues()`: What's the recommended way to deal with this? To create a list of available participants we have to read the room token from the command line first to create a `Room` object. However, since the command line options weren't interpreted yet, we can't rely on argument/option order (thus we can't just access word at index 2 to get the room token). What's the recommended way for situations like that?

Besides the above, we still have to decide what to do with the `TRoomCommand` trait, see https://github.com/nextcloud/spreed/pull/3465#pullrequestreview-406344263.

PhrozenByte:
@nickvergessen I'm currently thinking about removing the `--circle` options altogether. The options don't work right now anyway (the API interprets any circle ID in the context of the current user, and since there's no current user in the console it won't ever return a circle). Furthermore there's no way to get a circle's ID - you have to check the database. The only way to make this work would be to add two options, one for the circle owner and one for the circle name. Please correct me, but weren't circles designed to allow *users* to organize their contacts, not *admins*? Circles are like the contacts app, just with Nextcloud users. A console command rather works on a installation-wide basis, thus I feel like it won't be used much anyway, meaning the work isn't worth it in the end. What do you think?

nickvergessen:
Yeah, sounds right. Let's remove it

PhrozenByte:
Correction: I totally forgot about the `circles:manage:list` command. So I was wrong about the "you have to check the database" part to get a circle ID. Anyway, this still means that one would have to list all existing circles with `circles:manage:list`, then check its members using `circles:members:list` to be sure to choose the right circle and finally feed its ID into `talk:room:*`. If someone chooses to do that one could also feed the output of `circles:members:list` to `talk:room:*`... So I still believe the effort isn't worth it. I've removed the code just now; see ac41ca22 if one wants to re-add this later.

Furthermore I've just implemented bash completion for the missing participant options (like `--owner` and `--moderator`).

Thus this is ready for review and merge @nickvergessen :rocket: 

Regarding `TRoomCommand` I'd vote for leaving it as-is. Anyway, I'm open for suggestions. See https://github.com/nextcloud/spreed/pull/3465#pullrequestreview-406344263.

jancborchardt:
Hi @PhrozenByte! We have a Talk development chat for all contributors where we can invite you with a guest account – is it ok to add you there? Would be cool to have you :)

nickvergessen:
Hmm seems like the PR slipped, will look later today, sorry about that!

lib/Manager.php
```diff
@@ -224,6 +224,43 @@ public function resetAssignedSignalingServers(ICache $cache): void {
...
+
+		$query->setMaxResults($limit)
+			->setFirstResult($offset)
+			->orderBy('token', 'ASC');
+		$result = $query->execute();
+
+		$rooms = [];
+		while ($row = $result->fetch()) {
+			if ($row['token'] === null) {
+				// FIXME Temporary solution for the Talk6 release
```
nickvergessen, line 253: 😨

PhrozenByte: Nothing is more definitive than the temporary :laughing: No idea whether this is needed... I just copied it from `Manager::getRoomForParticipant()`. Is the check still necessary?

lib/Command/Room/TRoomCommand.php
```diff
@@ -335,4 +304,60 @@ protected function removeRoomModerators(Room $room, array $userIds): void {
 			$room->setParticipantType($participant, Participant::USER);
 		}
 	}
+
+	protected function completeTokenValues(CompletionContext $context): array {
+		/** @var Manager $roomManager */
+		$roomManager = \OC::$server->query(Manager::class);
+		return array_map(function (Room $room) {
+			return $room->getToken();
+		}, $roomManager->searchRoomsByToken($context->getCurrentWord()));
```
nickvergessen, line 313:
```suggestion
		}, $this->manager->searchRoomsByToken($context->getCurrentWord()));
```
And just move
```php
	/** @var Manager */
	public $manager;
```

from all classes to the trait.

lib/Command/Room/TRoomCommand.php
```diff
@@ -335,4 +304,60 @@ protected function removeRoomModerators(Room $room, array $userIds): void {
...
+	}
+
+	protected function completeUserValues(CompletionContext $context): array {
+		$userManager = \OC::$server->getUserManager();
```
nickvergessen, line 317: Use DI in the command and inject it into this method?

Same for the other managers?

PhrozenByte: Done :+1: :smiley:

PhrozenByte:
> Hi @PhrozenByte! We have a Talk development chat for all contributors where we can invite you with a guest account – is it ok to add you there? Would be cool to have you :)

Sure, go ahead @jancborchardt :+1: :smiley:

nickvergessen:
Emaile at `<my github username>@nextcloud.com` with the email address we should use

nickvergessen:
Let's backport this because of the fixes to the injection and the removal of `--circle`

PR #3555, opened 2020-05-14
Only add chat reference id capability when it exists

### LINKED ISSUE #3401: Only add "chat-reference-id" when the DB scheme has been updated.
The server should expose the "chat-reference-id" capability only when the DB has updated its scheme and has the referece-id column.

nickvergessen:
Fix #3401

### DIFF
--- a/lib/Capabilities.php
+++ b/lib/Capabilities.php
@@ -54,59 +54,57 @@ class Capabilities implements IPublicCapability {
 			return [];
 		}
 
-		$maxChatLength = 1000;
-		if (version_compare($this->serverConfig->getSystemValueString('version', '0.0.0'), '16.0.2', '>=')) {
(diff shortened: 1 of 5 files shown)


lib/Capabilities.php
```diff
@@ -54,59 +54,58 @@ public function getCapabilities(): array {
...
+				'invite-groups-and-mails',
+				'locked-one-to-one-rooms',
+				'read-only-rooms',
+				'chat-read-marker',
+				'webinary-lobby',
+				'start-call-flag',
+				'chat-replies',
+				'circles-support',
+				'force-mute',
+				'chat-reference-id',
```
Ivansss, line 83: chat-reference-id should be removed from the default capability list

nickvergessen: 🤦

PR #3559, opened 2020-05-14
Add drop shadow again on icons in the topbar too

nickvergessen:
### Before
Bildschirmfoto von 2020-05-14 13-34-54

### After
Bildschirmfoto von 2020-05-14 13-34-12

It's at least a bit better, but the `…` is very small so the shadow is also small

### DIFF
--- a/css/icons.scss
+++ b/css/icons.scss
@@ -46,11 +46,14 @@
 	&.in-call {
 		.app-navigation-toggle,
 		.action-item__menutoggle {
-			color: #fff;
+			color: #ffffff;
+			filter: drop-shadow(1px 1px 4px var(--color-box-shadow));
 		}
 	}
 
 	.forced-white {
+		filter: drop-shadow(1px 1px 4px var(--color-box-shadow));
+
 		&.icon-menu-people {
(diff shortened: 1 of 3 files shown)


css/icons.scss
```diff
@@ -46,11 +46,14 @@
...
 		.app-navigation-toggle,
 		.action-item__menutoggle {
-			color: #fff;
+			color: #ffffff;
+			filter: drop-shadow(1px 1px 4px var(--color-box-shadow));
 		}
 	}
 
 	.forced-white {
+		filter: drop-shadow(1px 1px 4px var(--color-box-shadow));
```
jancborchardt, line 55: We don’t need both the circle _and_ the drop-shadow, right? We can remove the drop-shadows I’d say

nickvergessen: This also affects the other icons which don't have a circle around them.
But we can have a look at it again when this is live

jancborchardt: Yeah, no big deal :) Looks nicer if the icons with the circle don’t have the drop-shadow, just something for the long list cc @ma12-co

Issue #3483, opened 2020-05-08
Call view in sidebars is broken


nickvergessen:
## Steps to reproduce
1. Go to a file
2. Share it
3. Go to the Talk tab and start the call on the file

### Expected behaviour
You have a call and see mostly the other user, or a grid, or ...

### Actual behaviour
You overlap almost everything.

nickvergessen:
cc @jancborchardt what was the expected result again?

jancborchardt:
- Only show the promoted speaker in large
- Your own video in quite small overlaid on the bottom right
- Only audio and video controls for you (fade out after some seconds of not moving mouse / interacting keyboard over it, just like in Talk)
- Names won’t fit so no need to display them.
- The screensharing button doesn’t need to be there either as no one will screenshare in there. ;)

PR #3563, opened 2020-05-14
Fix the callview in sidebars

ma12-co:
Screenshot from 2020-05-14 15-12-47

fix: #3483

### DIFF
--- a/src/FilesSidebarCallViewApp.vue
+++ b/src/FilesSidebarCallViewApp.vue
@@ -25,6 +25,7 @@
 		<CallView
 			v-show="isInCall"
 			:token="token"
+			:is-sidebar="true"
 			:use-constrained-layout="true" />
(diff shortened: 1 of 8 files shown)


src/FilesSidebarCallViewApp.vue
```diff
@@ -25,6 +25,7 @@
 		<CallView
 			v-show="isInCall"
 			:token="token"
+			:is-sidebar="true"
 			:use-constrained-layout="true" />
```
nickvergessen, line 29: or those two the same?

nickvergessen:
> Bildschirmfoto von 2020-05-14 15-28-28

So In the files sidebar the black borders are left and right and in the public sharing sidebar top and bottom.
Would it make sense to change the height on both so with 3:4 video it has no black border?

PR #3583, opened 2020-05-15
added warning about mandatory distributed caching with HPB

wiswedel:
Signed-off-by: Sascha Wiswedel <sascha.wiswedel@nextcloud.com>


nickvergessen:
I think the readme is not the right place for this information.
I added it to the admin interface with: #3565

wiswedel:
> I added it to the admin interface with: #3565

Ah, great. So this resolves [our discussion](https://github.com/nextcloud-gmbh/customers/issues/466#issuecomment-628001945) in the end. Even better.

> I think the readme is not the right place for this information.

We don't have a real admin- and user documentation for Talk, so this was the only suitable place I could find.

Issue #3415, opened 2020-04-28
Inconsistent state when accidentally joining with 2 devices


jospoortvliet:
## Steps to reproduce
1. somebody calls you, you join in the browser
2. instead of turning off your ringing tablet, you accidentaly slide-to-join, then cancel that again

### Expected behaviour
Nothing special, you just have your call

### Actual behaviour
1. a spinning second avatar of you joins the call
2. after a while you suddenly get dropped back in the normal chat view, out of the call, join call button is there - but you can still hear the other participants and they can hear you
3. click join call and you're back, with a warning about leaving the page (?) that you can dismiss.

## Browser
Chromium, latest - 81.0.4.something

### Browser log
Sorry, totally failed to capture that

## Talk app
cloud.nextcloud.com setup

nickvergessen:
Already being dealt with in #3338

nickvergessen:
This is solved.

The rest will be dealt in #1587

PR #3591, opened 2020-05-15
"409 Conflict" on joining a second time in the same conversation

nickvergessen:
Ref #3415

### DIFF
--- a/lib/Controller/RoomController.php
+++ b/lib/Controller/RoomController.php
@@ -1193,6 +1193,10 @@ class RoomController extends AEnvironmentAwareController {
 				], Http::STATUS_CONFLICT);
 			}
 
+			if ($previousSession->getInCallFlags() !== Participant::FLAG_DISCONNECTED) {
+				$room->changeInCall($previousSession, Participant::FLAG_DISCONNECTED);
+			}
+
 			if ($this->userId === null) {
 				$room->removeParticipantBySession($previousSession, Room::PARTICIPANT_LEFT);
 			} else {
--- a/lib/Controller/RoomController.php
+++ b/lib/Controller/RoomController.php
@@ -1169,14 +1169,26 @@ class RoomController extends AEnvironmentAwareController {
 			return new DataResponse([], Http::STATUS_NOT_FOUND);
 		}
 
-		if ($force === false && $this->userId !== null) {
-			try {
-				$participant = $room->getParticipant($this->userId);
-				if ($participant->getSessionId() !== '0') {
-					return new DataResponse([], Http::STATUS_CONFLICT);
+		if ($force === false) {
+			if ($this->userId !== null) {
+				try {
+					$participant = $room->getParticipant($this->userId);
+					if ($participant->getSessionId() !== '0') {
+						return new DataResponse([], Http::STATUS_CONFLICT);
+					}
+				} catch (ParticipantNotFoundException $e) {
+					// All fine, carry on
 				}
-			} catch (ParticipantNotFoundException $e) {
-				// All fine, carry on
+			} else {
+				$session = $this->session->getSessionForRoom($token);
+				try {
+					$participant = $room->getParticipantBySession($session);
+					if ($participant->getSessionId() !== '0') {
+						return new DataResponse([], Http::STATUS_CONFLICT);
+					}
+				} catch (ParticipantNotFoundException $e) {
+					// All fine, carry on
+				}
 			}
 		}
 
--- a/lib/Controller/RoomController.php
+++ b/lib/Controller/RoomController.php
@@ -1174,7 +1174,11 @@ class RoomController extends AEnvironmentAwareController {
 				try {
 					$participant = $room->getParticipant($this->userId);
 					if ($participant->getSessionId() !== '0') {
-						return new DataResponse([], Http::STATUS_CONFLICT);
+						return new DataResponse([
+							'sessionId' => $participant->getSessionId(),
+							'inCall' => $participant->getInCallFlags(),
+							'lastPing' => $participant->getLastPing(),
+						], Http::STATUS_CONFLICT);
 					}
 				} catch (ParticipantNotFoundException $e) {
 					// All fine, carry on
@@ -1184,7 +1188,11 @@ class RoomController extends AEnvironmentAwareController {
 				try {
 					$participant = $room->getParticipantBySession($session);
 					if ($participant->getSessionId() !== '0') {
-						return new DataResponse([], Http::STATUS_CONFLICT);
+						return new DataResponse([
+							'sessionId' => $participant->getSessionId(),
+							'inCall' => $participant->getInCallFlags(),
+							'lastPing' => $participant->getLastPing(),
+						], Http::STATUS_CONFLICT);
 					}
 				} catch (ParticipantNotFoundException $e) {
 					// All fine, carry on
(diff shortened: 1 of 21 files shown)


nickvergessen:
This is part 1 of solving #3415 

It asks the user whether they want to join a conversation when the previous session pinged in the last 40 seconds (or 60 when it was/is in a call). Reloading within the same tab will always force join automatically.

Part 2 is handling the previous session, in case the user picked the "force join" option. That will be handled in #3338

nickvergessen:
Oh, missed to leave the old session on the signaling server in case of a force join

lib/Controller/RoomController.php
```diff
@@ -1144,15 +1144,47 @@ public function setPassword(string $password): DataResponse {
...
+					'sessionId' => $previousSession->getSessionId(),
+					'inCall' => $previousSession->getInCallFlags(),
+					'lastPing' => $previousSession->getLastPing(),
+				], Http::STATUS_CONFLICT);
+			}
+
+			if ($this->userId === null) {
+				$room->removeParticipantBySession($previousSession, Room::PARTICIPANT_LEFT);
+			} else {
+				$room->leaveRoomAsParticipant($previousSession);
```
nickvergessen, line 1184: Doesn't work with external signaling it seems

nickvergessen:
Now works also acceptable while chatting with the HPB (with https://github.com/nextcloud/server/pull/21417 applied):

> Bildschirmfoto von 2020-06-15 17-00-25


Todo:
- [ ] Check behaviour when in a call

nickvergessen:
Remaining issue:
- [ ] The "new" sessions UI shows as "isInCall" while the question is prompted and then toggles out of the call when clicking anything. I guess we should set the "isInCall" only when the current window/session joined the call.

nickvergessen:
Ready to review :tada:

nickvergessen:
> Additionally, if a conversation is created and a call is immediately started in that conversation the UI will not change to the call view.

This could be a result of my last commit.

I didn't think of guests and handle them at all, as they are not in the risk to continue in parallel in other conversations/mobile apps and then join an active conversation by mistake.

Let me check again if it works on my maschine.

nickvergessen:
> * Without HPB:
>   * With a user:
>      * Open the same conversation in another tab and choose Join here: both the previous tab and the new tab are redirected to Duplicate session
>      * Open the same conversation in another tab and choose Leave this page: the previous tab is redirected to Duplicate session (but only sometimes, other times it works)

Those both work pretty fine on my server. Maybe you can check the `oc_talk_participants` between the individual steps and check which session is stored?

> Additionally, if a conversation is created and a call is immediately started in that conversation the UI will not change to the call view.

Awkward: So I added my debug console logs again:
```js
			console.log('CallButton::isInCall')
			console.log(SessionStorage.getItem('joined_conversation'))
			console.log(this.token)
			console.log(this.participant.inCall)
			console.log(SessionStorage.getItem('joined_conversation') === this.token
				&& this.participant.inCall !== PARTICIPANT.CALL_FLAG.DISCONNECTED)
```

and it works again. Once this is removed, it seems that the computeds are not "refreshed".

nickvergessen:
Should be working better now :-X

nickvergessen:
The only thing that still needs work now is joining as guest and pressing "continue here" as it doesn't kick the other window apparently:

> Ignore unknown event {type: "bye", bye: {…}}bye: {reason: "room_join_timeout"}type: "bye"__proto__: Object
(anonymous) @ signaling.js:767

:runner:

nickvergessen:
Also fixed now.

nickvergessen:
> When the dialog to leave or join is shown clicking on the close button in the top left closes the dialog without selecting any option, which leaves the UI in a weird state. Unfortunately the close button is shown by default in the dialogs and there is no parameter to disable it in OC.dialogs, so this needs to be fixed in the server.

Worked around it for the time-being

nickvergessen:
> User B is redirected to the Duplicate session page instead of to the The conversation does not exist page.

Only happened with public conversations. Is fixed now too without HPB, but with the HPB this is not fixable. We can not distinguish the reason why we got the `disinvite` message sent. Maybe we create an issue in the HPB and see if we can fix it in a a followup?

src/mixins/duplicateSessionHandler.js
```diff
@@ -0,0 +1,55 @@
...
+	},
+
+	methods: {
+		duplicateSessionTriggered() {
+			this.isLeavingAfterSessionConflict = true
+			SessionStorage.removeItem('joined_conversation')
+			this.$nextTick(() => {
+				// Need to delay until next tick, otherwise the PreventUnload is still being triggered
+				// Putting the window in front with the warning and irritating the user
+				window.location = generateUrl('/apps/spreed/duplicate-session')
```
danxuliu, line 49: Redirecting to `duplicate-session` should be done only in the main Talk UI. When Talk is in the sidebar there should be no redirection, the message should be shown in the sidebar itself. That could be done in a follow up pull request, though.

nickvergessen:
> The issue seems to be that the room token stored in the session storage is not cleared, so when the conversation is joined again it is forced without asking. I guess that it could be solved by clearing the room token when the room is left, although that would mean that when the page is reloaded the token is not retained, which as far as I understand was the reason to add the session storage in first place.

Exactly. Also page reload is happening much more often than repeatingly toggle in the same conversation between two tabs. (Also note when you would first join in Tab B or you join any other conversation first in either A or B, it would work as expected, so it's very specific case you need to take to intentionally break it).

> However, given the changes added to this pull request since its beginning, is the session storage still needed or could it be replaced by just a variable that is not persisted between page reloads?

Yes, the session storage is still needed IMO as the page reload is much more important than intentionally breaking it with toggling between two tabs. Otherwise when your internet drops, your internet comes back, you do a page refresh and are greeted by "Duplicate session" which is way too confusing.

danxuliu:
> Yes, the session storage is still needed IMO as the page reload is much more important than intentionally breaking it with toggling between two tabs. Otherwise when your internet drops, your internet comes back, you do a page refresh and are greeted by "Duplicate session" which is way too confusing.

Right, I had not thought about Internet drops. Then yes, I agree that this is more important than the other case.

These should be the last issues:

- Share a file by link
- Copy the link
- Open the link in a private window (either as a guest or as a user, happens in both cases)
- Join the conversation
- Start a call

The call button shows _Join the call_ instead of _Leave the call_. This is a regression introduced in this pull request.

- Share a file by link
- Copy the link
- As the owner of the file, start a call in the conversation (either in Talk o in the Files app)
- In a private window, open the public link
- Join the call
- In a different tab of the private window, open the public link and join the conversation there

The call view will be shown for a while, but it will be eventually removed. However, then it is not possible to join the conversation again in that tab (the console shows `Participant not found`).

- Share a file
- Open the conversation in the Files app
- Start a call
- Open the Files app in another tab
- Join the conversation in that other tab

The call view will be shown and it will not go away. Same would happen if the public share page is used in this case (a user instead of a guest like in the previous scenario). Leaving the call and joining again works, though.

nickvergessen:
> The call button shows Join the call instead of Leave the call. This is a regression introduced in this pull request.

Fixed

nickvergessen:
> The call view will be shown for a while, but it will be eventually removed. However, then it is not possible to join the conversation again in that tab (the console shows Participant not found).


```js
			// FIXME The participant will not be updated with the server data
			// when the conversation is got again (as "addParticipantOnce" is
			// used), although that should not be a problem given that only the
			// "inCall" flag (which is locally updated when joining and leaving
			// a call) is currently used.
```

Well joy! Will take another look tomorrow and make sure we update the call state and the session id for guests on this.

nickvergessen:
All fixed, let's get this in finally :ship:

PR #3619, opened 2020-05-19
Correctly trim user list on name generation

nickvergessen:
### Steps
1. Have a conversation with many participants (or fake it by extending the string here: https://github.com/nextcloud/spreed/blob/c9bc798a4fb53dbfa174a861f32495fa1e8efa48/lib/Manager.php#L778 )
2. In the database clear the room name from oc_talk_rooms (to simulate old server)
3. Try to load the UI and check the room list

### Before
Loading circle for ever

### After
List is loaded

### DIFF
--- a/lib/Manager.php
+++ b/lib/Manager.php
@@ -776,8 +776,8 @@ class Manager {
 		}
 
 		$roomName = implode(', ', $displayNames);
-		if (strlen($roomName) > 128) {
-			$roomName = substr($roomName, 120) . '…';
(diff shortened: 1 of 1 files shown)


nickvergessen:
Yeah sqlite has no column length.
You needed 375+ chars user list to cause the bug. The first 120 chars were cut off instead of used and so if the remaining list was longer than the 255 db space it did boom.
The second problem was when cutting was done in a utf8 char causing a try of saving invalid bytes

PR #3622, opened 2020-05-19
If available creatae a cacheentry

### LINKED ISSUE #3440: Shares should query parent and permissions
I see a lot of queries of the form:

```sql
SELECT `storage`
FROM `oc_filecache`
WHERE `fileid` = 'XYZ' 
```

As far as I can tell this is due to: https://github.com/nextcloud/server/blob/caff1023ea72bb2ea94130e18a2a6e2ccf819e5f/apps/files_sharing/lib/SharedMount.php#L249

Hence we should do something like https://github.com/nextcloud/server/blob/cde6d768de0c48d470e6585d9b688c241187e229/lib/private/Share20/DefaultShareProvider.php#L1053 as  well.

@icewind1991

rullzer:
Fixes #3440

Signed-off-by: Roeland Jago Douma <roeland@famdouma.nl>

### DIFF
--- a/lib/Share/RoomShareProvider.php
+++ b/lib/Share/RoomShareProvider.php
@@ -28,6 +28,7 @@ declare(strict_types=1);
 
 namespace OCA\Talk\Share;
 
+use OC\Files\Cache\Cache;
 use OCA\Talk\Events\ParticipantEvent;
 use OCA\Talk\Events\RemoveUserEvent;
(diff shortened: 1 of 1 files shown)


nickvergessen:
Can this be backported to 18 and 17?

rullzer:
Not sure. I mean you could check. There is and check if the data is there so if it is not it should just work. But I guess only one way to find out ;)

Issue #3270, opened 2020-04-06
Support default settings for mute/unmute and camera off/on


philipreimer:
**Is your feature request related to a problem? Please describe.**
Everytime I join a Nextcloud Talk call using the browser, microphone and camera are active by default. I always have to mute myself and disable the camera.
I have made the experience that joining a call muted is favorable, especially in group calls, and especially if not everyone uses a headset. When everyone joins unmuted there often is a lot of disturbing background noise.

**Describe the solution you'd like**
There should be a way to configure the default settings for microphone and camera, which are applied automatically when joining a call.

A new section in the Personal Settings dedicated to Talk could be created containing two checkboxes:
* join calls muted by default
* join calls with camera off by default

**Describe alternatives you've considered**

The global Talk settings, which can only be changed by Administrators, could be extended to include two checkboxes:
* call participants join muted by default
* call participanst join with camera off by default

wbob:
I have the opposite experience for one-on-one chats: when joining calls after allowing mic+cam in the browser, I have to activate the inputs by clicking the symbols each, something I need to explain to a user unfamiliar with the interface initially. In one-on-ones, it's clear we both want to talk right away.

Though I understand your argument for a different default for group calls. I somehow think browsers should have a default handling, but couldn't find if the behaviour is meant to be standardized.

nickvergessen:
@wbob just upgrade your Talk to 8.0.7, that comes with default enabled and remembers the state in your browser on a per conversation level.

wbob:
@nickvergessen - very kind to hook me up with #3205, thank you. I really tried to search for prior issues but didn't land on #2485.

On topic: one proposition to the aforementioned issue of @philipreimer could be to make use of more than one join-button ("join with audio&camera", "join with audio only", "join muted") for group calls as surfacing as "topbar" discussion also in #2969. It's an interesting UX challenge.

philipreimer:
@wbob @nickvergessen I missed #2485 when looking for existing issues. Reading the comments there it seems that people think having a setting to start calls muted by default might be confusing for the user. So the idea is now to have separate buttons when joining a call? Is there a specific issue that I could follow?

Then this could be closed.

nickvergessen:
No, I guess we will use this issue for that topic.

kzar79:
[...]

PR #3661, opened 2020-05-25
Always start with disabled media by default

nickvergessen:

### DIFF
--- a/src/utils/webrtc/models/LocalMediaModel.js
+++ b/src/utils/webrtc/models/LocalMediaModel.js
@@ -181,7 +181,7 @@ LocalMediaModel.prototype = {
 	_setInitialMediaState: function(configuration) {
 		if (configuration.audio !== false) {
 			this.set('audioAvailable', true)
-			if (!localStorage.getItem('audioDisabled_' + this.get('token')) || this.get('audioEnabled')) {
+			if (localStorage.getItem('audioEnabled_' + this.get('token')) || this.get('audioEnabled')) {
 				this.enableAudio()
 			} else {
 				this.disableAudio()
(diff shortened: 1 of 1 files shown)


nickvergessen:
@jancborchardt @ma12-co thoughts on this?
We get quite some requests from bigger installations for this, while in the "home" case the old behaviour might be the desired one. I can understand both worlds but don't want to have something like a "this is a big instance" setting, or even a setting per conversation.

jancborchardt:
Yeah, we can do it this way or the other – don't really mind either way. Starting without video always is fine as it preserves bandwidth and privacy.

The real problem is that at the moment there is no way to get a preview of your webcam in the app without also broadcasting it.

nickvergessen:
Then lets do it after https://github.com/nextcloud/nextcloud-browser-storage/pull/2 has a release which breaks the browser storage key values anyway

jancborchardt:
Ah right, just one note: For 1on1 calls, _audio_ should be on by default, right? Otherwise I foresee a lot of confusion and possibly reports about sound not working.

ma12-co:
I don't like this.. I like the current behavior (remembering of the current state) and my Idea for the future was to implement a split button showing the last used option.

59698241-45f45e00-91bd-11e9-8b17-eafd8d0eb7d8


The option in the button would be:

* Join with audio 
* Join with video 
* Join silently
... Or something like that

To satisfy the needs of bigger installation, "silent join" could be enforced by moderators on a per conversation basis in the conversations settings tab (sidebar). But for most people hitting the call button means opening your mic or video if it's a video call. We've had many reports about how counter intuitive the old behavior was.

PVince81:
relates to the "split button" bit of https://github.com/nextcloud/spreed/issues/3270

nickvergessen:
Closing as the split button is not an option.

PR #3684, opened 2020-05-26
Reload page on changing conversation and disable start/join call when Talk hash is dirty

### LINKED ISSUE #3672: Force page refresh on conversation change when talk-hash was modified
## Steps to reproduce
1. Open a conversation
2. Change a talk admin setting in another tab 
3. Wait until you see the notice that talk was updated

### Expected behaviour
On changing of the conversation a page reload is done

### Actual behaviour
You can still change conversations and even start calls with the "old" UI being loaded

nickvergessen:
### Steps
1. Load a conversation
2. Modify a config or place `context.commit('setInitialNextcloudTalkHash', hash + hash)` in
https://github.com/nextcloud/spreed/blob/fcd7960c18709fa516cc1757395cf8ff26a72fd7/src/store/talkHashStore.js#L65 before step 1
3. After the next room (list) refresh you should see `X-Nextcloud-Talk-Hash marked dirty:  RANDOM_SHA1` in the console
4. "Start call" button is now disabled
5. Clicking on another conversation will make a page reload

Fix #3672

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -56,6 +56,7 @@ import {
 } from './utils/webrtc/index'
 import { emit } from '@nextcloud/event-bus'
 import browserCheck from './mixins/browserCheck'
+import talkHashCheck from './mixins/talkHashCheck'
 
 export default {
 	name: 'App',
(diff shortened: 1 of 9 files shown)


nickvergessen:
> Note that, when the external signaling server is used, there is no periodic fetching of the conversation when Talk is in the sidebars;

Hmm, I didn't even know that. But yeah, the main issue is people being active on talk all day. I assume you don't idle in the open sidebar 2-3h and then start a call.

PR #3686, opened 2020-05-26
Don't request avatars for user null

nickvergessen:
### Steps
1. Have a call with a guest
2. See a request for `https://localhost/index.php/avatar/null/300`


src/components/CallView/shared/VideoBackground.vue
```diff
@@ -25,8 +25,9 @@
 		<img
 			v-if="hasPicture"
 			:src="backgroundImage"
-			class="video-background__picture">
-		<div v-if="!hasPicture"
+			class="video-background__picture"
+			alt="">
```
jancborchardt, line 29: :heart:

PR #3687, opened 2020-05-26
Store the guest name in browser storage

nickvergessen:
### Step
1. As a guest join a conversation
2. Set a guest user name
3. Do a page refresh (e.g. because being prompted by the "Talk was updated" notice)
4. Check that your name is still set

### DIFF
--- a/src/components/SetGuestUsername.vue
+++ b/src/components/SetGuestUsername.vue
@@ -62,6 +62,7 @@ export default {
 		return {
 			guestUserName: '',
 			isEditingUsername: false,
+			delayHandleUserNameFromBrowserStorage: false,
 		}
 	},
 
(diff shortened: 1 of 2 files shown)


src/components/SetGuestUsername.vue
```diff
@@ -62,18 +63,46 @@ export default {
...
+			if (this.delayHandleUserNameFromBrowserStorage) {
+				console.debug('Saving guest name from browser storage to the session')
+				this.handleChooseUserName()
+				this.delayHandleUserNameFromBrowserStorage = false
+			}
+		},
+	},
+
+	mounted() {
+		this.guestUserName = BrowserStorage.getItem('guestName')
```
nickvergessen, line 93: Todo https://github.com/nextcloud/nextcloud-browser-storage/issues/3

nickvergessen:
Ready for review now.

Added a todo, to use `@nextcloud/browser-storage` or `OCP` when https://github.com/nextcloud/nextcloud-browser-storage/issues/3 got fixed in server. But since this is unlikely to backport, we just do this manually in the old way for now.

PR #3692, opened 2020-05-27
Add bottombar in screensharings

ma12-co:
## How to test

- create a call with at least 2 participants
- share the screen with one of them 

### before this pr:
no name or unfollow button were displayed;

### after this pr: 
participant name and unfollow button are displayed

### DIFF
--- a/src/components/CallView/shared/VideoBottomBar.vue
+++ b/src/components/CallView/shared/VideoBottomBar.vue
@@ -1,0 +1,270 @@
+<!--
+  - @copyright Copyright (c) 2020 Marco Ambrosini <marcoambrosini@pm.me>
+  -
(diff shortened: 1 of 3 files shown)


src/components/CallView/shared/VideoBottomBar.vue
```diff
@@ -0,0 +1,288 @@
...
+		boldenNameIndicator() {
+			return this.isSpeaking || this.isSelected
+		},
+
+		hasSelectedVideo() {
+			return this.$store.getters.selectedVideoPeerId !== null
+		},
+
+		stopFollowingLabel() {
+			return t('spreed', `Stop following`)
```
nickvergessen, line 174:
```suggestion
			return t('spreed', 'Stop following')
```

PR #3708, opened 2020-05-28
Bugfix/noid/switch to promoted view upon screenshare

ma12-co:


nickvergessen:
Doesn't seem to work for the person doing the screenshare.

PR #3738, opened 2020-06-04
Special handling of CLI administrators in system messages

### LINKED ISSUE #3664: Special handling of CLI interactions when managing users
## Steps to reproduce
1. Add a user via CLI to a conversation (#3465)
2. Chat shows the user was added by an administrator

### Expected behaviour
Clear message that the user was added by an admin

### Actual behaviour
Bildschirmfoto von 2020-05-25 14-28-54

cc @PhrozenByte in case you want to look into it, otherwise I might also have a look at some point

nickvergessen:
Fix #3664

### DIFF
--- a/lib/Chat/Parser/SystemMessage.php
+++ b/lib/Chat/Parser/SystemMessage.php
@@ -98,16 +98,22 @@ class SystemMessage {
 			$currentUserIsActor = $parsedParameters['actor']['type'] === 'guest' &&
 				sha1($participant->getSessionId()) === $parsedParameters['actor']['id'];
 		}
+		$cliIsActor = $parsedParameters['actor']['type'] === 'guest' &&
+			'guest/cli' === $parsedParameters['actor']['id'];
 
 		if ($message === 'conversation_created') {
 			$parsedMessage = $this->l->t('{actor} created the conversation');
 			if ($currentUserIsActor) {
 				$parsedMessage = $this->l->t('You created the conversation');
+			} elseif ($cliIsActor) {
+				$parsedMessage = $this->l->t('An administrator created the conversation');
 			}
 		} elseif ($message === 'conversation_renamed') {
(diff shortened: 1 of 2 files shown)


lib/Chat/Parser/SystemMessage.php
```diff
@@ -221,6 +259,11 @@ public function parseMessage(Message $chatMessage): void {
 				$parsedMessage = $this->l->t('You promoted {user} to moderator');
 			} elseif ($participant->isGuest() && $participant->getSessionId() === $parsedParameters['user']['id']) {
 				$parsedMessage = $this->l->t('{actor} promoted you to moderator');
+				if ($cliIsActor) {
+					$parsedMessage = $this->l->t('An administrator promoted you to moderator');
+				}
+			} elseif ($cliIsActor) {
+				$parsedMessage = $this->l->t('An administrator promoted {user} to moderator');
```
PhrozenByte, line 266: `guest_moderator_promoted` is never reached since we can't promote guests to moderators via CLI.

lib/Chat/Parser/SystemMessage.php
```diff
@@ -229,6 +272,11 @@ public function parseMessage(Message $chatMessage): void {
 				$parsedMessage = $this->l->t('You demoted {user} from moderator');
 			} elseif ($participant->isGuest() && $participant->getSessionId() === $parsedParameters['user']['id']) {
 				$parsedMessage = $this->l->t('{actor} demoted you from moderator');
+				if ($cliIsActor) {
+					$parsedMessage = $this->l->t('An administrator demoted you from moderator');
+				}
+			} elseif ($cliIsActor) {
+				$parsedMessage = $this->l->t('An administrator demoted {user} from moderator');
```
PhrozenByte, line 279: `guest_moderator_demoted` is never reached since we can't demote guests from moderators via CLI.

lib/Chat/Parser/SystemMessage.php
nickvergessen: well we can't so far, I'd like to keep it in case we do it at some point.

PR #3739, opened 2020-06-04
Equalize background blur

### LINKED ISSUE #3542: Equalize the blur between stripe videos and promoted videos
The blur radius is expressed in pixels atm so if the video is very big (promoted), the background results in a much less blurry picture.

ma12-co:
Add same amount of relative blur regardless of the actual size of the image.

fix #3542

Before:
Screenshot from 2020-06-04 11-27-25

After:
Screenshot from 2020-06-04 11-28-03

### DIFF
--- a/src/components/CallView/Grid/Grid.vue
+++ b/src/components/CallView/Grid/Grid.vue
@@ -364,12 +364,7 @@ export default {
 
 		// Blur radius for each background in the grid
 		videoBackgroundBlur() {
-			// The amount of blur
-			const amount = this.$store.getters.videoBackgroundBlur
-			// Represents the surface of the element
-			const surfaceMultiplier = (this.videoWidth * this.videoHeight) / 1000
-			// Calculate the blur
-			return `filter: blur(${surfaceMultiplier * amount}px)`
+			return this.$store.getters.getBlurFilter(this.videoWidth, this.videoHeight)
 		},
 	},
 
--- a/src/components/CallView/Grid/Grid.vue
+++ b/src/components/CallView/Grid/Grid.vue
@@ -49,6 +49,7 @@
 							:is-selected="isSelected(callParticipantModel)"
 							:fit-video="false"
 							:video-container-aspect-ratio="videoContainerAspectRatio"
+							:video-background-blur="videoBackgroundBlur"
 							:shared-data="sharedDatas[callParticipantModel.attributes.peerId]"
 							@click-video="handleClickVideo($event, callParticipantModel.attributes.peerId)" />
 					</template>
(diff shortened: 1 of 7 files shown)


src/components/CallView/shared/VideoBackground.vue
```diff
@@ -21,10 +21,14 @@
 
 <template>
 	<div class="video-backgroundbackground">
+		<ResizeObserver
```
nickvergessen, line 24: Can we avoid having the resize imported there? I'm not sure it scales well.
If you are in a call with 25 people we watch 25 elements for resize all the time,
but we could also just trigger it when the grid is updated?

ma12-co: I don't think it can be avoided, do you have something in mind?
Also notice that this is not a grid-only problem: it affects every video component

ma12-co: Ok, I moved the logic for the grid videos to the grid component and left the resize observer only for the "big" background.

src/components/CallView/shared/Video.vue
```diff
@@ -48,7 +48,8 @@
 				class="avatar-container">
 				<VideoBackground
 					:display-name="model.attributes.name"
-					:user="model.attributes.userId" />
+					:user="model.attributes.userId"
+					:grid-blur="videoBacgroundBlur" />
```
nickvergessen, line 52:
```suggestion
					:grid-blur="videoBackgroundBlur" />
```

src/components/CallView/Grid/Grid.vue
```diff
@@ -355,6 +361,16 @@ export default {
...
 		},
+
+		// Blur radius for each bacground in the grid
+		videoBackgroundBlur() {
+			// The amount of blur
+			const amount = this.$store.getters.videoBackgroundBlur
+			// Represents the surface of the element
+			const surfaceMultiplier = (this.videoWidth * this.videoHeight) / 100
+			// Calculate the blur
+			return `filter: blur(${surfaceMultiplier * amount}px)`
```
nickvergessen, line 372: extract the math to another computed and add the blur factor to the grid dev details view?

src/components/CallView/shared/VideoBackground.vue
```diff
@@ -49,11 +58,16 @@ export default {
 			type: String,
 			default: '',
 		},
+		gridBlur: {
+			type: String,
+			required: true,
```
nickvergessen, line 63: So this is required, but you don't set it in LocalVideo

nickvergessen:
Rebased and added one last commit to fix the blur size on the localVideo when switching to the grid.

PR #3740, opened 2020-06-04
Fix talk:room:* command completion for participant options

PhrozenByte:
We can't deal with guest participants in CLI commands (we rely on user ids) and `array_map()` currently returns a empty string for guest participants that is stripped later anyway. This fix makes this behaviour just more explicit.

I don't think that we should support guest participants in CLI commands in general anyway, can't think of a reasonable use case. What do you guys think?

Guess we should backport this to stable19, too.

### DIFF
--- a/lib/Command/Room/TRoomCommand.php
+++ b/lib/Command/Room/TRoomCommand.php
@@ -369,8 +369,13 @@ trait TRoomCommand {
 			return [];
 		}
 
-		return array_map(function (Participant $participant) {
-			return $participant->getUser();
-		}, $room->searchParticipants($context->getCurrentWord()));
+		$users = [];
+		foreach ($room->searchParticipants($context->getCurrentWord()) as $participant) {
+			if (!$participant->isGuest()) {
+				$users[] = $participant->getUser();
+			}
+		}
+
+		return $users;
 	}
 }


nickvergessen:
I also noticed issues with users that contain a space. can you have a look?
I guess we need to quote them somehow?

PhrozenByte:
Indeed. But no issue on our side, it's an issue with `symfony-console-completion` that was fixed more than a year ago. https://github.com/nextcloud/3rdparty/pull/388 must be merged. Can you initiate this?

lib/Command/Room/TRoomCommand.php
```diff
@@ -369,8 +369,13 @@ protected function completeParticipantValues(CompletionContext $context): array
...
-			return $participant->getUser();
-		}, $room->searchParticipants($context->getCurrentWord()));
+		$users = [];
+		foreach ($room->searchParticipants($context->getCurrentWord()) as $participant) {
+			if (!$participant->isGuest()) {
+				$users[] = $participant->getUser();
+			}
+		}
+
+		return $users;
```
nickvergessen, line 379: return array_map(function (Participant $participant) {
```suggestion
return array_filter(array_map(function (Participant $participant) {
			return $participant->getUser();
		}, $room->searchParticipants($context->getCurrentWord())));
```

should work too?

PhrozenByte: It indeed should work, too. But that would be `O(2)` complexity compared to `O(1)`. It's not like it makes much of a difference, I also feel like `foreach` is more readable than `array_filter()` + `array_map()`. But if you want it changed I'll change it. Just let me know :smiley:

nickvergessen:
> Indeed. But no issue on our side, it's an issue with symfony-console-completion that was probably (didn't check, but the changelog reads about fine) fixed more than a year ago. nextcloud/3rdparty#388 must be merged. Can you initiate this?

Done and merged, but I guess it will only come with Nextcloud 20 :/

PR #3743, opened 2020-06-04
Defer sending the notifications when we interact multiple times

nickvergessen:
- [x] Requires https://github.com/nextcloud/server/pull/21238
- [x] Has impact after https://github.com/nextcloud/notifications/pull/652


nickvergessen:
> Should the flushing be added to CheckHostedSignalingServer.php too?

Since this is in a background job, I don't care too much.

PR #3744, opened 2020-06-04
Dispatch an event to signal interest in defering the notifications

nickvergessen:
Backportable hack for the biggest issue of https://github.com/nextcloud/spreed/pull/3743

This can be backported easily and when the notifications app is running https://github.com/nextcloud/notifications/pull/651 it will help making everything faster

### DIFF
--- a/lib/Notification/Listener.php
+++ b/lib/Notification/Listener.php
@@ -29,7 +29,6 @@ use OCA\Talk\Events\RoomEvent;
 use OCA\Talk\Room;
 use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\EventDispatcher\Event;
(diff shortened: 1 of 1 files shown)


nickvergessen:
To test:

1. Also apply the notifications app PR: https://github.com/nextcloud/notifications/pull/651
2. On top of that apply this in the notifications app https://gist.github.com/nickvergessen/8f53db1960b4d4933072b920edcbb74f
3. Create a conversation with multiple users (the more that have a push notification device assigned the better)
4. Start a call in that conversation
5. Check the log file `/var/www/data/notifications.log` from 2.

Before there was one request to the push proxy per user, now it's 1 per proxy, so should be 1 for everyone. This should be noticeable when comparing the duration

PR #3745, opened 2020-06-04
Less database queries when parsing multiple notifications

nickvergessen:
Currently when a call is started, we run:
1. O(n) queries to get the same room for every one
2. O(n) queries to check if the participant is in the conversation

We can reduce 1. to O(1) and remove 2. as we just looped outside over the participants to generate these notifications to begin with.

### DIFF
--- a/lib/Notification/Notifier.php
+++ b/lib/Notification/Notifier.php
@@ -75,6 +75,8 @@ class Notifier implements INotifier {
 
 	/** @var Room[] */
 	protected $rooms = [];
+	/** @var Participant[][] */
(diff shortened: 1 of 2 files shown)


nickvergessen:
It is already tested with:
https://github.com/nextcloud/spreed/pull/3745/files#diff-c11fe75116fbd13fc688d342253b70c8R245
```php
		$room->expects($this->once())
			->method('getParticipant')
```

I was just too lazy to rename it.

nickvergessen:
Okay, then no backport to stable18

PR #3748, opened 2020-06-05
Make the key for participants stable

nickvergessen:
Before | After
---|---
n-th menu stays open | menu on the same participant stays open
before | after


nickvergessen:
Well then no backport to 18

PR #3749, opened 2020-06-05
Use the correct token to check the participant

nickvergessen:
Was only noticable when your servers post request was blasing fast.
You didn't switch conversations then, as we checked the participant index with teh right token, but when getting the actual participant we took another token.


nickvergessen:
Okay then no backport to 18

PR #3285, opened 2020-04-07
Ping all session ids with 1 query from the external signaling

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>

PR #3755, opened 2020-06-05
Make sure pinging session ids only uses indexes

nickvergessen:
Since sessions are unique globally (apart from '0' which is filtered out before)
we can update the ping only based on the index instead of
index + additional where check

See https://github.com/nextcloud/spreed/pull/3285#issuecomment-639433523 for analysis


nickvergessen:
No, the Ping function uses user, token and session which is a unique key and therefor even better/faster/less locking

PR #3757, opened 2020-06-05
Don't ping the session while waiting for chat messages

nickvergessen:


nickvergessen:
Actually let's think about this a bit more, as it seems the mobile clients don't do signaling when the just chat

PR #3816, opened 2020-06-16
Use 'number' for countable 'videos'

ChristophWurst:
*videos* is a *countable* word, hence *number* should be use instead of *amount*.

Ref https://writingexplained.org/amount-vs-number-difference

### DIFF
--- a/src/components/TopBar/TopBar.vue
+++ b/src/components/TopBar/TopBar.vue
@@ -241,7 +241,7 @@ export default {
 		},
 
 		layoutHintText() {
-			return t('Spreed', `The amount of videos don't fit in the window. Maximize or switch to 'promoted view' for a better experience.`)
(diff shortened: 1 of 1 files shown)


jancborchardt:
Good find. Even better if we are specific (and shorter):
> Too many videos to fit in the window. Maximize or switch to 'promoted view' for a better experience.

src/components/TopBar/TopBar.vue
```diff
@@ -241,7 +241,7 @@ export default {
 		},
 
 		layoutHintText() {
-			return t('Spreed', `The amount of videos don't fit in the window. Maximize or switch to 'promoted view' for a better experience.`)
+			return t('Spreed', `Too many videos to fit in the window. Maximize or switch to 'promoted view' for a better experience.`)
```
nickvergessen, line 244:
```suggestion
			return t('Spreed', `Too many videos to fit in the window. Maximize the window or switch to 'promoted view' for a better experience.`)
```

PR #3817, opened 2020-06-16
Use usernameToColor from @nextcloud/vue

nickvergessen:
- [x] Requires https://github.com/nextcloud/nextcloud-vue/pull/1158


nickvergessen:
Ready to review

PR #3822, opened 2020-06-17
Services: Start room service

nickvergessen:
Currently only room creation is done, but I prefer small steps in the migration.
Integration tests should confirm that everything still works as intended

### DIFF
--- a/lib/Command/Room/Create.php
+++ b/lib/Command/Room/Create.php
@@ -27,6 +27,7 @@ namespace OCA\Talk\Command\Room;
 
 use InvalidArgumentException;
 use OC\Core\Command\Base;
+use OCA\Talk\Room;
 use Stecman\Component\Symfony\Console\BashCompletion\CompletionContext;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
@@ -98,7 +99,8 @@ class Create extends Base {
 			return 1;
 		}
 
-		$room = $public ? $this->manager->createPublicRoom($name): $this->manager->createGroupRoom($name);
+		$roomType = $public ? Room::PUBLIC_CALL : Room::GROUP_CALL;
+		$room = $this->roomService->createConversation($roomType, $name);
 
 		try {
 			$this->setRoomReadOnly($room, $readonly);
--- a/lib/Command/Room/Create.php
+++ b/lib/Command/Room/Create.php
@@ -93,14 +93,13 @@ class Create extends Base {
 		$owner = $input->getOption('owner');
 		$moderators = $input->getOption('moderator');
 
-		$name = trim($name);
-		if (!$this->validateRoomName($name)) {
+		$roomType = $public ? Room::PUBLIC_CALL : Room::GROUP_CALL;
+		try {
+			$room = $this->roomService->createConversation($roomType, $name);
+		} catch (\InvalidArgumentException $e) {
 			$output->writeln('<error>Invalid room name.</error>');
 			return 1;
 		}
-
-		$roomType = $public ? Room::PUBLIC_CALL : Room::GROUP_CALL;
-		$room = $this->roomService->createConversation($roomType, $name);
 
 		try {
 			$this->setRoomReadOnly($room, $readonly);
--- a/lib/Command/Room/Create.php
+++ b/lib/Command/Room/Create.php
@@ -97,8 +97,11 @@ class Create extends Base {
 		try {
 			$room = $this->roomService->createConversation($roomType, $name);
 		} catch (\InvalidArgumentException $e) {
-			$output->writeln('<error>Invalid room name.</error>');
-			return 1;
+			if ($e->getMessage() === 'name') {
+				$output->writeln('<error>Invalid room name.</error>');
+				return 1;
+			}
+			throw $e;
 		}
 
 		try {
--- a/lib/Command/Room/Create.php
+++ b/lib/Command/Room/Create.php
@@ -96,7 +96,7 @@ class Create extends Base {
 		$roomType = $public ? Room::PUBLIC_CALL : Room::GROUP_CALL;
 		try {
 			$room = $this->roomService->createConversation($roomType, $name);
-		} catch (\InvalidArgumentException $e) {
+		} catch (InvalidArgumentException $e) {
 			if ($e->getMessage() === 'name') {
 				$output->writeln('<error>Invalid room name.</error>');
 				return 1;
--- a/lib/Controller/RoomController.php
+++ b/lib/Controller/RoomController.php
@@ -27,6 +27,7 @@ declare(strict_types=1);
 
 namespace OCA\Talk\Controller;
 
+use InvalidArgumentException;
 use OCA\Circles\Api\v1\Circles;
 use OCA\Circles\Model\Member;
 use OCA\Talk\Chat\ChatManager;
@@ -612,7 +613,7 @@ class RoomController extends AEnvironmentAwareController {
 		try {
 			$room = $this->roomService->createOneToOneConversation($currentUser, $targetUser);
 			return new DataResponse($this->formatRoom($room, $room->getParticipant($currentUser->getUID())), Http::STATUS_CREATED);
-		} catch (\InvalidArgumentException $e) {
+		} catch (InvalidArgumentException $e) {
 			// Same current and target user
 			return new DataResponse([], Http::STATUS_FORBIDDEN);
 		} catch (RoomNotFoundException $e) {
@@ -736,7 +737,7 @@ class RoomController extends AEnvironmentAwareController {
 		// Create the room
 		try {
 			$room = $this->roomService->createConversation($roomType, $roomName, $currentUser);
-		} catch (\InvalidArgumentException $e) {
+		} catch (InvalidArgumentException $e) {
 			return new DataResponse([], Http::STATUS_BAD_REQUEST);
 		}
 
--- a/lib/Controller/RoomController.php
+++ b/lib/Controller/RoomController.php
@@ -631,7 +631,8 @@ class RoomController extends AEnvironmentAwareController {
 		}
 
 		// Create the room
-		$room = $this->roomService->createConversation(Room::GROUP_CALL, $targetGroup->getDisplayName(), $currentUser);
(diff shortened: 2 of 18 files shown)


tests/php/Command/Room/CreateTest.php
```diff
@@ -286,17 +290,18 @@ public function validProvider(): array {
...
 
-		$this->manager
-			->method('createPublicRoom')
-			->willReturnCallback(function (string $name = ''): Room {
-				return $this->roomMockContainer->create(['name' => $name, 'type' => Room::PUBLIC_CALL]);
-			});
+		if ($input['name'] !== 'PHPUnit Test Room') {
+			$this->roomService
+				->method('createConversation')
+				->willThrowException(new \InvalidArgumentException('name'));
```
PhrozenByte, line 297: I feel like this is an hack (i.e. we would create another technical debt here). The fact that behaviour differs (for invalid names a exception is thrown, for anything else not; just like with `UpdateTest` vs. `CreateTest` - for the latter it's an exception, for `UpdateTest` it's not) is a problem.

lib/Command/Room/Create.php
```diff
@@ -92,14 +93,14 @@ protected function execute(InputInterface $input, OutputInterface $output): ?int
...
 
-		$name = trim($name);
-		if (!$this->validateRoomName($name)) {
+		$roomType = $public ? Room::PUBLIC_CALL : Room::GROUP_CALL;
+		try {
+			$room = $this->roomService->createConversation($roomType, $name);
+		} catch (\InvalidArgumentException $e) {
 			$output->writeln('<error>Invalid room name.</error>');
 			return 1;
 		}
```
PhrozenByte, line 102: It is never really tested whether the room name is at fault... The fact that `InvalidArgumentException` is thrown for invalid names only right now is just a implementation detail of `RoomService::createConversation()`.

lib/Service/RoomService.php
```diff
@@ -0,0 +1,86 @@
...
+				'userId' => $owner->getUID(),
+				'participantType' => Participant::OWNER,
+			]);
+		}
+
+		return $room;
+	}
+
+	public function prepareConversationName(string $objectName): string {
+		return trim(mb_substr(trim($objectName), 0, 64));
```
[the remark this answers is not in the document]
nickvergessen: > I'm not sure whether silently truncating the room name is a good idea? I feel like this should be an error.

Since we sometimes automatically generate the name e.g. for file conversations or video verification, as well as when you select a group or circle, automatic truncate is the best idea I guess.

For manual name setting an error is better I agree, which is why `createEmptyRoom` doesn't prepare the name and instead returns a 400 as before.

tests/php/Command/Room/CreateTest.php
```diff
@@ -318,18 +323,6 @@ public function invalidProvider(): array {
...
 			],
-			[
-				[
-					'name' => '  ',
-				],
-				"Invalid room name.\n",
-			],
-			[
-				[
-					'name' => str_repeat('x', 256),
```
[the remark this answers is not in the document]
nickvergessen: It is now part of the RoomService and the creation throws an exception. The CreateCommand itself does not interact with the room name anymore. So it will be covered by unit tests for the RoomService later.

tests/php/Command/Room/CreateTest.php
```diff
@@ -318,18 +323,6 @@ public function invalidProvider(): array {
 				],
 				"Invalid room name.\n",
 			],
-			[
-				[
-					'name' => '  ',
```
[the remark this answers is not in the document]
nickvergessen: Same as above

tests/php/Command/Room/CreateTest.php
nickvergessen: Well this is just temporarily until the other validations are moved from the controller to the service

tests/php/Command/Room/CreateTest.php
```diff
@@ -53,7 +57,16 @@ public function setUp(): void {
 		$this->registerGroupManagerMock();
 
 		$this->manager = $this->createMock(Manager::class);
-		$this->command = new Create($this->manager, $this->userManager, $this->groupManager);
+		$this->roomService = $this->getMockBuilder(RoomService::class)
+			->setMethods(['createConversation'])
+			->setConstructorArgs([$this->manager])
+			->getMock();
```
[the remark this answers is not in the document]
nickvergessen: It's a mistake from an attempt

lib/Command/Room/Create.php
nickvergessen: Yeah, future "to be written" tests for the RoomService will make sure it throws on invalid room names

nickvergessen: Full tests written now

tests/php/Command/Room/CreateTest.php
nickvergessen: type and object data also throws now

nickvergessen:
Thanks for the quick review @PhrozenByte should be all fixed now

tests/php/Command/Room/CreateTest.php
PhrozenByte: Very good :+1: 

Only one last thing: We shouldn't hard code the valid room name in the unit test. How about the following and adding a third data parameter to `invalidProvider` for those data sets which are expected to fail with an exception:

```php
	public function testInvalid(array $input, string $expectedOutput, ?Exception $expectedException = null): void {
		if ($expectedException !== null) {
			$this->roomService
				->method('createConversation')
				->willThrowException($expectedException);
		} else {
			$this->roomService
				->method('createConversation')
				->willReturnCallback(function (int $type, string $name): Room {
					return $this->roomMockContainer->create(['name' => $name, 'type' => $type]);
				});
		}
```

lib/Command/Room/Create.php
PhrozenByte: Excellent :+1: 

However, my comment rather was about assuming that a `InvalidArgumentException` is thrown for invalid room names only. In other words: There's a `$e->getMessage() === 'name'` check missing.

lib/Controller/RoomController.php
```diff
@@ -581,40 +586,28 @@ public function createRoom(int $roomType, string $invite = '', string $roomName
...
-			return new DataResponse([], Http::STATUS_FORBIDDEN);
-		}
-
-		// If room exists: Reuse that one, otherwise create a new one.
 		try {
-			$room = $this->manager->getOne2OneRoom($this->userId, $targetUser->getUID());
-			$room->ensureOneToOneRoomIsFilled();
+			$room = $this->roomService->createOneToOneConversation($currentUser, $targetUser);
 			return new DataResponse($this->formatRoom($room, $room->getParticipant($currentUser->getUID())), Http::STATUS_OK);
+		} catch (\InvalidArgumentException $e) {
```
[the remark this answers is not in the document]
PhrozenByte: Still a FQN.

lib/Service/RoomService.php
```diff
@@ -62,10 +68,42 @@ public function createOneToOneConversation(IUser $actor, IUser $targetUser): Roo
...
 	}
 
+	/**
+	 * @param int $type
+	 * @param string $name
+	 * @param IUser|null $owner
+	 * @param string $objectType
+	 * @param string $objectId
+	 * @return Room
+	 * @throws InvalidArgumentException on too long or empty names
```
PhrozenByte, line 78: Wrong description.

lib/Controller/RoomController.php
```diff
@@ -731,27 +726,19 @@ protected function createCircleRoom(string $targetCircleId): DataResponse {
...
+
 		// Create the room
-		if ($public) {
-			$room = $this->manager->createPublicRoom($roomName);
-		} else {
-			$room = $this->manager->createGroupRoom($roomName);
+		try {
+			$room = $this->roomService->createConversation($roomType, $roomName, $currentUser);
+		} catch (\InvalidArgumentException $e) {
+			return new DataResponse([], Http::STATUS_BAD_REQUEST);
```
PhrozenByte, line 740: Should we expose what field is at fault in the response?

tests/php/Service/RoomServiceTest.php
```diff
@@ -0,0 +1,253 @@
...
+		$this->service = new RoomService(
+			$this->manager
+		);
+	}
+
+	public function testCreateOneToOneConversationWithSameUser(): void {
+		$user = $this->createMock(IUser::class);
+		$user->expects($this->exactly(2))
+			->method('getUID')
+			->willReturn('uid');
```
PhrozenByte, line 56: Isn't the fact how many times `getUID` is called an implementation detail that doesn't affect the result?

```suggestion
		$user->method('getUID')
			->willReturn('uid');
```

Same for all usages below.

lib/Controller/RoomController.php
nickvergessen: I'm not sure it's of a lot of help atm, the only invalid input is the name, as object data can not be specified via API and the type is hard coded in the controller.

tests/php/Service/RoomServiceTest.php
nickvergessen: done

lib/Service/RoomService.php
nickvergessen: extended

lib/Command/Room/Create.php
nickvergessen: done

tests/php/Command/Room/CreateTest.php
nickvergessen: well if okay for you I would skip that for now, until all parts have been moved over to RoomServce and then we can revisit the test structure.

nickvergessen:
Mind to have another look @PhrozenByte ?

tests/php/Command/Room/CreateTest.php
PhrozenByte: Sounds reasonable :+1:

tests/php/Service/RoomServiceTest.php
PhrozenByte: All except this one :smile:

nickvergessen: Ops

nickvergessen:
Fixed the one `expects()` and rebased on latest master.

Thanks for the review!

Issue #1164, opened 2018-08-29
Chat flow experience


nickvergessen:
After a little discussion we decided to handle the chat like the following:

The conversation list contains the last read message id, from there on the flow should be the following:
* [x] A `-- 🆕 --` marker is shown at 1/3 of the screen
* [x] Below it a set of new messages is loaded (covering the bottom 2/3rds of the screen)
* [x] Above the marker old messages are being loaded, serving context for the new messages

More details:
* [x] When there are too few "new" messages available, of course the `-- 🆕 --` marker should be moved down (and hidden when there are none at all) and more old messages should be shown instead.
* [x] When you scroll up, the read marker is not changed.
* [x] When you scroll down, the read marker should be set to the last message id that was displayed on the screen.
* [ ] If you scroll up, after scrolling down, *ideally* the `-- 🆕 --` marker is moved after the message that is the last one that was on the screen.
* [ ] If you are far in the past or not "at the bottom" and there is a new chat message written just now, a "⏬" button should be displayed, which allows you to jump to the bottom of the conversation. (current implementations of iOS and Android are fine)
* [ ] If there are unread mentions between `-- 🆕 --` and the bottom of the conversation (room API `unreadMention`), a `@` button should be displayed, which allows you to jump to the next unread mention.

```
  old
  old
-- 🆕 --
  new
  new
  new     
  new     @  ⏬
```

### Sub issues

- [ ] #3825

nickvergessen:
# Pre 17
Currently both mobile clients seem to do the following:
1. `lookIntoFuture=0&lastKnownMessageId=-1` or `lookINtoFuture=0&lastKnownMessageId=0`
2. From the response they take the newest message `<23>` (not the one from the header) to start waiting for new messages: `lookIntoFuture=1&lastKnownMessageId=<23>`

# 17
Instead the following should happen (after #2135  is in, so beta2, but you can assume it works like this in all 7, so when the `chat-read-marker` capability is there)
1. Get `lastReadMessage` `<42>` from the conversation list for this conversation
2. Load history: `lookIntoFuture=0&lastKnownMessageId=<42>&includeLastKnown=1`
3. Load future (can happen in parallel): `lookIntoFuture=1&lastKnownMessageId=<42>`

# 17 but nice :heart_eyes_cat: 
[...]

PR #3825, opened 2020-06-18
Add unread message marker in the chat view

### LINKED ISSUE #1788: Indicator for read position
I left the browser open while doing something else, chat is going on. What did i miss in the conversation? 

Show an indicator for new messages that happened while you were not active.

IRCCloud does it like this:
D0844F44-5B79-4F50-B4BE-EC95347B0F97

**Additional context**
Possibly related to #798. Should messages that occurred while i tabbed out be considered unread?

nickvergessen:
~~Fix #1788~~
Fixes https://github.com/nextcloud/spreed/issues/1164

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -28,6 +28,8 @@ the main body of the message as well as a quote.
 	<li
 		:id="`message_${id}`"
 		ref="message"
+		:data-message-id="id"
+		:data-seen="seen"
 		class="message"
 		:class="{'hover': showActions && !isSystemMessage && !isDeletedMessage, 'system' : isSystemMessage}"
 		@mouseover="handleMouseover"
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -169,7 +169,7 @@ the main body of the message as well as a quote.
 			<div v-if="isLastReadMessage"
 				v-observe-visibility="lastReadMessageVisibilityChanged">
 				<div v-if="!isLastMessage" class="new-message-marker">
-					<span>----- {{ t('spreed', 'Unread messages') }} -----</span>
+					<span>{{ t('spreed', 'Unread messages') }}</span>
 				</div>
 			</div>
 		</div>
@@ -801,14 +801,21 @@ export default {
 }
 
 .new-message-marker {
-	margin: 20px 0;
-	text-align: center;
+	position: relative;
+	margin: 40px 15px 0 -45px;
+	border-top: 1px solid var(--color-border);
 
 	span {
-		background-color: var(--color-primary-light);
-		color: var(--color-text-lighter);
-		padding: 10px  20px;
+		position: absolute;
+		top: 0;
+		left: 50%;
+		transform: translateX(-50%) translateY(-50%);
+		padding: 0 7px 0 7px;
+		text-align: center;
+		white-space: nowrap;
+
 		border-radius: var(--border-radius);
+		background-color: var(--color-main-background);
 	}
 }
 
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -30,6 +30,8 @@ the main body of the message as well as a quote.
 		ref="message"
 		:data-message-id="id"
 		:data-seen="seen"
+		:data-next-message-id="nextMessageId"
+		:data-previous-message-id="previousMessageId"
 		class="message">
 		<div
 			:class="{'hover': showActions && !isSystemMessage && !isDeletedMessage, 'system' : isSystemMessage}"
@@ -351,6 +353,16 @@ export default {
 		sendingFailure: {
 			type: String,
 			default: '',
+		},
+
+		previousMessageId: {
+			type: Number,
+			default: 0,
+		},
+
+		nextMessageId: {
+			type: Number,
+			default: 0,
 		},
 
 		lastReadMessageId: {
@@ -700,7 +712,7 @@ export default {
 			// update in backend
 			this.$store.dispatch('updateLastReadMessage', {
 				token: this.token,
-				id: this.id,
+				id: this.previousMessageId,
 			})
 			// update visually
 			this.$store.dispatch('setVisualLastReadMessageId', {
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -352,6 +352,11 @@ export default {
 			type: String,
 			default: '',
 		},
+
+		lastReadMessageId: {
+			type: Number,
+			default: 0,
+		},
 	},
 
 	data() {
@@ -368,7 +373,9 @@ export default {
 
 	computed: {
 		isLastReadMessage() {
-			return this.id === this.conversation.lastReadMessage
+			// note: not reading lastReadMessage from the conversation as we want to define it externally
+			// to have closer control on marker's visibility behavior
+			return this.id === this.lastReadMessageId
 				&& (!this.conversation.lastMessage
 				|| this.id !== this.conversation.lastMessage.id)
 		},
@@ -690,7 +697,16 @@ export default {
 		},
 
 		handleMarkAsUnread() {
-			this.$store.dispatch('updateLastReadMessage', { token: this.token, id: this.id })
+			// update in backend
+			this.$store.dispatch('updateLastReadMessage', {
+				token: this.token,
+				id: this.id,
+			})
+			// update visually
+			this.$store.dispatch('setVisualLastReadMessageId', {
+				token: this.token,
+				id: this.conversation.lastReadMessage,
+			})
 		},
 	},
 }
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -144,6 +144,12 @@ the main body of the message as well as a quote.
 								@click.stop.prevent="handleCopyMessageLink">
 								{{ t('spreed', 'Copy message link') }}
 							</ActionButton>
+							<ActionButton
+								v-if="isReplyable"
+								:close-after-click="true"
+								@click.stop="handleMarkAsUnread">
+								{{ t('spreed', 'Mark as unread') }}
+							</ActionButton>
 							<ActionSeparator v-if="messageActions.length > 0" />
 							<template
 								v-for="action in messageActions">
@@ -374,7 +380,7 @@ export default {
 		},
 
 		hasActionsMenu() {
-			return (this.isPrivateReplyable || this.isDeleteable || this.messageActions.length > 0) && !this.isConversationReadOnly
+			return (this.isPrivateReplyable || this.isReplyable || this.isDeleteable || this.messageActions.length > 0) && !this.isConversationReadOnly
 		},
 
 		isConversationReadOnly() {
@@ -683,6 +689,10 @@ export default {
 			} catch (error) {
 				showError(t('spreed', 'The link could not be copied.'))
 			}
+		},
+
+		handleMarkAsUnread() {
+			this.$store.dispatch('updateLastReadMessage', { token: this.token, id: this.id })
 		},
 	},
 }
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -32,137 +32,143 @@ the main body of the message as well as a quote.
 		:class="{'hover': showActions && !isSystemMessage && !isDeletedMessage, 'system' : isSystemMessage}"
 		@mouseover="handleMouseover"
 		@mouseleave="handleMouseleave">
-		<div v-if="isFirstMessage && showAuthor"
-			class="message__author"
-			role="heading"
-			aria-level="4">
-			{{ actorDisplayName }}
-		</div>
-		<div
-			ref="messageMain"
-			class="message__main">
-			<div v-if="isSingleEmoji"
-				class="message__main__text">
-				<Quote v-if="parent" :parent-id="parent" v-bind="quote" />
-				<div class="single-emoji">
-					{{ message }}
+		<div>
+			<div v-if="isFirstMessage && showAuthor"
+				class="message__author"
+				role="heading"
+				aria-level="4">
+				{{ actorDisplayName }}
+			</div>
+			<div
+				ref="messageMain"
+				class="message__main">
+				<div v-if="isSingleEmoji"
+					class="message__main__text">
+					<Quote v-if="parent" :parent-id="parent" v-bind="quote" />
+					<div class="single-emoji">
+						{{ message }}
+					</div>
 				</div>
-			</div>
-			<div v-else-if="showJoinCallButton" class="message__main__text call-started">
-				<RichText :text="message" :arguments="richParameters" :autolink="true" />
-				<CallButton />
-			</div>
-			<div v-else-if="isDeletedMessage" class="message__main__text deleted-message">
-				<RichText :text="message" :arguments="richParameters" :autolink="true" />
-			</div>
-			<div v-else class="message__main__text" :class="{'system-message': isSystemMessage}">
-				<Quote v-if="parent" :parent-id="parent" v-bind="quote" />
-				<RichText :text="message" :arguments="richParameters" :autolink="true" />
-			</div>
-			<div v-if="!isDeletedMessage" class="message__main__right">
-				<span
-					v-tooltip.auto="messageDate"
-					class="date"
-					:style="{'visibility': hasDate ? 'visible' : 'hidden'}"
-					:class="{'date--self': showSentIcon}">{{ messageTime }}</span>
-				<!-- Message delivery status indicators -->
-				<div v-if="sendingFailure"
-					v-tooltip.auto="sendingErrorIconTooltip"
-					class="message-status sending-failed"
-					:class="{'retry-option': sendingErrorCanRetry}"
-					:aria-label="sendingErrorIconTooltip"
-					tabindex="0"
-					@mouseover="showReloadButton = true"
-					@focus="showReloadButton = true"
-					@mouseleave="showReloadButton = true"
-					@blur="showReloadButton = true">
-					<button
-						v-if="sendingErrorCanRetry && showReloadButton"
-						class="nc-button nc-button__main--dark"
-						@click="handleRetry">
-						<Reload
+				<div v-else-if="showJoinCallButton" class="message__main__text call-started">
+					<RichText :text="message" :arguments="richParameters" :autolink="true" />
+					<CallButton />
+				</div>
+				<div v-else-if="isDeletedMessage" class="message__main__text deleted-message">
+					<RichText :text="message" :arguments="richParameters" :autolink="true" />
+				</div>
+				<div v-else class="message__main__text" :class="{'system-message': isSystemMessage}">
+					<Quote v-if="parent" :parent-id="parent" v-bind="quote" />
+					<RichText :text="message" :arguments="richParameters" :autolink="true" />
+				</div>
+				<div v-if="!isDeletedMessage" class="message__main__right">
+					<span
+						v-tooltip.auto="messageDate"
+						class="date"
+						:style="{'visibility': hasDate ? 'visible' : 'hidden'}"
+						:class="{'date--self': showSentIcon}">{{ messageTime }}</span>
+					<!-- Message delivery status indicators -->
+					<div v-if="sendingFailure"
+						v-tooltip.auto="sendingErrorIconTooltip"
+						class="message-status sending-failed"
+						:class="{'retry-option': sendingErrorCanRetry}"
+						:aria-label="sendingErrorIconTooltip"
+						tabindex="0"
+						@mouseover="showReloadButton = true"
+						@focus="showReloadButton = true"
+						@mouseleave="showReloadButton = true"
+						@blur="showReloadButton = true">
+						<button
+							v-if="sendingErrorCanRetry && showReloadButton"
+							class="nc-button nc-button__main--dark"
+							@click="handleRetry">
+							<Reload
+								decorative
+								title=""
+								:size="16" />
+						</button>
+						<AlertCircle v-else
 							decorative
 							title=""
 							:size="16" />
-					</button>
-					<AlertCircle v-else
-						decorative
-						title=""
-						:size="16" />
-				</div>
-				<div v-else-if="isTemporary && !isTemporaryUpload || isDeleting"
-					v-tooltip.auto="loadingIconTooltip"
-					class="icon-loading-small message-status"
-					:aria-label="loadingIconTooltip" />
-				<div v-else-if="showCommonReadIcon"
-					v-tooltip.auto="commonReadIconTooltip"
-					class="message-status"
-					:aria-label="commonReadIconTooltip">
-					<CheckAll decorative
-						title=""
-						:size="16" />
-				</div>
-				<div v-else-if="showSentIcon"
-					v-tooltip.auto="sentIconTooltip"
-					class="message-status"
-					:aria-label="sentIconTooltip">
-					<Check decorative
-						title=""
-						:size="16" />
-				</div>
-				<!-- Message Actions -->
-				<div
-					v-show="showActions"
-					class="message__main__right__actions"
-					:class="{ 'tall' : isTallEnough }">
-					<Actions
-						v-show="isReplyable">
-						<ActionButton
-							icon="icon-reply"
-							@click.stop="handleReply">
-							{{ t('spreed', 'Reply') }}
-						</ActionButton>
-					</Actions>
-					<Actions
-						v-show="hasActionsMenu"
-						:force-menu="true"
-						container="#content-vue">
-						<ActionButton
-							v-if="isPrivateReplyable"
-							icon="icon-user"
-							:close-after-click="true"
-							@click.stop="handlePrivateReply">
-							{{ t('spreed', 'Reply privately') }}
-						</ActionButton>
-						<ActionButton
-							v-if="isReplyable"
-							icon="icon-external"
-							:close-after-click="true"
-							@click.stop.prevent="handleCopyMessageLink">
-							{{ t('spreed', 'Copy message link') }}
-						</ActionButton>
-						<ActionSeparator v-if="messageActions.length > 0" />
-						<template
-							v-for="action in messageActions">
+					</div>
+					<div v-else-if="isTemporary && !isTemporaryUpload || isDeleting"
+						v-tooltip.auto="loadingIconTooltip"
+						class="icon-loading-small message-status"
+						:aria-label="loadingIconTooltip" />
+					<div v-else-if="showCommonReadIcon"
+						v-tooltip.auto="commonReadIconTooltip"
+						class="message-status"
+						:aria-label="commonReadIconTooltip">
+						<CheckAll decorative
+							title=""
+							:size="16" />
+					</div>
+					<div v-else-if="showSentIcon"
+						v-tooltip.auto="sentIconTooltip"
+						class="message-status"
+						:aria-label="sentIconTooltip">
+						<Check decorative
+							title=""
+							:size="16" />
+					</div>
+					<!-- Message Actions -->
+					<div
+						v-show="showActions"
+						class="message__main__right__actions"
+						:class="{ 'tall' : isTallEnough }">
+						<Actions
+							v-show="isReplyable">
 							<ActionButton
-								:key="action.label"
-								:icon="action.icon"
-								:close-after-click="true"
-								@click="action.callback(messageAPIData)">
-								{{ action.label }}
+								icon="icon-reply"
+								@click.stop="handleReply">
+								{{ t('spreed', 'Reply') }}
 							</ActionButton>
-						</template>
-						<template v-if="isDeleteable">
-							<ActionSeparator v-if="isReplyable || messageActions.length > 0" />
+						</Actions>
+						<Actions
+							v-show="hasActionsMenu"
+							:force-menu="true"
+							container="#content-vue">
 							<ActionButton
-								icon="icon-delete"
+								v-if="isPrivateReplyable"
+								icon="icon-user"
 								:close-after-click="true"
-								@click.stop="handleDelete">
-								{{ t('spreed', 'Delete') }}
+								@click.stop="handlePrivateReply">
+								{{ t('spreed', 'Reply privately') }}
 							</ActionButton>
-						</template>
-					</Actions>
+							<ActionButton
+								v-if="isReplyable"
+								icon="icon-external"
+								:close-after-click="true"
+								@click.stop.prevent="handleCopyMessageLink">
+								{{ t('spreed', 'Copy message link') }}
+							</ActionButton>
+							<ActionSeparator v-if="messageActions.length > 0" />
+							<template
+								v-for="action in messageActions">
+								<ActionButton
+									:key="action.label"
+									:icon="action.icon"
+									:close-after-click="true"
+									@click="action.callback(messageAPIData)">
+									{{ action.label }}
+								</ActionButton>
+							</template>
+							<template v-if="isDeleteable">
+								<ActionSeparator v-if="isReplyable || messageActions.length > 0" />
+								<ActionButton
+									icon="icon-delete"
+									:close-after-click="true"
+									@click.stop="handleDelete">
+									{{ t('spreed', 'Delete') }}
+								</ActionButton>
+							</template>
+						</Actions>
+					</div>
 				</div>
+			</div>
+			<div v-if="lastReadMessage === id"
+				class="new-message-marker">
+				<span>{{ t('spreed', 'New messages') }}</span>
 			</div>
 		</div>
 	</li>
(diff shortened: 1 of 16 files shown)


PVince81:
what's missing ? from reading the code it seems like it could already work.

let me know if I should take over

nickvergessen:
Handling the update of the value is still missing. once the next bunch of messages is loaded and the room list refreshes, the read marker will be on the newest message

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -197,6 +204,14 @@ export default {
 			type: Number,
 			default: 0,
 		},
+
+		/**
+		 * Id of the last read message when the chat was opened
+		 */
+		lastReadMessage: {
```
PVince81, line 211: feels a bit overkill having this property on all components.

wondering if we should read it directly from the messages store (the purpose of stores is to give access to commonly used attributes and remove the need to pass them in everywhere)

src/components/MessagesList/MessagesList.vue
```diff
@@ -424,8 +439,11 @@ export default {
 
 		/**
 		 * Creates a long polling request for a new message.
+		 * @param {boolean} scrollToBottom Whether we should try to automatically scroll to the bottom
 		 */
-		async getNewMessages() {
+		async getNewMessages(scrollToBottom) {
+			scrollToBottom = scrollToBottom === undefined ? true : scrollToBottom
```
PVince81, line 445: @nickvergessen note for the future: you can use default values for arguments in ES6 :-)

I'll adjust after rebase

PVince81:
rebased

PVince81:
> Handling the update of the value is still missing. once the next bunch of messages is loaded and the room list refreshes, the read marker will be on the newest message

Seems there's more to it.
The original ticket mentions how the read marker would move when scrolling up and down based on multiple conditions. The latter is challenging as it requires detecting visible/invisible elements. Not sure if you intended to have this in the scope of this PR ?

PVince81:
I had a quick look at ways for detecting the element visible at the top:
- https://github.com/AkatQuas/vue-visibility-sensor
   - cons: requires adding this to every Message DOM component, which could trigger a lot of logic when we have lots of messages
- document.elementFromPoint https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/elementFromPoint
   - pros: supported by all browsers
   - cons: marked as experimental, so not sure
   - cons: need some calculation gymnastics to find a good coordinate to query to make sure that we'll land on a message element

PVince81:
I'll go with `elementFromPoint` and `getBoundingClientRect` which are said to work on Edge at least.
I have something working locally.

Next up: tracking whether the unread marker was actually visible on screen or not.

Then we only update the unread marker position if the marker was seen at least one and the detected top element is below the marker.

src/components/MessagesList/MessagesGroup/Message/Message.vue
nickvergessen: I guess it makes sense to store it somewhere.

nickvergessen:
We have `vue-observe-visibility` already, can it do the job?

PVince81:
> We have `vue-observe-visibility` already, can it do the job?

Yes. But I won't set it on all the message elements as observing them all won't scale.
It will be useful for tracking the visibility of the read marker at least.

PVince81:
I've pushed some experiment which currently works for me in Chromium.

TODOs:[ ] fix left sidebar to not "mark as read" immediately when clicking, and the message count bubble
- [ ] remove console logs
- [ ] double check if we should show the marker at the top or at the bottom of message, might need to "shift by one"
- [ ] test: more testing with various scrolling scenarios
- [ ] test: double check if works with multi-line messages, the latter need to be fully visible to be considered as read

- [ ] check more todos from original ticket with special buttons to scroll to mention, etc (maybe for another PR)

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -331,10 +324,19 @@ export default {
 			isTallEnough: false,
 			showReloadButton: false,
 			isDeleting: false,
+			// whether the message was seen, only used if this was marked as last read message
+			seen: false,
```
PVince81, line 328: actually now thinking, this only means that the marker was seen, not the full message's bounding box
so it could cause situations where we mark as seen when it's not

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -546,6 +548,12 @@ export default {
 	},
 
 	methods: {
+		lastReadMessageVisibilityChanged(isVisible) {
+			if (isVisible) {
+				this.seen = true
```
PVince81, line 553: optional: move this to store or a global space ? (would save having an extra bool for every message)

PVince81:
- [ ] explore this alternative for finding the "element fully visible at the bottom of the screen":
   - pick the message at the current read marker
   - iterate towards the end of the list
   - check for each Message element if it is fully visible in the container
   - this way, we find the first message that is no more fully visible => set the read marker on the one preceding it
   - => this is slightly more efficient than trying to iterate over ALL messages and likely more reliable than the coordinates test

PVince81:
alright... with the current code I managed to have the point targetting the sapce between two message elements, so it couldn't find the last visible message...
I'll attempt the alternative approach proposed in the previous comment

PVince81:
- [ ] BUG: seems scrolledToBottom attribute is slightly buggy/unreliable

PVince81:
slowly getting there... it still feels a bit unreliable.

what works for me:

1. have a chatter bot post messages often
2. open the conversation in the UI
3. scroll to bottom
4. wait: here it autoscrolls
5. switch to another tab and wait for a few messages
6. go back: read marker is here
7. scroll down, it moves down with you

and a case without being at the bottom:


1. have a chatter bot post messages often
2. open the conversation in the UI
3. scroll up a few pages
4. wait for a few pages of messages to be posted
5. slowly scroll down to the read marker, see how it always moves to the last visible message

it will still need some tweaks to feel more natural

- [ ] BUG: if scrolling down too fast after seeing the read marker, it might move down a big further but moving fast makes it "not seen any more". so one would need to slowly scroll down to have the marker reset. => might be solvable by having a global "seen" property instead of a local one. The seen property would then fully clear once we reach the bottom.

PVince81:
- [ ] BUG: some read markers seem to be stuck on "0" in the database
- [ ] BUG: switching between conversations seldom displays the read marker (might be relater to the previous entry)

So far the read marker only works when staying in the same conversation.
Might need to add more logic to make "0" be equivalent to the last message.

SQL for debugging/watching: `select token, last_read_message, (select message from oc_comments c where c.id=a.last_read_message) message, last_mention_message, last_message from oc_talk_attendees a, oc_talk_rooms r where r.id=a.room_id and actor_id='admin';`

PVince81:
- [ ] BUG: clicking "scroll to bottom" while being pages above the marker might mark it as seen while quickly scrolling by, shouldn't reset it: use a flag to prevent marker processing while fast-scrolling

- [ ] TODO: add a message on top to "scroll to last read message" (will only work if the message pages are fully loaded, same issue like with search)

PVince81:
> some read markers seem to be stuck on "0" in the database

I was able to fix this by clearing it in the frontend. But it feels wrong.
The marker is likely set to 0 initially when creating a new room which has no messages. As soon as the first message appears the backend should set it to that value. It seems this is only an issue for conversation creators.

- [ ] techdebt: update last_read_message marker from 0 to last_message after posting first message to a new room

nickvergessen:
https://github.com/nextcloud/spreed/blob/9e55a41e0880219eb993c5690198f159120ccdd5/lib/Service/ParticipantService.php#L276-L279

+ there is always the message "X created the conversation". Even before the owner is added. So this should work Oo

PVince81:
> there is always the message "X created the conversation". Even before the owner is added. So this should work Oo

yeah, I saw those codes when grepping.
maybe the order is wrong and the message is posted later, I'll debug...

PVince81:
Without debugging I think I know what happens: after creating the room we trigger "EVENT_AFTER_ROOM_CREATE" which posts the message. But we're still operating on the old Room object which doesn't have the last message id.

So either we update the object within that even which is a bit dirty, or we requery the last message after triggering the event.

PVince81:
I've pushed a fix separately as the issue can be triggered with occ also: https://github.com/nextcloud/spreed/pull/5190

PVince81:
I just had a weird case where the message that was marked as unread was not even displayed in the chat, so I was never able to see the unread marker so it wouldn't clear. After refreshing the page it appears.

This sounds similar to https://github.com/nextcloud/spreed/issues/5188

PVince81:
- [ ] for the API which sets the read marker position, would be good to receive a response containing the new "unreadMessages" so that we can update the bubble in the left sidebar. Otherwise need to wait for the next poll call.

PVince81:
Based on my research in https://github.com/nextcloud/spreed/issues/5188 it seems the problem I observe is specifically affecting the message on which the current read marker is on.

I found some steps to reproduce:

1. First go on master
2. Create two conversations A and B
3. Use the bots from https://gist.github.com/PVince81/300f03d277a1eb8c876a27b5ddae932e to populate 100+ messages
4. Stop the bots
5. Take note of the last counter value in both conversations
6. Switch to this PR
7. Hack messagesService to prevent updating the read marker. This means that currently the read markers are located at the end of the conversations (due to master's auto-updating)
8. Run the bots again, let them populate 100+ more messages. This is the place where the counters restart with 1.
9. Refresh the page on conversation A, then scroll to the moment where the counters restarted with 1.
10. Now switch to conversation B, then scroll to where the counters restarted with 1.

In conversation A, you'll see that the conversation initially locates itself around the read marker, and you'll see the marker visible also. Then after loading it will automatically scroll to bottom. So far so good, no issue with conversation A.

When switching to conversation B, you'll notice that there is no read marker around the place where the counters were restarted. This is because the marked message is missing. You can confirm this by checking the last_read_marker in the database and looking for that element with `document.getElementById('message_$messageId')`.

Good, so now it should be possible to debug this further.

PVince81:
If you refresh on conversation B instead of A, then switch to A, it is A that will have the message missing.
So there's something in the logic for loading a second conversation that is buggy.

PVince81:
Okay, I'm onto something.
In the window where the "read marker" message with id 89 is missing I see two calls:
- `https://vvortex.local/ocs/v2.php/apps/spreed/api/v1/chat/4oq6c9zq?setReadMarker=0&lookIntoFuture=0&lastKnownMessageId=89&includeLastKnown=0`
- `https://vvortex.local/ocs/v2.php/apps/spreed/api/v1/chat/4oq6c9zq?setReadMarker=0&lookIntoFuture=1&lastKnownMessageId=89&includeLastKnown=0`.

None of these contain the message with id 89 because we didn't tell it to include the last known one.
I think we should at least set one of these two to include the last known message.

With a full page refresh there only one call which does include the current one:
`https://vvortex.local/ocs/v2.php/apps/spreed/api/v1/chat/4oq6c9zq?setReadMarker=0&lookIntoFuture=0&lastKnownMessageId=89&includeLastKnown=1`

Maybe the latter is about initially loading while the case where it's missing is about scrolling/pagination.

PVince81:
more clues!

it seems that in the broken case the first fetch was cancelled:
`https://vvortex.local/ocs/v2.php/apps/spreed/api/v1/chat/4oq6c9zq?setReadMarker=0&lookIntoFuture=0&lastKnownMessageId=89&includeLastKnown=1` from `fetchMessages`, but the subsequent calls still go through.

image

Maybe a bug in the cancelling. I just switched to room "4oq6c9zq" (visible with the "active" call) so it shouldn't cancel that call.

PVince81:
also looks like the two calls to "active" are out of order:
First there's a POST to /active to join 4oq6c9zq, then the cancelled call and only then the DELETE /active from the conversation I just left. I suspect that the latter code path is likely triggering the cancelling.

Not sure if this order is intended or a side effect.

PVince81:
Aha, I've observed the same order on c.nc.com.

But the call did not get cancelled.

My theory: if you're unlucky and the fetch call is too slow, then the DELETE /active might cancel it. But if you already got an answer, then the message in question has already been loaded.

Next up: digging into the code :pick:

PVince81:
doesn't always trigger, need to be "unlucky"

I managed to catch another case:
image

It appears that the second call to "getOldMessages" if it happens too early will cancel the former one, the latter being the one with "also return the last known id". It looks more and more like a conflict of "cancellables".
So either we need the second getOldMessages to wait for the first one to finish, or use two cancellable objects for the different parameters combinations. Or even an array of cancellables.

I suspect that this second call is also triggered by automatic scrolling of the viewport.

Still to find out: where the second call originates.
It doesn't seem related to the join/leave parts.

PVince81:
Okay, I think I found a way to reproduce the issue consistently:

1. Initially, following the steps from https://github.com/nextcloud/spreed/pull/3825#issuecomment-780719458
2. Patch the backend to add some delay for that request (just in case)
```diff
diff --git a/lib/Chat/ChatManager.php b/lib/Chat/ChatManager.php
index 2209dfa72..355acdc26 100644
--- a/lib/Chat/ChatManager.php
+++ b/lib/Chat/ChatManager.php
@@ -339,6 +339,9 @@ class ChatManager {
         *         timeout expired.
         */
        public function getHistory(Room $chat, int $offset, int $limit, bool $includeLastKnown): array {
+               if ($includeLastKnown) {
+                       sleep(5);
+               }
                return $this->commentsManager->getForObjectSince('chat', (string) $chat->getId(), $offset, 'desc', $limit, $includeLastKnown);
        }
```
3. After opening converstion A, don't forget to scroll down to the very bottom. This will make the viewport keep its scrolling position initially.
4. Then switch to conversation B.

I added some logging and could confirm that it's "handleScroll" that is triggering the `getOldMessages()` that will cancel the first important one.

Ok, I guess that's enough proof... Now to work on a fix.

PVince81:
I made a separate PR here: https://github.com/nextcloud/spreed/pull/5198

PVince81:
I've cherry-picked https://github.com/nextcloud/spreed/pull/5198 here because otherwise it blocks further work. Because if you can't always see the message with the read marker, the whole logic of "seen it, update when scrolling" won't work.

I already works better now but still feel clumsy.

When a conversation is active and scrolled to bottom, new messages come in. Not sure if we should consider these as read or not. Maybe only if the window is focused we could set the marker at the bottom.

PVince81:
- [ ] BUG: after seeing read marker many pages above, clicking "scroll to bottom" doesn't update the marker. Probably need to tweak the "don't update marker on automatic scroll" flag

PVince81:
- [ ] BUG: when uploading one or multiple pics, also automatically set the read marker to the last posted message (works for regular messages)

PVince81:
- [ ] scroll behavior
   - Currently it seems that the browser remembers where I was scrolled when switching between already loaded conversations, probably it remembers based on the URL
   - However when opening the conversation after a shift+refresh, the position seems more random
   - [ ] make scrolling more consistent:
      - scroll to last read message initially (when possible with the current pagination)
      - doing so will also scroll to bottom if the last read message is at the bottom already

- [ ] add button to "scroll to last read message" regardless of position (above or below the read marker when not visible) ?

PVince81:
If we want to go for a minimal solution first, I feel like while it's nice to see the marker displayed, as a user it feels wrong to not be able to scroll to it nor even being scrolled there initially. I always end up try to scroll up and down manually to see where the read marker actually is. I'd say before merging we should at least provide a way to jump there, whether it's up or down. And solve the few bugs.

@ma12-co @nickvergessen would also be nice if you could try out the current PR and give feedback on your impressions.

PVince81:
- [ ] BUG: something wrong with reactivity: despite the update of lastReadMessage, the matching Message instance's "isLastReadMessage" doesn't always evaluate to true :-(

PVince81:
- [ ] BUG: it seems there's another hidden process that sets the read marker to the bottom of the chat from times to times :thinking:

PVince81:
- [ ] optional: "menu entry: mark conversation as read" or raise ticket. After this PR marking as read is just a matter of resetting the read marker

PVince81:
I managed to improve the behavior a lot.

One issue was that lastMessage was not updated when getting new messages, so even if the marker was on whatever was the previous lastMessage, you wouldn't see it because it still believed that it was the last one... I've added store mutations to update lastMessage without a server call directly after we got new messages. (no need to wait until the next conversation store refresh)

Also, I managed to make it always scroll to the last read message first. I had to use some hacks to prevent the "scroll into view" to trigger the logic that moves the message marker, as it was not a manual scroll.

There are still a few issues now: if you stay scrolled to bottom, you'll see the read marker pop up then disappear again (unless you minimize the window). This doesn't look very nice UX-wise. I'm not sure yet if I should simply assume that if you're at the bottom, that all messages can be assumed read if `document.hasFocus()` is true...
It's actually the smoothScrollToBottom that is wrongly triggering the scroll event for resetting the marker. I'll investigate this next.

PVince81:
- [ ] BUG: scroll to last read message should show at least in the middle of the screen, but ideally 1/3 of the screen as per spec from https://github.com/nextcloud/spreed/issues/1164. Might require custom code for scroll into view or some library that does that.

- [ ] BUG: read-marker jumpy when scrolled to bottom

PVince81:
- [ ] BUG: reset indicator directly when no scrollbar present

image

PVince81:
- [ ] directly update lastMessage and lastReadMarker after posting a message (no delay waiting for the next conversation refresh)

PVince81:
- [ ] don't scroll on sticky when window is not visible, stay at the read marker
   - remove logic in smoothScrollToBottom that handles invisible window
   - check that read marker works as it should in such scenarios when new messages are coming in while invisible

PVince81:
- [ ] windowIsVisible vs hasFocus in regards to when to keep or clear the read marker
   - windowIsVisible only changes when window is completely invisible like switching tabs, but not when on another desktop window
   - hasFocus seems more reliable but requires the user to click the window from times to times to confirm they have read, otherwise the chat might be scrolling on an unseen external monitor

PVince81:
I was able to fix a few issues.

My current struggle is that I want to keep the read marker in his actual position when the window is unfocussed / invisible, or if the window is focussed always clear it to bottom. But currently there are scroll handler conflicts and it's hard to detect whether a scroll was manual or automatic (scrollIntoView has no callbacks).

PVince81:
Okay, it starts to look good.

I made a demo video for the basics here: https://cloud.nextcloud.com/s/SGGsa2XKCn8fQLr

Included in the demo:
- sticky at bottom (aka scrolledToBottom mode):
   - focussed window: no marker is shown
   - unfocussed window: marker is shown and the view does not scroll further
   - hidden tab: same as unfocussed window
- scrolls to first unread message when switching conversation (currently at bottom, needs to be fixed to be in the middle)
- marker sets itself to last visible message while scrolling down
- when scrolled up (non-sticky), the marker stays at the last read message

Not in demo:
- re-focussing on the window will remove the read marker, if currently visible in the page
- re-focussing on the window will NOT remove the read marker if there are so many new messages that the marker disappeared at the top

Please let me know what you think so far.
I'll focus on fixing the remaining issues.

PVince81:
some feedback and next steps:

- [ ] try having the read marker at the top of the window instead of bottom, except when scrolled to bottom
- [ ] jumpy marker is not nice:
    - simply clear it completely when seen ? (would get rid of the complex logic to find the next visible item to mark)
    - or keep it in place until "change of direction" ?

PVince81:
when scrolling down the marker now updates itself to the top-most visible element, it feels much better now

PVince81:
and the marker looks less jumpy now as I've increased the debounce from 1s to 3s

PVince81:
squashed all these detours into a single commit

then pushed a fix to remove the automatic "mark as read" in the store when joining.
now "mark as read" in the store only happens when clearing the read marker.
otherwise the unread number will be updated periodically when the conversation list refreshes.

src/components/MessagesList/MessagesGroup/Message/Message.vue
PVince81: it seems to work fine that way

PVince81:
here's a new video with the marker on top: https://cloud.nextcloud.com/s/B8GL4kfHPztoxG6


https://user-images.githubusercontent.com/277525/108973933-3d884200-7685-11eb-992e-6541263d0dc8.mp4

PVince81:
- [ ] to test: if read marker works properly while in a call and chat is collapsed / expanded

PVince81:
rebased to solve conflict

PVince81:
I changed the marker a bit...

image

PVince81:
> when uploading one or multiple pics, also automatically set the read marker to the last posted message (works for regular messages) 

Seems in the backend it's not possible to find out the origin of the share call, so we can't automatically set the read marker on the backend only when posted in the Talk UI.

On the frontend side, it seems that it "works like magic": when uploading an attachment, we don't actually do a "processMessage" afterwards like for regular messages. Instead, we call the share API to create it and rely on the pending "getNewMessages()" to catch it.

Now I'm not sure if we want to try and hook into that scenario, like telling "getNewMessages()" that the next incoming messages might be attachments from ours, so we might want to mark them as read automatically. But these could be intertwined with other user's messages, so it's risky. I think I'll leave this outside the scope of this PR for now, needs further research.

PVince81:
- [ ] BUG: seems the sidebar chat doesn't know when it's hidden despite focus on call view, so the logic to "stay where it is" is not working correctly (this was likely already not working before but is more apparent now)

PVince81:
Scroll to unread now works also when reopening the sidebar while in a call.

Sadly the sidebar takes 100+ ms to open itself for whatever reason so I've put a delay in there... We should investigate why the right sidebar is taking that long and is not animated like the left sidebar.

- [ ] raise ticket for right sidebar delays

PVince81:
- [ ] center indicator: "Center based on the text container width for sure (the one which also contains avatar and time stamp), that will look best :)"

src/services/messagesService.js
```diff
@@ -138,10 +140,28 @@ const postRichObjectToConversation = async function(token, { objectType, objectI
...
+ * @param {int} lastReadMessage id of the last read message to set
+ */
+const updateLastReadMessage = async function(token, lastReadMessage) {
+	try {
+		const response = await axios.post(generateOcsUrl('apps/spreed/api/v1', 2) + `chat/${token}/read`, {
+			lastReadMessage,
+		})
+		return response
+	} catch (error) {
+		console.error(`Error while updating the last read message to {lastReadMessage}`, error)
```
nickvergessen, line 156: don't catch in services anymore, the front end should do that and handle it gracefully.

src/store/messagesStore.js
```diff
@@ -321,6 +324,29 @@ const actions = {
...
+		const conversation = Object.assign({}, context.getters.conversations[token])
+		if (!conversation || conversation.lastReadMessage === id) {
+			return
+		}
+
+		// optimistic early commit to avoid indicator flickering
+		context.commit('updateConversationLastReadMessage', { token, lastReadMessage: id })
+		await updateLastReadMessage(token, id)
+
+		// TODO: update unread counters if possible (otherwise user needs to wait for conversation list refresh)
```
nickvergessen, line 348: Should the readmarker endpoint reply with that?

src/services/messagesService.js
PVince81: good point, I forgot to remove after copy-pasting. thanks for the reminder

src/components/MessagesList/MessagesList.vue
```diff
@@ -137,6 +142,24 @@ export default {
...
+		 *
+		 * @returns {object} Vue component of the message or null if not found
+		 */
+		unreadMessageComponent() {
+			let el = document.getElementById('message_' + this.conversation.lastReadMessage)
+			if (el) {
+				el = el.closest('.message')
+			}
+
+			return el?.__vue__
```
nickvergessen, line 160: This looks illegal.

src/store/messagesStore.js
PVince81: Obsolete, I'll remove. There was a bug that was hiding the fact that this already works.

src/components/MessagesList/MessagesList.vue
PVince81: Not too happy about it either.
I can work it around by storing the "seen" attribute also on the DOM to be able to access it from here.

src/components/MessagesList/MessagesList.vue
```diff
@@ -583,6 +660,121 @@ export default {
...
+		 * Find the next message element following the given message DOM element.
+		 *
+		 * This traverses the next messages and message groups to find the next one.
+		 *
+		 * @param {object} messageEl DOM element for message to start with
+		 * @returns {object} DOM element for the next message or null if none found
+		 */
+		findNextMessageElement(messageEl) {
+			// pick the previous message
+			let searchEl = messageEl.nextElementSibling
```
nickvergessen, line 677: Does this also work with the "group by author" feature we have and with system messages in between?

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -32,121 +32,129 @@ the main body of the message as well as a quote.
...
+							@click.stop="handleDelete">
+							{{ t('spreed', 'Delete') }}
+						</ActionButton>
+					</Actions>
+				</div>
+			</div>
+			<div v-if="isLastReadMessage"
+				v-observe-visibility="lastReadMessageVisibilityChanged">
+				<div v-if="!isLastMessage" class="new-message-marker">
+					<span>----- {{ t('spreed', 'Unread messages') }} -----</span>
```
nickvergessen, line 156: let's use the old date time styling instead of ascii, see removed code in https://github.com/nextcloud/spreed/pull/4582/files

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -744,6 +769,18 @@ export default {
 	100% { background-color: rgba(var(--color-background-hover), 0); }
 }
 
+.new-message-marker {
+	margin: 20px 0;
+	text-align: center;
+
+	span {
+		background-color: var(--color-primary-light);
+		color: var(--color-text-lighter);
```
nickvergessen, line 778: I think this is too little contrast depending on the colors, cc @jancborchardt for the math
Bildschirmfoto von 2021-03-01 11-05-28

nickvergessen:
So when I open the chat in 2 containered tabs with user 1 and user 2 and write and then switch to the other tab, I see a flicker of the unread marker and it immediately disappears again. this feels weird, especially as it makes things "jump" up.

So I wonder if:
1. the marker should be floating on top of the message instead of being inserted moving everything up/down
2. the marker should have a minimum display time somehow?

src/components/MessagesList/MessagesList.vue
PVince81: it works with the message groups, if you read further below you can see how it's exiting the current group and entering the next one.

for system messages I'll double check

PVince81:
@nickvergessen did you test with scrolling ?

If you don't have a scrollbar, the read marker will be discarded as soon as you focus the window. Because there's no way to scroll down to make it disappear (special handling). Maybe adding a delay could help here.

If there's a scrollbar, then you can scroll down and the marker follows. And once you reach the bottom it gets cleared.

PVince81:
I've now added a delay (max 3 seconds) by calling `debounceUpdateReadMarkerAfterScroll` in `focusMessage` in case no scroll is possible.

src/components/MessagesList/MessagesGroup/Message/Message.vue
PVince81: it would look something like this:
image

and would obsolete your comment about the background color below

but if we make it "floating" instead of "in-layout" we probably can't use the non-bubble style as it wouldn't overlap well on top of other messages

src/components/MessagesList/MessagesGroup/Message/Message.vue
PVince81: obsoleted by the previous request to have the bar look like the old dates:
image

src/components/MessagesList/MessagesList.vue
PVince81: I did a test with a lot of system messages, then scrolled down and waited 3s to have the marker update, then scrolled up again:
image

this means the detection within those groups also work.
they actually use the same classes "message-group" and "message", so this was expected

PVince81:
- [ ] look into making the marker floating

because the current one might overlap:
image

src/components/MessagesList/MessagesList.vue
PVince81: removed the aberration and replaced with DOM attributes "data-message-id" and "data-seen"

PVince81:
I can't think of any good design where the marker would be floating. This means overlapping on top of older messages and it wouldn't look good. The closest I thought of is having a wide bar with dark background with transparency, and the text on top. But even that didn't look well.

As an alternative, I tried to compensate for the jumpiness of the marker when it moves, but it seems that the jump size is not predictable. It's definitely not equal to the height of the marker. I'd expect the browser to not scroll the viewport when removing then insert an element with the exact same height.

src/components/MessagesList/MessagesGroup/Message/Message.vue
nickvergessen: Well not requests but thoughts, don't know what's a good idea

PVince81:
@ma12-co would be good if you could check the current design of the marker in this PR, and whether you have ideas if there's a way to make it a floating bubble without looking weird when overlapping messages.

PVince81:
I've now added "mark as read" in the left sidebar menu and "mark as unread" on the message context menu.

This has highlighted further issues:
- [ ] bug: mark as unread would need to move the marker to the **previous** message.
   - move the marker **before** the message ?
   - the problem might change or disappear if we go with a floating indicator
- [ ] bug: moving the marker with "mark as unread" sometimes causes micro-scrolling which triggers the scroll handler, which in turn removes the read marker... need a way to the scroll handler to ignore that one scroll

PVince81:
Tested with IE Edge and the marker scrolling logic works the same.

ma12-co:
So from my standpoint we shouldn't hide/move the marker while in the conversation. What I would do is: 
- If the tab is active and you're in a conversation => Marker stays put forever no matter what;
- if you switch to another conversation and comeback => Marker jumps to last read message;
- if you switch tab and comeback => Marker jumps to last read message;

PVince81:
> So from my standpoint we shouldn't hide/move the marker while in the conversation. What I would do is:
> 
>     * If the tab is active and you're in a conversation => Marker stays put forever no matter what;
> 
>     * if you switch to another conversation and comeback => Marker jumps to last read message;
> 
>     * if you switch tab and comeback => Marker jumps to last read message;

Seems to make sense. So instead of calling "processReadMarkerAfterScroll" after a scroll event, it would only be called when the window/tab is blurred / becoming invisible, and when just before switching away to another conversation.

PVince81:
@nickvergessen do you agree with the above proposed change ? I'm asking because the original ticket https://github.com/nextcloud/spreed/issues/1164 only mentions moving the marker after scroll

nickvergessen:
> Seems to make sense. So instead of calling "processReadMarkerAfterScroll" after a scroll event, it would only be called when the window/tab is blurred / becoming invisible, and when just before switching away to another conversation.

So from my pov those things can (and should?) Be disconnected. Setting the readmarker on the api (do the status is reflected on other devices) should happen as in this PR. I would basically just change that the readmarker in the UI is only read from the store on those named events (visibility change/leave/join). This would let the state be stable in the UI (no jump/flash on new messages, etc) and still up to date info on other devices when they get the room list

PVince81:
@nickvergessen so if I understand correctly:

- when scrolling down, we do dispatch the action that saves/update the read marker to the server, but we don't update its value in the store (no commit)
- whenever the conversation/window/tab is focussed again, we do an explicit fetch for the current conversation to retrieve the lastReadMessage from the server, which might either be the one from the former bullet point or from the clients.

However, please note that we already have periodic reloading of all conversations as part of the left sidebar (not sure if in place for guests though). So the read marker will jump at the time where this load happens. I believe it's every 30 seconds.

If that's ok with everyone I'll implement it that way.

nickvergessen:
No, that's not what i ment. Let's talk about it tomorrow in speech as it is easier to bring the thoughts around in that

PVince81:
Summary of discussion

- Visually the read marker has its own position, independent from what's in the store
   - [ ] The visual read marker position is stored in a separate attribute "visualLastReadMessageId" (working title)
   - [ ] The separate attribute "visualLastReadMessageId" is only updated whenever the window/tab is focused again. When it does, the value from the Vuex store is copied into it.
   - This will make sure the user never sees the marker jumping directly (it jumps only when you're not looking)
- Clicking on "mark as read"
   - [ ] Set and save lastReadMessage to the **previous** message, 
   - [ ] Tweak wording to make it clearer that it's "unread starting from here"

- [ ] move marker outside the hoverable message div
- [ ] tweak message marker style

PVince81:
Okay, I'm done.

I've reworked the code to be able to find any previous or next message based on the approach I used for the keyboard nav WIP: https://github.com/nextcloud/spreed/pull/4666. Basically, the Message component now is aware of the previous and next message ids.

This means that when applying as unread we can easily do so on the previous message's id.

And the code that looks for the next visible element uses the next message ids attribute for iterating instead of the more complex group/message DOM element iterations.

The remaining tasks are about design, maybe @ma12-co can jump in for the wording and the appearance of the marker ?

nickvergessen:
> Conflicting files
src/components/MessagesList/MessagesGroup/Message/Message.vue

nickvergessen:
It's a bit weird to have the unread marker start showing for own items. Basically when you have read everything and reload, it shows on the next one instead of not at all:
Peek 2021-03-08 13-41

PVince81:
> It's a bit weird to have the unread marker start showing for own items. Basically when you have read everything and reload, it shows on the next one instead of not at all:
> Peek 2021-03-08 13-41

okay, it seems I forgot to also force-update the marker when posting...

PVince81:
Fixed case where posting new messages: the read marker is now updated.

Also fixed issue with marker update when being scrolled to bottom, blur then focus again.

ma12-co (approved):
:ok_hand:

PR #3827, opened 2020-06-19
Move to event listeners

nickvergessen:


lib/Listener/BeforeUserLoggedOutListener.php
```diff
@@ -0,0 +1,70 @@
...
+
+class BeforeUserLoggedOutListener implements IEventListener {
+
+	/** @var Manager */
+	private $manager;
+	/** @var TalkSession */
+	private $talkSession;
+
+	public function __construct(Manager $manager,
+TalkSession $talkSession) {
```
ChristophWurst, line 43: indentation

nickvergessen: Oo how did that happen

PR #3872, opened 2020-07-03
Document system requirements

### LINKED ISSUE #3139: Document system requirements
Nextcloud Talk currently has various undocumented system requirements, prerequisites and limitations that should be documented.

For instance, you can't run Nextcloud Talk on a Apache webserver with both HTTP/2 and the Apache `mpm_prefork` module - it will slow down Nextcloud Talk beyond usability. See #602 and #2211. You must use Apache `mpm_event` instead.

Furthermore Nextcloud Talk has some significant system requirements for both the server and client. Servers will face additional load - users on a dedicated server should just be aware of this fact, but more importantly, users of shared webhosters might do something illegal by installing Nextcloud Talk.

Clients will experience a massive resource usage when connecting multiple users at a time - affecting both CPU and RAM usage, as well as bandwidth usage. This effectively limits the number of concurrent users to about 5 people, depending on the client's abilities. That's due to technical limitations of WebRTC, but Nextcloud Talk should document this nevertheless.

The fact that you might need a TURN server is at least mentioned in the app's `README.md`, but without explaining what the TURN server is actually for and under which circumstances one is required. Neither STUN nor signalling servers are even mentioned nor explained what they are for.

Nextcloud Talk kinda is on a "either it works on the first try or simply forget about it" state regarding its docs right now... Very frustrating if it doesn't work on the first try. :unamused:

nickvergessen:
Fix #3139

### DIFF
--- a/docs/system-requirements.md
+++ b/docs/system-requirements.md
@@ -1,0 +1,46 @@
+# Server
+
+Nextcloud Talk will significantly increase load on your server, depending on the number of concurrent active participants - no matter whether they use audio call, video call or just the chat. This is due to required signalling and message pulling. By default Talk will use its integrated signalling server, which works just fine for small and medium installations. If you're running Talk on a dedicated or decent virtual server, you shouldn't experience any problems. However, if you're running Talk on a shared webhoster, you might indeed experience issues. Also note that many shared webhosters particularly prohibit running chat server software like Nextcloud Talk; please make sure to check your provider's Terms & Conditions. In this case you might be required to upgrade your package.
+
+Additional to normal Nextcloud requirements the following constraints apply to use Nextcloud Talk:
+
+## HTTPS
+ 
+HTTPS is required to be able to use WebRTC (the video call technology of browsers used by Nextcloud Talk calls).
+
+## Database
+
+* SQLite: must not be used, to grant a decent experience for chats and calls
+* MySQL/Maria DB: Must enable utf8mb4 support as per documentation at [Enabling MySQL 4-byte support](https://docs.nextcloud.com/server/latest/admin_manual/configuration_database/mysql_4byte_support.html)
+
+## Webserver
+
+Apache and Nginx must use:
+
+* PHP FPM + mpm_events or
+* PHP + mpm_prefork
+
+Other combinations will not work due to the long polling used for chat and signaling messages, see [this issue](https://github.com/nextcloud/spreed/issues/2211#issuecomment-610198026) for details.
+
+## TURN server
+
+A TURN server running on **port 443** (or 80) is required in almost all scenarios, see  [Configuring coTURN](TURN.md) for more details.
+
+# Browsers
+
+## Recommended
+
+* Firefox: latest
+* Chrome/Chromium: latest
+
+## Supported
+
+* Firefox / Firefox ESR: 52 or later
+* Chrome / Chromium: 49 or later
+* Edge: latest
+* Safari: 13 or later
+
+# Mobile apps
+
+* Android: 5 or later
+* iOS: 10 or later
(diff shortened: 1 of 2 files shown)


docs/system-requirements.md
```diff
@@ -0,0 +1,44 @@
+# System requirements
+
+## Server
+Additional to normal Nextcloud requirements the following constraints apply to use Nextcloud Talk:
```
PhrozenByte, line 4: We should talk about increased server load due to signalling. For example, a few weeks ago I failed using Talk on a decent shared webhosting package ([all-inkl.com Premium](https://all-inkl.com/webhosting/premium/)) with about 25 active users. Impossible to use - and I'm talking about just Chat here.

```suggestion
Nextcloud Talk will significantly increase load on your server, depending on the number of concurrent active conversations - no matter whether they use Audio, Video or just Chat. This is due to required signalling. By default Talk will use its integrated signalling server, which works just fine for small and medium installations. If you're running Talk on a dedicated or decent virtual server, you shouldn't experience any problems. However, if you're running Talk on a shared webhoster, you might indeed experience issues. Please also note that many shared webhosters particularly prohibit running chat server software like Talk; please make sure to check your providers's Terms & Conditions. In this case you might be required to upgrade to either a dedicated server or use Talk's [High-performance Back-end (HPB)](https://www.spreed.eu/contact-nextcloud-talk-high-performance-backend/) for signalling.

Additional to normal Nextcloud requirements the following constraints apply to use Nextcloud Talk:
```

docs/system-requirements.md
```diff
@@ -0,0 +1,44 @@
...
+
+### HTTPS
+ 
+HTTPS is required to be able to use WebRTC (the video call technic of browsers used by Nextcloud Talk calls).
+
+### Database
```
PhrozenByte, line 10:
```suggestion
### Database

```

docs/system-requirements.md
```diff
@@ -0,0 +1,44 @@
...
+* SQLite: must not be used, to grant a decent experience for chats and calls
+* MySQL/Maria DB: Must enable utf8mb4 support** as per documentation at [Enabling MySQL 4-byte support](https://docs.nextcloud.com/server/latest/admin_manual/configuration_database/mysql_4byte_support.html)
```
PhrozenByte, line 12: What are the two `**` behind "utf8mb support" for?

docs/system-requirements.md
```diff
@@ -0,0 +1,44 @@
...
+Apache and Nginx must use:
+
+* PHP FPM + mpm_events or
+* PHP + mpm_prefork
+
+Other combinations will not work due to the long polling used for chat and signaling messages, see [this issue](https://github.com/nextcloud/spreed/issues/2211#issuecomment-610198026) for details.
+
+### TURN server
+
+A TURN server running on **port 443** (or 80) is required in almost all scenarios, see  [Configuring coTURN](TURN.md) for more details.
```
PhrozenByte, line 25: This reads like one always needs a TURN server. In my experience a TURN server isn't always required.

docs/system-requirements.md
nickvergessen: I reformated it later, before I used `**` to make it bold

docs/system-requirements.md
nickvergessen: Well as soon as a mobile phone is involved you need one 99% of the time. If people want to ignore it fine by me, but at least we told them up front. The UI will tell it again later when you try a call and a connection can not be established.

docs/system-requirements.md
nickvergessen: I modified it a bit to be less "aggressive"

PR #3885, opened 2020-07-06
Correctly initialise blur in chrome/chromium

nickvergessen:
Before bluring of local avatars is missing in Chrome/Chromium after toggling between grid view and promoted view.

### DIFF
--- a/src/components/CallView/shared/VideoBackground.vue
+++ b/src/components/CallView/shared/VideoBackground.vue
@@ -21,7 +21,9 @@
 
 <template>
 	<div class="video-backgroundbackground">
-		<div class="darken">
+		<div
+			ref="darkener"
+			class="darken">
 			<ResizeObserver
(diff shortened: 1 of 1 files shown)


src/components/CallView/shared/VideoBackground.vue
```diff
@@ -21,7 +21,8 @@
 
 <template>
 	<div class="video-backgroundbackground">
-		<div class="darken">
+		<div class="darken"
+			ref="placeholder">
```
ma12-co, line 25: can we improve the name? maybe "darkener"?

nickvergessen: darkener is not useful at all, as the thing isn't actually darker. Maybe `backgroundAvatar` ?

ma12-co: It should darken, that should be the only purpose of that div

PR #3890, opened 2020-07-10
🍱 Dashboard adaption for Nextcloud 20

juliushaertl:
For https://github.com/nextcloud/server/pull/21346

This implements a basic dashboard panel for Talk that lists the recent conversations. 

- [x] Create conversation button in empty content view is non-functional
- [ ] ~~Figure out if we can join a call directly by clicking on the start call button~~
- [x] Show a loading placeholders
- [x] Properly style conversation list to match the talk sidebar

image

cc @nickvergessen @jancborchardt @ma12-co

### DIFF
--- a/lib/Dashboard/TalkPanel.php
+++ b/lib/Dashboard/TalkPanel.php
@@ -1,0 +1,70 @@
+<?php
+/**
+ * @copyright Copyright (c) 2020 Julius Härtl <jus@bitgrid.net>
+ *
+ * @author Julius Härtl <jus@bitgrid.net>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Talk\Dashboard;
+
+
+class TalkPanel implements \OCP\Dashboard\IPanel {
+
+	/**
+	 * @inheritDoc
+	 */
+	public function getId(): string {
+		return 'spreed';
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function getTitle(): string {
+		return 'Talk';
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function getOrder(): int {
+		return 10;
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function getIconClass(): string {
+		return 'icon-talk';
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function getUrl(): ?string {
+		return \OC::$server->getURLGenerator()->getAbsoluteURL('/apps/spreed');
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function load(): void {
+		\OC_Util::addScript('spreed', 'dashboard');
+	}
+}
--- a/lib/Dashboard/TalkPanel.php
+++ b/lib/Dashboard/TalkPanel.php
@@ -23,8 +23,18 @@
 
 namespace OCA\Talk\Dashboard;
 
+use OCP\IL10N;
 
 class TalkPanel implements \OCP\Dashboard\IPanel {
+
+	/** @var IL10N */
+	private $l10n;
+
+	public function __construct(
+		IL10N $l10n
+	) {
+		$this->l10n = $l10n;
+	}
 
 	/**
 	 * @inheritDoc
@@ -37,7 +47,7 @@ class TalkPanel implements \OCP\Dashboard\IPanel {
 	 * @inheritDoc
 	 */
 	public function getTitle(): string {
-		return 'Talk';
+		return $this->l10n->t('Talk mentions');
 	}
 
 	/**
--- a/lib/Dashboard/TalkPanel.php
+++ b/lib/Dashboard/TalkPanel.php
@@ -24,6 +24,7 @@
 namespace OCA\Talk\Dashboard;
 
 use OCP\IL10N;
+use OCP\Util;
 
 class TalkPanel implements \OCP\Dashboard\IPanel {
 
@@ -75,6 +76,6 @@ class TalkPanel implements \OCP\Dashboard\IPanel {
 	 * @inheritDoc
 	 */
 	public function load(): void {
-		\OC_Util::addScript('spreed', 'dashboard');
+		Util::addScript('spreed', 'dashboard');
 	}
 }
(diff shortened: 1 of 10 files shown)


juliushaertl:
@nickvergessen Is there a way to limit the results on https://nextcloud-talk.readthedocs.io/en/latest/conversation/#get-user-s-conversations in the request already? Any suggestion how to handle updates better than refetching every x seconds there?

lib/Dashboard/TalkPanel.php
```diff
@@ -0,0 +1,70 @@
...
+	 */
+	public function getUrl(): ?string {
+		return \OC::$server->getURLGenerator()->getAbsoluteURL('/apps/spreed');
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function load(): void {
+		\OC_Util::addScript('spreed', 'dashboard');
```
nickvergessen, line 68:
```suggestion
		\OCP\Util::addScript('spreed', 'dashboard');
```

src/views/Dashboard.vue
```diff
@@ -0,0 +1,115 @@
...
+	components: { ConversationIcon, EmptyContent },
+	data() {
+		return {
+			roomOptions: [],
+		}
+	},
+	computed: {
+		callLink() {
+			return (conversation) => {
+				return '/index.php/call/' + conversation.token
```
nickvergessen, line 67:
```suggestion
				return generateUrl('/call/' + conversation.token)
```

src/views/Dashboard.vue
```diff
@@ -0,0 +1,115 @@
...
+			}
+		},
+	},
+	beforeMount() {
+		this.fetchRooms()
+		setInterval(() => this.fetchRooms(), 5000)
```
nickvergessen, line 73: This needs to be slower to not bring down the servers for unnecessary idling people.

nickvergessen:
> Is there a way to limit the results on https://nextcloud-talk.readthedocs.io/en/latest/conversation/#get-user-s-conversations in the request already?

No, and actually the results are not sorted at the moment on the server side. What would be a desired value, then we can add it. I would just like to not see this abused in other clients then by falsly assuming things on the limit and then asking for an offset etc.

> Any suggestion how to handle updates better than refetching every x seconds there?

Not really, as I don't want to get signaling involved on the dashboard page. Also I think 5 second is way too frequent.

src/views/Dashboard.vue
```diff
@@ -0,0 +1,140 @@
...
+
+		&:hover {
+			background-color: var(--color-background-hover);
+			border-radius: var(--border-radius-large);
+		}
+
+		.conversation-icon {
+			position: relative;
+
+			// Do not show favorite or call icons
```
nickvergessen, line 101: No, instead don't set hide-call-icon and hide-favorite-icon to false in the ConversationIcon component

lib/Dashboard/TalkPanel.php
```diff
@@ -0,0 +1,80 @@
...
+	 */
+	public function getId(): string {
+		return 'spreed';
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function getTitle(): string {
+		return $this->l10n->t('Talk mentions');
```
nickvergessen, line 50: @jancborchardt this is not what it does atm. It just shows the recent conversations, its not filtering by mentions (which will also be quite empty most of the times (unless the start of the day e.g.)

jancborchardt: Right, of course. @juliushaertl @nickvergessen is it planned to do the filtering in this PR or in an upcoming one? If in an upcoming one, we can say "Talk conversations" here.

nickvergessen: like mentioned, I'm not even sure we should filter this way. It will be mostly empty all the time. I think recent conversations is much more helpful then only mentions.

jancborchardt: Well, _if_ there are mentions, they should for sure be prioritized though.

We have the same issue in the Talk app itself, and in a Dashboard we really need to be concise. :)

ma12-co:
@juliushaertl could we have a "create new conversation" always and not only in the empty content? That way we could have a quick way to create a meeting like Jitzi and Zoom.
This could be a plus button similar to what we have in talk's conversations list:

Screenshot from 2020-08-03 10-51-56

Also, since each app has this "+" action, every dashboard list could have this, maybe it could appear when hovering the list? cc @jancborchardt

jancborchardt:
@ma12-co yup, exactly that is defined in the spec at https://github.com/nextcloud/server/issues/20930

lib/Dashboard/TalkWidget.php
```diff
@@ -0,0 +1,81 @@
...
+	 */
+	public function getIconClass(): string {
+		return 'icon-talk';
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function getUrl(): ?string {
+		return \OC::$server->getURLGenerator()->getAbsoluteURL('/apps/spreed');
```
nickvergessen, line 72: Nop

lib/Dashboard/TalkWidget.php
```diff
@@ -0,0 +1,81 @@
...
+	 */
+	public function getId(): string {
+		return 'spreed';
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function getTitle(): string {
+		return $this->l10n->t('Talk mentions');
```
nickvergessen, line 51: Is it a mention widget now or talk conversations?
I see you sort mentions and calls on top, but the caption seems wrong to me

src/views/Dashboard.vue
```diff
@@ -0,0 +1,141 @@
...
+					<ConversationIcon
+						:item="item"
+						:hide-favorite="true"
+						:hide-call="true" />
+				</template>
+			</DashboardWidgetItem>
+		</template>
+		<template v-slot:empty-content>
+			<EmptyContent icon="icon-talk">
+				{{ t('spreed', 'Join a conversation or start a new one') }}
```
nickvergessen, line 42: You can't really join when you don't have any. So that needs a bit adjusting here

src/views/Dashboard.vue
```diff
@@ -0,0 +1,141 @@
...
+				this.loading = false
+			})
+		},
+	},
+}
+</script>
+
+<style lang="scss" scoped>
+	.empty-content {
+		text-align: center;
```
nickvergessen, line 139: Should this be part of the default component?

src/views/Dashboard.vue
```diff
@@ -0,0 +1,141 @@
...
+	beforeMount() {
+		this.fetchRooms()
+		// FIXME: reduce interval if user not active
+		setInterval(() => this.fetchRooms(), ROOM_POLLING_INTERVAL * 1000)
+	},
+	methods: {
+		fetchRooms() {
+			axios.get(generateOcsUrl('/apps/spreed/api/v1', 2) + 'room').then((response) => {
+				const rooms = response.data.ocs.data
+				rooms.sort(propertySort(['-hasCall', '-unreadMention', '-lastActivity']))
```
nickvergessen, line 128: So unread mentions are sorted top, but there is no visual indicator, which might be very confusing?

src/views/Dashboard.vue
```diff
@@ -0,0 +1,141 @@
...
+		:loading="loading"
+		@hide="() => {}"
+		@markDone="() => {}">
+		<template v-slot:default="{ item }">
+			<DashboardWidgetItem :item="getWidgetItem(item)">
+				<template v-slot:avatar>
+					<ConversationIcon
+						:item="item"
+						:hide-favorite="true"
+						:hide-call="true" />
```
nickvergessen, line 36: Since you sort by it, i would show the call icon too

nickvergessen:
Bildschirmfoto von 2020-08-17 16-03-57

Updated as discussed with @jancborchardt and @juliushaertl to show unread mentions and calls only and otherwise an empty content with the last 4 conversations.

PR #3910, opened 2020-07-16
Presort the mention suggestions to fix postgres integration tests

nickvergessen:
* See https://drone.nextcloud.com/nextcloud/spreed/8031/20/4
* See https://drone.nextcloud.com/nextcloud/spreed/8031/22/4


nickvergessen:
Well we are not sorting them by ID, you just expect that the input order is the output order. which is true for everything apart from postgres. there items can be sorted different to autoincrement due to temporary journals.
Also I'm somewhat unwilling to add an order by to make the queries a bit heavier just for the sake of it. Can also add a sleep(1) after creating the conversations and participants and ... But I don't think it is all worth it.

PR #3931, opened 2020-07-21
Add scroll to bottom button

ma12-co:
- [x] Add the button and make it scroll down;
- [x] Fix delay in appearing/disappearing of the button;
- [x] Fix transition (fade in and out of the button);
- [x] Animate scroll;

Peek 2020-07-21 21-19

Also adds smooth scroll to the whole messageslist

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -53,6 +53,7 @@ get the messagesList array and loop through the list to generate the messages.
 		</template>
 		<transition name="fade">
 			<button v-show="!isScrolledToBottom"
+				:aria-label="scrollTobottomAriaLabel"
 				class="scroll-to-bottom"
 				@click="scrollToBottom">
 				<ChevronDown decorative
@@ -210,6 +211,10 @@ export default {
 
 		chatIdentifier() {
 			return this.token + ':' + this.isParticipant + ':' + this.isInLobby
+		},
+
+		scrollToBottomAriaLabel() {
+			return t('spreed', 'Scroll to bottom')
 		},
 	},
 
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -51,6 +51,13 @@ get the messagesList array and loop through the list to generate the messages.
 				type="messages"
 				:count="15" />
 		</template>
+		<transition name="fade">
+			<button v-show="!isScrolledToBottom"
+				class="scroll-to-bottom"
+				@click="scrollToBottom">
+				<ChevronDown :size="24" fill-color="#fff" />
+			</button>
+		</transition>
 	</div>
 </template>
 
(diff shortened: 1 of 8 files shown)


src/components/MessagesList/MessagesList.vue
```diff
@@ -50,6 +50,13 @@ get the messagesList array and loop through the list to generate the messages.
 				v-for="n in 15"
 				:key="n" />
 		</template>
+		<transition name="fade">
+			<button v-show="!isScrolledToBottom"
+				class="scroll-to-bottom"
+				@click="scrollToBottom">
+				<ChevronDown :size="24" fill-color="#fff" />
```
nickvergessen, line 57: only works when the color on primary is "white"

src/components/MessagesList/MessagesList.vue
```diff
@@ -580,6 +590,8 @@ export default {
 </script>
 
 <style lang="scss" scoped>
+@import '../../assets/variables.scss';
```
nickvergessen, line 593: needed?

src/components/MessagesList/MessagesList.vue
ma12-co: We're going to have a design hack Tuesday next week and by the end of that we'll have a way to use material design icons with theming

src/components/MessagesList/MessagesList.vue
ma12-co: Yes, for the fade transition

src/components/MessagesList/MessagesList.vue
nickvergessen: `var(--color-primary-text)` it is !

nickvergessen:
Fixed the color thing and a "random" bug with the button not disappearing correctly.

src/components/MessagesList/MessagesList.vue
```diff
@@ -51,6 +51,13 @@ get the messagesList array and loop through the list to generate the messages.
 				type="messages"
 				:count="15" />
 		</template>
+		<transition name="fade">
+			<button v-show="!isScrolledToBottom"
+				class="scroll-to-bottom"
+				@click="scrollToBottom">
+				<ChevronDown :size="24" fill-color="#fff" />
```
nickvergessen, line 58: Tooltip says `Chevron Down icon`.
We have the same issue in quite some other places (overflowing video list is `Chevron Right icon` etc.
I guess we need to set the title ourselves

src/components/MessagesList/MessagesList.vue
```diff
@@ -51,6 +51,13 @@ get the messagesList array and loop through the list to generate the messages.
 				type="messages"
 				:count="15" />
 		</template>
+		<transition name="fade">
+			<button v-show="!isScrolledToBottom"
+				class="scroll-to-bottom"
+				@click="scrollToBottom">
```
nickvergessen, line 57: aria?

nickvergessen (approved):

PR #3934, opened 2020-07-23
Add shortcut to makefile

ma12-co:
`make d` 
d as 'develop'

Quick shortcut to build and watch for changes. Handy when switching branches 


ma12-co:
:cry:

PR #3948, opened 2020-07-27
Handle referenceId for system messages e.g. when uploading a file

nickvergessen:


nickvergessen:
Failing unit tests are fixed with https://github.com/nextcloud/server/pull/22027

nickvergessen:
> One little thing that I have noticed: if you upload a large file and send new messages while the file is being uploaded the file will appear after those messages. Honestly I do not know if a user would expect the file to appear in the message list where it started to upload or where it ended to upload.

Hmm yeah well it's the same for when your sending a message takes longer the order could be different if you dont wait for the finishing.

PR #3953, opened 2020-07-29
Always allow local HPBs

nickvergessen:


nickvergessen:
> As this removes the need to set allow_local_remote_servers system wide when using a local HPB …

Exactly that was the idea ;)

PR #3962, opened 2020-07-31
Only show browser warning when interacting with talk

### LINKED ISSUE #3844: "The browser you're using is not fully supported" for GNOME Web (Epiphany) 3.36.1
GNOME Web uses the WebKitGTK rendering engine, and as such is at least on par with Safari, if not more up to date. Yet every time I log in with Web, I am presented with the "The browser you're using is not fully supported" nag, which tells me to go install Safari, amongst others.

### Steps to reproduce
1. Install GNOME Web 3.36
2. Log in to Next Cloud account

### Expected behaviour

No upgrade nag is presented.

### Actual behaviour

An upgrade nag is presented.

### Server configuration

**Operating system:** Debian 10

**Web server:** Apache 2.4.38

**Database:** Maria 10.3.22

**PHP version:** 7.3

**Nextcloud version:** 19.0.0

**Updated from an older Nextcloud/ownCloud or fresh install:** Fresh

**Where did you install Nextcloud from:** Downloaded tar.gz

nickvergessen:
Fix #3844

### DIFF
--- a/src/FilesSidebarCallViewApp.vue
+++ b/src/FilesSidebarCallViewApp.vue
@@ -34,7 +34,6 @@
 import { PARTICIPANT } from './constants'
 import CallView from './components/CallView/CallView'
 import PreventUnload from 'vue-prevent-unload'
-import browserCheck from './mixins/browserCheck'
 import duplicateSessionHandler from './mixins/duplicateSessionHandler'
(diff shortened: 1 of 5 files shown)


nickvergessen:
Outside of Talk itself, we should only prompt the warning when the user clicks the buttons to interact with Talk.

PR #3963, opened 2020-07-31
Fix Internet Explorer 11 not loading because Element.append() does no…

nickvergessen:
…t exist

Before main Talk worked, but all sidebar integrations still failed


nickvergessen:
Right, that also works :+1:

PR #3988, opened 2020-08-03
Feature/3949/allow editing before upload

ma12-co:
Follow-up:
- [ ] Prevent duplicate uploads
- [ ]  Upload indicator should be overlaid on image, and just vanish when it’s uploaded
- [ ]  Re-drop files on modal

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
@@ -23,8 +23,8 @@
 
 <template>
 	<file-preview v-bind="filePreview"
-		class="container"
-		:class="{ 'is-viewer-available': isViewerAvailable, 'upload-editor': isUploadEditor }"
+		class="file-preview"
+		:class="{ 'file-preview--viewer-available': isViewerAvailable, 'file-preview--upload-editor': isUploadEditor }"
 		@click="showPreview">
 		<img v-if="!isLoading && !failed"
 			:class="previewSizeClass"
@@ -37,6 +37,10 @@
 		<span v-if="isLoading"
 			class="preview loading" />
 		<strong>{{ name }}</strong>
+		<button v-if="isUploadEditor"
+			class="remove-file primary">
+			<Close class="remove-file__icon" @click="$emit('remove-file', id)" />
+		</button>
 		<ProgressBar v-if="isTemporaryUpload && !isUploadEditor" :value="uploadProgress" />
 	</file-preview>
 </template>
@@ -44,12 +48,14 @@
 <script>
 import { generateUrl, imagePath } from '@nextcloud/router'
 import ProgressBar from '@nextcloud/vue/dist/Components/ProgressBar'
+import Close from 'vue-material-design-icons/Close'
 
 export default {
 	name: 'FilePreview',
 
 	components: {
 		ProgressBar,
+		Close,
 	},
 
 	props: {
@@ -227,8 +233,10 @@ export default {
 </script>
 
 <style lang="scss" scoped>
+@import '../../../../../assets/variables.scss';
 
-.container {
+.file-preview {
+	position: relative;
 	width: 100%;
 	/* The file preview can not be a block; otherwise it would fill the whole
 	width of the container and the loading icon would not be centered on the
@@ -243,6 +251,9 @@ export default {
 		/* Trick to keep the same position while adding a padding to show
 			* the background. */
 		box-sizing: content-box !important;
+		.remove-file {
+			visibility: visible;
+		}
 	}
 
 	.preview {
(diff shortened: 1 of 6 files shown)


src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
```diff
@@ -213,20 +235,14 @@ export default {
...
-			box-sizing: content-box !important;
-			padding: 10px;
-			margin: -10px;
-		}
+	&:hover,
+	&:focus {
+		background-color: var(--color-background-hover);
+		/* Trick to keep the same position while adding a padding to show
+			* the background. */
+		box-sizing: content-box !important;
```
danxuliu: While this looks good in the upload editor in my opinion the look in the messages list has regressed. Therefore I would suggest to have both the previous style and the new one, and use one or the other depending on whether the `upload-editor` class is set or not.

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
```diff
@@ -26,7 +26,7 @@
 		class="file-preview"
 		:class="{ 'file-preview--viewer-available': isViewerAvailable, 'file-preview--upload-editor': isUploadEditor }"
 		@click="showPreview">
-		<img v-if="!isLoading && !failed"
+		<img v-if="(!isLoading && !failed) || hasTemporaryImageUrl"
```
danxuliu, line 29: Should the preview be shown when both `hasTemporaryImageUrl` and `localUrl` are true (as `previewUrl` checks both)?

src/components/UploadEditor.vue
```diff
@@ -29,30 +29,28 @@
 			type="file"
 			class="hidden-visually"
 			@change="handleFileInput">
-		<div class="upload-editor">
```
danxuliu: It feels strange to have `upload-editor__previews` and `upload-editor__actions` without a `upload-editor` parent element. Would it be possible to keep it instead?

src/components/UploadEditor.vue
```diff
@@ -143,33 +141,43 @@ export default {
...
+		}
 	}
 	&__add-more {
-		width: 100px;
-		height: 100px;
+		width: 80px;
+		height: 80px;
 		border: none;
-		border-radius: var(--border-radius-large);
+		border-radius: 40px;
```
danxuliu, line 164: I think that the radius variable for full rounded borders is `var(--border-radius-pill)`.

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
ma12-co: What do you think has regressed in particular? Before there was an unneeded hover feedback on the preview itself that partially covered the sender's username.

src/components/UploadEditor.vue
ma12-co: That div messed things up a lot. Adding the upload-editor class to the outer modal.

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
danxuliu: > there was an unneeded hover feedback on the preview itself

I think that the hover feedback should be there, as it shows the user that she can interact with the file. The problem is that the whole message is "highlighted" on hover when showing the reply button, and that hides the specific feedback for the file. Anyway, I would be fine with removing the hover feedback on the file due to that overlap (but I would prefer to find a better solution nevertheless).

> that partially covered the sender's username

[I know](https://github.com/nextcloud/spreed/pull/2778#issuecomment-598234022), it has been haunting me since I added it :-P

However, even if it covers the sender's username I prefer that small, focused frame on the file:
Message-File-Focus-Before

Rather than highlighting the whole message body (even if it actually matches the real clickable area):
Message-File-Focus-After

For me the previous version feels and looks much better when navigating with the keyboard. Also the change is not directly related to editing the upload, so in any case I would address it in a separate pull request too.

ma12-co: So this is the current behavior (before this pr):
Peek 2020-08-13 09-43

So you get the big hover feedback of the message but no clickability unless you hover the preview, and since the hover color is the same, you get the hover feedback of the preview only through unwanted stuff:
- overflow from the message on the left;
- invading the name on top;

The keyboard navigation is a bug with how messages hover feedback is implemented. For a reference of how it should really look refer to the mouse navigation feedback. Plus, I think that a bigger click area is generally better, since that space is unused anyway... I implemented the change in this pr because I'm adapting the filepreview component. 


Peek 2020-08-13 09-31

nickvergessen:
Follow up:

- [ ] Ctrl+V with an copied image works on the first try, but on follow up pastes, it just replaces the previous image.
- [ ] When the dialog is open, drag+drop does not work anymore. Would be nice to be able to still drag+drop on top of the modal to add more images to the process

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
```diff
@@ -20,14 +22,15 @@
...
 
 <template>
-	<a :href="link"
-		class="container"
-		:class="{ 'is-viewer-available': isViewerAvailable }"
-		target="_blank"
-		rel="noopener noreferrer"
-		@click="showPreview">
-		<img v-if="!isLoading && !failed"
+	<file-preview v-bind="filePreview"
```
nickvergessen, line 25: This file really bothers me. The shared part is actually so little, that it should just be 2 maintainable subcomponents instead.

PR #3993, opened 2020-08-05
💬 Status adaption for Nextcloud 20

nickvergessen:
- [x] Add away/dnd state on mentions
- [x] Add away/dnd state on the participant list
- [ ] Show status icon as overlay on avatars? => Follow up after nextcloud-vue support

### Mentions

* Status shown when away or dnd

### Participant list

* Status shown when away or dnd
* People with status sorted to the end of their section (moderator, online, user status)

> Bildschirmfoto von 2020-08-05 17-10-03

### DIFF
--- a/css/At.scss
+++ b/css/At.scss
@@ -150,8 +150,8 @@
 		}
 
 		.atwho-cur {
-			background: var(--color-primary);
-			color: var(--color-primary-text);
+			background: var(--color-primary-light);
+			color: var(--color-main-text);
 		}
 	}
 }
--- a/docs/chat.md
+++ b/docs/chat.md
@@ -116,6 +116,7 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
     ------|------|------------
     `search` | string | Search term for name suggestions (should at least be 1 character)
     `limit` | int | Number of suggestions to receive (20 by default)
+    `includeStatus` | bool | Whether the user status information also needs to be loaded
 
 * Response:
     - Status code:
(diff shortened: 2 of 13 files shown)


nickvergessen:
Depends on https://github.com/nextcloud/server/pull/22106, but works otherwise

nickvergessen:
Moving the status to a second line in the participant list is out of my knowledge

jancborchardt:
> Moving the status to a second line in the participant list is out of my knowledge

@ma12-co @skjnldsv to the rescue :)

nickvergessen:
Got some help, so it's now on a new line, but it should probably be trimmed in case the user has bad intentions
Bildschirmfoto von 2020-08-05 15-29-06

nickvergessen:
But we can do that as a follow up

css/At.scss
```diff
@@ -150,8 +150,12 @@
 		}
 
 		.atwho-cur {
-			background: var(--color-primary);
+			background: var(--color-primary-element);
```
jancborchardt, line 153: If we use --color-primary-light, then we don’t have to change the text color, and the status is also distinguished for the person who is selected. That way it also looks consistent with the global style as per Vue components and https://github.com/nextcloud/server/pull/22115

nickvergessen: I tried it together with Marco, it is much worse with certain combinations.

nickvergessen: nevermind, I used element-light, instead of only light

nickvergessen: fixed

PR #4010, opened 2020-08-14
[PoC] [WIP] Matterbridge integration

eneiluj:
Hey! Here is a proof-of-concept that integrates [matterbridge](https://github.com/42wim/matterbridge/) into Talk.

@gary-kim has recently implemented NC Talk compatibility to Matterbridge, which makes it possible to bridge it with many other instant messaging systems (centralized and distributed ones).

Matterbridge binary needs can be installed from https://apps.nextcloud.com/apps/talk_matterbridge

There's a new "bridge" section in the settings tab of the right sidebar for room moderators that lets them configure bridge parts.

Server side generates a bridge config file and runs the binary which connects to each parts of the bridge. The process is launched by the web server user and runs in the background. Its PID is stored and used to kill it when a moderator disables the bridge.

TODO:
* [x] make it work
* [ ] choose where to trigger checks to potentially relaunch dead bridges or kill zombie ones (checks are implemented)
* [ ] improve bridge state display in UI. only the "desired state" is shown for the moment. we could show the log file content.
* [ ] make it safer with more checks when launching the process.
* [ ] write config and log files in appdata (or somewhere else). currently written in /tmp


appinfo/routes.php
```diff
@@ -362,6 +362,33 @@
 				'token' => '^[a-z0-9]{4,30}$',
 			],
 		],
+		[
+			'name' => 'Room#getBridgeOfRoom',
+			'url' => '/api/{apiVersion}/room/{token}/bridge',
+			'verb' => 'GET',
+			'requirements' => [
+				'apiVersion' => 'v(1|2)',
```
nickvergessen, line 370:
```suggestion
				'apiVersion' => 'v1',
```

Same below. bridge api is only version 1 for now :)

lib/BridgeManager.php
```diff
@@ -0,0 +1,399 @@
...
+		$this->manager = $manager;
+		$this->l = $l;
+	}
+
+	public function getBridgeOfRoom(string $token): array {
+		$defaultParts = sprintf(
+			'{"enabled":false,"pid":0,"parts":[{"type":"nctalk","server":"","login":"","password":"","channel":"%s"}]}',
+			$token
+		);
+		$bridgeJSON = $this->config->getAppValue('spreed', 'bridge_' . $token, $defaultParts);
```
nickvergessen, line 67: Will this scale? App configs are loaded always and on each request. I think it should instead be stored in a dedicated database table.

appinfo/routes.php
nickvergessen: Maybe also split it up a bit, so use:
`/api/{apiVersion}/bridge/{token}` as route and a dedicated `BridgeController`

lib/Controller/RoomController.php
```diff
@@ -1363,4 +1369,91 @@ protected function demoteGuestFromModerator(Room $room, string $sessionId): Data
...
+	 * @NoAdminRequired
+	 *
+	 * @return DataResponse
+	 */
+	public function deleteBridgeOfRoom(string $token): DataResponse {
+		try {
+			$room = $this->manager->getRoomForParticipantByToken($token, $this->userId, true);
+			try {
+				$participant = $room->getParticipant($this->userId);
+				if ($participant->hasModeratorPermissions()) {
```
nickvergessen, line 1442: If this stays in talk, use `@RequireLoggedInModeratorParticipant` annotation instead.

lib/Controller/RoomController.php
```diff
@@ -1363,4 +1369,91 @@ protected function demoteGuestFromModerator(Room $room, string $sessionId): Data
...
+	 * @NoAdminRequired
+	 *
+	 * @return DataResponse
+	 */
+	public function editBridgeOfRoom(string $token, bool $enabled, array $parts = []): DataResponse {
+		try {
+			$room = $this->manager->getRoomForParticipantByToken($token, $this->userId, true);
+			try {
+				$participant = $room->getParticipant($this->userId);
+				if ($participant->hasModeratorPermissions()) {
```
nickvergessen, line 1412: `@RequireLoggedInModeratorParticipant`

lib/Controller/RoomController.php
```diff
@@ -1363,4 +1369,91 @@ protected function demoteGuestFromModerator(Room $room, string $sessionId): Data
...
+	 * @NoAdminRequired
+	 *
+	 * @return DataResponse
+	 */
+	public function getBridgeOfRoom(string $token): DataResponse {
+		try {
+			$room = $this->manager->getRoomForParticipantByToken($token, $this->userId, true);
+			try {
+				$participant = $room->getParticipant($this->userId);
+				if ($participant->hasModeratorPermissions()) {
```
nickvergessen, line 1385: `@RequireLoggedInModeratorParticipant`

lib/BridgeManager.php
```diff
@@ -0,0 +1,399 @@
...
+		// then actually delete the config
+		$bridgeJSON = $this->config->deleteAppValue('spreed', 'bridge_' . $token);
+		return true;
+	}
+
+	public function checkAllBridges() {
+		// TODO call this from time to time to make sure everything is running fine
+		$this->manager->forAllRooms(function ($room) {
+			$token = $room->getToken();
+			if ($room->getType() === Room::GROUP_CALL or $room->getType() === Room::PUBLIC_CALL) {
```
nickvergessen, line 106:
```suggestion
			if ($room->getType() === Room::GROUP_CALL || $room->getType() === Room::PUBLIC_CALL) {
```

Never use `and` and `or` in php, it always executes all conditions. You had an other occurance with `isset() and ...` which will still trigger an error because of using undefined.

eneiluj: Good to know. Thanks

lib/BridgeManager.php
eneiluj: Maybe I'm missing something here but it seems to me that accessing a dedicated DB table will be the same as calling getAppValue as app values are stored in the DB.

We need the bridge information when loading the sidebar tab and when performing config editions. We can't do it less often.

Do you think of a better (and/or more efficient) way to store/get bridge configs?

eneiluj:
@nickvergessen Thanks a lot for the review.

lib/BridgeManager.php
nickvergessen: > Maybe I'm missing something here but it seems to me that accessing a dedicated DB table will be the same as calling getAppValue as app values are stored in the DB.

Well but appconfig is loaded also when you do something in deck, ping the root of files as a sync client, check for notifications, etc. Now if someone setups up like hundreds of bridges it will load hundreds of entries every time.

eneiluj: Ok I'll put it in a dedicated table. Didn't think about the importance of keeping appconfig reasonably small. Thanks.

lib/Migration/Version10000Date20200819121721.php
```diff
@@ -0,0 +1,68 @@
...
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options) {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if (!$schema->hasTable('talk_bridges')) {
+			$table = $schema->createTable('talk_bridges');
+			$table->addColumn('id', Type::INTEGER, [
+				'autoincrement' => true,
+				'notnull' => true,
+				'length' => 11,
```
nickvergessen, line 50:
```suggestion
			$table->addColumn('id', Type::BIGINT, [
				'autoincrement' => true,
				'notnull' => true,
```

lib/Migration/Version10000Date20200819121721.php
```diff
@@ -0,0 +1,68 @@
...
+			]);
+			$table->addColumn('room_id', Type::INTEGER, [
+				'notnull' => true,
+				'length' => 11,
+				'unsigned' => true,
+				'default' => 0,
```
nickvergessen, line 56:
```suggestion
			$table->addColumn('room_id', Type::BIGINT, [
				'notnull' => true,
				'unsigned' => true,
```

lib/Migration/Version10000Date20200819121721.php
```diff
@@ -0,0 +1,68 @@
...
+			]);
+			$table->addColumn('json_values', Type::STRING, [
+				'notnull' => true,
+				'length' => 10000,
+				'default' => '{}',
```
nickvergessen, line 61:
```suggestion
			$table->addColumn('json_values', Type::TEXT, [
				'notnull' => true,
```

lib/BridgeManager.php
```diff
@@ -151,22 +156,23 @@ private function editBridgeConfig(string $token, array $newBridge) {
...
 	}
 
-	private function checkBotUser(string $token) {
+	private function checkBotUser(string $token, bool $create): array {
 		$botUserId = 'matterbridge-bot';
-		// TODO check user exists and create it if necessary
+		// check if user exists and create it if necessary
 		if (!$this->userManager->userExists($botUserId)) {
 			$pass = md5(strval(rand()));
+			$this->config->setAppValue('spreed', 'bot_pass', $pass);
```
nickvergessen, line 220: Make sure to add this to https://github.com/nextcloud/server/blob/cb057829f72c70e819f456edfadbb29d72dba832/lib/private/AppConfig.php#L51-L57

lib/BridgeManager.php
```diff
@@ -183,22 +189,36 @@ private function checkBotUser(string $token) {
...
+		// delete old bot app tokens for this room
+		$tokenName = 'spreed_' . $token;
 		$tokens = $this->tokenProvider->getTokenByUser($botUserId);
-		foreach ($tokens as $token) {
-			$j = $token->jsonSerialize();
-			$keys = implode(',',array_keys($j));
-			error_log('KKKKKK '.$keys.'||||');
-			// TODO revoke existing tokens for this room
+		foreach ($tokens as $t) {
+			$j = $t->jsonSerialize();
```
nickvergessen, line 241: instead use the dedicated methods of the interface? `$t->getName()` and so on?

lib/BridgeManager.php
eneiluj: Thanks a lot for the review!

I don't know if it even makes sense to store the bot password as I've seen app token generation works whatever 'password' parameter we give it.

Maybe I didn't grasp the meaning of the 'password' parameter when generating an app token. It appears not to be the user password. Do you know what it is used for?

nickvergessen: I think its needed to be able to unencrypt the user keys in the session. So in case of server side encryption you wouldn't be able to read files. So make sure it's stored and save

lib/BridgeManager.php
```diff
@@ -0,0 +1,624 @@
...
+				$qb->expr()->eq('room_id', $qb->createNamedParameter($roomId, IQueryBuilder::PARAM_INT))
+			);
+		$req = $qb->execute();
+		$jsonValues = '{"enabled":false,"pid":0,"parts":[]}';
+		while ($row = $req->fetch()) {
+			$jsonValues = $row['json_values'];
+			break;
+		}
+		$req->closeCursor();
+		$qb = $qb->resetQueryParts();
```
nickvergessen, line 580: not needed as `$qb` is local and destroyed after this function call

lib/BridgeManager.php
```diff
@@ -0,0 +1,624 @@
...
+
+		$exists = false;
+		$qb->select('json_values')
+			->from('talk_bridges', 'b')
+			->where(
+				$qb->expr()->eq('room_id', $qb->createNamedParameter($roomId, IQueryBuilder::PARAM_INT))
+			);
+		$req = $qb->execute();
+		while ($row = $req->fetch()) {
+			$exists = true;
```
nickvergessen, line 600: Instead of checking existance you could also insert and on `UniqueConstraintViolation` update.

src/components/BridgeSettings.vue
```diff
@@ -0,0 +1,515 @@
...
+							icon: 'icon-user',
+						},
+						password: {
+							type: 'password',
+							placeholder: t('spreed', 'User password'),
+							icon: 'icon-category-auth',
+						},
+						channel: {
+							type: 'text',
+							placeholder: t('spreed', 'Talk room'),
```
nickvergessen, line 127:
```suggestion
							placeholder: t('spreed', 'Talk conversation'),
```

src/components/BridgeSettings.vue
```diff
@@ -0,0 +1,515 @@
...
+	align-items: center;
+}
+
+#matterbridge-header {
+	padding-left: 15px;
+}
+
+.icon {
+	display: inline-block;
+	width: 8%;
```
nickvergessen, line 513: icons should always have a fixed width

lib/BridgeManager.php
```diff
@@ -0,0 +1,624 @@
...
+			if (count(preg_split('/\n/', $result)) > 2) {
+				return true;
+			}
+		} catch (Exception $e) {
+		}
+		return false;
+	}
+
+	private function getBridgeFromDb(string $token): array {
+		$room = $this->manager->getRoomByToken($token);
```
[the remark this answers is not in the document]
eneiluj: Ok right.

What about the config file name? `bridge-TOKEN.toml` or `bridge-ROOM_ID.toml`?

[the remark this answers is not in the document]
nickvergessen: I think token is better as when something is wrong you know the token as its in the url, but not the room id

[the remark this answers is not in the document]
eneiluj: Maybe you were thinking of another way to get the room in the controller. I get it with the RoomManager from the token. Well at least the manager part is done :grin:.

lib/BridgeManager.php
```diff
@@ -0,0 +1,624 @@
...
+
+		if (isset($bridge['pid']) && intval($bridge['pid']) !== 0) {
+			// config : there is a PID stored
+			error_log('pid is defined in config |||||');
+			$pid = intval($bridge['pid']);
+			$isRunning = $this->isRunning($pid);
+			// if bridge running and enabled is false : kill it
+			if ($isRunning) {
+				if ($bridge['enabled']) {
+					error_log('process running AND config enabled : doing nothing |||||');
```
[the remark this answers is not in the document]
eneiluj: What do you mean? If bridge is running, then both `$bridge['enabled']` boolean states are handled.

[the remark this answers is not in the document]
nickvergessen: we don't use `error_log` instead inject `Psr\Log\LoggerInterface` and use the `->error()` function there.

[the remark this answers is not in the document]
eneiluj: It's level 1 logs to avoid flooding. Any difference between `Psr\Log\LoggerInterface` and `OCP\ILogger`?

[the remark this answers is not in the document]
nickvergessen: ILogger is deprecated in favor of LoggerInterface

nickvergessen:
Binary repo is at https://github.com/nextcloud/talk_matterbridge

nickvergessen:
How installing looks:

Peek 2020-08-24 13-38

nickvergessen (approved):
Doesn't break the rest and should work as per Julien

PR #4016, opened 2020-08-17
Show the status on avatars

nickvergessen:
Bildschirmfoto von 2020-08-18 13-15-01

### DIFF
--- a/css/At.scss
+++ b/css/At.scss
@@ -125,7 +125,7 @@
 	// Override autocompletion panel items height, as they are too short
 	// for the avatars and also need some padding.
 	.atwho-li {
-		height: 48px;
+		height: 64px;
 		padding-top: 4px;
 		padding-bottom: 4px;
 	}
@@ -134,8 +134,8 @@
 	// is the default 16px. This is a temporary fix until it is fixed
 	// in the avatar component.
 	.atwho-li .icon-group-forced-white {
-		width: 32px;
-		height: 32px;
+		width: 44px;
+		height: 44px;
 	}
 
 	// Override all colors with our theming and dark mode
--- a/css/At.scss
+++ b/css/At.scss
@@ -111,11 +111,11 @@
 	// Override "max-width: 180px", as that makes the autocompletion panel too
 	// narrow.
 	.atwho-view {
-		max-width: unset;
+		max-width: 300px;
+		min-width: 250px;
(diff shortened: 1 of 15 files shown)


nickvergessen:
> Yep, it shouldn’t be shown in those little mention chips. (Neither in comments btw, in case that uses the same component.)

Well, then let me change the default for the mention chip

nickvergessen:
Done via https://github.com/nextcloud/nextcloud-vue/pull/1259/commits/6ae3c95dde9defdde6abbf3b5804330a913fbf2f

nickvergessen:
Aparently 2.5.0 has issues, so lets wait for 2.5.1

ma12-co (commented):
Nice! I would say that the status circle should be smaller though. Also Imo there's no need for an icon in the yellow and red one.

nickvergessen:
Invisible users are now "offline"

nickvergessen:
Bildschirmfoto von 2020-08-19 18-45-05

Looong status ellipsis now in the mention suggestion and the right sidebar.

nickvergessen:
The page reloads currently when clicking on the `Share file` action menu

nickvergessen:
Bildschirmfoto von 2020-08-24 16-19-08

Double scrollbars when the participant menu is opened. Seems it doesn't respect the boundaries element body or something?

nickvergessen:
Chat in the sidebar breaks the scrolling (it doesnt overflow anymore)

jancborchardt:
Also agree with @nickvergessen that it looks quite off with the huge avatars in the suggestions and the sidebar. We could try that again if we also increase the global text size and also increase the avatar size in the chat list. But otherwise it’s a bit too big.

nickvergessen:
> Also agree with nickvergessen that it looks quite off with the huge avatars in the suggestions and the sidebar. We could try that again if we also increase the global text size and also increase the avatar size in the chat list. But otherwise it’s a bit too big.

That issue has been solved in the meantime, there was a wrong offset making everything go wild.

nickvergessen:
> Add boundarieselement to upload action 

It was already working there, only the participants menu is wrong

ma12-co:
> It was already working there, only the participants menu is wrong

If you made the window small enough it would go underneath the sidebar before this commit for me

ma12-co:
Screenshot from 2020-08-26 13-59-02

PR #4022, opened 2020-08-19
Use SVG based placeholders like in unified search

nickvergessen:
- [ ] Requires https://github.com/nextcloud/server/pull/22111 in server

Peek 2020-08-19 15-08


skjnldsv:
> Requires [nextcloud/server#22111](https://github.com/nextcloud/server/pull/22111) in server

why does that requires the unified search?

nickvergessen:
> why does that requires the unified search?

Because that is where the SCSS vars for the colors are defined

nickvergessen:
See https://github.com/nextcloud/server/pull/22111/files first and second file

PR #4058, opened 2020-08-25
Conversations list auto scrolling

ma12-co:


src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -134,20 +135,25 @@ export default {
 		 *
 		 * @returns {String}
 		 */
+
```
nickvergessen, line 138: this is off, the doc block should be directly on the method

nickvergessen (approved):

PR #4061, opened 2020-08-26
Bugfix/noid/fix buttons

ma12-co:
44px all the buttons as discussed in design review.

also rounded searchboxes

Screenshot from 2020-08-26 15-47-08


nickvergessen (approved):
Works, but I fixed the conversation-creation icon to be the right color based on theme instead of `icon-white` which gets inverted by dark-mode instead of primary color

nickvergessen (approved):

PR #4079, opened 2020-08-31
Make sure matterbridge binary is executable

eneiluj:
File permissions are not preserved when Php extracts the talk_matterbridge app so the included binaries have 644 rights.

So we can either:
* Do it like this PR: add execution permission to the binary in Talk when checking its version. Might be a little bit repetitive but it's safe.
* Or do it in every talk_matterbridge migration scripts.
* Or do it via a post upgrade script (repair step) in talk_matterbridge.

I prefer the first solution because it's safer. It covers the cases where rights are changed or file is overwritten by a daring user after enabling matterbridge integration.

### DIFF
--- a/lib/MatterbridgeManager.php
+++ b/lib/MatterbridgeManager.php
@@ -40,6 +40,7 @@ use Doctrine\DBAL\Exception\UniqueConstraintViolationException;
 use Psr\Log\LoggerInterface;
 
 use OCA\Talk\Exceptions\ImpossibleToKillException;
+use OCA\Talk\Exceptions\WrongPermissionsException;
 use OCA\Talk\Exceptions\ParticipantNotFoundException;
 
 class MatterbridgeManager {
@@ -640,6 +641,33 @@ class MatterbridgeManager {
 		$binaryPath = $this->config->getAppValue('spreed', 'matterbridge_binary');
 		if (!file_exists($binaryPath)) {
 			return null;
+		}
+
+		// is www user the file owner?
+		$user = posix_getpwuid(posix_getuid());
+		$fileOwner = posix_getpwuid(fileowner($binaryPath));
+		$userIsOwner = $user['name'] === $fileOwner['name'];
+
+		// get file group and user groups
+		$fileGid = filegroup($binaryPath);
+		$myGids = posix_getgroups();
+
+		// check permissions
+		$perms = fileperms($binaryPath);
+		$uExec = (($perms & 0x0040) && !($perms & 0x0800));
+		$gExec = (($perms & 0x0008) && !($perms & 0x0400));
+		$aExec = (($perms & 0x0001) && !($perms & 0x0200));
+
+		// 3 ways to have the permission :
+		// * be the owner and have u+x perm
+		// * not be the owner, be in the file group and have g+x perm
+		// * have o+x perm
+		$execPerm = $aExec
+			|| ($user['name'] === $fileOwner['name'] && $uExec)
+			|| ($user['name'] !== $fileOwner['name'] && in_array($fileGid, $myGids) && $gExec);
+
+		if (!$execPerm) {
+			throw new WrongPermissionsException();
 		}
 
 		$cmd = escapeshellcmd($binaryPath) . ' ' . escapeshellarg('-version');
--- a/src/components/AdminSettings/MatterbridgeIntegration.vue
+++ b/src/components/AdminSettings/MatterbridgeIntegration.vue
@@ -49,6 +49,10 @@
 		<template v-else>
 			<p class="settings-hint" v-html="description" />
 
+			<p v-if="errorText" class="settings-hint">
+				{{ errorText }}
+			</p>
+
 			<p>
 				<button v-if="isInstalling">
 					<span class="icon icon-loading-small" />
@@ -82,6 +86,7 @@ export default {
 			matterbridgeEnabled: loadState('talk', 'matterbridge_enable'),
 			matterbridgeVersion: loadState('talk', 'matterbridge_version'),
 			isInstalling: false,
+			error: loadState('talk', 'matterbridge_error'),
 		}
 	},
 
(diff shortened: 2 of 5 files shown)


lib/MatterbridgeManager.php
```diff
@@ -642,6 +642,9 @@ public function getCurrentVersionFromBinary(): ?string {
 			return null;
 		}
 
+		$cmd = sprintf('chmod u+x %s', escapeshellarg($binaryPath));
+		@exec($cmd, $output, $returnCode);
```
nickvergessen, line 646: I would like to not do this. This is job of the sysadmin and nothing we should fix in the software.

eneiluj: What do you mean? In the current state, it does not work because extracted binaries don't have execution permissions (at least on my system).

Are you suggesting we tell the admin to `chmod`?

nickvergessen: For me the items from https://apps.nextcloud.com/apps/talk_matterbridge are executable when extracting the package.

nickvergessen: But yeah, the admin should decide on this and fix it. I'm too scared to have code in place that automatically makes a given config string executable and executes it.

eneiluj: > For me the items from https://apps.nextcloud.com/apps/talk_matterbridge are executable when extracting the package.

I guess it depends on Php config. Mine is default from Ubuntu and binaries are 644 when extracted.

> the admin should decide on this and fix it

Ok then should we check permissions and display the suggested fix only if necessary?

nickvergessen: yeah, just show something like "Binary is not executable, please make sure the file is owned by the correct user and can be executed." Maybe you can copy the logic from https://github.com/nextcloud/server/blob/cb057829f72c70e819f456edfadbb29d72dba832/console.php#L67-L76 to show the needed user

eneiluj:
So now, admin gets to know if the binary file owner is incorrect or if it has wrong permissions.

I would still like to fix it on our side since it adds potential confusion and extra actions for admins. Plus the problem will probably reappear on each talk_matterbridge new version.

How about checking (and fixing) this in talk_matterbridge migration script or post upgrade repair step? This way, check and fix are done once for each talk_matterbridge release.

src/components/AdminSettings/MatterbridgeIntegration.vue
```diff
@@ -138,12 +143,17 @@ export default {
 				this.matterbridgeVersion = response.data.ocs.data.version
 				this.matterbridgeEnabled = true
 				this.saveMatterbridgeEnabled()
-
-				this.isInstalling = false
+				this.errorText = ''
 			} catch (e) {
-				console.error(e)
 				showError(t('spreed', 'Failed to execute Matterbridge binary.'))
+				if (e.response.data.ocs.data.error === 'binary_permissions') {
```
nickvergessen, line 149: this should always show? Not only after clicking the enable/install button?

eneiluj: Right, done.

src/components/AdminSettings/MatterbridgeIntegration.vue
```diff
@@ -138,12 +152,12 @@ export default {
 				this.matterbridgeVersion = response.data.ocs.data.version
 				this.matterbridgeEnabled = true
 				this.saveMatterbridgeEnabled()
-
-				this.isInstalling = false
+				this.error = ''
 			} catch (e) {
-				console.error(e)
```
nickvergessen, line 144: Please keep this

lib/MatterbridgeManager.php
```diff
@@ -642,6 +643,33 @@ public function getCurrentVersionFromBinary(): ?string {
...
+
+		// check permissions
+		$perms = fileperms($binaryPath);
+		$uExec = (($perms & 0x0040) && !($perms & 0x0800));
+		$gExec = (($perms & 0x0008) && !($perms & 0x0400));
+		$aExec = (($perms & 0x0001) && !($perms & 0x0200));
+
+		// 3 ways to have the permission :
+		// * be the owner and have u+x perm
+		// * not be the owner, be in the file group and have g+x perm
```
nickvergessen, line 663: I can also be the owner and have execute only via the group? Should work too?

lib/MatterbridgeManager.php
```diff
@@ -642,6 +643,33 @@ public function getCurrentVersionFromBinary(): ?string {
...
+		// * have o+x perm
+		$execPerm = ($user['name'] === $fileOwner['name'] && $uExec)
+			|| ($user['name'] !== $fileOwner['name'] && in_array($fileGid, $myGids) && $gExec)
+			|| $aExec;
```
nickvergessen, line 667:
```suggestion
		$execPerm = $aExec
			|| ($user['name'] === $fileOwner['name'] && $uExec)
			|| (in_array($fileGid, $myGids) && $gExec);
```

Sorted from easy comparison to hardest as complained by phpstorm

src/components/AdminSettings/MatterbridgeIntegration.vue
```diff
@@ -138,12 +152,13 @@ export default {
 				this.matterbridgeVersion = response.data.ocs.data.version
 				this.matterbridgeEnabled = true
 				this.saveMatterbridgeEnabled()
-
-				this.isInstalling = false
+				this.error = ''
 			} catch (e) {
 				console.error(e)
 				showError(t('spreed', 'Failed to execute Matterbridge binary.'))
+				this.error = e.response.data.ocs.data.error
```
nickvergessen, line 159: There could be an error where this is undefined (e.g. maintenance mode). So it would need some sanity check and a fallback error message I guess?

lib/MatterbridgeManager.php
eneiluj: I tried and exec permission is not granted in this case

lib/MatterbridgeManager.php
eneiluj: Done. I kept the `$user['name'] !== $fileOwner['name']` associated with group permission because of previous explanation.

lib/MatterbridgeManager.php
nickvergessen: Okay, interesting to know :D

src/components/AdminSettings/MatterbridgeIntegration.vue
```diff
@@ -49,6 +49,10 @@
 		<template v-else>
 			<p class="settings-hint" v-html="description" />
 
+			<p class="settings-hint">
```
nickvergessen, line 52:
```suggestion
			<p v-if="errorText" class="settings-hint">
```

nickvergessen (approved):
Looks good otherwise

eneiluj:
Last fix done, rebased on master, squashed. Thanks for all the feedback.

nickvergessen (approved):

PR #4083, opened 2020-09-01
Show bridge process state and log

eneiluj:
Matterbridge process state is updated and displayed when saving bridge config and when showing log file content.

Seeing log file content is mandatory when setting up Ms-teams as Matterbridge provides a link to finalize sign-in in those logs.

Not sure about the "show mattberbridge log" button text and location.

### DIFF
--- a/lib/Controller/MatterbridgeController.php
+++ b/lib/Controller/MatterbridgeController.php
@@ -60,9 +60,25 @@ class MatterbridgeController extends AEnvironmentAwareController {
 	 * @return DataResponse
 	 */
 	public function getBridgeOfRoom(): DataResponse {
-		$this->bridgeManager->checkBridge($this->room);
+		$pid = $this->bridgeManager->checkBridge($this->room);
+		$logContent = $this->bridgeManager->getBridgeLog($this->room);
 		$bridge = $this->bridgeManager->getBridgeOfRoom($this->room);
+		$bridge['running'] = ($pid !== 0);
+		$bridge['log'] = $logContent;
 		return new DataResponse($bridge);
+	}
+
+	/**
+	 * Get bridge process information
+	 *
+	 * @NoAdminRequired
+	 * @RequireLoggedInModeratorParticipant
+	 *
+	 * @return DataResponse
+	 */
+	public function getBridgeProcessState(): DataResponse {
+		$result = $this->bridgeManager->getBridgeProcessState($this->room);
+		return new DataResponse($result);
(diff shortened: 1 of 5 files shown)


src/components/RightSidebar/Matterbridge/MatterbridgeSettings.vue
```diff
@@ -485,6 +508,24 @@ export default {
 				console.debug(exception)
 			}
 			this.loading = false
+			setTimeout(() => this.getBridgeProcessState(this.token), 4000)
```
nickvergessen, line 511: Why the delay? with the await above it should be good already? Or is this the "startup" time of matterbridge binary? Seems a bit weird to wait a random amount of time?

eneiluj: It's weird indeed. I just want to wait that the log file is a little bit populated. But as we get the log file content each time the user wants to see it, the delay now seems useless.

Still, it might not be so useless to wait a few seconds. We might see the process alive right after launching it but it will die in a few seconds. I think it has happened to me a few times.

What do you think?

nickvergessen: Well in this case I would say `editBridge` should return the info, and in case it was not there (yet) it can be refetched/checked after some time. But randomly refreshing after 4 seconds feels weird, in case it terminates after 30 sec it still shows as functional?

eneiluj:
Here is a clarified version. Like you said, it's better if editBridge returns the state.

So we now get process state every minute. The loop is relaunched when changing the room. We could use a much lower frequency. What do you think?

lib/Controller/MatterbridgeController.php
```diff
@@ -60,11 +60,27 @@ public function __construct(string $appName,
...
+	 * Get bridge process information
+	 *
+	 * @NoAdminRequired
+	 * @RequireLoggedInModeratorParticipant
+	 *
+	 * @return DataResponse
+	 */
+	public function getBridgeProcessState(): DataResponse {
+		$result = $this->bridgeManager->getBridgeProcessState($this->room);
+		return new DataResponse($result);
```
nickvergessen, line 81:
```suggestion
		$state = $this->bridgeManager->getBridgeProcessState($this->room);
		return new DataResponse($state);
```

nickvergessen (approved):
LGTM but didn't execute

PR #4086, opened 2020-09-02
Don't flicker the call state when participants are updated while join…

nickvergessen:
…ing/leaving

> Joining/Leaving a call is an async process and can take a while,
especially when starting a call and sending a lot of notifications.
Now we keep a force call state list for your session while joining/leaving.

Reopening #3584

### DIFF
--- a/src/store/participantsStore.js
+++ b/src/store/participantsStore.js
@@ -31,8 +31,11 @@ import {
 	leaveCall,
 } from '../services/callsService'
 import { PARTICIPANT } from '../constants'
+import { EventBus } from '../services/EventBus'
 
 const state = {
+	forceCallFlag: {
+	},
 	participants: {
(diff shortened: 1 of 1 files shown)


src/store/participantsStore.js
```diff
@@ -31,8 +31,11 @@ import {
 	leaveCall,
 } from '../services/callsService'
 import { PARTICIPANT } from '../constants'
+import { EventBus } from '../services/EventBus'
 
 const state = {
+	forceCallFlag: {
```
PVince81, line 37: add comment as it's not clear what "forcing a call" means

nickvergessen:
Closing for now. I still think it's better to be quick reactive (our company call is like ~20+ sec in this mixed state).
But it needs some more thinking

PR #4089, opened 2020-09-02
Show keyboard shortcuts in settings like in mail and calendar

nickvergessen:
Bildschirmfoto von 2020-09-02 13-32-44


ma12-co (changes requested):
This shouldn't be in the settings in my opinion. If you want to have it there for now anyway, at least make it the last section of the settings. Not before the input selection.

nickvergessen:
Calendar and mail have a link in the bottom left settings to open a modal which shows this.
Since settings already pops a modal for us, I think it's the most logical thing to do.

But sure we can move it to the end.

src/components/SettingsDialog/SettingsDialog.vue
```diff
@@ -158,6 +193,10 @@ export default {
 	&__input {
 		width: 100%;
 	}
+
+	dd {
```
ma12-co, line 197: Please always target elements with classes :) 

https://vue-loader.vuejs.org/guide/scoped-css.html#also-keep-in-mind

ma12-co (changes requested):

ma12-co (approved):

Issue #4097, opened 2020-09-04
matterbridge: restart bridges automatically


jospoortvliet:
Right now, if a matterbridge bridge process crashes or the server is rebooted, the bridge remains down. Would be good to have a cronjob that runs every now and then to check if all configured bridges are running and, when one is down, restart it.

eneiluj:
@nickvergessen Would you be ok with an hourly cron job calling `checkAllBridges()`?

nickvergessen:
Can also run every 5 minutes, don't care too much.

jospoortvliet:
and #4119 fixes this!

PR #4119, opened 2020-09-07
Job to check Matterbridge processes

eneiluj:
refs #4097

Start the ones that should be running, kill the ones that are not related with a room in config.

### DIFF
--- a/lib/BackgroundJob/CheckMatterbridges.php
+++ b/lib/BackgroundJob/CheckMatterbridges.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 /**
- * @copyright Copyright (c) 2017 Joas Schilling <coding@schilljs.com>
+ * @copyright Copyright (c) 2020 Julien Veyssier <eneiluj@posteo.net>
  *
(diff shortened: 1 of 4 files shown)


lib/BackgroundJob/CheckMatterbridges.php
```diff
@@ -0,0 +1,57 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Talk\BackgroundJob;
+
+use OC\BackgroundJob\TimedJob;
```
nickvergessen, line 26:
```suggestion
use OCP\BackgroundJob\TimedJob;
```

nickvergessen (changes requested):
Use public api as commented

eneiluj:
I took the liberty to fix this for other background jobs (where I got the error from :grin:).

nickvergessen (approved):

Issue #4099, opened 2020-09-04
Log matterbridge config changes in chat


nickvergessen:
When a moderator adds/updates/delete matterbridge config manually in the web UI, it should be written into the chat as a system message, to make clear to users that their chat messages might be mirrored to another server.

jospoortvliet:
fixed in #4121

PR #4121, opened 2020-09-07
Log matterbridge start/stop actions in related room

eneiluj:
refs #4099 

I just had trouble including the bridged systems in the system messages. Hints are welcome :grin:.

### DIFF
--- a/lib/Chat/Parser/SystemMessage.php
+++ b/lib/Chat/Parser/SystemMessage.php
@@ -289,6 +289,21 @@ class SystemMessage {
 					$parsedMessage = $this->l->t('You shared a file which is no longer available');
 				}
 			}
+		} elseif ($message === 'matterbridge_config_added') {
+			$parsedMessage = $this->l->t('{actor} set up Matterbridge to synchronize this conversation with other chats.');
+			if ($currentUserIsActor) {
+				$parsedMessage = $this->l->t('You set up Matterbridge to synchronize this conversation with other chats.');
+			}
+		} elseif ($message === 'matterbridge_config_edited') {
+			$parsedMessage = $this->l->t('{actor} updated the Matterbridge configuration.');
+			if ($currentUserIsActor) {
+				$parsedMessage = $this->l->t('You updated the Matterbridge configuration.');
+			}
+		} elseif ($message === 'matterbridge_config_removed') {
+			$parsedMessage = $this->l->t('{actor} removed the Matterbridge configuration.');
+			if ($currentUserIsActor) {
+				$parsedMessage = $this->l->t('You removed the Matterbridge configuration.');
+			}
 		} else {
 			throw new \OutOfBoundsException('Unknown subject');
 		}
--- a/lib/MatterbridgeManager.php
+++ b/lib/MatterbridgeManager.php
@@ -38,10 +38,12 @@ use OCP\Security\ISecureRandom;
 use OCP\IAvatarManager;
 use Doctrine\DBAL\Exception\UniqueConstraintViolationException;
 use Psr\Log\LoggerInterface;
+use OCP\AppFramework\Utility\ITimeFactory;
 
 use OCA\Talk\Exceptions\ImpossibleToKillException;
 use OCA\Talk\Exceptions\WrongPermissionsException;
 use OCA\Talk\Exceptions\ParticipantNotFoundException;
+use OCA\Talk\Chat\ChatManager;
 
 class MatterbridgeManager {
 	/** @var IDBConnection */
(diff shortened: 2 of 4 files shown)


lib/MatterbridgeManager.php
```diff
@@ -493,6 +503,7 @@ private function checkBridgeProcess(Room $room, array $bridge, bool $relaunch =
 					$killed = $this->killPid($bridge['pid']);
 					if ($killed) {
 						$pid = 0;
+						$this->sendSystemBridgeStop($room);
```
nickvergessen, line 506: This is not what I had in mind. I would just do it when a bridge config is done. Not based on the real process.

`{actor} set up Matterbridge to synchronize this conversation with other chats.`
`{actor} updated the Matterbridge configuration.`
`{actor} removed the Matterbridge configuration.`

eneiluj: > I would just do it when a bridge config is done

Not sure I understand what you mean. That's more or less what's done here. When the config is generated and the bridge is started or stopped, there is a system message. Do you mean you would like to trigger a system message each time the config file is edited?

I think it makes more sense to notify only when the bridge is started and stopped because the config can be edited many times without starting the bridge. The important information for the user is to know when messages are forwarded or not. Config change events seem less pertinent in a user point of view.

nickvergessen: > Do you mean you would like to trigger a system message each time the config file is edited?

Yeah and only then. Not otherwise you will create a chat message in each chat when the server is rebooted e.g.
I want users to be aware when their chat messages are leaked to other chats, they don't need to know if this services failed right now or not. And other moderators can then check what service was configured when "User X changed the matterbridge configuration" is posted into the chat. Logging who changed something crucial like this is the minimum. It's like we log when a chat is allowed/disallowing guests.

eneiluj: Ok. Tiny remaining problem is that currently it will produce many system messages because the config is saved each time a part is added/removed. I'll prevent that as the save button is enough.

eneiluj:
New stuff:
* rebased on master and squashed
* do not save when deleting a bridge part anymore
* only notify (system message) when:
    * config was empty and is not anymore
    * config was not empty but now is
    * config has changed

lib/MatterbridgeManager.php
```diff
@@ -548,6 +560,106 @@ private function checkBridgeProcess(Room $room, array $bridge, bool $relaunch =
...
+		foreach ($part1 as $key => $val) {
+			if (!isset($part2[$key]) || $val !== $part2[$key]) {
+				return false;
+			}
+		}
+
+		return true;
+	}
+
+	private function sendSystemBridgeAdded(Room $room, string $userId): void {
```
nickvergessen, line 630: Combine all of those 3 to one and add `'matterbridge_config_added'` as a parameter?

eneiluj:
Factorized system message methods and rebased on master.

lib/Chat/Parser/SystemMessage.php
```diff
@@ -289,6 +289,12 @@ public function parseMessage(Message $chatMessage): void {
 					$parsedMessage = $this->l->t('You shared a file which is no longer available');
 				}
 			}
+		} elseif ($message === 'matterbridge_config_added') {
+			$parsedMessage = $this->l->t('{actor} set up Matterbridge to synchronize this conversation with other chats.');
+		} elseif ($message === 'matterbridge_config_edited') {
+			$parsedMessage = $this->l->t('{actor} updated the Matterbridge configuration.');
+		} elseif ($message === 'matterbridge_config_removed') {
+			$parsedMessage = $this->l->t('{actor} removed the Matterbridge configuration.');
```
nickvergessen, line 297:
```suggestion
		} elseif ($message === 'matterbridge_config_added') {
			$parsedMessage = $this->l->t('{actor} set up Matterbridge to synchronize this conversation with other chats.');
			if ($currentUserIsActor) {
					$parsedMessage = $this->l->t('You set up Matterbridge to synchronize this conversation with other chats.');
			}
		} elseif ($message === 'matterbridge_config_edited') {
			$parsedMessage = $this->l->t('{actor} updated the Matterbridge configuration.');
			if ($currentUserIsActor) {
					$parsedMessage = $this->l->t('You updated the Matterbridge configuration.');
			}
		} elseif ($message === 'matterbridge_config_removed') {
			$parsedMessage = $this->l->t('{actor} removed the Matterbridge configuration.');
			if ($currentUserIsActor) {
					$parsedMessage = $this->l->t('You removed the Matterbridge configuration.');
			}
```

nickvergessen (approved):

PR #4122, opened 2020-09-07
Kill matterbridge before editing the config file

eneiluj:
Matterbridge is supposed to watch changes in its config file but, for example, it does not disconnect from a removed part and continues forwarding messages there.

I think it safer to kill it and relaunch a new process when editing the config file.


nickvergessen (approved):
🦈

PR #4127, opened 2020-09-08
Add a default on NotNull column

nickvergessen:
Ref https://github.com/nextcloud/server/pull/22643

Since we always set it to `'0'` anyway, this is only an issue when updating from a previous version on Oracle. Since it didn't install, that is a non issue and we don't need to update the default on current versions.


danxuliu (approved):
Tested (although not with Oracle :shrug:) and works :+1:

PR #4137, opened 2020-09-11
Don't leak the status when the app is disabled for the user

nickvergessen:

### DIFF
--- a/lib/Controller/RoomController.php
+++ b/lib/Controller/RoomController.php
@@ -842,7 +842,10 @@ class RoomController extends AEnvironmentAwareController {
 		$results = [];
 
 		$statuses = [];
-		if ($this->userId !== null && $includeStatus && count($participants['users']) < 100) {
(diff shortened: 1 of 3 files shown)


lib/Controller/RoomController.php
```diff
@@ -842,7 +842,10 @@ public function getParticipants(bool $includeStatus = false): DataResponse {
 		$results = [];
 
 		$statuses = [];
-		if ($this->userId !== null && $includeStatus && count($participants['users']) < 100) {
+		if ($this->userId !== null
+			&& $includeStatus
+			&& count($participants['users']) < 100
+			&& $this->appManager->isEnabledForUser('user_status')) {
```
nickvergessen, line 848: appmanager is already in the roomcontroller to check for circles, just in case someone is wondering

georgehrke (approved):
Code looks good 👍

kesselb:
:disappointed:

PR #4165, opened 2020-09-16
Be more robust against failures when trying to get the owner of the s…

### LINKED ISSUE #3381: Path for room attachement doesn't work
Tried with Talk 9.0.0-dev.5 on c.nc.com. I have set the path for attachments to "/Talk Attachments" but attachments are still added to my root.

nickvergessen:
…hare we are trying to mount

If we fail to get the owner of the view from the cache,
e.g. because the user never logged in but a cron job runs
We fallback to calculating the owner from the root of the view:

Fix #3381 

### Steps

1. Create a conversation and invite users
2. Upload an attachment
3. Without opening the conversations for other users, run the `OCA\Files\BackgroundJob\ScanFiles` and see 2 errors being logged:

Level | App | Message
-- | -- | --
Fatal | files_sharing | Error while trying to create shared mount
Error | files_sharing | OCP\Files\NotFoundException: / not found while trying to get owner

With this patch :heavy_check_mark:

### DIFF
--- a/lib/Share/Listener.php
+++ b/lib/Share/Listener.php
@@ -79,7 +79,23 @@ class Listener {
 		}
 
 		if ($event->getParent() === RoomShareProvider::TALK_FOLDER_PLACEHOLDER) {
-			$parent = $this->config->getAttachmentFolder($view->getOwner('/'));
+			try {
+				$userId = $view->getOwner('/');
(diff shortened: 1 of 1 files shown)


MorrisJobke (approved):
Looks good 👍

PR #4187, opened 2020-09-21
Cache the last msg id of a chat in MemoryCache to save O(n) queries p…

### LINKED ISSUE #4184: Save last message id in redis
Basically https://github.com/nextcloud/spreed/blob/7b3f1f37a2f231b8a50d7eb2eee039affdfc33c1/lib/Chat/ChatManager.php#L280-L287 could check if the last message in redis is still the same as the requested one.

If not, it runs the actual query.

Just have to make sure to also clear the value when a message is posted or created.

nickvergessen:
…er second

Before n queries per second when n is the number of active users in a chat
After 1 query per 30 seconds per chat

Fix #4184


danxuliu (approved):
Tested and works :+1:

PR #4189, opened 2020-09-21
Less participant queries

nickvergessen:
Should reduce the number of those queries a lot:

```sql
SELECT * FROM `oc_talk_participants` WHERE (`user_id` = 'xxx') AND (`room_id` = '1234')
```

and will instead left join the participant on the initial query. That in itself is not a big help. But in case we ask for the same participant more than once (hint, happens every time), we save N-1 queries

it was 1122 queries of the 7873 in our query log


rullzer (approved):
Noice!

danxuliu (changes requested):
Unit tests need to be adjusted.

danxuliu (approved):
Tested and works :+1:

PR #4190, opened 2020-09-21
If there are no sessions, we don't need to run a query

nickvergessen:
Should remove all those queries from our query log:

```sql
SELECT * FROM `oc_talk_guests` WHERE `session_hash` IN (NULL)
```

Called from
https://github.com/nextcloud/spreed/blob/291e2c75f7b97cd8c1194b5a54cbf6b97193e981/lib/Controller/RoomController.php#L877-L881


rullzer (approved):
Nice catch

PR #4196, opened 2020-09-22
Bugfix/noid/matterbridge

nickvergessen:


lib/MatterbridgeManager.php
```diff
@@ -666,28 +665,51 @@ private function launchMatterbridge(Room $room): int {
 
 	/**
 	 * kill the mattermost processes (owned by web server unix user) that do not match with any room
+	 * @param bool $killAll
 	 */
-	public function killZombieBridges(): void {
+	public function killZombieBridges(bool $killAll = false): void {
```
eneiluj, line 670: This could be named `killBridges` and have a `$zombieOnly` boolean parameter. This would make it less confusive when it's called from `stopAllBridges`.

lib/MatterbridgeManager.php
```diff
@@ -728,19 +750,26 @@ private function isRunning(int $pid): bool {
...
-		$this->manager->forAllRooms(function ($room) {
-			if ($room->getType() === Room::GROUP_CALL || $room->getType() === Room::PUBLIC_CALL) {
-				$bridge = $this->getBridgeOfRoom($room);
-				// disable bridge in stored config
+		$query = $this->db->getQueryBuilder();
+
+		$query->select('*')
+			->from('talk_bridges')
+			->where($query->expr()->like('json_values', $query->createNamedParameter(
+				$this->db->escapeLikeParameter('{"enabled":true') . '%'
```
eneiluj, line 758: Yeah this would be much prettier if `enabled` and `pid` were DB columns.

eneiluj (approved):
All good. Non blocking remarks. Thanks a lot.

lib/MatterbridgeManager.php
```diff
@@ -754,17 +756,17 @@ public function stopAllBridges(): bool {
...
-			if ($bridge['enabled']) {
-				$bridge['enabled'] = false;
-				$this->saveBridgeToDb((int) $row['room_id'], $bridge);
-			}
+			$bridge = [
+				'enabled' => true,
+				'pid' => (int) $row['pid'],
+				'parts' => json_decode($row['json_values'], true),
+			];
+			$bridge['enabled'] = false;
```
nickvergessen, line 768:
```suggestion
			$bridge = [
				'enabled' => false,
				'pid' => (int) $row['pid'],
				'parts' => json_decode($row['json_values'], true),
			];
```

lib/Migration/Version11000Date20200922161218.php
```diff
@@ -0,0 +1,114 @@
...
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options) {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable('talk_bridges')) {
+			$table = $schema->getTable('talk_bridges');
+			if (!$table->hasColumn('enabled')) {
+				$table->addColumn('enabled', Type::INTEGER, [
```
nickvergessen, line 58:
```suggestion
				$table->addColumn('enabled', Type::SMALLINT, [
```

eneiluj (approved):
Yep.

nickvergessen:
Need to backport because the background job is broken and the performance after fixing it is too bad.

PR #4198, opened 2020-09-22
Change the upload button name from upload to send

GretaD:
I think its more clear if we name the upload button at the second step of sending a file to chat as send, instead of upload, since it give you the impression you will upload a new file and not send it to the chat.

uploadbutton

### DIFF
--- a/src/components/UploadEditor.vue
+++ b/src/components/UploadEditor.vue
@@ -54,10 +54,10 @@
 		</transition-group>
 		<div class="upload-editor__actions">
 			<button @click="handleDismiss">
-				Dismiss
+				{{ t('spreed', 'Dismiss') }}
 			</button>
 			<button class="primary" @click="handleUpload">
-				Send
+				{{ t('spreed', 'Send') }}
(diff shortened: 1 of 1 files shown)


src/components/UploadEditor.vue
```diff
@@ -57,7 +57,7 @@
 				Dismiss
 			</button>
 			<button class="primary" @click="handleUpload">
-				Upload
+				Send
```
nickvergessen, line 60:
```suggestion
				{{ t('spreed', 'Send') }}
```

🙈 I hope "Send" is already translated

nickvergessen (changes requested):

src/components/UploadEditor.vue
```diff
@@ -57,7 +57,7 @@
 				Dismiss
```
nickvergessen, line 57: Also misisng translation here

nickvergessen (changes requested):

nickvergessen (approved):

PR #4200, opened 2020-09-22
Save the "count unread messages" query, when the last read is the las…

nickvergessen:
…t room messages


rullzer (approved):
:rocket:

PR #4209, opened 2020-09-23
Fix update query

nickvergessen:


danxuliu (approved):
Tested and works :+1:

PR #4214, opened 2020-09-23
Set the user status when a mobile client polls for messages

nickvergessen:


Ivansss (approved):
Tested and works 👍

PR #4231, opened 2020-09-25
✔️✔️ Read status for chat messages

### LINKED ISSUE #1012: Indicators for Sent/Received/Read text messages
Hi NC devs !!

I logged a Feature request (https://github.com/nextcloud/talk-android/issues/219) on Android Talk repo but I was told there's a prerequisite on the server side, hence this new issue :-)

Could the server enable Talk clients to get status indicators for message : Sent/Received/Read ?
The aim is for Clients to be able to get the same features as Signal/Whatsapp have.

Thanks.

---

Android: https://github.com/nextcloud/talk-android/issues/219

nickvergessen:
- [x] Add capability for the read-privacy
- [x] Add API to allow setting the read-privacy
- [x] Expose the current setting value on capability
- [x] Add "last common read message" to conversation API
- [x] Add "last common read message" header to all chat requests
- [x] Add integration test for the setting
- [x] Basic UI checks so @ma12-co can do the polishing
- [ ] Beautiful UI by @ma12-co 

Fix #1012

### DIFF
--- a/lib/Service/ParticipantService.php
+++ b/lib/Service/ParticipantService.php
@@ -245,7 +245,7 @@ class ParticipantService {
 		foreach ($participants as $participant) {
 			$readPrivacy = Participant::PRIVACY_PUBLIC;
 			if ($participant['actorType'] === Attendee::ACTOR_USERS) {
-				$readPrivacy = (int) $this->serverConfig->getUserValue($participant['actorId'], 'spreed', 'read_status_privacy', (string) Participant::PRIVACY_PUBLIC);
+				$readPrivacy = $this->talkConfig->getUserReadPrivacy($participant['actorId']);
 			}
 
 			$attendee = new Attendee();
--- a/lib/Service/ParticipantService.php
+++ b/lib/Service/ParticipantService.php
@@ -47,12 +47,15 @@ use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\Comments\IComment;
 use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\EventDispatcher\IEventDispatcher;
+use OCP\IConfig;
 use OCP\IDBConnection;
 use OCP\IUser;
 use OCP\IUserManager;
 use OCP\Security\ISecureRandom;
 
 class ParticipantService {
+	/** @var IConfig */
+	protected $serverConfig;
 	/** @var Config */
 	protected $talkConfig;
 	/** @var AttendeeMapper */
@@ -72,7 +75,8 @@ class ParticipantService {
 	/** @var ITimeFactory */
 	private $timeFactory;
 
-	public function __construct(Config $talkConfig,
+	public function __construct(IConfig $serverConfig,
+								Config $talkConfig,
 								AttendeeMapper $attendeeMapper,
 								SessionMapper $sessionMapper,
 								SessionService $sessionService,
@@ -81,6 +85,7 @@ class ParticipantService {
 								IEventDispatcher $dispatcher,
 								IUserManager $userManager,
 								ITimeFactory $timeFactory) {
+		$this->serverConfig = $serverConfig;
 		$this->talkConfig = $talkConfig;
 		$this->attendeeMapper = $attendeeMapper;
 		$this->sessionMapper = $sessionMapper;
@@ -238,12 +243,18 @@ class ParticipantService {
 		}
 
 		foreach ($participants as $participant) {
+			$readPrivacy = Participant::PRIVACY_PUBLIC;
+			if ($participant['actorType'] === Attendee::ACTOR_USERS) {
+				$readPrivacy = (int) $this->serverConfig->getUserValue($participant['actorId'], 'spreed', 'read_status_privacy', (string) Participant::PRIVACY_PUBLIC);
+			}
+
 			$attendee = new Attendee();
 			$attendee->setRoomId($room->getId());
 			$attendee->setActorType($participant['actorType']);
 			$attendee->setActorId($participant['actorId']);
 			$attendee->setParticipantType($participant['participantType'] ?? Participant::USER);
 			$attendee->setLastReadMessage($lastMessage);
+			$attendee->setReadPrivacy($readPrivacy);
 			$this->attendeeMapper->insert($attendee);
 		}
 
@@ -468,6 +479,15 @@ class ParticipantService {
 			->where($query->expr()->eq('room_id', $query->createNamedParameter($room->getId(), IQueryBuilder::PARAM_INT)))
 			->andWhere($query->expr()->eq('actor_type', $query->createNamedParameter(Attendee::ACTOR_USERS)))
 			->andWhere($query->expr()->in('actor_id', $query->createNamedParameter($userIds, IQueryBuilder::PARAM_STR_ARRAY)));
+		$query->execute();
+	}
+
+	public function updateReadPrivacyForActor(string $actorType, string $actorId, int $readPrivacy): void {
+		$query = $this->connection->getQueryBuilder();
+		$query->update('talk_attendees')
+			->set('read_privacy', $query->createNamedParameter($readPrivacy, IQueryBuilder::PARAM_INT))
+			->where($query->expr()->eq('actor_type', $query->createNamedParameter($actorType)))
+			->andWhere($query->expr()->eq('actor_id', $query->createNamedParameter($actorId)));
 		$query->execute();
 	}
 
--- a/lib/Service/ParticipantService.php
+++ b/lib/Service/ParticipantService.php
@@ -491,6 +491,46 @@ class ParticipantService {
 		$query->execute();
 	}
 
+	public function getLastCommonReadChatMessage(Room $room): int {
+		$query = $this->connection->getQueryBuilder();
+		$query->selectAlias($query->func()->min('last_read_message'), 'last_common_read_message')
+			->from('talk_attendees')
+			->where($query->expr()->eq('actor_type', $query->createNamedParameter(Attendee::ACTOR_USERS)))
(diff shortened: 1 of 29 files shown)


nickvergessen:
How do we want to handle the status?

## Pre-thought
There is a privacy setting for the read status. If it is set to private, the reading state of this person is totally ignored and they can also not see any read-status.

## Current

We store the "last read message id" per attendee in a conversation.

Current "message states"
Visual | Status
---|---
⭕️ (loading circle) | Message has been "submit" and is currently transferred to the server and waiting for the request to finish.
12:34 (timestamp) | Message was received by the server and is published

## Original proposal
The proposal was:
> The screenshot shows the message delivery status indicator. A simple check mark means that the message was delivered to the server, a check mark in a circle means that the message was seen by the recipient. If the message was not yet send to the server a spinning circle will be shown

## Thoughts

### One-to-one

For one-to-one chats this basically means (pseudo queries):
```sql
SELECT last_read_message FROM oc_talk_attendees WHERE room_id = X AND actor_type = 'users' AND actor_id != YOU;
```

### Groups
For a group chat the question is:
* Does the state change when:
    1. One participant read it
    2. All participants read it

Telegram e.g. uses double checkmarks if someone read it, Whatsapp uses blue checkmarks when everyone read it, but it has a list of user->time to check individually who read it.

We pick `2.` here, so the query is.
```sql
SELECT MIN(last_read_message) FROM oc_talk_attendees WHERE room_id = X AND actor_type = 'users' AND actor_id != YOU GROUP BY room_id;
```

If the result is `null` it means no one read it => `✓`
If the result is an ID it means everything older and this message are  => `✓✓` everything afterwards `✓`

Visual | Status
---|---
⭕️ (loading circle) | Message has been "submit" and is currently transferred to the server and waiting for the request to finish.
12:34 (timestamp) | Message was received by the server and is published
12:34 ✓ | Message was sent successfully and it could or could not be read by some users
12:34 ✓✓ | Message was read by everyone

nickvergessen:
Technical question: How do we update the read status. Responding to the long polling request doesn't sound like a good idea, because that will make spam requests all the time when you write something in a busy chat, but if we just add it as a header on all those requests, it means the read status could be delayed in updating by <= 30 seconds.

Similar question applies to "other conversations". If we should show the read status in the conversation list, I would query the info there and just udpate it every 30seconds. It's also in line with current scenarios for a reply there, so that sounds good?

ma12-co:
Result of coworking with @nickvergessen on this :) 

Frame 1asd

nickvergessen:
Stage 1 ready for review. UI will be polished by @ma12-co  in a follow up.

ma12-co (commented):
> Beautiful UI by @ma12-co 
:rofl:

nickvergessen:
Had forgotten to do the fix ups. done now.

lib/Service/ParticipantService.php
```diff
@@ -471,6 +482,49 @@ public function markUsersAsMentioned(Room $room, array $userIds, int $messageId)
...
+			->andWhere($query->expr()->eq('read_privacy', $query->createNamedParameter(Participant::PRIVACY_PUBLIC, IQueryBuilder::PARAM_INT)));
+
+		$result = $query->execute();
+		$row = $result->fetch();
+		$result->closeCursor();
+
+		return (int) ($row['last_common_read_message'] ?? 0);
+	}
+
+	public function getLastCommonReadChatMessageForMultipleRooms(array $roomIds): array {
```
PVince81, line 509: PHPDoc would be nice, especially when the return value is not a plain indexed array

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -59,6 +59,16 @@ the main body of the message as well as a quote.
 			<div class="message__main__right">
 				<div v-if="isTemporary && !isTemporaryUpload" class="icon-loading-small" />
 				<span v-if="hasDate" v-tooltip.auto="messageDate">{{ messageTime }}</span>
+				<CheckAll
```
PVince81, line 62: would be good to add a v-tooltip and also an aria-label for these icons

src/store/settingsStore.js
```diff
@@ -0,0 +1,58 @@
...
+const actions = {
+
+	/**
+	 * Updates the token
+	 *
+	 * @param {object} context default store context;
+	 * @param {string} privacy The new selected privacy
+	 */
+	updateReadStatusPrivacy(context, privacy) {
+		context.commit('updateReadStatusPrivacy', privacy)
```
PVince81, line 54: the store's action (make it async) should call the service to save the action, and when it's done, commit the result

tests/integration/features/chat/read-status.feature
```diff
@@ -0,0 +1,35 @@
+Feature: public
+  Background:
+    Given user "participant1" exists
+    Given user "participant2" exists
+
+  Scenario: User toggles the read privacy
```
PVince81, line 6: would be nice to also add a scenario with multiple users with "public" setting and demonstrate the "last common read message id", which would validate the "min()" function.

or is this already somehow covered in the second part of this scenario ?
a separate scenario could make it clearer that it is about the "common read", not directly about the "privacy flag"

PVince81 (commented):
See comments on the code. Haven't tested so far.

src/components/MessagesList/MessagesGroup/Message/Message.vue
nickvergessen: https://github.com/nextcloud/spreed/pull/4231#issuecomment-741777019

> UI will be polished by @ma12-co in a follow up.

We should get the API in asap, so mobile apps can start their development and maybe finish before xmas

lib/Service/ParticipantService.php
nickvergessen: it's a map of roomid => messageid, but sure will add

tests/integration/features/chat/read-status.feature
nickvergessen: yes it is, as it always considers your and the other persons read status. let me try to make it explicit

src/components/SettingsDialog/SettingsDialog.vue
```diff
@@ -167,6 +187,17 @@ export default {
 				})
 		},
 
+		async toggleReadStatusPrivacy() {
+			try {
+				await this.$store.commit(
```
PVince81, line 192: this should be a dispatch of that action, then the action does the commit.
I don't think this code works as it would update the value directly without going through the action

PVince81 (changes requested):

PVince81:
UX note: while testing this I intuitively went to the personal settings, didn't think of going to the talk specific cog icon @ma12-co

src/components/SettingsDialog/SettingsDialog.vue
PVince81: also: would be good to disable the checkbox while saving.
then when saving completed, display a toast notification "Your privacy settings were saved" which inspires confidence.

we used the same pattern for conversation settings dialog

otherwise when I click the checkbox I'm leaving wondering whether it's saved and every time I find myself scrolling down the dialog to find the save button.

=> could be done separately

PVince81: let me adjust this, quick change

PVince81: https://github.com/nextcloud/spreed/pull/4231/commits/bd9aead8c1e1eb94e665c3ec44279757a21ebcfa

nickvergessen: Ah yeah it worked before as the action was basically skipped it still worked. But now the backend request is missing 🙈

PVince81:
More comments from my tests (could be done separatly as part of improving UX):

- [ ] hide the indicator just like we do for the time part (not sure why we even hide it)

image

PVince81:
> Technical question: How do we update the read status. Responding to the long polling request doesn't sound like a good idea, because that will make spam requests all the time when you write something in a busy chat, but if we just add it as a header on all those requests, it means the read status could be delayed in updating by <= 30 seconds.
> 
> Similar question applies to "other conversations". If we should show the read status in the conversation list, I would query the info there and just udpate it every 30seconds. It's also in line with current scenarios for a reply there, so that sounds good?

What was the decision ? Leaving the current logic as is ?

During a call, I found it interesting that it makes it possible to see who doesn't have the chat panel open. Maybe in the future we could always collapse the chat panel (with localstorage preference) and only open it when there's a new message. Or pop a notification for the message to prompt people to open it. Or something.
Having it collapsed by default would then reduce the load as well.

PVince81:
Another remark: the read notification applies to everyone, not only people in the call. Not sure if someone might expect to get a double mark as soon as people within the call read it, minus external ones... Something to discuss separately I guess.

PVince81 (approved):
Fine to merge considering that UX will be worked on after that :+1:

nickvergessen:
> What was the decision ? Leaving the current logic as is ?

Yeah the "on every chat request" which is a max of 30s is good enough for now.

> During a call, I found it interesting that it makes it possible to see who doesn't have the chat panel open.

You can't see who read it. Just if everyone read it (in stage 1 at least)

> Another remark: the read notification applies to everyone, not only people in the call. Not sure if someone might expect to get a double mark as soon as people within the call read it, minus external ones... Something to discuss separately I guess.

It applies to all users (not guests) everything else would be totally unexpected.

PVince81:
raised https://github.com/nextcloud/spreed/issues/4738 for the UX bits

maximelehericy:
> Result of coworking with @nickvergessen on this :)
> 
> Frame 1asd

@marcoambrosini @nickvergessen is there plans to pursue the development of step 2 (providing details when people read messages) ?
Thanks for your help 🙏

nickvergessen:
Not at the moment

PR #4238, opened 2020-09-26
l10n: Change the sign of the text string

Valdnet:


src/components/SettingsDialog/SettingsDialog.vue
```diff
@@ -152,7 +152,7 @@ export default {
 				.build()
 			picker.pick()
 				.then(async(path) => {
-					console.debug(`Path '${path}' selected for talk attachments`)
+					console.debug('Path '${path}' selected for talk attachments')
```
nickvergessen, line 155: This breaks the log that is generated as path is not evaluated anymore

nickvergessen (changes requested):

src/components/SettingsDialog/SettingsDialog.vue
Valdnet: I corrected :wink:.

PR #4264, opened 2020-10-02
Fix participants list search box overlap

PVince81:
When scrolling up, the participants are not visible any more above the
search box.

image

This is a minimal quick-fix.

A proper way might need to reconsider the structure of the panel and move the scrolling container to start below the search box (vertically).

### DIFF
--- a/src/components/RightSidebar/RightSidebar.vue
+++ b/src/components/RightSidebar/RightSidebar.vue
@@ -258,6 +258,16 @@ export default {
 	display: flex;
 }
 
+/**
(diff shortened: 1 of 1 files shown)


ma12-co (approved):

skjnldsv:
A little reminder for you @ma12-co and @nickvergessen. you should not override components styling. 
If it's there and not fitting your needs it's either because this was not designed for it or might need an adjustment for everyone in the components.
The outline is an accessibility feature and is here on purpose. :)

PR #4266, opened 2020-10-02
Focus message field after upload or sharing file

PVince81:
After starting an upload or sharing a file, set the focus back to the message field.

- [x] ~~focus happens before the upload dialog is closed~~ => not an issue
- [x] ~~BUG: when inserting emoji, cursor always appears at the beginning~~ => abandoning emoji approach for now until the [new rich input field](https://github.com/nextcloud/nextcloud-vue/pull/1433) is there

### Tests
- [x] TEST: upload file with the button, see that focus is back on the message field
- [x] TEST: upload file with drag and drop, see that focus is back on the message field
- [x] TEST: share a file, see that focus is back on the message field


ma12-co (approved):

PVince81:
@nickvergessen is a backport of this interesting as it's a minor UX addition ?

PR #4287, opened 2020-10-05
Import dialogs scss

### LINKED ISSUE #4281: Add generic Chromium support
Nextcloud Talk tells me that my browser is not supported, even that it of course, as it's Opera, based on Chromium and works nicely with all aspects of the app. Worse, the banner makes the generally wrong information true by hiding parts of the apps UI, making them unusable:

I think the blue hidden button at the top right there was to toggle the participants list, but I am unable to toggle it now 😢.

image

- Support all Chromium-based browsers in a generic way, which, aside of Opera, would include Vivaldi, Brave, Iron and some others that deserve more attention instead of feel like being forced to use one of the big four...
- Add a little `[x]` hide button to hide the notification, probably even remember that to not bother users again with it.

- If it's insisted to only support original Chrome and Chromium Edge as Chromium-based browsers and wanted to keep showing this to users on every access, at least let's make it a notification behind the bell instead of a banner that blocks parts of the UI, the banner does not make it's own message become true.

**Additional context**
Many thanks for the great conference and value speeches and presentations btw! I hope it's okay that parts of the conversation is included in the screen above, right after Julius Härtl finished great demonstrations of Nextcloud Flow and Dashboard widget development 👏.

nickvergessen:
Fix #4281

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -24,11 +24,11 @@ module.exports = {
 		rules: [
 			{
 				test: /\.css$/,
-				use: ['vue-style-loader', 'css-loader'],
+				use: ['style-loader', 'css-loader'],
 			},
 			{
 				test: /\.scss$/,
-				use: ['vue-style-loader', 'css-loader', 'sass-loader'],
+				use: ['style-loader', 'css-loader', 'sass-loader'],
 			},
 			{
 				test: /\.(js|vue)$/,
@@ -79,10 +79,7 @@ module.exports = {
 			},
 			{
 				test: /\.(png|jpg|gif|svg)$/,
-				loader: 'file-loader',
-				options: {
-					name: '[name].[ext]?[hash]',
-				},
+				loader: 'url-loader',
 			},
 		],
 	},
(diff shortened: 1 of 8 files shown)


webpack.common.js
```diff
@@ -79,10 +79,7 @@ module.exports = {
 			},
 			{
 				test: /\.(png|jpg|gif|svg)$/,
-				loader: 'file-loader',
-				options: {
-					name: '[name].[ext]?[hash]',
-				},
+				loader: 'url-loader',
```
nickvergessen, line 82: changes are copied from https://github.com/nextcloud/webpack-vue-config/blob/master/webpack.js

skjnldsv: You can't use the webpack-vue-config directly?

nickvergessen: well this needs backporting so lets do it step for step

skjnldsv (approved):
sure then!
Love the fact that you didn't even had the `vue-style-loader` as a dependency ^^'

ma12-co (approved):

Issue #4284, opened 2020-10-05
Flickering background gradient when call speaker changes


PVince81:
## Steps to reproduce
1. Create a few users with avatars set
2. Have a call with 3-4 people A, B, C, D, everyone with video disabled
3. Have person A unmute and talk, then mute again
4. Have person B unmute and talk, then mute again
5. Repeat 2-3 and have person C observe the view

Whenever a person speaks, their view (avatar + background) is moved to the main view.

### Expected behaviour

The blurry/gradient background should appear directly

### Actual behaviour

When the speaker's view appear on top, first there is some generic background, and after a short delay the blurry/gradient background appears. When it happens frequently, it becomes annoying.

## Talk app

**Talk app version:** 10.0.0

## Browser

**Microphone available:** yes

**Camera available:** yes

**Operating system:** openSUSE

**Browser name:** Firefox

**Browser version:** 80.0

**Nextcloud configuration:** cloud.nextcloud.com

PVince81:
updated to mention that avatars need to be set, as it's the blur generated based on the pic that is delayed

there might be ways to cache this by hiding the picture element off-screen... and deleting in case of resize

PVince81:
A quicker way to reproduce this:

1. Have three participants, all with an avatar, participate in a call
1. Reduce horizontally the window until one participant's grid cell disappears
1. Then increase horizontal size again to make it reappear, then observe

When reappearing, the same effect occurs: first black, then the avatar appears, then blur.

The network console shows that the avatar gets re-requested. So maybe caching the avatar here would help.

PVince81:
PR here with quick fix: https://github.com/nextcloud/spreed/pull/4292

PVince81:
I looked into adding a fade animation whenever the blur+background and avatar are loaded. For the background alone it doesn't seem to work, no fade happens despite preloading.

And for detecting whenever the avatar is loaded, it needs changes on a higher level. Basically being able to know in advance what avatar sizes will be needed by the subcomponents of Vue's "video" element to be able to preload them. This doesn't feel right.

Furthermore I had the feeling that avatar caching in general doesn't work at all. I see network requests for the same avatar sizes repeated at least two times. One is fetching the "is this a picture" header, so the second call should be cached, but it still hits the server. We might want to first look into the avatar caching issue before smoothing the transitions.

- [ ] raise ticket to recheck avatar caching which seems to be not working
- [ ] raise ticket for smoother load experience, with fade in after a panel is loaded

[...]

PR #4292, opened 2020-10-05
Reuse video background / avatar during call

PVince81:
Whenever a call is running, if the presenter changes we now reuse the
video background and avatar elements to avoid flicker due to reloading.

This is achieved by using v-show instead of v-if for the specified
elements.

Note: when resizing the window if participants disappear and you make them appear again, their avatar is reloaded. I think this is ok. The mini-flicker also happens on first load. Most important was to fix periodic flickering during presenter change when multiple people are speaking.

Fixes https://github.com/nextcloud/spreed/issues/4284

### DIFF
--- a/src/components/CallView/CallView.vue
+++ b/src/components/CallView/CallView.vue
@@ -27,7 +27,7 @@
 			<div v-if="showPromoted" ref="videoContainer" class="video__promoted autopilot">
 				<template v-for="callParticipantModel in reversedCallParticipantModels">
 					<Video
-						v-if="sharedDatas[callParticipantModel.attributes.peerId].promoted"
+						v-show="sharedDatas[callParticipantModel.attributes.peerId].promoted"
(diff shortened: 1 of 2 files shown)


PVince81:
- [ ] BUG: somehow managed to have missing blur when adding a 4th participant and fiddling. I saw setBlur called with 0, 0 and never again, and it seems that resizing the window doesn't trigger the resize observer ?!

- [ ] check potential perf issue: number of Vue "video" components will increase as a lot are hidden (so far was not able to debug further as laptop went too slow after 4 participants, could not inspect the tree...)

PVince81:
I've found a potential alternative approach here https://github.com/nextcloud/spreed/issues/4284#issuecomment-707122567 that requires to know in advance whether a user has an avatar or not as that request is causing delays, even when read from cache.

PVince81:
obsoleted by https://github.com/nextcloud/spreed/issues/4284#issuecomment-709340940

PR #4296, opened 2020-10-05
Remove pulse animation from call button

PVince81:
It causes unnecessary load on Chrome browser.

Discovered and confirmed by @danxuliu

### DIFF
--- a/src/components/TopBar/CallButton.vue
+++ b/src/components/TopBar/CallButton.vue
@@ -235,10 +235,6 @@ export default {
 	.icon {
 		opacity: 1;
 		margin-right: 8px;
-
-		&.icon-incoming-call {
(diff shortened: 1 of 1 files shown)


danxuliu (approved):
Tested (also in Firefox) and works :+1: 

I will miss the animation, I liked it :-P

PVince81:
squashed. thanks for deleting the other bit!

PVince81:
I had another quick try, this is my chromium gpu state:
image

I tried enabling hardware accelerated rasterization.
I also tried to set "position: absolute" on the animated icon and various other hacks.

But no luck: still too much CPU used.

PVince81:
@nickvergessen backports to minor as it's a quick fix ?

Issue #1647, opened 2019-03-20
Add an UI option for "read-only" conversations


nickvergessen:
With #1640 conversations can be marked read-only.
This is currently only available via the APIs, since it was needed as a base for #1616 

We should add an option for moderators to allow toggling the state in the callinfoview?
Maybe a (un)lock-icon next  to the conversation name or the copy-link/make public/private checkbox?

Also the change event somehow does not toggle the editable-state of the chat-input.

@nextcloud/designers opinions?

skjnldsv:
We should be careful not to make the user confused about the password protection and the locking state of the conversation. For read-only I'm not sure I would put an icon. Putting everything without any interaction is pretty self-explanatory :thinking: 
Though the title of the conversation could have `(read-only)` in it ?

nickvergessen:
No, we may use another icon maybe, and its pretty obvious when its locked, as chat input and call button are disabled. Also a system message is disabled and you can revert it easily.

jancborchardt:
Yep – I also think an icon is not needed here and would possibly be confusing. Appending (read-only) to the title is also a possibility, not sure if needed though.

> and its pretty obvious when its locked, as chat input and call button are disabled.

@nickvergessen exactly. :+1:  In addition I’d say the chat input placeholder should say "Conversation is read-only for you" so it’s 100% obvious (or something similar if you find something shorter so it also works better on mobile).

Supergamerx3000:
how can i use this function now?
can I make channels read-only via a command?

nickvergessen:
No, via the API or directly in the database

PVince81:
seems this discussion is old and might refer to an older version of the UI as I didn't find any "callinfoview".

I agree that it doesn't need to be a clickable icon as this kind of action is not that frequent.

Here's a proposal based on the master version of Talk:

1) for moderators only, add lock/unlock action in the three dots menu behind the conversation name on the left sidebar
![image](https://user-images.githubusercontent.com/277525/94912026-53fbea00-04a7-11eb-8d5c-fe011c804b26.png)

2) when locked, the action says "Unlock conversation" (also moderators only)

3) when locked, the conversation title has "(read-only)" appended in a faded color:
   - in the left conversation list
   - in the top of the sidebar panel on the right

4) New message form appears but is disabled with placeholder "Conversation is read-only"


As I understand the conversation is locked for everyone ? Or can moderators still post messages ?

nickvergessen:
[...]

Issue #3886, opened 2020-07-08
🛂 Chat permission


jakobroehrl:
On a big talk group is often a lot of noise, so maybe it's helpful to restrict posting to moderators

nickvergessen:
Closing this in favor of #383 or do we close #383 as "done" since #4331 is merged and your PR #4487 basically closes this here?

jakobroehrl:
@nickvergessen 
This issue is about locking a conversation EXCEPT the moderators, so IMHO it should be open, maybe I can implement it for NC22?

But we can close https://github.com/nextcloud/spreed/issues/383 
I don't know if I want to continue with https://github.com/nextcloud/spreed/pull/4487

xbluemonkx:
posting here to get updates on this ticket. we would like to have that feature too.

marcoambrosini:
UI-wise:

- Input is grayed out with a "no permissions message"
- No reaction symbol to add new reactions (you can still see existing reactions from other ppl) 
- Attachment button disabled
- Forwarding messages to the conversation with no chat permissions should be blocked too

nickvergessen:
Closing as it's implemented on all clients

xbluemonkx:
is this already available in talk 14?

PR #4331, opened 2020-10-08
Add checkbox for locking conversation

PVince81:
Added logic for setting a conversation to read-only.

Fixes https://github.com/nextcloud/spreed/issues/1647

### Todos

- [x] read-only switch
- [x] proper error handling when sending a message and it doesn't go through: put it back
- [x] hide "Start call" button when read-only
- [x] hide "NewMessageForm" when read-only or leave visible but disabled ?
- [x] remove "Reply" buttons in messages list
- [ ] add "(read-only)" text in conversation title
   - to be discussed as one might prefer to call it "(locked)", "(archived)" or whatever and the suffix interferes with renaming. maybe lock icon still better

### Manual Tests
- [x] check that read-only checkbox not visible for non-moderators
- [x] test removed actions:
   - [x] start call
   - [x] reply
   - [x] upload by drag and drop
- [x] check that guests get the same read-only experience

### Automated Tests
- [x] check what scenarios could be added => seems read-only backend integration tests already exist

### DIFF
--- a/src/components/ChatView.vue
+++ b/src/components/ChatView.vue
@@ -44,7 +44,7 @@
 		</transition>
 		<MessagesList
 			:token="token" />
-		<NewMessageForm v-if="!isReadOnly" />
+		<NewMessageForm />
 	</div>
 </template>
 
--- a/src/components/ChatView.vue
+++ b/src/components/ChatView.vue
@@ -44,7 +44,7 @@
 		</transition>
 		<MessagesList
 			:token="token" />
-		<NewMessageForm />
+		<NewMessageForm v-if="!isReadOnly" />
 	</div>
 </template>
 
@@ -84,7 +84,7 @@ export default {
 			if (this.isGuest) {
 				return t('spreed', 'You need to be logged in to upload files')
 			} else if (this.isReadOnly) {
-				return t('spreed', 'This conversation is read only')
+				return t('spreed', 'This conversation is read-only')
 			} else {
 				return t('spreed', 'Drop your files to upload')
 			}
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -59,7 +59,7 @@ the main body of the message as well as a quote.
 					{{ messageTime }}
 				</h6>
 				<Actions
-					v-show="showActions && hasActions"
+					v-show="showActions && hasActions && !isConversationReadOnly"
 					class="message__main__right__actions">
 					<ActionButton
 						v-if="isReplyable"
@@ -86,7 +86,7 @@ import RichText from '@juliushaertl/vue-richtext'
 import Quote from '../../../Quote'
 import { EventBus } from '../../../../services/EventBus'
 import emojiRegex from 'emoji-regex'
-import { PARTICIPANT } from '../../../../constants'
+import { PARTICIPANT, CONVERSATION } from '../../../../constants'
 import moment from '@nextcloud/moment'
 
 export default {
@@ -218,6 +218,11 @@ export default {
 	computed: {
 		hasActions() {
 			return this.isReplyable
+		},
+
+		isConversationReadOnly() {
+			const conversation = this.$store.getters.conversation(this.token)
+			return conversation.readOnly === CONVERSATION.STATE.READ_ONLY
 		},
 
 		isSystemMessage() {
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -59,7 +59,7 @@ the main body of the message as well as a quote.
 					{{ messageTime }}
 				</h6>
 				<Actions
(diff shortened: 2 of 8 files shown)


src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -59,7 +59,7 @@ the main body of the message as well as a quote.
 					{{ messageTime }}
 				</h6>
 				<Actions
-					v-show="showActions && hasActions"
+					v-show="showActions && hasActions && !isConversationReadOnly"
```
nickvergessen, line 62: This should go into `isReplyable` instead. Other (future) actions might still be possible

src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -257,7 +257,28 @@ export default {
...
 				} catch (error) {
-					console.debug(`error while submitting message ${error}`)
+					let statusCode = null
+					console.debug(`error while submitting message ${error}`, error)
+					if (error.isAxiosError) {
+						statusCode = error.response.status
+					}
+					// 403 when room is read-only, 412 when switched to lobby mode
+					if (statusCode === 403 || statusCode === 412) {
+						OC.Notification.show(
```
nickvergessen, line 267:
```suggestion
						showError(
```

same below

src/components/MessagesList/MessagesGroup/Message/Message.vue
PVince81: adjusted, moved into the single action as attribute

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -220,6 +220,11 @@ export default {
 			return this.isReplyable
```
nickvergessen, line 220:
```suggestion
			return this.isReplyable && !this.isConversationReadOnly
```

src/components/ChatView.vue
```diff
@@ -44,7 +44,7 @@
 		</transition>
 		<MessagesList
 			:token="token" />
-		<NewMessageForm />
+		<NewMessageForm v-if="!isReadOnly" />
```
nickvergessen, line 47: I think @jancborchardt always wanted to keep the input form with a message that the conversation is locked? Or am I mixing things up now?

nickvergessen (changes requested):
See first comment

nickvergessen:
> Conflicting files
src/store/conversationsStore.js

src/components/ChatView.vue
PVince81: I tried it at first but was more complicated as it requires adding a disabled property to many elements, so I removed it.
Also, most of the time the bottom-most message in the converstion is "admin has locked conversation" or so, so it's already clear that it's locked (and no one can post after this message)

@jancborchardt thoughts ?

PVince81: this is how it looks currently, please note the "you locked the conversation" message:

image

@jancborchardt let me know if this is fine, otherwise I can look into adding a disabled field

PVince81:
adjusted based on review and conflict fixed

src/components/ChatView.vue
PVince81: as discussed in chat, we'll go for the following approach when locked:

- hide all buttons around the message form
- display a disabled input field with the text "This conversation has been locked"

nickvergessen: Ref https://github.blog/2014-06-09-locking-conversations/

PVince81:
adjusted, here is how it looks like now:
image

PVince81:
- [ ] BUG? seems settings tab is gone on sidebar

nickvergessen (approved):
:shark:

jakobroehrl:
@PVince81 @nickvergessen 
Wouldn't it be better letting moderators write in locked conversations?
see https://github.com/nextcloud/spreed/issues/3886

nickvergessen:
For now this is not what they are designed for

jakobroehrl:
OK, can I try to implement a read only setting for !moderators?

jakobroehrl:
I need the function to lock the conversation for not moderator users. 
Could I extend it?

Eg as a dropdown:
group is writeable for
 - all group users
 - only moderators
 - nobody

@PVince81 @nickvergessen What do you think?

nickvergessen:
Well the mobile clients would need to understand that too. So not possible for 11. We can see for 12

fenvarien:
For education scenarios this feature would be very important and helpful. A moderator (teacher) should be able to write in locked chats too. So you could create chats with information for students and parents which are locked and other ones, which are used for the ongoing communication between teachters, staff, students and so on.

jakobroehrl:
> For education scenarios this feature would be very important and helpful. A moderator (teacher) should be able to write in locked chats too. So you could create chats with information for students and parents which are locked and other ones, which are used for the ongoing communication between teachters, staff, students and so on.

Nobody can write on locked conversations. 
Have a look here: https://github.com/nextcloud/spreed/issues/3886

PR #4334, opened 2020-10-08
Update conversation name directly after call returns

PVince81:
After renaming a conversation, set the name back into the state for
immediate display instead of waiting for the next signal/event many
seconds later.

I fixed this because it annoyed me to not see the change immediately, made it look broken.

Maybe this implementation is too naive and there is more complexity on the server side than I saw ?
So far I didn't understand the difference between "name", "displayName" and "roomName", but "displayName" is what gets displayed. Are there other scenarios I need to understand and test ?


nickvergessen:
name is the user id in case of 1-to-1 conversations and used to be e.g. empty when you didnt set any name (but we prevent this now), while the display name is the group name you set, the user display name in 1-to-1 conversations and used to be the "list of participant displaynames" in case the group name was empty.

So that should be good enough, I cant remember an intentionally reason why it shouldn't be immediate like this.

nickvergessen (approved):

PR #4336, opened 2020-10-09
Add Psalm

nickvergessen:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,9 +2,14 @@
 	"scripts": {
 		"cs:check": "php-cs-fixer fix --dry-run --diff",
 		"cs:fix": "php-cs-fixer fix",
-		"lint": "find . -name \\*.php -not -path './vendor/*' -not -path './build/*' -not -path './tests/integration/vendor/*' -print0 | xargs -0 -n1 php -l"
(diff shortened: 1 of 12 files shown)


composer.json
```diff
@@ -2,9 +2,14 @@
...
+		"lint": "find . -name \\*.php -not -path './vendor/*' -not -path './build/*' -not -path './tests/integration/vendor/*' -print0 | xargs -0 -n1 php -l",
+		"psalm": "psalm",
+		"psalm:fix": "psalm --alter --issues=InvalidReturnType,InvalidNullableReturnType,MissingParamType,InvalidFalsableReturnType"
 	},
+	"minimum-stability": "dev",
 	"require-dev": {
-		"nextcloud/coding-standard": "^0.3.0"
+		"nextcloud/coding-standard": "^0.3.0",
+		"vimeo/psalm": "^3.16",
+		"christophwurst/nextcloud": "dev-master"
```
ChristophWurst, line 13: you can drop it here and only install on CI

nickvergessen: but then one can not run it locally?
Or is it recomment to install this and psalm locally manually?

ChristophWurst: Yeah, right. You can leave it then. Or install the OCP package only on demand when you want to psalm locally. I'd keep psalm definitely in composer.json as that locks the version.

PVince81:
seems it fails now on master https://github.com/nextcloud/spreed/issues/4404

Issue #4192, opened 2020-09-21
make scroll to bottom button gray


ma12-co:
set max width on connection quality warnings

nickvergessen:
Is that maybe fixed by importing the toast.scss  in  #4287

PR #4358, opened 2020-10-13
Max width on call view tooltips

ma12-co:
Fix: #4192

### DIFF
--- a/src/components/CallView/shared/LocalMediaControls.vue
+++ b/src/components/CallView/shared/LocalMediaControls.vue
@@ -590,9 +590,10 @@ export default {
 }
 
 .hint {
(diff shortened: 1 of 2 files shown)


src/components/CallView/shared/LocalMediaControls.vue
```diff
@@ -590,9 +590,10 @@ export default {
 }
 
 .hint {
-	padding: 4px;
+	padding: 12px;
+	max-width: 300px;
```
nickvergessen, line 594: do we face the same issue with the "please use the speaker view" popover and others?

nickvergessen (approved):

Issue #4245, opened 2020-09-29
Add option to hide video thumbnails of users while someone shares the screen


croessner:
**Is your feature request related to a problem? Please describe.**
If you have several people in a video call and one shares the screen, there is no option to only have showing the shared screen. So the shared screen is very small. It would be nice, if you could hide the videos from other users while watching the shared screen. Like a dock that can be hidden.

**Describe the solution you'd like**
It would be nice to have a toggle that allows having the shared screen in "real" full screen and hiding all other elements.

**Describe alternatives you've considered**
I searched if there exists some hidden occ command for that or options for config.php, but had no luck. Customizing the talk "view" would be really nice :-)

**Additional context**
none

kingfisher77:
That would be perfect for us either!

nickvergessen:
I thought an issue for this would exist already for a long time. It's definetly on our todo list at least.

PVince81:
maybe could be an additional option between "grid view", "speaker view" and maybe add "full speaker view" there ?

since this would add a third mode it might be a good idea to make it a dropdown

nickvergessen:
No, there should just be a small arrow on top/over the video strip to allow folding it away.

nickvergessen:
No, there should just be a small arrow on top/over the video strip to allow folding it away.

PVince81:
Hiding the bottom bar also means hiding one own's avatar/video, so the microphone, camera and other options would become invisible.

In that case we could move them at the bottom-center of the full view in that mode.

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/4363

PVince81:
I've raised https://github.com/nextcloud/spreed/issues/4441 for the next step: automatic collapse whenever a screen share starts

majkinetor:
I am still having this.

Interesting, you can zoom OUT the page in browser to make shared screen bigger and avatars very small.

I am currently blocking this via uBlock Origin and rule

```
! 2021-02-23 https://cloud.nil.rs
cloud.nil.rs##.avatar-container
```

Microfon and stuff can still be accessed via other menus.

PR #4363, opened 2020-10-13
Add ability to collapse the grid view

PVince81:
To be able to focus on the video / screen of the current presenter.

Fixes https://github.com/nextcloud/spreed/issues/4245

### Todos

- [x] Fix design/positioning of button
- [x] Move microphone/video/etc actions to the bottom of the screen when the grid is collapsed
- [x] Test multiple call modes to make sure it looks fine
- [x] Test with sidebar view - blocked by https://github.com/nextcloud/server/issues/23445

### DIFF
--- a/src/assets/variables.scss
+++ b/src/assets/variables.scss
@@ -101,3 +101,26 @@ $chat-line-height: 1.6em;
 		transition: all 150ms ease-in-out;
 	}
 }
+
+.slide-down {
+	&-enter {
+		transform: translateY(50%);
+		opacity: 0;
+	}
+	&-enter-to {
+		transform: translateY(0);
+		opacity: 1;
+	}
+	&-leave {
+		transform: translateY(0);
+		opacity: 1;
+	}
+	&-leave-to {
+		transform: translateY(50%);
+		opacity: 0;
+	}
+	&-enter-active,
+	&-leave-active {
+		transition: all 150ms ease-in-out;
+	}
+}
--- a/src/assets/variables.scss
+++ b/src/assets/variables.scss
@@ -122,5 +122,7 @@ $chat-line-height: 1.6em;
 	&-enter-active,
 	&-leave-active {
 		transition: all 150ms ease-in-out;
+		/* force top container to resize during animation */
+		position: absolute !important;
 	}
 }
--- a/src/components/CallView/CallView.vue
+++ b/src/components/CallView/CallView.vue
@@ -92,6 +92,7 @@
 			<Grid
 				v-if="showGrid"
 				v-bind="$attrs"
+				:is-sidebar="isSidebar"
 				:is-stripe="!isGrid"
 				:token="token"
 				:fit-video="true"
--- a/src/components/CallView/CallView.vue
+++ b/src/components/CallView/CallView.vue
@@ -545,6 +545,7 @@ export default {
 	width: 100%;
 	height: 100%;
 	top: 0;
+	overflow: hidden;
 	display: -webkit-box;
 	display: -moz-box;
 	display: -ms-flexbox;
--- a/src/components/CallView/CallView.vue
+++ b/src/components/CallView/CallView.vue
@@ -23,6 +23,12 @@
 	<div id="call-container">
 		<EmptyCallView v-if="!remoteParticipantsCount && !screenSharingActive && !isGrid" />
 		<div id="videos">
+			<LocalMediaControls
+				class="local-media-controls"
+				:model="localMediaModel"
+				:local-call-participant-model="localCallParticipantModel"
+				:screen-sharing-button-hidden="isSidebar"
+				@switch-screen-to-id="$emit('switchScreenToId', $event)" />
 			<!-- Promoted "autopilot" mode -->
 			<div v-if="showPromoted" ref="videoContainer" class="video__promoted autopilot">
 				<template v-for="callParticipantModel in reversedCallParticipantModels">
@@ -92,7 +98,6 @@
 			<Grid
 				v-if="showGrid"
 				v-bind="$attrs"
-				:is-sidebar="isSidebar"
 				:is-stripe="!isGrid"
 				:token="token"
 				:fit-video="true"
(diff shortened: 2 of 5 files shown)


PVince81:
After talking with @ma12-co we decided to keep my PR and I've taken a similar approach from https://github.com/nextcloud/spreed/pull/4365 and added (and fixed) the slide animation.

Next up is extracting the local call controls to a higher layer to make them always visible and available even when collapsed.

PVince81:
- [ ] sort out isBig attribute and expected logic
- [ ] wire up other missing attributes

PVince81:
current state when open:
image

and collapsed:
image

PVince81:
most stuff sorted out, but one thing remaining:

- [ ] network quality warning broken when grid is collapsed: quickest would be to move it back to the media controls location as it creates an extra temporary icon.

I do wonder if there is another kind of interaction we could use to show that warning and not clutter that bottom right corner with a bubble.

PVince81:
next challenge: 
image

PVince81:
ok, it will look like this:
image

PVince81:
- [ ] BUG: missing mute and video buttons in sidebar mode

(tested using this PR merged with https://github.com/nextcloud/spreed/pull/4291)

PVince81:
I've fixed the sidebar mode. It required me to move the LocalMediaControls from the Grid to the CallView.

I've also tested the "network connection warning" mode and it looks fine in sidebar mode.

PVince81:
image

ma12-co (changes requested):
shouldn't this transition be applied to the .grid-main-wrapper div?
At the moment the transition feels a bit weird, it seems like .grid-main-wrapper comes in without transition and its content slides up in it

ma12-co (commented):
* missing hover feedback, I'd say we should have it like the buttons in the topbar;
* when active, the button has white background;
* Missing height and width of 44px;

Very nice work :+1:

PVince81:
- [ ] BUG: grid view video sizes are not correct / layout borken
- [ ] BUG: once grid view fixed, the bottom right video will not always have 300px width, so need to find a way to adjust
- [ ] TODO: https://github.com/nextcloud/spreed/pull/4363#pullrequestreview-509422426

PVince81:
- [ ] BUG: when "waiting for others to join" the stripe mode seems engaged, it shouldn't

PVince81:
- [ ] should we make one to one mode also collapsible ?
    - keep the chevron button but don't use stripe mode

ma12-co:
>  should we make one to one mode also collapsible ?

Yes, I'd say it's a desirable thing to have, especially when the other participant is screen-sharing

PVince81:
grid view is fixed.
in that mode now LocalMediaControls are present inside the LocalVideo. otherwise it uses the outside instance.

next up: one on one view and the remaining bits.

For one on one I'll see if I can reuse the grid/stripe view but with its left part disabled

ma12-co:
> next up: one on one view and the remaining bits.

Do you want to do it in this pr of follow-up?

PVince81:
@ma12-co already done. I want to address all checkboxes within this PR.

Feel free to have a quick design review.

- [ ] BUG: "waiting for others" has black background

PVince81:
- [ ] retest sidebar mode

PVince81:
button styles fixed now

PVince81:
for the transition I tried different containements but the top part didn't animate while the bottom part slid down.
I think this is because we used `translate` so it would only translate the stripe's absolute position without affecting the page's layout. Need to look into maybe animating another property.

PVince81:
Seems it might be possible to temporarily set `position: absolute; bottom 0;` on the ".wrapper" element while running the transition. This forces the top panel to expand directly before we start the slide down.

I'll study transitions more in detail to find out where exactly to put that as my trials did not succeed yet.

PVince81:
I've fixed the slide animation now with a small hack. I didn't want to reorganize all the containers for now.
We could look into the latter separately.

Tests collapsing:
- [x] TEST: "waiting for others"
- [x] TEST: one to one
- [x] TEST: one to many
- [x] TEST: grid view still works (and no collapse button there)
- [x] TEST: test with screen share on then collapse
- [x] TEST: sidebar mode still works as before

ma12-co (commented):
I get this error when switching to grid: 
Screenshot from 2020-10-19 13-05-18

PVince81:
when testing I had the latest nextcloud-vue master included.
retrying with the latest spreed master

PVince81:
Got it, the exact steps are:

1. Start a call with three people
1. Collapse the strip
1. Switch to grid view
1. Switch back to speaker view

The reason is that `isStripe()' is abused to re-generate the grid in `makeGrid`, and the latter function is trying to access the grid object while it's not visible. I'll try adding a check there to avoid calling `makeGrid` when collapsed.

Too many states :-S

PVince81:
@ma12-co I've fixed the issue when switching back and forth to grid while collapsed

ma12-co (approved):
Amazing :) 
One last cool thing would be a tooltip that shows up on hover that says something like "Hide videos" and "Show videos"

PVince81:
it's not videos if you only see avatars there, so not sure about the text.
maybe we can look into tooltips separately as I noticed there are more tooltips missing in the call view

danxuliu:
Since this pull request when the call view is shown in the sidebar the media controls are a bit off from the video (I have tested with nextcloud/server@fd178b9^ and also rebasing #4291 and using it against nextcloud/server@766590d204^, so it should not be a problem of an outdated style from server):

**Before:**
Files-Sidebar-Controls-Before

**After:**
Files-Sidebar-Controls-After

Is that intended, or known?

PVince81:
I didn't observe this before. Can you make a ticket and provide steps to reproduce and maybe also window size ?
With all the different tests I made I didn't see that.

The reason is that now the controls are not part of the video box any more, they are outside. But how did you manage to have the video box smaller than usual ?

ma12-co:
We really need to implement tests for the call view. There are so many scenarios that I's becoming impossible to test all of them  whenever someone changes something there

PVince81:
I've raised https://github.com/nextcloud/spreed/issues/4440 for adding automated tests for the frontend

PVince81:
@ma12-co while we might be able to test the code and logic, we can't test layout and styling stuff, so if a background color looks off or a container is not stretched correctly, we'd still need some manual testing to find that out. However we wouldn't need to retest all the logic.

PR #4374, opened 2020-10-15
Only show the conflict warning when the session was in a call

### LINKED ISSUE #4366: Better duplicate session UX
I find this warning annoying and most of the time useless. I always end up clicking "Join here" anyway.
```
Duplicate session
You are trying to join a conversation while having an active session in another window or device. 
This is currently not supported by Nextcloud Talk. What do you want to do?
```

So this is my proposal to improve this:

let's say that:
* I have a tab open somewhere and I joined **conversation A** in that **tab 1**;
* I have talk open in **tab 2** and i click on **conversation A** there too;

#### User is not participating to a call in **tab 1**
=> Just join the conversation in  **tab 2**, without any warning and display the "You joined the conversation in another window or device." notice in **tab 1**;
We could also change the title of **tab 1** into `duplicate session` so that if things get messy one can easily see at a glance which tabs are safe to close.

#### User had previously joined a call in **tab 1**
=> Show a warning that says that there's a call undergoing in **conversation A** in some other tab, and joining here would interrupt the call. 

Do you see any drawbacks with this approach?  

cc @nextcloud/designers

nickvergessen:
Fix #4366

### DIFF
--- a/lib/Controller/RoomController.php
+++ b/lib/Controller/RoomController.php
@@ -1229,8 +1229,8 @@ class RoomController extends AEnvironmentAwareController {
 		}
 
 		if ($previousSession instanceof Participant && $previousSession->getSessionId() !== '0') {
-			// Previous session was active
-			if ($force === false) {
+			if ($force === false && $previousSession->getInCallFlags() !== Participant::FLAG_DISCONNECTED) {
+				// Previous session was active in the call, show a warning
 				return new DataResponse([
 					'sessionId' => $previousSession->getSessionId(),
 					'inCall' => $previousSession->getInCallFlags(),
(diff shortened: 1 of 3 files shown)


ma12-co (changes requested):
Can we update the warning message and change the title of the tab?

nickvergessen:
Any suggestions?

ma12-co:
Warning: "You've joined a call in this conversation on another device, if you join here, the call will be interrupted."
Title of the "logged out" tab: "Duplicate session"

nickvergessen:
Added the page title as that is already translated. Let's do the update of the other message after the maintenance release ?

ma12-co (approved):

PR #4379, opened 2020-10-15
Cache hasAvatar to avoid call background flickering

PVince81:
Whenever the speaker changes, the avatar background can flicker due to
HTTP calls. This fix improves it a bit by caching the information
whether the user has an avatar or not, so allows a less delayed loading of
the avatar background.

This only affects the avatar background.
For the avatar itself, you also need https://github.com/nextcloud/nextcloud-vue/pull/1457 in place.
Best is to test both together.

### DIFF
--- a/src/components/CallView/shared/VideoBackground.vue
+++ b/src/components/CallView/shared/VideoBackground.vue
@@ -47,6 +47,19 @@ import usernameToColor from '@nextcloud/vue/dist/Functions/usernameToColor'
 import { generateUrl } from '@nextcloud/router'
 import { ResizeObserver } from 'vue-resize'
 
+// note: this info is shared with the Avatar component
+function getUserHasAvatar(userId) {
+	const flag = window.sessionStorage.getItem('userHasAvatar-' + userId)
+	if (typeof flag === 'string') {
+		return Boolean(flag)
+	}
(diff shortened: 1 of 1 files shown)


ma12-co (commented):
I get this artifact both when switching videos manually and when doing it by activating/deactivating the mics of other users with fake streams on, thereby changing the current promoted person: 

Peek 2020-10-15 20-14

PVince81:
@ma12-co this PR is only about the background and that one looks stable/fine in your video.

For the avatar's "white circle" please also check out https://github.com/nextcloud/nextcloud-vue/pull/1457.

Make sure to repeat the operations multiple times: the first time you'll still see artifacts and after the cache is populated, switching should not show any glitch (except for "stop following" which I observed locally)

PVince81:
Adjusted to use browser storage. However I'm not using the `BrowserStorage` service as it's in the "talk" namespace.
Need to use the "nextcloud" namespace as this is also where the Avatar vue stores the same state: https://github.com/nextcloud/nextcloud-vue/pull/1457

ma12-co (approved):

PVince81:
backport to stable20 here https://github.com/nextcloud/spreed/pull/4514

Issue #4378, opened 2020-10-15
Messagelist doesn't scroll properly


ma12-co:
## Steps to reproduce
1. Be in conversation 1
2. Hit refresh
3. Click on a conv with no new messages

### Expected behavior
The chat should scroll to the bottom 

### Actual behaviour
The scroller scrolls to 50% height

nickvergessen:
> So, when the bug happens, I have 2 requests for the history taking place
> 
> got it locally now :)
>
> okay, so basically when you enter the conversation, the view is empty, and then when the items are added, you are for a quick moment at the top of the list, so another history grab is called
>
> so yeah, thats it
> 
> it then triggers https://github.com/nextcloud/spreed/blob/515cfe8c8fbb68e711f002cf487501861775b95a/src/components/MessagesList/MessagesList.vue#L544-L550
> 
> so basically it would be enough to hardcode https://github.com/nextcloud/spreed/blob/515cfe8c8fbb68e711f002cf487501861775b95a/src/components/MessagesList/MessagesList.vue#L378 to only scroll down, no matter what

PR #4380, opened 2020-10-15
Fix chat scrolling

ma12-co:
fix: #4378 
fix: #4377

Removes global smooth scrolling of the container and smooth scrolls it only when needed

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -55,7 +55,7 @@ get the messagesList array and loop through the list to generate the messages.
 			<button v-show="!isScrolledToBottom"
 				:aria-label="scrollToBottomAriaLabel"
 				class="scroll-to-bottom"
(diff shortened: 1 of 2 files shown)


PVince81:
how to verify ? https://github.com/nextcloud/spreed/issues/4377 doesn't have steps listed

ma12-co:
> how to verify ? #4377 doesn't have steps listed

I debugged this with Joas and we didn't really find a reliable way to reproduce it, but we're pretty sure those bugs were introduced when I made the whole dive smooth-scrolling. This pr reverts that and animates the scroll only when strictly needed.

nickvergessen:
So it happened quite often when your history is long enough (300+ messages). inserting the 200 items from the history is taking a bit so at some point it beams you to the top, thinking you scrolled, and therefor requests more history and when that is loaded you are set to the top of the old history (middle of the screen).

nickvergessen (approved):

PR #4405, opened 2020-10-19
Fix psalm warnings about missing class in listeners

PVince81:


lib/Chat/SystemMessage/Listener.php
```diff
@@ -246,7 +246,7 @@ public static function register(IEventDispatcher $dispatcher): void {
 			$room = $manager->getRoomByToken($share->getSharedWith());
 			$listener->sendSystemMessage($room, 'file_shared', ['share' => $share->getId()]);
 		};
-		$dispatcher->addListener('OCP\Share::postShare', $listener);
+		$dispatcher->addListener(\OCP\Share::class . '::' . 'postShare', $listener);
```
nickvergessen, line 249: The warning is wrong, it's not a class actually:
https://github.com/nextcloud/server/blob/d9015a8c94bfd71fe484618a06d276701d3bf9ff/lib/private/Share20/Manager.php#L812

We should advice psalm to ignore this line for now.

Issue #4388, opened 2020-10-16
Black background for "waiting for others"


PVince81:
## Steps to reproduce
1. Start a call and stay alone
2. Admire the message "waiting for others"

### Expected behaviour
Another less dark background, maybe grey like in the stripe ?

### Actual behaviour
Black background.
It's so black that it looks like a bug
<img width="472" alt="image" src="https://user-images.githubusercontent.com/277525/96263806-baaef680-0fc3-11eb-91ff-d9aa0406c320.png">
## Talk app

**Talk app version:** 10.0.0 (from git v10.0.0)

@ma12-co any ideas how to make it better ?

ma12-co:
Makes sense! Sure we could match the stripe selected video background :)

PVince81:
fix is here https://github.com/nextcloud/spreed/pull/4408

PR #4408, opened 2020-10-19
Fix empty call background color

PVince81:
Make it the same grey like the stripe background and adjust text color
to be readable.

Fixes https://github.com/nextcloud/spreed/issues/4388

image

### DIFF
--- a/src/components/CallView/shared/EmptyCallView.vue
+++ b/src/components/CallView/shared/EmptyCallView.vue
@@ -147,7 +147,7 @@ export default {
 }
 </script>
 
-<style lang="scss">
+<style lang="scss" scoped>
 
 .empty-call-view {
 	height: 100%;
@@ -157,7 +157,7 @@ export default {
 	flex-direction: column;
 	align-content: center;
 	justify-content: center;
-	color: var(--color-text-maxcontrast);
+	background-color: var(--color-loading-dark);
 	text-align: center;
 	.icon {
 		background-size: 64px;
@@ -168,5 +168,10 @@ export default {
 	button {
 		margin: 4px auto;
 	}
+
+	h2, p {
+		color: var(--color-primary-text);
+	}
 }
+
 </style>


ma12-co (changes requested):
Now that I see it I think that  it should be a darker gray and the text should stay white.
Maybe var(--color-loading-dark)??

PVince81:
do we have a darker grey somewhere in the presets or should I invent a value ?

ma12-co (approved):
:+1:

src/components/CallView/shared/EmptyCallView.vue
```diff
@@ -168,5 +168,10 @@ export default {
 	button {
 		margin: 4px auto;
 	}
+
+	h2, p {
+		color: var(--color-primary-text);
```
nickvergessen, line 173: You want white, not primary text, otherwise you have black text on dark grey background if your primary theming color is bright

PVince81: should I hardcode white here ? then maybe I need to hardcode the background color as well instead of `--color-loading-dark`, if we agree that this is not something we want to theme.

as far as I can see here, only primary is defined: https://github.com/nextcloud/nextcloud-vue/blob/master/styleguide/assets/variables.css but if one would want to invert the colors they'd need to override "color-background-dark" with a light color ? or do we need a new `primary-text-background` color ?

nickvergessen: We dont want the screen to be blue or something so we should use anything related to primary.
Also we dont want it to flip ot white with dark mode, so mostlikely the loading color is wrong too.

But can be tested easily, just search for "accessibility" in your dev nextcloud, turn on dark mode, search for theming and use "#ffffff" as primary color and then start a call.

Expected result is a dark screen with readable font color

nickvergessen: https://github.com/nextcloud/server/blob/master/apps/accessibility/css/dark.scss#L16
Dark mode makes loading bright

PVince81: it seems that whatever combination of variables I pick, it cannot look good in both dark mode and regular mode

so if we anyway don't want to invert the colors I'll hardcode white and dark grey in this view

PVince81: => https://github.com/nextcloud/spreed/pull/4416

Issue #4407, opened 2020-10-19
Mic/cam error toast persists, even when problem is resolved


ChristophWurst:
## Steps to reproduce
1. Select no mic nor cam
2. Join call
3. See "Error while accessing microphone & camera"
4. Pic the actual mic
5. Join call

### Expected behaviour
Toast disappears (after the call at the latest)

### Actual behaviour
Toast persists forever.

## Talk app

**Talk app version:** company instance

PVince81:
Yep, noticed that as well as I'm usually testing with permissions denied.
A delay would be nice.

There are other variations of the error message IIRC

PVince81:
second one is "Access to microphone & camera was denied"

PVince81:
This has annoyed me long enough, here's the fix https://github.com/nextcloud/spreed/pull/4409

PR #4409, opened 2020-10-19
Improve call error messages UX

PVince81:
Added default timeout instead of permanent for messages related to
denied permissions.

Fixes https://github.com/nextcloud/spreed/issues/4407

### DIFF
--- a/src/components/CallView/shared/LocalVideo.vue
+++ b/src/components/CallView/shared/LocalVideo.vue
@@ -122,6 +122,12 @@ export default {
 		},
 	},
 
+	data() {
+		return {
+			notificationHandle: null,
+		}
+	},
+
 	computed: {
 
 		videoContainerClass() {
@@ -208,15 +214,15 @@ export default {
 			handler: function(error) {
 				if (error) {
 					if (error.name === 'NotAllowedError') {
-						showError(t('spreed', 'Access to camera was denied'))
+						this.notificationHandle = showError(t('spreed', 'Access to camera was denied'))
 					} else if (error.name === 'NotReadableError' || error.name === 'AbortError') {
 						// when camera in use, Chrome gives NotReadableError, Firefox gives AbortError
-						showError(t('spreed', 'Error while accessing camera: it is likely in use by another program'), {
+						this.notificationHandle = showError(t('spreed', 'Error while accessing camera: it is likely in use by another program'), {
 							timeout: TOAST_PERMANENT_TIMEOUT,
 						})
 					} else {
 						console.error('Error while accessing camera: ', error.message, error.name)
-						showError(t('spreed', 'Error while accessing camera'), {
+						this.notificationHandle = showError(t('spreed', 'Error while accessing camera'), {
 							timeout: TOAST_PERMANENT_TIMEOUT,
 						})
 					}
@@ -232,6 +238,9 @@ export default {
 	},
 
 	destroyed() {
+		if (this.notificationHandle) {
+			this.notificationHandle.hideToast()
+		}
 		if (this.localCallParticipantModel) {
 			this.localCallParticipantModel.off('forcedMute', this._handleForcedMute)
 		}
--- a/src/components/CallView/shared/LocalVideo.vue
+++ b/src/components/CallView/shared/LocalVideo.vue
@@ -78,7 +78,7 @@ import SHA1 from 'crypto-js/sha1'
 import {
 	showError,
 	showInfo,
-	TOAST_DEFAULT_TIMEOUT,
+	TOAST_PERMANENT_TIMEOUT,
 } from '@nextcloud/dialogs'
 import video from '../../../mixins/video.js'
 import VideoBackground from './VideoBackground'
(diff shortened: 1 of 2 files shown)


ChristophWurst (approved):

ma12-co (approved):

src/utils/webrtc/webrtc.js
```diff
@@ -935,10 +938,11 @@ export default function initWebRTC(signaling, _callParticipantCollection, _local
 		} else {
 			message = t('spreed', 'Error while accessing microphone & camera')
 			console.error('Error while accessing microphone & camera: ', error.message || error.name)
+			timeout = TOAST_DEFAULT_TIMEOUT
 		}
 
 		showError(message, {
-			timeout: TOAST_PERMANENT_TIMEOUT,
+			timeout: timeout,
```
nickvergessen, line 945: I dont think this should go away after 7sec. But it should close when you make a successful connection. For that we can store the return of showError in a variable and call variable.hide() or hideToast() or what the function is called, when a second request was successful

src/components/CallView/shared/LocalVideo.vue
```diff
@@ -295,7 +295,7 @@ export default {
 			handler: function(localStreamVideoError) {
 				if (localStreamVideoError) {
 					showError(t('spreed', 'Error while accessing camera'), {
-						timeout: TOAST_PERMANENT_TIMEOUT,
+						timeout: TOAST_DEFAULT_TIMEOUT,
```
nickvergessen, line 298: Instead of hardcoding to default, remove the timeout. But also here I would prefer to not timeout and just remove it when it now works successfully

nickvergessen (changes requested):

src/components/CallView/shared/LocalVideo.vue
PVince81: Sadly I see this message repeatedly when the camera is disabled, this is why I wanted it to go away.
Would need to investigate more in detail to see why sometimes it says "camera was not allowed" and sometimes "Error while accessing camera". From what I remember, if I disallow the camera the first time after seeing the dialog, the error is specific to that. However, next time when I open the browser it will fail.

Needs further research

src/utils/webrtc/webrtc.js
PVince81: Might apply to some messages, yes, but not for "Access to microphone & camera was denied"

src/components/CallView/shared/LocalVideo.vue
nickvergessen: Well, `timeout: TOAST_DEFAULT_TIMEOUT,` is not needed, so just remove the option for now

src/utils/webrtc/webrtc.js
```diff
@@ -935,10 +938,11 @@ export default function initWebRTC(signaling, _callParticipantCollection, _local
 		} else {
 			message = t('spreed', 'Error while accessing microphone & camera')
 			console.error('Error while accessing microphone & camera: ', error.message || error.name)
+			timeout = TOAST_DEFAULT_TIMEOUT
```
nickvergessen, line 941: I think `Error while accessing microphone & camera` is rather important one? should be permanent?

src/utils/webrtc/webrtc.js
nickvergessen: > Might apply to some messages, yes, but not for "Access to microphone & camera was denied"

But even that should immediatelly go away when you leave the call and join again and this time accept it. So save the return of `showError` in a var and in the beginning of this method check if `"var"` is not empty and then call the hideToast method, like in
https://github.com/nextcloud/spreed/blob/bd78555cc31e653171ec68a1f87eafe2304d7bba/src/services/conversationsService.js#L37-L50

PVince81:
I did a local test with both Firefox and Chromium with different states, they give the same results:
- deny both: "Access to microphone & camera was denied"
- allow mic only: "Error while accessing camera"
- allow camera only: "Access to microphone & camera was denied"
- when camera busy in other app: "Error while accessing camera"
- camera unplugged / unavailable: no error displayed

Now I remember that when I test locally often times one browser is using the camera so the other is always showing "Error while accessing camera". I'm not sure if we can detect the case "camera already in use" vs other errors.

I'll leave the "Error while accessing camera" in permanent mode and see if we can easily make it go away when it becomes irrelevant like leaving the call (but the button for leaving the call is blocked visually by the notification...)

src/utils/webrtc/webrtc.js
PVince81: done

src/utils/webrtc/webrtc.js
PVince81: I've separated the error in more detailed ones. In case the error is unknown, it will stay permanent.

PVince81:
I've adjusted the PR to be slightly more precise for some messages. The timeout was adjusted for messages that I think can be auto-dismissed like permission ones.

ma12-co (approved):
:elephant:

PVince81:
Would be good to hear also from @nickvergessen or @danxuliu to see if the error detection is acceptable in this form.

src/components/CallView/shared/LocalVideo.vue
```diff
@@ -205,23 +205,40 @@ export default {
...
+						})
+					}
 				}
 			},
 		},
 
 	},
 
+	created() {
+		this.notificationHandle = null
```
nickvergessen, line 230: should be in `data() {` so it's always defined and tracked?

nickvergessen (changes requested):
Looks good apart from the missing init

src/components/CallView/shared/LocalVideo.vue
PVince81: I put it here as the Vue docs said it's in private space.
I'm fine moving it to data then

nickvergessen: Let's keep it consistent for now with the rest of our code.

PVince81:
adjusted (squashed) and retested

nickvergessen (approved):

PVince81:
some quick win here ^ (and it annoyed me again during testing on stable20)

PR #4439, opened 2020-10-22
Don't copy array for reverse iteration

PVince81:
Copying an array just for the sake of reverse iteration is a bit wasteful, especially with bigger arrays.

Tested by sending quickly multiple messages. They were all properly replaced in the list. (hope it's the correct test case)


nickvergessen:
@skjnldsv @juliushaertl can I have your expertise?

skjnldsv (approved):
Code make sense yes :)

PR #4449, opened 2020-10-23
Correctly leave the room not the call in integration tests

nickvergessen:


PVince81 (approved):
:+1: makes sense

Issue #3824, opened 2020-06-18
Remember grid view state


nickvergessen:
### Steps to reproduce
1. Join a call and select the grid view (should it be default?)
2. Share a screen
3. Unshare the screen

#### Expected behaviour
You are in the grid view

#### Actual behaviour
You are in promoted view

### Steps to reproduce
1. Join a call and select the grid view
2. Follow a video
3. Click "Unfollow"

#### Expected behaviour
You are in the grid view

#### Actual behaviour
You are in promoted view

### Steps to reproduce
1. Join a call and select the grid view
2. Quit the call
3. Reload
4. Join the call again

#### Expected behaviour
You are in the grid view

#### Actual behaviour
You are in promoted view

PVince81:
could use "BrowserStorage"

also:
- [ ] remember collapsed state ? (once https://github.com/nextcloud/spreed/pull/4365 is merged)

nickvergessen:
I think collapse should be done when a screenshare is coming  in, I would not remember this for now.

For grid, I think grid should be the default when in group calls.
I wouldn't really store it, or we need to store it on a per-conversation basis, because e.g. in one-to-one you mostlikely wouldn't want grid, but in group calls you would.

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/4451

PR #4451, opened 2020-10-23
Remember grid view state per call

PVince81:
Fixes https://github.com/nextcloud/spreed/issues/3824

Please note that joinCall and leaveCall are actions already defined in the participantsStore.
Nevertheless, vuex will call the action on both the stores! :smile:

### DIFF
--- a/src/store/callViewStore.js
+++ b/src/store/callViewStore.js
@@ -20,6 +20,11 @@
  *
  */
 
+import BrowserStorage from '../services/BrowserStorage'
+import {
+	CONVERSATION,
+} from '../constants'
+
 const state = {
 	isGrid: false,
 	selectedVideoPeerId: null,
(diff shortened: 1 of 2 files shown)


src/store/callViewStore.js
```diff
@@ -55,6 +57,19 @@ const actions = {
 	selectedVideoPeerId(context, value) {
 		context.commit('selectedVideoPeerId', value)
 	},
+
+	joinCall(context, { token }) {
+		let isGrid = BrowserStorage.getItem('callprefs-' + token + '-isgrid')
+		if (isGrid === null) {
+			// default to speaker view
+			isGrid = false
```
nickvergessen, line 65: please `true` for group conversations

PVince81:
@nickvergessen the default for group or public conversation is now grid. (I saw other places in the code where public was also considered a group conversation)

I've also adjusted the implementation to store the preference as soon as it's changed instead of upon leaving the call: because when leaving the call by closing the browser window it doesn't call that handler.

Please re-review

jancborchardt (approved):
Nice improvement! :) :+1: UX-wise, didn’t test

ma12-co (approved):

Issue #4376, opened 2020-10-15
Dashboard: Show favourite rooms at the top


jakobroehrl:
Sort it like in the talk room overview: first the favourites then the other rooms

nickvergessen:
Should only happen on the "No mentions" half-empty content view I guess?

cc @nextcloud/designers-talk

jakobroehrl:
> Should only happen on the "No mentions" half-empty content view I guess?

Exactly

nickvergessen:
https://github.com/nextcloud/spreed/blob/557de2c1842262fc40c5698fdcbc5c40c9cf0e47/src/views/Dashboard.vue#L166 needs to be extended to also check for the favorite status.

PR #4464, opened 2020-10-26
New dashboard: Unread and favourites conversations

jakobroehrl:
Tried to duplicate the code corresponding https://github.com/nextcloud/spreed/pull/3890, but my new dashboard isn't shown
Can anybody look over it?


nickvergessen:
Just fix the sorting in the existing dashboard as per https://github.com/nextcloud/spreed/issues/4376 ?
if there is no call/mention it would then show favorites and unread messages automatically?

jakobroehrl:
Hmm, I think you are right. 
Let's close this.

Issue #3746, opened 2020-06-04
Better preview of the shared files


tiotrom:
I would suggest to make the preview of images and videos, and also text documents, a bit bigger. Perhaps make them as wide as the chat window itself or at least 50% wide. Just a suggestion. Thanks!

jcastang:
Hello,

Tried a 50% live modification on browser. But it depends on the size of the image. 
After trying a couple of settings, 384px/384px seems a good value on most screen resolution. (1080p)

Maybe just a change in CSS ?

tiotrom:
What would be the proper css code to change it for all files and not only images? Also, isn't it that the perview is just a very small thumbnail image and not a good quality one?

jcastang:
> What would be the proper css code to change it for all files and not only images? Also, isn't it that the perview is just a very small thumbnail image and not a good quality one?

Do not really know. I just live-edited CSS using dev mode in FF. 
I found that 256x256 or 384x384 px are nicer than orignal 128x128px size.

As I am not involved in coding Talk, I donner know what IS the best way to mâle rhis change permanent.

PVince81:
PR here https://github.com/nextcloud/spreed/pull/4469 that makes the images wider

PR #4469, opened 2020-10-26
Chat previews now expanded with aspect ratio

PVince81:
Instead of cropping the previews in the conversation to a square, they
are now using a limited height but their width is computed based on
aspect ratio. This is done by passing different arguments to the
previews endpoint so it delivers a preview with the matching size.

Adjusted styles around the preview to make it look better.

Fixes https://github.com/nextcloud/spreed/issues/3746

This was mentionned in https://github.com/nextcloud/spreed/issues/1805#issuecomment-716571572

image

### Manual Tests

- [x] wide pictures
- [x] narrow pictures
- [x] loading icon looks fine
- [x] reduced preview when using reply in pic

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
@@ -162,9 +162,9 @@ export default {
 			}
 
 			const previewSize = Math.ceil(this.previewSize * window.devicePixelRatio)
-			return generateUrl('/core/preview?fileId={fileId}&x={width}&y={height}', {
+			// expand width but keep a max height
+			return generateUrl('/core/preview?fileId={fileId}&x=-1&y={height}&a=1', {
 				fileId: this.id,
-				width: previewSize,
 				height: previewSize,
 			})
 		},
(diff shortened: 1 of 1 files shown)


skjnldsv:
I'm fine with this, but this is up to @jancborchardt's call :)

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
```diff
@@ -290,14 +290,18 @@ export default {
 		object-fit: cover;
 	}
 
+	.loading {
+		display: inline-block;
+		height: 128px;
```
nickvergessen, line 295: I think we once also discussed to use more height, e.g. 300? cc @jancborchardt 
But not sure what the endpoint does if the image is actually smalled?

PVince81: I can check with it. otherwise there's a "fill=true" option which I'm not sure what it does, might fit

PVince81: I've tested with a picture of 50x50 and the preview provider doesn't make it bigger.
However the image tag will scale it up, so it will have the same height as the other pictures:
image

PVince81: okay I found the ticket with the decision to make it 384px: https://github.com/nextcloud/spreed/issues/3746#issuecomment-641979030

I'll adjust

PVince81: with new height:
image

and for a wide image:
image

PVince81:
closing in favor of https://github.com/nextcloud/spreed/pull/4472 which improves the same code further (and I'd like to avoid conflicts)

Issue #4471, opened 2020-10-26
Previews do not render for guests


PVince81:
## Steps to reproduce
1. Create a public room
2. Post a few pictures (jpg)
3. Copy the guest link
4. Log out
5. Open the link
6. Scroll to the pictures in the chat

### Expected behaviour
Previews rendered

### Actual behaviour
Previews not rendered in the chat, instead I just see an icon.
But I can click them and see them as "public link share" view.

## Talk app

**Talk app version:** 11.0.0-dev.2 (git master 02fb50c233ebf365f2bdf65b680b6bd663cd1b2a)

PVince81:
As far as I could see in FilePreviews.vue there is no code path to make it generate a public preview URL.

I'll see if I can handle it as part of https://github.com/nextcloud/spreed/issues/1805 where I'm touching that code anyway.

PVince81:
PR here https://github.com/nextcloud/spreed/pull/4472

Issue #3543, opened 2020-05-13
Rename the images you send by CTRL + V


Valdnet:
When sending files via drag'n'drop, I suggest changing the file name from e.g.: image (2) to the current date and time of sending (20200513_202433). In my opinion it will be more professional.

![obraz](https://user-images.githubusercontent.com/47037905/81850481-51c8db80-9558-11ea-8b1c-12f5126db9c5.png)

vix95:
Hi, I think that it's a good idea and it looks more pro as well. But this one should refer only for images by CTRL + V. The real image file name is ok for me, if it's dragged&dropped from Desktop for instance.

Valdnet:
> But this one should refer only for images pasted by CTRL + V.

I agree with you. After thorough checking of the sending of specific files, the names should be unchanged and a tag (2) added etc.

PVince81:
When uploading more than 4 files, the dialog already looks crowded and file names are ellipsized already.
It wouldn't look ideal to have a pencil icon there to rename inline, nor would it look nice to have another modal on top for renaming.

The layout of that upload dialog might need to be reworked a bit.

nickvergessen:
Just rename them with a timestamp

PVince81:
I think it depends on the purpose of the file name within a conversation.

Usually the file name is always displayed within the chat. Some people might want the name to be like a description of the picture itself, or hide the name completely and post an own description in the next message. If that name cannot be changed I still wonder whether we should display it at all then.

Another use of the file name would be to distinguish one file from others, for example in a series of photos or documents where one wants to download the correct/intended one.

Always automatically renaming the file names would break the above use cases.
This is why intuitively I thought we should allow renaming the files within that dialog so that the name appears correctly and also on disk. Note: renaming the file on disk after the fact does not rename it for other users, so it has to be done before upload (and implicit share)

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/4485

PR #4472, opened 2020-10-26
Improve previews for guests + GIF support

PVince81:
As this touched the same code, I've gathered them into a single PR.

### Improve preview cropping

Chat previews now expanded with aspect ratio
    
Instead of cropping the previews in the conversation to a square, they are now using a limited height but their width is computed based on aspect ratio. This is done by passing different arguments to the previews endpoint so it delivers a preview with the matching size.     

Also expanded image preview height to 384 as it looks better.

Fixes https://github.com/nextcloud/spreed/issues/3746

### Render GIFs directly

Use direct URL to render GIFs (animated!) when they have a small size.

The maximum size is configurable with `occ config:app:set spreed max-gif-size --value=X` and exposed as the capability `spreed.config.previews.max-gif-size`

Fixes https://github.com/nextcloud/spreed/issues/1805

### Add support for guest previews

Guest users can now directly see images in the chat, and small GIFs are animated as well!

Fixes https://github.com/nextcloud/spreed/issues/4471

### DIFF
--- a/lib/Chat/Parser/SystemMessage.php
+++ b/lib/Chat/Parser/SystemMessage.php
@@ -323,6 +323,7 @@ class SystemMessage {
 		$share = $this->shareProvider->getShareById($shareId);
 		$node = $share->getNode();
 		$name = $node->getName();
+		$size = $node->getSize();
 		$path = $name;
 
 		if (!$participant->isGuest()) {
@@ -349,6 +350,7 @@ class SystemMessage {
 					$fullPath = $userNode->getPath();
 					$pathSegments = explode('/', $fullPath, 4);
 					$name = $userNode->getName();
+					$size = $userNode->getSize();
 					$path = $pathSegments[3] ?? $path;
 				}
 			} else {
@@ -370,6 +372,7 @@ class SystemMessage {
 			'type' => 'file',
 			'id' => (string) $node->getId(),
 			'name' => $name,
+			'size' => $size,
 			'path' => $path,
 			'link' => $url,
 			'mimetype' => $node->getMimeType(),
--- a/package-lock.json
+++ b/package-lock.json
@@ -4919,6 +4919,7 @@
 					"resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-4.3.0.tgz",
 					"integrity": "sha512-zbB9rCJAT1rbjiVDb2hqKFHNYLxgtk8NURxZ3IZwD3F6NtxbXZQCnnSi1Lkx+IDohdPlFp222wVALIheZJQSEg==",
 					"dev": true,
+					"optional": true,
 					"requires": {
 						"color-convert": "^2.0.1"
 					}
@@ -4983,6 +4984,7 @@
 					"resolved": "https://registry.npmjs.org/color-convert/-/color-convert-2.0.1.tgz",
 					"integrity": "sha512-RRECPsj7iu/xb5oKYcsFHSppFNnsj/52OVTRKb4zP5onXwVF3zVmmToNcOfGC+CRDpfK/U584fMg38ZHCaElKQ==",
 					"dev": true,
+					"optional": true,
 					"requires": {
 						"color-name": "~1.1.4"
 					}
@@ -4991,7 +4993,8 @@
 					"version": "1.1.4",
 					"resolved": "https://registry.npmjs.org/color-name/-/color-name-1.1.4.tgz",
 					"integrity": "sha512-dOy+3AuW3a2wNbZHIuMZpTcgjGuLU/uBL/ubcZF9OXbDo8ff4O8yVp5Bf0efS8uEoYo5q4Fx7dY9OgQGXgAsQA==",
-					"dev": true
+					"dev": true,
+					"optional": true
 				},
 				"electron-to-chromium": {
 					"version": "1.3.582",
@@ -5009,7 +5012,8 @@
 					"version": "3.0.0",
 					"resolved": "https://registry.npmjs.org/emojis-list/-/emojis-list-3.0.0.tgz",
 					"integrity": "sha512-/kyM18EfinwXZbno9FyUGeFh87KC8HRQBQGildHZbEuRyWFOmv1U10o9BBp8XVZDVNNuQKyIGIu5ZYAAXJ0V2Q==",
-					"dev": true
+					"dev": true,
+					"optional": true
 				},
 				"escalade": {
 					"version": "3.1.1",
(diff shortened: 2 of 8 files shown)


PVince81:
- [ ] BUG: previews in upload dialog are too big now

PVince81:
- [x] BUG: need a max-width for when the preview is too wide, otherwise it pushes the increases chat width => I've added a max-width: 100% to the preview (squashed)

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
```diff
@@ -153,20 +159,47 @@ export default {
...
+			if (this.previewAvailable !== 'yes') {
 				return OC.MimeType.getIconUrl(this.mimetype)
 			}
 
+			// max size of a gif for which we allow direct embedding
+			let maxGifSize = 3145728
+			const caps = getCapabilities()
+			if (caps) {
+				maxGifSize = caps.spreed.config.previews['max-gif-size']
+			}
```
skjnldsv, line 176:
```suggestion
			const maxGifSize = getCapabilities()?.caps?.spreed?.config?.previews?.['max-gif-size'] || 3145728
```

PVince81: depends what hurts the eyes more :smile:

skjnldsv: You mean using kebab-case keys? :wink:

PVince81: no? I? mean? the? question? marks? :smile:

PVince81:
- [ ] squash after review

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
skjnldsv: It's the new fancy optional chaining :heart_eyes: 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

Love it!

skjnldsv (approved):
Works fine!
Maybe we should have width/height constrains, it can lead to really big images 
dev skjnldsv com_apps_spreed_not-found

PVince81:
@skjnldsv the height will always be the same. For the width I added max-width: 100% so it doesn't become wider than the chat window. Not sure if we need more.

Thanks for the approval, I'll squash.

lib/Chat/Parser/SystemMessage.php
```diff
@@ -370,6 +372,7 @@ protected function getFileFromShare(Participant $participant, string $shareId):
 			'type' => 'file',
 			'id' => (string) $node->getId(),
 			'name' => $name,
+			'size' => $size,
```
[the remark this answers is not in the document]
PVince81: pushed here https://github.com/nextcloud/server/pull/23712

this means no direct backporting of the GIF pieces. but we could backport the non-GIF improvements at least

jancborchardt:
Very nice @PVince81! :) Tiny enhancements which would make it look even better (but you decide whether to do it here or in a separate pull request):
- Give images a `border-radius: var(--border-radius)` so it looks more integrated
- Cut the image name from below since it looks technical, rarely has value and other chat solutions don’t show it either. Can still be shown as "title" attribute if we think it’s necessary.

PVince81:
> Cut the image name from below since it looks technical

@jancborchardt also see related discussion about renaming files. If we remove the display name then renaming the files is maybe not needed any more: https://github.com/nextcloud/spreed/issues/3543

PVince81:
- [ ] test image layout when conversation happens in sidebar

PVince81:
Note: to test sidebar mode need to merge with https://github.com/nextcloud/spreed/pull/4291 and upload the files through the regular talk view.

- [ ] BUG?: small pictures get stretched (ex: 50x50 pixels)
    - it seems this was already the case before but was less noticable
- [ ] BUG?: image is cropped in sidebar mode instead of having its size reduced: 
    - reduce size in sidebar mode
    - maybe we can use a smaller vertical height there instead

PVince81:
The issue was that sometimes the height is anyway the one we got from the preview, but sometimes when we render the real image like a GIF, we need max-height.

With the latest commit the image resizing and keeping the ratio is solved, both in the sidebar and also regular chat view.
Smaller pictures (like 50x50) are rendered small, which also means people could in theory upload small pictures of custom emojis if they wanted to without them being stretched...

@skjnldsv mind having another glance with your GIF collection and play with the resize ?

package-lock.json
```diff
@@ -4919,6 +4919,7 @@
 					"resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-4.3.0.tgz",
 					"integrity": "sha512-zbB9rCJAT1rbjiVDb2hqKFHNYLxgtk8NURxZ3IZwD3F6NtxbXZQCnnSi1Lkx+IDohdPlFp222wVALIheZJQSEg==",
 					"dev": true,
+					"optional": true,
```
nickvergessen, line 4922: what was it with the optional again to not get it in and out again all the time?

nickvergessen (approved):

skjnldsv:
Capture d’écran_2020-10-29_08-53-24
A weird display bug for the second (75Mo gif)

Otherwise works fine. Maybe we should add an indicator to why it wont auto play if it's over 3Mo

package-lock.json
PVince81: no idea, this got set automatically when installing the new package

nickvergessen:
> A weird display bug for the second (75Mo gif)

I see this too sometimes on our instance, I guess this happens when generating the preview takes too long or errors on first attempt or something. But it's a server issue

skjnldsv:
> > A weird display bug for the second (75Mo gif)
> 
> I see this too sometimes on our instance, I guess this happens when generating the preview takes too long or errors on first attempt or something. But it's a server issue
> 

But the mime icon is really small now, no?

nickvergessen:
Well the icon you see there is the placeholder icon in case of an error I think. because the mimetype icon of svg is the image icon.

PVince81:
here's a fix for the icon height: https://github.com/nextcloud/spreed/pull/4537

PVince81:
and here the play button to make it clearer that it can be played: https://github.com/nextcloud/spreed/pull/4538

PR #4195, opened 2020-09-22
Use new app settings components

ma12-co:
- [x] Requires https://github.com/nextcloud/nextcloud-vue/pull/1286

Issue #2946, opened 2020-02-13
scrolling issue (participants list)


schiessle:
Tested with Talk 8.0.3, in large rooms where you can scroll the participants list the user scroll through the "Add participants to the conversation" input field:

![image](https://user-images.githubusercontent.com/1589737/74441915-e136f380-4e70-11ea-87e3-d8249bd1ed74.png)

I would expect that the scroll are stops below the input field.

PVince81:
another "position: sticky" nightmare...

it seems it's not possible to increase the padding as the element will move down and still have some space at the top

making it "position: absolute" doesn't make the scroll work correctly, unless we add some top margin to make up for it.

maybe need to switch to another way of doing scrolling there

PVince81:
quickfix here: https://github.com/nextcloud/spreed/pull/4264

PVince81:
PR was merged, closing

PR #4473, opened 2020-10-27
Bump @nextcloud/vue to 3.1.0

PVince81:
This brings in:
- the new edit component (which we can replace later in https://github.com/nextcloud/spreed/pull/4333)
- Scrollable app sidebar (which is the reason for major bump)
- Avatar caching.
- Other small things

### Todos

- [x] cleanup sidebar css hacks
- [x] ~~potentially blocked by https://github.com/nextcloud/spreed/pull/4195 (issues mentioned there)~~ => https://github.com/nextcloud/spreed/pull/4473#issuecomment-717083734

### Tests

I did a quick test to see if all still works:
- [x] scrolling in left sidebar still works fine
- [x] scrolling in participant list works fine
- [x] sending (typing)/receiving messages still works, with emojis
- [x] remove hacks
- [x] check settings panel for regression
- [x] verify chat scrolling with lots of messages when conversation in sidebar
- [x] verify regular chat scrolling

Obsoletes https://github.com/nextcloud/spreed/pull/4457

### DIFF
--- a/src/components/RightSidebar/RightSidebar.vue
+++ b/src/components/RightSidebar/RightSidebar.vue
@@ -27,6 +27,7 @@
 		:title="title"
 		:starred="isFavorited"
 		:title-editable="canModerate && isRenamingConversation"
+		:class="'active-tab-' + activeTab"
 		@update:active="handleUpdateActive"
 		@update:starred="onFavoriteChange"
 		@update:title="handleUpdateTitle"
@@ -112,7 +113,7 @@ export default {
 
 	data() {
 		return {
-			activeTab: null,
+			activeTab: 'participants',
 			contactsLoading: false,
 			// The conversation name (while editing)
 			conversationName: '',
@@ -258,31 +259,6 @@ export default {
  * nextcloud-vue for ".app-sidebar". */
 #app-sidebar {
 	display: flex;
-}
-
-/**
- * Replace padding with margin to make the participant list scroll
- * properly behind the top field
- */
-#tab-participants {
-	margin-top: 10px;
-	padding-top: 0;
-	outline: none; /* remove the weird border that appears on focus */
-}
-
-/* Force scroll bars in tabs content instead of in whole sidebar. */
-::v-deep .app-sidebar-tabs {
-	height: calc(100% - 80px);
-
-	&__content {
-		overflow: hidden;
-
-		section {
-			height: 100%;
-
-			overflow-y: auto;
-		}
-	}
 }
 
 .app-sidebar-tabs__content #tab-chat {
--- a/package-lock.json
+++ b/package-lock.json
@@ -4107,24 +4107,29 @@
 			}
 		},
 		"@nextcloud/vue": {
-			"version": "2.7.0",
-			"resolved": "https://registry.npmjs.org/@nextcloud/vue/-/vue-2.7.0.tgz",
-			"integrity": "sha512-iZxTUWsDvfQdi3FRqEBfv5MGqJPPISk9FEXhX98wnWX37YaDwx9qOpvAQo+/6bjVVwwu8E7oWJN24NE4rxWhBg==",
+			"version": "3.1.0",
+			"resolved": "https://registry.npmjs.org/@nextcloud/vue/-/vue-3.1.0.tgz",
+			"integrity": "sha512-BB3/6RM0yPw4+McZvda8CaL0JEFb4lXQ/7l8tZ1zgrY2jDCZpwnEMsizzK16KeX4dzx1HNyt4PPBlWaSoVvfuQ==",
 			"requires": {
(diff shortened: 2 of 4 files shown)


skjnldsv (commented):
please also cleanup old css hacks
https://github.com/nextcloud/spreed/blob/fc1cbfce48c2cb3a4ca6a12bce422ee6f118a10b/src/components/RightSidebar/RightSidebar.vue#L273-L286

PVince81:
I had a quick look at https://github.com/nextcloud/spreed/pull/4195 and I don't see any difference before and after the Vue update, so I'd say it's not a blocker. The PR in question needs further work and might require further changes in nextcloud-vue, for which we can do future releases + version bumps.

PVince81:
@skjnldsv when I remove the whole block you mentioned, there's a double scrollbar again.

I compared the sidebar before and after the vue update, removing or adding the hack back, I saw no difference in behavior.
So I'm not sure how the new tab scrollbar approach is supposed to help in this situation.

skjnldsv:
> @skjnldsv when I remove the whole block you mentioned (specifically, the overflow:hidden), there's a double scrollbar again.

Yeah, it's because of https://github.com/nextcloud/spreed/blob/38d0870c18ddd1e54e0bc3cc48e03aeb2ca78b76/src/components/RightSidebar/RightSidebar.vue#L268-L270
You need to remove this as well.
Please do not remove the outline! This is an accessibility feature and is here on purpose.
Generally speaking you should not override components styling. If it's here, it's here on purpose and for all apps.

PVince81:
Indeed. so this will reopen https://github.com/nextcloud/spreed/issues/2946 and we need to find a different solution for it, maybe rework the tab contents layout. Ideally the scroll container should start below the field instead of having the field floating and overlapping. This means having a scroll container even deeper inside the tab content and not using the tab content's own scroller.

skjnldsv:
> Ideally the scroll container should start below the field instead of having the field floating and overlapping.

No need, just put the padding on the form, and it will have the background under it?
Position sticky is exactly why it was designed for :)

skjnldsv:
Also works perfectly fine for my by defaults

PVince81:
@skjnldsv easier said than done. I'll need some CSS (or even Vue) trickery to remove the padding on the parent ".app-sidebar-tabs__content" to add it back to the input field. But only do that when the participants tab is selected.

PVince81:
never mind, was looking at the wrong element...

src/components/RightSidebar/RightSidebar.vue
```diff
@@ -260,28 +260,11 @@ export default {
...
-	outline: none; /* remove the weird border that appears on focus */
-}
-
-/* Force scroll bars in tabs content instead of in whole sidebar. */
-::v-deep .app-sidebar-tabs {
-	height: calc(100% - 80px);
-
-	&__content {
-		overflow: hidden;
+	padding-top: 0px;
```
skjnldsv, line 264: I still don't see why that is necessary :thinking:

PVince81: after my PR:
image

if I remove padding-top: 0
image

so I need to compensate by removing the 8px from the content (the top green border)
image

PVince81: adjusted and squashed based on more hints from @skjnldsv, thanks!

PVince81:
I've verified the scrolling with lots of messages in the files app sidebar and it worked fine! (had to use https://github.com/nextcloud/spreed/pull/4291 for testing)

If CI is green I think we can move forward with this PR.

skjnldsv (approved):

src/components/RightSidebar/RightSidebar.vue
```diff
@@ -260,29 +261,14 @@ export default {
...
-	&__content {
-		overflow: hidden;
-
-		section {
-			height: 100%;
-
-			overflow-y: auto;
-		}
-	}
+::v-deep .app-navigation-search {
```
skjnldsv, line 268: You most likely don't need v-deep as the root component SearchBox is used in the ParticipantTab component and will share the same vue-scope-id :)
https://github.com/nextcloud/spreed/blob/d0094567770ef799f039c6f57c1d33ca19f6f60e/src/components/RightSidebar/Participants/ParticipantsTab.vue#L275

You can move this there

PVince81: adjusted and squashed

at that point I forgot that I had direct access and that it wasn't an external component

skjnldsv: Yes, it gets confusing very fast :p

PVince81:
I forgot to verify the sidebar chat during a call (I only checked in the files app).
It has a double scrollbar now, raised here https://github.com/nextcloud/spreed/issues/4492

Issue #4354, opened 2020-10-12
Polite call notification for private calls


ChristophWurst:
**Is your feature request related to a problem? Please describe.**

When I receive a notification about a private call I see "Joas wants to talk with you" and IMO this sounds a bit aggressive.

**Describe the solution you'd like**

A slightly better wording like "Joas would like to talk with you".

**Describe alternatives you've considered**

n/a

**Additional context**

https://github.com/nextcloud/spreed/blob/5bd5779d356907e9c7ab3de6ad4bbdecb5857c94/lib/Notification/Notifier.php#L566

PR #4324, opened 2020-10-08
↕️ Split participants into 👤 Attendees and ⏳ Sessions

nickvergessen:
The old participant (table) has been split into 2 sub components:

* Participant (consists of)
  * Attendee (always there):
    + Participant identifier (to map the session data)
    + Room id/token
    + Identifier: User id, group id, email-address, random-id for guests
      - Actor type: `users`, `emails`, `guests` (future: `groups` and `circles` to track them to automatically add/remove users)
      - Actor id: userid, email address, session hash, group id, circle id
    + Participant type
    + Attendee favorited the room
    + Selected notification level
    + PIN (new for the upcoming SIP #4496 )
    + Last mention comment
    + Last read comment
    + Last joined call (for the call summary)

  * Session (optional, currently still limited to 1 per attendee):
    + Attendee ID
    + SessionId - random string for WebRTC
    + In call flag
    + Last ping (is currently only used to timeout users from the call, therefor it does not require to be kept in the participant data)

The most important thing is that users are no longer identified by `userId || sessionId`, as they might not have a session and not be a logged in user. the new identifier is actor type+id

Also I started to get rid of the queries and modifying functions in the room object. Attendee and Session are "ordinary" `OCP\AppFramework\Db\Entity` objects and I hope the Room will be too at some point. This entries are modified internally via the respective `OCP\AppFramework\Db\QBMapper`, but access to those should only be done via the ParticipantService (only exception being the Manager which needs to be able to use `QBMapper::mapRowToEntity` to create the objects and I don't want to expose this function on the service.

Fix #2020

PR #4475, opened 2020-10-27
Politer language for call notifications for private calls 

jakobroehrl:
Fix https://github.com/nextcloud/spreed/issues/4354

### DIFF
--- a/lib/Notification/Notifier.php
+++ b/lib/Notification/Notifier.php
@@ -569,7 +569,7 @@ class Notifier implements INotifier {
 			if ($user instanceof IUser) {
 				if ($this->notificationManager->isPreparingPushNotification() || $this->participantService->hasActiveSessionsInCall($room)) {
 					$notification = $this->addActionButton($notification, $l->t('Answer call'));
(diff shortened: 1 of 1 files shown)


lib/Notification/Notifier.php
```diff
@@ -563,7 +563,7 @@ protected function parseCall(INotification $notification, Room $room, IL10N $l):
 			if ($user instanceof IUser) {
 				if ($this->notificationManager->isPreparingPushNotification() || $room->hasSessionsInCall()) {
```
nickvergessen, line 564: this will create a conflict with #4324 so lets delay this until the other one is in, to avoid tones of rebases

nickvergessen (approved):

lib/Notification/Notifier.php
PVince81: the PR in question was merged, no conflict has appeared :-D

nickvergessen: Yeah, it is 2 lines apart and that seems to be enough...

ChristophWurst:
:heart: :handshake: :smile:

Issue #4448, opened 2020-10-23
Quick multi user direct message 


jakobroehrl:
**Is your feature request related to a problem? Please describe.**
Often I use mail to reach more then user because it's felt harder to create a talk group give it a name and send the message

**Describe the solution you'd like**
Implement a direct message function like in
rocket.chat: https://docs.rocket.chat/guides/user-guides/direct-messages-between-multiple-users
mattermost: https://docs.mattermost.com/help/getting-started/organizing-conversations.html#direct-messages-and-group-messages

Just choose your members and start a group chat without group name.

nickvergessen:
Duplicate of https://github.com/nextcloud/spreed/issues/4254 ?

jakobroehrl:
This issue is just about quick group creation without choosing a group name basicly.
https://github.com/nextcloud/spreed/issues/4254 is about the same with the difference that for the recipients shows as 1:1 conversation (not as a group)

jakobroehrl:
Maybe we want to change in the future the chat room creation.
For example: Choose the members first and preselect the groupname with some of the member names? 
Then the user can create quick or change the group name

nickvergessen:
Tracked in #4254

PR #4476, opened 2020-10-27
Quick multi user direct message

jakobroehrl:
solves: https://github.com/nextcloud/spreed/issues/4448
Create a group without groupname input (groupname is generated out of the member names)
grafik

### Todo
- [ ] New Icon
- [ ] CSS


jakobroehrl:
@nickvergessen Could you look generally over this?

Issue #2800, opened 2020-01-20
Add reply-privately feature


ma12-co:
Add ability to select a message in a group conversation and reply to that message in a private conversation with the sender.

nickvergessen:
Addressed in https://github.com/nextcloud/spreed/pull/4855

PR #4477, opened 2020-10-27
WIP: Private reply

jakobroehrl:
grafik
solves: https://github.com/nextcloud/spreed/issues/2800
This is only a 1st try, I don't know how hard it is do finish it. 
### Questions/Todo
- [x] change the chat
- [x] post quoted msg
- [ ] send new msg
- [ ] use another icon?

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -68,6 +68,13 @@ the main body of the message as well as a quote.
 						@click.stop="handleReply">
 						{{ t('spreed', 'Reply') }}
 					</ActionButton>
+					<ActionButton
+						v-if="isReplyable"
+						icon="icon-user"
+						:close-after-click="true"
+						@click.stop="handlePrivateReply">
+						{{ t('spreed', 'Reply private') }}
+					</ActionButton>
 				</Actions>
 			</div>
 		</div>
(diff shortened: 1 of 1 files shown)


nickvergessen:
> What should happen, if the user clicks on the "Reply private" Button? Change the chat and show the quote there? If yes, how can I change the chat?

The thing is we prohibit this. You can not quote messages from chat A in chat B, to avoid leaking of messages when people are removed from chats etc. So my suggestion (same as for forward) would be to just copy the original text message with a `> ` in front to the input, after changing to the one-to-one conversation:

```
> What should happen, if the user clicks on the "Reply private" Button? Change the chat and show the quote there? If yes, how can I change the chat?
```

But switching is not that easy, because you might not have the one-to-one chat before, so you first need to create it or rejoin it in case you left it. But all this can be done with a call to
https://github.com/nextcloud/spreed/blob/bd78555cc31e653171ec68a1f87eafe2304d7bba/src/services/conversationsService.js#L129-L137

From that function you should always get back the room data, so you can afterwards trigger a join for that token, but @ma12-co might be better in telling you where that code is

jakobroehrl:
@jancborchardt Should I use another icon?

nickvergessen:
Converting to draft for easier sorting.

nickvergessen:
> @jancborchardt Should I use another icon?

🏓

jakobroehrl:
@nickvergessen 
When could we pick up the work here?
see: https://github.com/nextcloud/spreed/pull/4477#issuecomment-717810331

jakobroehrl:
see: https://github.com/nextcloud/spreed/pull/4855

Issue #4481, opened 2020-10-28
Enforce use of ICommentsManager in all objects


PVince81:
See https://github.com/nextcloud/spreed/pull/4425#issuecomment-717816406 for some context.

~~Objects should never rely on ICommentsManager but receive the `OCA\Talk\Chat\CommentsManager` instead, as we rely on implementation details of that specific instance.
Accepting any other implementation (like core's comments app CommentsManager) could result in unwanted side-effects.~~

After research we noticed that it is possible to alias ICommentsManager to CommentsManager to make it safe to use within the app.

nickvergessen:
ChatManager already get's the right class.

PVince81:
I've updated the subject and top post to match the new approach we discussed with aliasing.

PR here: https://github.com/nextcloud/spreed/pull/4483

PR #4425, opened 2020-10-21
Upstream our comments manager extensions to server

nickvergessen:
Ref https://github.com/nextcloud/server/pull/23601

PR #4482, opened 2020-10-28
Revert "Sorter use ICommentsManager directly"

nickvergessen:
This reverts commit 97bd312022bcf34f2e723f0cb41bb2ea5faad0ee.

From #4425

### DIFF
--- a/lib/Chat/AutoComplete/Sorter.php
+++ b/lib/Chat/AutoComplete/Sorter.php
@@ -23,15 +23,15 @@ declare(strict_types=1);
 
 namespace OCA\Talk\Chat\AutoComplete;
 
-use OCP\Comments\ICommentsManager;
+use OCA\Talk\Chat\CommentsManager;
 use OCP\Collaboration\AutoComplete\ISorter;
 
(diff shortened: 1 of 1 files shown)


PVince81:
Fixes https://github.com/nextcloud/spreed/issues/4481

See that ticket, there are two more classes to adjust to not receive a ICommentsManager

PVince81:
tried locally with [stubs](https://psalm.dev/docs/running_psalm/configuration/#stubs) and defining `OC\Comments\Manager` thag implements `ICommentManager` but it seems Psalm is ignoring the interface's body and just checking whether it's there.

I tried (for science!) to copy the full ICommentManager, fixing the namespaces and uses, and using that as a stub file and all errors disappear... This confirms to me that stubs do work in general when everything is defined directly, but Psalm does not follow the `interface` bit to populate its model.

Another crazy thing I tried is to copy ICommentsManager as is and add it as stub. Same problem: Psalm doesn't connect both.

PVince81:
Superseded by https://github.com/nextcloud/spreed/pull/4483

PR #4483, opened 2020-10-28
Always use CommentsManager from the app

PVince81:
Fixes https://github.com/nextcloud/spreed/issues/4481

The application's container will now always map ICommentsManager to
the app's CommentsManager instance instead of the one from server.

All the usages were replaced to use ICommentsManager now except for
lib/Manager which requires access to the extra method.

This made it possible to reduce the number of skipped Psalm warnings and
made sure there is no possible mix up of manager classes.

Signed-off-by: Vincent Petry <vincent@nextcloud.com>

### DIFF
--- a/lib/Chat/AutoComplete/Sorter.php
+++ b/lib/Chat/AutoComplete/Sorter.php
@@ -23,15 +23,16 @@ declare(strict_types=1);
 
 namespace OCA\Talk\Chat\AutoComplete;
 
-use OCP\Comments\ICommentsManager;
+use OCA\Talk\Chat\CommentsManager;
 use OCP\Collaboration\AutoComplete\ISorter;
+use OCP\Comments\ICommentsManager;
 
 class Sorter implements ISorter {
 
 	/** @var ICommentsManager */
 	protected $commentsManager;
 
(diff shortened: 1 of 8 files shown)


nickvergessen (approved):

PVince81:
I stepped into the code with a debugger and noticed that it fails parsing the mention with "No Displayname resolver for this type registered"

I suspect that the server's CommentsManager had more services injected or registered. Might need to do the same with the app's instance now.

PVince81:
@nickvergessen please check my latest commit, not sure if this is already going too far.

This bit was missing: https://github.com/nextcloud/server/blob/v20.0.0/lib/private/Server.php#L1085

I've copy-pasted the implementation as I don't see a good/quick way to have it common.
Maybe this will eventually need to be moved into a dedicated manager in the server that apps can reuse.

nickvergessen:
Arg, well registering them again is wrong, but that is the reason we had it wrapped then....
I will have a look tomorrow

PVince81:
I checked your approach and after digging a bit I understood why it all makes sense, let's merge this then :+1:

PR #4487, opened 2020-10-29
Restrict posting to moderators

### LINKED ISSUE #3886: 🛂 Chat permission

jakobroehrl:
Fix #3886

grafik

### Todo:
- [x] set archived state in localstore
- [x] hide archived rooms
- [x] show archived room (filter icon)
- [ ] save state in database
- [ ] sort archived rooms to the end of list
- [ ] nice icon, where to place the filter icon?

### DIFF
--- a/src/components/LeftSidebar/ConversationsList/Conversation.vue
+++ b/src/components/LeftSidebar/ConversationsList/Conversation.vue
@@ -83,6 +83,11 @@
 			<ActionSeparator />
 
 			<ActionButton v-if="canLeaveConversation"
+				:icon="iconArchive"
+				@click.prevent.exact="toggleArchiveConversation">
+				{{ labelArchive }}
+			</ActionButton>
+			<ActionButton v-if="canLeaveConversation"
 				:icon="iconLeaveConversation"
 				@click.prevent.exact="leaveConversation">
 				{{ t('spreed', 'Leave conversation') }}
@@ -137,6 +142,7 @@ export default {
 					type: 0,
 					displayName: '',
 					isFavorite: false,
+					isArchived: false,
 					notificationLevel: 0,
 					lastMessage: {},
 				}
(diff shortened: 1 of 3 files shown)


src/components/LeftSidebar/ConversationsList/Conversation.vue
```diff
@@ -82,6 +82,11 @@
 
 			<ActionSeparator />
 
+			<ActionButton v-if="canLeaveConversation"
```
nickvergessen, line 85: Wrong permission check

src/components/LeftSidebar/ConversationsList/Conversation.vue
```diff
@@ -82,6 +82,11 @@
...
+				:icon="iconArchive"
+				@click.prevent.exact="toggleArchiveConversation">
+				{{ labelArchive }}
+			</ActionButton>
```
nickvergessen, line 89: Since this is a purely moderator option it should go into the moderator menu?

nickvergessen:
I would just sort archived rooms to the end (so invert favorites basically) or even not load them and instead have a filter icon where you can toggle them in and out (need something similar for for the "listed" rooms anyway

src/components/LeftSidebar/ConversationsList/Conversation.vue
jakobroehrl: I don't think it's only a moderator option.
Everybody should archiv a room to have a cleaner room list. 
But wait, this is not possible with this setup. This needs to be done on a user basis. 
So maybe I don't understand the aim here

nickvergessen: A per user archiving is way over the top from my pov.

nickvergessen:
Master is 22 now

nickvergessen:
New attempt in https://github.com/nextcloud/spreed/pull/5259

PR #4489, opened 2020-10-29
Also delete the room if the last moderator leaves …

### LINKED ISSUE #4435: Regain access to a lost "chat"
Say I create a chat. I am an admin. The I leave the chat. Can I find that chat again and take control over it? Like I created a public chat to test this, but I do not know how to regain access to that chat and delete it...

nickvergessen:
Fix #4435


nickvergessen:
I found this while working on #4324 so no tests and stuff as we need to backport this and I dont want to make the rebase of the sip too complicated. I will create an issue to write respective tests later.

PVince81 (approved):
Makes sense :+1:

Issue #4492, opened 2020-10-29
Double scrollbar in call view


PVince81:
## Steps to reproduce
1. Create a call (a single participant is enough)
2. Post some messages in the chat to make it scroll

### Expected behaviour
Single scroll bar

### Actual behaviour
Double scrollbar
<img width="501" alt="image" src="https://user-images.githubusercontent.com/277525/97624927-164b9c00-1a28-11eb-997d-6501d071eba5.png">

Regression introduced by https://github.com/nextcloud/spreed/pull/4473

Seems we might need to add back some of the hacks :frowning:

PVince81:
Fix is here https://github.com/nextcloud/spreed/pull/4493

PR #4493, opened 2020-10-29
Remove double scrollbar in call chat view

PVince81:
Fixes https://github.com/nextcloud/spreed/issues/4492

Seems this is already enough, tested on Firefox and Chromium.

### DIFF
--- a/src/components/NewMessageForm/NewMessageForm.vue
+++ b/src/components/NewMessageForm/NewMessageForm.vue
@@ -401,17 +401,18 @@ export default {
 @import '../../assets/variables';
 
 .wrapper {
+	position: sticky;
+	bottom: 0;
+	background-color: var(--color-main-background);
+	display: flex;
+	justify-content: center;
 	border-top: 1px solid var(--color-border-dark);
 	padding: 4px 0;
 }
 
 .new-message {
-	margin: auto;
 	max-width: $messages-list-max-width;
-	position: sticky;
-	position: -webkit-sticky;
-	bottom: 0;
-	background-color: var(--color-main-background);
+	flex: 1 1 100%;
 	&-form {
 		display: flex;
 		align-items: center;


skjnldsv:
Why is this not working in the first place?
What is taking extra height?

Also the #app-sidebar css above isn't used anymore, there is no such id in the vue components (unrelated to this pr though, but I just saw it)

PVince81:
the extra height is from the input box element (".wrapper"). if I hide it then the biggest scrollbar disappears.

one idea would be to change that one to position: absolute, bottom: 0 but then we need to reduce the chat container height accordingly to not hide messages there. And the height of that input box element (".wrapper") is not predictable as it can change depending on screen sizes

PVince81:
I tried that and with position: sticky it's the less worse result:
image

but the scrollbar appears on the right side of the field.

and with position:absolute it's even worse

so far I prefer the one-liner fix/hack, but if you have other ideas let me know :-/

skjnldsv:
It seems related to the file-upload input to me :thinking:

skjnldsv:
Also, lot to cleanup on the `.scroller` overflowing and the `.wrapper` should most likely be tyhe one carrying the sticky-bottom, not the `new-message`
```
.wrapper {
	border-top: 1px solid var(--color-border-dark);
	padding: 4px 0;
	position: sticky;
	bottom: 0;
	background-color: var(--color-main-background);
}
```

skjnldsv:
Yeah, it works like that.

Capture d’écran_2020-10-29_21-24-38
Capture d’écran_2020-10-29_21-24-45
Capture d’écran_2020-10-29_21-25-38

skjnldsv:
Capture d’écran_2020-10-29_21-26-34

PVince81:
does the scrollbar still near the field ?

seems we'll need to apply all this only specifically when in "tab chat" mode as we likely need to keep all these styles for the regular chat view

PVince81:
yeah, it still goes past bottom:
image

skjnldsv:
> yeah, it still goes past bottom:

That is how sticky works, it's all good.

PVince81:
> > yeah, it still goes past bottom:
> 
> That is how sticky works, it's all good.

But inconsistent with the main chat view where it doesn't do that. Probably fine for the sidebar though.

I tried leaving the call and switching to main call view, there the scrollbar is gone. So we'll need to add all these specifically for sidebar mode. Not sure why that would be better than the single line "height: 100%" ?

PVince81:
regarding file upload: `.hidden-visually` has "top: auto" if we set it also to "-10000px" the input field is properly "off the layout"

skjnldsv:
> I tried leaving the call and switching to main call view, there the scrollbar is gone. So we'll need to add all these specifically for sidebar mode. Not sure why that would be better than the single line "height: 100%" ?

Because there is a lot of conflicting unneeded files.
I went for the cleanup way. I prefer to remove stuff that was not properly done that adding another line here :)
Everything should be self container, the layout is quite simple

```
Sidebar tab
| Messages List
| Sticky writer
```

Yet,there are tons of css for something that almost default css would display :see_no_evil: 
Message list should just be displayed as is, no scroller as the sidebar have the scrolling ahdnling
And the writer should just be sticky to the bottom, simple as that :)

skjnldsv:
> regarding file upload: `.hidden-visually` has "top: auto" if we set it also to "-10000px" the input field is properly "off the layout"

It is conflicting because the sticky is not including the input field for some reason, so it is not taken into account, and seen as out of the view (so it seems on the quick check I jus did :thinking: )

PVince81:
@skjnldsv fix for hidden-visually: https://github.com/nextcloud/server/pull/23782

skjnldsv:
> I went for the cleanup way. I prefer to remove stuff that was not properly done that adding another line here :)

If you just want to fix this, without cleaning the other code (which is kinda unrelated to this pr, I went a bit overhead :see_no_evil: )
Just fixing the new-message display is enough
Capture d’écran_2020-10-29_21-41-14
Capture d’écran_2020-10-29_21-41-20

PVince81:
if we touch .new-message I'd need to either use a `v-deep` or to modify the style in NewMessageForm and make it conditional, unless you think that this change is safe enough to be used in the regular chat view so we could apply it for all cases

PVince81:
I've pushed the change for the fix that works without the extra cleanup, as the latter will need much more work.

Tested with:
- [x] regular chat view scrolling
- [x] call mode sidebar chat scrolling
- [x] files app sidebar mode scrolling (merged with a rebased https://github.com/nextcloud/spreed/pull/4291)

skjnldsv (approved):

PR #4495, opened 2020-10-29
Inject MimeTypeLoader into RoomShareProvider to not query it on each …

MorrisJobke:
…createShareObject call

Found during some profiling.

Bildschirmfoto 2020-10-29 um 22 46 17

### DIFF
--- a/lib/Share/RoomShareProvider.php
+++ b/lib/Share/RoomShareProvider.php
@@ -41,6 +41,7 @@ use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\EventDispatcher\IEventDispatcher;
 use OCP\Files\Folder;
+use OCP\Files\IMimeTypeLoader;
 use OCP\Files\Node;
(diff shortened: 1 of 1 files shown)


PVince81 (approved):
Looks fine to me :+1:

PVince81:
ohhhh, failure on master now https://github.com/nextcloud/spreed/actions/runs/341270807

PVince81:
raised as https://github.com/nextcloud/spreed/issues/4512

sometimes I have the feeling that touching some files might wake up Psalm...

MorrisJobke:
> sometimes I have the feeling that touching some files might wake up Psalm...

Most likely because it can now properly infer the type. Before it was only inferable from a private namespace method call? Need to check how server code is injected into psalm inside apps to properly understand this.

MorrisJobke:
> Most likely because it can now properly infer the type. Before it was only inferable from a private namespace method call? Need to check how server code is injected into psalm inside apps to properly understand this.

Ah wait ... this is the mime type loader not the event dispatcher ... then yes: no idea why it is triggered.

Issue #4441, opened 2020-10-23
Automatic collapse of stripe when a screen share starts


PVince81:
**Is your feature request related to a problem? Please describe.**
Since https://github.com/nextcloud/spreed/pull/4363 it is possible to collapse the strip at the bottom of the call view.
Usually when someone shares their screen, the purpose is for everyone to focus on just that. Instead of requiring every user to collapse the stripe to get a better view of the screen share, we should do that automatically.

**Describe the solution you'd like**
- Whenever a remote screen starts:
   - remember the preview view mode + collapsed state ("save1")
   - if in grid view mode, show the remote screen in full page
   - if in stripe view mode, automatically collapse the view
   - if follow mode is enabled, do nothing unless the followed view is the screen share itself
- If the user uncollapses or switches the view during that one screen share, reset "save1" to not remember any more
- Whenever a remote screen share stops
   - if "save1" state was set (means we're still in automatic mode), restore the view to what it was saved in "save1", then reset that value

Need to find out whether that implicit automatic mode is enough from an UX perspective or whether we need an explicit mode somewhere.

PVince81:
PR is here https://github.com/nextcloud/spreed/pull/4521

majkinetor:
Doesn't seeem to be working for me, that is, sometimes it does ,but mostly doesn't.

nickvergessen:
There is an option in Talk 11 to collapse the bar with a simple button click

demogorgonz:
> There is an option in Talk 11 to collapse the bar with a simple button click

Can you maybe make a screen shoot with the button ?

nickvergessen:
See https://github.com/nextcloud/spreed/pull/4363#issuecomment-708299399

demogorgonz:
> See [#4363 (comment)](https://github.com/nextcloud/spreed/pull/4363#issuecomment-708299399)

Thank you very much for the link, however on NC version `20.0.7` and Talk plugin version `10.0.6` this option/button doesn't appear.

PVince81:
you need NC 21 and Talk >= 11

PR #4521, opened 2020-11-03
Remember call view state during presentation

PVince81:
When a presentation starts (screen share), remember the grid + stripe
state then switch to presenter mode (no grid + no stripe).

When the presentation stops (last screen share stops), restores the grid
+ stripe state, unless the user has changed the state during the
presentation.

Stripe mode is now also remembered for each call.

Fixes https://github.com/nextcloud/spreed/issues/4441

- [x] will conflict with https://github.com/nextcloud/spreed/pull/4542, let's wait for the latter to be merged


nickvergessen (approved):
Works as decribed

nickvergessen:
Would have given bonus points if returning to grid view after follow-unfollow.

Issue #4527, opened 2020-11-03
Scroll to bottom button is missing


nickvergessen:
## Steps to reproduce
1. Scroll up in the chat view
2. Use the "Scroll to bottom" button to jump to the bottom again

### Expected behaviour
Button is there

### Actual behaviour
Button is not in vision, regression from the height-mess in the sidebar #4493

PVince81:
Works for me in the main conversation view.

But it's indeed missing in the sidebar chat mode.

PVince81:
also, the sidebar chat doesn't automatically scroll to bottom

I suspect that the wrong scroll container is used now by the code, considering that we changed where the overflow and scrollbars are happening.

nickvergessen:
Another related error:
1. Create a new conversation
2. Start call
3. Expect the chat input to be at the bottom of the sidebar

![Bildschirmfoto von 2020-11-04 11-21-44](https://user-images.githubusercontent.com/213943/98099644-19171880-1e90-11eb-8cba-899b8a97b091.png)

PVince81:
PR here for scrolling issues: https://github.com/nextcloud/spreed/pull/4529

PVince81:
new one liner that fixes both: https://github.com/nextcloud/spreed/pull/4530

PR #4529, opened 2020-11-04
Fix message list scroll container in sidebar

PVince81:
When embedded into the right sidebar, the scrolling is achieved by a
parent element instead of the MessagesList itself.

To be able to listen to and control the scrolling, now a reference to
the scroll container is passed in.

Fixes https://github.com/nextcloud/spreed/issues/4527

This issue was due to the recent changes in scrollbar containment in the sidebar.

#### Scrolling
- [x] scroll up in main chat view: chevron-down button appears
- [x] scroll up in call sidebar chat view: chevron-down button appears
- [ ] scroll up in files sidebar chat view: chevron-down button appears
- [x] auto scroll to bottom works in main chat view (sticky)
- [x] auto scroll to bottom works in call sidebar chat view (sticky)
- [ ] auto scroll to bottom works in files sidebar chat view (sticky)

#### Layout
- [ ] call sidebar has proper height when few messages
- [ ] files sidebar has proper height when few messages

### DIFF
--- a/src/components/ChatView.vue
+++ b/src/components/ChatView.vue
@@ -43,7 +43,8 @@
 			</div>
 		</transition>
 		<MessagesList
-			:token="token" />
+			:token="token"
+			:scroll-container="scrollContainer" />
 		<NewMessageForm />
 	</div>
 </template>
@@ -67,6 +68,17 @@ export default {
 		token: {
 			type: String,
 			required: true,
+		},
+
+		/**
+		 * The scroll container.
+		 *
+		 * Can be different depending on containment
+		 */
+		scrollContainer: {
+			type: HTMLElement,
+			required: false,
+			default: null,
 		},
 	},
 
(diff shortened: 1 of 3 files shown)


PVince81:
this took me a while as I first tried to use computed properties but bumped against this issue: https://stackoverflow.com/questions/43531755/using-refs-in-a-computed-property

One should not use "$refs" in computed properties as it might not be defined initially, and also $refs is not reactive so even if it becomes defined the computed property won't re-evaluate. Hence I've used watchers instead.

PVince81:
back to draft as I'm going to investigate the size issue https://github.com/nextcloud/spreed/issues/4527#issuecomment-721645929 and there's a slight risk that I might need to reshuffle the containers to make it work :see_no_evil:

PVince81:
I knew it... to fix the container I have to set height: 1px on it as per https://stackoverflow.com/a/21836870

zero-height-magic mkv

but doing so breaks the scrolling events and measurements on that element.

needs further research

PVince81:
I found a simpler fix that fixes both... https://github.com/nextcloud/spreed/pull/4530

PR #4531, opened 2020-11-04
Rework scroll into view of conversation after post

PVince81:
Fixes an issue where a chat in the files sidebar would trigger a warning
because the conversation list element don't exist there. See https://github.com/nextcloud/spreed/pull/4291#issuecomment-708979771

When posting a message, an event is now triggered so other components
could react on it.

The conversation list will now take the responsibility of scrolling the
current conversation into view after posting, instead of the
NewMessageForm.

The behavior is also adjusted to always scroll regardless where the
conversation is located in the list.

@ma12-co not sure why this was limited to only when it moves to the top.
I tested also by clicking on a conversation at the bottom of list, then scrolling up, then posting: it will now scroll down to highlight the conversation even when at the bottom.

Note: I don't understand when a conversation moves to the top and when not, it seemed to be random in my tests... could be another bug.

### DIFF
--- a/src/components/LeftSidebar/ConversationsList/ConversationsList.vue
+++ b/src/components/LeftSidebar/ConversationsList/ConversationsList.vue
@@ -75,13 +75,25 @@ export default {
 
 	mounted() {
 		EventBus.$on('routeChange', this.onRouteChange)
+		EventBus.$on('newMessagePosted', this.onMessagePosted)
 	},
 
 	beforeDestroy() {
 		EventBus.$off('routeChange', this.onRouteChange)
+		EventBus.$off('newMessagePosted', this.onMessagePosted)
 	},
 
 	methods: {
+		onMessagePosted({ token }) {
+			const conversation = document.getElementById(`conversation_${token}`)
+			this.$nextTick(() => {
+				conversation.scrollIntoView({
+					behavior: 'smooth',
+					block: 'start',
+					inline: 'nearest',
+				})
+			})
+		},
 		onRouteChange({ from, to }) {
 			if (from.name === 'conversation') {
 				leaveConversation(from.params.token)
--- a/src/components/NewMessageForm/NewMessageForm.vue
+++ b/src/components/NewMessageForm/NewMessageForm.vue
@@ -180,15 +180,7 @@ export default {
 		},
 	},
 
-	watch: {
-		token(newValue, oldValue) {
-			this.isCurrentConversationIsFirstInList()
-		},
-	},
-
 	mounted() {
-		this.isCurrentConversationIsFirstInList()
-
 		EventBus.$on('uploadStart', this.handleUploadStart)
 	},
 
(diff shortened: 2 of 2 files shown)


PVince81:
@danxuliu thanks for the addition, indeed I forgot to `off`. Squashed.

danxuliu (commented):
Tested and works as described. But I do not know why before the scrolling was done only if it was not the first item in the list*, so I can not approve :-) In any case, code looks good :+1:

*Note that as `conversationIsFirstInList` was updated only on mount and on token change if the conversation moved to the top once the list continued scrolling to the top whenever the current user posted a new message (until the conversation was changed). That was probably unintended, though.

> Note: I don't understand when a conversation moves to the top and when not, it seemed to be random in my tests... could be another bug.

For me it always moves to the top when the current user postes a message; I have not dug into the code, though.

src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -387,12 +369,6 @@ export default {
 
 			range.setStartAfter(emojiTextNode)
 		},
-
-		// Check whether the current conversation is the first in the conversations
-		// list and stores the value in the component's data.
-		isCurrentConversationIsFirstInList() {
-			this.conversationIsFirstInList = this.$store.getters.conversationsList.map(conversation => conversation.token).indexOf(this.token) === 0
```
nickvergessen, line 394: This didn't work when you had a favorite and posted in a normal conversation?

src/components/LeftSidebar/ConversationsList/ConversationsList.vue
```diff
@@ -75,13 +75,25 @@ export default {
...
+		EventBus.$off('newMessagePosted', this.onMessagePosted)
 	},
 
 	methods: {
+		onMessagePosted({ token }) {
+			const conversation = document.getElementById(`conversation_${token}`)
+			this.$nextTick(() => {
+				conversation.scrollIntoView({
+					behavior: 'smooth',
+					block: 'center',
```
nickvergessen, line 92: I find the center thing always weird. I think the top makes much more sense?

src/components/NewMessageForm/NewMessageForm.vue
PVince81: I just tested in this PR and having a favorite or normal conversation makes no difference.
It just makes sure that whatever room you posted in becomes visible in the left panel.

And when posting in a room that is invisible below the scroll panel, the room gets moved to the top, below favorites, highlighted, and it scrolls to that.

@ma12-co do you know why there was a check for "first room" ?

src/components/LeftSidebar/ConversationsList/ConversationsList.vue
PVince81: yeah, I'll change to "start" as I also observed some occurrences where it stopped in the middle of a row.

PVince81:
adjusted scrollIntoView to "top" instead of "center", rebased, squashed

src/components/NewMessageForm/NewMessageForm.vue
ma12-co: Don't remember honestly :/

ma12-co (approved):
tested and works

Issue #4494, opened 2020-10-29
Image label in chat view causing layout disturbances


PVince81:
## Steps to reproduce
1. Post a picture in landscape mode
2. Start a call
3. Open the chat sidebar
4. Reply to the message in question

### Expected behaviour
Time of message properly aligned.
No extra spacing below the file name label.
In the reply box, also no extra spacing below the file name label.

### Actual behaviour
Time of message properly pushed to the right.
Extra spacing below the file name label.

<img width="590" alt="image" src="https://user-images.githubusercontent.com/277525/97631625-8d396280-1a31-11eb-8b35-2323bc5a94a6.png">

In the reply box, also extra spacing below the file name label.

<img width="300" alt="image" src="https://user-images.githubusercontent.com/277525/97631726-ae01b800-1a31-11eb-9cc7-107de735e3b4.png">

## Talk app

**Talk app version:** 11.0 pre-alpha git master (062916bc71966e99bbf7a50b1c193f9c88013575)

## Notes

Likely introduced or worsened through the changes that allow the image to expand horizontally in https://github.com/nextcloud/spreed/pull/4472

I didn't find where the extra spacing is coming from.
Seems we need to change it to another display mode, not block.

jancborchardt:
As also mentioned elsewhere – other chat apps don’t show the image title as it’s rarely useful and looks rather technical. Let’s just cut it. :)

PVince81:
PR here https://github.com/nextcloud/spreed/pull/4534

PVince81:
As discussed in https://github.com/nextcloud/spreed/pull/4534, the file name will still be preserved for non-image file types.

As for the layout, I found that the order of elements matter, more precisely the spaces that are introduced by HTML markup can have an influence...

This is the current state:
<img width="641" alt="image" src="https://user-images.githubusercontent.com/277525/98384338-ee72be80-204d-11eb-9433-8a5baa342038.png">

And if we move the "strong" element to the end of the comment blocks the extra space vanishes:
<img width="641" alt="image" src="https://user-images.githubusercontent.com/277525/98384382-fe8a9e00-204d-11eb-9725-56a1abe7062e.png">

But it also means moving the file name below the progress bar on upload. I think it's not that bad as it would make the progress bar closer to the preview, as they fit well together.

PR #4534, opened 2020-11-04
Move the filename of attachments to a tooltip

PVince81:
Instead of displaying file names below attachments in a conversations,
like for images, it is now displayed only on hover in a tooltip.

Howver, in the upload editor the file names are still shown as they can
be helpful in that situation.

This cleans up the message list a bit and also solves some layout issues
with images in narrow views like the sidebar.

~~Fixes https://github.com/nextcloud/spreed/issues/4494~~ the latter ticket will still exist for cases where file name is visible

image

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
@@ -29,17 +29,20 @@
 		@click="handleClick"
 		@keydown.enter="handleClick">
 		<img v-if="(!isLoading && !failed)"
+			v-tooltip.auto="previewTooltip"
 			:class="previewImageClass"
 			class="file-preview__image"
 			alt=""
 			:src="previewUrl">
 		<img v-if="!isLoading && failed"
+			v-tooltip.auto="previewTooltip"
 			:class="previewImageClass"
 			alt=""
 			:src="defaultIconUrl">
 		<span v-if="isLoading"
+			v-tooltip.auto="previewTooltip"
 			class="preview loading" />
-		<strong>{{ name }}</strong>
+		<strong v-if="isUploadEditor">{{ name }}</strong>
 		<button v-if="isUploadEditor"
 			tabindex="1"
 			:aria-label="removeAriaLabel"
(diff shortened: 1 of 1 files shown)


ma12-co:
I think that if anything, we should apply this only to images, as for any other kind of file the title might be the only hint to what's in it.

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
```diff
@@ -29,17 +29,20 @@
...
 		<img v-if="!isLoading && failed"
+			v-tooltip.auto="previewTooltip"
 			:class="previewSizeClass"
 			alt=""
 			:src="defaultIconUrl">
 		<span v-if="isLoading"
+			v-tooltip.auto="previewTooltip"
 			class="preview loading" />
-		<strong>{{ name }}</strong>
+		<strong v-if="isUploadEditor">{{ name }}</strong>
```
nickvergessen, line 45: I would show the name when we don't have an image preview but a mimetype only. Also the preview of a `.md` file is rather useless and the name adds very useful information and therefor should be displayed by default

PVince81: makes sense, so the condition becomes, display file name if:
- mime type is not "image/" + something
or
- there is no preview available for this file
(previews could be disabled for some images by config)

PVince81: adjusted

jancborchardt: Yup, agreed – I also said the filename should only be removed for images. :) thanks for adjusting!

ma12-co:
Should we just get rid of the filename in case of images?? cc @jancborchardt 

If not could we: 
* Display the tooltip to the right of the image so that it doesn't cover up other elements;
* Delay its appearance by 0,5 to 1 second;

PVince81:
> Should we just get rid of the filename in case of images?? cc @jancborchardt
> 
> If not could we:
> 
>     * Display the tooltip to the right of the image so that it doesn't cover up other elements;
> 
>     * Delay its appearance by 0,5 to 1 second;

I'd rather we still have a way to display the file name. In some cases that name might be meaningful.

I could change the tooltip settings to appear on the right, however, note that in sidebar mode there might be no room for a tooltip on the right. Might work assuming that the tooltip lib is able to deal with such scenarios and automatically shift.

ma12-co:
> Might work assuming that the tooltip lib is able to deal with such scenarios and automatically shift.

It should yes

PVince81:
@ma12-co rebased, and adjusted tooltip, see last commit

ma12-co (approved):

PR #4538, opened 2020-11-05
Add play button on top of video or gif previews

PVince81:
### Description

Whenever a video preview of GIF preview (not direct) is displayed, add a play button to make it clear that what is here is not a picture but a playable video.

2020-11-05 11-48-28 mkv

### Testing

When testing, make sure to have the `Movie` preview provider enabled.

### Todos

- [x] REQUIRES https://github.com/nextcloud/spreed/pull/4537 for the preview type detection
- [x] improve video button style (misaligned, blurry)

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
@@ -28,16 +28,22 @@
 		:class="{ 'file-preview--viewer-available': isViewerAvailable, 'file-preview--upload-editor': isUploadEditor }"
 		@click="handleClick"
 		@keydown.enter="handleClick">
-		<img v-if="(!isLoading && !failed)"
-			v-tooltip="previewTooltip"
-			:class="previewImageClass"
-			class="file-preview__image"
-			alt=""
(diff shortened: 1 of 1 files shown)


PVince81:
rebased and conflicts solved

PVince81:
I managed to improve the button style:
image

ma12-co:
Shouldn't we just autoplay them muted when they're in the viewport? Don't gifs already autoplay?

PVince81:
> Shouldn't we just autoplay them muted when they're in the viewport? Don't gifs already autoplay?

GIFs only autoplay when they have a small size, by default less then 3 MB.

For videos, we render the preview by default so no autoplay. The performance might be bad if we autoplay many videos within the thread, might need to do that only when they become visible or on hover. But I'd see this as a separate task/attempt.

As for playing the videos directly, this is more involved as we need to have the HTML 5 video player in place, ideally the same like the viewer app. Or maybe have a way for the viewer app to inject its player there. This is all more involved and would rather attempt this much later...

ma12-co (approved):
All righty then :)

PVince81:
rebased and conflict solved, retested

PR #4539, opened 2020-11-05
Only autorename uploads on paste

PVince81:
There was a mix up of handlers that made all uploads be automatically
renamed.

This properly rewires the handlers to only autorename whenever the event
was a paste.

### Tested
- [x] drag and drop: no rename
- [x] upload button: no rename
- [x] paste: autorename!


danxuliu (approved):
Tested and works :+1:

Issue #4536, opened 2020-11-05
Upload confirmation screen needs mouse click


nickvergessen:
## Steps to reproduce
1. Paste image content into the chat with Strg+V 
2. Upload confirmation dialog pops up
![Bildschirmfoto von 2020-11-05 10-03-39](https://user-images.githubusercontent.com/213943/98220335-6d80cd80-1f4e-11eb-8541-421e63bca086.png)
3. Try to submit with Enter 

### Expected behaviour
Submits

### Actual behaviour
Doesn't submit

PVince81:
thanks for raising, had it on my "list of to raise" :smile:

PVince81:
Fix is here: https://github.com/nextcloud/spreed/pull/4541

I was first wondering if we should skip the dialog altogether, but one might want to review whatever they pasted before submitting anyway.

nickvergessen:
We just added the dialog in the last release because too many people pasted into the wrong chat by accident and its currently quite complex to remove the share again.

PR #4541, opened 2020-11-05
Focus on submit button by default in upload dialog

PVince81:
Makes it possible to submit directly with enter after pasting

Fixes https://github.com/nextcloud/spreed/issues/4536


danxuliu (approved):
Tested and works :+1:

Issue #4500, opened 2020-10-30
Ellipsis long user names on sidebar


skjnldsv:
![Capture d’écran_2020-10-29_08-49-43](https://user-images.githubusercontent.com/14975046/97713619-5e1e0200-1ac0-11eb-9276-91d28cec775b.png)

nickvergessen:
It used to be ellipsed :thinking:

PVince81:
I have a quick solution for ellipsis that works for the participant list.

Still, I'm not happy to not have a tooltip there, so looking into the latter, and making it conditional with [ellipsis detection](https://stackoverflow.com/questions/7738117/html-text-overflow-ellipsis-detection)

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/4548

PVince81:
also no ellipsis on stable20 and Talk 10.0.1, same behavior there.
let's backport

PR #4548, opened 2020-11-06
Fix long participant name + status in participant list with tooltip

PVince81:
## Description

Added ellipsis on long participant names.
Added conditional tooltip on participant + status:
    
Whenever the participant name or status is ellipsized, a tooltip will be available for those entries.    
To check the "is ellipsized" condition, a ResizeObserver is used to update. To avoid trying to measure and update too many participants in the list, the observer only activates after 10+ characters, which is the limit of chinese characters that still fit within the sidebar minimum width of 300px.
Note that in the future once we use virtual scroll for the participant list, we likely won't need to worry about resize optimization.

Fixes https://github.com/nextcloud/spreed/issues/4500

### Long status ellipsis + tooltip
image

### Long user name + tooltip
image

Note: if the user name has no space, the tooltip breaks the layout, this is a bug in the tooltip library and needs to be addressed there.

### DIFF
--- a/src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
+++ b/src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
@@ -39,13 +39,15 @@
 			:offline="isOffline" />
 		<div class="participant-row__user-wrapper">
 			<div class="participant-row__user-descriptor">
-				<span class="participant-row__user-name">{{ computedName }}</span>
+				<span
+					v-tooltip.auto="userTooltip"
+					class="participant-row__user-name">{{ computedName }}</span>
 				<span v-if="showModeratorLabel" class="participant-row__moderator-indicator">({{ t('spreed', 'moderator') }})</span>
 				<span v-if="isGuest" class="participant-row__guest-indicator">({{ t('spreed', 'guest') }})</span>
 			</div>
-			<div v-if="getStatusMessage(participant)"
+			<div v-if="statusMessage"
 				class="participant-row__status">
-				<span>{{ getStatusMessage(participant) }}</span>
+				<span v-tooltip.auto="statusMessage">{{ statusMessage }}</span>
 			</div>
 		</div>
 		<div v-if="callIcon"
@@ -90,6 +92,7 @@
 <script>
 
 import ActionButton from '@nextcloud/vue/dist/Components/ActionButton'
+import Tooltip from '@nextcloud/vue/dist/Directives/Tooltip'
 import Actions from '@nextcloud/vue/dist/Components/Actions'
 import Microphone from 'vue-material-design-icons/Microphone'
 import Phone from 'vue-material-design-icons/Phone'
@@ -109,6 +112,10 @@ export default {
 		Microphone,
 		Phone,
 		Video,
+	},
+
+	directives: {
+		tooltip: Tooltip,
 	},
 
 	mixins: [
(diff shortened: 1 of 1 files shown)


PVince81:
@skjnldsv FYI in case you want to check it out

src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
```diff
@@ -263,6 +300,13 @@ export default {
 	},
 
 	methods: {
+		handleResize() {
+			const e = this.$refs.userName
+			this.showUserNameTooltip = (e && e.offsetWidth < e.scrollWidth)
+			const e2 = this.$refs.userStatus
+			this.showUserStatusTooltip = (e2 && e2.offsetWidth < e2.scrollWidth)
+		},
```
skjnldsv, line 308: Is this really needed? I feel like a simple would suffice.
It's fine if we just show it for all sizes. It would avoid the cost of a ResizeObserver as well as increase accessibility :)

PVince81: it looks shitty with short names...

image

now, there is another possible, reusable but more complex solution: extend the [Tooltip directive](https://nextcloud-vue-components.netlify.app/#/Directives) to be able to check an element's ellipsization only **on hover** and then decide whether to display or not. Or at least provide some events so that the caller could do the check on their own.

I'm fine removing all the conditionals here if we are fine with the above look...

skjnldsv: Well titles appear on long-hover only, I think it's fine.
I'm afraid a resizeobserver for each participant just doesn't scale.

PVince81: ok, I've removed the conditions

it's less bad than I think as the hover will only appear if you hove on the text itself, not the whole row, so it's ok

ma12-co: Couldn't we just observe the resize one level above? we only need to do it once for all participant components. I think that the bob tooltip looks really bad.

PVince81: Even if we observe in the ParticipantList, we still need to trigger an event to all the participant entries individually so they can perform the tooltip check.

The best solution is to have an on-demand check, basically whenever the user hovers, the check is done, and then the decision is made whether the tooltip appears or not. Ideal would be to have this as an option in the v-tooltip directive.

PVince81: I've raised https://github.com/nextcloud/nextcloud-vue/issues/1567 to address this there.

PVince81: if this is the lib we use there is a chance: https://github.com/Akryum/v-tooltip

I see there are event handlers, I'll see if I can prevent displaying...

PVince81: I checked the library and it seems we cannot use it that way.
See my attempts here: https://github.com/nextcloud/nextcloud-vue/issues/1567#issuecomment-725535802

Now I found another trick that works without touching the lib: within a "mouseover" handler, do the ellipsis check and quickly update the tooltip text afterwards. That seems to work for me. See https://github.com/nextcloud/spreed/pull/4548/commits/40d87afa9046632f1a0d3a9ea26d6776614787ac

ma12-co (approved):

PR #4551, opened 2020-11-06
Add tooltip for long room names.

PVince81:
Added tooltip for room names in case they are too long.
Note: the tooltip appears regardless of length.

- [x] REQUIRES: https://github.com/nextcloud/nextcloud-vue/pull/1555
- [x] REQUIRES: https://github.com/nextcloud/spreed/pull/4567
- [ ] add "is tooltip necessary check" using the same approach like https://github.com/nextcloud/spreed/pull/4548/commits/40d87afa9046632f1a0d3a9ea26d6776614787ac

### DIFF
--- a/src/components/RightSidebar/RightSidebar.vue
+++ b/src/components/RightSidebar/RightSidebar.vue
@@ -25,6 +25,7 @@
 		v-show="opened"
 		id="app-sidebar"
 		:title="title"
+		:title-tooltip="title"
 		:starred="isFavorited"
 		:title-editable="canModerate && isRenamingConversation"
 		:class="'active-tab-' + activeTab"


PVince81:
Rebased.

It turns out I can't easily or cleanly access that "h2" element from within the AppSideBar components.
So I think we should put the logic for "is there an ellipsis then show tootltip" to the Vue component instead and do so only if the tooltip text is the same as the title.

Not sure if the current PR is acceptable in that form then as the tooltip is always displaying.

nickvergessen:
Doesn't show a tooltip for me in FF82.0.3

PVince81:
@nickvergessen you might need to `make dev-setup` as the tooltip is provided by the latest nextcloud-vue update

nickvergessen (approved):

nickvergessen:
Right, I had that before, but to test another branch I went back to an older version it seems.

nickvergessen:
Should also do that on the left sidebar, but one thing after another.

Issue #4553, opened 2020-11-06
High CPU with LoadingPlaceholder animation


PVince81:
## Steps to reproduce
1. Apply this patch to display the placeholder forever:
```
diff --git a/src/components/MessagesList/MessagesList.vue b/src/components/MessagesList/MessagesList.vue
index 340341d6d..6918d365d 100644
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -46,7 +46,7 @@ get the messagesList array and loop through the list to generate the messages.
                        v-bind="item"
                        :messages="item"
                        @deleteMessage="handleDeleteMessage" />
-               <template v-if="!messagesGroupedByAuthor.length">
+               <template v-if="true || !messagesGroupedByAuthor.length">
                        <LoadingPlaceholder
                                type="messages"
                                :count="15" />
```
2. Open a conversation in the UI
3. Observe CPU usage

### Expected behaviour
Low CPU usage, but not zero.

### Actual behaviour
High CPU usage, see more info below.

I noticed this because somehow the http requests were stuck so that state stayed longer and my laptop fan kicked in very loudly!

## Talk app

**Talk app version:** 11.0 pre-alpha (git master 3af57c87b27ae4a9a22c58232e195277c1326f5e)

**Hardware:** Dell XPS 13 Laptop with Intel GPU

** Browser:** Chromium 86.0.4240.111 (openSUSE Build) (64-bit)

## More information

On my laptop with Intel GPU, the load is about 54% CPU for Chromium.

I tried replacing it with a simple pulse animation with opacity and the load reduced to ~22%.
Then I added a `translate: translateZ(0)` but it didn't improve much (https://stackoverflow.com/a/13293044)

Now the opacity animation doesn't look as cool, so I tried to reproduce it by having the list twice. One list with regular gradient and one with inverted one. Then use pulse animation to fade between both. The CPU load is still at 22%.

PR upcoming

PVince81:
@skjnldsv I heard that a similar component exists in the unified search, so might need adjusting there as well.

skjnldsv:
Hum, this is a simple svg animation iirc, shouldn't be that hard on the processor unless you show tons of them.
We should 3 on the unified search, I think it make sense to lower the number of entries show.

PVince81:
it's not that simple: we are changing the stop values of a gradient, which means recomputing the whole gradient for every frame.
simple would be using opacity in SVG.
[...]

PR #4554, opened 2020-11-06
Improve loading placeholder animation performance

PVince81:
Fixes https://github.com/nextcloud/spreed/issues/4553

Replaced the SVG gradient animation with a CSS opacity animation that fades between two lists, one with the regular gradient and one with the reversed one.

See ticket for details about analysis.

The fix reduces the CPU load for me from 54% to ~20 %

### DIFF
--- a/src/components/LoadingPlaceholder.vue
+++ b/src/components/LoadingPlaceholder.vue
@@ -51,7 +51,7 @@ each other by animating the opacities.
 						:fill="'url(#placeholder-gradient' + suffix + ')'">
 						<circle class="conversation-placeholder-icon" />
 						<rect class="conversation-placeholder-line-one" />
-						<rect class="conversation-placeholder-line-two" :style="{width: `width`}" />
+						<rect class="conversation-placeholder-line-two" :style="width" />
 					</svg>
 					<svg
 						v-if="type === 'messages'"
@@ -62,7 +62,7 @@ each other by animating the opacities.
 						<rect class="message-placeholder-line-one" />
 						<rect class="message-placeholder-line-two" />
 						<rect class="message-placeholder-line-three" />
-						<rect class="message-placeholder-line-four" :style="{width: `width`}" />
+						<rect class="message-placeholder-line-four" :style="width" />
 					</svg>
 				</li>
 			</ul>
@@ -97,7 +97,7 @@ export default {
 			const data = []
 			for (let i = 0; i < this.count; i++) {
 				// generate random widths
-				data.push(Math.floor(Math.random() * 20) + 30)
+				data.push('width: ' + (Math.floor(Math.random() * 20) + 30) + '%')
 			}
 			return data
 		},
--- a/src/components/LoadingPlaceholder.vue
+++ b/src/components/LoadingPlaceholder.vue
@@ -18,53 +18,56 @@
   - You should have received a copy of the GNU Affero General Public License
   - along with this program. If not, see <http://www.gnu.org/licenses/>.
 -->
+<docs>
+
+Displays a loading placeholder for conversation messages.
+
+The gradient animation is achieved by having two placeholder elements,
+with opposite gradient directions (regular and reverse) displayed on top
+of each other (overlapped with position: absolute) and then fading between
+each other by animating the opacities.
+
+</docs>
 
 <template>
-	<ul>
+	<div class="placeholder-main">
 		<!-- Placeholder animation -->
-		<svg class="placeholder-gradient">
-			<defs>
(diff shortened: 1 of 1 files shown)


PVince81:
If that feels a bit over the top, I had another local version where I only pulsed a single list. The CPU usage was almost the same, so if we want we could still stick with the gradient+opacity thing.

ma12-co (approved):
Nice, the only thing is that for me the placeholders are aligned to the left instead of centered!

src/components/LoadingPlaceholder.vue
```diff
@@ -20,51 +20,50 @@
...
-							repeatCount="indefinite" />
-					</stop>
-					<stop offset="100%" :stop-color="dark">
-						<animate attributeName="stop-color"
-							:values="`${dark}; ${light}; ${light}; ${dark}; ${dark}`"
-							dur="2s"
-							repeatCount="indefinite" />
-					</stop>
+				<linearGradient id="placeholder-gradient-regular">
+					<stop offset="0%" :stop-color="light" />
```
skjnldsv, line 28: You can't use variables there?

nickvergessen: for `0%` ?

PVince81: adjusted

skjnldsv: Lol no, for `stop-color="var(--color-blabla)"`

skjnldsv: Ah, ignore my comments, I think I misunderstood the code :thinking:

PVince81:
fixed centering and squashed.

I've compared the animation with the old one side by side and couldn't see any significant difference, they look visually the same.

Note: it seems the avatar size and bar alignment doesn't match whatever the actual list has. I attempted to adjust it but failed as the alignment changes with different zoom sizes... If we want alignment identical to the message list we might need to rework the svg markup and use HTML / CSS for alignment instead of CSS calculations. This is out of scope for the current PR. (no quick win sadly)

src/components/LoadingPlaceholder.vue
```diff
@@ -18,53 +18,56 @@
...
+					<svg
+						v-if="type === 'messages'"
+						class="message-placeholder"
+						xmlns="http://www.w3.org/2000/svg"
+						:fill="'url(#placeholder-gradient' + suffix + ')'">
+						<circle class="message-placeholder-icon" />
+						<rect class="message-placeholder-line-one" />
+						<rect class="message-placeholder-line-two" />
+						<rect class="message-placeholder-line-three" />
+						<rect class="message-placeholder-line-four" :style="{width: `width`}" />
```
nickvergessen, line 65: The last line doesn't seem to have a random length anymore.

Something like the following makes it work:
```diff
diff --git a/src/components/LoadingPlaceholder.vue b/src/components/LoadingPlaceholder.vue
index 49acbfc0e..e74f28c58 100644
--- a/src/components/LoadingPlaceholder.vue
+++ b/src/components/LoadingPlaceholder.vue
@@ -51,7 +51,7 @@ each other by animating the opacities.
                                                :fill="'url(#placeholder-gradient' + suffix + ')'">
                                                <circle class="conversation-placeholder-icon" />
                                                <rect class="conversation-placeholder-line-one" />
-                                               <rect class="conversation-placeholder-line-two" :style="{width: `width`}" />
+                                               <rect class="conversation-placeholder-line-two" :style="width" />
                                        </svg>
                                        <svg
                                                v-if="type === 'messages'"
@@ -62,7 +62,7 @@ each other by animating the opacities.
                                                <rect class="message-placeholder-line-one" />
                                                <rect class="message-placeholder-line-two" />
                                                <rect class="message-placeholder-line-three" />
-                                               <rect class="message-placeholder-line-four" :style="{width: `width`}" />
+                                               <rect class="message-placeholder-line-four" :style="width" />
                                        </svg>
                                </li>
                        </ul>
@@ -97,7 +97,7 @@ export default {
                        const data = []
                        for (let i = 0; i < this.count; i++) {
                                // generate random widths
-                               data.push(Math.floor(Math.random() * 20) + 30)
+                               data.push('width: ' + (Math.floor(Math.random() * 20) + 30) + '%')
                        }
                        return data
                },
```

nickvergessen (changes requested):

src/components/LoadingPlaceholder.vue
PVince81: Nice catch. Despite my compare the old and new this didn't catch my eye.

Tested and pushed.

nickvergessen (approved):

Issue #4562, opened 2020-11-09
Raise hand button


PVince81:
**Is your feature request related to a problem? Please describe.**
In a call with many participants, like in a remote school setting, it is better that people don't unmute themselves and just talk as it could cause a lot of talking into each other and can quickly become chaotic.

**Describe the solution you'd like**
A "raise hand" button that one can click before speaking.
The moderator can see who raised their hands and grant them "unmute" rights, maybe "ask to unmute" which the person can confirm.

Ideally I think we should show the raised hand state both in the people's video view and also in the call sidebar.
Then the moderator could click on that icon either in the video view or on the sidebar to grant them speaking rights.

## Initial limitations

- The first iteration will not include a "permanent mute" feature, so people would still be able to unmute instead of raising hands. Permanent mute has a separate ticket: https://github.com/nextcloud/spreed/issues/4565
- Sidebar icon not supported yet for guests as it needs some specific backend support for mapping guests there.

## Technical / design notes

- Add hand icon on the left side of the mute button for both own controls and other people's video views
- Add hand icon on the participant sidebar
- We can use the signaling channels (but not data channels) to send the necessary messages about who is raising hand and not (broadcast to moderators).
   - "ask to unmute" message => https://github.com/nextcloud/spreed/issues/4566
   - "person X raised hand" message
   - "person X lowered hand" message
   - can check how https://github.com/nextcloud/spreed/issues/3015 was done as it uses similar channels
- Whenever a moderator joins the room, we could optionally also query whoever raised their hands already.

## Questions

- should only moderators see raised hands ? I think in other solutions everyone sees that, and that way it also looks like a real classroom or meeting. Also: whenever someone is already speaking if they see raised hands they might finish their topic more quickly to let other people speak.

PVince81:
@nextcloud/designers-talk in case you have input

PVince81:
I've raised the following as separate tickets/features:
- ask to unmute: https://github.com/nextcloud/spreed/issues/4566
- permanent mute: https://github.com/nextcloud/spreed/issues/4565

they somehow go hand in hand for a complete experience.

we can keep the current ticket here only for the hand raising topic

PVince81:
PR here for first prototype: https://github.com/nextcloud/spreed/pull/4569

PVince81:
cc @tobiasKaminsky for clients awareness

so far the technical bits about how the info is sent is not set in stone yet (the prototype PR uses signaling)

[...]

PR #4569, opened 2020-11-09
Added "Raise Hand" feature

PVince81:
Fixes https://github.com/nextcloud/spreed/issues/4562

### Requires

- [x] support/quickfix for material design icon in ActionButton: https://github.com/nextcloud/nextcloud-vue/pull/1614

### Tasks

- [x] Added hand and hand-off icons (need better fitting ones later)
- [x] Added "raise-hand" capability
- [x] State changes for raising/lowering hand is broadcast to other participants
- [x] Display notification (toast) to ~~moderators~~ participants when someone raised a hand
- Inform newly joined participants of current raised hands => won't, see https://github.com/nextcloud/spreed/pull/4569#issuecomment-725983663
- [x] Show hand raised icon to sidebar participant list (when raised)
- ~~Add chat message "XYZ raised their hand" in the history ? Only for call people or everyone ? (if possible)~~ => probably not, would need backend state, and is likely not that useful
- ~~Disable hand raising in 1-1 calls (or if participants count <= 2)~~
   - => tried it, doesn't feel natural. and one could always be presenting something
   - one can also do a presentation to only one participant
   - the hand action is anyway now in the three dots menu, so doesn't disturb any more visually
   - implementing this means we need to automatically lower all hands when participant count is less than 3
   - sometimes one might want to raise the hand before the presenter joins, or the presenter just lost connection and is re-joining
- [x] Sort participant list by "raised hand time" first, and then non-raised-hand participants
- [x] ~~Hide local hand button if there is no local stream (as we currently can only broadcast when there are streams)~~ => not needed, there will be a fix for this soon
- [x] raising hand from SIP: https://github.com/nextcloud/spreed/pull/4496#issuecomment-721765591
- [x] Test with internal signaling
- [x] Test with HPB
- [ ] Test with phone users (SIP)
- [x] Tweak UX design

### Issues

- [x] broadcasting is unreliable (tested with internal signaling)
   - only Firefox seems to receive messages
   - not sure if the way I send the message is correct and whether we need to pass along the peer id and verify it (if available, as peers require streams AFAIK) @danxuliu I'll need your advice here

- [x] clicking "raise hand" doesn't toggle the hand icon

- [x] current icon design not fitting / not consistent

~~The hand is striked out differently. This is the [do not touch icon from Material Design](https://material.io/resources/icons/?search=hand&icon=do_not_touch&style=baseline).~~ obsolete, see newer screenshots below

- [x] clicking directly on the pixels of the three dots of the new menu doesn't do anything in some layouts, but clicking the border does... in any case it seems the actions menu has some issues with hover, cursor, etc...

- [x] bug: when leaving with hand raised, rejoining should have it lowered again. Probably LocalMediaModel is not cleared when leaving (approach might be obsolete with https://github.com/nextcloud/spreed/pull/4569#issuecomment-734757623)

- [x] use store value in video bottom bar vue? it seems the unwatcher has a delay compared to the model attribute
- [x] fix hand icon alignment and spacing in VideoBottomBar
image


src/utils/webrtc/models/CallParticipantModel.js
```diff
@@ -240,6 +242,12 @@ CallParticipantModel.prototype = {
 		}
 	},
 
+	_handleRaisedHand: function(data) {
+		// FIXME: rely on peer id ?
```
PVince81, line 246: @danxuliu should we check the peer id like for other methods ?

As far as I understood a peer id is only available if there's at least an audio or video stream ?

src/utils/webrtc/simplewebrtc/simplewebrtc.js
```diff
@@ -118,6 +118,12 @@ function SimpleWebRTC(opts) {
 				} else {
 					self.emit('mute', { id: message.payload.peerId })
 				}
+			} else if (message.payload.action === 'raiseHand') {
+				if (message.from === self.connection.getSessionId()) {
```
PVince81, line 122: Since we don't have the peer id here, I saw that "from" seems to contain an id. But it feels wrong as it didn't seem to map to peers.

This is different than "forceMute" as it's one participant sending to everyone else, not one moderator broadcasting to all with a specific peer id to mute.

@danxuliu any advice here ?

fancycode:
Would be great to show that also in the participant list. For phone-only users there won't be a video component and they should be able to raise their hand, too.

PVince81:
> Would be great to show that also in the participant list

Indeed, forgot to add it. Added as checkbox now.

PVince81:
silly idea: animate the hand icon, making it wave (CSS 3 rotate / skew), while the hand is raised

PVince81:
I've changed the message type from "control" to "raiseHand" and now the broadcasting work.

Now need to find out why the views are not updating correctly.

PVince81:
Okay, the basics work now (tested with internal signaling):
- clicking hand icon will toggle it
- peers will see a hand icon on the left of the display name whenever a remote hand is raised

ma12-co:
I think that this action should be buried in 3-dot menu in the bottom right corner of the localvideo, together with a link to the global audio and video settings.

ma12-co (commented):
I think that this action should be buried in 3-dot menu in the bottom right corner of the localvideo, together with a link to the global audio and video settings.

PVince81:
adding a new three dots menu in the local controls sounds like an idea.

I've seen variants where additionally to raising hand it was also possible to set a mood instead of the hand, like "smiling", "unhappy", "confused", "thumbs up", "thumbs down" to silently signal something to the presenter. Having a menu would make room for those.

ma12-co:
As per the icon, this would do

https://materialdesignicons.com/icon/hand-right

We don't need a "hand down" icon, we can just display this one when the hand is raised.

ma12-co:
> adding a new three dots menu in the local controls sounds like an idea.

Plus this is a borderline niche feature, so having it explicitly displayed by default could be distracting

nickvergessen:
>  Inform newly joined participants of current raised hands

wouldnt do that, as this requires a global state or something, instead of being a signaling message. e.g. with sip dialin this wont work

PVince81:
> > Inform newly joined participants of current raised hands
> 
> wouldnt do that, as this requires a global state or something, instead of being a signaling message. e.g. with sip dialin this wont work

We already do that for the muted state so we can show the icon. But it requires every participant to detect whenever a new person joined and then send them their state. Could get messy with a lot of participants.

PVince81:
I've moved the hand action into a new three dots menu:
image

When the hand is raised, there will be an extra button in the controls to show you so, and clicking it will lower the hand again and make it disappear:
image

PVince81:
- [ ] BUG: hand icon shortcut "h" key doesn't do anything, probably because it's in the menu
   - should we debounce it as well to avoid abuse ?

PVince81:
- [ ] clicking directly on the pixels of the three dots of the new menu doesn't do anything in some layouts, but clicking the border does... in any case it seems the actions menu has some issues with hover, cursor, etc...

fancycode:
Can we somehow reuse the flags introduced for "talking" in the SIP integration to signal the "raised hand" state?
https://github.com/nextcloud/spreed/pull/4496#issuecomment-721765591

Right now it seems to use datachannels which we try to get rid of (#397).

PVince81:
- [ ] reuse flags for hand raise event

Note that having a boolean flag will make it difficult to add other symbols like thumbs up / down in the future. But maybe we can rework this once we walk that path.

PVince81:
- [ ] raise enhancement: ability to follow a participant after clicking on them in the participant list.
   - would be useful so that when someone raises hand, people could click on that person to  make their video visible

PVince81:
- [ ] bug: when leaving with hand raised, rejoining should have it lowered again. Probably LocalMediaModel is not cleared when leaving (approach might be obsolete with https://github.com/nextcloud/spreed/pull/4569#issuecomment-734757623)

- [ ] use store value in video bottom bar vue? it seems the unwatcher has a delay compared to the model attribute

PVince81:
hand icon is now displayed in the participant list as per https://github.com/nextcloud/spreed/pull/4569/commits/9c1db586a98acd84796fea9c671be3b23b6adda2:

here we see Bob raised their hand, left side is admin and right side is Bob:
Screenshot_20201127_115503

- [ ] consolidate hand icons ? (`<Hand>` vs "icon-hand")

src/components/CallView/CallView.vue
```diff
@@ -404,6 +404,7 @@ export default {
 				this.raisedHandUnwatchers[removedModelId]()
 				// Not reactive, but not a problem
 				delete this.raisedHandUnwatchers[removedModelId]
+			    this.$store.dispatch('setParticipantHandRaised', { peerId: removedModelId, raised: false })
```
PVince81, line 407: eslint didn't see this tab/space mix... weird... will fix in next push

PVince81:
- [ ] BUG: hand icon in videos only visible after hover in grid view, like the mute icon. Might need to move it to another layer :-/

PVince81:
- [ ] check if we could use the peers store instead to read the hand state (which I just discovered)

src/components/CallView/CallView.vue
```diff
@@ -465,6 +479,27 @@ export default {
...
 		},
 
+		_handleParticipantRaisedHand(callParticipantModel, raisedHand) {
+			const nickName = callParticipantModel.attributes.name || callParticipantModel.attributes.userId
+			// sometimes the nick name is not available yet...
+			if (nickName) {
+				if (raisedHand) {
+					showMessage(t('spreed', 'Participant {nickName} raised their hand.', { nickName: nickName }))
+				} else {
+					showMessage(t('spreed', 'Participant {nickName} lowered their hand.', { nickName: nickName }))
```
nickvergessen, line 489: I think we shouldn't put out a toast notification for this. You are grabbing attention that someone does not want attention anymore... sounds weird.

img/hand-off.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" enable-background="new 0 0 24 24" height="24" viewBox="0 0 24 24" width="24"><g><rect fill="none" height="24" width="24"/><path d="M13,10.17l-2.5-2.5V2.25C10.5,1.56,11.06,1,11.75,1S13,1.56,13,2.25V10.17z M20,12.75V11V5.25C20,4.56,19.44,4,18.75,4 S17.5,4.56,17.5,5.25V11h-1V3.25C16.5,2.56,15.94,2,15.25,2S14,2.56,14,3.25v7.92l6,6V12.75z M9.5,4.25C9.5,3.56,8.94,3,8.25,3 c-0.67,0-1.2,0.53-1.24,1.18L9.5,6.67V4.25z M13,10.17l-2.5-2.5V2.25C10.5,1.56,11.06,1,11.75,1S13,1.56,13,2.25V10.17z M20,12.75 V11V5.25C20,4.56,19.44,4,18.75,4S17.5,4.56,17.5,5.25V11h-1V3.25C16.5,2.56,15.94,2,15.25,2S14,2.56,14,3.25v7.92l6,6V12.75z M9.5,4.25C9.5,3.56,8.94,3,8.25,3c-0.67,0-1.2,0.53-1.24,1.18L9.5,6.67V4.25z M21.19,21.19L2.81,2.81L1.39,4.22l5.63,5.63L7,9.83 v4.3c-1.11-0.64-2.58-1.47-2.6-1.48c-0.17-0.09-0.34-0.14-0.54-0.14c-0.26,0-0.5,0.09-0.7,0.26C3.12,12.78,2,13.88,2,13.88 l6.8,7.18c0.57,0.6,1.35,0.94,2.18,0.94H17c0.62,0,1.18-0.19,1.65-0.52l-0.02-0.02l1.15,1.15L21.19,21.19z"/></g></svg>
```
nickvergessen, line 1: Are there no icons in material design we could use? Would make it much easier for the mobile apps to follow.
cc @jancborchardt any icon suggestion?

PVince81: seems I took it from material.io not material design... I checked https://materialdesignicons.com and there are a few interesting icons:
- a raised hand
- a left hand
- a right hand

I think we could use the raised hand as the indicator and a "left hand" (or right) for the action.

PVince81: I tried making the switch but have some troubles with the menu item: https://github.com/nextcloud/nextcloud-vue/issues/1613

PVince81: I made the switch, will need some extra fix in nextcloud-vue

src/components/CallView/CallView.vue
PVince81: I also wasn't sure, was thinking of visually impaired people who might want to get notified. But it would likely be overwhelming for them as well.

Removed.

PVince81:
moved to material design icons.

the menu icon for "raise hand" is not aligned, needs PR in nextcloud-vue: https://github.com/nextcloud/nextcloud-vue/pull/1614

PVince81:
> Can we somehow reuse the flags introduced for "talking" in the SIP integration to signal the "raised hand" state?
> [#4496 (comment)](https://github.com/nextcloud/spreed/pull/4496#issuecomment-721765591)
> 
> Right now it seems to use datachannels which we try to get rid of (#397).

I think I'd like to send it through the signaling server after https://github.com/nextcloud/spreed/pull/4686 is merged, which will make it work also for participants without peers. Main reason is that a simple flag will not be extensible if we ever decide to go further than just hand raising, with other statuses like "thumbs up", "thumbs down", etc.

PVince81:
current state:
image

PVince81:
- [ ] add `.once` to short key

src/components/CallView/shared/LocalMediaControls.vue
```diff
@@ -82,6 +81,42 @@
 					</li>
 				</ul>
 			</div>
+			<button
+				v-shortkey="['h']"
```
nickvergessen, line 85: Jitsi uses `r` for `raise hand`.
Can someone check teams? If they also use `r` we should use `r` so we are inline with expectations of users.

src/components/CallView/shared/VideoBottomBar.vue
```diff
@@ -24,6 +24,14 @@
 		<div v-if="!isSidebar"
 			class="bottom-bar"
 			:class="{'bottom-bar--video-on' : hasShadow, 'bottom-bar--big': isBig }">
+			<div class="bottom-bar__statusIndicator">
+				<transition name="fade">
+					<Hand
+						class="wave-hand-animation"
+						v-if="!connectionStateFailedNoRestart && model.attributes.raisedHand"
```
nickvergessen, line 31: Attribute "v-if" should go before "class"

nickvergessen: I guess the if should even be on `<div class="bottom-bar__statusIndicator">`?

src/components/SettingsDialog/SettingsDialog.vue
```diff
@@ -88,6 +88,12 @@
 						{{ t('spreed', 'Push to talk or push to mute') }}
 					</dd>
 				</div>
+				<div>
+					<dt><kbd>H</kbd></dt>
+					<dd class="shortcut-description">
+						{{ t('spreed', 'Raise/lower hand') }}
```
nickvergessen, line 94:
```suggestion
						{{ t('spreed', 'Raise or lower hand') }}
```

src/components/CallView/shared/LocalMediaControls.vue
```diff
@@ -503,6 +556,21 @@ export default {
...
+			const raisedHand = !this.model.attributes.raisedHand
+			if (this.raisingHandNotification) {
+				this.raisingHandNotification.hideToast()
+				this.raisingHandNotification = null
+			}
+			if (raisedHand) {
+				this.raisingHandNotification = showMessage(t('spreed', 'You are now raising your hand.'))
+			} else {
+				this.raisingHandNotification = showMessage(t('spreed', 'You are no longer raising your hand.'))
+			}
```
nickvergessen, line 569: I'm not sure I would show a message for this at all.

src/components/CallView/shared/LocalMediaControls.vue
```diff
@@ -82,6 +81,42 @@
 					</li>
 				</ul>
 			</div>
+			<button
```
nickvergessen, line 84: clicking the button focuses your own video :rofl:

PVince81:
- [ ] fix hand icon alignment and spacing in VideoBottomBar
image

src/components/CallView/shared/LocalMediaControls.vue
PVince81: this is mostly visual feedback for an icon, for example if you'd use the keyboard shortcut.
on the other hand, for this action specifically you'd see the hand disappear at the bottom

I'll remove it

PVince81:
I've fixed all outstanding issues, so I think this is ready for review.

The PR from @danxuliu https://github.com/nextcloud/spreed/pull/4686 is not needed for testing, but need to make sure there's at least an audio peer for the raise hand signal to go through. Later on it will work also without peer. As far as I saw I'm already using the correct method that will be adjusted in https://github.com/nextcloud/spreed/pull/4686, so I expect minimal to no changes needed after that PR is merged.

src/components/CallView/shared/LocalMediaControls.vue
PVince81: I've changed it to "r" for now. we can change it later once someone can check on teams

docs/capabilities.md
```diff
@@ -56,8 +56,9 @@ title: Capabilities
...
-* `config => chat => read-privacy` - See `chat-read-status`
-* `phonebook-search` - Is present when the server has the endpoint to search for phone numbers to find matches in the accounts list
 * `listable-rooms` - Conversations can searched for even when not joined. A "listable" attribute set on rooms defines the scope of who can find it.
+* `phonebook-search` - Is present when the server has the endpoint to search for phone numbers to find matches in the accounts list
+* `raise-hand` - Participants can raise or lower hand, the state change is sent through signaling messages.
+* `config => chat => read-privacy` - See `chat-read-status`
+* `config => previews => max-gif-size` - Maximum size in bytes below which a GIF can be embedded directly in the page at render time. Bigger files will be rendered statically using the preview endpoint instead. Can be set with `occ config:app:set spreed max-gif-size --value=X` where X is the new value in bytes. Defaults to 3 MB.
```
nickvergessen, line 64: alphabetic and features first? Fine by me, just have to remember it in the future

src/components/CallView/shared/VideoBottomBar.vue
```diff
@@ -24,6 +24,15 @@
 		<div v-if="!isSidebar"
 			class="bottom-bar"
 			:class="{'bottom-bar--video-on' : hasShadow, 'bottom-bar--big': isBig }">
+			<transition name="fade">
+				<div
+					v-if="!connectionStateFailedNoRestart && model.attributes.raisedHand"
+					class="bottom-bar__statusIndicator">
+					<Hand
```
nickvergessen, line 31: decorative + title?

src/components/RightSidebar/Participants/CurrentParticipants/CurrentParticipants.vue
```diff
@@ -145,6 +145,12 @@ export default {
 				return p1inCall ? -1 : 1
 			}
 
+			const p1HandRaised = this.$store.getters.isParticipantRaisedHand(participant1.sessionId)
+			const p2HandRaised = this.$store.getters.isParticipantRaisedHand(participant2.sessionId)
+			if (p1HandRaised !== p2HandRaised) {
+				return p1HandRaised ? -1 : 1
```
nickvergessen, line 151: Makes sense. the problem is if sort by that I would assume it's by order of the raising and not the remaining things?
But we don't have a timestamp or something :/

src/components/CallView/CallView.vue
```diff
@@ -467,6 +481,23 @@ export default {
 			}
 		},
 
+		_handleParticipantRaisedHand(callParticipantModel, raisedHand) {
+			const nickName = callParticipantModel.attributes.name || callParticipantModel.attributes.userId
+			// sometimes the nick name is not available yet...
+			if (nickName) {
+				if (raisedHand) {
+					showMessage(t('spreed', 'Participant {nickName} raised their hand.', { nickName: nickName }))
```
nickvergessen, line 489:
```suggestion
					showMessage(t('spreed', '{nickName} raised their hand.', { nickName: nickName }))
```

src/components/CallView/CallView.vue
```diff
@@ -467,6 +481,23 @@ export default {
...
+				}
+			} else {
+				if (raisedHand) {
+					showMessage(t('spreed', 'A participant raised their hand.'))
```
nickvergessen, line 493:
```suggestion
					showMessage(t('spreed', 'A guest raised their hand.'))
```

src/components/RightSidebar/Participants/CurrentParticipants/CurrentParticipants.vue
PVince81: right. in theory we can add a timestamp but it wouldn't be authoritative as it would be based on whenever the signal/event is received, so everyone might see something different

unless we send the timestamp from the sender and use that instead (would be easy to cheat but don't think we should care)

docs/capabilities.md
PVince81: Yeah. After several rebases I saw stuff being added one by one and it became messy. Looks better this way :-)

src/components/RightSidebar/Participants/CurrentParticipants/CurrentParticipants.vue
nickvergessen: > unless we send the timestamp from the sender and use that instead (would be easy to cheat but don't think we should care)

sounds good enough. Is that easily doable?

PVince81: yeah, I'll have a try.

there is another issue to look into: it seems reactivity is not fully working from inside this sort function, probably because accessing the hand state is "too remote".
I had cases where the list didn't re-sort automatically, it only did after a full refresh

PVince81: as discussed in chat: I'll remove the sorting for now and we can investigate separately

PVince81:
I've removed the sorting thing and rebased.

Also adjusted the icon with decorative and empty title

nickvergessen:
Can you `git rm tags*` please :)

nickvergessen (approved):

PR #4582, opened 2020-11-12
Improve date separators

ma12-co:
| Before      | After |
| ----------- | ----------- |
|   Screenshot from 2020-11-12 11-21-20 | Screenshot from 2020-11-12 11-20-54   |

### DIFF
--- a/src/components/MessagesList/MessagesGroup/MessagesGroup.vue
+++ b/src/components/MessagesList/MessagesGroup/MessagesGroup.vue
@@ -141,25 +141,17 @@ export default {
 	&__date-header {
 		display: block;
 		text-align: center;
-
-		margin: 40px 15px 0;
-		border-top: 1px solid var(--color-border);
 		padding-top: 20px;
 		position: relative;
+		margin: 20px 0;
 
 		.date {
 			content: attr(data-date);
-			position: absolute;
-			top: 0;
-			left: 50%;
-			transform: translateX(-50%) translateY(-50%);
-			padding: 0 7px 0 7px;
-
-			text-align: center;
-			white-space: nowrap;
-
+			background-color: var(--color-background-dark);
+			font-weight: bold;
+			padding: 6px 12px;
 			color: var(--color-text-maxcontrast);
-			background-color: var(--color-main-background);
+			border-radius: var(--border-radius-pill);
 		}
 	}
 }


src/components/MessagesList/MessagesGroup/MessagesGroup.vue
```diff
@@ -141,25 +141,17 @@ export default {
...
-			top: 0;
-			left: 50%;
-			transform: translateX(-50%) translateY(-50%);
-			padding: 0 7px 0 7px;
-
-			text-align: center;
-			white-space: nowrap;
-
+			background-color: var(--color-background-dark);
+			font-weight: bold;
```
nickvergessen, line 151: dont like the bold thing. it's not too important.

src/components/MessagesList/MessagesGroup/MessagesGroup.vue
```diff
@@ -141,25 +141,17 @@ export default {
...
+			padding: 6px 12px;
```
nickvergessen, line 152: Also I would reduce the padding and make it is small as possible (similar to the mention chips).

nickvergessen:
Would it be possible to make the last date sticky? so when scrolling into the past you can see which day you are looking at?

src/components/MessagesList/MessagesGroup/MessagesGroup.vue
ma12-co: It's pretty important, and it's only once a day tops!

src/components/MessagesList/MessagesGroup/MessagesGroup.vue
ma12-co: Mentions don't have sufficient padding for their font size. If I'm to reduce the padding I'd have to reduce the font size too, which I'm not too fond of...

ma12-co:
> Would it be possible to make the last date sticky? so when scrolling into the past you can see which day you are looking at?

good Idea!

jancborchardt:
Agree with @nickvergessen on both points:
- Let’s make the last date sticky, that’s a definite issue with the current date separator :)
- The bold is too present – it makes the date divider the one of the most important things, which it absolutely is not. 
 → As extension of that, the pill button style on the divider makes it even more present and even makes it look interactive, way more present and important than it should be.

Additional input:
- The text being set in `var(--color-text-maxcontrast)` and regular font size is the proper amount of importance.
- Removing the border line is fine, maybe we need to check if the amount of whitespace is good (should also not be too much because then it will look weird if you chat at midnight).
- We need to add a `--color-main-background-translucent` to the date divider so that when it’s sticky, it doesn’t get underlapped by the text.

ma12-co:
So @jancborchardt no change other the stickyness of the current thing?

ma12-co:
Closed this and opened an issue for that #4610

PR #4584, opened 2020-11-12
New message form facelift

ma12-co:
Peek 2020-11-12 14-27

### DIFF
--- a/src/components/ChatView.vue
+++ b/src/components/ChatView.vue
@@ -45,9 +45,12 @@
 		<MessagesList
 			role="region"
 			:aria-label="t('spreed', 'Conversation messages')"
-			:token="token" />
+			:is-chat-scrolled-to-bottom="isChatScrolledToBottom"
+			:token="token"
+			@setChatScrolledToBottom="setScrollStatus" />
 		<NewMessageForm
 			role="region"
+			:is-chat-scrolled-to-bottom="isChatScrolledToBottom"
 			:aria-label="t('spreed', 'Post message')" />
 	</div>
 </template>
@@ -77,6 +80,12 @@ export default {
 	data: function() {
 		return {
 			isDraggingOver: false,
+			/**
+			 * Initialised as true as when we open a new conversation we're scrolling to
+			 * the bottom for now. In the future when we'll open the conversation close
+			 * to the scroll position of the last read message, we will need to change this.
+			 */
+			isChatScrolledToBottom: true,
 		}
 	},
 
@@ -127,6 +136,10 @@ export default {
 			const uploadId = new Date().getTime()
 			// Uploads and shares the files
 			processFiles(files, this.token, uploadId)
+		},
+
+		setScrollStatus(payload) {
+			this.isChatScrolledToBottom = payload
 		},
 	},
 
--- a/src/components/ChatView.vue
+++ b/src/components/ChatView.vue
@@ -165,7 +165,7 @@ export default {
 	background: var(--color-primary-light);
 	z-index: 11;
 	display: flex;
-	box-shadow: 0px 0px 36px var(--color-box-shadow);
+	box-shadow: 0 0 36px var(--color-box-shadow);
 	border-radius: var(--border-radius);
 	opacity: 90%;
 }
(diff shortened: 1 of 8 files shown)


src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -420,6 +427,15 @@ export default {
 			margin-top: auto;
 			background-color: transparent;
 			border: none;
+			margin: 0 4px;
+			color: var(--color-main-text);
+			opacity: .9;
+		}
+		&__quote {
+			margin: 0px 16px 12px 24px;
```
nickvergessen, line 435:
```suggestion
			margin: 0 16px 12px 24px;
```

ma12-co:
@joas pointed out that there are too many shades of gray. I could remove the circle around the X button when it's not hovered so it's bee one less shade. What do you think?

jancborchardt:
As briefly mentioned in our call, I would say if we change it, then I’d expect the _input field_ to be grey, not the block around it.

Because the current style (please always do before / after comparison screenshots :) actually looks very light and nice like @jenniferpiperek designed it, and changing to having a whole grey block down there is a bit massive. (As @nickvergessen also pointed out.)

(And then again, input fields are usually not grey but always white with a border around it, so I see a high risk in breaking that convention. On first glance it would look like a disabled button or disabled input field more than a regular input field.)

ma12-co:
> input fields are usually not grey but always white with a border around it, so I see a high risk in breaking that convention.

Pushed a second attempt with border on the input field

Peek 2020-12-18 13-13

jancborchardt:
Looks better, and then we don't need the additional horizontal line anymore? Could either be done with a fade to white, or only showing the line when not scrolled all the way to the bottom (that's what e.g. Signal does).

ma12-co:
Done it the signal way, looks pretty darn good to me!

Peek 2020-12-29 11-36

jancborchardt:
Looking great now! :) The commits seem messed up though, as e.g. also the button-looking date dividers which we don’t want are in here?

ma12-co:
> Looking great now! :) The commits seem messed up though, as e.g. also the button-looking date dividers which we don’t want are in here?

Yep needs some cleanup. 
It's based on the date-buttons branch because of the messagelist alignment issues fixed there. We don't want them? I thought that the discussion was still open, [please see the updates!](#4779 )

src/components/ChatView.vue
```diff
@@ -77,6 +80,12 @@ export default {
 	data: function() {
 		return {
 			isDraggingOver: false,
+			/**
+			 * Initialised as true as when we open a new conversation we're scrolling to
+			 * the bottom for now. In the future when we'll open the conversation close
+			 * to the scroll position of the last red message, we wil need to change this.
```
nickvergessen, line 86:
```suggestion
			 * to the scroll position of the last read message, we will need to change this.
```

ma12-co: done!

src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -96,8 +98,12 @@
 					:disabled="isReadOnly"
 					type="submit"
 					:aria-label="t('spreed', 'Send message')"
-					class="new-message-form__button submit icon-confirm-fade"
-					@click.prevent="handleSubmit" />
+					class="new-message-form__button submit"
+					@click.prevent="handleSubmit">
+					<Send
+						:size="20"
```
nickvergessen, line 104: no title, so its always "Send" instead of being translated

src/components/NewMessageForm/AdvancedInput/AdvancedInput.vue
```diff
@@ -126,7 +126,7 @@ function isDefaultAtWhoStyleSheet(sheet) {
 			// rules. Moreover, a rather complex rule is needed so the style
 			// sheet is not mismatched with a different atwho stylesheet, for
 			// example, the one added by the Comments app.
-			if (cssRule.cssText === '.atwho-view { color: rgb(0, 0, 0); border-radius: 3px; box-shadow: rgba(0, 0, 0, 0.1) 0px 0px 5px; min-width: 120px; z-index: 11110 !important; }') {
+			if (cssRule.cssText === '.atwho-view { color: rgb(0, 0, 0); border-radius: 3px; box-shadow: rgba(0, 0, 0, 0.1) 0 0 5px; min-width: 120px; z-index: 11110 !important; }') {
```
nickvergessen, line 129: this breaks

nickvergessen:
```suggestion
			if (cssRule.cssText === '.atwho-view { color: rgb(0, 0, 0); border-radius: 3px; box-shadow: rgba(0, 0, 0, 0.1) 0px 0px 5px; min-width: 120px; z-index: 11110 !important; }') {
```

Issue #4516, opened 2020-11-02
Highlight the original message when scrolling on a reply


nickvergessen:
## Steps to reproduce
1. Click on the original message part of a reply
> ![Bildschirmfoto von 2020-11-02 13-05-13](https://user-images.githubusercontent.com/213943/97866552-6cf8f480-1d0c-11eb-8d0b-227aa4b935cf.png)
2. Chat scrolls up to original message, but (due to mouse position) a "random" message is highlighted instead of the message you clicked on.
> ![Bildschirmfoto von 2020-11-02 13-05-18](https://user-images.githubusercontent.com/213943/97866632-9023a400-1d0c-11eb-8230-c556ec1eeb5a.png)

### Expected behaviour
The message you clicked on should be highlighted (maybe even more than a mouse hover and "sticky" (so hovering a different message does not loose it)

I also currently find the "message is scrolled to the middle" part confusing, but the correct (forced) highlight will solve this.

### Actual behaviour
Message under the cursor is highlighted.

PVince81:
Seems this would introduce the concept of message highlighting that sticks (unless this is a regression, but I haven't seen such highlight before except on hober)

how long to highlight it ?
should clicking elsewhere remove the highlight ?
should we also allow to highlight by simply clicking on a message ?

nickvergessen:
No strong feelings, something like https://stackoverflow.com/a/64617156 would be good enough from my pov

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/4585

I've used the same highlight style like when highlighting uploaded files in the files app.

PVince81:
Would be great to have the same effect when coming from search.
So far I haven't found the code path that reads the URL param with the message id.

PVince81:
> Would be great to have the same effect when coming from search.
> So far I haven't found the code path that reads the URL param with the message id.

added to my PR: https://github.com/nextcloud/spreed/pull/4585

PR #4585, opened 2020-11-12
Highlight quote after scrolling to it

PVince81:
Fixes https://github.com/nextcloud/spreed/issues/4516

When clicking on a quote, we scroll to it and now there's a new
highlight animation that fades out

The animation and scrolling logic was moved from the Quote vue to the
Message vue. The Quote vue will now only send an event to tell that we
want to focus a specific message.

A DOM-level method was used for triggering the highlight to be able to
access it with document.getElementById() as it's cheaper than trying to
look up the message by id within the collection of MessagesList.

If the latter is not wanted we should then rework the messages structure inside of MessagesList to have the messages in an associative array with message id as key. Or build yet another index-like structure to be able to find the messages. So far for this use case it feels a bit like overkill.

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -457,9 +457,12 @@ export default {
 	margin: -6px 0;
 }
 
+.hover, .highlight-animation {
+	border-radius: 8px;
+}
+
 .hover {
 	background-color: var(--color-background-hover);
-	border-radius: 8px;
 }
 
 .highlight-animation {
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -463,12 +463,12 @@ export default {
 }
 
 .highlight-animation {
-	animation: highlight-animation 2s 1;
+	animation: highlight-animation 5s 1;
 }
 
-// TODO: define these colors globally (these are borrowed from the files app upload.css)
 @keyframes highlight-animation {
-	0% { background-color: rgba(255, 255, 140, 1); }
-	100% { background-color: rgba(0, 0, 0, 0); }
+	0% { background-color: var(--color-background-hover); }
+	50% { background-color: var(--color-background-hover); }
+	100% { background-color: rgba(var(--color-background-hover), 0); }
 }
 </style>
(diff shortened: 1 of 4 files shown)


src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -439,4 +461,14 @@ export default {
...
 	border-radius: 8px;
 }
+
+.highlight-animation {
+	animation: highlight-animation 2s 1;
+}
+
+// TODO: define these colors globally (these are borrowed from the files app upload.css)
+@keyframes highlight-animation {
+	0% { background-color: rgba(255, 255, 140, 1); }
```
nickvergessen, line 471: Need to introduce scss vars in server so it's themed somehow and dark aware

PVince81: I now simply switched to the existing hover variable

PVince81:
More improvements:
- switched to existing hover variable
- increased animation time
- fixed "click on quote in new message form"
- enhancement: when coming from search, if the URL anchor contains a message id, it is now also highlighted!

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -439,4 +461,14 @@ export default {
 	background-color: var(--color-background-hover);
 	border-radius: 8px;
 }
+
+.highlight-animation {
```
nickvergessen, line 465: missing the border-radius compared to the hover?

nickvergessen (changes requested):
Looks good apart from the missing border radius

src/components/MessagesList/MessagesGroup/Message/Message.vue
PVince81: I didn't intend to add one, but you're right that it might look better.
I'll add it

PVince81:
Now that https://github.com/nextcloud/spreed/issues/4592 is merged the behavior might change slightly.
It is possible that I'll need to add extra handling since now we have a scenario where the page doesn't reload.

- [ ] recheck "click search result for the current conversation" behavior

src/components/MessagesList/MessagesGroup/Message/Message.vue
PVince81: Added

PVince81:
I'm not done yet... Something is fishy: it should be possible to smooth scroll despite the "browser's anchor logic teleports to location before we get a chance to smooth scroll"

PVince81:
I managed to improve it a bit with a tiny delay hack.

At least now when clicking the search result the entry will end up in the middle of the page, even though the scroll direction might be wrong. The best fix would be to add a "preventDefault" somehow on the search result link... Something for another time once I get to study a bit how the latter works.

nickvergessen (approved):

PR #4588, opened 2020-11-13
Fix casing of IURLGenerator to fix DI

rullzer:
Seems the non deprecated way we use is https://github.com/nextcloud/server/pull/24037
is more case sensitive.

In any case. This hopefully fixes it.


danxuliu (approved):
Tested and works :+1: (but let's wait for the full CI to run).

Thanks!

nickvergessen (approved):

nickvergessen:
okay then no backport to 19

PR #4589, opened 2020-11-13
Test against oracle

nickvergessen:
- [x] Based on #4735

### DIFF
--- a/lib/Manager.php
+++ b/lib/Manager.php
@@ -242,8 +242,9 @@ class Manager {
 
 	public function resetAssignedSignalingServers(ICache $cache): void {
 		$query = $this->db->getQueryBuilder();
-		$query->select('r.*')
-			->from('talk_rooms', 'r')
+		$helper = new SelectHelper();
+		$helper->selectRoomsTable($query);
+		$query->from('talk_rooms', 'r')
 			->where($query->expr()->isNotNull('r.assigned_hpb'));
 
 		$result = $query->execute();
@@ -354,9 +355,9 @@ class Manager {
 			$allowedListedTypes[] = Room::LISTABLE_USERS;
 		}
 		$query = $this->db->getQueryBuilder();
-		$query->select('r.*')
-			->selectAlias('r.id', 'r_id')
-			->from('talk_rooms', 'r')
+		$helper = new SelectHelper();
+		$helper->selectRoomsTable($query);
+		$query->from('talk_rooms', 'r')
 			->leftJoin('r', 'talk_attendees', 'a', $query->expr()->andX(
 				$query->expr()->eq('a.actor_id', $query->createNamedParameter($userId)),
 				$query->expr()->eq('a.actor_type', $query->createNamedParameter(Attendee::ACTOR_USERS)),
--- a/lib/Manager.php
+++ b/lib/Manager.php
@@ -120,9 +120,9 @@ class Manager {
 
 	public function forAllRooms(callable $callback): void {
 		$query = $this->db->getQueryBuilder();
-		$query->select('*')
+		$query->select('r.*')
 			->selectAlias('id', 'r_id')
-			->from('talk_rooms');
+			->from('talk_rooms', 'r');
 
 		$result = $query->execute();
 		while ($row = $result->fetch()) {
@@ -241,9 +241,9 @@ class Manager {
 
 	public function resetAssignedSignalingServers(ICache $cache): void {
 		$query = $this->db->getQueryBuilder();
-		$query->select('*')
-			->from('talk_rooms')
-			->where($query->expr()->isNotNull('assigned_hpb'));
+		$query->select('r.*')
+			->from('talk_rooms', 'r')
+			->where($query->expr()->isNotNull('r.assigned_hpb'));
 
 		$result = $query->execute();
 		while ($row = $result->fetch()) {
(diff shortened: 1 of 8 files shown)


PVince81:
A nice start...
```
Nextcloud was successfully installed
An exception occurred while executing 'INSERT INTO "oc_talk_commands" ("app", "command", "name", "script", "response", "enabled") VALUES(?, ?, ?, ?, ?, ?)' with params ["", "help", "talk", "", 1, 3]:

ORA-01400: cannot insert NULL into ("AUTOTEST"."oc_talk_commands"."script")
```

nickvergessen:
- [ ] Missing "update" step for the help command

lib/Migration/Version11000Date20201209142525.php
```diff
@@ -103,7 +103,7 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 			]);
 
 			$table->setPrimaryKey(['id']);
-			$table->addUniqueIndex(['session_hash'], 'tg_session_hash');
+			$table->addUniqueIndex(['session_hash'], 'tgn_session_hash');
```
jancborchardt, line 106: This fixes my upgrade issue, so :+1: to that part for sure. :)

nickvergessen:
Rebased to add "listable" column on the select

lib/Model/SelectHelper.php
```diff
@@ -0,0 +1,85 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Talk\Model;
+
+use OCP\DB\QueryBuilder\IQueryBuilder;
+
+class SelectHelper {
+	public function selectRoomsTable(IQueryBuilder $query, string $alias = 'r'): void {
```
PVince81, line 29: for readability on the calling side, I think I wouldn't make the alias optional.
I came here to check as I was confused where the alias comes from

lib/Model/SelectHelper.php
```diff
@@ -0,0 +1,85 @@
...
+			->addSelect($alias . 'password')
+			->addSelect($alias . 'active_guests')
+			->addSelect($alias . 'active_since')
+			->addSelect($alias . 'last_activity')
+			->addSelect($alias . 'last_message')
+			->addSelect($alias . 'lobby_timer')
+			->addSelect($alias . 'object_type')
+			->addSelect($alias . 'object_id')
+			->addSelect($alias . 'listable')
+			->selectAlias($alias . 'id', 'r_id');
```
PVince81, line 51: r_id should take the "r" from the alias, for consistency

lib/Manager.php
```diff
@@ -241,9 +242,9 @@ public function loadLastCommentInfo(int $id): ?IComment {
 
 	public function resetAssignedSignalingServers(ICache $cache): void {
 		$query = $this->db->getQueryBuilder();
-		$query->select('*')
-			->from('talk_rooms')
-			->where($query->expr()->isNotNull('assigned_hpb'));
+		$query->select('r.*')
```
PVince81, line 245: did you miss replacing this one with the helper ?

lib/Model/CommandMapper.php
```diff
@@ -85,8 +85,13 @@ public function find(string $app, string $cmd): Command {
 		$query = $this->db->getQueryBuilder();
 		$query->select('*')
 			->from($this->getTableName())
-			->where($query->expr()->eq('app', $query->createNamedParameter($app)))
-			->andWhere($query->expr()->eq('command', $query->createNamedParameter($cmd)));
+			->where($query->expr()->eq('command', $query->createNamedParameter($cmd)));
```
PVince81, line 88: should the select * above be replaced as well for Oracle ?

lib/Model/SelectHelper.php
nickvergessen: The alias is optional?

lib/Model/SelectHelper.php
nickvergessen: No, it needs to be `r_id` to match https://github.com/nextcloud/spreed/blob/70679077ca966f48e5dd710b75613b090742e915/lib/Manager.php#L178

lib/Manager.php
nickvergessen: yes, although in this case it's not a problem as we only request against 1 DB. but could do for consistency

lib/Model/CommandMapper.php
nickvergessen: No, it's only one table being queried, so it's fine.

lib/Model/SelectHelper.php
nickvergessen: Ah, you would **n't**.

Well it's r for room, a for attendee and s for session. Your IDE (if you have one) will also tell you that. Not a huge fan of adding hundreds of `'(a|r|s)` in the code just for lazy non-IDE users

PVince81: it's not only for non-IDE but also when reading the code on Github for example, like when reviewing code changes or viewing a commit / diff on Github or CLI

but as you like, not a big issue to infer the alias

nickvergessen:
Okay actually the problem was we first installed composer deps and then set the php version, so it ran composer with 8.0 before switching to 7.4
Nevertheless this led to the relevant fix in the dependency so 8.0 will work in the future.

PVince81 (approved):
:+1:

Issue #4586, opened 2020-11-12
Cannot grab scroll cursor when on top


PVince81:
## Steps to reproduce
1. Have a chat with many messages
2. Scroll to the top with the mouse's scroll wheel
3. Now try grabbing the scrollbar's cursor to scroll back down

### Expected behaviour
Scrollbar cursor always reachable.

### Actual behaviour
Dragging the cursor does nothing when the cursor is on the row with the call buttons.

<img width="221" alt="image" src="https://user-images.githubusercontent.com/277525/98970240-01830400-2510-11eb-8a83-6aab23b24372.png">

The cursor can be grabbed when outside that zone.
We may want to move the buttons a tiny bit to the left.

## Talk app

**Talk app version:** 11.0 pre-alpha git master 9f535433509441da8fddb3f86eb8ed4bfcbd0de1

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/4591

PR #4591, opened 2020-11-13
Increase top bar space from the right

PVince81:
When in chat view, increase the space on the right of the top right to
make sure the user can still grab the scroll bar.

In call view that space is set back to zero.

Fixes https://github.com/nextcloud/spreed/issues/4586

### DIFF
--- a/src/components/TopBar/TopBar.vue
+++ b/src/components/TopBar/TopBar.vue
@@ -20,7 +20,7 @@
 -->
 
 <template>
-	<div class="top-bar">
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):
Works in FF, not sure if 12px is enough for all browsers?

PVince81:
I developed this on Chromium where it was enough, I believe Chromium's scrollbar is smaller.
On the other hand, I think we styled the scrollbar so it should look the same in all browsers anyway.

Issue #4592, opened 2020-11-13
Clicking search result from current chat triggers conversation not found error


PVince81:
## Steps to reproduce
1. Create a conversation
2. Type some text with the word "example" in it
3. Open the search box on the top bar
4. Type "example"
5. Click the result from the conversation

### Expected behaviour
Page reload to the matching conversation. (ideally no page reload, but that's something for another time)

### Actual behaviour
Sometimes, it seems that leaving the conversation also triggers a 404. Maybe some code has already left the conversation but the code that polls is still running:
<img width="353" alt="image" src="https://user-images.githubusercontent.com/277525/99067281-b964ef80-25aa-11eb-93e2-db91a6aab654.png">

~~This happens fairly often and is annoying.~~

It only happens whenever the clicked search result is targetted at the currently open conversation.
If it targets another conversation the switch happens without issue.

## Talk app

**Talk app version:** 11.0 git master fd02d1a41e89043582523d4945e68c9ab013bf46

**Signaling server:** None

Potential indirect fix: https://github.com/nextcloud/spreed/pull/4338

What I currently don't understand is why the 404 is happening in the first place.
Maybe it's just a result of cancelling the requests.
Furthermore, the "unload" handler is synchronous, so nothing else should be able to happen or fail.

Sometimes it also seems to happen when I click on other links, even other conversations in the left sidebar.

PVince81:
The access log reflects what I see in the network console:
```
127.0.0.1 - - [13/Nov/2020:12:24:38 +0100] "POST /ocs/v2.php/apps/spreed/api/v3/room/amgxdf2n/participants/active HTTP/1.1" 200 707
127.0.0.1 - - [13/Nov/2020:12:24:38 +0100] "DELETE /ocs/v2.php/apps/spreed/api/v3/room/amgxdf2n/participants/active HTTP/1.1" 200 81
127.0.0.1 - - [13/Nov/2020:12:24:38 +0100] "GET /ocs/v2.php/apps/spreed/api/v1/signaling/settings?token=amgxdf2n HTTP/1.1" 200 255
127.0.0.1 - - [13/Nov/2020:12:24:38 +0100] "GET /ocs/v2.php/apps/spreed/api/v3/room/amgxdf2n/participants?includeStatus=true HTTP/1.1" 
200 354
[...]

PR #4593, opened 2020-11-13
Don't leave conversation on URL hash change

PVince81:
Whenever only the URL hash changes despite the token staying the same, a
route change event is still triggered.

This fix avoids leaving and re-joining the same conversation.

Fixes an issue where clicking a search result would trigger both the
event that leaves the conversation within a route change and a page
unload event, which also leaves the same conversation.

Fixes https://github.com/nextcloud/spreed/issues/4592

### Test cases

- searching within current conversation: https://github.com/nextcloud/spreed/issues/4592#issue-742379306
- joining non-current conversation X through search link, then clicking X again to clear the hash: https://github.com/nextcloud/spreed/issues/4592#issuecomment-726745488

### DIFF
--- a/src/components/LeftSidebar/ConversationsList/ConversationsList.vue
+++ b/src/components/LeftSidebar/ConversationsList/ConversationsList.vue
@@ -95,6 +95,13 @@ export default {
 			})
 		},
 		onRouteChange({ from, to }) {
+			if (from.name === 'conversation'
+				&& to.name === 'conversation'
+				&& from.token === to.token) {
+
+				// this is triggered when the hash in the URL changes
+				return
+			}
 			if (from.name === 'conversation') {
 				leaveConversation(from.params.token)
 			}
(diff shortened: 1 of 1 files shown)


PVince81:
Note: when testing using the steps from the ticket you'll noticed that the search box stays open and the page simply scrolls to the result. We need to fix unified search to automatically close itself when clicking a link. So far it seems to assume that every link would lead to a page reload.

src/components/LeftSidebar/ConversationsList/ConversationsList.vue
```diff
@@ -95,6 +95,13 @@ export default {
 			})
 		},
 		onRouteChange({ from, to }) {
+			if (from.name === 'conversation'
+				&& to.name === 'conversation'
+				&& from.token === to.token) {
+
```
nickvergessen, line 101:
```suggestion
```

nickvergessen (approved):

PR #4600, opened 2020-11-13
Fix preview spacing

PVince81:
Fixes https://github.com/nextcloud/spreed/issues/4494

- Removes spacing of label under the preview 
- Reduce "small" preview size from 64 to 32 to prevent scrollbars in the reply box
- Attempt to move the preview size values to SCSS variables to be imported also in JS

The latter with SCSS variables doesn't work, the imported structure is always an empty hash.
This should work according to many sources:
- https://til.hashrocket.com/posts/sxbrscjuqu-share-scss-variables-with-javascript
- https://css-tricks.com/getting-javascript-to-talk-to-css-and-sass/
- https://stackoverflow.com/questions/56523785/exporting-scss-variables-to-js-auto-looping-variables
- https://github.com/webpack-contrib/css-loader (search for ":export")

I guess I'll revert the last commit and stick to hard-coded values for now :-/

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
@@ -54,7 +54,9 @@
 			<Close class="remove-file__icon" decorative @click="$emit('remove-file', id)" />
 		</button>
 		<ProgressBar v-if="isTemporaryUpload && !isUploadEditor" :value="uploadProgress" />
-		<strong v-if="shouldShowFileName">{{ name }}</strong>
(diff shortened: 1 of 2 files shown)


nickvergessen (approved):
can't see huge improvements but also doesn't look more crappy.

nickvergessen:
- [ ] Mimetype icon in the preview of the new-message box is still big and scrolls

PVince81:
I've pushed yet another fix, I forgot to fix the sidebar case, see below:

Before:
image

After:
image

And I've fixed the mime icon size:
image

nickvergessen (approved):

PR #4608, opened 2020-11-15
Add Verbose output support to ActiveCalls API

atemp:
This PR adds verbose output to `talk:active-calls` if `-v` is used. It is useful to find out which call is currently running, if this blocks eg. an update-script.


atemp:
does not work with nc21dev/master will reopen once i have test env installed

PR #4620, opened 2020-11-17
Adjust call icon positioning in participant list

PVince81:
Fixed call icon vertical alignment.
Adjusted ellipsis width calculation to take the icon into account when
present.

@nickvergessen as discussed

### Before:
image

### After:
image

with status line:
image

PR #4621, opened 2020-11-17
Add tooltips to call icons in participant list

PVince81:
Best tested with https://github.com/nextcloud/spreed/pull/4620 :
image

I wasn't sure about the wording and didn't want to delay the above PR, so made this separate PR.
Let me know what you think.

### DIFF
--- a/src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
+++ b/src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
@@ -64,8 +64,8 @@
 		</div>
 		<div v-if="callIcon"
 			v-tooltip.auto="callIconTooltip"
-			:aria-label="callIconTooltip"
 			class="participant-row__callstate-icon">
+			<span class="hidden-visually">{{ callIconTooltip }}</span>
 			<Microphone
 				v-if="callIcon === 'audio'"
 				:size="24"
--- a/src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
+++ b/src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
@@ -63,18 +63,23 @@
 			</div>
 		</div>
 		<div v-if="callIcon"
+			v-tooltip.auto="callIconTooltip"
+			:aria-label="callIconTooltip"
 			class="participant-row__callstate-icon">
 			<Microphone
 				v-if="callIcon === 'audio'"
 				:size="24"
+				title=""
 				decorative />
 			<Phone
 				v-if="callIcon === 'phone'"
 				:size="24"
+				title=""
 				decorative />
 			<Video
 				v-if="callIcon === 'video'"
 				:size="24"
+				title=""
 				decorative />
 		</div>
 		<Actions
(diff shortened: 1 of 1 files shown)


src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
```diff
@@ -57,18 +57,22 @@
...
 				decorative />
 			<Phone
 				v-if="callIcon === 'phone'"
 				:size="24"
+				title=""
 				decorative />
 			<Video
 				v-if="callIcon === 'video'"
 				:size="24"
+				title=""
```
nickvergessen, line 75: 🧐

src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
```diff
@@ -246,6 +250,16 @@ export default {
 			}
 			return 'audio'
 		},
+		callIconTooltip() {
+			if (this.callIcon === 'audio') {
+				return t('spreed', 'Joined with audio')
+			} else if (this.callIcon === 'video') {
+				return t('spreed', 'Joined with video')
+			} else if (this.callIcon === 'phone') {
+				return t('spreed', 'Joined with phone')
```
nickvergessen, line 259:
```suggestion
				return t('spreed', 'Joined via phone')
```

src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
PVince81: yep... otherwise there's a "title" tip that says "Video icon"... seems to be built-in

src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
PVince81: adjusted (only this one)

src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
nickvergessen: Should not happen with the `decorative` attribute

PVince81: then probably a bug, when I remove the title attribute it looks like this:
image

PVince81: ok, I managed to make it appear elsewhere as well:
image

According to the docs, the purpose of "decorative" is not to remove the title: https://github.com/robcresswell/vue-material-design-icons#props

unless it did before and it's a regression from a recent update

PVince81: nope, this also happens on Talk 10.0, so it was always there but no one hovered there long enough to see it.

I'll keep title empty here and make another PR to fix the other buttons.

PVince81: pushed https://github.com/nextcloud/spreed/pull/4633 (I also adjusted the titles there, so whichever is merged first wins)

PVince81:
I tried adding aria-label on the div or the icons, but the screen reader only reads them if I explicitly click on "Alice".
But if I come with keyboard navigation it only says "Alice (pause) Participant settings push button".
When clicking it says "Alice joined with audio (pause) Participant settings push button".

So for the meantime I'll just set the aria label on the div. I anyway think we should revamp the participant list somehow to be able to read out more useful info like "Participant alice, moderator, is online, joined with audio"

PVince81:
@danxuliu please review

PVince81:
I found a way to make the screen reader say the text using hidden-visually.
It seems an aria-label on a plain div doesn't work anyway.

ma12-co (approved):

PR #4622, opened 2020-11-17
cloak password fields

wiswedel:
Disclosing passwords can be dangerous when having multiple admins with different trust levels on an instance.
Also, it's quite common to cloak password fields.

Signed-off-by: Sascha Wiswedel <sascha.wiswedel@nextcloud.com>


nickvergessen:
Changing the type on the input only changes the displaying on your browser. You can open the browser development console to change back the type and see the value again.
Also there is an API for admins that allows to grab all app configs, so this is not helping at all.
For nextcloud all admins are equal.

PR #4637, opened 2020-11-18
[stable20] Add a phone icon for SIP phone users

PVince81:
Fixes warning about missing "callIcon" property.


danxuliu (approved):
Tested and works :+1:

Issue #4631, opened 2020-11-18
No Participant refresh after 10.0.2


apg1980:
## Steps to reproduce
1. Open a Conversation with more than 1 Participants

### Expected behaviour
List of all Partiticants with Status Indicator should displayed

### Actual behaviour
All Participants are not shown its greyed out.
![grafik](https://user-images.githubusercontent.com/9514965/99492851-95b8f500-296e-11eb-95b2-30e912630544.png)
Refresh Browser with F5 the Participants are there (this behaviour is in all Rooms):
![grafik](https://user-images.githubusercontent.com/9514965/99492985-bed98580-296e-11eb-95ac-5bf1de54e425.png)
Also the Status bouble is quite big.

## Browser
Latest Firefox, Edge

**Microphone available:** yes/no
yes
**Camera available:** yes/no
yes
**Operating system:** Windows/Ubuntu/...
Ubuntu 20.04
**Browser name:** Firefox/Chrome/...

**Browser version:** 50.1/55/...

### Browser log

<details>
```
Insert your browser log here, this could for example include:
a) The javascript console log
`Current participant found 2 CallButton.vue:101
The request has been canceled 
Object { message: "canceled" }
MessagesList.vue:489
The request has been canceled 
Object { message: "canceled" }
MessagesList.vue:489

​`
## Talk app

**Talk app version:10.0.2 (see apps admin page: `/index.php/settings/apps`)

**Custom TURN server configured:** yes

**Custom STUN server configured:** yes


## Server configuration
<!--
You can use the Issue Template application to prefill most of the required information: https://apps.nextcloud.com/apps/issuetemplate
-->


**Operating system**: Ubuntu 20.04

**Web server:** Nginx

**Database:** MySQL/Maria/SQLite/PostgreSQL

**PHP version: 7.04

**Nextcloud Version:** 20.01


<details>
```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your server installation folder
```
</details>

**Nextcloud configuration:**

<details>
```
If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your Nextcloud installation folder

or

Insert your config.php content here
Make sure to remove all sensitive content such as passwords. (e.g. database password, passwordsalt, secret, smtp password, …)
```
</details>

### Server log (data/nextcloud.log)
<details>
```
Insert your server log here
```
</details>

PVince81:
[...]

PR #4639, opened 2020-11-18
[stable20] Revert "Don't leave conversation on URL hash change"

PVince81:
Fixes https://github.com/nextcloud/spreed/issues/4631 

Bug also observed on master. I'll see if fixable without a revert


PVince81:
no need for revert, I found the cause... the token attribute is under "params".

fix incoming...

PR #4640, opened 2020-11-18
Grab the token from the correct attribute

PVince81:
When the URL changes, the check whether we switched conversations need
to read the token for the correct attribute, which is "params".

Fixes https://github.com/nextcloud/spreed/issues/4631


danxuliu (approved):
Tested and works :+1:

PR #4617, opened 2020-11-17
Fix "Copy link" button not clickable in empty calls

Chartman123:
This fixes https://github.com/nextcloud/spreed/issues/4311 by adding "z-index: 1;" to the "empty-call-view button" css.

PR #4641, opened 2020-11-18
Fix "Copy link" button not clickable (resend)

### LINKED ISSUE #4311: "Copy link" at middle of screen isn't clickable/doesn't work
This has spanned multiple major versions of nextCloud now, so this is a longer standing bug. When I start a call, there's a blue button "Copy link" in the middle of the window. Clicking on it never copies the link, it does nothing. I have to go to the 3-dot menu to get the link copied.

I'm on NC20, but this has happened in NC18 and earlier versions. Latest version of Talk and previous versions too.

PVince81:
Resend of https://github.com/nextcloud/spreed/pull/4617

Fixes #4311

### DIFF
--- a/src/components/CallView/shared/EmptyCallView.vue
+++ b/src/components/CallView/shared/EmptyCallView.vue
@@ -167,6 +167,7 @@ export default {
 	}
 	button {
 		margin: 4px auto;
+		z-index: 1;
 	}
 
 	h2, p {


Chartman123 (approved):
Works as stated in original PR

ma12-co (approved):
:elephant:

nickvergessen:
Backport?

Chartman123:
As it's a rather small change, I'd propose to baxkport it, too 👍🏻

Issue #4636, opened 2020-11-18
Public sidebar regressions


PVince81:
From @danxuliu from https://github.com/nextcloud/spreed/pull/4530#issuecomment-729567681

### Scenario 1

- Open the Files app
- Share a file by link
- Set a password
- Enable video verification
- Copy the link
- In a private window, open the link
- Request password

The layout of the video verification sidebar is broken.

### Scenario 2

- Open the Files app
- Share a file by link
- Copy the link
- In a private window, open the link
- Join the conversation

The layout of the sidebar is broken.

PVince81:
so this is how video verification looks on Talk 10.0:
<img width="476" alt="image" src="https://user-images.githubusercontent.com/277525/99534657-0b8d8280-29a8-11eb-9afc-cbc52808929a.png">

PVince81:
Fix is here: https://github.com/nextcloud/spreed/pull/4643

PR #4643, opened 2020-11-18
Fix public page calls

PVince81:
Fixes chat in sidebar for video verification and when chatting about a
file from a public link.

I tried to set "position: relative" for chatView in general but it would then break scrolling during a logged in call, so I went for a more localized solution (for now).

Fixes https://github.com/nextcloud/spreed/issues/4636

Note: the video won't appear in video verification because of that other bug related to "disable grid view in sidebar": https://github.com/nextcloud/spreed/pull/4573


danxuliu (approved):
Tested and works :+1:

PR #4644, opened 2020-11-18
Add changelog entries for 10.0.3

PVince81:


PVince81:
converted to draft as I'm not sure if I can release this

PR #4648, opened 2020-11-19
Add missing aria label on new message form

PVince81:
Tested with screen reader on orca + firefox.

It will now read out the placeholder text when focusing on the new message form.


PVince81:
moved to https://github.com/nextcloud/spreed/pull/4652

PR #4652, opened 2020-11-19
A few accessibility tweaks

PVince81:
See individual commit descriptions.

I originally tried to go further and add invisible text to make it say "Message from Alice: hello" but somehow the screen reader doesn't read it in once sentence.
Also I was hoping that ul/li would make navigation easier like making it read each row in the participant list or messages list, but it rather reads individual blocks inside a row. Maybe we'll need to introduce links there at some point... would make it possible to link to messages. In the conversation list we have links and it works well there.

Needs further separate research.

### DIFF
--- a/src/components/ChatView.vue
+++ b/src/components/ChatView.vue
@@ -43,8 +43,12 @@
 			</div>
 		</transition>
 		<MessagesList
+			role="region"
+			:aria-label="t('spreed', 'Conversation messages')"
 			:token="token" />
-		<NewMessageForm />
+		<NewMessageForm
+			role="region"
+			:aria-label="t('spreed', 'Post message')" />
 	</div>
 </template>
 
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -25,7 +25,7 @@ the main body of the message as well as a quote.
 </docs>
 
 <template>
-	<div
+	<li
 		:id="`message_${id}`"
 		ref="message"
 		class="message"
@@ -75,7 +75,7 @@ the main body of the message as well as a quote.
 				</Actions>
 			</div>
 		</div>
-	</div>
+	</li>
 </template>
 
 <script>
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -33,7 +33,9 @@ the main body of the message as well as a quote.
 		@mouseover="showActions=true"
 		@mouseleave="showActions=false">
 		<div v-if="isFirstMessage && showAuthor" class="message__author">
-			<h6>{{ actorDisplayName }}</h6>
+			<h6 role="presentation" aria-hidden="true">
+				{{ actorDisplayName }}
+			</h6>
 		</div>
 		<div
 			ref="messageMain"
@@ -56,7 +58,8 @@ the main body of the message as well as a quote.
 			<div class="message__main__right">
 				<div v-if="isTemporary && !isTemporaryUpload" class="icon-loading-small" />
 				<h6 v-if="hasDate"
-					v-tooltip.auto="messageDate">
+					v-tooltip.auto="messageDate"
+					role="presentation">
 					{{ messageTime }}
 				</h6>
 				<Actions
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -32,10 +32,11 @@ the main body of the message as well as a quote.
 		:class="{'hover': showActions && !isSystemMessage, 'system' : isSystemMessage}"
 		@mouseover="showActions=true"
 		@mouseleave="showActions=false">
-		<div v-if="isFirstMessage && showAuthor" class="message__author">
-			<h3 role="heading">
-				{{ actorDisplayName }}
-			</h3>
+		<div v-if="isFirstMessage && showAuthor"
+			class="message__author"
+			role="heading"
+			aria-level="4">
+			{{ actorDisplayName }}
 		</div>
 		<div
 			ref="messageMain"
@@ -57,10 +58,7 @@ the main body of the message as well as a quote.
 			</div>
 			<div class="message__main__right">
 				<div v-if="isTemporary && !isTemporaryUpload" class="icon-loading-small" />
-				<span v-if="hasDate"
-					  v-tooltip.auto="messageDate"/>
-					{{ messageTime }}
-				</span>
+				<span v-if="hasDate" v-tooltip.auto="messageDate">{{ messageTime }}</span>
 				<Actions
 					v-show="showActions && hasActions"
 					class="message__main__right__actions">
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -33,9 +33,9 @@ the main body of the message as well as a quote.
 		@mouseover="showActions=true"
 		@mouseleave="showActions=false">
 		<div v-if="isFirstMessage && showAuthor" class="message__author">
-			<h6 role="heading" aria-level="2">
+			<h3 role="heading">
 				{{ actorDisplayName }}
-			</h6>
+			</h3>
 		</div>
 		<div
 			ref="messageMain"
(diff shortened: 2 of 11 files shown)


PVince81:
regarding the "conversation list navigation" region, it will only fully work once https://github.com/nextcloud/nextcloud-vue/pull/1585 is merged and updated here.

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -25,15 +25,17 @@ the main body of the message as well as a quote.
...
+	<li
 		:id="`message_${id}`"
 		ref="message"
 		class="message"
 		:class="{'hover': showActions && !isSystemMessage, 'system' : isSystemMessage}"
 		@mouseover="showActions=true"
 		@mouseleave="showActions=false">
 		<div v-if="isFirstMessage && showAuthor" class="message__author">
-			<h6>{{ actorDisplayName }}</h6>
+			<h6 role="presentation" aria-hidden="true">
```
PVince81, line 36: this shouldn't be hidden, leftover from a previous attempt

PVince81: resolved

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -25,15 +25,17 @@ the main body of the message as well as a quote.
...
+	<li
 		:id="`message_${id}`"
 		ref="message"
 		class="message"
 		:class="{'hover': showActions && !isSystemMessage, 'system' : isSystemMessage}"
 		@mouseover="showActions=true"
 		@mouseleave="showActions=false">
 		<div v-if="isFirstMessage && showAuthor" class="message__author">
-			<h6>{{ actorDisplayName }}</h6>
+			<h6 role="heading" aria-level="2">
```
jancborchardt, line 36: Btw I’d bet that we are skipping some heading levels here like h4 and h5, which is semantically incorrect. We should use the level which makes sense for the markup, and then style it to fit how we want it to look like.

src/components/MessagesList/MessagesGroup/MessagesGroup.vue
```diff
@@ -31,7 +31,7 @@
 					:author-id="actorId"
 					:display-name="actorDisplayName" />
 			</div>
-			<div class="messages">
+			<ul class="messages">
```
jancborchardt, line 34: The list is `ol` though, as it’s ordered, no?

src/components/Quote.vue
```diff
@@ -30,7 +30,7 @@ components.
 		@click.prevent="handleQuoteClick">
 		<div class="quote__main">
 			<div class="quote__main__author">
-				<h6>{{ getDisplayName }}</h6>
+				<h6 role="presentation">{{ getDisplayName }}</h6>
```
jancborchardt, line 33: Same as the heading level comment above:
> Btw I’d bet that we are skipping some heading levels here like h4 and h5, which is semantically incorrect. We should use the level which makes sense for the markup, and then style it to fit how we want it to look like.

src/components/MessagesList/MessagesGroup/Message/Message.vue
PVince81: yeah, seems h6 was used in many places mostly for formatting purposes.

I could make it a h2 if it looks identical

src/components/MessagesList/MessagesGroup/MessagesGroup.vue
PVince81: indeed, had the same thought.
technically doesn't make a difference for assistive technologies from what I read

src/components/Quote.vue
PVince81: indeed, missed that one. maybe for the quote we could make it h3 as it's one level deeper

src/components/MessagesList/MessagesGroup/Message/Message.vue
jancborchardt: Should it be h2 though? Isn’t it:
1. h1 is the title "Nextcloud" or rather "Nextcloud Talk"
2. h2 would e.g. be the room name
3. h3 would be stuff like the message author

And then `.message__author h3` would need styling to make it look as before.

src/components/Quote.vue
jancborchardt: I’d assume it stays the same level though. A heading wouldn’t go one level down just cause it’s in a quote. So h3, as per the additional comment above?

src/components/ChatView.vue
```diff
@@ -43,8 +43,12 @@
 			</div>
 		</transition>
 		<MessagesList
+			role="region"
+			:aria-label="t('spreed', 'Conversation messages')"
 			:token="token" />
-		<NewMessageForm />
+		<NewMessageForm
+			role="region"
+			:aria-label="t('spreed', 'Post message')" />
```
ma12-co, line 51: Shouldn't this  go within the newmessageform component?

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -25,15 +25,17 @@ the main body of the message as well as a quote.
 </docs>
 
 <template>
-	<div
+	<li
```
ma12-co, line 28: Please make sure to address the server's css that targets unordered lists

src/components/MessagesList/MessagesGroup/MessagesGroup.vue
ma12-co: The only difference afaik is visual in default styles. Bullets in <ul> and umbers in <ol>

src/components/MessagesList/MessagesGroup/Message/Message.vue
PVince81: @jancborchardt I've used one heading level for the day section as it's useful for grouping and navigation

src/components/ChatView.vue
PVince81: not this one: see I've defined a "role=region" which means that screen readers will announce this section as a region.
with a screen reader you have the option to jump directly to a specific region of the screen from anywhere.

PVince81: seems I had a separate PR for the other missing aria-label, see https://github.com/nextcloud/spreed/pull/4648

src/components/MessagesList/MessagesGroup/Message/Message.vue
PVince81: here I didn't see any visual differences so we might be fine

another alternative would be to use "role=listitem" here, this way the styles would not change

src/components/Quote.vue
PVince81: with the screen reader you use the up/down arrow keys in browse mode and it will read the current lines.
if you'd navigate over this one it will speak out "$authorName heading level three". it could cause confusion as the user would believe it's a new section on the same level

the "blockquote" bit is read out when you hit down arrow after the heading.
maybe if we move this heading inside the blockquote it will be more understandable, as the blockquote would get announced first and the heading as being inside

src/components/MessagesList/MessagesGroup/Message/Message.vue
ma12-co: No no, that's fine :)

src/components/ChatView.vue
ma12-co: Just sayinh that maybe those 2 lines should go here? 
https://github.com/nextcloud/spreed/blob/1bce4b8bdf8ea820c513bef83afdebe124a705ff/src/components/NewMessageForm/NewMessageForm.vue#L24
Aren't they going to end up on the newmessageform root element anyway?

PVince81: yes and no

We need to keep the aria-label here so that the region itself has a name. The screen reader will announce it "Entering post message region" when tabbing into it, even if you tab onto the file upload button first so we should not set the region on the post message field. 

I agree that the field itself also needs an aria-label.

src/components/MessagesList/MessagesGroup/Message/Message.vue
PVince81: I've moved away from using "<h?>"  tags as I didn't want to fight against the styles trying to reproduce the existing style.

Instead, I used the alternative form of adding `role='heading'  aria-level='x'` to define the heading, which is valid and will be parsed as such by screen readers.

The hierarchy is now:
1. h1 app title
2. h2 room name
3. h3 date separator for message list grouped by date (useful for jumping to a different date)
4. h4 message author and message quote author

On the date item on the right I simply used a span as it's not really a header.

src/components/Quote.vue
PVince81: I've changed it to the same level as author now

src/components/ChatView.vue
PVince81: I've cherry-picked my changes from the other PR here to make it more complete.

Please let me know whether you understood the region thing. I can demo it to you if you like.

ma12-co (approved):
tested and I see no design regressions due to tag changes :)

src/components/MessagesList/MessagesGroup/MessagesGroup.vue
nickvergessen: Also when scrolling up new items would be added and therefor all numbers would change. sounds confusing?

PR #4653, opened 2020-11-19
Improve Matterbridge settings design

eneiluj:
Hey there,

Here are a few design improvements in Matterbridge room settings.

* Move "show matterbridge log" button next to the bridge switch and make it more discreet
* Reorder add/save/toggle elements. Add is now first to emphasize it.
* Display icons for each external service (in the multiselect and in the bridge part config component)
* Bridge switch label is now reflecting current desired state (enabled/disabled)
* The process state in the parenthesis now suggests to "check matterbridge log" when the bridge is enabled but the process is not running.
* Fold the whole bridge settings area on title click. Title icon changes to east/south caret on hover.
* Make the multiselect placeholder more explicit
* Add icon-external to external link to documentation

Any other idea?

How would you improve the multiselect appearance? The only way I found to display icons on the right was to switch on "user-select" prop which adds avatars to all items. I think it's kind of ok :grin:.


ma12-co (commented):
@eneiluj since this will be moved in the settings dialog component, I'd say that there's no need to make it collapsible as it's going to have its own setting section.

ma12-co:
Other than that, I think it'd be nice to have separate "view" and "edit" mode in every configured bridge. E.g. we could have a three dot menu where the current delete button is, and have all the entries in there:

- Info;
- edit;
- delete;

eneiluj:
@ma12-co Thanks for the review!

> no need to make it collapsible as it's going to have its own setting section

Oh indeed. Done.

> it'd be nice to have separate "view" and "edit" mode in every configured bridge

Yes it's much nicer. It's like that now. New parts are in edition mode. There is an edition switch in actions menu. All parts are put back in readonly mode when saving.

eneiluj:
Hey, rebased on master, squashed. Ready for review.

img/bridge-services/matrix.svg
```diff
@@ -0,0 +1,25 @@
...
+
+
+
+
+
+
+
+
+
+
```
nickvergessen, line 24: Something is wrong.

please use the `build/image-optimization.sh` of server

src/components/RightSidebar/Matterbridge/MatterbridgeSettings.vue
```diff
@@ -565,38 +602,168 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-.loading {
-	margin-top: 30px;
+::v-deep .icon-slack {
+	mask: url('./../../../../img/bridge-services/slack.svg') no-repeat;
```
nickvergessen, line 606: what magic is `mask` can't we use it just as a normal `img` or as a background image, like with all other icons?

src/components/RightSidebar/Matterbridge/MatterbridgeSettings.vue
```diff
@@ -20,59 +20,75 @@
...
+						:user-select="true"
+						:internal-search="true"
+						@input="clickAddPart">
+						<template #option="{option}">
+							<span :class="option.icon" />
+							{{ option.displayName }}
+						</template>
+					</Multiselect>
+				</div>
 				<ActionButton
```
nickvergessen, line 58: can you make this a button, disabled when unchanged, only getting enabled when content was changed, so it's much clearer you need to save things?

img/bridge-services/matrix.svg
eneiluj: That's what I used. I manually removed the blanks.

src/components/RightSidebar/Matterbridge/MatterbridgeSettings.vue
eneiluj: It's an easy way to support the dark theme. Other icons have their background-image changed with global CSS variables. While we could also declare CSS variables and make them change depending on dark theme, I find it more convenient to use masks and `background-color: var(--color-main-text)`. As you wish.

Other advantage of using masks, it gets rid of the image color.

src/components/RightSidebar/Matterbridge/MatterbridgeSettings.vue
eneiluj: We can also show it only after changes (add, edit, delete a part). I pushed it like that.

ActionButton has a disabled state but it's visually not very clear it's disabled. Making this a real button would not look as nice imho.

Do you still prefer a button with disabled text?

ma12-co (commented):
I find it confusing that the multiselect uses the avatar component for the services. Could we use the services icons instead?

eneiluj:
@ma12-co Yeah the gif is out of date, sorry. Multiselect options now have the service icon on the left side, no more avatar.

src/components/RightSidebar/Matterbridge/MatterbridgeSettings.vue
nickvergessen: Well ActionButton should only be used in an Actions-menu. So just use a plain `<button>` html element here.

That button can appear once you changed something. (Although I'm currently not sure. Since this will all be moved to the new settings dialog anyway, right?)

src/components/RightSidebar/Matterbridge/MatterbridgeSettings.vue
nickvergessen: https://caniuse.com/?search=mask

It's not supported in all necessary browsers. So let's just use the current guidelines. make the svgs black and use the normal way to make it automatically adjust to dark theme.

img/bridge-services/matrix.svg
nickvergessen: Weird, okay then

jancborchardt:
Nice, good progress! :) Some details

- [ ] When adding a new service, focus could go to the first field of that service so you can directly start editing
- [ ] The "Save" action should be styled as a button? And be below the "Enable/Disable" checkmark section
- [ ] The checkmark changes text which is confusing. Instead of "Disabled (not running)" and "Enabled (running)" it should say: "Enable bridge" always, and then it’s either checked or not
- [ ] The "Matterbridge" in "Show matterbridge log" should be written with capital M as it’s a project name :) 
- [ ] The "End edition" label should be called "Save"?
- [ ] The "Edit" and "Save" actions should be outside of the menu since otherwise it’s not immediately obvious how to edit entries. Like clicking into them doesn’t work even though the empty fields show.
- [ ] To make the view more compact, when not editing, I’d also say that when not editing, we can collapse an entry to only show the bridge platform name and server URL – this will also make it clear it’s a different view.

src/components/RightSidebar/Matterbridge/MatterbridgeSettings.vue
eneiluj: Right. Icons are now `` using black svg images. `theme--dark` body class is now used to apply a CSS filter to those images. [CSS filter support](https://caniuse.com/css-filters).

src/components/RightSidebar/Matterbridge/MatterbridgeSettings.vue
eneiluj: Right, no more ActionButton (and ActionCheckbox) there.

> Since this will all be moved to the new settings dialog anyway, right?

Why would it change the save button apparition conditions?

eneiluj:
@jancborchardt Thanks a lot for the review.

A few more things:
* the "enable" switch is hidden when there is no bridged service configured
* when there is no more bridged service (let's say we just deleted the only one there) and we save, the bridge is automatically disabled

> The "End edition" label should be called "Save"?

Well as there is a button to save the whole config I would say we need to:
* either keep different labels to validate edition and saving config
* or try to get rid of the save button...but it would feel weird to save on each edition action. Any idea?

bridge

src/components/RightSidebar/Matterbridge/BridgePart.vue
```diff
@@ -20,26 +20,47 @@
 -->
 
 <template>
-	<div class="part">
+	<div :class="{ part: true, readonly: !editing }">
 		<h3>
+			<img class="icon-service"
+				:src="type.iconUrl" />
```
nickvergessen, line 26:
```suggestion
				:src="type.iconUrl">
```

src/components/RightSidebar/Matterbridge/BridgePart.vue
```diff
@@ -20,26 +20,47 @@
...
-				:href="type.infoTarget" />
-			<button v-if="deletable"
-				class="icon icon-delete"
-				@click="$emit('deletePart')" />
+			<Actions
+				:force-menu="false">
+				<ActionButton
+					:icon="editing ? 'icon-checkmark' : 'icon-rename'"
+					@click="onEditClick">
+					{{ editing ? t('spreed', 'Ok'): t('spreed', 'Edit') }}
```
nickvergessen, line 35: "Ok" is not an action. what do you mean? Save?

src/components/RightSidebar/Matterbridge/BridgePart.vue
```diff
@@ -20,26 +20,47 @@
...
+				</ActionButton>
+			</Actions>
+			<Actions
+				:force-menu="true"
+				placement="bottom">
+				<ActionLink
+					icon="icon-info"
+					title="More information"
```
nickvergessen, line 43: missing translation

src/components/RightSidebar/Matterbridge/BridgePart.vue
```diff
@@ -97,12 +129,27 @@ export default {
...
+		},
 	},
 
-	beforeMount() {
+	mounted() {
+		this.focusMainField()
 	},
 
-	beforeDestroy() {
+	watch: {
```
nickvergessen, line 149: The "watch" property should be above the "mounted" property on line 145

src/components/RightSidebar/Matterbridge/BridgePart.vue
eneiluj: Well this button does not really save, it just validates the edition. Would "Validate" be ok?

nickvergessen: Check?
Or just do it automatically after a timeout (so you can use `debounce()`)

eneiluj: What about allowing edition only if the bridge is disabled and save each time a part is validated? So we can call this edition validation button "save" and get rid of the big general "save" button.

nickvergessen: Sounds like it would help to get rid of confusion and help to avoid misbehaviour

eneiluj:
3 more changes:
* the settings can't be edited when bridge is enabled (multiselect is hidden and parts actions are removed)
* bridge settings are saved each time user validates the edition of a part
* fix system messages (sometimes not triggered after edition) and add one after enabling and disabling the bridge

jancborchardt:
> or try to get rid of the save button...but it would feel weird to save on each edition action. Any idea?

I’d say it’s fine to get rid of the big save button, as I was not quite sure what it did. It would be fine if each single bridge service would have an edit/save state, then it’s pretty clear.

eneiluj:
@jancborchardt No more big save button. Bridge config is saved each time a part is edited or deleted.

bridge

eneiluj:
Rebased on master.

src/components/RightSidebar/Matterbridge/MatterbridgeSettings.vue
```diff
@@ -20,59 +20,79 @@
...
+						:options="formatedTypes"
+						:user-select="true"
+						:internal-search="true"
+						@input="clickAddPart">
+						<template #option="{option}">
+							<img class="icon-multiselect-service"
+								:src="option.iconUrl">
+							{{ option.displayName }}
+						</template>
+					</Multiselect>
```
nickvergessen, line 58: This breaks line with Dyslexia font from the accessibility settings:
Bildschirmfoto von 2020-12-15 13-38-49

src/components/RightSidebar/Matterbridge/MatterbridgeSettings.vue
```diff
@@ -441,9 +485,10 @@ export default {
 							icon: 'icon-group',
 						},
 					},
+					mainField: 'channel',
 				},
 			},
-			newPartPlaceholder: t('spreed', 'Add new bridged channel'),
+			newPartPlaceholder: t('spreed', 'Add new bridged channel to current room'),
```
nickvergessen, line 491:
```suggestion
			newPartPlaceholder: t('spreed', 'Add new bridged channel to current conversation'),
```

src/components/RightSidebar/Matterbridge/MatterbridgeSettings.vue
```diff
@@ -565,38 +622,118 @@ export default {
...
+		.save-changes {
+			width: 100%;
+			text-align: left;
+
+			.icon-checkmark {
+				margin: 0 10px 0 2px;
+			}
+		}
+		.multiselect {
+			width: calc(100% - 44px);
```
nickvergessen, line 693:
```suggestion
			width: calc(100% - 46px);
```
seems to fix the dyslexia

nickvergessen (changes requested):

nickvergessen (approved):

PR #4666, opened 2020-11-23
WIP keyboard nav attempt on messages list

PVince81:
Adds keyboard navigation on the messages list:

### Todos
- [x] click a message to focus/select, then arrows to move
- [ ] tab should move to the reply button (it does but invisible)
- [ ] arrow up on NewMessageForm should jump to last message
- [ ] arrow down after last message should jump to NewMessageForm

### Issues
- [ ] tab should not move between messages
- [ ] reply button focus conflict between mouse and keyboard, might need to rework the hover concept
- [ ] warnings in log due to the quick and dirty approach for next-message and previous-message attributes

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -28,8 +28,13 @@ the main body of the message as well as a quote.
 	<div
 		:id="`message_${id}`"
 		ref="message"
+		tabindex="0"
 		class="message"
 		:class="{'hover': showActions && !isSystemMessage, 'system' : isSystemMessage}"
+		@keydown.up.prevent="jumpToPrevious"
+		@keydown.down.prevent="jumpToNext"
+		@focus="showActions=true"
+		@blur="showActions=false"
 		@mouseover="showActions=true"
 		@mouseleave="showActions=false">
 		<div v-if="isFirstMessage && showAuthor" class="message__author">
@@ -155,6 +160,20 @@ export default {
 		id: {
 			type: [String, Number],
 			required: true,
+		},
+		/**
+		 * Id of the next message
+		 */
+		nextMessage: {
+			type: [String, Number],
+			default: null,
+		},
+		/**
+		 * Id of the previous message
+		 */
+		previousMessage: {
+			type: [String, Number],
+			default: null,
 		},
 		/**
 		 * If true, it displays the message author on top of the message.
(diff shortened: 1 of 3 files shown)


PVince81:
Ideas:
- [ ] skip non-interactable messages (like system messages) with the focus mode (screen readers can use browser)
- [ ] focus on quotes to be able to jump to original message
- [ ] focus on images/videos/etc to be able to summon the viewer

jancborchardt:
@PVince81 @ma12-co any update on this? Especially as we’re introducing more actions like deleting messages and creating a card, people who navigate with a keyboard are otherwise left out.

PVince81:
I made a little bit of progress and managed to make the hover persist, but it requires the Actions component to emit focus and blur events: https://github.com/nextcloud/nextcloud-vue/pull/1757

Another issue to fix in the lib:
- [ ] Actions menu, when opened with keyboard's enter key, doesn't focus on the first element, so arrow keys don't work (opening with mouse works and makes up/down work...)

PVince81:
- [ ] should we allow moving over system messages ? the current PR allows it with the keyboard but as there are no actions there it might not be useful

- [ ] BUG? since the buttons are now hidden-visually it's possible to tab over ALL of them which makes it tedious to exit the message list. However I can't remove `hidden-visually` as it breaks the hover/focus persistence on the row... Tricky stuff...

PVince81:
I think before going full keyboard nav, I'll try to have smaller bits done separately.
The next bit would be the part that keeps the focus active on the row while the menu is open, even when hovering away.

jancborchardt:
> I think before going full keyboard nav, I'll try to have smaller bits done separately.

Sure, step by step is absolutely fine. :) And when in doubt how to do specific things, probably Slack or MS Teams is the best place to check since if I remember correctly they do accessibility pretty nicely.

PVince81:
> > I think before going full keyboard nav, I'll try to have smaller bits done separately.
> 
> Sure, step by step is absolutely fine. :) And when in doubt how to do specific things, probably Slack or MS Teams is the best place to check since if I remember correctly they do accessibility pretty nicely.

Yeah, I'll need to recheck Slack. But from what I remember the keyboard navigation there was a bit messed up so wasn't fully convinced.

PVince81:
I've rebased this to bring in the focus/blur events for Actions. But it stopped working: for some reason when the mouse exits the actions components it also triggers a blur event ?!

nickvergessen:
I will close the PR for now (as there was no active development since 6 months). Anyone can pick it up whenever they want and reopen the PR.

PR #4576, opened 2020-11-11
Move conversation menu options to a dialog

PVince81:
For https://github.com/nextcloud/spreed/issues/4007

- [x] REQUIRES: release of nextcloud-vue with [scroll container fix](https://github.com/nextcloud/nextcloud-vue/pull/1573)

### Current state

See https://github.com/nextcloud/spreed/pull/4576#issuecomment-731219492

### Old state from first attempt

image

image

### Tasks

- [x] move sharing settings
- [x] add/copy "copy link" also to the settings tab
- [x] move lobby settings to settings tab
- [x] move SIP options to settings tab
- ~~remove lobby + locking panels while in call~~
- [ ] disable locking checkbox while in a call => will do separately in https://github.com/nextcloud/spreed/pull/4671
- [x] improve design
- [x] add more hints

### Tests

- [x] retest permissions and switches
   - [x] owner/moderator
   - [x] guest moderator
   - [x] non-moderator

PR #4671, opened 2020-11-25
Feature/noid/conversation settings error handling

PVince81:
Follow up of https://github.com/nextcloud/spreed/pull/4576 for:
- error handling to reset checkboxes after an error
- disable lock conversation field

- [x] REQUIRES: [base PR](https://github.com/nextcloud/spreed/pull/4576)
- [x] REQUIRES: [participant mixin](https://github.com/nextcloud/spreed/pull/4672)

### DIFF
--- a/src/components/ConversationSettings/LockingSettings.vue
+++ b/src/components/ConversationSettings/LockingSettings.vue
@@ -31,19 +31,27 @@
 				class="checkbox"
 				name="moderation_settings_lock_conversation_checkbox"
 				:checked="isReadOnly"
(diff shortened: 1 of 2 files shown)


PVince81:
- [ ] some troubles because the conversation object is not getting updated despite the call... the dialog is tied to an older instance

PVince81:
- [ ] fix service calls to actually throw out the exception instead of skipping, otherwise we can't catch error cases to display the error

nickvergessen:
Status?

PVince81:
not worth the effort => closed

PR #4672, opened 2020-11-25
Introduce participant mixin

PVince81:
To remove duplication of the "get current participant" logic all over
the place and also for future use.

Open issue:
- [x] seems one of the participant method was not implemented like the other, could be a bug ?! see https://github.com/nextcloud/spreed/pull/4672/files#r530502314 => I've fixed it, the below test will show if it worked

Tested based on the diff:
- [x] files sidebar call
- [x] public share sidebar
- [x] conversation messages
- [x] call button
- [x] app.vue: warn leaving when in call
- [x] mainview.vue: is in lobby check
- [x] newgroupconversation.vue: test creating a conversation while being in a call

### DIFF
--- a/src/components/LeftSidebar/NewGroupConversation/NewGroupConversation.vue
+++ b/src/components/LeftSidebar/NewGroupConversation/NewGroupConversation.vue
@@ -189,6 +189,7 @@ export default {
 		},
 
 		participant() {
+			// TODO: use participant mixin ??
 			if (this.$store.getters.getToken()) {
 				return {
 					inCall: PARTICIPANT.CALL_FLAG.DISCONNECTED,
--- a/src/components/LeftSidebar/NewGroupConversation/NewGroupConversation.vue
+++ b/src/components/LeftSidebar/NewGroupConversation/NewGroupConversation.vue
@@ -133,7 +133,8 @@ import {
 } from '../../../services/conversationsService'
 import { generateUrl } from '@nextcloud/router'
 import PasswordProtect from './PasswordProtect/PasswordProtect'
-import { PARTICIPANT } from '../../../constants'
+import isInCall from '../../../mixins/isInCall'
+import participant from '../../../mixins/participant'
 import Tooltip from '@nextcloud/vue/dist/Directives/Tooltip'
 
 export default {
@@ -153,6 +154,11 @@ export default {
 		PasswordProtect,
 		Plus,
 	},
+
+	mixins: [
+		isInCall,
+		participant,
+	],
 
 	data() {
 		return {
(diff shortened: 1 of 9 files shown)


src/components/LeftSidebar/NewGroupConversation/NewGroupConversation.vue
```diff
@@ -189,6 +189,7 @@ export default {
 		},
 
 		participant() {
+			// TODO: use participant mixin ??
```
PVince81, line 192: not sure why the below check is not `!this.$store.getters.getToken()` ??

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -97,6 +98,10 @@ export default {
 		tooltip: Tooltip,
 	},
 
+	mixins: [
+		participant,
+	],
+
 	components: {
```
nickvergessen, line 105: GitHub Actions / eslint node12.x

src/components/MessagesList/MessagesGroup/Message/Message.vue#L105

The "components" property should be above the "mixins" property on line 101

PVince81: adjusted

src/components/LeftSidebar/NewGroupConversation/NewGroupConversation.vue
nickvergessen: Yeah, the thing is this.token is not initialized or injected into this component.
It's only used after creating a conversation and then it contains the new token, independent from which conversation you are currently in.

So by switching to the mixin you need to make this.token a computed instead of a data attribute and the computed returns `this.$store.getters.getToken()` and all existing references of `this.token` need to change to `this.newToken` or something like this.

PVince81: When reading the code I had the impression that the logic is only here to detect whether there's a call in progress already, as you can create new conversations while being in a call, and when we do so we don't want to switch to the new conversation but stay.

And for that check we need "isInCall" which in turn needs "participant". But the current implementation (before this PR) would say "participant disconnected" even if a token was set, so "isInCall" would likely be false. I can test this to confirm.

PVince81: Okay, maybe we are talking about different things. My question is about the inverted negation. It should be `if (!this.$store.getters.getToken())  instead of `if (this.$store.getters.getToken())`

PVince81: regarding the computed prop, I now made the two mixins use the getter from the store, which reduces dependencies and potential side effects: https://github.com/nextcloud/spreed/pull/4672/commits/a38299001b6f89561c7002fe429489ae8998f48c

nickvergessen: ah right, so yeah that looks like a bug, maybe from copy paste and then replacing `this.token`

nickvergessen (approved):

PVince81:
I retested the checkboxes items in the OP and all went fine, merging

PR #4685, opened 2020-11-27
Unregister refreshPeerList on unmount

PVince81:
Fix copy-paste issue, replacing "$on" with "$off" when unmounting the
call view.

Might need a backport or cherry-pick in case this will be part of a bigger backport related to SIP.


nickvergessen (approved):

nickvergessen:
added to the backport in #4612

PR #4689, opened 2020-11-27
Follow up for SIP changes

### LINKED ISSUE #4677: Follow up for SIP changes
## Remaining steps:
- [x] Add back indexes for the queries
- [x] Add capability when sip can be enabled by the participant for mobile clients
- [x] Add docs
- [ ] Add integration tests (if possible)

nickvergessen:
* Make sure the key calculation is always unique
before it was s#0 for all email invites
* guests should no see the dial-in info panel
* Always generate PINs for logged in users when SIP is enabled (not only on join)
* Add capability and docs

Integration tests will be done in a new PR based on #4698 

Fix #4677


PVince81:
seems my local env (php7-7.4.12-1.1.x86_64 with NC server master c1161e27b47d1f5cd7b6f3ec0c229738bbc5c271) has trouble matching the numeric route and the signaling calls give a 404, for example "https://myhost/ocs/v2.php/apps/spreed/api/v2/signaling/8351248634" gives a 404.

In UrlMatcher::matchCollection:
- $regex = (string [86]) `#^/ocsapp/apps/spreed/api/(?P<apiVersion>v1)/signaling/(?P<token>[a-z0-9]{4,30})/?$#sD`
- $pathinfo = (string [47]) `/ocsapp/apps/spreed/api/v2/signaling/8351248634`

it fails to match on line 158.

Maybe it is expecting a string but we're giving a number ?
https://github.com/nextcloud/3rdparty/blob/master/symfony/routing/Matcher/UrlMatcher.php#L153

In the route definition I only see v1 specified: https://github.com/nextcloud/spreed/blob/669ab421158c3730dd17ecd0cc3130ef8d757c14/appinfo/routes.php#L99

Did you forget to push some commits ?

nickvergessen:
Fixed

nickvergessen:
Rebased to solve the TopBar conflict

lib/Controller/RoomController.php
```diff
@@ -1644,4 +1654,69 @@ protected function changeParticipantType(?int $attendeeId, ?string $userId, ?str
...
+	 * @return DataResponse
+	 */
+	public function setSIPEnabled(int $state): DataResponse {
+		$user = $this->userManager->get($this->userId);
+		if (!$user instanceof IUser) {
+			return new DataResponse([], Http::STATUS_UNAUTHORIZED);
+		}
+
+		if (!$this->talkConfig->canUserEnableSIP($user)) {
+			return new DataResponse([], Http::STATUS_UNAUTHORIZED);
```
PVince81, line 1709: should this be "forbidden" instead since unauthorized would return a 401 and make some clients think that the user is not authenticated (which is the case for the previous if statement).

lib/Controller/SignalingController.php
```diff
@@ -118,10 +118,13 @@ public function __construct(string $appName,
 	/**
 	 * @PublicPage
 	 *
+	 * @param string $apiVersion
 	 * @param string $token
 	 * @return DataResponse
 	 */
-	public function getSettings(string $token = ''): DataResponse {
+	public function getSettings(string $apiVersion, string $token = ''): DataResponse {
+		$apiV = (int) substr($apiVersion, 1);
```
PVince81, line 126: I guess we don't worry that this might start breaking once we reach v10 :-D

PVince81:
- [ ] improve panel style, doesn't look that nice (could be done separately)

image

PVince81 (commented):
See minor comments.

Will approve after I'm done testing.

PVince81:
- [ ] idea: use telephone link (`<a href="tel:123-456-7890">123-456-7890</a>`) so that if someone opened Talk web on mobile, they can call right away. Could be part of the above.

lib/Controller/SignalingController.php
nickvergessen: substr(…, 1) removes the first char (the `v`) not takes the last one. So it's #futureProof

lib/Controller/RoomController.php
nickvergessen: right

nickvergessen:
> idea: use telephone link (<a href="tel:123-456-7890">123-456-7890</a>) so that if someone opened Talk web on mobile, they can call right away. Could be part of the above.
> 
> actually, this is part of the dialin info, so would require markdown there or something formattable.
> => enhancement for later

Yeah Markup doesn't really work with the clients. Also you can have multiple numbers and the user needs to be aware which one to use instead of clicking the first one. Also if clicking, it should be `phonnumber,00token,pin` which would automatically verify the user already.

PVince81:
> Users that are already in the room need to rejoin as well to get a PIN

In my local test with internal signaling I didn't need to rejoin.
Had admin and bob in the channel, with settings panel open.
After enabling SIP and waiting a bit, the dial-in info + pin appeared there.

nickvergessen:
Yeah that's why https://github.com/nextcloud/spreed/issues/4677#issuecomment-734913425 is ticked ;)

PVince81 (approved):
:+1: looks fine to me

PR #4696, opened 2020-11-30
Adjust local media control position in sidebar

PVince81:
When in sidebar, use the same size as the video for properly centering
the media controls.

Fixes https://github.com/nextcloud/spreed/issues/4668

This was caused by the changes related to the stripe collapsing.


nickvergessen (approved):
Works

PR #4697, opened 2020-11-30
Remove duplicate top bar height scss variable

PVince81:
Seems both come from the same commit by @ma12-co :-)


ma12-co (approved):

nickvergessen (approved):
:see_no_evil:

PR #4699, opened 2020-11-30
Disallow following self in sidebar mode

PVince81:
Remove hover effect and ignore clicks on a local video when in sidebar
mode, to prevent following self.

Fixes https://github.com/nextcloud/spreed/issues/4574


nickvergessen (approved):

ma12-co (approved):

PVince81:
FYI I checked with Talk 10.0 and apparently you could follow yourself there. But then you'd lose media controls, etc. Not very useful. I still think preventing this is the right approach.

Should we backport or would that qualify as loss of feature ?

Issue #4688, opened 2020-11-27
stopFollowingLabel is not defined


nickvergessen:
## Steps to reproduce
1. Start a call with 2 users
2. Click on a video to follow it
3. Back button is very small and hidden

### Expected behaviour
Button is visible and clickable

### Actual behaviour
Button is behind controls:

![Bildschirmfoto von 2020-11-27 16-33-17](https://user-images.githubusercontent.com/213943/100464671-5016cd80-30ce-11eb-920a-75cf512dff97.png)

Also there is a console error:
```
Property or method "stopFollowingLabel" is not defined on the instance but referenced during render. Make sure that this property is reactive, either in the data option, or for class-based components, by initializing the property. See: https://vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties.

found in

---> <LocalVideo> at src/components/CallView/shared/LocalVideo.vue
       <CallView> at src/components/CallView/CallView.vue
         <MainView> at src/views/MainView.vue
           <AppContent>
             <Content>
               <App> at src/App.vue
                 <Root>
```

PVince81:
indeed... maybe it got eaten by some conflict resolution.

I'll look into it.

PVince81:
This PR fixes it: https://github.com/nextcloud/spreed/pull/4700

Issue #3518, opened 2020-05-13
Go back to autopilot mode if re-clicking selected participant


marcoambrosini:
Go back to autopilot mode if re-clicking selected participant in the promoted view

nickvergessen:
Should also go back to grid-view when you follow+unfollow

PVince81:
> Should also go back to grid-view when you follow+unfollow

Similar to https://github.com/nextcloud/spreed/pull/4521 except that instead of a remote screen share as trigger it's the follow mode. After merging we could reuse the mechanism introduced for remembering the last state.

PVince81:
@ma12-co do you still intend to work on this as you're assigned ? Otherwise I'd gladly take over as I already took care of some similar bits.

ma12-co:
The problem with this is that we could have a total mess with the feedbacks. Probably we should just keep the stop following button only?

PVince81:
> Should also go back to grid-view when you follow+unfollow

this bit was adressed in https://github.com/nextcloud/spreed/pull/4700

PR #4700, opened 2020-11-30
Fix follow self + remember state after unfollowing + collapse stripe

PVince81:
Don't show controls when following own video.
Fix missing label for going back.

Fixes https://github.com/nextcloud/spreed/issues/4688

After seeing this I was bothered that my own video was still visible, so I directly looked into hiding it, which led me to do this for all the follow modes:

After following, start presentation mode which will internally remember the last state. When unfollowing, stop presentation mode which will restore the previous state, unless tempered with previously.
    
This will now also automatically collapse the stripe and the own video when following.

This adresses the comment in https://github.com/nextcloud/spreed/issues/3518#issuecomment-676600694

### DIFF
--- a/src/components/CallView/CallView.vue
+++ b/src/components/CallView/CallView.vue
@@ -530,7 +530,7 @@ export default {
 			if (this.isSidebar) {
 				return
 			}
-			this.$store.dispatch('setCallViewMode', { isGrid: false })
+			this.$store.dispatch('startPresentation')
 			this.$store.dispatch('selectedVideoPeerId', peerId)
 			this.isLocalVideoSelected = false
 		},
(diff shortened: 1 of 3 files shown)


nickvergessen:
> After following, start presentation mode which will internally remember the last state. When unfollowing, stop presentation mode which will restore the previous state, unless tempered with previously.

Not sure this is really intended. At least from my PoV I would still like to see the miniature videos of others while following the main talking person. But could be me.

PVince81:
> Not sure this is really intended. At least from my PoV I would still like to see the miniature videos of others while following the main talking person. But could be me.

you can always uncollapse the stripe again which will cancel the "remember last state", so when unfollowing it will stay

But I agree, I was also unsure about this bit.
If we want it that way we'll need to improve startPresentation to allow keeping the stripe uncollapsed for that specific mode.

nickvergessen (approved):

Issue #1763, opened 2019-04-30
Public vs private rooms - Allow users to freely join a room


schiessle:
I had this idea while using Talk. With "public rooms" I don't mean rooms I share by a public link. I want to make a different distinction here.

We have two types of rooms at the moment:

1. 1-1 chat rooms
2. group chat rooms

Both have one in common, you have to invite your chat partner(s). Think about a larger setup of an organization, your sports club, etc. Every time someone new joins your organization you have to add them to the relevant rooms manually. If someone leaves you have to remove them again. A lot of manual work with a high risk to make mistakes.

That's where my idea of "private vs public" rooms starts.

I would define private rooms as rooms where the users in the room decide who can join and who not, by inviting users, sending links around, etc. Basically what we have now.

Public rooms would be rooms someone creates and which are then open to join for everyone. Users could search for public rooms on the left hand sidebar like they search for users and groups today. They can join any of this rooms and participate in the discussion. A bit like I can join a IRC server, look around which channels exists and join the one I'm interested in.

Opinions?

sunjam:
[Circles integration](https://github.com/nextcloud/spreed/issues/860) is a possibility for helping add this functionality since you can create your own "groups" of users as public, private, closed, or secret circles.  Once conversations can be started with a circle, it will be possible to start group discussions.

jancborchardt:
Totally agree! This could also be useful for guests, where you then have the AppNavigation shown with all the conversations specifically set to "Public". In our case all the public team channels.

robertfromont:
I'm looking to replace Slack with NextCloud Talk, and this is the main thing I would need to achieve that; for already-registered users, a set of Conversations they can join without needing a specific invitation.
So that's a :+1: from me!

jcastang:
Same thing for me ! 
Being able to join a room without invitation is a must-have.

PVince81:
reuse the existing public rooms, but make them listable with an extra checkbox in the room settings:
- any guest (public link access) can find it (for public rooms)
- any logged in user can find it (for private rooms)

- make it possible to set/unset the listable property after creation?
- where do guests see that list?

PVince81:
- use the opportunity to move most of the actions menu into the settings tab ? (except full screen)

PVince81:
- only display listed rooms in the left panel:    
    - when searched in top left room search
    - click plus button, ability to join a room instead of creating
    - because otherwise: too crowded and performance impact

PVince81:
[...]

PR #4706, opened 2020-12-01
Add listable scope attribute for conversations

PVince81:
### Related issue

Fixes https://github.com/nextcloud/spreed/issues/1763

### Description

Added ability to set a conversation as listable for regular users and/orguest users from the guest app.
This means those users will be able to search for those rooms and join them even if they haven't joined before.

### Todos
- [x] add new conversation property (mostly copy-pasted from read-only)
   - [x] DB migration for new property
   - [x] backend endpoint to set property
   - [x] settings frontend
   - [x] ~~detect guest app in UI~~ => obsolete ?
- [x] search provider impl
   - [x] impl
- [x] implement search results code based on listable flags
   - [x] show listable only in search results, not regular list
   - [x] filter by scope (regular user vs guest)
   - [x] backend endpoint "/listed-room"
   - [x] frontend
- [x] allow joining listable conversations
- [x] unit tests
   - [x] SystemMessageTest
   - [x] SystemMessage/ListenerTest
   - [x] RoomController, Manager, ParticipantService => cover with integration test ?
   - [x] ~~ListedConversationSearch (if we agree to keep it)~~ removed

### Issues
- [x] BUG: somehow the dialog always shows "Everyone" by default
- [ ] sort out todos in LeftSidebar
   - [x] use ConversationsList ?
   - [ ] sort out "no search results" case, currently nothing appears if no results
   - [x] optional: cancel search when continuing to type? => later in https://github.com/nextcloud/spreed/pull/4742
- [x] implement recognizing guest users (see FIXMEs and TODOs)

### Testing
- [x] verify that cli + defaults there work correctly
- [x] test direct join to private listed room when knowing the URL
- [x] test join via left sidebar
- [ ] test join via global search result
- [x] test with guest users from guest app
- [x] test perms
- [ ] integration tests (see read-only ones)
   - [x] only moderators can change listable property
   - [x] can join room when knowing URL
      - [x] when listed + private (+ system message visible)
      - [x] when listed + public (+ system message visible)
      - [x] when non-listed + public  (+ system message NOT visible)
   - [x] cannot join room when non-listed and private
   - [x] listed rooms of public and private types appear in search results from "/listed-rooms" API, with search term
   - [x] ~~listed rooms of public and private types appear in search results from global search API, with search term~~ => removed

### Open topics

- [x] clarify if flags concept was correct, see https://github.com/nextcloud/spreed/issues/1763#issuecomment-736021636 => adjusted to be a scope, not bitmask any more
- [x] unsure if "all" state is a good idea in regards to extensibility => obsolete
- [x] instead of checkbox, use dropdown for a single-shot choice ? => obsolete
- [x] USER vs USER_SELF_JOINED, see https://github.com/nextcloud/spreed/pull/4706#issuecomment-739883178
- [x] add listable field already in creation dialog ?
- [ ] review discuss the UX concept as the current idea of listable x public/private is already very confusing
- [ ] might need yet another visual indicator in the list to show listed rooms vs non-listed (for moderators to know).
- [x] ~~should we forbid joining listable rooms when calling "/active" endpoint if version used is not v3 ?~~ => no
- [x] password is only asked if not alowed to join a listed room resulting in USER_SELF_JOINED instead of USER, is this ok ?
- [x] should we remove listed conversations from the global search ? => yes, as discussed. they'd cause too much noise there.
- [x] should we force passing in a search term in the "listed-room" API ? or listing all is fine on API level ?


appinfo/routes.php
```diff
@@ -245,6 +245,15 @@
 				'token' => '^[a-z0-9]{4,30}$',
 			],
 		],
+		[
+			'name' => 'Room#setListable',
+			'url' => '/api/{apiVersion}/room/{token}/listable',
+			'verb' => 'PUT',
+			'requirements' => [
+				'apiVersion' => 'v(1|2|3)',
```
PVince81, line 253: should this be limited to v3 ? @nickvergessen

lib/Command/Room/Create.php
```diff
@@ -89,10 +94,21 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
 		$password = $input->getOption('password');
 		$owner = $input->getOption('owner');
 		$moderators = $input->getOption('moderator');
 
+		if (!in_array($readOnly, [null, '0', '1'], true)) {
+			$output->writeln('<error>Invalid value for option "--readonly" given.</error>');
+			return 1;
+		}
+
+		if (!in_array($listable, [null, '0', '1', '2', '3'], true)) {
```
PVince81, line 107: seems I messed up the default

lib/Migration/Version2100Date20201201102528.php
```diff
@@ -0,0 +1,60 @@
...
+namespace OCA\Talk\Migration;
+
+use Closure;
+use Doctrine\DBAL\Types\Type;
+use OCP\DB\ISchemaWrapper;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+/**
+ * Auto-generated migration step: Please modify to your needs!
```
PVince81, line 33: note to self: adjust comment

src/services/conversationsService.js
```diff
@@ -324,6 +324,22 @@ const changeReadOnlyState = async function(token, readOnly) {
...
 }
 
+/**
+ * Change the listable flags
+ * @param {string} token The token of the conversation to be modified
+ * @param {int} listable The new listable flags to set
+ */
+const changeListable = async function(token, listable) {
+	try {
+		const response = await axios.put(generateOcsUrl('apps/spreed/api/v2', 2) + `room/${token}/listable`, {
```
PVince81, line 334: v3 or whatever we decide for the version

appinfo/routes.php
nickvergessen: yes

docs/chat.md
```diff
@@ -147,6 +147,9 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
 * `call_ended` - Call with {user1}, {user2}, {user3}, {user4} and {user5} (Duration 30:23)
 * `read_only_off` - {actor} unlocked the conversation
 * `read_only` - {actor} locked the conversation
+* `listable_participants` - {actor} made the conversation listable only for participants
```
nickvergessen, line 150: This makes no sense 🙈 i would call it not listable if that is the current state

docs/chat.md
```diff
@@ -147,6 +147,9 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
...
+* `listable_users` - {actor} made the conversation listable for regular users
+* `listable_all` - {actor} made the conversation listable for everone which includes users and guest users
```
nickvergessen, line 152:
```suggestion
* `listable_all` - {actor} made the conversation listable for everone which includes users and guests
```

docs/constants.md
```diff
@@ -12,6 +12,11 @@ title: Constants
 * `0` read-write
 * `1` read-only
 
+## Listable scope
+* `0` participants only
+* `1` regular users only, excluding guests
```
nickvergessen, line 17: Is this guest app users or guests?

lib/Chat/SystemMessage/Listener.php
```diff
@@ -163,6 +163,24 @@ public static function register(IEventDispatcher $dispatcher): void {
 				$listener->sendSystemMessage($room, 'read_only_off');
 			}
 		});
+		$dispatcher->addListener(Room::EVENT_AFTER_LISTABLE_SET, static function (ModifyRoomEvent $event) {
+			$room = $event->getRoom();
+
+			if ($room->getType() === Room::CHANGELOG_CONVERSATION) {
+				return;
```
nickvergessen, line 170: This case should never be triggered

lib/Command/Room/Update.php
```diff
@@ -92,6 +98,11 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 			return 1;
 		}
 
+		if (!in_array($listable, [null, '0', '1', '2', '3'], true)) {
```
nickvergessen, line 101: Further down suggestions only mention 0-2

lib/Migration/Version2100Date20201201102528.php
```diff
@@ -0,0 +1,60 @@
...
+			if (!$table->hasColumn('listable')) {
+				$table->addColumn('listable', Type::INTEGER, [
+					'notnull' => true,
+					'length' => 6,
+					'default' => 0,
+				]);
+			}
+		}
+
+		return $schema;
```
nickvergessen, line 58: Only return this when the column needs to be added

lib/Room.php
```diff
@@ -62,6 +62,21 @@ class Room {
 	public const READ_WRITE = 0;
 	public const READ_ONLY = 1;
 
+	/**
+	 * Only visible when joined
+	 */
+	public const LISTABLE_PARTICIPANTS = 0;
+
+	/**
+	 * Searchable by all regular users and moderators, even when not joined, excluding guest users
```
nickvergessen, line 71: Guest app users?

lib/Room.php
```diff
@@ -729,6 +755,42 @@ public function setReadOnly(int $newState): bool {
...
+	 * 						`self::GROUP_CALL` and `self::PUBLIC_CALL`
+	 * @return bool True when the change was valid, false otherwise
+	 */
+	public function setListable(int $newState): bool {
+		$oldState = $this->getListable();
+		if ($newState === $oldState) {
+			return true;
+		}
+
+		if (!in_array($this->getType(), [self::GROUP_CALL, self::PUBLIC_CALL, self::CHANGELOG_CONVERSATION], true)) {
```
nickvergessen, line 770: Why changelog? So you can rejoin your own changelog?

lib/Room.php
```diff
@@ -729,6 +755,42 @@ public function setReadOnly(int $newState): bool {
...
+			return false;
+		}
+
+		if ($newState < 0 || $newState > 3) {
```
nickvergessen, line 774: `!in_array` woth the constants? Makes it future saver to extend by searching for the constants

lib/Command/Room/Create.php
nickvergessen, line 107: Use the constants (and cast to string?)

lib/Search/ListedConversationSearch.php
```diff
@@ -0,0 +1,111 @@
...
+use OCA\Talk\Manager;
+use OCP\IL10N;
+use OCP\IURLGenerator;
+use OCP\IUser;
+use OCP\Search\IProvider;
+use OCP\Search\ISearchQuery;
+use OCP\Search\SearchResult;
+use OCP\Search\SearchResultEntry;
+
+class ListedConversationSearch implements IProvider {
```
nickvergessen, line 36: This should not work via the search from my pov, but the conversation listing/filtering

src/components/ConversationSettings/ModerationSettings.vue
```diff
@@ -21,6 +21,23 @@
 
 <template>
 	<div>
+		<div class="app-settings-subsection">
+			<div id="moderation_settings_listable_users_hint" class="app-settings-section__hint">
+				{{ t('spreed', 'Defines who can find this conversation') }}
+			</div>
+			<div>
+				<label for="moderation_settings_listable_users_conversation_input">{{ t('spreed', 'Listable for') }}</label>
```
nickvergessen, line 29: Invalid string concatenation?

src/components/ConversationSettings/ModerationSettings.vue
```diff
@@ -91,25 +108,36 @@
...
 
 <script>
+import Multiselect from '@nextcloud/vue/dist/Components/Multiselect'
 import { showError, showSuccess } from '@nextcloud/dialogs'
 import { CONVERSATION, WEBINAR } from '../../constants'
 import DatetimePicker from '@nextcloud/vue/dist/Components/DatetimePicker'
 import SipSettings from './SipSettings'
 
+const listableOptions = [
+	{ value: 2, label: t('spreed', 'Everyone') },
```
nickvergessen, line 118: src/constants.js 😎

src/components/ConversationSettings/ModerationSettings.vue
```diff
@@ -91,25 +108,36 @@
...
+	{ value: 1, label: t('spreed', 'Regular users, without guests') },
```
nickvergessen, line 119: Needs to be clear if its the guest app and to avoid confusion should only say registered/logged in users when guests is not there

nickvergessen (changes requested):

docs/chat.md
PVince81: here with "participant" is implied that the room is only listable by people who already joined them

but I'm fine changing it to "listable_none"

docs/constants.md
PVince81: as originally clarified in the  ticket https://github.com/nextcloud/spreed/issues/1763#issuecomment-721008640 it's for "guest users from the guest app"

because for public guests (invited by token) we don't have a UI for searching for rooms or switching

lib/Search/ListedConversationSearch.php
PVince81: yes and no

it's still useful to be able to find public conversations with the global search

I have a few new commits that use this provider also for the left side. But that case feels strange.
I'll look into conversation filtering next.

PVince81:
I've pushed a few more commits which I had locally which make the search result visible in the left sidebar.

I'm not too happy as it uses the global search provider and it feels wrong.

Tomorrow I'll look into the endpoint that lists joined conversations and see if I can extend it somehow.

nickvergessen:
We once also discussed a dedicated view for them? Having it on the normal endpoint would make integration for the mobile clients pretty easy.

PVince81:
- [ ] make sure that other endpoints also allow retrieving listable rooms by token (like getSingleRoom, etc)

PVince81:
> We once also discussed a dedicated view for them? Having it on the normal endpoint would make integration for the mobile clients pretty easy.

yeah, but as a first step we agreed to not handle "guests outside the guest app" for now due to the missing view

PVince81:
Adjusted after review.

Also modified to use a new endpoint "/listed-room" to search and used the `Conversation` component to display results.
Regarding the endpoint, I'm not sure how many fields we want / don't want to be populated, I hoped what I picked is enough for now. A lot of values might be null/zero/default as they are not relevant when the user is not a participant (has not joined the room).

Next up: making it possible to actually join those listable rooms

PVince81:
- [ ] test listable rooms with password...

PVince81:
- [ ] might need yet another visual indicator in the list to show listed rooms vs non-listed.

somehow I dislike the term as it somehow conflicts with "public" vs "non-public". might need to come up with better terms before this is finalized

PVince81:
It is now possible to join listable rooms by clicking on the left sidebar.

Next up: improve the UX after clicking:
- [ ] some JS errors appear
- [ ] it doesn't directly appear in the list and need to wait for the regular refresh => fetch it earlier + scroll to it

PVince81:
after a lot of trial and error I managed to make the transition look clean when clicking on a listable entry: the left sidebar will appear with an entry saying "Joining..." until it gets refreshed.

as an added bonus, the left sidebar now scrolls to the item after joining.

still unresolved issues:
- participant not found in ["updateSessionId" action](https://github.com/nextcloud/spreed/blob/2fc5379f97ec72a7ea5f50519316d33b733b7eed/src/services/participantsService.js#L54)
   - because at the time of joining it's not available yet and might need initialization that only happens further down the road.
   - also somehow it seems the lines are swapped, shouldn't we first set the current participant before calling getCurrentParticipant ?

nickvergessen:
> after a lot of trial and error I managed to make the transition look clean when clicking on a listable entry: the left sidebar will appear with an entry saying "Joining..." until it gets refreshed.

The response of the join should contain all the conversation info iirc?

PVince81:
@nickvergessen if the POST to "/active" does, then could try and add it to the store right away.

So far I noticed that the conversation is re-fetched later on which automagically makes it work, somehow.

I still do wonder how non-listable conversations can work correctly in this code path, isn't getParticipantIdentifier() outdated here https://github.com/nextcloud/spreed/blob/2fc5379f97ec72a7ea5f50519316d33b733b7eed/src/services/participantsService.js#L56 ? it might have the value from the previous room, unless I misunderstand something in the "join room dance"

docs/conversation.md
```diff
@@ -4,6 +4,63 @@
...
+        + `200 OK`
+        + `401 Unauthorized` when the user is not logged in
+
+    - Header:
+
+        field | type | Description
+        ------|------|------------
+        `searchTerm` | string | search term
+
+    - Data:
```
nickvergessen, line 23: Use this instead of duplicating it:
```suggestion
    - Data: See array definition in `Get user´s conversations`
```

We do already for some other things too and the difference should be 0

lib/Command/Room/Create.php
```diff
@@ -89,10 +94,26 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 		$groups = $input->getOption('group');
 		$public = $input->getOption('public');
 		$readonly = $input->getOption('readonly');
+		$listable = $input->getOption('listable');
 		$password = $input->getOption('password');
 		$owner = $input->getOption('owner');
 		$moderators = $input->getOption('moderator');
 
+		if (!in_array($readOnly, [null, '0', '1'], true)) {
```
nickvergessen, line 102: constants here too?

lib/Command/Room/Create.php
```diff
@@ -140,6 +162,14 @@ public function completeOptionValues($optionName, CompletionContext $context) {
 			case 'owner':
 			case 'moderator':
 				return $this->completeParticipantValues($context);
+			case 'readonly':
+				return ['1', '0'];
```
nickvergessen, line 166: constants

lib/Controller/RoomController.php
```diff
@@ -210,6 +211,37 @@ public function getRooms(int $noStatusUpdate = 0): DataResponse {
...
+		$rooms = $this->manager->getListedRoomsForUser($this->userId, $searchTerm);
+
+		$return = [];
+		foreach ($rooms as $room) {
+			try {
+				$roomData = $this->formatRoom($room, null);
+				// since formatRoom will break early due to having no participant,
+				// we populate the remaining base attributes here
+				$return[] = $this->populateBaseRoomData($roomData, $room, $this->userId);
+				// TODO: should we populate more ?
```
nickvergessen, line 235: yeah lets stick to the normal output please. Same array as with normal conversations. Otherwise we have follow up regressions with search results not having the same attributes as conversations from the conversation list

PVince81: we do have the normal output but with zeroed values, as some cannot be populated before we join

I'll check if there's more that we can populate though

PVince81:
participant not found error solved by adding the conversation early, as discussed: https://github.com/nextcloud/spreed/pull/4706/commits/2baed21e41225e41c440ed4821c6111231ab5c6f

lib/Controller/RoomController.php
```diff
@@ -587,19 +620,12 @@ protected function formatRoomV2andV3(Room $room, ?Participant $currentParticipan
 		$attendee = $currentParticipant->getAttendee();
 		$userId = $attendee->getActorType() === Attendee::ACTOR_USERS ? $attendee->getActorId() : '';
 
+		$roomData = $this->populateBaseRoomData($roomData, $room, $userId);
```
nickvergessen, line 623: I would just use the `if ($isSIPBridgeRequest) {` block too when:
```php
if ($isSIPBridgeRequest || ($room->getListable() && !$currentParticipant instanceof Participant)) {
```

lib/Controller/RoomController.php
```diff
@@ -710,6 +736,33 @@ protected function formatRoomV2andV3(Room $room, ?Participant $currentParticipan
...
+
+		return array_merge($roomData, [
+			'name' => $room->getName(),
+			'displayName' => $room->getDisplayName($userId),
+			'objectType' => $room->getObjectType(),
+			'objectId' => $room->getObjectId(),
+			'readOnly' => $room->getReadOnly(),
+			'listable' => $room->getListable(),
+			'hasCall' => $room->getActiveSince() instanceof \DateTimeInterface,
+			'lobbyState' => $room->getLobbyState(),
```
nickvergessen, line 762: missing lobbyTimer :boom:

lib/Manager.php
```diff
@@ -329,6 +335,54 @@ public function getRoomsForUser(string $userId, bool $includeLastMessage = false
...
+			->leftJoin('a', 'talk_sessions', 's', $query->expr()->andX(
+				$query->expr()->eq('a.id', 's.attendee_id')
+			))
+			->where($query->expr()->isNull('a.id'))
+			->andWhere($query->expr()->in('r.type', $query->createNamedParameter($allowedRoomTypes, IQueryBuilder::PARAM_INT_ARRAY)))
+			->andWhere($query->expr()->in('r.listable', $query->createNamedParameter($allowedListedTypes, IQueryBuilder::PARAM_INT_ARRAY)));
+
+		if ($term !== '') {
+			$query->andWhere(
+				$query->expr()->ilike('name', $query->createNamedParameter(
```
nickvergessen, line 369:
```suggestion
				$query->expr()->iLike('name', $query->createNamedParameter(
```

lib/Migration/Version2100Date20201201102528.php
```diff
@@ -0,0 +1,62 @@
...
+use Closure;
+use Doctrine\DBAL\Types\Type;
+use OCP\DB\ISchemaWrapper;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+/**
+ * Add listable column to the rooms table.
+ */
+class Version2100Date20201201102528 extends SimpleMigrationStep {
```
nickvergessen, line 35:
```suggestion
class Version11000Date20201201102528 extends SimpleMigrationStep {
```

lib/Migration/Version2100Date20201201102528.php
```diff
@@ -0,0 +1,62 @@
...
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable('talk_rooms')) {
+			$table = $schema->getTable('talk_rooms');
+
+			if (!$table->hasColumn('listable')) {
+				$table->addColumn('listable', Type::INTEGER, [
+					'notnull' => true,
```
nickvergessen, line 51:
```suggestion
					'notnull' => false,
```

lib/Service/ParticipantService.php
```diff
@@ -161,12 +166,32 @@ public function joinRoom(Room $room, IUser $user, string $password, bool $passed
...
 
-			// User joining a public room, without being invited
-			$this->addUsers($room, [[
-				'actorType' => Attendee::ACTOR_USERS,
-				'actorId' => $user->getUID(),
-				'participantType' => Participant::USER_SELF_JOINED,
-			]]);
+			// User joining a group call through listing
+			if (
+				$room->getType() === Room::GROUP_CALL && (
```
nickvergessen, line 171: or public

src/components/LeftSidebar/ConversationsList/Conversation.vue
```diff
@@ -192,6 +197,11 @@ export default {
 		},
 
 		conversationInformation() {
+			// temporary item while joining
+			if (!this.isSearchResult && !this.item.actorId) {
+				return t('spreed', 'Joining conversation...')
```
nickvergessen, line 202:
```suggestion
				return t('spreed', 'Joining conversation …')
```

src/services/conversationsService.js
```diff
@@ -110,7 +137,7 @@ const searchPossibleConversations = async function(searchText, token, onlyUsers)
 	}
 
 	try {
-		return await axios.get(generateOcsUrl('core/autocomplete', 2) + `get`, {
+		return axios.get(generateOcsUrl('core/autocomplete', 2) + `get`, {
```
nickvergessen, line 140: 😐 this changes the return from actual data to a promise. doesn't look intended. So better not hide this in a big feature PR?

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -672,6 +676,32 @@ public function userChangesReadOnlyStateOfTheRoom($user, $newState, $identifier,
...
+	 *
+	 * @param string $user
+	 * @param string $newState
+	 * @param string $identifier
+	 * @param string $statusCode
+	 * @param string $apiVersion
+	 */
+	public function userChangesListableScopeOfTheRoom($user, $identifier, $newState, $statusCode, $apiVersion = 'v3') {
+		$this->setCurrentUser($user);
+		if ($newState === 'joined') {
```
nickvergessen, line 690: participants?

src/services/conversationsService.js
PVince81: some leftover from a previous attempt, I'll remove

lib/Service/ParticipantService.php
PVince81: public is already handled below, I've left it to "self joined" but still need to test whether it works that way or makes sense

nickvergessen: When you click on a listed conversation you should always be fully added

lib/Migration/Version2100Date20201201102528.php
PVince81: why have both 0 and null ?
I picked this from read-only which is also not nullable.

tests/integration/features/bootstrap/FeatureContext.php
PVince81: => "none"

lib/Migration/Version2100Date20201201102528.php
nickvergessen: 0 and `''` are stored as null by Oracle. Don't ask :speak_no_evil: just allow nullable

lib/Service/ParticipantService.php
```diff
@@ -161,12 +166,32 @@ public function joinRoom(Room $room, IUser $user, string $password, bool $passed
...
+					// this check should have happened earlier already but let's stay defensive
+					$room->getListable() === Room::LISTABLE_ALL || (
+						$room->getListable() === Room::LISTABLE_USERS && !$this->groupManager->isInGroup($user->getUID(), 'guest_app')
```
nickvergessen, line 174: use the function from the OCA\Talk\Manager instead, so we don't have multiple places with this magic string?

PVince81: I couldn't inject Manager without causing an infinite loop of dependencies.

is there an approach for this ?

PVince81:
- [ ] BUG: public link now automatically joins call (not happening on master)

lib/Chat/SystemMessage/Listener.php
```diff
@@ -200,9 +214,21 @@ public static function register(IEventDispatcher $dispatcher): void {
...
+				}
+
+				$userJoinedFileRoom = $room->getObjectType() === 'file' && $participantType !== Participant::USER_SELF_JOINED;
+
+				// add a message "X joined the conversaion", whenever user $userId:
+				if (
+					// - has joined a file room but not through a public link
+					$userJoinedFileRoom
+					// - has been added by another user
+					|| $userId !== $participant['actorId']
```
nickvergessen, line 229: So the idea of this was always to not show that you added yourself, when you create a conversation. Maybe you can add this as a comment

lib/Controller/RoomController.php
```diff
@@ -210,6 +211,33 @@ public function getRooms(int $noStatusUpdate = 0): DataResponse {
...
+	 * Get listed rooms with optional search term
+	 *
+	 * @NoAdminRequired
+	 *
+	 * @param string $searchTerm search term
+	 * @return DataResponse
+	 */
+	public function getListedRooms(string $searchTerm = ''): DataResponse {
+		$event = new UserEvent($this->userId);
+		$this->dispatcher->dispatch(self::EVENT_BEFORE_LISTED_ROOMS_GET, $event);
```
nickvergessen, line 224: This should not be needed. Drop it?

We only need it in the other function to check if the changelog needs to be readded and new messages posted

lib/Service/ParticipantService.php
nickvergessen: Add the function to another service which can be injected in both, but well too much effort. just keep it like this for now.

src/services/conversationsService.js
```diff
@@ -88,6 +88,33 @@ const checkTalkVersionHash = function(response) {
...
+
+		if (maintenanceWarning) {
+			maintenanceWarning.hideToast()
+			maintenanceWarning = null
+		}
+
+		return response
+	} catch (error) {
+		if (error.response && error.response.status === 503 && !maintenanceWarning) {
+			maintenanceWarning = showError(t('spreed', 'Nextcloud is in maintenance mode, please reload the page'), {
```
nickvergessen, line 111: This should be handled globally. Feels weird that this search would trigger it but posting a message would not?

nickvergessen: so something for later?

PVince81: I copy-pasted this from another method. Agreed that this should be global. I'll remove it

lib/Controller/RoomController.php
PVince81: Ok. I just added it for consistency, but if we think no one will use it then I can remove.

PVince81:
- [ ] recheck joining permissions and mode

This is how I think it should be working:
|                            | Regular user | Guest user (guest app) | Public guest (no account) |
|----------------------------|--------------|------------------------|---------------------------|
| Group call, non-listable   | Cannot join  | Cannot join            | Cannot join               |
| Public call, non-listable  | Join USER    | Join USER              | Join USER_SELF_JOINED     |
| Group call, user-listable  | Join USER    | Cannot join            | Cannot join               |
| Public call, user-listable | Join USER    | Join USER              | Join USER_SELF_JOINED     |
| Group call, all-listable   | Join USER    | Join USER              | Cannot join               |
| Public call, all-listable  | Join USER    | Join USER              | Join USER_SELF_JOINED     |

The USER_SELF_JOINED would only ever be for "guests that have no account" (also see https://github.com/nextcloud/spreed/issues/4717 as Talk 10 behaves differently for logged in users)

lib/Chat/SystemMessage/Listener.php
PVince81: I checked and now when creating it also doesn't show up, maybe because when creating you join as "USER_MODERATOR"

PVince81:
- [ ] forbid joining any listable rooms if API version is < v3 ?? this way we can say that the new behavior is only in v3

PVince81:
I've pushed some new integration tests.
Some fail because of my misunderstanding of the fact that a USER_SELF_JOINED is not returned as participant of a room. It seems that many existing tests work this around by using the call endpoint instead and counting the peers.

So probably I'll just remove the assertion about "is a participant" and simply rely on the return value of the "/active" call to check if the permission was granted.

nickvergessen:
From my POV its:

|                            | Regular user | Guest user (guest app) | Public guest (no account) |
|----------------------------|--------------|------------------------|---------------------------|
| **Group call, non-listable**   | Cannot join  | Cannot join            | Cannot join               |
| **Public call, non-listable**  | USER_SELF_JOINED    | USER_SELF_JOINED               | GUEST     |
| **Group call, user-listable**  | USER    | Cannot join            | Cannot join               |
| **Public call, user-listable** | USER    | USER_SELF_JOINED              | GUEST     |
| **Group call, all-listable**   | USER    | USER              | Cannot join               |
| **Public call, all-listable**  | USER    | USER              | GUEST     |

PVince81:
It took me a while to understand the self joined concept.
Based on my current understanding from https://github.com/nextcloud/spreed/issues/4717 I agree with your proposal as it would keep the behavior consistent until we reach a point where we can adjust it.

- [ ] also make regular user joining "non-listable public" use "USER_SELF_JOINED"

nickvergessen:
> It took me a while to understand the self joined concept.

TL;DR: Someone adds you: "USER", you access/join it on your own "USER_SELF_JOINED"

PVince81:
rebased

and I've adjusted the acceptance tests to encode the latter table into test scenarios.

- [ ] troubles with guest user from guest app: https://github.com/nextcloud/guests/issues/505 (the step that adds the user to guest_app doesn't work but [fails silently](https://github.com/nextcloud/guests/issues/504))

PVince81:
I managed to create guest account users (let's call them that for now?) by adding an endpoint to spreedcheats that grabs the GuestManager from the guests app directly.

So now next step is adjusting the actual behavior to match the expected one from the table above.

PVince81:
okay, now the joining-listable-rooms.feature is aligned with the table above and all tests are green

To fix them I also had to set a password for the guest account user and whitelist the spreed app, this is now done in a single test step so it can be done on-demand in scenarios.

nickvergessen:
Some more tests failing:

> 1) OCA\Talk\Tests\php\Controller\SignalingControllerTest::testLeaveRoomWithOldSession

PVince81:
sharing failures unrelated, also happens on master: https://github.com/nextcloud/spreed/issues/4736

nickvergessen:
Not convinced by that. Already see the reports on h1 that the conversation is set to invisible but still shown... But we have more time to find a good word, as long as the api and backend is settled

PVince81:
now also added in the "new conversation dialog":
image

I've reused the same component there but the saving happens later at creation, it needed an extra request.
Not sure if we should bundle "listable" in the initial creation request instead.

PVince81:
While developing this I was slightly bothered that search requests did not get cancelled while typing.
Since this is slightly off topic for this PR I pushed it to another one https://github.com/nextcloud/spreed/pull/4742 which depends on this one here due to the additional search query for listable conversations.

PVince81:
I've squashed some commits together that belonged together and resolved the DCO issue.

lib/Migration/Version11000Date20201201102528.php
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Talk\Migration;
+
+use Closure;
+use Doctrine\DBAL\Types\Type;
```
nickvergessen, line 27:
```suggestion
use Doctrine\DBAL\Types\Types;
```

PVince81:
- [ ] delete global search code as discussed with @nickvergessen

lib/Migration/Version11000Date20201201102528.php
```diff
@@ -0,0 +1,62 @@
...
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable('talk_rooms')) {
+			$table = $schema->getTable('talk_rooms');
+
+			if (!$table->hasColumn('listable')) {
+				$table->addColumn('listable', Type::INTEGER, [
+					'notnull' => false,
+					'length' => 6,
+					'default' => 0,
```
nickvergessen, line 53:
```suggestion
				$table->addColumn('listable', Types::SMALLINT, [
					'notnull' => false,
					'default' => 0,
					'unsigned' => true,
```

lib/Migration/Version11000Date20201201102528.php
```diff
@@ -0,0 +1,62 @@
...
+				]);
```
nickvergessen, line 54: We need an index on the column, otherwise the search will die at some point.

lib/Service/ParticipantService.php
```diff
@@ -166,12 +171,34 @@ public function joinRoom(Room $room, IUser $user, string $password, bool $passed
...
-				'participantType' => Participant::USER_SELF_JOINED,
-			]]);
+			// User joining a group or public call through listing
+			if (
+				($room->getType() === Room::GROUP_CALL || $room->getType() === Room::PUBLIC_CALL) && (
+					// this check should have happened earlier already but let's stay defensive
+					$room->getListable() === Room::LISTABLE_ALL || (
+						$room->getListable() === Room::LISTABLE_USERS &&
+						// queried here to avoid loop deps
+						!\OC::$server->query(\OCA\Talk\Manager::class)->isGuestUser($user->getUID())
```
nickvergessen, line 181:
```suggestion
						!\OC::$server->get(\OCA\Talk\Manager::class)->isGuestUser($user->getUID())
```

src/components/ConversationSettings/ListableSettings.vue
```diff
@@ -0,0 +1,131 @@
...
+			<label for="moderation_settings_listable_conversation_input">{{ t('spreed', 'Visible for') }}</label>
+			<Multiselect id="moderation_settings_listable_conversation_input"
+				v-model="listable"
+				:options="listableOptions"
+				:placeholder="t('spreed', 'Visible for')"
+				label="label"
+				track-by="value"
+				:disabled="isListableLoading"
+				aria-describedby="moderation_settings_listable_conversation_hint"
+				@input="saveListable" />
```
nickvergessen, line 37: Needs to be a bit wider
Bildschirmfoto von 2020-12-11 09-41-07

src/components/ConversationSettings/ModerationSettings.vue
```diff
@@ -21,6 +21,7 @@
 
 <template>
 	<div>
+		<ListableSettings :token="token" />
```
nickvergessen, line 24: Currently the settings block is captioned with `Meeting settings` which is weird.
Only the lobby refers to meetings. "listable" and "locking" should be `Conversation settings` instead of `Meeting settings`

src/services/conversationsService.js
```diff
@@ -324,9 +343,26 @@ const changeReadOnlyState = async function(token, readOnly) {
...
+ * @param {int} listable The new listable scope to set
+ */
+const changeListable = async function(token, listable) {
+	try {
+		const response = await axios.put(generateOcsUrl('apps/spreed/api/v3', 2) + `room/${token}/listable`, {
+			scope: listable,
+		})
+		return response
+	} catch (error) {
+		console.debug('Error while updating listable scope: ', error)
```
nickvergessen, line 358: Since you catch here, `await this.$store.dispatch('setListable', {` doesn't catch anything anymore and does not `showError`, but always shows the success message.
can be tested by putting
```php
```
In the first line of the function in the RoomController.

From my pov the services should never catch.

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -121,6 +133,81 @@ public function tearDown() {
...
+	}
+
+	/**
+	 * @Then /^user "([^"]*)" cannot find any listed rooms with (\d+)(?: \((v(1|2|3))\))?$/
+	 *
+	 * @param string $user
+	 * @param int $statusCode
+	 * @param string $apiVersion
+	 */
+	public function userCannotFindAnyListedRoomsWithStatus($user, $statusCode, $apiVersion = 'v1') {
```
nickvergessen, line 158:
```suggestion
	 * @Then /^user "([^"]*)" cannot find any listed rooms with (\d+)(?: \((v(3))\))?$/
	 *
	 * @param string $user
	 * @param int $statusCode
	 * @param string $apiVersion
	 */
	public function userCannotFindAnyListedRoomsWithStatus($user, $statusCode, $apiVersion = 'v3') {
```

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -121,6 +133,81 @@ public function tearDown() {
...
+	}
+
+	/**
+	 * @Then /^user "([^"]*)" cannot find any listed rooms with term "([^"]*)"(?: \((v(1|2|3))\))?$/
+	 *
+	 * @param string $user
+	 * @param string $term
+	 * @param string $apiVersion
+	 */
+	public function userCannotFindAnyListedRoomsWithTerm($user, $term, $apiVersion = 'v1') {
```
nickvergessen, line 171:
```suggestion
	 * @Then /^user "([^"]*)" cannot find any listed rooms with term "([^"]*)"(?: \((v(3))\))?$/
	 *
	 * @param string $user
	 * @param string $term
	 * @param string $apiVersion
	 */
	public function userCannotFindAnyListedRoomsWithTerm($user, $term, $apiVersion = 'v3') {
```

nickvergessen: and so on

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -793,6 +903,36 @@ public function userChangesReadOnlyStateOfTheRoom($user, $newState, $identifier,
 		$this->assertStatusCode($this->response, $statusCode);
 	}
 
+	/**
+	 * @Then /^user "([^"]*)" allows listing room "([^"]*)" for "(none|users|all|\d+)" with (\d+)(?: \((v(1|2|3))\))?$/
```
nickvergessen, line 907:
```suggestion
	 * @Then /^user "([^"]*)" allows listing room "([^"]*)" for "(none|users|all|\d+)" with (\d+)(?: \((v(3))\))?$/
```

nickvergessen (changes requested):

PVince81:
- [ ] add missing check for guests in resolveRoomDisplayName (see FIXME in lib/Manager.php)

lib/Search/ListedConversationSearch.php
PVince81: as discussed, we'll remove it

PVince81:
DCO again ? :-(

lib/Manager.php
```diff
@@ -447,8 +511,20 @@ public function getRoomForUserByToken(string $token, ?string $userId, bool $incl
...
-			return $room;
+		if ($userId !== null) {
+			// user already joined that room before
+			if ($row['actor_id'] === $userId) {
+				return $room;
+			}
+
+			// never joined before but found in listing
+			$listable = null;
+			if (isset($row['listable'])) {
```
nickvergessen, line 522: Why isset? The query will always select it?

nickvergessen:
> DCO again ? :-(

The revert is the problem.

PVince81:
removed redundant isset, fixed DCO issue, rebased

expecting green now :-)

nickvergessen:
Version bump got lost on rebase because master already is `11.0.0-dev.9`

docs/conversation.md
```diff
@@ -4,6 +4,63 @@
...
+        `lobbyTimer` | int | * | Timestamp when the lobby will be automatically disabled (only available with `webinary-lobby` capability)
+        `sipEnabled` | int | v3 | SIP enable status (0-1)
+        `canEnableSIP` | int | v3 | Whether the given user can enable SIP for this conversation. Note that when the token is not-numeric only, SIP can not be enabled even if the user is permitted and a moderator of the conversation
+        `unreadMessages` | int | * | Number of unread chat messages in the conversation (only available with `chat-v2` capability)
+        `unreadMention` | bool | * | Flag if the user was mentioned since their last visit
+        `lastReadMessage` | int | * | ID of the last read message in a room (only available with `chat-read-marker` capability)
+        `lastMessage` | message | * | Last message in a conversation if available, otherwise empty
+        `objectType` | string | * | The type of object that the conversation is associated with; "share:password" if the conversation is used to request a password for a share, otherwise empty
+        `objectId` | string | * | Share token if "objectType" is "share:password", otherwise empty
```
nickvergessen, line 62: Drop please

nickvergessen (approved):

PVince81:
as discussed, I've adjusted the listable dropdown width and tested with the dyslexia font:
image

PVince81:
we didn't backport the conversation settings: https://github.com/nextcloud/spreed/pull/4576

unless critical, I'd rather not backport these huge changes

nickvergessen:
we shouldn't backport this at all...

PVince81:
ah, seems you asked the bot to backport a single commit about the ctrl+c... not sure why there would be conflict, maybe the bot tried to backport all

I can do this manually

PVince81:
here we go: https://github.com/nextcloud/spreed/pull/4759

nickvergessen:
Ah, no the problem is you rebased and the sha of the commit didn't exist anymore.

PVince81:
this PR will remove the guest manager cheat: https://github.com/nextcloud/spreed/pull/4764

PR #4708, opened 2020-12-01
Don't select timestamp but selecting messages

PVince81:
Fixes copying of multiple messages.
Now only the message bodies are included and no more timestamps.

### Steps to reproduce

1. Send a message "one message"
2. Send a message "second line"
3. Send a message "third, repeat"
4. With the mouse, start dragging before the "o" and down to the end of "repeat" to select the messages
5. Right click + copy
6. Paste into a text editor

### Before:
Timestamps included:
```
one message

second line
15:41
third, repeat
```

### After:
No more timestamps:
```
one message
second line
third, repeat
```

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -425,6 +425,7 @@ export default {
 			justify-self: flex-start;
 			justify-content:  space-between;
 			position: relative;
+			user-select: none;
 			display: flex;
 			color: var(--color-text-maxcontrast);
 			font-size: $chat-font-size;


nickvergessen:
Not sure if that is actually good or bad.
I would like the timestamp, but from my POV it should be: `time1 user message1 time2 user message2` and currently you copy `user message1 time1 message2 (time2)`

PVince81:
yeah, that would require adding some ghost elements that are selectable (not sure if even possible) with the timestamp
but maybe at some point (if we redesign) we can move the timestamps to the left side and then it would work automagically

for the current state I prefer no timestamps than having a big blob of text that I need to manually cleanup before posting

nickvergessen:
CI died on that day, can you rebase?

nickvergessen (approved):

PVince81:
rebased

PR #4723, opened 2020-12-08
Update Spreed Cheats for 21

### LINKED ISSUE #4720: Move away from app.php to IBootstrap
The log often shows this line:
```
/appinfo/app.php is deprecated, use \OCP\AppFramework\Bootstrap\IBootstrap on the application class instead.
```

nickvergessen:
Fix #4720

### DIFF
--- a/tests/integration/spreedcheats/lib/AppInfo/Application.php
+++ b/tests/integration/spreedcheats/lib/AppInfo/Application.php
@@ -1,8 +1,8 @@
 <?php
+
+declare(strict_types=1);
 /**
- * @author Joas Schilling <coding@schilljs.com>
- *
- * @copyright Copyright (c) 2017 Joas Schilling <coding@schilljs.com>
+ * @copyright Copyright (c) 2020 Joas Schilling <coding@schilljs.com>
  *
  * @license GNU AGPL version 3 or any later version
  *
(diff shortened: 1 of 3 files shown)


tests/integration/spreedcheats/lib/AppInfo/Application.php
```diff
@@ -21,4 +21,23 @@
...
+use OCP\AppFramework\Bootstrap\IRegistrationContext;
+
+class Application extends App implements IBootstrap {
+	public const APP_ID = 'spreedcheats';
+	
+	public function __construct() {
+		parent::__construct(self::APP_ID);
+	}
+
+	public function register(IRegistrationContext $context): void {
```
PVince81, line 38: strange that this is not already defined in the base class to make it optional

but maybe it's very rare to have an app that uses zero services :smile:

PVince81 (approved):
:+1: goodbye flood of warnings

tests/integration/spreedcheats/lib/AppInfo/Application.php
nickvergessen: well only the IBootstrap is using this.
Also we throw a more aggressive warning when you use app.php and IBootstrap.

PR #4735, opened 2020-12-09
Create a new table for internal signaling and guest names with a PK

nickvergessen:

### DIFF
--- a/lib/Migration/Version11000Date20201209142525.php
+++ b/lib/Migration/Version11000Date20201209142525.php
@@ -1,0 +1,144 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2020, Joas Schilling <coding@schilljs.com>
+ *
+ * @author Joas Schilling <coding@schilljs.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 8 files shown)


PVince81:
@nickvergessen mind adding more context to the PR to explain what this is about and why we need it ? (our future selves will be thankful)

PVince81:
aha, the branch name contains a clue "cluster-compatibility"

nickvergessen:
Yeah, so tables need a primary key in order to work in Galera and other clusters, so they can be replicated easily.

Server was fixed in https://github.com/nextcloud/server/commit/d5df033ede1f2b19b87c365b1f808130023edf17#diff-723d474847ce3f8926be66be2147d580e43f9a021655b56b6942d683ed408b49

lib/Migration/Version11000Date20201209142525.php
```diff
@@ -0,0 +1,144 @@
...
+				'session_hash' => $insert->createParameter('session_hash'),
+				'display_name' => $insert->createParameter('display_name'),
+			]);
+
+		$query = $this->connection->getQueryBuilder();
+		$query->select('*')
+			->from('talk_guests');
+
+		$result = $query->execute();
+		while ($row = $result->fetch()) {
```
PVince81, line 135: I guess there was no way to make this into a single query like "insert into", which might not work with all DBs.

hope the number of guests is low on most setup so that the downtime is minimal

PVince81 (approved):
:+1: looks fine, see minor comment.

I did a quick test with some data in the table and the guest names were moved properly to the new table.

PVince81:
there's a chat integration test that failed, maybe rebase to try again

lib/Migration/Version11000Date20201209142525.php
nickvergessen: We did worse copies on other updates. But well that's why we try not to backport such things

nickvergessen:
It was actually a valid fail for inserting a guest name.

PVince81 (approved):
:+1: nice catch

PR #4742, opened 2020-12-10
Cancel search while typing

PVince81:
- [x] REQUIRES https://github.com/nextcloud/spreed/pull/4706 because of future conflicts

The code is in the very last commit if you want to have a peek.

### DIFF
--- a/src/services/conversationsService.js
+++ b/src/services/conversationsService.js
@@ -91,29 +91,24 @@ const checkTalkVersionHash = function(response) {
 /**
  * Fetch listed conversations
  * @param {string} searchTerm The string that will be used in the search query.
+ * @param {object} options options
  */
-const searchListedConversations = async function(searchTerm) {
-	const response = await axios.get(generateOcsUrl('apps/spreed/api/v3', 2) + 'listed-room', {
+const searchListedConversations = async function(searchTerm, options) {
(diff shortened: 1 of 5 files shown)


nickvergessen:
Sounds like a good idea

src/services/conversationsService.js
```diff
@@ -91,29 +91,24 @@ const checkTalkVersionHash = function(response) {
 /**
  * Fetch listed conversations
  * @param {string} searchTerm The string that will be used in the search query.
+ * @param {object} options options
  */
-const searchListedConversations = async function(searchTerm) {
-	const response = await axios.get(generateOcsUrl('apps/spreed/api/v3', 2) + 'listed-room', {
+const searchListedConversations = async function(searchTerm, options) {
+	return axios.get(generateOcsUrl('apps/spreed/api/v3', 2) + 'listed-room', Object.assign(options, {
```
PVince81, line 97: note: await is gone because redundant (if you leave it here eslint will warn about it)

nickvergessen: 🍿

nickvergessen (approved):

PVince81:
I was so focused on testing the cancellability that I forgot to test if the search actually works. A copy-paste issue had sneaked in, fixed here: https://github.com/nextcloud/spreed/pull/4835

Issue #4719, opened 2020-12-07
Telegram matterbridge does not send or receive messages


psuter80:
I can't get a matterbridge to telegram up and running. not sure if i'm doing it all right, so here are all the steps I've executed to set up the bridge: 

## Steps to reproduce
1. create a new telegram bot using bothfather and note the API Key  (1234567890:AAH7pHpFOoKL8PVy15s6Kkqnx31hZUviPUc)
2. create a new group in telegram and add the bot to it
3. send "/command" message to the group
4. navigate to https://api.telegram.org/bot1234567890:AAH7pHpFOoKL8PVy15s6Kkqnx31hZUviPUc/getUpdates?offset=0
5. note down the chat id field (negative number i.e.: -987654321)
6. in nextcloud, as admin enable talk app and install matterbridge via nextcloud admin UI
7. as user in nextcloud, add new group conversation without participants
8. open the group conversation and go to the settings tab on the right side 
9. add a new telegram bridge
10. enter api key and chat id as noted before and click save 
11. check the enabled box to start the matterbridge service 
12. send message to the chat group

### Expected behaviour
message should appear in telegram group

### Actual behaviour
message only appears in nextcloud talk group but is not sent to telegram group. the same is true for the opposite direction, no messages from telegram are displayed in nextcloud talk. 

## Talk app

**Talk app version:** 10.0.3
**Server/PHP:** PHP 7.2.24-0ubuntu0.18.04.7
**Browser:** Firefox 83.0 on Ubuntu 18.04

matterbridge log: 
```
time="2020-12-07T15:20:55+01:00" level=info msg="Running version 1.20.0 1a5353d7" prefix=main
time="2020-12-07T15:20:55+01:00" level=info msg="Parsing gateway myGateway" prefix=router
time="2020-12-07T15:20:55+01:00" level=info msg="Starting bridge: telegram.0 " prefix=router
time="2020-12-07T15:20:55+01:00" level=info msg=Connecting prefix=telegram
time="2020-12-07T15:20:55+01:00" level=info msg="Connection succeeded" prefix=telegram
time="2020-12-07T15:20:55+01:00" level=info msg="telegram.0: joining  (ID: telegram.0)" prefix=telegram
time="2020-12-07T15:20:55+01:00" level=info msg="Starting bridge: nctalk.1 " prefix=router
time="2020-12-07T15:20:55+01:00" level=info msg=Connecting prefix=nctalk
[...]

PR #4745, opened 2020-12-11
Fix MsTeams and Steam Matterbridge config

eneiluj:
As mentioned in #4719, `channel` is the key used for all services in `gateway` section.

MsTeams is now finally working.


nickvergessen (approved):

nickvergessen:
I guess we backport this?

Do existing configs have to be adjusted somehow?

eneiluj:
Yes we can backport. Configs are generated when bridge is saved which happens when it's toggled. No special adjustments needed.

PR #4746, opened 2020-12-11
Use proc_open to run system commands in bridge manager

eneiluj:
This clarifies how we run commands and launch Matterbridge binary. It silences the processes. Output is now parsed with Php instead of using pipes, grep, awk and wc (which might all be missing on some systems).

refs  #4743


nickvergessen (approved):
🐘 Christoph said it should be fine

PhrozenByte:
:heart:

PR #4757, opened 2020-12-14
Use a unique index name

nickvergessen:
Regression from #4735 

Only breaks on Oracle and SQLite, where index names have to be database-wide unique

Extracted from #4589 to make sure @jancborchardt can update on sqlite
Not adding a migration as it has not been released.


jancborchardt (approved):
Fixes the upgrade issue for me, thanks! :)

PVince81 (approved):
:+1:

PR #4762, opened 2020-12-15
Allow to respond with modified common read value on "not modified" messages

nickvergessen:
We need to set the status code to 200 when the header is modified,
because as per "section 10.3.5 of RFC 2616" entity headers shall be
stripped out on 304: https://stackoverflow.com/a/17822709
This means the header would not be sent to the clients at all, so
they wouldn't update the read marker when idling in the chat.
For the web interface this is not needed, as we update the info
every 30 seconds with the conversation list refresh and can therefor
save the query


tobiasKaminsky (approved):

PVince81 (approved):
makes sense :+1:

PR #4768, opened 2020-12-16
Move PHPUnit to GitHub actions 🚛

nickvergessen:


PVince81 (approved):
Looks fine, I only quickly scrolled to it.
Assuming that all steps still exist in the new config (I saw oci there)

:+1:

Issue #4523, opened 2020-11-03
Page asking for password of shared file contains "Request password" button


mknet:
## Steps to reproduce
1. Create a password protected file share link
2. Call the respective URL
3. In addition to the form for entering the password for the sharing link, a button with the value "Restore password" is rendered.

### Expected behaviour
I'd assume that this button shouldn't occur here. It's missleading since as I user I would press this button to send my entered password belonging to the share link

### Actual behaviour
The button appears and IMHO misleads the user

## Talk app

**Talk app version:** 10.0.1

If I deactivate the Talk app, the button is not rendered anymore.

nickvergessen:
The button does not display here. Can you post a screenshot and a list of installed apps?

mknet:
Thank you, @nickvergessen 
Here is the particular markup which is rendered in the DOM

`<div data-v-1ffb88f8=""><div data-v-1ffb88f8="" id="submit-wrapper" class="request-password-wrapper"><input data-v-1ffb88f8="" id="request-password-button" type="button" value="Passwort anfordern" class="primary"> <div data-v-1ffb88f8="" class="icon icon-confirm-white"></div></div> <!----></div>`

and screenshots showing the button and the list of installed apps.

![2020-11-04 14 24 51 a4532b700c27](https://user-images.githubusercontent.com/396667/98117407-f8f45300-1ea9-11eb-8fa1-87efd97bbc10.png)

![2020-11-04 14 32 32  15c597700ec8](https://user-images.githubusercontent.com/396667/98118114-02ca8600-1eab-11eb-9d2d-86bfbff0e538.png)

If it's an issue the screenshots are in German, drop me a line / comment.

Thanks

nickvergessen:
@danxuliu with some recent changes the button seems to appear now all the time and not only when video verification is on.
Can you have a look?

PVince81:
confirmed. when clicking the button an error occurs.

I'll have a look.

PVince81:
the button also appears in 10.0.3, so if it's a regression, likely happening even earlier

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/4769

PR #4769, opened 2020-12-16
Remove request password button when verif disabled

PVince81:
Whenever the video verification has not been set for a share, do not
display the button "Request password" in the public link page.

Somehow it looks like there was no test anywhere before, maybe it was always missing ?

Fixes https://github.com/nextcloud/spreed/issues/4523

- [x] TEST: no button when video verif was not ticked
- [x] TEST: button appears when video verif is ticked
- [x] TEST: verified that the talk sidebar still appears for link file share when video verif is off


nickvergessen (approved):
🙈

PR #4771, opened 2020-12-16
Updated wording for conversation visibility

PVince81:
As per feedback in chat from @jancborchardt 

OLD
image

NEW
image

Fixes https://github.com/nextcloud/spreed/issues/4773

### TODOs:

- [x] wording changes
- [x] detect guest app


PVince81:
Updated to match the proposal from https://github.com/nextcloud/spreed/issues/4773

I still need to find out how to detect the guest app, hopefully without having to change it...
Ideally the guest app should publish a capability which it doesn't currently.

PVince81:
Here we go. I've exposed the guest app enabled state through initial state, similar to how it was done for the circles app.

Please review

ma12-co (commented):
I still think that "make conversation accessible to registered users" is not clear, especially while in the process of creating a group... 
Possible undesired train of thought: "registered to what?" to talk? to this conversation? That is weird, I haven't added anybody yet.. Oh, wait.. registered to nextcloud? This nextcloud or all nextclouds? 

Possible improvements:
 - Make conversation accessible to all registered users (except guests);
 - Make conversation accessible to users registered in this instance (except guests);

PVince81:
I'm not a fan of "accessible" because even unlisted conversations are accessible if a public link is allowed.
Also the texts are a bit too long.

PVince81:
another proposal:

```
Appears in search results for:
- participants only (or "never"?)
- users with an account on this instance
- guest users with an account on this instance
```

then for the system messages it would become:
- "You made this conversation appear in search results for participants only"
- "You made this conversation appear in search results for users with an account on this instance"
- "You made this conversation appear in search results for guest users with an account on this instance"

a bit long thought, but when appearing in a system message it might be ok as people usually won't read it, it's only for future reference

thoughts ? @jancborchardt @ma12-co

jancborchardt:
Let’s just go with one choice for now, cause either is an improvement on the current wording. :) 

I would say either the existing one with

> Make conversation accessible to registered users

or

> Make conversation findable for registered users

@PVince81 just pick one and let’s go with it and get it merged – let’s improve step by step and not have a small wording question block an improvement for so long. :)

nickvergessen:
Well i'd like to avoid tranlsators to have to rediscuss this on a weekly basis too...

jancborchardt:
Considering that at some stage these conversations will not only be accessible via search but also possibly via the "+" menu (or on the bottom of the list, or wherever we put them, "findable" is not a futureproof wording, so we should for sure go for:

> Make conversation accessible to registered users

lib/Chat/Parser/SystemMessage.php
```diff
@@ -166,25 +166,25 @@ public function parseMessage(Message $chatMessage): void {
...
 			if ($currentUserIsActor) {
-				$parsedMessage = $this->l->t('You made the conversation visible for registered users only');
+				$parsedMessage = $this->l->t('You made the conversation accessible to registered users only');
 			} elseif ($cliIsActor) {
-				$parsedMessage = $this->l->t('An administrator made the visible for registered users only');
+				$parsedMessage = $this->l->t('An administrator made the accessible to registered users only');
 			}
 		} elseif ($message === 'listable_all') {
-			$parsedMessage = $this->l->t('{actor} made the conversation visible for everyone');
+			$parsedMessage = $this->l->t('{actor} made the conversation accessible to everyone');
```
jancborchardt, line 183: "everyone" is still confusing wording.
```suggestion
			$parsedMessage = $this->l->t('{actor} made the conversation accessible to registered users and guest app users');
```

Otherwise great! :) :+1:

lib/Chat/Parser/SystemMessage.php
```diff
@@ -166,25 +166,25 @@ public function parseMessage(Message $chatMessage): void {
 				$parsedMessage = $this->l->t('An administrator locked the conversation');
 			}
 		} elseif ($message === 'listable_none') {
-			$parsedMessage = $this->l->t('{actor} made the conversation invisible');
+			$parsedMessage = $this->l->t('{actor} made the conversation accessible to participants');
```
nickvergessen, line 169: but it could still be public :/ "freely accessible" maybe?

src/components/ConversationSettings/ListableSettings.vue
```diff
@@ -97,30 +100,50 @@ export default {
...
+					this.lastNotification.hideToast()
+					this.lastNotification = null
+				}
+				if (listable === CONVERSATION.LISTABLE.NONE) {
+					this.lastNotification = showSuccess(t('spreed', 'You made the conversation accessible to participants'))
+				} else if (listable === CONVERSATION.LISTABLE.USERS) {
+					this.lastNotification = showSuccess(t('spreed', 'You made the conversation accessible to registered users only'))
+				} else if (listable === CONVERSATION.LISTABLE.ALL) {
+					this.lastNotification = showSuccess(t('spreed', 'You made the conversation accessible to everyone'))
 				}
```
nickvergessen, line 142: there is a system message in the chat. shouldn't really spam with a success notification on top right helping people to get blind to notifications

nickvergessen:
How about just calling it "Open conversations"

src/components/ConversationSettings/ListableSettings.vue
PVince81: Only applies if you actually see the system message.
In the case of new conversation creation, there should be no such system message shown currently.

When using the settings dialog, if you change the dropdown it will save directly without closing the dialog, but you don't necessarily see the chat messages from there.

In general I think it's a good idea to have some kind of visual feedback for autosave fields. Here this is achieved with a notification. If we can come up with a different approach like have tick marks or so (system wide, as design guideline), it might work better. @jancborchardt thoughts ?

nickvergessen: The visual feedback should be the chat, or the checkmark goes to spinner and only turns into a checkmark/empty once the response is there. Otherwise you see 5 popup notifications when you organize a meeting and it's just making you blind.

PVince81: checkmark going to spinner is an idea, with a minimum delay in case it's too fast, so it looks like it's saving (I recently read about UI improvements where delays were added to give the user the impression that it's properly computing/saving)

nickvergessen:
Updated the wording to the final discussion results:
* everyone => registered and guest app users
* accessible => open
* "accessible to participants only" => limited to participants

nickvergessen (approved):

Issue #4754, opened 2020-12-14
Disable picture in picture for own local video


PVince81:
## Steps to reproduce
1. On Talk master
2. Start a call with another participant
3. Hover on own local video

### Expected behaviour
No picture in picture offer as it doesn't make much sense to watch self. (makes more sense for the presenter video)

### Actual behaviour
Mirrored picture in picture button
<img width="265" alt="image" src="https://user-images.githubusercontent.com/277525/102074465-21eef880-3e05-11eb-90d0-a20665085045.png">

## Versions

**Talk app version:** git master

**Browser:** Firefox 83.0

I heard discussions somewhere about the mirroring. Until this is resolved I suggest to try disabling this completly, if possible.

nickvergessen:
Well from my pov we should disable all PiP apart from the the promoted windows, see #4674

PVince81:
let me try this... the mirror thing looks too ugly

PVince81:
https://www.w3.org/TR/picture-in-picture/#disable-pip

PVince81:
doesn't seem to work, maybe Firefox (tested with 83.0) is not up to date yet with that spec:
<img width="659" alt="image" src="https://user-images.githubusercontent.com/277525/102339728-e20b4b00-3f95-11eb-97fd-1c35b601246b.png">

and a closed ticket from Mozilla as won't fix https://bugzilla.mozilla.org/show_bug.cgi?id=1611831 :open_mouth: 

let me check chromium...

PVince81:
in Chromium 87.0.4280.88 the picture in picture option is more hidden, need to right click on the video

when setting the attribute above the menu entry appears grayed out

so might still be worth adding and hope Firefox will reconsider adding it later...

but: this would not fix the flipped overlay issue which is the most annoying thing

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/4772

PR #4772, opened 2020-12-16
Disable PIP for non-promoted videos

PVince81:
Disabled in grid view, own video and also the media settings.

Fixes https://github.com/nextcloud/spreed/issues/4754

Tested in Chromium 87.
You need to right click to see the "Picture in Picture" option.
Note that in grid view for some reason right click doesn't show the same menu, so I used the inspector to verify.
In promoted view the video context menu works and allows picture in picture mode.

Doesn't work in Firefox because of https://bugzilla.mozilla.org/show_bug.cgi?id=1611831

### DIFF
--- a/src/components/CallView/shared/Screen.vue
+++ b/src/components/CallView/shared/Screen.vue
@@ -22,6 +22,7 @@
 	<div :id="screenContainerId" class="screenContainer">
 		<video v-show="(localMediaModel && localMediaModel.attributes.localScreen) || (callParticipantModel && callParticipantModel.attributes.screen)"
 			ref="screen"
+			:disablePictureInPicture="!isBig"
 			class="screen"
 			:class="screenClass" />
 		<VideoBottomBar v-if="isBig"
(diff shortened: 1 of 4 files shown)


src/components/CallView/shared/Screen.vue
```diff
@@ -22,6 +22,7 @@
 	<div :id="screenContainerId" class="screenContainer">
 		<video v-show="(localMediaModel && localMediaModel.attributes.localScreen) || (callParticipantModel && callParticipantModel.attributes.screen)"
 			ref="screen"
+			:disablePictureInPicture="!isBig"
```
nickvergessen, line 25: So here you insert a boolean while in `src/components/CallView/shared/LocalVideo.vue` its a string?

PVince81: if I remove the ":" then the rendered result will be `<video ... disablePictureInPicture="!isBig"` so I have to keep it so the variable gets evaluated

in LocalVideo I didn't add the ":" because `<video>` is not a Vue component so "true" ends up rendered as string anyway

nickvergessen: then
```suggestion
			:disablePictureInPicture="!isBig ? 'true' : 'false'"
```
?

PVince81: your proposal only makes sense if isBig can be null or undefined in which case it would become the string "null" or "undefined".

anyway, better safe than sorry, I take it

nickvergessen (approved):

PR #4778, opened 2020-12-17
WIP debug peerId matching

PVince81:
I have two users, admin and Bob.

When bob connects, the admin browser receives this peer (from refreshPeerList):
image

However, the signaling message comes from a different peer id:
image

which matches the one from the callParticipantModel which we use for setting the raised hand state:
image

### DIFF
--- a/src/components/CallView/CallView.vue
+++ b/src/components/CallView/CallView.vue
@@ -495,6 +495,7 @@ export default {
 			}
 
 			// update in callViewStore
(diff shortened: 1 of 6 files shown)


PVince81:
how to match the peerId from the localParticipantModel to the one from that refresh list ? (this is with HPB enabled as per ticket)

or should we simply use actorId as here it's only about hand raised ?

@nickvergessen @danxuliu

PVince81:
we have `nextcloudSessionId` now, closing

PR #4741, opened 2020-12-10
Message status ux improvements

ma12-co:
fix #4738 

Peek 2020-12-10 17-16

Signed-off-by: Marco Ambrosini <marcoambrosini@pm.me>

PR #4779, opened 2020-12-17
Feature/noid/center system messages

ma12-co:
As discussed some time ago with @jancborchardt, let's center align the text for system message and add some padding to the sides, in order to create a visual separation between real chat messages and system ones. 

based on #4741 

Before | After
---|---
Peek 2020-12-17 13-11 | Peek 2020-12-17 13-07

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -463,6 +463,8 @@ export default {
 
 			&.system-message {
 				color: var(--color-text-maxcontrast);
+				text-align: center;
+				padding: 0 20px;
 			}
 
 			::v-deep .rich-text--wrapper {
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -465,6 +465,7 @@ export default {
 				color: var(--color-text-maxcontrast);
 				text-align: center;
 				padding: 0 20px;
+				width: 100%;
 			}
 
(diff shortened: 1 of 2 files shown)


nickvergessen:
makes the date separator basically invisible

ma12-co:
> makes the date separator basically invisible

we can tackle that

ma12-co:
Fixed some alignment issues and reintroduced the gray chips for dates, albeit without the bold font so that they don't look like buttons.

Screenshot from 2020-12-17 14-59-29

jancborchardt:
Only change I'd make is instead of having a background for the date divider (which makes it look too much like it's interactive), there should still be the horizontal line.

It should still have a background of color-main-background and look pill-style, so that when it's sticky on scrolling it can properly overlap the chat.

(This is also what we discussed before.)

ma12-co:
Yes Jan. Already discussed but as you can see there are new issues being raised. I did this in a coworking session with Joas to address his concern with the current date separators being too similar to centered system messages.
The style of this element is pretty far from any other button we have. If anything, it might resemble a disabled button so I don't think that the concern about the clickability of it is still grounded. Don't you think?

ma12-co:
What about this? This is half the size of our buttons 
Screenshot from 2020-12-17 15-30-24

nickvergessen:
I also think it's far enough to be confused with a button (also there is no cursor change).
It's the same with all other messengers

jancborchardt:
Sure, let’s just go with it.

nickvergessen:
Messages overflow the length of the window with long displaynames (as they don't wrap) and then pushes off the timestamp

nickvergessen (approved):

Issue #4776, opened 2020-12-17
Change location or access of config file


KaitoKuroba:
While testing I realized that the .toml config file for matterbridge is created under /tmp. It also seems to be that the default access rights of the web application user were used. Since the config file contains sensible data this seems like a security problem for me.

Wouldn't it be better to put the .toml and .log file somewhere under the data folder of the nextcloud installation? This folder should have better access rights in most environments. Alternatively the matterbridge app should at least make shure that the access rights to the files are as small as possible.

KaitoKuroba:
A change of the location should also solve this issue: #4549

eneiluj:
Thanks for the suggestion. It should indeed be done.

I started to do it in #4782.

alaincao:
May I ask to increase the weight of this issue? Because my understanding is that it's a bug:  
After a restart of the computer, the `/tmp` directory is cleaned (on Debian, at least).  
So the configuration does not seem to survive a server reboot.  
```
time="2021-01-26T12:03:00Z" level=fatal msg="Failed to read configuration file: &os.PathError{Op:"open", Path:"/tmp/bridge-cs3sx4on.toml", Err:0x2}" prefix=config
```
This is particularily useful when used in a Docker when an upgrade of the container completely wipes the filesystem.  
A configuration option would be incredibly useful.

eneiluj:
@alaincao This should be fixed by #5040 which checks if the config file exists before launching the bridge and creates it again if it's missing. Thanks for reporting the problem.

alaincao:
Ok, thanks a bunch!

Issue #4549, opened 2020-11-06
Talk_Matterbridge not able to start/write the log file in /tmp 


poltrian:
@eneiluj

## Steps to reproduce
1. open Talk
2. Add Bridge
3. Add your Data
4.Press "Save

### Expected behaviour
nothing happend, will not saved

### Actual behaviour
nothing

## Talk app

**Talk app version:** 10.01 on Nextcloud 20.0.1

**Custom Signaling server configured:** no 

**Custom TURN server configured:** yes/

**Custom STUN server configured:** yes/

eneiluj:
Could you give more details?

So you define some bridged channels, then press save and nothing happens, right?
* Is there any error in browser console?
* Could you check the network tab when saving? What's the response for the "api/v1/bridge/xxxxx" request?
* Is there a "/tmp" on your server's system? Is it writeable by your webserver's user?

I remember you gave more information when we talked in the public Talk room...but don't remember what :grin:.

poltrian:
Hi,

Network IRC Freenode
await fetch("https://xxxxxx.de/ocs/v2.php/apps/spreed/api/v1/chat/ffsavsaz?lookIntoFuture=1&lastKnownMessageId=3698&includeLastKnown=0", {
    "credentials": "include",
    "headers": {
        "User-Agent": "Mozilla/5.0 (X11; Linux x86_64; rv:82.0) Gecko/20100101 Firefox/82.0",
        "Accept": "application/json, text/plain, */*",
        "Accept-Language": "de,en-US;q=0.7,en;q=0.3",
        "requesttoken": "4ld1gw4nIHgHgMHaj2ZAx1jM/mZ3KBqPfD8uE42uK3U=:yRE2yEYTbDc/6/iYuAsriw34yStCGEzXTQh+dMPtXAY="
    },
    "method": "GET",
    "mode": "cors"
});
Error is 404
Konsole:
OC.getCapabilities is deprecated and will be removed in Nextcloud 21. See @nextcloud/capabilities 3 capabilities.js:32:9
$ is deprecated: The global jQuery is deprecated. It will be updated to v3.x in Nextcloud 21. In later versions of Nextcloud it might be removed completely. Please ship your own. globals.js:60:15
jQuery is deprecated: The global jQuery is deprecated. It will be updated to v3.x in Nextcloud 21. In later versions of Nextcloud it might be removed completely. Please ship your own. 3 globals.js:60:15
Setting current user App.vue:329
[...]

PR #4782, opened 2020-12-17
Adjust matterbridge config file permissions

eneiluj:
This sets 0600 permissions (read and write for owner only) to Matterbridge config files each time they are written.

refs #4776 #4549

@nickvergessen Would you have a hint about a good place to write the config files instead of `/tmp`? We need to know the absolute path to give it to the binary then. I guess the question is: How can we get an absolute path, somewhere in Nextcloud, where it's ok to write the config files?

### DIFF
--- a/lib/MatterbridgeManager.php
+++ b/lib/MatterbridgeManager.php
@@ -253,6 +253,7 @@ class MatterbridgeManager {
 		$configPath = sprintf('/tmp/bridge-%s.toml', $room->getToken());
 		$configContent = $this->generateConfig($newBridge);
 		file_put_contents($configPath, $configContent);
+		chmod($configPath, 0600);
 	}
 
 	/**


eneiluj:
@nickvergessen I rebased this on master. I think it can be merged and I'll take care of the config location in another PR.

nickvergessen:
> Would you have a hint about a good place to write the config files instead of /tmp? We need to know the absolute path to give it to the binary then. I guess the question is: How can we get an absolute path, somewhere in Nextcloud, where it's ok to write the config files?

So the best place would be data directory. That is garanteed writable, etc.
The problem is with "S3 primary storage" it's not local. so then you would need to use 

https://github.com/nextcloud/server/blob/9ce3ea3368937f827171d1ee3aa6a202d93e6cd2/lib/public/Files/Storage/IStorage.php#L318-L326 the problem is this is not available on the Node API and therefor not available on IAppData as far as I can see. Maybe you can ask Roeland to confirm and then request adding it for 22

PR #4815, opened 2020-12-20
Update .gitignore

yahsashay:
changed the first character of every word in the comment upper case.


nickvergessen:
Actually we only uppercase the first word everywhere at Nextcloud

Issue #4818, opened 2020-12-21
exports is not defined on master


PVince81:
## Steps to reproduce
1. On master 7d7999ec28472e34edc316d87439ad902a921b1e
2. `make dev-setup && make build-js`
3. Open the web UI

### Expected behaviour
Web UI loads

### Actual behaviour
<img width="627" alt="image" src="https://user-images.githubusercontent.com/277525/102754711-a98cb800-436d-11eb-9b6d-5f99c035dc75.png">

### Notes
Caused by https://github.com/nextcloud/spreed/pull/4811

We have no way to prevent dependabot to trigger this patch update.
Doesn't make sense that a patch update breaks compilation so suddenly ?

There are some ideas that it might be related to the "modules" babel option which isn't set in talk.
Also note that other repos like nextcloud-event-bus also haven't had that update yet, so there's a risk that breakage will happen there as well soon.

PVince81:
```
talk@11.0.0 /srv/www/htdocs/server/apps/spreed
├── @babel/preset-env@7.12.11 
└─┬ @nextcloud/axios@1.5.0
  └── @babel/preset-env@7.12.11  deduped
```

with 7.12.10 it works fine, so the problem is caused by changes specifically in 7.2.11

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/4819

PVince81:
as mentioned in the PR, the change will break with IE 11 so not possible to use it here.

as suggested in https://github.com/nextcloud/spreed/pull/4819#issuecomment-748881555 the next part is to:
- [ ] revert the update
- [ ] tell dependabot to not update that lib any more until further notice (to get master unblocked)
- [ ] continue investigating the root cause

PVince81:
but first, there are other issues already on master with IE 11: https://github.com/nextcloud/spreed/issues/4820

nickvergessen:
https://github.com/nextcloud/notifications/pull/825#pullrequestreview-561665184 the problem is fixed by removing the `babel-loader-exclude-node-modules-except`
So when IE11 is dropped we can update. The origin is https://github.com/babel/babel/pull/10862

PVince81:
okay, so should we move the ticket to backlog / 22 so we just discard the workaround at that time ?

we already added the "upgrade exclusion" to avoid further regressions.

PVince81:
this is a diff of the generated talk.js, on the left with present-env 7.12.10 and on the right with 7.12.11:
```
[...]

PR #4819, opened 2020-12-21
Fix exports issue related to nextcloud-event-bus

PVince81:
- [x] check that it still runs in Edge

This is a partial revert of https://github.com/nextcloud/spreed/pull/3961 so better not backport to versions that need to support IE11.

Fixes https://github.com/nextcloud/spreed/issues/4818

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -59,7 +59,6 @@ module.exports = {
 					'hot-patcher',
 					'nextcloud-vue-collections',
 					'semver',
-					'@nextcloud/event-bus',
 					'webdav',
 				]),
 			},


PVince81:
Tested with Edge, the page loads fine.

nickvergessen:
> This is a partial revert of #3961 so better not backport to versions that need to support IE11.

The last IE version is Nextcloud 22 IIRC?

PVince81:
> > This is a partial revert of #3961 so better not backport to versions that need to support IE11.
> 
> The last IE version is Nextcloud 22 IIRC?

I found this (still open) ticket on server: https://github.com/nextcloud/server/issues/20855
As far as I understand:
- IE11 deprecated in NC 20, Edge still supported
- Edge still supported in NC 21, IE 11 not any more (which fits with my statement in this PR)
- Edge also dropped in NC 22

PVince81:
we might not need to backport if we don't upgrade babel-env-preset in stable vesions

nickvergessen:
No, the ticket is IE11 will be deprecated but should be mostly functional until 22.
Edge is Chrome based and updates automatically, so not a problem at all.
But a blank page in talk is not "deprecated".

Another fix idea would be to downgrade the problematic lib again and add it to the ignore list of dependabot

PVince81:
okay, so will retest with IE 11 to see if, by any chance, the old problem magically disappeared:
- [ ] retest with IE 11

ChristophWurst (approved):

PVince81:
with IE11 it does NOT work, an error about the event bus appear:
ie11-with-talk-pr

as we can see it did not get transpiled for IE 11

maybe the transpiling should be done in the repo itself ?

also, it seems talk alpha 2 build is broken as well with IE 11 due to nextcloud-vue, raised here: https://github.com/nextcloud/spreed/issues/4820

PVince81:
revert PR is here https://github.com/nextcloud/spreed/pull/4821

PVince81:
right, so the problem is that semver is not transpiled: https://github.com/nextcloud/nextcloud-event-bus/issues/145

PR #4829, opened 2020-12-22
Transpile more deps for IE 11

PVince81:
In the listed order, every time I added a new exclude, a new one popped up. So I added them one by one.
Now in this PR we've reached a state where the Talk page loads in IE 11 but is missing some controls, likely due to the many JS errors happening in nextcloud-vue components (but no more syntax errors).

ie11-many-excludes

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -21,7 +21,3 @@ updates:
     versions:
     - ">= 4.a"
     - "< 5"
-  # remove after fixing https://github.com/nextcloud/spreed/issues/4818
-  - dependency-name: "@babel/preset-env"
-    versions:
-    - ">= 7.12.11"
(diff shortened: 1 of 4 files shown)


PVince81:
the last error seems to say that `VPopover` expects a string instead of a number, and we do pass in a number. We pass it indirectly to `Popover` which then copies it over to `VPopover`. Not sure why the property requirements would be different/buggy with IE 11...

PVince81:
in components/Popover.vue of the lib it **does** accept a number:
```
    delay: {
      type: [String, Number, Object],
      default: () => getDefault('defaultDelay'),
    },
```

nickvergessen:
> Conflicting files
package-lock.json

nickvergessen:
Allows chatting in IE11 again

PVince81:
let me solve the conflict.

indeed the chat works for me, despite the million of errors in the console about data type validation in Vue properties...

nickvergessen (approved):
Fixes IE11 loading again

nickvergessen:
Merge?

Issue #4833, opened 2020-12-22
Don't scroll conversation list that often


PVince81:
## Steps to reproduce
1. Have a lot of conversations in the left sidebar
2. Join a conversation at the bottom of the list
3. Post a message or click on a conversation in the left sidebar

### Expected behaviour
No scrolling when joining and none when posting or receiving a message, nor when clicking on a conversation.

### Actual behaviour
Scrolling happens when joining and none when posting or receiving a message, nor when clicking on a conversation.

## Talk app

**Talk app version:** v11.0.0-alpha.2

## Notes

As discussed, I'll remove the scrolling for these two events. @jancborchardt @ChristophWurst @ma12-co 

However, scrolling will then happen when joining a newly created conversation (when picking a user from the search) or when joining a listed conversation.

PVince81:
Now I'm unsure why there was a "scroll on message post" in the first place.
I suspect it might be because sometimes the conversation might move in the left sidebar if new conversations appear, for example when you got invited. But I guess this is a corner case anyway so fine to not scroll in such cases.

PVince81:
follow up with missing bits to cleanup: https://github.com/nextcloud/spreed/pull/4843

PR #4836, opened 2020-12-22
Scroll to conversation only in specific cases

PVince81:
Fixes https://github.com/nextcloud/spreed/issues/4833

Scroll to conversation only in specific cases:
- scroll into view when joining a conversation from the search results
- scroll into view initially when loading the page
- don't scroll on click
- don't scroll when posting or receiving a message

To test this, make sure to have a lot of favorite channels at the top, then join a channel that would appear at the bottom.

I already lost some time trying to make the scrolling itself work, so now I've left a timeout in there. Not sure what the left sidebar list is doing that prevents us to scroll it earlier even when the element already exists. To be researched later separately...


ma12-co (approved):

PVince81:
I missed some bits, now there are errors in console, fixed here: https://github.com/nextcloud/spreed/pull/4843

Issue #4770, opened 2020-12-16
Sort participant list by raised hand timestamp


PVince81:
**Is your feature request related to a problem? Please describe.**
Find out who raised the hand first in a very big call where many people raised their hands.

**Describe the solution you'd like**
People who raised their hand appear at the top of the participant list, sorted by timestamp: whoever raised the hand first appears on top, etc.

Follow up of https://github.com/nextcloud/spreed/pull/4569

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/4838

PR #4838, opened 2020-12-23
Sort participants by raised hands

PVince81:
Fixes https://github.com/nextcloud/spreed/issues/4770

Participants who raised their hand will appear on top of the list.
For this, the "raisedHand" state was exended to also include a timestamp. The timestamp is sent by whoever raised the hand to have consistent values across participants.

Tested with regular users and also guest users without accounts.

Please use internal signaling when testing because with HPB it doesn't work because of https://github.com/nextcloud/spreed/issues/4777

### DIFF
--- a/src/utils/webrtc/models/LocalMediaModel.js
+++ b/src/utils/webrtc/models/LocalMediaModel.js
@@ -228,7 +228,7 @@ LocalMediaModel.prototype = {
 			this.set('videoAvailable', false)
 		}
 
-		this.set('raisedHand', false)
+		this.set('raisedHand', { state: false, timestamp: Date.now() })
 	},
 
 	_handleLocalStreamChanged: function(localStream) {
@@ -441,11 +441,16 @@ LocalMediaModel.prototype = {
 			throw new Error('WebRtc not initialized yet')
 		}
 
-		this._webRtc.sendToAll('raiseHand', { raised: raised })
+		const raisedHand = {
+			state: raised,
+			timestamp: Date.now(),
+		}
+
+		this._webRtc.sendToAll('raiseHand', raisedHand)
 
 		// Set state locally too, as even when sending to all the sender will not
 		// receive the message.
-		this.set('raisedHand', raised)
+		this.set('raisedHand', raisedHand)
 	},
 
 }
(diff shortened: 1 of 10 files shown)


ma12-co (approved):
:elephant:

nickvergessen:
We could send the session id manually:

```diff
diff --git a/src/components/CallView/CallView.vue b/src/components/CallView/CallView.vue
index 546d3d2ef..128a009ca 100644
--- a/src/components/CallView/CallView.vue
+++ b/src/components/CallView/CallView.vue
@@ -500,8 +503,7 @@ export default {
 
                        // update in callViewStore
                        this.$store.dispatch('setParticipantHandRaised', {
-                               // FIXME: when using HPB sessionId doesn't match
-                               sessionId: callParticipantModel.attributes.peerId,
+                               sessionId: raisedHand?.sessionId ? raisedHand?.sessionId : callParticipantModel.attributes.peerId,
                                raisedHand: raisedHand,
                        })
                },
diff --git a/src/utils/webrtc/models/LocalMediaModel.js b/src/utils/webrtc/models/LocalMediaModel.js
index b17b579a2..4ddb185d6 100644
--- a/src/utils/webrtc/models/LocalMediaModel.js
+++ b/src/utils/webrtc/models/LocalMediaModel.js
@@ -444,6 +444,7 @@ LocalMediaModel.prototype = {
                const raisedHand = {
                        state: raised,
                        timestamp: Date.now(),
+                       sessionId: this._webRtc.connection?.nextcloudSessionId,
                }
 
                this._webRtc.sendToAll('raiseHand', raisedHand)

```
Works for sending, but if a user leaves and rejoins their hand is still up, as the removing of a user is only removing by peer id.
Maybe we just always send the session id? cc @danxuliu

src/utils/webrtc/models/LocalMediaModel.js
```diff
@@ -444,6 +444,7 @@ LocalMediaModel.prototype = {
 		const raisedHand = {
 			state: raised,
 			timestamp: Date.now(),
+			sessionId: this._webRtc.connection?.nextcloudSessionId,
```
PVince81, line 447: nice one! I didn't look deep enough

Issue #4834, opened 2020-12-22
Add Conversation API version 3 in capabilities when available


Ivansss:
The last one added to capabilities was conversation-v2 but conversation-v3 was never added.

PVince81:
Seems this was added as description for the "sip-support" capability. It was probably from the time where v3 was assumed to be only for SIP support.

Meanwhile we also added more stuff in v3 like listable rooms.

So to me it makes sense to add another one. I'll send a PR.

PVince81:
I've raised https://github.com/nextcloud/spreed/pull/4839 to add the "conversations-v3" capability.

However, I think clients should anyway rely on the specific feature capability.
Probing for "conversations-v3" doesn't guarantee that features like "listable-rooms" and "chat-read-status" exist.

PR #4839, opened 2020-12-23
Add conversation-v3 capability for API v3

PVince81:
Fixes https://github.com/nextcloud/spreed/issues/4834

Added capability to signal the existence of the v3 API.

**NOTES**:
- I changed the docs to say Talk 10.1 for the "sip-support" capability and "conversation-v3" because we only backported these to stable20.1, not stable20.
- the existence of "conversation-v3" doesn't mean that all the other features in v3 are available but only that this endpoint version exists. For specific features, these need to check their corresponding capabilities (ex: "listable-rooms", "chat-read-status")


Ivansss (approved):
LGTM 👍

nickvergessen (approved):

PR #4855, opened 2020-12-28
Private reply without quote

jakobroehrl:
solves: #2800
image

works only in group chats, not for guest and my own messages
continuation of https://github.com/nextcloud/spreed/pull/4477

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -122,6 +122,13 @@ the main body of the message as well as a quote.
 						@click.stop="handleReply">
 						{{ t('spreed', 'Reply') }}
 					</ActionButton>
+					<ActionButton
+						v-if="isPrivateReplyable"
+						icon="icon-user"
+						:close-after-click="true"
+						@click.stop="handlePrivateReply">
+						{{ t('spreed', 'Reply private') }}
+					</ActionButton>
 					<template
 						v-for="action in messageActions">
 						<ActionButton
@@ -132,13 +139,6 @@ the main body of the message as well as a quote.
 							{{ action.label }}
 						</ActionButton>
 					</template>
-					<ActionButton
-						v-if="isPrivateReplyable"
-						icon="icon-user"
-						:close-after-click="true"
-						@click.stop="handlePrivateReply">
-						{{ t('spreed', 'Reply private') }}
-					</ActionButton>
 					<ActionButton
 						v-if="isDeleteable"
 						icon="icon-delete"
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -170,7 +170,7 @@ import Quote from '../../../Quote'
 import participant from '../../../../mixins/participant'
 import { EventBus } from '../../../../services/EventBus'
 import emojiRegex from 'emoji-regex'
-import { PARTICIPANT, CONVERSATION } from '../../../../constants'
+import { PARTICIPANT, CONVERSATION, ATTENDEE } from '../../../../constants'
 import moment from '@nextcloud/moment'
 import {
 	showError,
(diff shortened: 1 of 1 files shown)


lib/Model/Message.php
```diff
@@ -161,6 +161,16 @@ public function isReplyable(): bool {
...
 	}
 
+	/**
+	 * Specifies whether a message can be replied private to
+	 */
+	public function isPrivateReplyable(): bool {
+		return $this->getMessageType() !== 'system' &&
+			$this->getMessageType() !== 'command' &&
+			$this->getRoom()->getType() !== Room::ONE_TO_ONE_CALL &&
+			\in_array($this->getActorType(), [Attendee::ACTOR_USERS, Attendee::ACTOR_GUESTS]);
```
nickvergessen, line 171: Should only work for `Attendee::ACTOR_USERS` and also only when the current user is `Attendee::ACTOR_USERS` because otherwise they can't have a private chat at all.

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -410,7 +426,7 @@ export default {
 			// again another time
 			this.$refs.message.classList.remove('highlight-animation')
 		},
-		handleReply() {
+		handleReply(t) {
```
nickvergessen, line 429:
```suggestion
		handleReply(token) {
```

t is for translations, please don't overwrite it

nickvergessen:
> we actually need to find a way, because atm private replies don't display in the messages list

As said before, this will not work and I will not make it work. We will not be able to show messages of another chat 1 in the private chat 2. Checking if users have/retain access and so on is way to difficult. That's why I posted it in https://github.com/nextcloud/spreed/pull/4477#issuecomment-717810331

If you reply, we need to copy it somehow into the chat input, either with:
```
> Original message

Private reply
```

or
```
Original author: "Original message"

Private reply
```

or something.

lib/Model/Message.php
nickvergessen: also same as with delete, I would do this check in the UI and not output it as a flag.

szaimen:
``` 
> Original message

Private reply
```

Sound good to me. This is e.g. how Whatsapp handles this. Of course it would need to display the whole message in quotes in this case and not a shortened one.

jakobroehrl:
@nickvergessen 
I think this looks good, could you look over it?

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -132,6 +132,13 @@ the main body of the message as well as a quote.
 							{{ action.label }}
 						</ActionButton>
 					</template>
+					<ActionButton
+						v-if="isPrivateReplyable"
+						icon="icon-user"
+						:close-after-click="true"
+						@click.stop="handlePrivateReply">
+						{{ t('spreed', 'Reply private') }}
```
nickvergessen, line 140:
```suggestion
						{{ t('spreed', 'Reply privately') }}
```

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -494,6 +502,12 @@ export default {
 					|| this.isMyMsg)
 		},
 
+		isPrivateReplyable() {
+			return (this.conversation.type === 3
+			&& !this.isMyMsg
+			&& this.actorType !== 'guests')
```
nickvergessen, line 508:
```suggestion
			return this.conversation.type === 3
				&& !this.isMyMsg
				&& this.actorType === ATTENDEE.ACTOR_TYPE.USERS
				&& this.$store.getters.getActorType() === ATTENDEE.ACTOR_TYPE.USERS
```

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -132,6 +132,13 @@ the main body of the message as well as a quote.
 							{{ action.label }}
 						</ActionButton>
 					</template>
+					<ActionButton
```
nickvergessen, line 135: this should be moved up right after Reply

nickvergessen (approved):

PR #4856, opened 2020-12-29
Do not load full room object for share queries

rullzer:
Saves a lot of work on propfinds for example

### DIFF
--- a/lib/Manager.php
+++ b/lib/Manager.php
@@ -342,6 +342,33 @@ class Manager {
 	}
 
 	/**
+	 * @param string $userId
+	 * @return string[]
+	 */
+	public function getRoomTokensForUser(string $userId): array {
+		$query = $this->db->getQueryBuilder();
+		$query->select('r.token')
+			->from('talk_attendees', 'a')
+			->leftJoin('a', 'talk_rooms', 'r', $query->expr()->eq('a.room_id', 'r.id'))
+			->where($query->expr()->eq('a.actor_id', $query->createNamedParameter($userId)))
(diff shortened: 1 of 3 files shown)


lib/Manager.php
```diff
@@ -341,6 +341,39 @@ public function getRoomsForUser(string $userId, bool $includeLastMessage = false
...
+	public function getRoomTokensForUser(string $userId): array {
+		$query = $this->db->getQueryBuilder();
+		$query->select('token')
+			->from('talk_rooms', 'r')
+			->leftJoin('r', 'talk_attendees', 'a', $query->expr()->andX(
+				$query->expr()->eq('a.actor_id', $query->createNamedParameter($userId)),
+				$query->expr()->eq('a.actor_type', $query->createNamedParameter(Attendee::ACTOR_USERS)),
+				$query->expr()->eq('a.room_id', 'r.id')
+			))
+			->leftJoin('a', 'talk_sessions', 's', $query->expr()->andX(
```
nickvergessen, line 357: You are not using it so drop it

lib/Manager.php
```diff
@@ -341,6 +341,39 @@ public function getRoomsForUser(string $userId, bool $includeLastMessage = false
...
+				$query->expr()->eq('a.id', 's.attendee_id')
+			))
+			->where($query->expr()->isNotNull('a.id'));
```
nickvergessen, line 360: In this case it might even make more sense to select from this table and left join the rooms instead

nickvergessen (changes requested):

nickvergessen (approved):

PR #4479, opened 2020-10-27
Pending: Pinning messages

jakobroehrl:
solves: https://github.com/nextcloud/spreed/issues/3390

grafik

### Todo
- [x] restrict to moderators in groups
- [x] store if a msg is pinned in the local strore
- [ ] store if a msg is pinned in the database (needs backend work)
- [ ] Showing pinned messages on top of the chat room

Issue #3390, opened 2020-04-25
Pinning messages


jakobroehrl:
In a big room it's sometimes hard to find important messages.
Maybe it would be nice to favourite some of them and have the possibility to show only favourited messages.

What do you think about this?

nickvergessen:
Favorites not so much, but I couldimagine to ping some messages into a chat which then are shown always on top, similar to the featue in Telegram and Line

Nyco:
Favorites and pins usually do not work in terms of UX and usage.

jancborchardt:
@Nyco is there some user research about that? Whenever doing a :-1: on a comment please explain it more, otherwise it comes off as a bit unfriendly. ;)

Agree with @nickvergessen, others offer this, Slack as well. To be on par with expectations, this is something we should have at some point.

Nyco:
> Whenever doing a 👎 on a comment please explain it more, otherwise it comes off as a bit unfriendly. ;)

Oh sorry, I apologise.
I am a bit confused, I thought thumbs up and down were simples votes.

> @Nyco is there some user research about that?

Yes, we had interviewed people on the use of personal chats (Telegram, WhatsApp, Messenger,...) and team chats (Slack, Discord, Teams...), and it turns out people just don't use pins and favorites for messages: either they just don't see (anymore) or filter out that feature, or they just don't see any need. Sometimes they also find the UI to be polluting.

nickvergessen:
Nearly all my Telegram and Line group chats have a pinned message.
I guess it's just a matter of habbits.

jancborchardt:
Yeah, first off it would be good to know _how many_ people were interviewed, since that could be anecdotal or biased. I would also take interviews with a grain of salt in general – asking people what they do or want generally gives you a less accurate answer than just watching what they actually do. (Please do clarify if I’m misunderstanding what you mean by "interviews".)

HammyHavoc:
Being able to favourite messages and link to them directly is pretty important and one of the major features of Rocket.Chat that we're sorely missing.

PVince81:
> Being able to favourite messages and link to them directly is pretty important and one of the major features of Rocket.Chat that we're sorely missing.

Direct link to messages is technically already possible.
When searching for a message it generates a URL that makes the list scroll to the target message.
What's missing is the ability to copy that message link directly from the message row.

PVince81:
I've raised https://github.com/nextcloud/spreed/issues/5097 for copying the direct link

[...]

PR #4857, opened 2020-12-29
Pinning messages 2

jakobroehrl:

grafik
grafik
grafik

continuation of #4479
solves: #3390

Todo
 - [x] store if a msg is pinned in the local strore
 - [ ] store if a msg is pinned in the database
 - [ ] Showing pinned messages on top of the chat room

### DIFF
--- a/lib/Model/Message.php
+++ b/lib/Model/Message.php
@@ -66,6 +66,9 @@ class Message {
 	/** @var string */
 	protected $actorDisplayName = '';
 
+	/** @var bool */
+	protected $isPinned = false;
+
 	public function __construct(Room $room,
 								Participant $participant,
 								IComment $comment,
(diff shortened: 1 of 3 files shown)


jakobroehrl:
@nickvergessen 
How to store the pinning status of a message in the database?
Add a route Chat#pinMessage and add a methode in the ChatController, right? From where I do the call to this new route?

nickvergessen:
So yeah, API wise I would say you send a post against a message endpoint, but for getting I think we need to return it on the room api otherwise it doesn't scale.
Also the info needs to be saved on the rooms table then, otherwise we need to query over all comments to find the matches.

But then again, I'm not really sure how we would want to display this, also do we allow 1 or more pins, etc. Because that heavily influences the database layout as well as the UI layout

PR #4861, opened 2020-12-29
Delete chat messages

### LINKED ISSUE #774: Allow to delete chat messages
Enhancement / feature request: 

Moderator as well as user who wrote a message should be able to remove a chat message again.
Use case: somebody copy/pasted a password into the chat by accident. Legally problematic text, bad language, etc.

Should be handled in similar way as with any other messenger.

As discussed with Joas: might make sense to instead of completing deleting the line, replace it with a placeholder (i.e. some gray bar or similar) so people know that a message/line has been removed, so that other messages around that still make sense in the context.

UI-wise: a trashcan icon on the right side that appears when the user hovers over a chat line.

jakobroehrl:
Fix #774

Todo
 - [x] Delete message content from database
 - [x] Add a system message so the old one can be deleted/updated in stores (web ui and mobile clients)
 - [x] Currently limited to 6h old messages
 - [x] Show a warning if matterbridge is active
 - [x] Write integration tests
 - [x] Add capability
 - [x] Adjust documentation
 - [x] change delete possibility: Only my messages within 6 hours after sending can be deleted

### DIFF
--- a/docs/chat.md
+++ b/docs/chat.md
@@ -96,6 +96,34 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
     - Data:
         The full message array of the new message, as defined in [Receive chat messages of a conversation](#receive-chat-messages-of-a-conversation)
 
+
+## Deleting a chat message
+
+* Method: `DELETE`
+* Endpoint: `/chat/{token}/{messageId}`
+
+* Response:
+    - Status code:
+        + `200 OK` - When deleting was successful
+        + `202 Accepted` - When deleting was successful but Matterbridge is enabled so the message was leaked to other services
+        + `400 Bad Request` The message is already older than 6 hours
+        + `403 Forbidden` When the message is not from the current user and the user not a moderator
+        + `403 Forbidden` When the conversation is read-only
+        + `404 Not Found` When the conversation or chat message could not be found for the participant
+        + `412 Precondition Failed` When the lobby is active and the user is not a moderator
+
+    - Header:
+
+        field | type | Description
+        ------|------|------------
+        `X-Chat-Last-Common-Read` | int | ID of the last message read by every user that has read privacy set to public. When the user themself has it set to private the value the header is not set (only available with `chat-read-status` capability)
+
+    - Data:
+        The full message array of the new system message "You deleted a message", as defined in [Receive chat messages of a conversation](#receive-chat-messages-of-a-conversation)
+        The parent message is the object of the deleted message with the replaced text "Message deleted by you".
+        This message should not be displayed to the user but instead be used to remove the original message from any cache/storage of the device.
+
+
 ## Mark chat as read
 
 * Method: `POST`
@@ -117,7 +145,6 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
         field | type | Description
         ------|------|------------
         `X-Chat-Last-Common-Read` | int | ID of the last message read by every user that has read privacy set to public. When the user themself has it set to private the value the header is not set (only available with `chat-read-status` capability)
-
 
 
 ## Get mention autocomplete suggestions
--- a/docs/chat.md
+++ b/docs/chat.md
@@ -44,7 +44,7 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
         `actorDisplayName` | string | Display name of the message author
         `timestamp` | int | Timestamp in seconds and UTC time zone
         `systemMessage` | string | empty for normal chat message or the type of the system message (untranslated)
-        `messageType` | string | Currently known types are `comment`, `system` and `command`
+        `messageType` | string | Currently known types are `comment`, `comment_deleted`, `system` and `command`
         `isReplyable` | bool | True if the user can post a reply to this message (only available with `chat-replies` capability)
         `referenceId` | string | A reference string that was given while posting the message to be able to identify a sent message again (only available with `chat-reference-id` capability)
         `message` | string | Message string with placeholders (see [Rich Object String](https://github.com/nextcloud/server/issues/1706))
@@ -177,3 +177,4 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
 * `moderator_demoted` - {actor} demoted {user} from moderator
 * `guest_moderator_promoted` - {actor} promoted {user} to moderator
 * `guest_moderator_demoted` - {actor} demoted {user} from moderator
+* `message_deleted` - Message deleted by {actor} (Should not be shown to the user)
--- a/docs/chat.md
+++ b/docs/chat.md
@@ -110,6 +110,7 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
         + `403 Forbidden` When the message is not from the current user and the user not a moderator
         + `403 Forbidden` When the conversation is read-only
         + `404 Not Found` When the conversation or chat message could not be found for the participant
+        + `405 Method Not Allowed` When the message is not a normal chat message
         + `412 Precondition Failed` When the lobby is active and the user is not a moderator
 
     - Header:
--- a/lib/Controller/ChatController.php
+++ b/lib/Controller/ChatController.php
@@ -466,6 +466,40 @@ class ChatController extends AEnvironmentAwareController {
 	 * @NoAdminRequired
 	 * @RequireParticipant
 	 *
+	 * @param int $messageId
+	 * @return DataResponse
+	 */
+	public function deleteMessage(int $messageId): DataResponse {
+		try {
+			$message = $this->chatManager->getComment($this->room, (string) $messageId);
+		} catch (NotFoundException $e) {
+			return new DataResponse([], Http::STATUS_NOT_FOUND);
+		}
+
+		$attendee = $this->participant->getAttendee();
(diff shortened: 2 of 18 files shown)


sudwhiwdh:
How will users learn what this feature means in a uncomplicated way, i.e. without reading any documentation before?
For example, in terms of the available time window of 15 minutes or how to deal with federated messages.

sudwhiwdh:
How will users learn what this feature/action means in a uncomplicated way, i.e. without reading any documentation before?
For example, in terms of the available time window of 15 minutes or how to deal with federated messages.

jakobroehrl:
> How will users learn what this feature/action means in a uncomplicated way, i.e. without reading any documentation before?
> For example, in terms of the available time window of 15 minutes 

Only give the option when it's possible.

> or how to deal with federated messages.

Hmm, do you have an idea. If that is a problem let's close this PR.

sudwhiwdh:
> > How will users learn what this feature/action means in a uncomplicated way, i.e. without reading any documentation before?
> > For example, in terms of the available time window of 15 minutes
> 
> Only give the option when it's possible.

However, this still does not explain to users in which time window they can/should react with the Delete action, does it?

> 
> > or how to deal with federated messages.
> 
> Hmm, do you have an idea. If that is a problem let's close this PR.

I don't know if that's a problem. What problems do you see in this case?

For me, it's definitely a matter of transparent communication so that there are no false expectations. It would be simply important for me that users understand in a comprehensible way from which server or device the message can be deleted with this action and for which people the message will still be visible after deletion.

Btw. is it technically possible to retrieve a message from a federated Nextcloud server within fifteen minutes?

nickvergessen:
So multiple things with this:

1. Deleting gives a wrong impression when "Matterbridge" is configured for this channel, as the message is synced already to other chats where we can no longer delete the message. From my personal POV we should either show a warning on deleting that the message was federated and always will be "public" information, or not allow deleting in matterbridged conversations at all.
2. Since the last message id of a conversation is set and the messages are downloaded by our mobile clients, deleting from the DB is not a good idea and doesn't help a lot. Instead we should only wipe the content of the message (if the database structure allows it to be empty) and at the same time we need to add a "hidden" message that the message with ID X was deleted, so the clients can remove the message from their local store as well.
3. Regarding time restriction, we checked a bit what others do and they are a bit more "lax", Whatsapp 1 hour, Signal 3 hours, Line IM 24 hours, Telegram infinite. So we would go with 6 hours for now.

I can help with the backend if you want.

lib/Model/Message.php
```diff
@@ -161,6 +161,13 @@ public function isReplyable(): bool {
 			\in_array($this->getActorType(), [Attendee::ACTOR_USERS, Attendee::ACTOR_GUESTS]);
 	}
 
+	/**
+	 * Only my messages within 15 minutes after sending can be deleted
+	 */
+	public function isDeleteable(): bool {
+		return true;
+	}
```
nickvergessen, line 169: So for now we think this is not needed on the API, as the expression is rather constant:
```
message time > time() - 6h && (author === current user || current user is moderator)
```
Also we will check this on the deletion endpoint and therefor the time state needs to be updated live, as when you open the chat in the morning and have it open all day the time might expire, but your browser would not refetch the state (same for the mobile clients anyway).

nickvergessen:
Converting to draft for backend work

jakobroehrl:
@nickvergessen Please ping me if you are ready with the backend work, I can test this, if you want.

nickvergessen:
Should be ready to play around with it, integration tests are still missing, same for capability and documentation

jakobroehrl:
I can delete, but after the page reload the deleted messaage is visible again

nickvergessen:
I forcepushed because of a rebase. make sure your pulling worked

jakobroehrl:
It's working super nice!

docs/chat.md
```diff
@@ -96,6 +96,32 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
...
+        + `412 Precondition Failed` When the lobby is active and the user is not a moderator
+
+    - Header:
+
+        field | type | Description
+        ------|------|------------
+        `X-Chat-Last-Common-Read` | int | ID of the last message read by every user that has read privacy set to public. When the user themself has it set to private the value the header is not set (only available with `chat-read-status` capability)
+
+    - Data:
+        TODO
```
nickvergessen, line 122: Todo

Tekgnome-svg:
This is amazing work. I was stunned that Talk didn't have the ability for user deletable messages.

mahibi:
i'm able to delete a message of my conversation partner without having admin or moderator rights. Via REST and WebUI. 
As far as i understood this should't be allowed? I expected 403 Forbidden because "the message is not from the current user and the user not a moderator".

nickvergessen:
Note: in a one-to-one chat both users are in fact moderators

Ivansss:
The lastMessage of a conversation should not be the "deletion system message"

Screenshot 2021-01-29 at 16 08 39

@nickvergessen

nickvergessen:
Hmm so after checking all system messages end up in the last message as well, but they also all display. I'm not sure what's happening if your read marker and the last message can not match. So I would tend to keep it like this for now and just not show it in the message list 🙈

nickvergessen:
Ready to review

lib/Controller/ChatController.php
```diff
@@ -462,6 +468,64 @@ public function receiveMessages(int $lookIntoFuture,
...
+		$attendee = $this->participant->getAttendee();
+		if (!$this->participant->hasModeratorPermissions(false)
+			&& ($message->getActorType() !== $attendee->getActorType()
+				|| $message->getActorId() !== $attendee->getActorId())) {
+			// Actor is not a moderator or not the owner of the message
+			return new DataResponse([], Http::STATUS_FORBIDDEN);
+		}
+
+		$maxDeleteAge = $this->timeFactory->getDateTime();
+		$maxDeleteAge->sub(new \DateInterval('PT6H'));
```
PVince81, line 496: I guess not much use of making this configurable ? (through occ at least)

src/services/messagesService.js
```diff
@@ -103,8 +103,20 @@ const postNewMessage = async function({ token, message, actorDisplayName, refere
 	return response
 }
 
+/**
+ * Posts a new messageto the server.
```
PVince81, line 107: Deletes a message from the server

src/store/messagesStore.js
```diff
@@ -127,6 +128,16 @@ const mutations = {
 		Vue.delete(state.messages[message.token], message.id)
 	},
 
+	/**
+	 * Deletes a message from the store.
+	 * @param {object} state current store state;
+	 * @param {object} message the message;
+	 * @param {string} tempMessage Placeholder message until deleting finished
```
PVince81, line 135: mismatched argument name

PVince81:
- [ ] BUG? submitting the DELETE action multiple times for the same message returns 200 with payload. Should we make that a 403 ?

PVince81:
- [ ] BUG: should not be able to delete system messages

Steps to reproduce:

1. Join and leave a call to generate system messages
2. Post a message then delete it
3. Copy the curl from the network console
4. With vue inspector, find the id of the system messages about joining/leaving the call
5. Use the curl DELETE and replace the id with the one from the system messages

Expected: 403 error
Actual: messages appear as deleted in the messages list

lib/Controller/ChatController.php
nickvergessen: No, we don't support deleting after 6h

PVince81:
nice work! I didn't find any other issues while attempting to break it :-)

ma12-co (changes requested):
Right now the reply button is shown on a deleted message. Upon trying to reply to it, an error is triggered. 

Screenshot from 2021-02-01 16-33-33

I would remove all the actions, the time and the hovering background change on deleted messages

nickvergessen:
> Right now the reply button is shown on a deleted message. Upon trying to reply to it, an error is triggered.

Doesn't work here:
Peek 2021-02-02 10-48

ma12-co:
Weird, can't reproduce anymore. Should I take care of the looks a bit?

ma12-co (approved):

PVince81:
some hardening needed, see bugs above.
can also be done separately.

your call @nickvergessen

nickvergessen:
Blocking deletion of system messages and share messages now.

PVince81 (approved):
:+1:

see conflict though

nickvergessen:
Ah, conflict from #5019 
Rebased

PVince81:
All green! Merge for 21 or hold until the 22 split ?

kinimodmeyer:
> Note: in a one-to-one chat both users are in fact moderators

can we not change the condition to allow this only for moderator in group chats?
sounds not 100% right when i write a important message to a employe and he can just delete them if he wants.
in group chats ppl understand that there is a "moderator/admin" who could do this.

PVince81:
@kinimodmeyer please feel free to raise a new ticket for this request

theoneandonly-vector:
"Currently limited to 6h old messages" is there a way to change this?

marcoambrosini:
@nickvergessen @jancborchardt should this be tweakable in the admin settings?

nickvergessen:
No

PR #4907, opened 2021-01-07
Move matterbridge settings into the conversation settings dialog

ma12-co:
Screenshot from 2021-01-07 15-44-45


eneiluj (commented):
Nice!

I don't know why the action button is not correctly sized.
plop

Also, Matterbridge log is displayed in a modal on top of the settings modal. It's not shocking but what do you think about that?

Except that, looking good.

nickvergessen (changes requested):
See comment from Julien

ma12-co:
Fixed the actions button hover feedback height and adjusted margins a bit

> Also, Matterbridge log is displayed in a modal on top of the settings modal. It's not shocking but what do you think about that?

While I agree that it's not very elegant, I think that overall this is an improvement, especially the fact that we're moving every conversation related setting in one place as opposed to having them scattered around the UI. So I would say let's move forward with this :)

nickvergessen:
Peek 2021-01-12 13-21

Can the hovers have the same color?

nickvergessen:
Okay it's a bug in the vue lib...

nickvergessen:
Reported at https://github.com/nextcloud/nextcloud-vue/issues/1665

nickvergessen (approved):

PR #4932, opened 2021-01-11
Add deck cards as dedicated simple rich object string parameters

nickvergessen:
Simple text link for now, with "Stack in Board" on hover

https://github.com/nextcloud/server/pull/25070


PVince81:
Not sure how to test this, I tried checking out the deck app (and bumped against https://github.com/nextcloud/deck/issues/2721).
After creating boards, lists, cards I tried copying their link from the browser URL (I didn't find a way to copy link from the UI) then pasted into a chat, but it didn't get formatted.

So I guess there must be another way to share cards that maybe is hidden or not enabled ?
Or is this only part of a future feature that requires more work ?

PVince81:
I've traced the tickets back to https://github.com/nextcloud/deck/issues/2630

The sharing action on the screenshot doesn't seem to be available on master, so probably not implemented yet.

PVince81 (approved):
:+1: code looks fine, could not test (see above)

Issue #4174, opened 2020-09-17
Unclear which replies are to you or to others


jancborchardt:
## Steps to reproduce
1. Go to a conversation which has messages with replies
2. Scroll through cause you were mentioned
3. Try and find the relevant replies for you

### Expected behaviour
It’s clear which replies are to messages from you, and which are to messages from others.

### Actual behaviour
All quotes look the same. It’s very difficult to see when a reply is to something you said or to someone else. You need to manually read the name.

One way would be e.g. to only make the left bar blue for the qoutes which are from you. The others could be grey like `var(--color-border-dark)`:

Before | After
-|-
![quotes before](https://user-images.githubusercontent.com/925062/93508750-12076b80-f91f-11ea-84bf-6d15330ca711.png) | ![quotes after](https://user-images.githubusercontent.com/925062/93508748-10d63e80-f91f-11ea-8ff8-1c2220f23693.png)

As discussed @ma12-co

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/4940

PR #4940, opened 2021-01-13
Added quote border difference when own message is quoted

PVince81:
Fixes https://github.com/nextcloud/spreed/issues/4174

image

### DIFF
--- a/src/components/Quote.vue
+++ b/src/components/Quote.vue
@@ -151,7 +151,13 @@ export default {
 		},
 
 		isOwnMessageQuoted() {
-			return this.actorId === this.$store.getters.getUserId()
+			let actorId = this.actorId
+			if (this.actorType === 'guests') {
+				actorId = 'guest/' + actorId
+			}
+
+			return actorId === this.$store.getters.getActorId()
+				&& this.actorType === this.$store.getters.getActorType()
 		},
 
 		isFileShareMessage() {
--- a/src/components/Quote.vue
+++ b/src/components/Quote.vue
@@ -27,6 +27,7 @@ components.
 <template>
 	<a href="#"
 		class="quote"
+		:class="{'quote-own-message': isOwnMessageQuoted}"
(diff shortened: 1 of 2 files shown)


src/components/Quote.vue
```diff
@@ -145,6 +150,10 @@ export default {
 			return displayName
 		},
 
+		isOwnMessageQuoted() {
+			return this.actorId === this.$store.getters.getUserId()
```
nickvergessen, line 154:
```suggestion
			return this.actorId === this.$store.getters.getActorId()
```
Then it also should work for guests

nickvergessen (changes requested):
Let's add guest support properly.

Currently any quote of a guest looks like it was "yours"

src/components/Quote.vue
PVince81: just tried with guests now, the actor type "guests" match but not the actor id:

in my test with a guest, this.actorId is "95e4a466030ba6a55d521429f5961e12a5397d07"
and this.$store.getters.getActorId() is "guest/5ddeb308573914adfda3d870c8beb6ac4dbf566b"

PVince81: ohh, maybe it's because I refreshed the page, so the actor id would change...

in any case I think I need to split the "/" there

PVince81: https://github.com/nextcloud/spreed/pull/4940/commits/366464cd419bf0433a64e1b555cd3e26660ecb02

PVince81:
@nickvergessen please check again, the guest quote should work now

As mentioned above, refreshing the page as guest would change the guest id, so need to stay on the same page.
Hopefully in the future we can find a way to keep those guest ids more stable (maybe with php session id or so)

nickvergessen (approved):

jancborchardt:
Btw @PVince81 great detail quality-of-life improvement which really helps daily, thank you! :)

Issue #4817, opened 2020-12-21
Talk 11 alpha 2: See other participants when I share my screen


framarini02:
when I share the screen, it is not possible to see the screen also other participants as in Talk 10.

PVince81:
@framarini02 you should be able to: see on the bottom right there's an icon to uncollapse the stripe

framarini02:
2if I press that button I can only see me. Instead I would like to have the possibility to see other people in that black bar that is created.

PVince81:
makes me think: maybe the first time the stripe automatically collapses we should show a tooltip there to inform users that they can bring it back

thoughts @ma12-co ?

PVince81:
> 2if I press that button I can only see me. Instead I would like to have the possibility to see other people in that black bar that is created.

hmmm then it's a bug, you should be able to see everyone as you do when not sharing

PVince81:
I've just tested and it works for me:
<img width="960" alt="Screenshot_20201221_110102" src="https://user-images.githubusercontent.com/277525/102764716-e2cc2480-437b-11eb-81e5-2d4b1c2f83d6.png">

(the colors are the videos from the test peers using Firefox fake streams)

I'm also able to switch to grid view to see other participants.

if you don't see anyone in the bottom bar, please fill the issue template and report logs, etc: https://raw.githubusercontent.com/nextcloud/spreed/master/.github/ISSUE_TEMPLATE/bug_report.md

framarini02:
Checking the problem I notice that if we are only two participants I can see only myself in the bar below the sharing. On the other hand, if we are three or more participants, I can see the others under the screen sharing.

PVince81:
might be similar to https://github.com/nextcloud/spreed/issues/4051

PVince81:
okay, I just compared these but they are not the same.

so here the bug is that the bottom stripe is not showing any participants if there would be only a single one there. it only starts showing starting with 2...

PVince81:
From what I see the logic is fine when no screen shares are involved, the stripe is hidden when less than two participants are present: https://github.com/nextcloud/spreed/blob/master/src/components/CallView/Grid/Grid.vue#L59

I tried changing it to "less than one" but then the user whose video is visible in the big screen is also visible in the stripe.

So one idea would be to keep the "less than two" logic but extend it with "unless there is a screen share in the top".

PVince81:
PR here that fixes it: https://github.com/nextcloud/spreed/pull/4941

[...]

PR #4941, opened 2021-01-13
Allow single video in stripe if screen share in progress

PVince81:
So far, the stripe was designed to not duplicate the video of whoever is
displayed in the top. However, when a screen share is in progress, that
user's video must move to the stripe so we can see them.

Fixes https://github.com/nextcloud/spreed/issues/4817

Screenshot_20210113_154128

Left side is the user sharing the screen (admin). Own video is visible in the stripe and also the green video from the other user.
Right side is the user receiving the screen share (alice). The video of "admin" is visible also in the stripe.

So far so good, however when you expand the window width, it looks bad:
Screenshot_20210113_154320

Thoughts @ma12-co ?

### DIFF
--- a/src/components/CallView/Grid/Grid.vue
+++ b/src/components/CallView/Grid/Grid.vue
@@ -20,7 +20,7 @@
 -->
 
 <template>
-	<div class="grid-main-wrapper" :class="{'is-grid': !isStripe, 'transparent': isLessThanTwo}">
+	<div class="grid-main-wrapper" :class="{'is-grid': !isStripe, 'transparent': isLessThanTwoVideos}">
 		<button v-if="isStripe"
 			class="stripe--collapse"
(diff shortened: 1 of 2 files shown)


src/components/CallView/Grid/Grid.vue
```diff
@@ -262,6 +262,10 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		screens: {
+			type: Array,
+			default: [],
```
nickvergessen, line 267: Type of the default value for 'screens' prop must be a function

PVince81:
- [ ] try making the stripe transparent when less than one video inside (minus own video)

nickvergessen (approved):

nickvergessen:
Oh, now the todo pops up. Didn't show before

Issue #4809, opened 2020-12-18
Conversation description sometimes not appearing when switching from changelog room


PVince81:
## Steps to reproduce
1. Create two rooms
2. Set a distinct description for both
3. Click on the change log room
4. Switch back and forth between the rooms and also the changelog room, don't always wait for the room to be fully loaded

### Expected behaviour
Description always loaded/displayed.

### Actual behaviour
Sometimes the room description is not displayed and it shows as if it was empty

## Talk app

**Talk app version:** v11.0.0-alpha.2

**Custom Signaling server configured:** none

## Browser

**Chromium:** 87.0.4280.88

## Notes

It is possible that the problem is caused by switching while the room hasn't fully loaded.

PVince81:
<img width="510" alt="image" src="https://user-images.githubusercontent.com/277525/102621702-40742d00-4140-11eb-90c6-7017514bf89d.png">

The description seems to be set on the "Description" object
<img width="627" alt="image" src="https://user-images.githubusercontent.com/277525/102621760-571a8400-4140-11eb-9de4-4eedb156a125.png">

but hasn't been passed down to the richcontenteditable:
<img width="644" alt="image" src="https://user-images.githubusercontent.com/277525/102621806-68fc2700-4140-11eb-8097-d58c1fa113a3.png">

not sure what mechanism we rely on for passing down, plain Vue reactivity ?

@ma12-co

PVince81:
Okay, I'm onto something: 
- in my previous analysis I mixed up "description" and "descriptionText"
- the property "description" of the `<Description>` component is properly updated.
- for some reason the watcher is not called, so the attribute `descriptionText` is not getting updated.

PVince81:
Ok, problem solved: the watcher is not called initially but only for property updates. So need to initialize "description" with "descriptionText" initially. PR here: https://github.com/nextcloud/spreed/pull/4942

PR #4942, opened 2021-01-13
Properly initialize descriptionText with description

PVince81:
The conversation descriptionText attribute must be initialized when the
component is created as the watcher is only called on updates, not
creation.

Fixes https://github.com/nextcloud/spreed/issues/4809


nickvergessen (commented):
🙈 🙈 🙈 🙈 🙈

nickvergessen (approved):

PR #4943, opened 2021-01-13
Correctly use the displayname for groups in admin settings

### LINKED ISSUE #4683: Admin settings for "limit to group" don't use group display name
Correct (SIP setting) | Incorrect (Limit to group)
---|---
Bildschirmfoto von 2020-11-27 12-12-29 | Bildschirmfoto von 2020-11-27 12-12-59

nickvergessen:
Fix #4683

To create such a group use Nextcloud 18 or higher occ:
```
GROUPID=$(uuidgen)
GROUPID="group${GROUPID:8:28}"

./occ -vvv group:add $GROUPID --display-name "Dragon Crew"
```

Before | After
---|---
Bildschirmfoto vom 2021-01-13 16-52-21 | Bildschirmfoto vom 2021-01-13 16-51-50

### DIFF
--- a/src/components/AdminSettings/AllowedGroups.vue
+++ b/src/components/AdminSettings/AllowedGroups.vue
@@ -45,6 +45,8 @@
 				:loading="loadingGroups"
 				:show-no-options="false"
 				:close-on-select="false"
+				track-by="id"
+				label="displayname"
 				@search-change="searchGroup" />
 
 			<button class="button primary"
@@ -67,6 +69,8 @@
 				:loading="loadingGroups"
 				:show-no-options="false"
 				:close-on-select="false"
+				track-by="id"
+				label="displayname"
 				@search-change="searchGroup" />
 
 			<button class="button primary"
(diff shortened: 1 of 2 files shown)


src/components/AdminSettings/AllowedGroups.vue
```diff
@@ -131,11 +135,15 @@ export default {
...
+		this.allowedGroups = loadState('talk', 'allowed_groups').sort(function(a, b) {
+			return a.displayname.localeCompare(b.displayname)
+		})
+		this.canStartConversations = loadState('talk', 'start_conversations').sort(function(a, b) {
+			return a.displayname.localeCompare(b.displayname)
+		})
 		this.startCalls = startCallOptions[parseInt(loadState('talk', 'start_calls'))]
-		this.groups = [...new Set(this.allowedGroups.concat(this.canStartConversations))].sort(function(a, b) {
-			return a.localeCompare(b)
+		this.groups = Array.from(new Set(this.allowedGroups.concat(this.canStartConversations).map(g => JSON.stringify(g)))).map(g => JSON.parse(g)).sort(function(a, b) {
```
PVince81, line 145: some weird gymnastics here, I'm not really sure why

either move this into a function with an explicit name or add a code comment, if there is no better way

PVince81 (approved):
:+1: code looks fine, one minor comment

Tested and seen working for both the fields in the talk admin settings

src/components/AdminSettings/AllowedGroups.vue
nickvergessen: I'm bad at JS and just copied SO until it worked.
we have two arrays with objects (id + name) and want to merge them into a list without duplicates.

PVince81: the Set thing is fine.
I was mostly wondering about the converting to JSON string then parsing again.

ok, let's try and make it more readable to understand it:
```
Array.from(
   new Set(
      this.allowedGroups.concat(this.canStartConversations)
         .map(g => JSON.stringify(g))
   )
)
.map(g => JSON.parse(g))
.sort(...)
```

hmmmm, so we are trying to make the list unique by using the JSON **as string** as the key for uniqueness, so it only works with the assumption that the JSON object's key order is always the same

I guess the alternative would be to construct an object where keys are the id and values the JSON object, for deduplicating by id, and then retrieving only the values.

Either way I guess the approach is fine, please at least add a comment to clarify deduplication to avoid future head scratching.

PVince81 (approved):
:+1:

PR #4951, opened 2021-01-14
Cannot disable sip

nickvergessen:
🙈 I wonder why `LockingSettings.vue` and `SipSettings.vue` are so different when they do the same thing :D

### DIFF
--- a/src/components/ConversationSettings/LockingSettings.vue
+++ b/src/components/ConversationSettings/LockingSettings.vue
@@ -68,20 +68,6 @@ export default {
 		},
 	},
 
-	watch: {
-		value(value) {
-			this.listable = value
-		},
-	},
-
-	mounted() {
-		if (this.token) {
-			this.listable = this.value || this.conversation.listable
-		} else {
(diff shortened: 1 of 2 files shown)


PVince81:
> see_no_evil I wonder why `LockingSettings.vue` and `SipSettings.vue` are so different when they do the same thing :D

because they grew from some initial confusion and eventually it seems I only adjusted a single one later on

PVince81 (approved):
:+1:

nickvergessen:
Which is good and which is bad? so I can use either and do a string replace?

PVince81:
> Which is good and which is bad? so I can use either and do a string replace?

I don't remember. But one of them might not properly reset the checkbox state in case of error.

PVince81:
I had some WIP for error handling and trying to reset the states here: https://github.com/nextcloud/spreed/pull/4671

nickvergessen:
Seems this is only broken in 21, works fine in 20.
So it's a regression from  #4576

PR #4956, opened 2021-01-14
Fix display names with sip users

nickvergessen:


PVince81:
@nickvergessen issue description and steps to reproduce, to understand the problem ?

src/components/CallView/shared/Video.vue
```diff
@@ -225,22 +225,53 @@ export default {
 			return peerData
 		},
 
+		participant() {
```
PVince81, line 228: reuse or integrate into the participant mixin ?

note that the mixin would return the disconnected flag instead of just an empty object

src/utils/webrtc/webrtc.js
```diff
@@ -417,7 +417,7 @@ export default function initWebRTC(signaling, _callParticipantCollection, _local
 			callParticipantModel.set('speaking', (event.flags & PARTICIPANT.SIP_FLAG.SPEAKING) > 0)
 			callParticipantModel.set('audioAvailable', (event.flags & PARTICIPANT.SIP_FLAG.MUTE_MICROPHONE) === 0)
 			callParticipantModel.set('raisedHand', {
-				state: (event.flags & PARTICIPANT.SIP_FLAG.RAISE_HAND) === 0,
+				state: (event.flags & PARTICIPANT.SIP_FLAG.RAISE_HAND) !== 0,
```
PVince81, line 420: nice catch! I couldn't test back then so it didn't catch the eye

PVince81 (commented):
Code looks fine in general as far as I could understand.

Please check my comment about the participant mixin to avoid further code duplication.

nickvergessen:
> issue description and steps to reproduce, to understand the problem ?

Currently SIP users have no display name for guests and logged in users, it just shows the user id/email addresss/guest id

src/utils/webrtc/webrtc.js
nickvergessen: yeah I saw it flashing by everytime the user muted and unmuted, that's how I noticed

src/components/CallView/shared/Video.vue
nickvergessen: participant mixin is always the current user, this here is peerId, not yourself.

PVince81 (approved):
:+1:

PR #4959, opened 2021-01-15
Update @babel/preset-env and fix commonjs loading

PVince81:
Update @babel/preset-env to 7.12.11 has revealed an issue with the event
bus library related to the CommonJS format.

The webpack config was adjusted to properly load CommonJS libraries.

Fixes https://github.com/nextcloud/spreed/issues/4818


PVince81:
the fix is included in https://github.com/nextcloud/spreed/pull/4829, closing

PR #4974, opened 2021-01-19
Remove "unsupported browser" warning for Opera

pr0t4zy:
This adds Opera to the popup  "The browser you're using is not fully supported by Nextcloud Talk. Please use the latest version of Mozilla Firefox, Microsoft Edge, Google Chrome or Apple Safari." and removes the popup (for Opera).

Note: Opera v36 is based on Chromium v49.

### DIFF
--- a/src/mixins/browserCheck.js
+++ b/src/mixins/browserCheck.js
@@ -48,6 +48,9 @@ const browserCheck = {
 		isChrome() {
 			return this.browser.name === 'Chrome' || this.browser.name === 'Chromium'
 		},
+		isOpera() {
+			return this.browser.name === 'Opera'
+		},
 		isSafari() {
 			return this.browser.name === 'Safari' || this.browser.name === 'Mobile Safari'
 		},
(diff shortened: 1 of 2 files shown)


nickvergessen:
So the main issue/question is, if this is fully working with the webrtc-adapter lib we use:
https://github.com/webrtcHacks/adapter/blob/master/src/js/utils.js#L167-L176

pr0t4zy:
Opera will pass the check in linked utils.js. It will get "chrome" identification, there is an explicit comment: 
> // Chrome, Chromium, Webview, Opera.

Everything will work fine.

src/mixins/browserCheck.js
```diff
@@ -65,19 +68,21 @@ const browserCheck = {
 		isFullySupported() {
 			return (this.isFirefox && this.majorVersion >= 52)
 			|| (this.isChrome && this.majorVersion >= 49)
+			|| (this.isOpera && this.majorVersion >= 36)
```
nickvergessen, line 71: So yeah tested and all fine… with Opera 73.
From my POV we can add opera, but then we shouldn't claim such an old version we are not testing with. I would just go with 72 for now.

pr0t4zy: Ok, changed,

nickvergessen (approved):

pr0t4zy:
Thanks for approving!

PR #4975, opened 2021-01-19
Allow to resend temporary messages if the failure was not permissions

### LINKED ISSUE #2454: Error handling for temporary messages
Temporary messages are local messages that show up in the chat view as soon as the user hits enter. For now, while waiting for the server response we display a little spinner instead of the message time. Once we get a confirmation from the server we display the timestamp. Atm we don't handle unsent messages.

Screenshot_20191112_103056

Things to do: 

- [x] Properly style a temporary message like regular messages (avatar / no avatar);
- [x] Display the temporary message in the conversations preview text. 
- [ ] Set a timeout and add a warning if no response from the server is received. 

^ for this last one I'm not sure how to display the warning.. Red text on top of the message? Red message? What do you think @nextcloud/designers-talk  @nextcloud/talk?  Also I was thinking to display the delete and retry actions explicitly (no 3 dot menu) on the right.

nickvergessen:
Fix #2454 

Peek 2021-01-19 21-03

Also supports parallel errors much better now. If previous you would send each line as a message:
```
Lorem
ipsum
dolor
sit
amet
```
And they all failed e.g. due to network, you got "amet" recovered into the input while the rest is gone. Now you have to click on them one by one, but therefore can also recover them all.

### DIFF
--- a/src/assets/buttons.scss
+++ b/src/assets/buttons.scss
@@ -1,0 +1,57 @@
+/**
+ * @copyright Copyright (c) 2020 Marco Ambrosini <marcoambrosini@pm.me>
+ *
+ * @author Marco Ambrosini <marcoambrosini@pm.me>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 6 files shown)


nickvergessen:
For testing I patched:
```diff
diff --git a/lib/Controller/ChatController.php b/lib/Controller/ChatController.php
index 9aecf0fbe..539d14188 100644
--- a/lib/Controller/ChatController.php
+++ b/lib/Controller/ChatController.php
@@ -165,6 +165,7 @@ class ChatController extends AEnvironmentAwareController {
         *         found".
         */
        public function sendMessage(string $message, string $actorDisplayName = '', string $referenceId = '', int $replyTo = 0): DataResponse {
+               sleep(7);
                if ($this->userId === null) {
                        $actorType = Attendee::ACTOR_GUESTS;
                        $sessionId = $this->session->getSessionForRoom($this->room->getToken());
```
and restarted apache in the meantime

nickvergessen:
- [ ] @ma12-co can you turn the button in the retry-possible case into a correct button with 44px shade, etc.

ma12-co:
I started a file in which to start to apply the new buttons styles with the goal to upstream it once it's ready

nickvergessen:
Thanks, button works ✅  (can approve as it's my PR)

ma12-co (approved):

Issue #4147, opened 2020-09-14
Avatars and images are above avatar menu


jospoortvliet:
![Screenshot_20200914_100740](https://user-images.githubusercontent.com/551757/93060309-3854af00-f672-11ea-9533-4fb12894b576.png)

I guess the screenshot says it all.

This is on cloud.nextcloud.com with latest beta and all that 😭

jospoortvliet:
related to https://github.com/nextcloud/spreed/issues/4131 and https://github.com/nextcloud/spreed/issues/2575

nickvergessen:
Arg, what a painful back and forth...

PVince81:
Just noticed that too

It seems the avatar element has `position: sticky` and removing it solves the issue.

Apparently the sticky is there to make sure the avatar stays visible when scrolling down.

It appears that increasing the z-index for the popover menu doesn't make it appear higher.
And making the menu "sticky" as well breaks its layout.

PVince81:
Hmmm, tricky: I see that the popovermenu itself is also a child of the avatar, so it's inside a sticky element too...

One interesting thing to see is that when scrolling down, the "sticky" avatar element does get covered by the hamburger menu on top. So there is definitely a way to have another element overlap it.

I found a solution for z-index that requires to set position:relative + z-index: 1 on the currently highlighted element. I'll see if we can trigger this through code...

PVince81:
Not a simple fix: it seems the Avatar element is from the library or server. We could extend it to emit an event whenever the popup opens or not.

I tried another thing with z-index, but it requires setting it on every row in descending order, based on the id... and cannot be negative... feels like a big hack. `<div class="message-group" :style="{'z-index': 100000-id, 'position', 'relative'}">` ...

PVince81:
Would it be bad to not have the avatars be sticky any more ? It only means that when scrolling down one might not see the avatar for grouped messages.

Removing the sticky attribute would solve the overlap problem.

nickvergessen:
cc @jancborchardt can you check the last comment

PVince81:
some hope for a fix here https://github.com/nextcloud/nextcloud-vue/pull/831 as it changes the DOM structure, or at least more possibilities to explore for a fix

jancborchardt:
> Would it be bad to not have the avatars be sticky any more ? It only means that when scrolling down one might not see the avatar for grouped messages.

[...]

PR #4976, opened 2021-01-20
Bump @nextcloud-vue to 3.5.1

PVince81:
Brings in the updated Avatar component which now uses Popover.

Fixes issue with dropdown menu overlap with other avatars in the
messages list.

Fixes https://github.com/nextcloud/spreed/issues/4147

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3251,9 +3251,9 @@
 			}
 		},
 		"@nextcloud/vue": {
-			"version": "3.4.0",
-			"resolved": "https://registry.npmjs.org/@nextcloud/vue/-/vue-3.4.0.tgz",
-			"integrity": "sha512-xUI2CHQwNxmr42ewqlgWBUCUqH4PKsbMloE0S8y1NYYihth5MA0cutRJigP2qDv8n412YiHbrFfeoTh44f9loA==",
+			"version": "3.5.0",
+			"resolved": "https://registry.npmjs.org/@nextcloud/vue/-/vue-3.5.0.tgz",
+			"integrity": "sha512-2hDn9sPwQ6J0zxqXOYZYglhUynO4VojGa95/FLp+XHVCd+wmEKRZBw4yoQ61vQASMLYpjxgloQGr9CL2wuU3nw==",
 			"requires": {
 				"@nextcloud/auth": "^1.2.3",
 				"@nextcloud/axios": "^1.3.2",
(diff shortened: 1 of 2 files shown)


nickvergessen (changes requested):
- [x] Regressions: Icons are not inverted any longer on dark mode
- [ ] Instead of scrolling the popover now just overlaps the avatar. I think elipsis would be better

Bildschirmfoto von 2021-01-20 10-45-02

nickvergessen:
> Regressions: Icons are not inverted any longer on dark mode

I guess this needs fixing in server, as https://github.com/nextcloud/server/blob/f04b182b9447231b32cd5b5f8060314dee4fd801/apps/accessibility/css/dark.scss#L63 no longer matches.

PVince81:
> Regressions: Icons are not inverted any longer on dark mode

Seems the difference with the other popover menus is that in the popover menus we use "icon-x" classes for icons, but for this actions menu it's image URLs directly.

PVince81:
> > Regressions: Icons are not inverted any longer on dark mode
> 
> I guess this needs fixing in server, as https://github.com/nextcloud/server/blob/f04b182b9447231b32cd5b5f8060314dee4fd801/apps/accessibility/css/dark.scss#L63 no longer matches.

Dark mode fixes here:
- [ ] add classes for more specific targetting (not wanting to target images in a non-menu inside popover): https://github.com/nextcloud/nextcloud-vue/pull/1674
- [ ] server changes to dark.css to make it look better (screenshots included there): https://github.com/nextcloud/server/pull/25229

PVince81:
> Instead of scrolling the popover now just overlaps the avatar. I think elipsis would be better

Ellipsis means we'd need a tooltip also to be able to display the full row as there might be words behind the name like "sprechen" in German.

I looked at the actions menu and apparently it wraps there with a fixed max width, so we could do the same here:
image

Eventually once we switch to actions menu for avatars (see https://github.com/nextcloud/nextcloud-vue/pull/1670) this will be in place as well implicitly.

For now I'll manually adjust these to the avatar's popover menu.

PVince81:
- [ ] long text fixes (screenshots included there): https://github.com/nextcloud/nextcloud-vue/pull/1675

PVince81:
bumped to 3.5.1 now

PVince81:
all merged, I've retested this PR and it works.

ma12-co (approved):

PR #4977, opened 2021-01-20
allow some more padding for the search abort button

wiswedel:
before

talk_before

after
talk_after


PVince81 (approved):
:+1: tested, works. Nice catch!

Issue #2758, opened 2020-01-16
Open @-mention stays when navigating away


nickvergessen:
## Steps to reproduce
1. Share a file to a group/user
2. Join conversation and start typing `@t`
3. Navigate to another file/item/sidebar tab

### Expected behaviour
Autocomplete should close

### Actual behaviour
Autocomplete is stuck on this position on your screen until a page reload happens

> ![Peek 2020-01-16 08-58](https://user-images.githubusercontent.com/213943/72504693-9a88b600-383e-11ea-955e-fa649c65b095.gif)

nickvergessen:
cc @danxuliu if you can still have a look

PVince81:
Still happening in 11.0.0-alpha.3

PVince81:
seems the popover in that specific case is missing some handler to close when clicking outside.
doesn't matter where I click outside, it never closes.

seems to be a bug in the original library, see how in this demo it also doesn't close when clicking outside: https://fritx.github.io/vue-at/#/en/quickstart

so we'll likely need to write our own "click outside" handler in our "AdvancedInput" component for proper closing

PVince81:
bug report still open: https://github.com/fritx/vue-at/issues/114
with a potential solution to implement on our side: https://github.com/fritx/vue-at/issues/114#issuecomment-565777450

PVince81:
I had a quick local try with only the blur event and while closing works, now clicking on a mention in the dropdown closes it too early. I had to add a setTimeout() before closing of at least 500ms... That might explain why the proposed solution of the author also involves at timer :-/

PVince81:
PR here https://github.com/nextcloud/spreed/pull/4978 with a timeout handler

PR #4978, opened 2021-01-20
Close at-mention dropdown on blur

PVince81:
When clicking outside the at-mention dropdown or exiting the input
field, the at-mention dropdown is now closed automatically.

Fixes https://github.com/nextcloud/spreed/issues/2758

Also see https://github.com/nextcloud/spreed/issues/2758#issuecomment-763488114

### DIFF
--- a/src/components/NewMessageForm/AdvancedInput/AdvancedInput.vue
+++ b/src/components/NewMessageForm/AdvancedInput/AdvancedInput.vue
@@ -85,6 +85,7 @@
 			@shortkey="focusInput"
 			@keydown.enter="handleKeydownEnter"
 			@keydown.esc.prevent="handleKeydownEsc"
+			@blur="onBlur"
 			@paste="onPaste" />
 	</At>
 </template>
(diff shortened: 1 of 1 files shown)


src/components/NewMessageForm/AdvancedInput/AdvancedInput.vue
```diff
@@ -249,6 +250,15 @@ export default {
 		EventBus.$off('focusChatInput', this.focusInput)
 	},
 	methods: {
+		onBlur() {
+			// requires a short delay to avoid blocking click event handlers
+			// from vue-at which also have some delay in place...
+			// a setTimeout was recommended by the library author here:
+			// https://github.com/fritx/vue-at/issues/114#issuecomment-565777450
+			setTimeout(() => {
```
nickvergessen, line 258: `this.$next()` could help/work?

PVince81: no, nextTick is the equivalent of setTimeout with a value of zero
I even tried with 50 and it was not high enough...

I saw in the source code of the lib that it has some timeouts already in place.

PVince81: and even the lib author suggested to do that, which surprised me at first... probably he knew about the timeouts issue...

ma12-co (approved):

PR #4979, opened 2021-01-20
Fix scrolling chat when in the background

nickvergessen:
Before scrollTo() was used, but this does not work when the window is not visible.
So now when the chat is in the background we scroll only half a line and stop following.
The user will see that something happened and have the anchor to jump to the bottom.

Same author (half a line) | New author (half avatar)
---|---
Bildschirmfoto von 2021-01-20 16-35-17 | Bildschirmfoto von 2021-01-20 16-33-34

cc @ma12-co @jancborchardt  The main question is if the "scroll only half" behaviour is actually wanted, or if we should just scrolldown fully (as it was attempted to, but didn't do because scrollTo() only works on visible windows)

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -588,11 +588,24 @@ export default {
 		 */
 		smoothScrollToBottom() {
 			this.$nextTick(function() {
-				this.scroller.scrollTo({
-					top: this.scroller.scrollHeight,
-					behavior: 'smooth',
-					 })
-				this.setChatScrolledToBottom(true)
+				if (this.$store.getters.windowIsVisible()) {
+					// scrollTo is used when the user is watching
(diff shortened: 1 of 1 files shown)


src/components/MessagesList/MessagesList.vue
```diff
@@ -588,11 +588,24 @@ export default {
 		 */
 		smoothScrollToBottom() {
 			this.$nextTick(function() {
-				this.scroller.scrollTo({
-					top: this.scroller.scrollHeight,
-					behavior: 'smooth',
-					 })
-				this.setChatScrolledToBottom(true)
```
nickvergessen, line 595: I think this is actually what the biggest problem was. It removed the scroll-to-bottom button, although it wasn't scrolled to the bottom

ma12-co (approved):

PR #4988, opened 2021-01-21
Fix "Undefined index: userId"

nickvergessen:
Regression from the participant rewrite in  #4324


PVince81 (approved):
:+1:

PVince81 (approved):
:+1: makes sense

Issue #4983, opened 2021-01-20
[11.x] Lobby start time cannot be removed


wiswedel:
Until Talk 10.x, the start time for calls with lobby was optional. A once set start time could be removed from the input field and what remained was the help text that the start time is optional:

![image](https://user-images.githubusercontent.com/1804221/105214685-9d547000-5b50-11eb-8d8f-67873e38dc10.png)

That is no longer possible with 11.x: Once the checkbox "enable lobby" is checked, the start time is pre-filled with "now".
You neither see the hint that the start time is optional after all, nor can you remove the pre-filled time. The UI simply immediately puts it back into the input field if you try deleting it from there.

nickvergessen:
On 20 the displaying always failed

wiswedel:
> On 20 the displaying always failed

not on my test system on 20.0.4 with Talk 10.0.5 🤔

PVince81:
I went back in time to the moment where this feature was merged and at that time it was possible to clear the lobby field value. I did remember that it worked back then.

So maybe there were recent changes to the date picker field that disallowed empty values.

PVince81:
Fix is here https://github.com/nextcloud/spreed/pull/4990 with some bonus UX adjustments.

PR #4990, opened 2021-01-21
Fix lobby time input

PVince81:
Fixes https://github.com/nextcloud/spreed/issues/4983

- Bug: Lobby timer can now be cleared. When doing so it is saved directly.
- UX: Adjusted minute selection to have an interval of 5.
- UX: Adjusted default date to be the start of the next hour.
- UX: Removed extra click for confirmation as we anyway have the save button.
- Internal: Changed the data type for communicating with date picker to use timestamps, since this is what we store anyway.

### DIFF
--- a/src/components/ConversationSettings/LobbySettings.vue
+++ b/src/components/ConversationSettings/LobbySettings.vue
@@ -49,25 +49,17 @@
 					<div>
 						<label for="moderation_settings_lobby_timer_field">{{ t('spreed', 'Meeting start time') }}</label>
 					</div>
-					<div>
-						<DatetimePicker
-							id="moderation_settings_lobby_timer_field"
-							aria-describedby="moderation_settings_lobby_timer_hint"
-							:value="lobbyTimer"
-							:default-value="defaultLobbyTimer"
-							:placeholder="t('spreed', 'Start time (optional)')"
-							:disabled="lobbyTimerFieldDisabled"
-							type="datetime"
-							:input-class="['mx-input', { focusable: !lobbyTimerFieldDisabled }]"
-							v-bind="dateTimePickerAttrs"
-							@change="setNewLobbyTimer" />
-						<button
-							id="moderation_settings_lobby_timer_submit"
-							:aria-label="t('spreed', 'Save meeting start time')"
-							:disabled="lobbyTimerFieldDisabled"
-							type="submit"
-							class="icon icon-confirm-fade" />
-					</div>
+					<DatetimePicker
+						id="moderation_settings_lobby_timer_field"
+						aria-describedby="moderation_settings_lobby_timer_hint"
+						:value="lobbyTimer"
+						:default-value="defaultLobbyTimer"
+						:placeholder="t('spreed', 'Start time (optional)')"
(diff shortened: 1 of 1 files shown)


src/components/ConversationSettings/LobbySettings.vue
```diff
@@ -54,6 +54,7 @@
 							id="moderation_settings_lobby_timer_field"
 							aria-describedby="moderation_settings_lobby_timer_hint"
 							:value="lobbyTimer"
+							:defaultValue="defaultLobbyTimer"
```
nickvergessen, line 57: Attribute ':defaultValue' must be hyphenated

```suggestion
							:default-value="defaultLobbyTimer"
```

wiswedel:
Thanks for taking care, @PVince81. I'm not qualified to approve or equipped to test the code, so I'll leave that to @nickvergessen.

nickvergessen:
- [ ] Time selector does not close anymore when selecting a minute option
- [ ] When you select a time and click "anywhere" (empty space in the conversation settings modal) the popup will close, but you don't know whether you have to press the arrow button or not. It looks the same like when you close the time picker with clicking on the arrow

nickvergessen:
I don't think it's related, but:
* When the conversation settings open you can scroll the complete window on the background
* When you trigger a notification scrolling goes totally wild
* After a view seconds of the notification all is fixed again
Peek 2021-01-21 16-18

PVince81:
When the date picker was in the menu, you had to click on an "OK" button (confirm: true), and that would trigger a save.

Now that we have a separate save button, you have to click that to close the popup.
Otherwise closing the popup when selecting the minute is not a good idea as the person selecting the date/time might not be done

If all this is confusing I could remove the save button and bring the back the "OK in popup for saving". When clearing it anyway saves automatically now. Would that help to improve UX ?

nickvergessen:
> Otherwise closing the popup when selecting the minute is not a good idea as the person selecting the date/time might not be done

I ment we should close the date/time picker when you click on a minute and then make the previous (should be) disable save button non-disabled and primary (to make it obvious you need to save now).

But since this is so weird, I would prefer to just have the OK button on the time-picker again. Felt much more natural and well after selecting the time, you are done... there is mostly nothing you want to do more in the date-time selector

PVince81:
> But since this is so weird, I would prefer to just have the OK button on the time-picker again. Felt much more natural and well after selecting the time, you are done... there is mostly nothing you want to do more in the date-time selector

Yeah, let's try that

PVince81:
I don't see any scrollbars in regular mode. Probably the dyslexia font that is pushing stuff further than expected, same with https://github.com/nextcloud/spreed/issues/4993

PVince81:
@nickvergessen please re-review, now you need to click "Ok" in the popup for saving.

PVince81:
@nickvergessen I tried dark mode + dyslexia font with various window widths but could not reproduce the scrolling problem when the notification is visible. Would need further research separately to find out what other conditions are involved.

nickvergessen (approved):

nickvergessen:
Behaviour is a bit weird when another moderator changes things in parallel.
But okay for now

Issue #4982, opened 2021-01-20
Make lobby being active more visible to moderators


wiswedel:
**Is your feature request related to a problem? Please describe.**
When the lobby is active, there is a full-fledged sentence in the chat protocol that the lobby is active, but moderators don't see the fact that their guests are waiting outside very prominent.
Hence they would be waiting in an empty call until they either remember to disable the lobby or the set time has been reached anyway.

**Describe the solution you'd like**
It would be good if the moderator would be informed more prominently that the lobby is still active even though the moderator is already present or has even started the call.

wiswedel:
see also https://github.com/nextcloud/spreed/issues/4981

Issue #4981, opened 2021-01-20
Make lobby disabling more accessible


wiswedel:
**Is your feature request related to a problem? Please describe.**
When a moderator has activated the lobby with a meeting start date, the joining limitation for guests automatically drops once the start time is reached.
However, it is appropriate to start the meeting a few minutes early because for example the guests are already assembled in the lobby.

In order to disable that lobby manually, the moderator currently has to go to the ... menu, then to the conversation settings, scroll down and uncheck the lobby setting.

**Describe the solution you'd like**
A prominent button in the sidebar to disable the lobby with one click and open the doors for the guests waiting outside

wiswedel:
see also https://github.com/nextcloud/spreed/issues/4982

PVince81:
Are we ok to implement the proposal or do we want to explore other design alternatives ?

Possible alternative: have a "disable lobby" button in the three dots menu, only present if lobby as enabled.

Possible use case: the moderators already started a call in which they discuss preparations and they might want to disable the lobby without stopping the call. Having the button either in the right sidebar or top bar menu would work.

Thoughts ? @nextcloud/designers-talk

nickvergessen:
There should be a direct "Disable lobby" button in the sidebar under the title/description if lobby is enabled.

wiswedel:
> There should be a direct "Disable lobby" button in the sidebar under the title/description if lobby is enabled.

👍 that'll also close https://github.com/nextcloud/spreed/issues/4982

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/4997

PR #4997, opened 2021-01-22
New button to disable lobby in sidebar

PVince81:
image

Fixes https://github.com/nextcloud/spreed/issues/4981
Fixes https://github.com/nextcloud/spreed/issues/4982

Note: I tried adding more information like the remaining time until meeting start, but it would take too much space and make the sidebar look too crowded. Also the moment library doesn't generate nice compact human readable texts for remaining durations. So if we do want this: needs further research.

### DIFF
--- a/src/components/RightSidebar/LobbyStatus.vue
+++ b/src/components/RightSidebar/LobbyStatus.vue
@@ -1,0 +1,65 @@
+<!--
+  - @copyright Copyright (c) 2020 Vincent Petry <vincent@nextcloud.com>
+  -
+  - @author Vincent Petry <vincent@nextcloud.com>
+  -
+  - @license GNU AGPL version 3 or any later version
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+-->
+
+<template>
+	<button @click="disableLobby">
+		{{ t('spreed', 'Disable lobby' ) }}
+	</button>
+</template>
+
+<script>
+import { showError, showSuccess } from '@nextcloud/dialogs'
+
+export default {
+	name: 'LobbyStatus',
+
+	props: {
+		token: {
+			type: String,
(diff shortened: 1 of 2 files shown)


ma12-co:
I don't think this button should be there. We could create a new action button in the moderation menu when the lobby is activated. That would make it for a 2 click operation, wouldn't it be enough? If not I would say we add a circular unlabeled button beside the moderation menu (only when the lobby is enabled)

ma12-co (changes requested):
I don't think this button should be there. We could create a new action button in the moderation menu when the lobby is activated. That would make it for a 2 click operation, wouldn't it be enough? If not I would say we add a circular unlabeled button beside the moderation menu (only when the lobby is enabled)

PVince81:
@ma12-co we should also fulfil https://github.com/nextcloud/spreed/issues/4982 which requires a way to find out whether the lobby is enabled without extra clicks. The button being present would already be a sign.

nickvergessen:
> I don't think this button should be there.

Only the disable is there and that must be. It must be really prominent and is mostly the next thing to be done in meetings.

ma12-co:
What about 
> circular unlabeled button beside the moderation menu (only when the lobby is enabled)

PVince81:
> What about
> 
> > circular unlabeled button beside the moderation menu (only when the lobby is enabled)

unlabeled with what icon ?

nickvergessen (approved):
Works and good enough for now

ma12-co:
> unlabeled with what icon ?

Maybe message lock?

nickvergessen:
It needs to be super obvious. So if it's only an icon, it needs to be primary to make sure it's right into the eye.
But I think a labeled button is much better and more obvious.

ma12-co:
> it needs to be primary

I'd have suggested yellow (warning).

nickvergessen:
> > unlabeled with what icon ?
> 
> Maybe message lock?

Or the lobby button, so it's obvious that it's the lobby...
https://github.com/nextcloud/spreed/blob/master/img/lobby.svg

PVince81:
but clicking a button that looks like a bell doesn't tell me what the button does intuitively, so would need a tooltip to make it clear it's about disabling it, not enabling

nickvergessen:
Well so as you can see in the ticket people have the sidebar open and check the participant list in such cases. So it's not arbitrary really.

ma12-co:
> but clicking a button that looks like a bell doesn't tell me what the button does intuitively, so would need a tooltip to make it clear it's about disabling it, not enabling

Agree on that a bell doesn't work

wiswedel (approved):
Great implementation. Does exactly what it promises (and the wording refers 1:1 to the "enable" mechanism 👍).

PVince81:
Let's move forward for now. We can iterate on the design later on when having another look bird's eye vue at the lobby feature.

nickvergessen:
Icon in question is `RoomService` https://material.io/resources/icons/?search=service&icon=room_service&style=baseline

PR #4998, opened 2021-01-22
Fix description icon sizes

ma12-co:

### DIFF
--- a/css/icons.scss
+++ b/css/icons.scss
@@ -113,6 +113,11 @@
 			filter: initial;
 		}
 	}
+
+	// Patch material design icons' span wrong height
+	.material-design-icon {
+		display: contents;
+	}
 }
 
 // The atwho panel is a direct child of the body, so it is not affected by
--- a/src/assets/buttons.scss
+++ b/src/assets/buttons.scss
@@ -32,6 +32,7 @@
 	display: flex;
 	align-items: center;
 	justify-content: center;
+	margin: 0;
 	&:not(.primary) {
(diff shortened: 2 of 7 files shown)


PVince81:
seems the icons were not vertically centered on the buttons before:
image

and now when smaller it makes it slightly more obvious:
image

ma12-co:
Fixed!

css/icons.scss
```diff
@@ -113,6 +113,11 @@
 			filter: initial;
 		}
 	}
+
+	// Patch material design icons' span wrong height
+	.material-design-icon {
+		display: contents;
```
PVince81, line 119: risky change, not sure if this is intended (in call sidebar):

before:

image

after:

image

PVince81: oh, apparently it also reduces the emoji icon's side in the regular messages view, so it's not about narrow panels

PVince81:
the regression is now gone

but now the buttons are too close to that blue bottom bar:
image

what is that bar anyway ? maybe the border of the field, we could try and hide it

PVince81:
image

css/icons.scss
ma12-co: This I fixed

ma12-co:
Margins fixed, ready to review again!

PVince81:
the blue bar is still there, but as it was there before I guess it's intentional.
the distance from the bar looks fine now:
image

image

PVince81 (approved):
:+1:

PR #5015, opened 2021-01-25
Make background black again

nickvergessen:

### DIFF
--- a/src/assets/variables.scss
+++ b/src/assets/variables.scss
@@ -70,6 +70,7 @@ $message-form-max-height: 180px;
 
 $top-bar-height: 60px;
 
+$color-call-background: #444444;
 $color-guests-avatar: #ffffff;
 $color-background-guests-avatar: #b9b9b9;
 
--- a/src/components/CallView/CallView.vue
+++ b/src/components/CallView/CallView.vue
@@ -620,13 +620,13 @@ export default {
 	width: 100%;
 	height: 100%;
 	overflow: hidden;
-	background-color: black;
+	background-color: $color-call-background;
 }
 
 #call-container {
 	width: 100%;
 	height: 100%;
-	background-color: #000;
+	background-color: $color-call-background;
 }
 
 #videos {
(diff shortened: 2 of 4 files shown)


PVince81:
Why ? Screenshots before/after ?

As far as I remember this explicit background was needed for the stripe to not look shitty in some scenarios, perhaps when there are less participants. Maybe recent changes removed the need to have it that way ?

nickvergessen:
> Why?

Because the call screen should be "dark" 

It was introduced by "Add ability to collapse the grid view" (without discussion or comment):
https://github.com/nextcloud/spreed/commit/b885bddae6426e2a841d62f8e909cf6ed0b1e3c3

If this is colored at all, we could use `#444444` like in the empty call view:
https://github.com/nextcloud/spreed/blob/732200a0586dc6df39e29755b15103fe21a5c3e9/src/components/CallView/shared/EmptyCallView.vue#L165

But the current gray is way to light and feels totally out of place


> Screenshots before/after ?

Before | After
---|---
Bildschirmfoto von 2021-01-26 08-52-05 | Bildschirmfoto von 2021-01-26 08-51-57

Before | After
---|---
Bildschirmfoto von 2021-01-26 08-51-38 | Bildschirmfoto von 2021-01-26 08-51-50

PVince81:
in the first set of screenshots the grey background was added to make it clearer that it's a stripe, I think it was a request from @ma12-co 
in grid mode it indeed looks better with black

@ma12-co are you fine with the change above ?

nickvergessen:
Added screenshots of `#444444` above as well

ma12-co:
> I think it was a request from @ma12-co
 
No no :)

I agree that the gray is too light. Out of the 3 screenshots I prefer #444444

nickvergessen:
My only problem with `#444444` is that it's the background of the "promoted placeholder" just without the icon + shadow

ma12-co:
Please change the bkg to the whole callview (first few lines of css in callview.vue)

nickvergessen:
Fixed, ready again

PVince81 (approved):
:+1:

PR #5017, opened 2021-01-26
Add hover effects for grid collapser

### LINKED ISSUE #5016: Arrow to hide the video stream on the button not visible on white background
The arrow to hide the video streams on the button:

image

Is not visible on white background. Which happens quite often when people share slides:

image

Maybe put a shadow around it? Or turn it to black on bright backgrounds (if this is possible)?

nickvergessen:
Before | After
---|---
Peek 2021-01-26 10-56 | Peek 2021-01-26 10-54

The main problem is when the button is on e.g. white screenshare, see #5016  for screenshots

Fix #5016

### DIFF
--- a/src/components/CallView/Grid/Grid.vue
+++ b/src/components/CallView/Grid/Grid.vue
@@ -847,10 +847,16 @@ button.stripe--collapse {
 	border: 0;
 	background: none;
 	opacity: .7;
+	padding: 0;
 
-	&:hover,
-	&:focus {
-		opacity: 1;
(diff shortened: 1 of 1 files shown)


ma12-co (changes requested):
Could you try and apply the new classes? 
.nc-button and.nc-button__main

nickvergessen:
Doesn't work too good (no hover when the callview is hovered, etc).

Let's get this in and look at a proper class later

ma12-co (approved):
Ok

PR #5019, opened 2021-01-26
Allow other apps to register message actions

ma12-co:
How to test: 

* Open the developer console and type:
```
window.OCA.Talk.registerMessageAction({
	label: 'test',
	callback: (messageAPIData) => console.log(messageAPIData),
	icon: 'icon-user',
})
```
* Open a message 3 dot menu;

* Click on the added action;

* Check if the response object is logged in the console

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -119,6 +119,16 @@ the main body of the message as well as a quote.
 						@click.stop="handleReply">
 						{{ t('spreed', 'Reply') }}
 					</ActionButton>
+					<template
+						v-for="action in messageActions">
+						<ActionButton
+							:key="action.label"
+							:icon="action.icon"
+							:close-after-click="true"
+							@click="action.callback(messageAPIData)">
+							{{ action.label }}
+						</ActionButton>
+					</template>
 				</Actions>
 			</div>
 		</div>
@@ -287,6 +297,10 @@ export default {
 	},
 
 	computed: {
+		messageObject() {
+			return this.$store.getters.message(this.token, this.id)
+		},
+
 		hasActions() {
 			return this.isReplyable && !this.isConversationReadOnly
 		},
@@ -430,6 +444,18 @@ export default {
 				return t('spreed', 'Not enough free space to upload file')
 			}
 			return t('spreed', 'You can not send messages to this conversation at the moment')
+		},
+
+		messageActions() {
+			return this.$store.getters.messageActions
+		},
+
+		messageAPIData() {
+			return {
+				message: this.messageObject,
+				metadata: this.conversation,
+				apiVersion: 'v3',
+			}
 		},
 
 	},
--- a/src/init.js
+++ b/src/init.js
@@ -1,0 +1,48 @@
+/**
+ * @copyright Copyright (c) 2020 John Molakvoæ <skjnldsv@protonmail.com>
+ *
+ * @author Marco Ambrosini <marcoambrosini@pm.me>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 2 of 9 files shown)


src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -119,6 +119,16 @@ the main body of the message as well as a quote.
 						@click.stop="handleReply">
 						{{ t('spreed', 'Reply') }}
 					</ActionButton>
+					<template
+						v-for="action in messageActions">
+						<ActionButton
+							:key="action.label"
+							icon="action.icon"
```
nickvergessen, line 126:
```suggestion
							:icon="action.icon"
```

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -119,6 +119,16 @@ the main body of the message as well as a quote.
 						@click.stop="handleReply">
 						{{ t('spreed', 'Reply') }}
 					</ActionButton>
+					<template
+						v-for="action in messageActions">
+						<ActionButton
+							:key="action.label"
+							:icon="action.icon"
+							:close-after-click="true"
+							@click="action.callback">
```
nickvergessen, line 128:
```suggestion
							@click="action.callback(messageAPIData)">
```

So here we need to handover a public API message object (which we can not change frequently.

I would for the same of simplicity tend to just hand over the full object from the store so https://nextcloud-talk.readthedocs.io/en/latest/chat/#receive-chat-messages-of-a-conversation is the matching documentation

ma12-co: oh sure

juliushaertl: In addition to the message data, details about the conversation (either in the message itself or as a separate object) would probably also be good to have at hand as the integrating app.

src/main.js
```diff
@@ -154,3 +154,15 @@ Sidebar.prototype.close = function() {
 Object.assign(window.OCA.Files, {
 	Sidebar: new Sidebar(),
 })
+
+if (!window.OCA.Talk) {
+	window.OCA.Talk = {}
+}
+window.OCA.Talk.registerMessageAction = (label, callback, icon) => {
```
juliushaertl, line 161: JSdoc annotations for this and especially for the callback parameters would be nice

nickvergessen: Do we also need a "works on this message" call back? so we can have apps that allow interacting with e.g. system messages, old messages, etc. and others that don't (similar to deletion only working 6 hours)

src/components/MessagesList/MessagesGroup/Message/Message.vue
ma12-co: What sort of info? Would the conversation token be enough?

src/main.js
ma12-co: added comments!

src/components/MessagesList/MessagesGroup/Message/Message.vue
juliushaertl: token and title would be ideal i'd say for being able to immediately show some context as the action and generating a link to it.

src/init.js
```diff
@@ -0,0 +1,48 @@
...
+}
+
+/**
+ * Frontend message API for adding actions to talk messages.
+ * @param {*} Object the wrapping object.
+ * @param {String} label the action label.
+ * @param {Function} callback the callback function. This function will receive
+ * the messageAPIData object as a parameter and be triggered by a click on the
+ * action.
+ * @param {String} icon the action label. E.g. "icon-reply"
```
nickvergessen, line 39: one thing to keep in mind is that this icon needs to be globally loaded, right?
Because it shows on shares in the sidebar, files app sidebar, talk, ....

nickvergessen:
So I used:

```js
window.OCA.Talk.registerMessageAction({
	label: 'test',
	callback: (messageAPIData) => console.log(messageAPIData),
	icon: 'icon-reply',
})
```

but the icon is still black on dark mode. I guess there is some specialty required to stop the auto-invert of something as it might happen twice (action inverts it and the img is actually already inverted)

ma12-co:
> but the icon is still black on dark mode. I guess there is some specialty required to stop the auto-invert of something as it might happen twice (action inverts it and the img is actually already inverted)

hmm, this is a weird one, this happens with the original reply button too. And it seems to disappear only on hover. Anyhow seems it's not related to this pr as it's reproducible on master

Peek 2021-02-01 16-15

src/init.js
juliushaertl: Yep, should be fine though since those integrating apps load their scripts anyways for that.


Though for that is there already an event like LoadAdditionalScripts we have the legacy event for collaboration resources (`\OCP\Collaboration\Resources::loadAdditionalScript`)

ma12-co: updated the docs for this

nickvergessen (approved):

juliushaertl:
Tested and works with me WIP implementation :+1:

Issue #4138, opened 2020-09-11
abort infinite upload for guest users without storage


call-me-matt:
## Steps to reproduce
1. connect as a guest user
2. upload a picture to the chatroom

### Expected behaviour
The picture should be [shown to everybody](https://github.com/nextcloud/spreed/issues/3343) or alternatively the upload should be cancelled with an error message. Or, even better, the attachment icon [should be hidden](https://github.com/nextcloud/spreed/issues/3430) if no write permission is granted to the guest.

### Actual behaviour
An error message is displayed (briefly), but the uploaded picture is also shown in the chat. To the guest it is not obvious that the shared picture is only visible to her/him and why it disappears with a refresh of the page.

## Talk app

**Talk app version:** (see apps admin page: `/index.php/settings/apps`) I'm only a guest, but should be the latest and greatest.

PVince81:
ok, I was confused about the term "guest". so this is not a conversation accessed by public link but as user from the [guest app](https://github.com/nextcloud/guests)

PVince81:
I just tried it locally and can confirm the error.

So what we could do is have an initial Webdav call to the configured attachment folder (see my note in https://github.com/nextcloud/spreed/issues/3343#issuecomment-717368948 about allowing to configure it, will default to "/" for now), and if that one is writable (and has space left, more than 0), then enable all the upload operations.

PVince81:
with Webdav I was mostly thinking about frontend only.

maybe there's also a way to bundle that information through any of the other calls that might come from the Talk API

Tasks:
- [ ] add error verification in case the storage became unavailable/blocked/forbidden while uploading, display an error message and discard the temporary message
- [ ] get and store quota and perms of attachments folder upon loading the app
- [ ] if perms does not allow uploading, remove the upload components from view (don't render)
- [ ] if quota does not allow uploading, display the upload components disabled with a tooltip saying "not enough space" or alike
- [ ] after uploading a bunch of files, re-query the attachment folder's quota and perms
- [ ] if the attachment folder was changed in the settings, re-query as well

PVince81:
I thought about a simpler approach: detect the guest user in the capabilities manager and return "attachments.allowed = false" in case of a guest user.

But then it came to my mind that while a guest app user cannot upload into their "/Talk" folder, they could always reconfigure the attachment folder to point at a writable received share. A friendly user could share some of their personal space for them to be able to post attachments.

[...]

PR #5036, opened 2021-01-27
Hide upload button with zero quota

PVince81:
Whenever a user has zero quota, for example like guest app users, don't
display the upload button.

The quota value is retrieved for the attachment folder.

Fixes https://github.com/nextcloud/spreed/issues/4138

### DIFF
--- a/lib/TInitialState.php
+++ b/lib/TInitialState.php
@@ -98,17 +98,23 @@ trait TInitialState {
 		);
 
 		$attachmentFolder = $this->talkConfig->getAttachmentFolder($user->getUID());
+		$freeSpace = 0;
 
 		if ($attachmentFolder) {
 			try {
 				$userFolder = $rootFolder->getUserFolder($user->getUID());
 
-				if (!$userFolder->nodeExists($attachmentFolder)) {
-					$userFolder->newFolder($attachmentFolder);
+				try {
+					if (!$userFolder->nodeExists($attachmentFolder)) {
+						$userFolder->newFolder($attachmentFolder);
+					}
+
+					$freeSpace = $userFolder->get($attachmentFolder)->getFreeSpace();
+				} catch (NotPermittedException $e) {
+					$attachmentFolder = '/';
+					$this->serverConfig->setUserValue($user->getUID(), 'spreed', 'attachment_folder', '/');
+					$freeSpace = $userFolder->getFreeSpace();
 				}
-			} catch (NotPermittedException $e) {
-				$attachmentFolder = '/';
-				$this->serverConfig->setUserValue($user->getUID(), 'spreed', 'attachment_folder', '/');
 			} catch (NoUserException $e) {
 			}
 		}
@@ -116,6 +122,11 @@ trait TInitialState {
 		$this->initialState->provideInitialState(
 			'attachment_folder',
 			$attachmentFolder
+		);
+
+		$this->initialState->provideInitialState(
+			'attachment_folder_free_space',
+			$freeSpace
 		);
 
 		$this->initialState->provideInitialState(
@@ -144,6 +155,11 @@ trait TInitialState {
 
 		$this->initialState->provideInitialState(
 			'attachment_folder',
+			''
+		);
+
+		$this->initialState->provideInitialState(
+			'attachment_folder_free_space',
 			''
 		);
 
--- a/src/components/NewMessageForm/NewMessageForm.vue
+++ b/src/components/NewMessageForm/NewMessageForm.vue
@@ -45,14 +45,14 @@
 						:aria-label="t('spreed', 'Share files to the conversation')"
 						:aria-haspopup="true">
 						<ActionButton
-							v-if="canShareAndUploadFiles"
+							v-if="canUploadFiles"
 							:close-after-click="true"
 							icon="icon-upload"
 							@click.prevent="clickImportInput">
 							{{ t('spreed', 'Upload new files') }}
 						</ActionButton>
 						<ActionButton
-							v-if="canShareAndUploadFiles"
+							v-if="canShareFiles"
 							:close-after-click="true"
 							icon="icon-folder"
 							@click.prevent="handleFileShare">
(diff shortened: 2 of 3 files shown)


lib/TInitialState.php
```diff
@@ -98,17 +98,23 @@ protected function publishInitialStateForUser(IUser $user, IRootFolder $rootFold
...
 		$attachmentFolder = $this->talkConfig->getAttachmentFolder($user->getUID());
+		$freeSpace = 0;
 
 		if ($attachmentFolder) {
 			try {
 				$userFolder = $rootFolder->getUserFolder($user->getUID());
 
-				if (!$userFolder->nodeExists($attachmentFolder)) {
-					$userFolder->newFolder($attachmentFolder);
+				try {
```
nickvergessen, line 107: is this expensive (query and loadwise) because if so we might just want to do it for guest app users and assume a properly configured normal case

PVince81: Not sure if expensive, usually I don't think so. This is also queried on every propfind in the files app.

"normal case" can vary: there are use cases where regular users are created with zero quota outside of the guests app (for those who don't want to use the app). In the latter use case users could also be told to only work on an external storage that is mounted for everyone.

SystemKeeper:
Shouldn’t this distinguish between upload and share? If a normal user is created without quota, he‘s not allowed to upload something, but he could still share something that was shared with him. This PR would hide both options..?

PVince81:
> Shouldn’t this distinguish between upload and share? If a normal user is created without quota, he‘s not allowed to upload something, but he could still share something that was shared with him. This PR would hide both options..?

I have tested this scenario as follows:

1. Admin shares a folder "pics" with guest user guest1 and allows everything, even reshares
2. Admin uploads a few pics into the "pics" directory
3. Admin starts a conversation and invites Guest1
4. Guest1 logs in and accesses the conversation
5. Guest1 selects "Share" from the attachment menu and pics one of the pictures out of the "pics" directory
:boom: 

The request fails with a 500 error that returns a page:
image

So it seems this is explicitly blocked by the backend.
If we want this as a feature, need to look into it separately and also examine potential side effects as I don't fully understand what else the backend is doing, like creating shares.

SystemKeeper:
I've tested this yesterday with @Ivansss and it worked just fine. He shared a folder with me, I uploaded a picture and in talk shared in our conversation. So only difference was, that I uploaded the picture, but I don't see why it should behave differently there.

PVince81:
> I've tested this yesterday with @Ivansss and it worked just fine. He shared a folder with me, I uploaded a picture and in talk shared in our conversation. So only difference was, that I uploaded the picture, but I don't see why it should behave differently there.

@SystemKeeper were you logged in as a guest user from the "guests" app or as a regular user ?
Guest users from the "guests" app are blocked by the API.

PVince81:
Alright, I think I see your point. You could be a regular user (not guest app user) but have zero quota.

Then the API would allow it.

This means we need another mechanism to detect if the current user is a guest app user.
So far I haven't found any in the frontend.

SystemKeeper:
@PVince81 I'm a guest user on cloud.nextcloud.com

SystemKeeper:
> Alright, I think I see your point. You could be a regular user (not guest app user) but have zero quota.

That was my main concern, yes!

PVince81:
I need to retest everything (including changing attachment folder as guest) because I had an incomplete local setup. I had set the guest whitelist of apps to "files,spreed" and apparently it breaks shares.

PVince81:
Okay, with the correct whitelist (default one + spreed) sharing from a received folder works fine.

I've adjusted the code to distinguish between "upload perm" and "share perm".

src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -45,14 +45,14 @@
 						:aria-label="t('spreed', 'Share files to the conversation')"
 						:aria-haspopup="true">
 						<ActionButton
-							v-if="canShareAndUploadFiles"
+							v-if="canUploadFiles"
```
nickvergessen, line 48: Should the Actions have `v-if="canUploadFiles && canShareFiles"` ?

src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -45,14 +45,14 @@
...
 							:close-after-click="true"
 							icon="icon-upload"
 							@click.prevent="clickImportInput">
 							{{ t('spreed', 'Upload new files') }}
 						</ActionButton>
 						<ActionButton
-							v-if="canShareAndUploadFiles"
+							v-if="canShareFiles"
 							:close-after-click="true"
 							icon="icon-folder"
```
nickvergessen, line 57: this gets hover feedback when the mouse is in the chat view, while the other other one does not
Peek 2021-01-29 10-25

ma12-co: There's this problem also with the topbar buttons, it was introduced recently. I don't think it's related to this pr

nickvergessen: ok

src/components/NewMessageForm/NewMessageForm.vue
PVince81: the canUploadFiles computed attr is already based on the same conditions like canShareFiles, so not needed

src/components/NewMessageForm/NewMessageForm.vue
PVince81: I'm seeing slightly different symptoms:

1. hovering outside there is no circle at all
2. hovering on the message list (when empty), the circle appears light grey
3. hovering on the button makes the circle dark grey

Anyway, raised as https://github.com/nextcloud/spreed/issues/5049

src/components/NewMessageForm/NewMessageForm.vue
PVince81: as discussed, I've now added an extra v-if on the div that contains Actions to avoid rendering it empty

nickvergessen (approved):

PR #5038, opened 2021-01-28
l10n: Add word "file"

Valdnet:


PVince81 (approved):
I purposefully left it out to keep it short.
But if you think it looks better that way, fine :+1:

Valdnet:
I understand. But then in a similar message, you should also not use "file" 😉.
Thanks for the merge.

Issue #3462, opened 2020-05-05
Cannot add a participant to a room if the user joined the call via public link


skjnldsv:
Veeeeeerry specific edge case :grin: 

## Steps to reproduce
1. Be user0 on a server
2. Create a room with share link enabled 
3. Login as user1
4. Open talk
5. See that the Room is NOT there
4. Access the call via share link while still being logged
5. As user0, try to add user1 to the room

### Expected behaviour
You can add it to the room

### Actual behaviour
You can't add it to the room as it appear in the participant list, while *technically* not being part of the room *official* user list.

PVince81:
I assume that the wanted solution here is that if a moderator is adding a user while they are "USER_SELF_JOINED" we simply convert them to "USER" so they become a "true" participant.

nickvergessen:
Yeah

PVince81:
I just tried the steps and looking at the UI I don't know how as user0 I would come to the idea of trying to add user1 since I already see that user in the participant list.

And even if we allowed this, it means tweaking the UI to make it visually clear that this user1 is not "fully joined", so that user0 could decide to add the user permanently.

nickvergessen:
This comes from our instance where people open public conversations and ask to be added, but you as a moderator have to kick them before you can search and add them.
I guess it would also be fine to allow searching them and clicking them would add them. so basically 
https://github.com/nextcloud/spreed/blob/459858a69527311bf60ffa553d45ba2a7eeec032/lib/Chat/AutoComplete/SearchPlugin.php#L103-L110
would need to ignore `USER_SELF_JOINED` so they are still returned as options

PVince81:
Also an interesting effect in the current implementation is that if you add the group of the self-joined users, there is no error but they don't get permanently added either.

So this will need handling as well.

nickvergessen:
https://github.com/nextcloud/spreed/blob/6a019ed418f86185fb19ca520dc576f128758fcb/lib/Controller/RoomController.php#L1250-L1252

I guess `$this->participantService->getParticipantUserIds(` should be replaced with `getParticipantsForRoom` then so we can check the type and run an update.

PVince81:
Draft PR here https://github.com/nextcloud/spreed/pull/5048

PR #5048, opened 2021-01-29
Allow converting self-joined users participants to regular participants

PVince81:
Fixes https://github.com/nextcloud/spreed/issues/3462

- [x] adjust unit tests
- [x] TEST: adding directly
- [x] TEST: adding through group
- [x] TEST: adding group with empty "to add" doesn't fail

### DIFF
--- a/lib/Controller/RoomController.php
+++ b/lib/Controller/RoomController.php
@@ -1303,19 +1303,21 @@ class RoomController extends AEnvironmentAwareController {
 
 		// attempt adding the listed users to the room
 		// existing users with USER_SELF_JOINED will get converted to regular USER participants
-		foreach ($participantsToAdd as $participantToAdd) {
+		foreach ($participantsToAdd as $index => $participantToAdd) {
 			$existingParticipant = $participantsByUserId[$participantToAdd['actorId']] ?? null;
 
 			if ($existingParticipant !== null) {
+				unset($participantsToAdd[$index]);
 				if ($existingParticipant->getAttendee()->getParticipantType() !== Participant::USER_SELF_JOINED) {
 					// user is already a regular participant, skip
 					continue;
 				}
 				$this->participantService->updateParticipantType($this->room, $existingParticipant, Participant::USER);
-			} else {
-				$this->participantService->addUsers($this->room, [$participantToAdd]);
 			}
 		}
+
+		// add the remaining users in batch
+		$this->participantService->addUsers($this->room, $participantsToAdd);
 
 		return new DataResponse([]);
 	}
--- a/lib/Controller/RoomController.php
+++ b/lib/Controller/RoomController.php
@@ -1222,23 +1222,28 @@ class RoomController extends AEnvironmentAwareController {
 			return new DataResponse([], Http::STATUS_BAD_REQUEST);
 		}
 
-		$participants = $this->participantService->getParticipantUserIds($this->room);
+		$participants = $this->participantService->getParticipantsForRoom($this->room);
+		$participantsByUserId = [];
+		foreach ($participants as $participant) {
+			if ($participant->getAttendee()->getActorType() === Attendee::ACTOR_USERS) {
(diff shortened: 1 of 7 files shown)


lib/Controller/RoomController.php
```diff
@@ -1222,23 +1222,28 @@ public function addParticipantToRoom(string $newParticipant, string $source = 'u
...
 			if (!$newUser instanceof IUser) {
 				return new DataResponse([], Http::STATUS_NOT_FOUND);
 			}
 
-			if (\in_array($newParticipant, $participants, true)) {
-				return new DataResponse([]);
-			}
-
-			$this->participantService->addUsers($this->room, [[
+			$participantsToAdd[] = [
```
PVince81, line 1243: am thinking to make this a simple user id array and adding the actorType only at the end

nickvergessen: But not all participantsToAdd are users? There are already emails for guests, and we will have groups and circles tracking at some point too.

lib/Controller/RoomController.php
```diff
@@ -1316,7 +1301,23 @@ public function addParticipantToRoom(string $newParticipant, string $source = 'u
...
+			$existingParticipant = $participantsByUserId[$participantToAdd['actorId']] ?? null;
+
+			if ($existingParticipant !== null) {
+				if ($existingParticipant->getAttendee()->getParticipantType() !== Participant::USER_SELF_JOINED) {
+					// user is already a regular participant, skip
+					continue;
+				}
+				$this->participantService->updateParticipantType($this->room, $existingParticipant, Participant::USER);
+			} else {
+				$this->participantService->addUsers($this->room, [$participantToAdd]);
```
nickvergessen, line 1316: can we still group them? otherwise this will trigger too many events when adding a group with a lot of users.

lib/Controller/RoomController.php
PVince81: > But not all participantsToAdd are users?
In the current implementation it is, I simply refactored the code to have the adding logic at the end in a common location to avoid duplicating the existence check. Please check further below.

One exception indeed is the email logic, but that one is not adding any user and has its own block

lib/Controller/RoomController.php
PVince81: ah! indeed, I was assuming the function was simply looping over each entry on its own.
I missed the part about events, thanks for pointing out.

I'll adjust the impl to do the "addUsers" in a single batch.

Hmmm, then maybe we should do the same for updating the participant types as it has its own event.

PVince81: Feels a bit overkill to create a new ModifyParticipantsEvent (plural) for a corner case.

I guess I'll only batch the `addUsers` one and leave the other be.

lib/Controller/RoomController.php
```diff
@@ -1316,7 +1301,25 @@ public function addParticipantToRoom(string $newParticipant, string $source = 'u
...
 		}
 
-		return new DataResponse();
+		// attempt adding the listed users to the room
+		// existing users with USER_SELF_JOINED will get converted to regular USER participants
+		foreach ($participantsToAdd as $index => $participantToAdd) {
+			$existingParticipant = $participantsByUserId[$participantToAdd['actorId']] ?? null;
+
+			if ($existingParticipant !== null) {
+				unset($participantsToAdd[$index]);
```
PVince81, line 1310: potential frown here ? I didn't want to make yet another copy of the array... and apparently PHP allows for that, so yay!

lib/Controller/RoomController.php
nickvergessen: > I'll only batch the addUsers one and leave the other be.

yes

PVince81:
tests added and batching issue solved for addUsers, but left alone for the rare promotion case

nickvergessen (approved):

nickvergessen (approved):

PR #5052, opened 2021-01-29
Allow resending invitation emails

PVince81:
- [x] resend to all
- [x] option in three dots menu to resend to an individual email guest
- [x] BUG: PIN not regenerated after enabling SIP for a conversation, so the resent email does not contain a PIN either

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -344,6 +344,15 @@ return [
 			],
 		],
 		[
+			'name' => 'Room#resendEmails',
+			'url' => '/api/{apiVersion}/room/{token}/participants/emails',
+			'verb' => 'POST',
+			'requirements' => [
+				'apiVersion' => 'v(1|2|3)',
+				'token' => '^[a-z0-9]{4,30}$',
+			],
+		],
+		[
 			'name' => 'Room#leaveRoom',
 			'url' => '/api/{apiVersion}/room/{token}/participants/active',
 			'verb' => 'DELETE',
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -348,7 +348,7 @@ return [
 			'url' => '/api/{apiVersion}/room/{token}/participants/emails',
 			'verb' => 'POST',
 			'requirements' => [
-				'apiVersion' => 'v(1|2|3)',
+				'apiVersion' => 'v3',
 				'token' => '^[a-z0-9]{4,30}$',
 			],
 		],
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -344,8 +344,8 @@ return [
 			],
 		],
 		[
-			'name' => 'Room#resendEmails',
-			'url' => '/api/{apiVersion}/room/{token}/participants/emails',
+			'name' => 'Room#resendInvitations',
+			'url' => '/api/{apiVersion}/room/{token}/participants/resend-invitations',
 			'verb' => 'POST',
 			'requirements' => [
 				'apiVersion' => 'v3',
--- a/docs/participant.md
+++ b/docs/participant.md
@@ -172,6 +172,12 @@
 
 * Method: `POST`
 * Endpoint: `/room/{token}/participants/emails`
+* Data:
+
+    field | type | Description
+    ------|------|------------
+    `participant` | string or null | v3 | User for whom to resend email, or null to send to all email actors
+
 * Response:
     - Status code:
         + `200 OK`
--- a/docs/participant.md
+++ b/docs/participant.md
@@ -176,14 +176,13 @@
 
     field | type | Description
     ------|------|------------
-    `participant`| string or null | v3 | User for whom to resend invitations, or null to send to all email actors
     `attendeeId` | int or null | v3 | Attendee id can be used for guests and users
 
 * Response:
     - Status code:
         + `200 OK`
         + `403 Forbidden` When the current user is not a moderator or owner
-        + `404 Not Found` When the given participant was not found in the conversation
+        + `404 Not Found` When the given attendee was not found in the conversation
 
 ## Leave a conversation (not available for call and chat anymore)
 
--- a/docs/participant.md
+++ b/docs/participant.md
@@ -171,18 +171,18 @@
 ## Resend participant emails
 
 * Method: `POST`
-* Endpoint: `/room/{token}/participants/emails`
+* Endpoint: `/room/{token}/participants/resend-invitations`
 * Data:
 
     field | type | Description
     ------|------|------------
-    `participant` | string or null | v3 | User for whom to resend email, or null to send to all email actors
+    `participant` | string or null | v3 | User for whom to resend invitations, or null to send to all email actors
 
 * Response:
     - Status code:
         + `200 OK`
         + `403 Forbidden` When the current user is not a moderator or owner
-        + `404 Not Found` When the conversation could not be found
+        + `404 Not Found` When the given participant was not found in the conversation
 
 ## Leave a conversation (not available for call and chat anymore)
 
--- a/docs/participant.md
+++ b/docs/participant.md
@@ -176,7 +176,8 @@
 
     field | type | Description
     ------|------|------------
-    `participant` | string or null | v3 | User for whom to resend invitations, or null to send to all email actors
+    `participant`| string or null | v3 | User for whom to resend invitations, or null to send to all email actors
+    `attendeeId` | int or null | v3 | Attendee id can be used for guests and users
 
 * Response:
     - Status code:
--- a/docs/participant.md
+++ b/docs/participant.md
@@ -168,6 +168,16 @@
         `inCall` | int | Flags whether the conflicting session is in a potential call
         `lastPing` | int | Timestamp of the last ping of the conflicting session
 
+## Resend participant emails
+
+* Method: `POST`
+* Endpoint: `/room/{token}/participants/emails`
+* Response:
+    - Status code:
+        + `200 OK`
+        + `403 Forbidden` When the current user is not a moderator or owner
+        + `404 Not Found` When the conversation could not be found
+
 ## Leave a conversation (not available for call and chat anymore)
 
 * Method: `DELETE`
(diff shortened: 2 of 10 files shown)


lib/Controller/RoomController.php
```diff
@@ -1811,4 +1811,22 @@ public function setSIPEnabled(int $state): DataResponse {
...
+	/**
+	 * @NoAdminRequired
+	 * @RequireModeratorParticipant
+	 *
+	 * @return DataResponse
+	 */
+	public function resendEmails(): DataResponse {
+		$participants = $this->participantService->getParticipantsForRoom($this->room);
+		foreach ($participants as $participant) {
+			if ($participant->getAttendee()->getActorType() === Attendee::ACTOR_EMAILS) {
```
PVince81, line 1824: should I also check if participantType is GUEST or is this safe already ?

appinfo/routes.php
```diff
@@ -343,6 +343,15 @@
 				'token' => '^[a-z0-9]{4,30}$',
 			],
 		],
+		[
+			'name' => 'Room#resendEmails',
+			'url' => '/api/{apiVersion}/room/{token}/participants/emails',
```
nickvergessen, line 348:
```suggestion
			'name' => 'Room#resendInvitations',
			'url' => '/api/{apiVersion}/room/{token}/participants/resend-invitations',
```

Let's be prepared for non-email invites in the future

lib/Controller/RoomController.php
nickvergessen: participant type could be guest or guest_moderator

nickvergessen: So no, dont

src/components/ConversationSettings/LinkShareSettings.vue
```diff
@@ -88,6 +88,13 @@
 				<span class="icon icon-clippy" />{{ t('spreed', 'Copy conversation link') }}
 			</button>
 		</div>
+		<div class="app-settings-subsection" v-if="isSharedPublicly">
```
nickvergessen, line 91: > GitHub Actions / eslint node12.x
> 
> src/components/ConversationSettings/LinkShareSettings.vue#L91
> Attribute "v-if" should go before "class"

src/components/ConversationSettings/LinkShareSettings.vue
```diff
@@ -218,6 +226,17 @@ export default {
 			// workaround for https://github.com/Inndy/vue-clipboard2/issues/105
 			this.$refs.copyLinkButton.focus()
 		},
+
+		async handleResendInvitations() {
+			this.isSendingInvitations = true
+			try {
+				await this.$store.dispatch('resendEmailInvitations', { token: this.token })
+				showSuccess(t('spreed', 'Email invitations have been sent'))
```
nickvergessen, line 234:
```suggestion
				showSuccess(t('spreed', 'Email invitations sent'))
```

nickvergessen:
I guess it looks good. The question of course is how long it can take to send e.g. 100 messages 🤔 
So we should show a decent loading spinner

lib/Controller/RoomController.php
PVince81: indeed!

PVince81:
@nickvergessen adjusted and added a decent spinner...

As mentioned before, we should probably also add an action for resend in the three dots menu of participants.

PVince81:
Button:
image

Spinner while it's running:
image

Action in the participant list:
image

A notification will appear after success.

src/components/ConversationSettings/LinkShareSettings.vue
```diff
@@ -88,6 +88,14 @@
 				<span class="icon icon-clippy" />{{ t('spreed', 'Copy conversation link') }}
 			</button>
 		</div>
+		<div v-if="isSharedPublicly" class="app-settings-subsection">
+			<button
+				:disabled="isSendingInvitations"
+				@click.prevent="handleResendInvitations">
+				<span class="icon icon-mail" />{{ t('spreed', 'Resend email invitations') }}
```
ma12-co, line 95:
```suggestion
				<span class="icon icon-mail" />{{ t('spreed', 'Resend invitation e-mails') }}
```

PVince81:
wording adjusted

docs/participant.md
```diff
@@ -168,6 +168,23 @@
...
+## Resend participant emails
+
+* Method: `POST`
+* Endpoint: `/room/{token}/participants/resend-invitations`
+* Data:
+
+    field | type | Description
+    ------|------|------------
+    `participant`| string or null | v3 | User for whom to resend invitations, or null to send to all email actors
+    `attendeeId` | int or null | v3 | Attendee id can be used for guests and users
```
nickvergessen, line 180: so either we allow `actor id + type` or `attendee id`
but no mixing (just because older endpoints grew with legacy)

lib/Controller/RoomController.php
```diff
@@ -1811,4 +1811,41 @@ public function setSIPEnabled(int $state): DataResponse {
...
+					$participants[] = $this->room->getParticipant($participant);
+				}
+			} catch (ParticipantNotFoundException $e) {
+				return new DataResponse([], Http::STATUS_NOT_FOUND);
+			}
+		} else {
+			$participants = $this->participantService->getParticipantsForRoom($this->room);
+		}
+
+		\OCP\Util::writeLog('spreed', '#### participants: ' . $participants , \OCP\ILogger::DEBUG);
```
nickvergessen, line 1841: no way

docs/participant.md
PVince81: alright, I copied it from there. I'll limit it to attendee id then.

src/components/ConversationSettings/LinkShareSettings.vue
```diff
@@ -88,6 +88,14 @@
 				<span class="icon icon-clippy" />{{ t('spreed', 'Copy conversation link') }}
 			</button>
 		</div>
+		<div v-if="isSharedPublicly" class="app-settings-subsection">
+			<button
+				:disabled="isSendingInvitations"
+				@click.prevent="handleResendInvitations">
+				<span class="icon icon-mail" />{{ t('spreed', 'Resend invitation e-mails') }}
```
nickvergessen, line 95:
```suggestion
				<span class="icon icon-mail" />{{ t('spreed', 'Resend invitations') }}
```

src/components/ConversationSettings/LinkShareSettings.vue
```diff
@@ -218,15 +227,30 @@ export default {
 			// workaround for https://github.com/Inndy/vue-clipboard2/issues/105
 			this.$refs.copyLinkButton.focus()
 		},
+
+		async handleResendInvitations() {
+			this.isSendingInvitations = true
+			try {
+				await this.$store.dispatch('resendInvitations', { token: this.token })
+				showSuccess(t('spreed', 'Email invitations sent'))
```
nickvergessen, line 235:
```suggestion
				showSuccess(t('spreed', 'Invitations sent'))
```

src/components/ConversationSettings/LinkShareSettings.vue
```diff
@@ -218,15 +227,30 @@ export default {
...
+			} catch (e) {
+				showError(t('spreed', 'Error occurred when sending email invitations'))
```
nickvergessen, line 237:
```suggestion
				showError(t('spreed', 'Error occurred when sending invitations'))
```

PVince81:
@nickvergessen adjusted and retested.

src/services/participantsService.js
```diff
@@ -251,6 +251,19 @@ const setGuestUserName = async(token, userName) => {
...
+/**
+ * Resends email invitations for the given conversation.
+ * If no userId is set, send to all applicable participants.
+ *
+ * @param {string} token conversation token
+ * @param {int} attendeeId attendee id to target, or null for all
+ */
+const resendInvitations = async(token, { attendeeId = null }) => {
+	await axios.post(generateOcsUrl('apps/spreed/api/v3/room', 2) + token + '/participants/resend-invitations', {
+		attendeeId: attendeeId,
```
nickvergessen, line 263:
```suggestion
		attendeeId,
```

Should work even with the `null` case right?

src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
```diff
@@ -114,6 +114,13 @@
 				@click="promoteToModerator">
 				{{ t('spreed', 'Promote to moderator') }}
 			</ActionButton>
+			<ActionButton v-if="isEmailActor"
```
nickvergessen, line 117: But logged in users have pins too, when SIP is enabled. So also show the option in that case?

src/services/participantsService.js
PVince81: yeah, it's just a syntax thing

src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
PVince81: I did a test with users who have emails. If I add those to a SIP-enabled conversation they never receive an invitation email as they are not ACTOR_EMAILS.

Should we make the resend feature still send an email to those people who never received one in the first place ? This means moving the scope from ACTOR_EMAILS to all user actors who have an email address set.

nickvergessen: Well I think it would be nice, as they might not have their nextcloud log in at hand, but emails + dialin would be possible.
But well then lets keep it for now, as I forgot email addresses are optional 🙈

PVince81:
Adjusted. Hopefully final!

nickvergessen:
So from the discussions here and in the ticket, I would have looked for the option down in the SIP block, if that is where relevant information is changed:
Bildschirmfoto von 2021-02-01 14-18-27

Depending on the screen size you might not see it up there when toggling.
Maybe we should show it at the bottom too when you just enabled SIP?

PVince81:
@nickvergessen I had it at the bottom in my first iteration and even only displayed it when SIP was enabled.
I had decided to move it up because I realized that email participants also exist in non-SIP mode and one might want to resend also in that situation.

If you say SIP-only is the way to go I can move it back down.
Then I'd also remove the "resend invitation" for individual people if SIP is disabled.

Thoughts ?

PVince81:
> Maybe we should show it at the bottom too when you just enabled SIP?

Okay, I missed the "too".
So we could display it in both locations to make it easy to find. Makes sense.

nickvergessen:
No, I think it makes sense to have it in general.
But after turning on SIP, informing the moderator that resending might make sense sounds useful and like the number 1 usecase of it.

Issue #5021, opened 2021-01-26
Mute/unmute icon changes position on click


ma12-co:
![Peek 2021-01-26 12-26](https://user-images.githubusercontent.com/26852655/105844825-bedead00-5fd1-11eb-8166-86517d0a998d.gif)

Ivansss:
@ma12-co That's an issue in the SVG. I had the same issue in the iOS app.

nickvergessen:
Should just switch to the Material Icon after 21 is split of

PVince81:
draft PR https://github.com/nextcloud/spreed/pull/5058

PVince81:
I went further and replaced most of the icons from the call view with material design icons here: https://github.com/nextcloud/spreed/pull/5058

This was useful to prevent alignment/offset differences.

PR #5058, opened 2021-02-01
Replace call icons in buttons bar with material design icons

PVince81:
Fixes https://github.com/nextcloud/spreed/issues/5021

- [x] settings page
   - [x] replace
   - [x] test some of the "no audio available" cases
- [x] video bottom bar
   - [x] replace
   - [x] test
- [x] local media controls
   - [x] replace
   - [x] fix container alignment
- [ ] replace icons in screen share menu => move to actions menu as the icons currently don't align well, but switching to actions menu require support for material design icon as trigger button (TODO: raise)
- [x] "alert-circle" for ICE fail and other icon-error instances
- [x] ~~"signal" for connection speed problem~~ there's already a material design icon there
- [x] force mute in menu
- [x] test all in dark mode
- [x] test in speaker view
- [x] test in grid view
- [x] test in sidebar mode
- [x] test in video verification mode

### DIFF
--- a/css/icons.scss
+++ b/css/icons.scss
@@ -64,12 +64,6 @@
 		&.icon-grid-view {
 			background-image: url(icon-color-path('toggle-pictures', 'actions', 'fff', 1, true));
 		}
-		&.icon-audio {
-			background-image: url(icon-color-path('audio', 'actions', 'fff', 1, true));
-		}
-		&.icon-audio-off {
-			background-image: url(icon-color-path('audio-off', 'actions', 'fff', 1, true));
-		}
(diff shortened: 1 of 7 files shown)


PVince81:
seems that we don't need width and height on the buttons bar if we use only material design icons
when removing those the icon aligns itself properly in the button

so I'll try replacing the other icons as well for a consistent look

PVince81:
It looks like this now
image

The paddings might not perfectly match the old ones but I found that with the icons have thinner lines so the current padding matches better.

ma12-co (approved):

PR #5059, opened 2021-02-01
Bring up the "new group conv" dialog for group results

nickvergessen:
Before | After
---|---
Peek 2021-02-01 16-47 | Peek 2021-02-01 16-46

Should prevent some group conversations that are created by accident from time to time, when you think you are just searching for it.

* Prefills the conversation name with the search result label
* Selects the group/circle already as participant


jancborchardt (approved):
Awesome, perfect fix! :)

ma12-co (approved):
:clap:

PR #5060, opened 2021-02-01
fix spelling of "most likely" in multiple files

meonkeys:


nickvergessen (changes requested):
l10n/ Folder should not be modified. It's handled by transifex

meonkeys:
> l10n/ Folder should not be modified. It's handled by transifex

Ah, thank you @nickvergessen ! I'll remove `l10n/` changes from my PR.

I'm worried only changing `src/components/AdminSettings/TurnServers.vue` will break the linkage with the existing translations... or is this what you prefer?

I created a transifex account in case I need one.

nickvergessen (approved):

nickvergessen:
Yeah translations will break, but the old one is still there so they can copy it from there

PR #5061, opened 2021-02-02
Fix unwanted background changes in some buttons

### LINKED ISSUE #5049: Strange hover highlights
There is strange hover feedback on the "Share file to chat" option when the user has no quota and hovers the message list

ma12-co:
fix #5049

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -118,7 +118,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php-versions }}
-          extensions: mbstring, iconv, fileinfo, intl, mysql, pdo_mysql
+          extensions: mbstring, iconv, fileinfo, intl, mysql, pdo_mysql, gd, zip
           tools: phpunit:8.5.2
           coverage: none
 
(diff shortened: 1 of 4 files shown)


nickvergessen:
Screenshot  or description which buttons are affected?

nickvergessen:
Ah didnt see the linked issue

nickvergessen (changes requested):
Also affects the sidebar open/close handle. it's fine while not in a call. But I think in the call view we actually need the hover, to make the button visible  (not only when you know where to hover). Just imagine a white presentation or wall background (Ref https://github.com/nextcloud/spreed/issues/5016 )

ma12-co:
Done

Screenshot from 2021-02-03 15-22-32

PVince81:
Please rebase to get some master changes. This is how I tested it.

- Call view looks fine.
- Ghost hover is gone

Still, there's a hover inconsistency between the attachment button and the emoji picker button:
image
image

@ma12-co mind fixing that as well ?

ma12-co:
> Still, there's a hover inconsistency between the attachment button and the emoji picker button:

Fixed it here for now @PVince81

PVince81:
I just switched computers and now I'm on my "small screen with bigger fonts" and noticed this:
image

We should add a bit of padding here.

But as this is unrelated, let's do it separately.

PVince81 (approved):
:+1:

PR #5065, opened 2021-02-02
Fix forced active action looks

ma12-co:
Peek 2021-02-02 14-50

### DIFF
--- a/src/components/LeftSidebar/ConversationsList/Conversation.vue
+++ b/src/components/LeftSidebar/ConversationsList/Conversation.vue
@@ -415,6 +415,6 @@ export default {
 }
 
 .forced-active {
-	box-shadow: inset 4px 0 var(--color-primary-element);
+	background-color: var(--color-primary-light) !important //Overrides gray hover feedback;
 }
 </style>


nickvergessen:
While this fixes the forced-active, the hover is still "broken" for me?
You run vue master or something?

ma12-co:
https://github.com/nextcloud/nextcloud-vue/pull/1692

Nope not master, the fix for the hover is there ^

PVince81 (approved):
Tested, works. The blue thing is now visible when it wasn't before. :+1:

jancborchardt:
This should be in the Vue components, no @ma12-co? cc @skjnldsv

nickvergessen:
It's in https://github.com/nextcloud/nextcloud-vue/pull/1692

ma12-co:
I think @jancborchardt  is talking about the hacked buttons into a radio. 
We already have a ActionRadio component in the vue lib. Not sure why this approach has been adopted here in the first place. Maybe for the icons? 
In any event I think we could remove this and use the already existing `actionradio` component, which looks clearer IMO as it's impossible to confuse it with other `ActionButton`. 
Screenshot from 2021-02-22 13-10-16

PR #5072, opened 2021-02-03
[stable21] Fix find-listed integration test outline

PVince81:
Fix the test to actually use the variable from the example list.

### DIFF
--- a/tests/integration/features/conversation/find-listed.feature
+++ b/tests/integration/features/conversation/find-listed.feature
@@ -14,7 +14,7 @@ Feature: conversation/find-listed
       | roomName | public-room |
     When user "creator" allows listing room "group-room" for "none" with 200
     And user "creator" allows listing room "public-room" for "none" with 200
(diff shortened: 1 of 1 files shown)


nickvergessen:
> App "Talk" cannot be installed because it is not compatible with this version of the server.

Not based on https://github.com/nextcloud/spreed/pull/5067 ?

PVince81:
> > App "Talk" cannot be installed because it is not compatible with this version of the server.
> 
> Not based on #5067 ?

Strange...

I've rebased it.

Also: did we forget to replace CORE_BRANCH in drone to stable21 ?

PVince81:
I've pushed https://github.com/nextcloud/spreed/pull/5081 to set the core branch to stable21 as we did for stable20

ma12-co (approved):
:elephant:

PVince81:
Rebased to catch up with the CI fixes

PVince81:
another fix for CI is here: https://github.com/nextcloud/spreed/pull/5098

PR #5074, opened 2021-02-03
Improve deck card UI

ma12-co:
Peek 2021-02-03 13-46

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/DeckCard.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/DeckCard.vue
@@ -21,11 +21,21 @@
 
 <template>
 	<a
-		v-tooltip.bottom="deckLocation"
-		class="external"
+		class="deck-card"
 		:href="link"
+		:aria-label="deckCardAriaLabel"
 		target="_blank">
-		<strong>{{ name }}</strong>
+		<div class="deck-card__lineone">
+			<div class="icon-deck" />
+			<div class="title">
+				{{ name }}
+			</div>
+		</div>
+		<div class="deck-card__linetwo">
+			<div>
+				{{ deckLocation }}
+			</div>
+		</div>
 	</a>
 </template>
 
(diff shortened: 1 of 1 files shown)


src/components/MessagesList/MessagesGroup/Message/MessagePart/DeckCard.vue
```diff
@@ -73,6 +83,54 @@ export default {
...
 </script>
+
+<style lang="scss" scoped>
+.deck-card {
+	display: flex;
+	transition: box-shadow 0.1s ease-in-out;
+	box-shadow: 0 0 2px 0 var(--color-box-shadow);
+	border-radius: var(--border-radius-large);
+	font-size: 100%;
+	background-color: var(--color-main-background);
```
nickvergessen, line 100: Bildschirmfoto von 2021-02-03 16-19-51

Might be my personal preference, but the "cut out" look makes it feel a bit weird when hovering on dark mode?

nickvergessen: On on non-hover it looks just like it would be out of place position, as the boarder is hardly visible

ma12-co: addressed!

ma12-co:
ready to review again!

src/components/MessagesList/MessagesGroup/Message/MessagePart/DeckCard.vue
```diff
@@ -21,11 +21,21 @@
...
 	<a
-		v-tooltip.bottom="deckLocation"
-		class="external"
+		class="deck-card"
 		:href="link"
+		:aria-label="deckCardAriaLabel"
 		target="_blank">
-		<strong>{{ name }}</strong>
+		<div class="deck-card__lineone">
+			<div class="icon-deck" />
```
nickvergessen, line 29: icon is undefined when deck is disabled. Maybe we can copy and redefine it for that case?

nickvergessen: Also the icon is black on dark mode 🤷

ma12-co: > icon is undefined when deck is disabled. Maybe we can copy and redefine it for that case?

Silly question maybe, but if deck is unavailable what's the point of all this? Shouldn't we display a warning instead of the card?

nickvergessen: Well ignore that part

nickvergessen: Dark mode seems healed now 🤔

nickvergessen (approved):

Issue #4552, opened 2020-11-06
Error "Duplicate session" after deleting room when joined call


fancycode:
## Steps to reproduce
1. Create room and join it
2. Join call
3. Delete the room

### Expected behaviour
No error should be shown.

### Actual behaviour
An error about a duplicate session is shown. This doesn't happen if I don't join the call in step 2.

## Talk app

**Talk app version:** as installed on s.n.c

## Browser

**Browser name:** Chrome

@nickvergessen FYI

PVince81:
need to verify if https://github.com/nextcloud/spreed/pull/4593 (which is merged now) fixes this issue

PVince81:
still happening on master

to clarify: to delete the room I used a separate browser tab and did not join that second room while deleting.
then after a few seconds the room with the call would reload the page and say "duplicate session".

I wonder if maybe it's just a misinterpretation of the error code.

PVince81:
![image](https://user-images.githubusercontent.com/277525/106884140-33f06780-66e1-11eb-9f2d-191a0079b399.png)

maybe the attempt to leave the conversation returned an error

I'll dig further

PVince81:
Fix is here https://github.com/nextcloud/spreed/pull/5082

Issue #4670, opened 2020-11-25
Frequent redirects to not-found when switching between conversations


fritz-k:
Hi,

opening the same conversation again afterwards eventually leads to success. Roughly a quarter to a half of all requests get redirected to not-found.

## Steps to reproduce
1. Open an existing conversation

### Expected behaviour
The page shows the selected conversation.

### Actual behaviour
The page shows the selected conversation for a few seconds and then loads `index.php/apps/spreed/not-found` with the page stating that "The conversation does not exist".

## Talk app

**Talk app version:** 10.0.3

**Custom Signaling server configured:** no

**Custom TURN server configured:** no

**Custom STUN server configured:** no

## Browser

**Microphone available:** yes

**Camera available:** yes

**Operating system:** Linux; Ubuntu, Arch and NixOS

**Browser name:** Firefox & Chrom

**Browser version:** Various, including FF 82.0.2

### Browser log

<details>

```js
10:37:21.639 Current participant found CallButton.vue:101
10:37:22.293 GEThttps://nextcloud.ourdomain.de/index.php/avatar/user4/44?v=3
10:37:22.321 XHRPOSThttps://nextcloud.ourdomain.de/ocs/v2.php/apps/spreed/api/v2/room/kjvse3k9/participants/active
[HTTP/2 200 OK 402ms]

10:37:22.468 XHRGEThttps://nextcloud.ourdomain.de/ocs/v2.php/apps/spreed/api/v1/chat/kjvse3k9?lookIntoFuture=0&lastKnownMessageId=10212&includeLastKnown=1
[HTTP/2 200 OK 343ms]

10:37:22.471 XHRGEThttps://nextcloud.ourdomain.de/ocs/v2.php/apps/user_status/api/v1/statuses/user4
[HTTP/2 200 OK 364ms]

10:37:22.481 XHRGEThttps://nextcloud.ourdomain.de/ocs/v2.php/collaboration/resources/room/kjvse3k9?format=json
[HTTP/2 200 OK 303ms]

10:37:22.870 XHRGEThttps://nextcloud.ourdomain.de/ocs/v2.php/apps/spreed/api/v1/signaling/settings?token=kjvse3k9
[HTTP/2 200 OK 603ms]

10:37:23.550 GEThttps://nextcloud.ourdomain.de/index.php/avatar/user1/32
10:37:23.569 GEThttps://nextcloud.ourdomain.de/index.php/avatar/user4/32?v=3
[...]

Issue #1587, opened 2019-02-27
The same user can't join a room multiple times


fancycode:
While looking into the fix for #1586 reported by @danxuliu, I noticed a similar issue can happen with the default PHP backend if you join the same room with different browsers but the same user. While different browsers is probably not a real issue, using a mobile client and the browser could be a potential scenario.

## How to reproduce:
* Join room with user in browser A
* Join room with user in browser B
* Navigate away in browser A

### What happens:
After a couple of seconds, browser B shows the user as no longer in the room and also navigates away - from the room (to `/apps/spreed`) after some more seconds.

### What should happen instead:
Browser B should stay in the room / call.

nickvergessen:
This would require a major rework of the API and the database, as it means we need to split the participants from the sessions.

Also what should happen when a user joins twice? Do they connect to themselves too?

nickvergessen:
Also this means that all the participant handling needs to be done on user+session even for logged in users, because you might want to mute one of the sessions but not the other one.

nickvergessen:
How should 2 videos for the same user be displayed?

fancycode:
> Also what should happen when a user joins twice? Do they connect to themselves too?

Why not? You could call from your mobile to your desktop?

> How should 2 videos for the same user be displayed?

I don't get that. For subscribers it doesn't matter if it's the same user connected with two devices streaming the video or if it's two different users. On the publishing side, the two devices are also independent.

Of course a valid solution to this issue would also be to reject the behaviour completely and show the second session an error like "You are already connected to this room." when trying to connect. Right now the behaviour is undefined/confusing in that the one session leaving a room can kick out the other session of a user.

fancycode:
> Also this means that all the participant handling needs to be done on user+session even for logged in users, because you might want to mute one of the sessions but not the other one.

Yes, all that handling should only be done on sessions.

nickvergessen:
> Of course a valid solution to this issue would also be to reject the behaviour completely and show the second session an error like "You are already connected to this room." when trying to connect. Right now the behaviour is undefined/confusing in that the one session leaving a room can kick out the other session of a user.

The reason we do not do this is, that if your connection to the server is disconnected, and you reload, you can not join the call again for 90 seconds until your old session is timed out.

parafin:
[...]

PR #5082, opened 2021-02-04
Fix redirect when deleting current conversation

PVince81:
When the current conversation was deleted while in a call, the redirect
now targets the not-found page instead of the one about duplicate
session.

Added event bus handler for the not found page which now also removes
the session store's conversation token.

Fixes https://github.com/nextcloud/spreed/issues/4552

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -59,7 +59,7 @@ import {
 } from './utils/webrtc/index'
 import { emit } from '@nextcloud/event-bus'
 import browserCheck from './mixins/browserCheck'
-import duplicateSessionHandler from './mixins/duplicateSessionHandler'
+import sessionIssueHandler from './mixins/sessionIssueHandler'
 import isInCall from './mixins/isInCall'
 import participant from './mixins/participant'
 import talkHashCheck from './mixins/talkHashCheck'
@@ -85,7 +85,7 @@ export default {
 	mixins: [
 		browserCheck,
 		talkHashCheck,
-		duplicateSessionHandler,
+		sessionIssueHandler,
 		isInCall,
 		participant,
 	],
@@ -112,7 +112,7 @@ export default {
 		},
 
 		warnLeaving() {
-			return !this.isLeavingAfterSessionConflict && this.isInCall
+			return !this.isLeavingAfterSessionIssue && this.isInCall
 		},
 
 		/**
@@ -219,7 +219,7 @@ export default {
 				// We have to do this synchronously, because in unload and beforeunload
 				// Promises, async and await are prohibited.
 				signalingKill()
-				if (!this.isLeavingAfterSessionConflict) {
+				if (!this.isLeavingAfterSessionIssue) {
 					leaveConversationSync(this.token)
 				}
 			}
@@ -232,8 +232,9 @@ export default {
 					console.info('Conversations received, but the current conversation is not in the list, trying to get potential public conversation manually')
 					this.refreshCurrentConversation()
 				} else {
-					console.info('Conversation received, but the current conversation is not in the list. Redirecting to /apps/spreed')
-					this.$router.push('/apps/spreed/not-found')
+					console.info('Conversation received, but the current conversation is not in the list. Redirecting to not found page')
+					this.$router.push({ name: 'notfound', params: { skipLeaveWarning: true } })
+					this.$store.dispatch('updateToken', '')
 				}
 			}
 		})
@@ -287,7 +288,7 @@ export default {
 		 * Global before guard, this is called whenever a navigation is triggered.
 		*/
 		Router.beforeEach((to, from, next) => {
-			if (this.isInCall) {
+			if (this.warnLeaving && !to.params?.skipLeaveWarning) {
 				OC.dialogs.confirmDestructive(
 					t('spreed', 'Navigating away from the page will leave the call in {conversation}', {
 						conversation: this.getConversationName(this.token),
@@ -437,7 +438,8 @@ export default {
 				})
 			} catch (exception) {
 				console.info('Conversation received, but the current conversation is not in the list. Redirecting to /apps/spreed')
-				this.$router.push('/apps/spreed/not-found')
+				this.$router.push({ name: 'notfound', params: { skipLeaveWarning: true } })
+				this.$store.dispatch('updateToken', '')
 				this.$store.dispatch('hideSidebar')
 			} finally {
 				this.isRefreshingCurrentConversation = false
--- a/src/components/LeftSidebar/ConversationsList/Conversation.vue
+++ b/src/components/LeftSidebar/ConversationsList/Conversation.vue
@@ -84,11 +84,13 @@
 			<ActionSeparator />
 
 			<ActionButton v-if="canLeaveConversation"
+				:close-after-click="true"
 				:icon="iconLeaveConversation"
 				@click.prevent.exact="leaveConversation">
 				{{ t('spreed', 'Leave conversation') }}
 			</ActionButton>
 			<ActionButton v-if="canDeleteConversation"
+				:close-after-click="true"
 				icon="icon-delete-critical"
 				class="critical"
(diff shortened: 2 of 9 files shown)


src/App.vue
```diff
@@ -233,11 +234,23 @@ export default {
...
 
+		EventBus.$on('conversationNotFound', () => {
+			// this will skip showing a warning during a call and
+			// also skip auto-leaving the current conversation, which does not exist anyway
+			this.isLeavingAfterSessionConflict = true
+			SessionStorage.removeItem('joined_conversation')
+			this.$nextTick(() => {
+				// Need to delay until next tick, otherwise the PreventUnload is still being triggered
+				// Putting the window in front with the warning and irritating the user
+				window.location = generateUrl('/apps/spreed/not-found')
```
nickvergessen, line 250: Why this over `this.$router.push('/apps/spreed/not-found')` ?

nickvergessen: Also this doesn't seem to work. For me the "Leave call" popup is still shown (without HPB)

PVince81: ah, I copied this from mixins/duplicateSessionHandler which also doesn't use router push.
I'll check if it works with that.

PVince81: > Also this doesn't seem to work. For me the "Leave call" popup is still shown (without HPB)

It works if the deletion is done outside the current browser window as per original reproduction steps.

But indeed if you delete it within the same window you get the warnings. I'll try to add special handling for that scenario.

PVince81: I've tried using router.push() but it's triggering the "you are leaving call" dialog.
When adding a flag to inhibit it, it will display the not found page but with a lot of JS errors...

I don't think it's worth trying to protect against this at this stage (but worth as part of improving the not found page separately).

I'll leave window.location here then. I'm sure this must have been the reason why it was used also for the duplicate session case.

PVince81:
@nickvergessen I've pushed a fix to prevent leaving or deleting a conversation while in a call. This will prevent the can of worms of trying to handle warning dialogs, etc.

As a bonus I also added a "close that freaky popup already when I click".

Regarding router push, see my comment above. It's also a can of worms.

src/components/LeftSidebar/ConversationsList/Conversation.vue
```diff
@@ -84,11 +84,15 @@
 			<ActionSeparator />
 
 			<ActionButton v-if="canLeaveConversation"
+				:close-after-click="true"
+				:disabled="item.hasCall"
```
nickvergessen, line 88: * So I create a conversation
* Invite the full company
* Start a call
* No one can leave it
* muhahaha

Maybe only when you are part of the call?

src/App.vue
nickvergessen: Well signaling.js is not Vue, so it can't use router.push()
It smells so trappy to change this. There must be a reason we did it like this and not do a page reload all the time.

PVince81: > It smells so trappy to change this.

I'd appreciate a more diplomatic tone...

I'll have another look and compare with the original behavior and see why the errors I saw didn't happen there.

src/components/LeftSidebar/ConversationsList/Conversation.vue
PVince81: I knew something was fishy... Couldn't be that easy.

To detect "is in call" need to bring in two mixins :-/

I'll see if I can still solve this differently and inhibit the dialog while the redirect is in progress.

PVince81:
I've changed the approach. The new approach doesn't refactor the old code.

Since we'll need this kind of handling outside of App.vue, I've put the code into the mixin duplicateSessionHandler which I renamed to sessionIssueHandler.

Cases to cover and test:
- [ ] deleting session from outside
   - [x] regular call
   - [ ] sidebar call
   - [ ] video verification
   - [ ] guest call
- [ ] deleting from left sidebar while in the same call: todo, see https://github.com/nextcloud/spreed/pull/5082#discussion_r573885095

PVince81:
looks like we were redirecting to talk's "not found" page already in a sidebar call, despite being in the files app :-/
I'll make it the same now, we can try and find a cleaner solution for that later... maybe need to embed a "not found" in the sidebar tab and make sure it's recoverable

src/components/LeftSidebar/ConversationsList/Conversation.vue
PVince81: I can try moving this computed property to the store. Store getters are supposed to be like global computed properties and we've needed this one in enough places already.

PVince81: ok, as discussed this is a can of worms

PVince81:
it seems that when deleting the current conversation there's already another handler that sets the route because it reaches the sessionIssueHandler, this is what triggers that dialog.
I might need a second mechanism to inhibit the dialog in this code path.

PVince81:
Aha, finally found something: when setting the route to not-found it doesn't automatically clear the token, this is why I've seen a lot of JS errors because the right sidebar is still present with its chat tab.

PVince81:
After some research and tests I realized the following:
- we can push the "notfound" route instead of doing a full page reload:
   - the JS errors I saw before were only related to the fact that we forgot to clear the current token in some places
   - I've applied this now to this PR and it looks smooth when getting kicked out...
   - instead of a semi-global attribute for saying "please don't warn about leaving", we can use a route param when using named route, see https://router.vuejs.org/guide/essentials/navigation.html
- duplicate session handling must trigger a page reload to avoid side effects
   - in my local tests if I simply push the duplicate session route, it interferes with the other tab where I click "Join now" and that other tab gets a "not found"
   - there's probably something not cleant up properly. Considering that we'll [get rid of dupe sessions soonish](https://github.com/nextcloud/spreed/issues/1587) I think it's not worth the effort fixing this.

@nickvergessen please re-review

PVince81:
@nickvergessen any concerns against merging this ?

I've noticed this might tie in with my discovery in https://github.com/nextcloud/spreed/issues/4670#issuecomment-781474242 where we could also make the "go to not found" event only work if the token we're leaving is also the one on the page.
This could be achieved by passing the token from the 404 error to the event. Then the event handler compares the passed token with the one in store. If it's the same, redirect. Else: ignore.

nickvergessen:
It's a bit weird to be asked whether you want to leave the conversation. You have no chance to decide this. The conversation is gone, so it should fully terminate already, and not only show the correct "The conversation does not exist - Join a conversation or start a new one! " at the end. 
Bildschirmfoto von 2021-03-01 11-34-19

Left video (the moderator in the other tab) froze already.

For the moderator I see our custom "leave call" dialog for some split second:

Peek 2021-03-01 11-37

PVince81:
When I tested in chromium back then I didn't see the browser's native dialog, because now we're using the router, it should switch you to the "not found" page without a page refresh. Strange that it didn't work.

I'll recheck and test with Firefox then.

PVince81:
With the following I did not see any browser dialog:

1. Login as admin in Chromium
2. Create a conversation and invite user "alice"
3. Start a call
4. In Firefox login as "alice"
5. Join the conversation
6. Back in Chromium, open the left sidebar and delete the conversation

=> in the admin browser I land on the welcome page
=> in alice's browser I land on the "not found" page

Not sure if you used different steps or if it's a cache issue ?

nickvergessen (approved):

nickvergessen:
So I'm currently on the "Not found" page with Ivan in a call. We still hear each others as the webrtc stuff all didn't get canceled, but we can't change mic/camera/screenshare status anymore.

So I propose to go back to the redirect at least while being in a call.

PVince81:
> So I'm currently on the "Not found" page with Ivan in a call. We still hear each others as the webrtc stuff all didn't get canceled, but we can't change mic/camera/screenshare status anymore.
> 
> So I propose to go back to the redirect at least while being in a call.

PR here: https://github.com/nextcloud/spreed/pull/5595

PR #5084, opened 2021-02-05
Fix conversation switching lag

ma12-co:
Immediately display messages from the clicked conversation even if not joined yet with the signaling server.
In order to avoid errors, the call button and the newmessagegorm components are disabled while opening the new session.

TODO: 
- [x] Emoji button stays disabled;

https://user-images.githubusercontent.com/26852655/107543910-90f88b80-6bc1-11eb-8c58-16d7d43e6000.mp4

### DIFF
--- a/src/components/LeftSidebar/ConversationsList/Conversation.vue
+++ b/src/components/LeftSidebar/ConversationsList/Conversation.vue
@@ -115,7 +115,6 @@ import {
 } from '../../../services/conversationsService'
 import { generateUrl } from '@nextcloud/router'
 import { CONVERSATION, PARTICIPANT } from '../../../constants'
-import { EventBus } from '../../../services/EventBus'
 
 export default {
 	name: 'Conversation',
@@ -373,9 +372,6 @@ export default {
 
 		// forward click event
 		onClick(event) {
-			EventBus.$emit('conversationClicked', {
-				token: this.item.token,
-			})
 			this.$emit('click', event)
 		},
 	},
--- a/src/components/LeftSidebar/ConversationsList/Conversation.vue
+++ b/src/components/LeftSidebar/ConversationsList/Conversation.vue
@@ -115,6 +115,7 @@ import {
 } from '../../../services/conversationsService'
 import { generateUrl } from '@nextcloud/router'
 import { CONVERSATION, PARTICIPANT } from '../../../constants'
+import { EventBus } from '../../../services/EventBus'
 
 export default {
 	name: 'Conversation',
@@ -372,6 +373,9 @@ export default {
 
 		// forward click event
 		onClick(event) {
+			EventBus.$emit('conversationClicked', {
+				token: this.item.token,
+			})
 			this.$emit('click', event)
 		},
 	},
--- a/src/components/NewMessageForm/NewMessageForm.vue
+++ b/src/components/NewMessageForm/NewMessageForm.vue
@@ -64,7 +64,7 @@
 					<EmojiPicker @select="addEmoji">
 						<button
 							type="button"
-							disabled="disabled"
+							:disabled="disabled"
 							class="nc-button nc-button__main"
 							:aria-label="t('spreed', 'Add emoji')"
 							:aria-haspopup="true">
--- a/src/components/NewMessageForm/NewMessageForm.vue
+++ b/src/components/NewMessageForm/NewMessageForm.vue
@@ -40,6 +40,7 @@
 				<div
 					v-if="canUploadFiles || canShareFiles">
 					<Actions
+						ref="uploadMenu"
 						default-icon="icon-clip-add-file"
 						:aria-label="t('spreed', 'Share files to the conversation')"
 						:aria-haspopup="true">
@@ -59,11 +60,11 @@
 						</ActionButton>
 					</Actions>
 				</div>
-				<div
-					v-if="!isReadOnly">
+				<div>
 					<EmojiPicker @select="addEmoji">
 						<button
 							type="button"
+							disabled="disabled"
 							class="nc-button nc-button__main"
 							:aria-label="t('spreed', 'Add emoji')"
 							:aria-haspopup="true">
@@ -86,7 +87,7 @@
 						ref="advancedInput"
 						v-model="text"
 						:token="token"
-						:active-input="!isReadOnly"
+						:active-input="!disabled"
 						:placeholder-text="placeholderText"
 						:aria-label="placeholderText"
 						@update:contentEditable="contentEditableToParsed"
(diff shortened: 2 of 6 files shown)


nickvergessen:
You are not removing anything so we now pull new messages twice?

ma12-co:
> You are not removing anything so we now pull new messages twice?

What do you mean??

src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -175,12 +176,13 @@ export default {
 			}
 		},
 
-		isReadOnly() {
+		disabled() {
 			return this.conversation.readOnly === CONVERSATION.STATE.READ_ONLY
+			|| !this.currentConversationIsJoined
```
nickvergessen, line 181:
```suggestion
		disabledActions() {
			return this.conversation.readOnly === CONVERSATION.STATE.READ_ONLY
				|| !this.currentConversationIsJoined
```

Would rename the computed to reflect that only parts are disabled and not the complete component

src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -175,12 +176,13 @@ export default {
...
 		},
 
 		placeholderText() {
-			return this.isReadOnly
+			return this.disabled
```
nickvergessen, line 185: This is not correct. Now you replace the placeholder of the input with 'This conversation has been locked' while joining. So just copy the original value for this:
```suggestion
			return this.conversation.readOnly === CONVERSATION.STATE.READ_ONLY
```

src/utils/signaling.js
```diff
@@ -3,11 +3,13 @@
  * @copyright Copyright (c) 2019 Ivan Sein <ivan@nextcloud.com>
  * @copyright Copyright (c) 2019 Joachim Bauch <bauch@struktur.de>
  * @copyright Copyright (c) 2019 Joas Schilling <coding@schilljs.com>
+ * @copyright Copyright (c) 2020 Marco Ambrosini <marcoambrosini@pm.me>
```
nickvergessen, line 6: 🙄

src/components/NewMessageForm/NewMessageForm.vue
ma12-co: Aren't all the parts of the component (Actions + iIput field) disabled?

nickvergessen: Did you want to allow writing messages already and just not sending them?

ma12-co: Not at this stage we said

ma12-co:
Ping @nickvergessen

PVince81:
I like how joining feels much faster!

Now when there are new messages that haven't been loaded, you see the old state of the message list with "Joining conversation at the bottom". At this point, do we want to also display a placeholder below the existing messages in the place where new messages might appear ? There's a way to find out if there are new messages by checking that the lastMessage id is different. Not sure if that's a good idea though. :smile:

src/components/LeftSidebar/ConversationsList/Conversation.vue
```diff
@@ -372,6 +373,9 @@ export default {
 
 		// forward click event
 		onClick(event) {
+			EventBus.$emit('conversationClicked', {
```
PVince81, line 376: seems this isn't used anywhere, maybe it was part of a previous attempt.
remove ?

src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -207,6 +216,16 @@ export default {
...
 		},
+
+		currentConversationIsJoined() {
+			return this.$store.getters.currentConversationIsJoined
+		},
+	},
+
+	watch: {
+		disabled(newValue) {
+			this.$refs.uploadMenu.$refs.menuButton.disabled = newValue
```
PVince81, line 227: are we missing a "disabled" attribute in uploadMenu ?

let's add a "FIXME" here with a link to a ticket to do so.
otherwise this approach looks really ugly

src/init.js
```diff
@@ -46,3 +47,8 @@ window.OCA.Talk.registerMessageAction = ({ label, callback, icon }) => {
 	}
 	store.dispatch('addMessageAction', messageAction)
 }
+
+EventBus.$on('Signaling::joinRoom', (payload) => {
+	const token = payload[0]
+	store.dispatch('handleRoomJoined', token)
```
PVince81, line 53: I'm a bit worried that we have too many different "joined" states and the naming is going to become confusing.
Here I guess we are "fully joined" which is both the API based joined (`/chat/{token}/active`) and joined on signaling side.

I don't have any better idea currently, maybe we need to work on these confusions separately.

PVince81 (changes requested):

src/components/NewMessageForm/NewMessageForm.vue
nickvergessen: So actually the disable is missing on the Actions component.
Please upstream it to the vue lib directly.

src/init.js
ma12-co: We need these 2 joined states in order to speed up the UI... One is "UI joined" and the other one is real, "signaling joined"

ma12-co: Before there was only 1 state, so that the UI was waiting for the signaling requests to be done

src/store/tokenStore.js
```diff
@@ -80,6 +95,10 @@ const actions = {
 	updateTokenAndFileIdForToken(context, { newToken, newFileId }) {
 		context.commit('updateTokenAndFileIdForToken', { newToken, newFileId })
 	},
+
+	handleRoomJoined({ commit }, token) {
```
PVince81, line 99: why not then call it updateLastJoinedConversationToken ? then at least we have only one term to think about.

I like that the "lastJoinedConversationToken" is well-documented, so my brain can associate the dual-states to that, but having `handleRoomJoined' would mean yet another wording for the same thing. Hence my query for renaming it.

src/init.js
nickvergessen: It was not about the state, but about the names.

ma12-co:
Addressed the disabled state of action here https://github.com/nextcloud/nextcloud-vue/pull/1724

src/init.js
ma12-co: My bad, didn't get this right

src/store/tokenStore.js
ma12-co: addressed!

PVince81 (approved):
:+1:

PVince81:
follow up here to clean up some bits: https://github.com/nextcloud/spreed/pull/5266

PR #5103, opened 2021-02-09
Split reporting of candidates by protocol

nickvergessen:
### Before
1. Have a turn that only has working UDP (no TCP)
2. Configure it in admin settings for both
3. Use the test function

### Expected
Reported error for TCP

### Actually
Reported as working


PVince81 (approved):
Makes sense :+1:

PR #5105, opened 2021-02-09
Make sure that project resources scripts event is properly emitted

juliushaertl:
`user_uid` doesn't seem to be set automatically, so the event to dispatch for loading scripts for the projects integration is never emitted (in [tempates/index.php](https://github.com/nextcloud/spreed/blob/84d42604e60aa1c31ecf698dcc069f36c494148a/templates/index.php#L10))

I can't find where this broke currently so there might be a different way to solve this depending on what changed, however this at least makes it work again :)

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -44,6 +44,7 @@ use OCP\AppFramework\Http\Response;
 use OCP\AppFramework\Http\TemplateResponse;
 use OCP\AppFramework\Http\Template\PublicTemplateResponse;
 use OCP\AppFramework\Services\IInitialState;
+use OCP\EventDispatcher\GenericEvent;
 use OCP\EventDispatcher\IEventDispatcher;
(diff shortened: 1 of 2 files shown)


lib/Controller/PageController.php
```diff
@@ -255,7 +255,7 @@ public function index(string $token = '', string $callUser = '', string $passwor
 			$this->eventDispatcher->dispatchTyped(new LoadViewer());
 		}
 
-		$response = new TemplateResponse($this->appName, 'index');
+		$response = new TemplateResponse($this->appName, 'index', ['user_uid' => $user ? $user->getUID() : '']);
```
nickvergessen, line 258: Is this a typo and meant to be `user_id` in the template?

nickvergessen (approved):

juliushaertl:
:+1:

PVince81:
please rebase, the tests should pass now that they were fixed on master

Issue #5106, opened 2021-02-09
Register Deck action for sharing Deck with a conversation


PVince81:
### Description

- Main issue: https://github.com/nextcloud/deck/issues/2630
- [ ] REQUIRES: https://github.com/nextcloud/deck/pull/2772

Just like we already have for files, now also allow sharing Deck cards with a conversation.

### Implementation ideas

- [x] detect deck on the frontend side by detecting `window.OCA.Deck` namespace
- [x] register an action using `window.OCA.Deck.registerCardAction`
- [x] display a conversation selection menu (see https://github.com/nextcloud/spreed/blob/master/src/collections.js and views/RoomSelector)
- [x] send sharing action to backend API using https://nextcloud-talk.readthedocs.io/en/latest/chat/#share-a-rich-object-to-the-chat
   - data to send for a valid deck card: https://github.com/nextcloud/server/blob/078ecac188525490f7bc560c212070167249263b/lib/public/RichObjectStrings/Definitions.php#L247-L283

- what about unsharing ? => not for now

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/5110

PR #5110, opened 2021-02-09
Add sharing deck card with a conversation

PVince81:
Added deck action for sharing/linking that uses the RoomSelector.
Improved RoomSelector

Fixes https://github.com/nextcloud/spreed/issues/5106
Requires https://github.com/nextcloud/deck/pull/2772 (+ info.xml max-version hack) for testing.

Issues for later:
- adding conditional tooltip in case of ellipsis: would require having a reusable Conversation.vue in the future
- [x] the link inside the toast is not visually distinguishable, raised here: https://github.com/nextcloud/nextcloud-dialogs/issues/286
- [x] search field would be nice in RoomSelector: TODO raise, reusable room list component would be nice => raised https://github.com/nextcloud/spreed/issues/5117
- [x] ~~check for lobby state + moderator status to see if postable while filtering the list ?~~ => not now

Open topics:
- is the registration for the script in the right place or do we want another special event-thingy for deck to tell us when to register ? (like OCA\Files\Event\LoadSidebar)

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -38,6 +38,7 @@ use OCA\Talk\Config;
 use OCA\Talk\Dashboard\TalkWidget;
 use OCA\Talk\Events\ChatEvent;
 use OCA\Talk\Events\RoomEvent;
+use OCA\Talk\Deck\DeckPluginLoader;
 use OCA\Talk\Files\Listener as FilesListener;
 use OCA\Talk\Files\TemplateLoader as FilesTemplateLoader;
 use OCA\Talk\Flow\Operation;
@@ -99,6 +100,7 @@ class Application extends App implements IBootstrap {
 		$context->registerEventListener(BeforeTemplateRenderedEvent::class, PublicShareAuthTemplateLoader::class);
 		$context->registerEventListener(\OCP\AppFramework\Http\Events\BeforeTemplateRenderedEvent::class, UnifiedSearchCSSLoader::class);
 		$context->registerEventListener(UserChangedEvent::class, UserDisplayNameListener::class);
+		$context->registerEventListener(\OCP\AppFramework\Http\Events\BeforeTemplateRenderedEvent::class, DeckPluginLoader::class);
 
 		$context->registerSearchProvider(ConversationSearch::class);
 		$context->registerSearchProvider(CurrentMessageSearch::class);
--- a/src/deck.js
+++ b/src/deck.js
@@ -1,0 +1,78 @@
+/*
+ * @copyright Copyright (c) 2020 Vincent Petry <vincent@nextcloud.com>
+ *
+ * @author Vincent Petry <vincent@nextcloud.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+import Vue from 'vue'
+import { generateFilePath, generateUrl } from '@nextcloud/router'
+import { getRequestToken } from '@nextcloud/auth'
+import { translate, translatePlural } from '@nextcloud/l10n'
+import { showSuccess, showError } from '@nextcloud/dialogs'
+import { postRichObjectToConversation } from './services/messagesService'
+
+// CSP config for webpack dynamic chunk loading
+// eslint-disable-next-line
+__webpack_nonce__ = btoa(getRequestToken())
+
+// Correct the root of the app for chunk loading
+// OC.linkTo matches the apps folders
+// OC.generateUrl ensure the index.php (or not)
+// We do not want the index.php since we're loading files
+// eslint-disable-next-line
+__webpack_public_path__ = generateFilePath('spreed', '', 'js/')
+
+Vue.prototype.t = translate
+Vue.prototype.n = translatePlural
+Vue.prototype.OC = OC
+Vue.prototype.OCA = OCA
+
+document.addEventListener('DOMContentLoaded', function() {
+
+	if (!window.OCA.Deck) {
+		return
+	}
+
+	window.OCA.Deck.registerCardAction({
+		label: t('spreed', 'Post to a conversation'),
+		icon: 'icon-talk',
+		callback: (card) => {
+			OCP.Collaboration.trigger('room').then(async(token) => {
+				try {
+					const response = await postRichObjectToConversation(token, {
+						objectType: 'deck-card',
+						objectId: card.id,
+						metaData: JSON.stringify(card),
+					})
+					const messageId = response.data.ocs.data.id
+					const targetUrl = generateUrl('/call/{token}#message_{messageId}', { token, messageId })
+					showSuccess(t('spreed', 'Deck card has been posted to the selected <a href="{link}">conversation</a>.', {
+						link: targetUrl,
+					}), {
+						isHTML: true,
+					})
+				} catch (exception) {
+					console.error('Error posting deck card to conversation', exception, exception.response?.status)
+					showError(t('spreed', 'An error occurred while posting deck card to conversation.'))
+				}
+			})
+		},
+	})
+
+})
--- a/src/deck.js
+++ b/src/deck.js
@@ -26,53 +26,86 @@ import { getRequestToken } from '@nextcloud/auth'
 import { translate, translatePlural } from '@nextcloud/l10n'
 import { showSuccess, showError } from '@nextcloud/dialogs'
 import { postRichObjectToConversation } from './services/messagesService'
+import RoomSelector from './views/RoomSelector'
 
-// CSP config for webpack dynamic chunk loading
-// eslint-disable-next-line
-__webpack_nonce__ = btoa(getRequestToken())
-
-// Correct the root of the app for chunk loading
-// OC.linkTo matches the apps folders
-// OC.generateUrl ensure the index.php (or not)
-// We do not want the index.php since we're loading files
-// eslint-disable-next-line
-__webpack_public_path__ = generateFilePath('spreed', '', 'js/')
-
-Vue.prototype.t = translate
-Vue.prototype.n = translatePlural
-Vue.prototype.OC = OC
-Vue.prototype.OCA = OCA
-
-document.addEventListener('DOMContentLoaded', function() {
-
-	if (!window.OCA.Deck) {
-		return
+(function(OC, OCA, t, n) {
+	async function postCardToRoom(card, token) {
+		try {
+			const response = await postRichObjectToConversation(token, {
+				objectType: 'deck-card',
+				objectId: card.id,
+				metaData: JSON.stringify(card),
+			})
+			const messageId = response.data.ocs.data.id
+			const targetUrl = generateUrl('/call/{token}#message_{messageId}', { token, messageId })
+			showSuccess(t('spreed', 'Deck card has been posted to the selected <a href="{link}">conversation</a>.', {
+				link: targetUrl,
+			}), {
+				isHTML: true,
+			})
+		} catch (exception) {
+			console.error('Error posting deck card to conversation', exception, exception.response?.status)
+			if (exception.response?.status === 403) {
+				showError(t('spreed', 'No permission to post messages in this conversation'))
+			} else {
+				showError(t('spreed', 'An error occurred while posting deck card to conversation.'))
+			}
+		}
 	}
 
-	window.OCA.Deck.registerCardAction({
-		label: t('spreed', 'Post to a conversation'),
-		icon: 'icon-talk',
-		callback: (card) => {
-			OCP.Collaboration.trigger('room').then(async(token) => {
-				try {
-					const response = await postRichObjectToConversation(token, {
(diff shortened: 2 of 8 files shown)


nickvergessen:
> check for lobby state + moderator status to see if postable while filtering the list ?

Wouldn't care too much. Posting is blocked in the end, just show an error.

lib/AppInfo/Application.php
```diff
@@ -147,6 +147,7 @@ protected function registerCollaborationResourceProvider(IServerContainer $serve
 		$resourceManager->registerResourceProvider(ConversationProvider::class);
 		$server->getEventDispatcher()->addListener('\OCP\Collaboration\Resources::loadAdditionalScripts', function () {
 			\OCP\Util::addScript(self::APP_ID, 'collections');
+			\OCP\Util::addScript(self::APP_ID, 'deck');
```
nickvergessen, line 150: @juliushaertl is this the expected way to handle it?

src/deck.js
```diff
@@ -0,0 +1,82 @@
...
+Vue.prototype.OCA = OCA
+
+document.addEventListener('DOMContentLoaded', function() {
+
+	if (!window.OCA.Deck) {
+		return
+	}
+
+	window.OCA.Deck.registerCardAction({
+		label: t('spreed', 'Link to a conversation'),
```
nickvergessen, line 55:
```suggestion
		label: t('spreed', 'Post to a conversation'),
```

It's not linking. It's one time posting. Not revertable, no permissions added/modified, no collaboration resource linking done. Just cross posted link with name

src/deck.js
```diff
@@ -0,0 +1,82 @@
...
+		icon: 'icon-talk',
+		callback: (card) => {
+			OCP.Collaboration.trigger('room').then(async(token) => {
+				const tempId = 'temp-' + card.id + '-' + token + '-' + (new Date().getTime())
+				try {
+					const response = await axios.post(generateOcsUrl('apps/spreed/api/v1', 2) + `chat/${token}/share`, {
```
nickvergessen, line 61: move to chat service, as we are going to reuse this in the future.

src/deck.js
```diff
@@ -0,0 +1,82 @@
...
+						objectType: 'deck-card',
+						objectId: card.id,
+						metaData: JSON.stringify(card),
+						referenceId: Hex.stringify(SHA1(tempId)),
+					})
+					const messageId = response.data.ocs.data.id
+					const targetUrl = generateUrl('/call/{token}#message_{messageId}', { token, messageId })
+					showSuccess(t('spreed', 'Deck card has been linked with the selected <a href="{link}">conversation</a>.', {
```
nickvergessen, line 69: We don't have the conversation name, do we?

src/views/RoomSelector.vue
```diff
@@ -38,9 +38,16 @@
 							<span>{{ room.displayName }}</span>
 						</li>
 					</ul>
+					<div v-else-if="!loading">
+						{{ t('spreed', 'No available rooms found') }}
```
nickvergessen, line 42:
```suggestion
						{{ t('spreed', 'No conversations found') }}
```

src/deck.js
nickvergessen: `src/services/messagesService.js`

src/deck.js
nickvergessen:
```suggestion
					showSuccess(t('spreed', 'Deck card has been posted with the selected <a href="{link}">conversation</a>.', {
```

src/deck.js
```diff
@@ -0,0 +1,82 @@
...
+						link: targetUrl,
+					}), {
+						isHTML: true,
+					})
+				} catch (exception) {
+					console.error('Error sharing deck card to conversation', exception, exception.response?.status)
+					showError(t('spreed', 'An error occurred while sharing deck card to conversation.'))
```
nickvergessen, line 76: Here you say sharing, maybe that's even a better word to use than link/post

nickvergessen:
Looks good apart from the wordings

src/deck.js
PVince81: I hesitated a lot.
First I started with "sharing" because it was in the original mock up screenshot.
Then I saw "Link to conversation" in the hard-coded title of RoomSelector. At first believing that it was used elsewhere I simply went with that one. (note: we can't pass arguments to `trigger()` so have to settle for a window title globally)

Let's go with sharing then and I'll adjust the RoomSelector's title to also say so.
Or is that used elsewhere outside of Talk ?

src/deck.js
PVince81: > We don't have the conversation name, do we?

We only have the token. But we could replace it with the full conversation object.
However we can only afford to do this if we're sure that the RoomSelector is not used elsewhere outside of Talk, otherwise it would be an API change...

src/deck.js
PVince81: Meh... I'll go with "post", it sounds better.

Sharing makes it feel like you can unshare later.

nickvergessen (approved):

PVince81:
wait, the message service is not done yet

src/deck.js
PVince81: fixed

PVince81:
I've squashed our common changes + messagesService changes.

nickvergessen (dismissed):

lib/AppInfo/Application.php
juliushaertl: Is a bit to wide i'd say as we don't need this integration in files for example. For the complementary implementation of message actions i picked loading this in a BeforeTemplateRendered event: https://github.com/nextcloud/deck/pull/2795/files#diff-848ad1b5ff7f00ceb7ca05dc2abe8b592bb6711f3df7f01ae2b567508fe6906c

That is of course still somehow relying on the talk app route but felt cleaner then adding a dedicated event in the OCA namespace. But I have no strong oppinions on that.

src/deck.js
```diff
@@ -0,0 +1,78 @@
...
+
+	if (!window.OCA.Deck) {
+		return
+	}
+
+	window.OCA.Deck.registerCardAction({
+		label: t('spreed', 'Post to a conversation'),
+		icon: 'icon-talk',
+		callback: (card) => {
+			OCP.Collaboration.trigger('room').then(async(token) => {
```
juliushaertl, line 56: Not the most clean approach in my POV as those features should ideally be separated, but fine by me :see_no_evil: Works fine here of course as deck also registers for the Collaboration stuff already.

src/views/RoomSelector.vue
```diff
@@ -66,26 +74,13 @@ export default {
...
-			if (OCA.SpreedMe && OCA.SpreedMe.app.activeRoom) {
-				return OCA.SpreedMe.app.activeRoom.get('token')
-			}
-			return null
-		},
 		availableRooms() {
 			return this.rooms.filter((room) => {
-				return room.token !== this.currentRoom
-					&& room.type !== this.types.ROOM_TYPE_CHANGELOG
+				return room.readOnly !== CONVERSATION.STATE.READ_ONLY
```
juliushaertl, line 82: I assume this change is due to the reuse of the collaboration API, but this would allow to link a talk room to itself, where i'm not sure what happens when you try to do that in the backend.

PVince81: we don't use the RoomSelector within Talk, so not sure where else it's being used ?

but as you say it might be safer to also allow read-only rooms still

src/deck.js
PVince81: Why isn't it clean ? If we consider that room selection is a clean reusable dialog, then why not.
I do miss some way for configuring like passing in options.

lib/AppInfo/Application.php
PVince81: adjusted using a similar technique. See DeckLoaderPlugin.php

src/views/RoomSelector.vue
PVince81: adjusted to also allow listing read-only rooms

PVince81:
comments addressed, adjusted and squashed

lib/Listener/BeforeTemplateRenderedListener.php
```diff
@@ -0,0 +1,55 @@
...
+		if (!($event instanceof BeforeTemplateRenderedEvent)) {
+			return;
+		}
+
+		if (!$event->isLoggedIn()) {
+			return;
+		}
+
+		if (strpos($this->request->getPathInfo(), '/apps/deck') === 0) {
+			Util::addScript('deck', 'calendar');
```
nickvergessen, line 52: ???

src/views/RoomSelector.vue
```diff
@@ -66,26 +74,12 @@ export default {
...
 	computed: {
-		currentRoom() {
-			if (OCA.SpreedMe && OCA.SpreedMe.app.activeRoom) {
-				return OCA.SpreedMe.app.activeRoom.get('token')
-			}
-			return null
-		},
 		availableRooms() {
 			return this.rooms.filter((room) => {
-				return room.token !== this.currentRoom
```
nickvergessen, line 87: as per julius this is needed in the collaboration resource dialog. you shouldn't be able to bind the conversation to itself. Maybe add an option?

src/deck.js
nickvergessen: To me this is also a bit too much. Collaboration resources might have a different filtering or sorting or whatever. The function here is about a card action, not about `OCP.Collaboration`

lib/Listener/BeforeTemplateRenderedListener.php
PVince81: oops, I forgot to delete this file... the correct place is the DeckPluginLoader

src/deck.js
PVince81: Okay, I'll move this to a separate function, but will try to reuse/refactor common logic to avoid too much duplication.

src/views/RoomSelector.vue
PVince81: since we decided to move away from the collaboration dialog, this will become possible, yes

PVince81:
@nickvergessen see last commit, I've moved away from OCA.Collaboration and also made the dialog configurable.

Please review.

src/views/RoomSelector.vue
nickvergessen: so want to do this check in combination with `this.showPostableOnly` or need another way?

PVince81: I originally thought you were talking about the read-only case.

Now having a closer look at the mentioned line to answer: the "this.currentRoom" was based on an apparently old namespace "OCA.SpreedMe" which I didn't find elsewhere. So I assumed that this information was not available any more and is outdated. Since "this.currentRoom" relied on "OCA.SpeedMe" I removed it.

I could add it back and accept an optional room token but it this approach wouldn't be used anywhere. So maybe we keep my PR as is and add currentRoom back once the dialog is used in such way ?

nickvergessen: It is used in the collaboration resource feature where you can (without this check) link a conversation to itself. I guess it is broken since the vue rewrite

nickvergessen: The reason you can't find it atm is #5105

PVince81: If you can point me at a way to trigger it (if #5105 can bring it back), I can repair it to make it work again.

We might need to resort to similar hacks to detect the context, like if we're inside a conversation, because OCA.Collaboration.trigger() doesn't allow passing any arguments, so we can't set RoomSelector properties from there.

nickvergessen: #5105 is merged, so now:

1. Create a conversation
2. Open the sidebar in the conversation
3. Click on "Add to a project"
4. Click on "Link to a conversation"
5. See same conversation being offered

PVince81: Thanks, I'll try it out. Cool to finally see what this "add to project" is about.

Should the backend also prevent this "linking to self" ?

nickvergessen: > Should the backend also prevent this "linking to self" ?

It is (silently) ignoring it already
https://github.com/nextcloud/server/blob/5bf3d1bb384da56adbf205752be8f840aac3b0c5/core/Controller/CollaborationResourcesController.php#L135-L138

PVince81: eww, it seems this thing is even more buggy: you can link to another conversation or file multiple times
image

something to handle separately, seems that feature needs some love...

I'll now focus only on preventing the regression caused by my PR

nickvergessen: > eww, it seems this thing is even more buggy: you can link to another conversation or file multiple times

This is expected. because it could be different circles of people and projects. But anyway this is part of server and not Talk

PVince81:
Rebased and added commit for filtering out the current room.

To expose the current room I expose the full vue instance in `OC.Talk.instance` which I believe can be more useful in the future for components running on that page. Let me know if this is ok or if this is too much.

lib/Listener/BeforeTemplateRenderedListener.php
PVince81: annnd forgot again... deleting...

PVince81:
squashed, please re-review

nickvergessen:
Apps live in `OCA.` 🙈 

Also we are there already, so reuse (with same precausions) https://github.com/nextcloud/spreed/blob/a542d701637578296a85f40f2a8edc89c6d2b1e2/src/utils/webrtc/webrtc.js#L511-L514

PVince81:
> Apps live in OCA.

I did use OCA.Talk here: https://github.com/nextcloud/spreed/pull/5110/commits/c2c1fc1a51b44facbae88e6d1ce8b8d06dce6913#diff-27653c212e1cfe533e4eb2f7d0d3f89604c9de48a09583b4cbbbcbd08a07da79R160

I'll add a namespace check, just in case

PVince81:
extra check squashed into last commit

nickvergessen (approved):
Ready to merge, but I guess we wait for https://github.com/nextcloud/deck/pull/2772

PVince81:
rebased for CI

PVince81:
https://github.com/nextcloud/deck/pull/2772 was merged, merging as well...

is this something to backport ?

PR #5118, opened 2021-02-10
Show "Guest" instead of session hash for guests

### LINKED ISSUE #5109: Guests without displayname show actor ID in call view
## Steps to reproduce
1. Have a call
2. Join as a guest without setting a name ever
3. Others see you as `123456ACTOR_ID1243456` in the call view

nickvergessen:
Fix #5109 

The idea of this was to show the email address of invited SIP users, but in case of normal guests it's a bigger harm, so for now let's not do this.
In the future we might instead save the email address also as a display name for SIP invites.

### DIFF
--- a/src/components/CallView/shared/Video.vue
+++ b/src/components/CallView/shared/Video.vue
@@ -100,6 +100,7 @@ import AccountCircle from 'vue-material-design-icons/AccountCircle'
 import VideoBottomBar from './VideoBottomBar'
 import Screen from './Screen'
 import { EventBus } from '../../../services/EventBus'
+import { ATTENDEE } from '../../../constants'
 
 export default {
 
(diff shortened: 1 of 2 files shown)


src/components/CallView/shared/Video.vue
```diff
@@ -279,6 +279,9 @@ export default {
 
 			if (!participantName) {
 				participantName = this.peerData.displayName
+				if (!participantName && this.peerData.actorType === 'guests') {
```
PVince81, line 282: :bulb: Did you know: there is now a new constant `ATTENDEE.ACTOR_TYPE.GUESTS`

PVince81 (approved):
:+1: looks fine. Tested, works.

Seems sharing integration tests have failed

PVince81:
rebase needed to make tests pass

nickvergessen:
Yeah, also a potential rebase after #5119 (CallController touches two lines apart) so will wait until that one is merged

PR #5119, opened 2021-02-10
Store and use display name from attendees table

nickvergessen:
Could help a bit with https://github.com/nextcloud/server/pull/25440

Next step (for follow up PR) is to store the guest name also in this table and retain the attendee instead of the guests_name entry, so we can get rid of that table all together, but I don't want to backport that really.

### DIFF
--- a/lib/Controller/CallController.php
+++ b/lib/Controller/CallController.php
@@ -104,9 +104,13 @@ class CallController extends AEnvironmentAwareController {
 			if ($this->getAPIVersion() >= 3) {
 				$displayName = $participant->getAttendee()->getActorId();
 				if ($participant->getAttendee()->getActorType() === Attendee::ACTOR_USERS) {
-					$user = $this->userManager->get($participant->getAttendee()->getActorId());
-					if ($user instanceof IUser) {
-						$displayName = $user->getDisplayName();
+					if ($participant->getAttendee()->getDisplayName()) {
+						$displayName = $participant->getAttendee()->getDisplayName();
+					} else {
+						$user = $this->userManager->get($participant->getAttendee()->getActorId());
+						if ($user instanceof IUser) {
+							$displayName = $user->getDisplayName();
+						}
 					}
 				} else {
 					$displayName = $guestNames[$participant->getAttendee()->getActorId()] ?? $displayName;
--- a/lib/Controller/RoomController.php
+++ b/lib/Controller/RoomController.php
@@ -1161,11 +1161,6 @@ class RoomController extends AEnvironmentAwareController {
 
 			if ($participant->getAttendee()->getActorType() === Attendee::ACTOR_USERS) {
 				$userId = $participant->getAttendee()->getActorId();
-				$user = $this->userManager->get($userId);
-				if (!$user instanceof IUser) {
-					continue;
-				}
-
 				if ($result['lastPing'] > 0 && $result['lastPing'] <= $maxPingAge) {
 					$this->participantService->leaveRoomAsSession($this->room, $participant);
 				}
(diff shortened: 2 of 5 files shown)


lib/Controller/CallController.php
```diff
@@ -104,9 +104,13 @@ public function getPeersForCall(): DataResponse {
 			if ($this->getAPIVersion() >= 3) {
 				$displayName = $participant->getAttendee()->getActorId();
 				if ($participant->getAttendee()->getActorType() === Attendee::ACTOR_USERS) {
-					$user = $this->userManager->get($participant->getAttendee()->getActorId());
-					if ($user instanceof IUser) {
-						$displayName = $user->getDisplayName();
+					if ($participant->getAttendee()->getDisplayName()) {
```
nickvergessen, line 107:
```
cat /var/log/apache2/access.log.1 | grep "GET /ocs/v2.php/apps/spreed/api/v3/call/" | grep "09/Feb/2021:14" | wc -l
430```
Should save 430*40 = 16k  queries from the company call :D

File: lib/Controller/RoomController.php [hunk 2]
```diff
@@ -1161,19 +1161,21 @@ public function getParticipants(bool $includeStatus = false): DataResponse {
...
 				if ($result['lastPing'] > 0 && $result['lastPing'] <= $maxPingAge) {
 					$this->participantService->leaveRoomAsSession($this->room, $participant);
 				}
 
 				if ($this->getAPIVersion() < 3) {
 					$result['userId'] = $participant->getAttendee()->getActorId();
 				}
-				$result['displayName'] = (string) $user->getDisplayName();
+				$result['displayName'] = $participant->getAttendee()->getDisplayName();
+				if (!$result['displayName']) {
```
PVince81, line 1172: this is true even for the string "0" (corner case though)

lib/Controller/CallController.php
```diff
@@ -104,9 +104,13 @@ public function getPeersForCall(): DataResponse {
...
+						$displayName = $participant->getAttendee()->getDisplayName();
+					} else {
+						$user = $this->userManager->get($participant->getAttendee()->getActorId());
+						if ($user instanceof IUser) {
+							$displayName = $user->getDisplayName();
```
PVince81, line 112: shouldn't we also store the display name in the attendee table also after retrieving it here, if the purpose is for caching it there ?

otherwise it looks like we'd only store it there if the user changes the display name, which is probably seldom.
unless I missed some code path that is not directly visible in this diff ?

PVince81: same comment for the other similar block

if that duplicated logic gets bigger we could move it into a common utility method maybe

lib/Controller/RoomController.php
nickvergessen: That's not valid and respected in a lot of other places, so don't care

lib/Controller/CallController.php
nickvergessen: Let's improve step by step.
But I would really like to avoid potentially lot of update queries making this request even more heavier.

We can add a repairstep/background job or something. But future story.

PVince81: Fair enough.

So far I still don't understand when that attendee display name is getting populated. From my understanding it will almost never get any values, unless people are changing their display names while in a call ? So it seems it's almost never happening ?

PVince81: (let me actually test this, that might answer my question)

PVince81: Okay, after a test I saw that the display name is not populated in the attendees table, not even during a call.
This means that during a big call we'd still go to the user manager for display names (unless I missed something).

The only way to get an entry there is changing your display name in profile page.
So, from my understanding, the performance benefit is not achieved, at least not in this PR.

We'll then need the repair steps, background job or whatever is supposed to populate there, then.

Why not also populate the field at the time the user is getting added to the conversation ? We're anyway inserting new entries there.

nickvergessen: LDAP automatically triggers an update periodically as far as i know. But then again, inserting it via the hook is step 1, initial population for existing entries is step 2

PVince81 (approved):
:+1: looking forward to step 2

PVince81:
needs rebase to get the tests green

PR #5122, opened 2021-02-11
Fix sharing tests to work with attachment folders fix

PVince81:
Fixes https://github.com/nextcloud/spreed/issues/5120

This adjusts the tests to accomodate for the fix from https://github.com/nextcloud/spreed/pull/5076


PVince81:
so the sharing tests all passed but conversation ones failed with timeout after an hour... restarted

nickvergessen (approved):

PR #5129, opened 2021-02-11
Provide information for incoming call screens

### LINKED ISSUE #4931: Incoming call screen
Make sure the apis provide enough information for a incoming call screen.

## Solution

Description | Screenshot
---|---
When a call is placed, the application will always display information regarding the call status and whether it is an audio or video call, when technically possible. For example the distinction between audio and video call is not always possible if the call was initiated from the web interface. In this case the call always starts with audio only but video can be enabled at any time. On the left you see the outgoing call screen "ringing". On then right you see the incoming call screen, which indicated at the top that it is an audio call “Nextcloud Talk voice call”. | image

### Notes

* Distinction between audio and video call is not always possible. For example from the web interface a call is always started with audio only but video can be enabled at any time.

### Acceptance criteria

* The incoming call screen should indicate whether it is a audio or video call.
* The incoming call screen should indicate who is calling and whether it is an incoming or outdoing call.
* The call screen on both sides should indicate the current status, such as “calling”, “ringing”, “reconnection”, “connected”.

Ref https://github.com/nextcloud/talk-android/issues/956
Ref https://github.com/nextcloud/talk-ios/issues/465

nickvergessen:
Fix #4931

### DIFF
--- a/lib/Migration/Version11001Date20210211111527.php
+++ b/lib/Migration/Version11001Date20210211111527.php
@@ -1,0 +1,54 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2021, Joas Schilling <coding@schilljs.com>
+ *
+ * @author Joas Schilling <coding@schilljs.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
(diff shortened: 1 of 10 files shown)


nickvergessen:
Currently clients (web and ios tested) seem to always join the call without a parameter so all are added as `7` which is connected+audio+video

lib/Migration/Version11001Date20210211111527.php
```diff
@@ -0,0 +1,60 @@
...
+namespace OCA\Talk\Migration;
+
+use Closure;
+use Doctrine\DBAL\Types\Types;
+use OCP\DB\ISchemaWrapper;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+/**
+ * Auto-generated migration step: Please modify to your needs!
```
PVince81, line 35: stray comment

lib/Migration/Version11001Date20210211111527.php
```diff
@@ -0,0 +1,60 @@
...
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('talk_rooms');
+		if (!$table->hasColumn('call_flag')) {
+
+			// Auto increment id
```
PVince81, line 51: stray comment ?

PVince81 (approved):
Looks fine :+1: 

See minor comments

nickvergessen:
Fixed the comments and rebased on the fixed integration tests

Issue #3598, opened 2020-05-17
Inconsistent handling of deleted users


simonspa:
## Steps to reproduce
1. Create user account
2. Create conversation, chat, call...
3. Delete user account

### Expected behaviour
The UI should consistently refer to the deleted user in one way (I don't mind which)

### Actual behaviour
Within the conversation and the sidebar, there are four different ways of mentioning the user:

* In the conversation list, it say "Deleted user (username)"
* Chat entries of the user are marked as "[Unknown user name]"
* Call starts/joins by the user are marked with "deleted_users" (plural)
* The call summary entry still features the username

Summary in one screenshot:

![image](https://user-images.githubusercontent.com/1677436/82157587-9e772400-9882-11ea-9197-17d9f4adbd23.png)

(This is Talk 8.0.9)

nickvergessen:
Yeah, that looks aweful and it's due to the different components dealing with them in different ways all being combined by Talk.

PVince81:
@nextcloud/designers probably something to unify on the whole platform ? Which one of the four ? :smile:

jancborchardt:
Yep, should be unified, and that is into the form:
> Deleted user (`username`)

nickvergessen:
You mean user id I hope/guess?

> Deleted user (user id)

Since we can't load any display names anymore as the information is gone. Also note that in theory this is not fully GDPR compliant in most cases (e.g. user id was your email or full name based)

skjnldsv:
If we want, we can start using the Vue RichDocument rendering methods.
And define undefined mentions there.

> Also note that in theory this is not fully GDPR compliant in most cases (e.g. user id was your email or full name based)

Right, the data would still be in the message source doe, so not displaying it isn't being gdpr compliant :)

nickvergessen:
Well mentions is one thing, left over one-to-one conversations is another, also author names and more.

PR #5136, opened 2021-02-11
Add a repair step that updates the display name for existing users

nickvergessen:

### DIFF
--- a/lib/Migration/CacheUserDisplayNames.php
+++ b/lib/Migration/CacheUserDisplayNames.php
@@ -1,0 +1,78 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2021 Joas Schilling <coding@schilljs.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Talk\Migration;
+
+use OCA\Talk\Model\Attendee;
+use OCP\IDBConnection;
+use OCP\IUser;
+use OCP\IUserManager;
+use OCP\Migration\IOutput;
+use OCP\Migration\IRepairStep;
+
+class CacheUserDisplayNames implements IRepairStep {
+
+	/** @var IDBConnection */
+	protected $connection;
+	/** @var IUserManager */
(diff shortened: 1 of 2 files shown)


lib/Migration/CacheUserDisplayNames.php
```diff
@@ -0,0 +1,78 @@
...
+		$query->select('actor_id')
+			->from('talk_attendees')
+			->where($query->expr()->eq('actor_type', $query->createNamedParameter(Attendee::ACTOR_USERS)))
+			->andWhere($query->expr()->eq('display_name', $query->createNamedParameter('')))
+			->groupBy('actor_id');
+
+		$result = $query->execute();
+		while ($row = $result->fetch()) {
+			$user = $this->userManager->get($row['actor_id']);
+			if (!$user instanceof IUser) {
```
PVince81, line 67: do we have separate repair step to remove orphaned attendees ? (we had something similar for shares)
or maybe we want to keep them there for showing "Deleted user" or similar (ref https://github.com/nextcloud/spreed/issues/3598)

lib/Migration/CacheUserDisplayNames.php
```diff
@@ -0,0 +1,78 @@
...
+		$query = $this->connection->getQueryBuilder();
+		$query->select('actor_id')
+			->from('talk_attendees')
+			->where($query->expr()->eq('actor_type', $query->createNamedParameter(Attendee::ACTOR_USERS)))
+			->andWhere($query->expr()->eq('display_name', $query->createNamedParameter('')))
+			->groupBy('actor_id');
+
+		$result = $query->execute();
+		while ($row = $result->fetch()) {
+			$user = $this->userManager->get($row['actor_id']);
```
PVince81, line 66: not sure if we should worry about millions of users and make that an occ command instead of running during update

I guess it is less likely that we'll traverse all possible users (like LDAP) as it would mean every one exists at least once in a conversation

PVince81 (approved):
Tested, works :+1:

lib/Migration/CacheUserDisplayNames.php
nickvergessen: > do we have separate repair step to remove orphaned attendees ?

Don't want to risk that in case it's just the user backend wasn't loaded when the repair step is run after the update or something.

lib/Migration/CacheUserDisplayNames.php
nickvergessen: Since this is only executed once after the update, I don't mind timeouts (as big instances need to update via occ anyway) and so this should be fine.

nickvergessen: Also in case it would ever timeout, a follow up run would continue from where it broke because of the `>andWhere($query->expr()->eq('display_name', $query->createNamedParameter('')))` part checking for empty names

nickvergessen:
`int-sqlite-conversation` timedout, so I restartet tests

lib/Migration/CacheUserDisplayNames.php
PVince81: Yeah, makes sense.

My comment was more in the lines of "is the admin aware that this update will take longer than usual" as admins need to plan time for downtime. But we could add this to the release notes to make them aware.

nickvergessen: Well it's a simple query per users. We and other apps have done worse things. e.g. the 10.1 release copied all entries for participants to the attendees table and stuff like that.

PR #5137, opened 2021-02-11
Reuse attendee for guests

nickvergessen:
- [x] Depends on #5118 
- [x] Stop the abuse of the `sha1(sessionId)` as actorId in the web UI
  - [x] Notify/check Android mobile clients
  - [x] Notify/check iOS mobile clients
- [x] Reuse existing attendees when a guest refreshes
  - [x] Extend the creation/join event with the participant info if needed
- [x] Make the call summary reuse the attendee table also for guests, so we don't have "and 132 guests" when a guest rejoins with a weak internet connection

### Follow ups
- Move the storing of the displayname from oc_talk_guestnames to oc_talk_attendees for guests as well
- Make clicking a SIP link identify the guest as the invited email attendee (will break current SIP dual-join work around)

### DIFF
--- a/lib/Activity/Listener.php
+++ b/lib/Activity/Listener.php
@@ -120,14 +120,13 @@ class Listener {
 
 		$duration = $this->timeFactory->getTime() - $activeSince->getTimestamp();
 		$userIds = $this->participantService->getParticipantUserIds($room, $activeSince);
+		$numGuests = $this->participantService->getGuestCount($room, $activeSince);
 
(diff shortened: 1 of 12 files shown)


PVince81 (approved):
Reviewed, tested, works.
Glad to see the streamlined guest handling in the frontend :+1:

PVince81:
while testing I discovered https://github.com/nextcloud/spreed/issues/5144 which is unrelated and also happens on master.

nickvergessen:
Fixed up

PR #5153, opened 2021-02-12
Close left sidebar menu after leave or delete

PVince81:
Pushed separately so we can backport and reduce user's frustration


nickvergessen:
What is this about. the full item is removed and so is the menu?

PVince81:
Before:
- on a narrow screen, click on "Delete conversation" and be frustrated that the menu is still open and partly obfuscating the "are you sure" dialog

After:
- menu closes and be happy

PVince81:
screen from before after clicking the menu item:
image

ma12-co (approved):

PR #5181, opened 2021-02-15
Help translators

### LINKED ISSUE #5134: l18n: Strings not translatable without context
There are some strings which are not translatable without context.

Example:
https://github.com/nextcloud/spreed/blob/066087bd7648d384ae1d0b6d44dda88545acdada/src/components/AdminSettings/TurnServer.vue#L37

Do you developers think the user will understand them?

Is there a need to translate such strings?

nickvergessen:
Fix #5134


rakekniven (approved):
Let us see how this will be reflected at Transifex.

nickvergessen:
Adding backport label for later, so we only backport this when enough translations have been made because we are close to the release

PR #5189, opened 2021-02-16
Limit the session list on the database instead of selecting all attendees

nickvergessen:
Should save some load as we don't create objects and select items we then throw away

- [x] Need to check whether this breaks the index


PVince81:
while I can try and infer from the code why you made this change, it would still be useful to add some notes in the PR description and commit for our future selves

nickvergessen:
Done

PVince81 (approved):
:+1:

Issue #4566, opened 2020-11-09
Ask to unmute


PVince81:
**Is your feature request related to a problem? Please describe.**
Sometimes people are muted and we saw them talking on video and would like to unmute them remotely.
In a conversation with many participants, the moderator might want to give voice to people for example after they [raised hands](https://github.com/nextcloud/spreed/issues/4562).

**Describe the solution you'd like**
A moderator can click the muted button of participants, either in the video view or sidebar, to trigger "ask to unmute". The participant then receives a dialog box "X asked to unmute, proceed ?". If "Yes" is clicked, their microphone is unmuted, otherwise nothing happens. If the moderator clicks the button again, the dialog disappears (in case of mistake).

Note: For privacy reasons, one should never be able to directly unmute remotely.

PR #5194, opened 2021-02-17
👥 Multiple sessions 👥

nickvergessen:
spiderman

* 😿  For now only multiple Talk sessions in different PHP sessions will be supported. So you can be in the same conversation with mobile apps, SIP, browser, etc at the same time, but not in the same chat in 2 tabs of the same browser. The additional burden is currently too much to handle
- [x] Transform the unique attendee key on the sessions table to a normal index
- [x] Make sure that we always select the Talk session of the current PHP session because there could be more than 1
- [ ] In places where the session is not used, the parameter should be set to false/null, to avoid left joining an unneeded database table
- [x] Move integration tests to API v4
- [x] Move UI to API v4
- [x] Remove special handling of lower API versions as the routes don't match anymore
- [x] Add v4 capability and remove others


lib/Controller/RoomController.php
```diff
@@ -476,7 +477,7 @@ protected function formatRoomV1(Room $room, ?Participant $currentParticipant): a
 		$numActiveGuests = 0;
 		$cleanGuests = false;
 		$participantList = [];
-		$participants = $this->participantService->getParticipantsForRoom($room);
+		$participants = $this->participantService->getParticipantsForRoom($room, true); // FIXME NEEDS the session but can potentially kill APIv1?
```
nickvergessen, line 480: I think we can just not keep this alive anymore. we can drop this when mobile clients all speak APIv3 by default and just fallback to APIv1 when no APIv3 is provided. that should work for iOS already and Android is on the way.

lib/Controller/RoomController.php
```diff
@@ -1109,7 +1110,7 @@ public function getParticipants(bool $includeStatus = false): DataResponse {
 		}
 
 		$maxPingAge = $this->timeFactory->getTime() - 100;
-		$participants = $this->participantService->getParticipantsForRoom($this->room);
+		$participants = $this->participantService->getParticipantsForRoom($this->room, true); // FIXME should return some what session info?
```
nickvergessen, line 1113: This here however is more complicated. We want to show online-state and call icons, but it basically means we would need to select all sessions and then group by actor. We can of course do that, and return a combined call flag of all sessions, but which session id should be returned?
I think currently some clients abuse this for their call screen instead of `CallController::getPeersForCall` (even the web?).

From my POV this should be fixed to use the proper API and then the clients when receiving "please refresh participant list" they refresh participants and peer list for the internal signaling and refresh participants for external signaling + use the data of the event to update their peers.

nickvergessen:
Rebase on #5215 for compatibility

nickvergessen:
- [ ] Room API returns rooms twice when being 2 times in the conversation
- [ ] UI switches to call view for all sessions instead of only those that joined the call, could be caused by ^ or the RoomController returning data of a random session and the UI just using that participant info.

lib/Controller/RoomController.php
nickvergessen: this is still the question what to do.
I would introduce a v4 with the current changes that take place to make clear that only Talk sessions of the current PHP session are returned for the current user and basically would like to kill all older room APIs, as their behaviour is inconsistent to how it was before and we can't keep it backwards compatible.

nickvergessen:
Psalm should be happy with https://github.com/nextcloud/server/pull/25816

nickvergessen:
- [x] Move integration tests to API v4
- [ ] Move UI to API v4
- [ ] Remove special handling of lower API versions as the routes don't match anymore

nickvergessen:
Ready to review. Don't be scared about the size, it's mostly integration tests bumping the required version of the API from none (v1) or v2 or v3 to the new one-and-only v4

docs/call.md
```diff
@@ -1,8 +1,9 @@
 # Call API
 
-* Base endpoint for API v1 is: `/ocs/v2.php/apps/spreed/api/v1`
-* Base endpoint for API v2 is: `/ocs/v2.php/apps/spreed/api/v2`
-* Base endpoint for API v3 is: `/ocs/v2.php/apps/spreed/api/v3`
+* API v1: 🏁 Removed with API v4
+* API v2: 🏁 Removed with API v4
+* API v3: 🏁 Removed with API v4
+* API v4: Base endpoint `/ocs/v2.php/apps/spreed/api/v4`
```
PVince81, line 6: So this means all clients will stop working with Talk 12 and an upgrade will be mandatory.
If there are no scenarios where someone would not want to upgrade their apps (branding?) I guess this is ok.
Also it won't be possible to first upgrade all client apps first and then the server, so the migration will have to be done "all at once", not gradually, which could mean a bigger downtime for server and clients during that time (but I'm not 100% familiar how upgrades are done in enterprise)

Probably we'll need to make this very clear in release notes and communications.
"Don't upgrade to NC 22 if you can't upgrade all the client apps also at the same time".

Just thinking out loud. I understand that there is no way to keep compatibility.

nickvergessen: > Also it won't be possible to first upgrade all client apps first and then the server, so the migration will have to be done "all at once", not gradually, which could mean a bigger downtime for server and clients during that time (but I'm not 100% familiar how upgrades are done in enterprise)

No in theory apps can already start rolling out, the capability will tell them if v4 is available on the server and otherwise they use the old v1-3 apis. It's just not possible anymore to serve v1-3 from a current server. but clients can be compatible with v1-4 at the same time (with some if v4 then X else Y logic)

lib/Room.php
```diff
@@ -402,28 +403,47 @@ public function getPropertiesForSignaling(string $userId, bool $roomModified = t
...
+
+		if ($sessionId !== false) {
+			if ($sessionId !== null) {
+				$helper->selectSessionsTable($query);
+				$query->leftJoin('a', 'talk_sessions', 's', $query->expr()->andX(
+					$query->expr()->eq('s.session_id', $query->createNamedParameter($sessionId)),
+					$query->expr()->eq('a.id', 's.attendee_id')
+				));
+			} else {
+				$helper->selectSessionsTable($query); // FIXME PROBLEM
```
PVince81, line 442: fixme problem?

src/services/participantsService.js
```diff
@@ -158,7 +158,7 @@ const leaveConversation = async function(token) {
  * @param {string} token The conversation token;
  */
 const leaveConversationSync = function(token) {
-	axios.delete(generateOcsUrl('apps/spreed/api/v3/room', 2) + token + '/participants/active')
+	axios.delete(generateOcsUrl('apps/spreed/api/v4', 2) + `room/${token}/participants/active`)
```
PVince81, line 161: this would not escape the token arg, but it's ok

sadly generateOcsUrl doesn't seem to accept params, but generateUrl does: https://github.com/nextcloud/nextcloud-router/blob/master/lib/index.ts#L53

I've raised it here for later: https://github.com/nextcloud/nextcloud-router/issues/194

since the token is known to be safe, ok to keep

nickvergessen: Oh, I didn't know this as all APIs I use are OCS. sounds like a good enhancement

src/store/participantsStore.js
```diff
@@ -114,6 +125,32 @@ const mutations = {
...
+		} else {
+			if (!state.inCall[token]) {
+				Vue.set(state.inCall, token, {})
+			}
+			Vue.set(state.inCall[token], sessionId, flags)
+
+			if (!state.connecting[token]) {
+				Vue.set(state.connecting, token, {})
+			}
+			Vue.set(state.connecting[token], sessionId, flags)
```
PVince81, line 146: so we're setting the exact same values/structure to "connecting" like "inCall" ?
either redundant or a copy-paste mistake ?

since none of these have any other getters than `isInCall` and `isConnected` the data structure could be merged into a single one but still keep both getters ?

src/store/participantsStore.js
```diff
@@ -247,23 +284,48 @@ const actions = {
 	},
 
 	async joinCall({ commit, getters }, { token, participantIdentifier, flags }) {
+		if (!participantIdentifier?.sessionId) {
+			console.error('Trying to join call without sessionId')
```
PVince81, line 288: no return, so we still want to attempt the server call?

tests/integration/features/callapi/public.feature
```diff
@@ -5,105 +5,105 @@ Feature: callapi/public
...
+    And user "guest" joins call "room" with 200 (v4)
+    Then user "participant1" sees 2 peers in call "room" with 200 (v4)
+    And user "guest" sees 2 peers in call "room" with 200 (v4)
+    Then user "guest" leaves call "room" with 200 (v4)
+    Then user "participant1" sees 1 peers in call "room" with 200 (v4)
+    And user "guest" sees 1 peers in call "room" with 200 (v4)
+    Then user "guest" leaves room "room" with 200 (v4)
+    Then user "participant1" sees 1 peers in call "room" with 200 (v4)
+    And user "guest" sees 0 peers in call "room" with 404 (v4)
+    Then user "participant1" leaves call "room" with 200 (v4)
```
PVince81, line 104: after scrolling through pages and pages if sprinkled `(v4)` I wonder if we even need to mention it in the tests, considering that v4 is the new default and we're not going to test versions < v4 and only when v5 comes in, if incompatible, we would require to specify it

PVince81 (commented):
Some minor comments, especially some duplication in frontend.

Haven't tested yet

src/services/participantsService.js
PVince81: Yeah, I think it used to be that way when it was still part of server/core and didn't get ported that way.

lib/Room.php
nickvergessen: Fixed

src/store/participantsStore.js
nickvergessen: I find it easier with this explicit way instead of wrapping my head around existing and non-existing attributes and also we need to block "faulty" updates from the participant list update etc. But feel free to do it in a follow up PR when you think it's worth it

src/store/participantsStore.js
nickvergessen: right, fixed

tests/integration/features/callapi/public.feature
nickvergessen: So the thing is, the version was optional until now. I force it to now, because I noticed that all the tests used the v1 in 99% of the cases, so they didn't use/test the methods we use in the UI etc.
Sure we could go back to optional and just use v4 as fallback, but I'd rather be explicit making everyone aware that any other version is not tested.

src/store/participantsStore.js
PVince81: Not sure if you understood what I meant.

It seems that both `isInCall` and `isConnected` always contain the exact same data "flags".
There is no scenario I could find where they would contain different data, so why use two attributes instead of just one ?

PVince81: the attribute could be called `callFlags[token]` since it contains the flags.

PVince81: ah! okay, found it: so `finishedConnecting` is then deleting the token entry after success. that's the bit I missed.

in that case maybe just store a boolean in `connecting` instead of all the flags to avoid confusion

PVince81: I'm fine leaving everything as is. it's just likely that the next dev coming over here might wonder as well

tests/integration/features/callapi/public.feature
PVince81: agreed.

when scrolling further I saw we still had some v1 steps, so making it explicit will make sure there is no confusion

PVince81:
> In places where the session is not used, the parameter should be set to false/null, to avoid left joining an unneeded database table

See checkbox, is it done ? I saw a lot of places where you set it to false alreay.

PVince81:
Tests:
- [x] TEST: could join twice with same user
- [x] TEST: raising and lowering hand from both sessions properly updates the indicator in participant list for others (aggregated)
- [x] TEST: kicking out user from conversation terminated both sessions
- [ ] TEST: joining twice as the same guest
- [ ] TEST: files sidebar chat
- [ ] TEST: video verification (regresison check)
- [ ] TEST: joining+call twice as self in 1-1 conversation

lib/Events/RemoveParticipantEvent.php
```diff
@@ -31,14 +32,26 @@ class RemoveParticipantEvent extends ParticipantEvent {
...
 	protected $reason;
 
+	/** @var Session[] */
+	protected $sessions;
+
 	public function __construct(Room $room,
 								Participant $participant,
-								string $reason) {
+								string $reason,
+								array $sessions = []) {
```
nickvergessen, line 41: Since i moved the listener to the before event this modification might not be neededd

PVince81:
@nickvergessen seems the multi-session thing is not working with HPB (receiving duplicate session error), only internal signaling works. I didn't see any statement about this so not sure if it's a bug or whether my version of the HPB is outdated.

nickvergessen:
I didn't check HPB yet, it also needs some adjusting as it calls v3 api itself. I notified Jojo already about the changes and will test that part more deeply monday

PVince81:
- [ ] BUG: joining as the same guest twice in the same browser (not technically the same session) works for chatting but the call part is buggy

Not sure if we intended to support this. If not we should catch it and return a "duplicate session" error like for regular accounts.

nickvergessen:
> BUG: with HPB joining as the same guest twice in the same browser (not technically the same session) works for chatting but the call part is buggy

Same browser is the same php session and therefor not something we can fix atm

PVince81:
I think I'm done testing. Apart from the HPB case and some corner issues, it seemed to work fine :smile:

See what I tested here https://github.com/nextcloud/spreed/pull/5194#issuecomment-791420898
Will need to redo this once it works with HPB.

nickvergessen:
> * TEST: could join and call twice with same user
>    * HPB no_entry_sign getting duplicate session

So this is an easy fix. The guest thing however is impossible atm. Also for the HPB users break as they can join the same conversation twice with 2 different session ids in the same browser. The problem is we don't have access to the PHP session when the signaling server tells us "Hey, user X joined with session id Y" and so we always create a new session. The problem is one of them has a session that doesn't match the session id in the PHP session and therefor is not match/returned correctly.

nickvergessen:
> system messages for join/leave only displayed once when relevant 

So the question of course is what "relevant" is. I would just display it all the time for calls as it might be important to know that another device disconnected.

nickvergessen:
Added 3 commits and rebased.
Works as expected now with the HPB from my POV.

Only thing is guests see "Conversation not found" after their previous session gets kicked because of duplication.

PVince81:
I've tested with HPB and all seemed to work fine.

There's only a minor issue with raise hand because we don't aggregate HPB style session id in the participant list, but perhaps we can fix this separately.

PVince81 (approved):
:+1:

PVince81:
some integration tests are failing still: https://drone.nextcloud.com/nextcloud/spreed/1816/4/3
```
/drone/server/apps/spreed/tests/integration/features/conversation/one-to-one.feature:121
    /drone/server/apps/spreed/tests/integration/features/conversation/one-to-one.feature:130
    /drone/server/apps/spreed/tests/integration/features/conversation/remove-participant.feature:10
    /drone/server/apps/spreed/tests/integration/features/conversation/remove-participant.feature:18
    /drone/server/apps/spreed/tests/integration/features/conversation/remove-participant.feature:30
    /drone/server/apps/spreed/tests/integration/features/conversation/remove-participant.feature:67
    /drone/server/apps/spreed/tests/integration/features/conversation/remove-participant.feature:79
    /drone/server/apps/spreed/tests/integration/features/conversation/remove-participant.feature:94
```

nickvergessen:
Ah yeah... that is what the fixme was about 😅

PVince81:
Also, I still have some features in my mind like "[ask to unmute](https://github.com/nextcloud/spreed/issues/4566)" or "[mute from sidebar](https://github.com/nextcloud/spreed/issues/4412)". Now since we only show a participant once during a call there the UX for such would be more tricky as we'd need to either ask the user which session they want to mute (maybe only one is unmuted by mistake) or do it for all of them.

Also not sure if it's a good idea to temporarily show one participant as multiple entries in the participant list during a call. Or group them somehow with an expandable arrow. Feels like it's going too far.

nickvergessen:
> Also, I still have some features in my mind like "ask to unmute" or "mute from sidebar". Now since we only show a participant once during a call there the UX for such would be more tricky as we'd need to either ask the user which session they want to mute (maybe only one is unmuted by mistake) or do it for all of them.

Yeah, let's discuss this separatly. For now muting works on the call view and its okay for now.

nickvergessen:
> raising and lowering hand from both sessions properly updates the indicator in participant list for others (aggregated) no_entry_sign the participant list seems tied only to the first session so the hand only appears when first session raises/lowered, might need some tweaks for session id matching / aggregation => raise for separate fixing ?

https://github.com/nextcloud/spreed/issues/5349

> Only thing is guests see "Conversation not found" after their previous session gets kicked because of duplication, but lets fix that later.

https://github.com/nextcloud/spreed/issues/5350

PVince81:
:tada:

PR #5197, opened 2021-02-18
Check psalm security

nickvergessen:


nickvergessen:
Currently blocked because Psalm can only run in root and we therefore would scan the server part too.

nickvergessen:
There seems to be no progress on the github action, so this can't be fixed atm => closing

Issue #5188, opened 2021-02-16
Older messages sometimes missing until full page reload


LukasReschke:
## Steps to reproduce
1. Open a chat on c.nc.c
2. See content is missing in the backlog sometimes

### Expected behaviour

I see older messages.

### Actual behaviour

I don't see older messages in some cases. I have two different scenarios:

1. Group chats in which scrolling back sometimes misses messages
2. New chats that have no content when opened (reopening sometimes helps)

Example of a 1:1 chat where I have to refresh my browser to see the received message: (which I received yesterday already and wanted to revisit today)

![image](https://user-images.githubusercontent.com/878997/108062036-80c92c00-7059-11eb-9784-703210bb1b22.png)

![image](https://user-images.githubusercontent.com/878997/108062051-86267680-7059-11eb-90bb-60a3169df089.png)

nickvergessen:
Yeah I also sometimes experience an empty chat with only new messages (no history) lately from time to time.
However we didn't find any reproducable steps yet.

LukasReschke:
@nickvergessen it reproduces for me on rhwpfxci when selecting it in the sidebar. Is there any request you are particularly interested in to see? :)

PVince81:
I just had a local case where one message was missing in the middle of a chat until refresh. Not sure if related or different issue.
Sadly I refreshed too quickly and lost the calls.

PVince81:
I have an idea about a test: have a bot post messages that count up in two rooms. Then switch between the rooms manually from times to times and see if any numbers are missing.

PVince81:
I've managed to gather some incomplete clues about a second case of missing message:

1. the message with id 492 is also missing in the store

2. access log around that case (note: this is based on https://github.com/nextcloud/spreed/pull/3825 which doesn't reset the read marker)
```
127.0.0.1 - - [17/Feb/2021:15:31:21 +0100] "POST /ocs/v2.php/apps/spreed/api/v1/chat/va6i7nwy/read HTTP/1.1" 200 81
127.0.0.1 - - [17/Feb/2021:15:31:19 +0100] "GET /ocs/v2.php/apps/spreed/api/v1/chat/va6i7nwy?setReadMarker=0&lookIntoFuture=1&lastKnownMessageId=486&includeLastKnown=0 HTTP/1.1" 200 225
[...]

PR #5198, opened 2021-02-18
Prevent loading old messages twice on scroll

PVince81:
Whenever the old messages are already loading initially when opening a
conversation, a scroll event might call getOldMessages again which would
cancel the previous one.

The initial old messages load is using includeLastKnown=1 while the others use
0. So if it gets cancelled, we will miss the "last known" message as it
won't be included in subsequent calls.

This fix prevents repeated calls to getOldMessages to avoid the above
issue.

See https://github.com/nextcloud/spreed/pull/3825#issuecomment-780719458 for semi-reliable reproduction steps and https://github.com/nextcloud/spreed/pull/3825#issuecomment-781319116 for some information.

The handleScroll method was sometimes calling getOldMessages(false) while we still had getOldMessages(true), which led to having a missing message in the list. Usually the one with the read marker, hence why I discovered it only in the PR above.

With a bit of luck this might also fix https://github.com/nextcloud/spreed/issues/5188 as it will also prevent repeated calls to getOldMessages(false) if scrolling up/down multiple times when at the top of the container.

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -131,6 +131,8 @@ export default {
 			previousScrollTopValue: null,
 
 			pollingErrorTimeout: 1,
+
+			oldMessagesPromise: null,
 		}
 	},
 
@@ -429,7 +431,8 @@ export default {
 
 			// Make the request
 			try {
-				const messages = await request({ token, lastKnownMessageId, includeLastKnown: includeLastKnown ? '1' : '0' })
+				this.oldMessagesPromise = request({ token, lastKnownMessageId, includeLastKnown: includeLastKnown ? '1' : '0' })
+				const messages = await this.oldMessagesPromise
 				// Process each messages and adds it to the store
 				messages.data.ocs.data.forEach(message => {
 					if (message.actorType === 'guests') {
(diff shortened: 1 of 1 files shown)


nickvergessen:
Ay that's what I'm always experiencing

src/components/MessagesList/MessagesList.vue
```diff
@@ -429,7 +431,8 @@ export default {
 
 			// Make the request
 			try {
-				const messages = await request({ token, lastKnownMessageId, includeLastKnown: includeLastKnown ? '1' : '0' })
+				this.oldMessagesPromise = request({ token, lastKnownMessageId, includeLastKnown: includeLastKnown ? '1' : '0' })
```
nickvergessen, line 434: the component is reused, so you need to store this based on token?
Otherwise if you quickly change to 2 unloaded conversations the second one wouldn't load?

ma12-co: Or more simply reset the value to null when the `token` changes?

PVince81: if you switch to another conversation the call will get cancelled anyway, so you'll then get a new request

nickvergessen (approved):
Nice one, seems to work

nickvergessen:
Queuing backport for after 11.0.0 is out

PR #5212, opened 2021-02-19
Replace all message parameters and not only the last one

### LINKED ISSUE #5199: Tooltip in unified search doesn't show mentions correctly
image

We should probably switch to "v-tooltip" there anyway so we can display richer tooltips with the mention bubbles.

nickvergessen:
Fix #5199


PVince81 (approved):
:+1:

PR #5213, opened 2021-02-19
Drop guest moderators when changing a conversation to group conversation

### LINKED ISSUE #5193: Changing type from public to group conversation doesn't kill guest moderators
https://github.com/nextcloud/spreed/blob/e4556d8327afd71750f1110573c80f39645a8bbf/lib/Room.php#L795

### Expected behaviour
Guest moderators are killed immediately

### Actual behaviour
Guest moderators are missed, might break a tad later, but it should be cleaned here immediately

nickvergessen:
Fix #5193


PVince81 (approved):
:+1: makes sense

PVince81:
backports or later ?

PR #5216, opened 2021-02-19
Update it.json

stefanomarty:
Correct wrong translation.


nickvergessen:
Needs to be fixed in transifex directly

stefanomarty:
> Needs to be fixed in transifex directly

I'm sorry, I don't know how to do that, let's hope somebody will do it.

PR #5239, opened 2021-02-22
Adjust screenshoting

nickvergessen:
Updates the screenshot on the front page.
Others will be updated after the release only


PVince81 (approved):
:+1:

PR #5242, opened 2021-02-23
Try to catch the unique constraint violation on migrating the data

nickvergessen:
### Steps to reproduce
1. Install on stable20
2. Create a room and join it
3. In the oc_talk_participants table create a second entry without a session id (so use string `'0'`) for your user and that room
4. Checkout stable20.1 and try to update

### Before
💥  Unique constraint violation

### After
🏁  Only one attendee is added and all is fine.

### DIFF
--- a/lib/Migration/Version10000Date20201015134000.php
+++ b/lib/Migration/Version10000Date20201015134000.php
@@ -26,10 +26,12 @@ declare(strict_types=1);
 namespace OCA\Talk\Migration;
 
 use Closure;
+use Doctrine\DBAL\Exception\UniqueConstraintViolationException;
 use Doctrine\DBAL\Types\Types;
 use OCA\Talk\Model\Attendee;
 use OCA\Talk\Participant;
 use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\DB\Exception;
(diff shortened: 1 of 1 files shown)


lib/Migration/Version10000Date20201015134000.php
```diff
@@ -220,7 +222,22 @@ public function postSchemaChange(IOutput $output, Closure $schemaClosure, array
 				->setParameter('last_mention_message', (int) $row['last_mention_message'], IQueryBuilder::PARAM_INT)
 				;
 
-			$insert->execute();
+			try {
+				$insert->execute();
+			} catch (\Exception $e) {
+				if (get_class($e) === UniqueConstraintViolationException::class) {
```
PVince81, line 228: strange approach.
no multiple catch blocks ?

PVince81 (approved):
:+1:

lib/Migration/Version10000Date20201015134000.php
nickvergessen: I'm not sure you can catch classes that don't exist, just wanted to be on the save side because I know `::class` works without the class existing

PR #5252, opened 2021-02-24
Only show quick conversation settings menu only when hovering or focu…

marcoambrosini:
…sing a conversation


PVince81:
- [ ] verify that this is accessibility friendly (haven't tested yet)
   - need to be able to display that extra info also with keyboard
   - check if mobile devices also get access to that extra info

In my opinion we should keep the three dots instead of hover, like the avatar menu which we planned to contain more context information. Having it as a button makes it work for everybody.

PVince81:
maybe put the three dots on the conversation icon as if it was the conversation's avatar ?

ma12-co:
There's an ongoing discussion in the nc-vue repo for this, I'm just using this as a parking lot for now :)

marcoambrosini:
Done in nextcloud vue lib

PR #5257, opened 2021-02-24
Use addMessageForAllParticipants instead of looping manually

nickvergessen:
Split off from #5194


PVince81 (approved):
Makes sense, yes :+1:

PR #5258, opened 2021-02-24
Base queries on sessions where it makes sense

nickvergessen:
Split off from #5194


PVince81 (approved):
As far as I could understand, this seems to make sense :+1:

PR #5259, opened 2021-02-24
conversation state: Only Moderators can write

jakobroehrl:
solves: #3886

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -282,6 +282,15 @@ return [
 			],
 		],
 		[
+			'name' => 'Room#setConversationState',
+			'url' => '/api/{apiVersion}/room/{token}/conversation-state',
+			'verb' => 'PUT',
+			'requirements' => [
+				'apiVersion' => 'v(1|2|3)',
+				'token' => '^[a-z0-9]{4,30}$',
+			],
+		],
+		[
 			'name' => 'Room#setListable',
 			'url' => '/api/{apiVersion}/room/{token}/listable',
 			'verb' => 'PUT',
--- a/src/components/ConversationSettings/LockingSettings.vue
+++ b/src/components/ConversationSettings/LockingSettings.vue
@@ -34,6 +34,18 @@
 				:disabled="isReadOnlyStateLoading"
 				@change="toggleReadOnly">
 			<label for="moderation_settings_lock_conversation_checkbox">{{ t('spreed', 'Lock conversation') }}</label>
+
+			<h3>{{ t('spreed', 'Limit writing to a conversation') }}</h3>
+			<p>
+				<Multiselect id="writing_conversations"
+					v-model="writingConversations"
+					:options="writingConversationOptions"
+					:placeholder="t('spreed', 'Limit writing to conversations')"
+					label="label"
+					track-by="value"
+					:disabled="loading || loadingWritingConversations"
+					@input="saveWritingConversations" />
+			</p>
 		</div>
 	</div>
 </template>
@@ -41,9 +53,20 @@
 <script>
 import { showError, showSuccess } from '@nextcloud/dialogs'
 import { CONVERSATION } from '../../constants'
+import Multiselect from '@nextcloud/vue/dist/Components/Multiselect'
+
+const writingConversationOptions = [
+	{ value: 0, label: t('spreed', 'Everyone can read and write') },
+	{ value: 1, label: t('spreed', 'Lock conversation') },
+	{ value: 2, label: t('spreed', 'Only Moderators can write') },
+]
 
 export default {
 	name: 'LockingSettings',
+
+	components: {
+		Multiselect,
+	},
 
 	props: {
 		token: {
@@ -55,6 +78,9 @@ export default {
 	data() {
 		return {
 			isReadOnlyStateLoading: false,
+			writingConversationOptions,
+			writingConversations: writingConversationOptions[0],
+			loadingWritingConversations: false,
 		}
 	},
 
@@ -92,6 +118,19 @@ export default {
 				}
 			}
 			this.isReadOnlyStateLoading = false
+		},
+
+		async saveWritingConversations() {
+			this.loadingWritingConversations = true
+			try {
+				await this.$store.dispatch('setConversationState', {
+					token: this.token,
+					state: this.writingConversations,
+				})
+			} catch (e) {
+
+			}
+			this.loadingWritingConversations = false
 		},
 	},
 
--- a/src/services/conversationsService.js
+++ b/src/services/conversationsService.js
@@ -335,6 +335,22 @@ const changeReadOnlyState = async function(token, readOnly) {
 }
 
 /**
+ * Change the converstaion state
+ * @param {string} token The token of the conversation to be modified
+ * @param {int} state The new state to set
+ */
+const changeConversationState = async function(token, state) {
+	try {
+		const response = await axios.put(generateOcsUrl('apps/spreed/api/v3', 2) + `room/${token}/conversation-state`, {
+			state: state,
+		})
+		return response
+	} catch (error) {
+		console.debug('Error while updating converstation state: ', error)
+	}
+}
+
+/**
  * Change the listable scope
  * @param {string} token The token of the conversation to be modified
  * @param {int} listable The new listable scope to set
@@ -375,4 +391,5 @@ export {
 	setConversationPassword,
 	setConversationName,
 	setConversationDescription,
+	changeConversationState,
 }
--- a/src/store/conversationsStore.js
+++ b/src/store/conversationsStore.js
@@ -26,6 +26,7 @@ import {
 	setSIPEnabled,
 	changeLobbyState,
 	changeReadOnlyState,
+	changeConversationState,
 	changeListable,
 	addToFavorites,
 	removeFromFavorites,
@@ -239,6 +240,18 @@ const actions = {
 
 		await changeReadOnlyState(token, readOnly)
 		conversation.readOnly = readOnly
+
+		commit('addConversation', conversation)
+	},
+
+	async setConversationState({ commit, getters }, { token, state }) {
+		const conversation = Object.assign({}, getters.conversations[token])
+		if (!conversation) {
+			return
+		}
+
+		await changeConversationState(token, state)
+		conversation.state = state
 
 		commit('addConversation', conversation)
 	},
--- a/lib/Room.php
+++ b/lib/Room.php
@@ -62,6 +62,7 @@ class Room {
 
 	public const READ_WRITE = 0;
 	public const READ_ONLY = 1;
+	public const WRITE_ONLY_MOD = 2;
 
 	/**
 	 * Only visible when joined
--- a/src/constants.js
+++ b/src/constants.js
@@ -33,6 +33,7 @@ export const CONVERSATION = {
 	STATE: {
 		READ_WRITE: 0,
 		READ_ONLY: 1,
+		WRITE_ONLY_MOD: 2,
 	},
 	LISTABLE: {
 		NONE: 0,


appinfo/routes.php
```diff
@@ -281,6 +281,15 @@
 				'token' => '^[a-z0-9]{4,30}$',
 			],
 		],
+		[
+			'name' => 'Room#setConversationState',
+			'url' => '/api/{apiVersion}/room/{token}/conversation-state',
```
PVince81, line 286: how about reusing the "readOnly" attribute and the "/read-only" endpoint but with the new flag value ?

nickvergessen: Yeah sounds like a good idea

jakobroehrl:
@PVince81 
Do you have an idea how I get my state value here to set it in my new state dropdown:
src/components/ConversationSettings/LockingSettings.vue

src/services/conversationsService.js
```diff
@@ -334,6 +334,22 @@ const changeReadOnlyState = async function(token, readOnly) {
...
+ * @param {string} token The token of the conversation to be modified
+ * @param {int} state The new state to set
+ */
+const changeConversationState = async function(token, state) {
+	try {
+		const response = await axios.put(generateOcsUrl('apps/spreed/api/v3', 2) + `room/${token}/read-only`, {
+			state: state,
+		})
+		return response
+	} catch (error) {
```
PVince81, line 348: don't catch in the service.
I know we did this before but want to stop doing this for new calls.
The caller should handle the error accordingly.

src/store/conversationsStore.js
```diff
@@ -26,6 +26,7 @@ import {
 	setSIPEnabled,
 	changeLobbyState,
 	changeReadOnlyState,
+	changeConversationState,
```
PVince81, line 29: I suggest you simply reuse changeReadOnlyState and extend it with the new attribute values where needed, no need for an extra method.

This means you'll also be able to reuse this.conversation.readOnly to contain the new state as well

src/components/ConversationSettings/LockingSettings.vue
```diff
@@ -34,17 +34,40 @@
 				:disabled="isReadOnlyStateLoading"
 				@change="toggleReadOnly">
 			<label for="moderation_settings_lock_conversation_checkbox">{{ t('spreed', 'Lock conversation') }}</label>
+
+			<h3>{{ t('spreed', 'Limit writing to a conversation') }}</h3>
```
PVince81, line 38: please integrate with "lock conversation" which is another wording for the read-only state

src/components/ConversationSettings/LockingSettings.vue
```diff
@@ -34,17 +34,41 @@
...
+					v-model="writingConversations"
+					:options="writingConversationOptions"
+					:placeholder="t('spreed', 'Limit writing to conversations')"
+					label="label"
+					track-by="value"
+					:disabled="loading || loadingWritingConversations"
+					@input="saveWritingConversations" />
+			</p>
 		</div>
+		{{conversation}}
```
nickvergessen, line 50: 🧐

src/components/ConversationSettings/LockingSettings.vue
```diff
@@ -25,26 +25,38 @@
...
-				@change="toggleReadOnly">
-			<label for="moderation_settings_lock_conversation_checkbox">{{ t('spreed', 'Lock conversation') }}</label>
+			<h3>{{ t('spreed', 'Limit writing to a conversation') }}</h3>
+			<p>
+				<Multiselect id="limit_writing"
+					v-model="writingConversations"
+					:options="writingConversationOptions"
+					:placeholder="t('spreed', 'Limit writing to conversations')"
+					label="label"
+					track-by="value"
```
PVince81, line 35: please also set ":disabled" while saving. I hope that property is available on this component.

src/components/ConversationSettings/LockingSettings.vue
```diff
@@ -25,26 +25,38 @@
...
 
 <script>
 import { showError, showSuccess } from '@nextcloud/dialogs'
 import { CONVERSATION } from '../../constants'
+import Multiselect from '@nextcloud/vue/dist/Components/Multiselect'
+
+const writingConversationOptions = [
+	{ value: 0, label: t('spreed', 'Everyone can read and write') },
+	{ value: 1, label: t('spreed', 'Lock conversation') },
+	{ value: 2, label: t('spreed', 'Only Moderators can write') },
```
PVince81, line 50: not sure if we want to capitalize moderators here

src/components/ConversationSettings/LockingSettings.vue
```diff
@@ -55,6 +67,9 @@ export default {
 	data() {
 		return {
 			isReadOnlyStateLoading: false,
+			writingConversationOptions,
+			writingConversations: writingConversationOptions[this.isReadOnly],
```
PVince81, line 71: I think you should rename this and change the logic. It's not a boolean any more but an integer value.

Same in saveWritingConversations, instead of flipping the boolean you should take the value from the multiselect.

jakobroehrl:
@PVince81 
Thanks for looking over, now it's working! :)
1. Should I change the status messages, too? https://github.com/nextcloud/spreed/blob/e44215e984ef1b9e7145243b2bbf1204fb968c75/src/components/ConversationSettings/LockingSettings.vue#L105
2. Should I try to show the options lilke here?
image

src/components/ChatView.vue
```diff
@@ -97,12 +97,25 @@ export default {
...
+		currentParticipant() {
+			return this.$store.getters.conversation(this.token) || {
+				sessionId: '0',
+				participantType: this.$store.getters.getUserId() !== null ? PARTICIPANT.TYPE.USER : PARTICIPANT.TYPE.GUEST,
+			}
+		},
+
+		selfIsModerator() {
+			return this.currentParticipant.participantType === PARTICIPANT.TYPE.OWNER
+				|| this.currentParticipant.participantType === PARTICIPANT.TYPE.MODERATOR
```
nickvergessen, line 116: What about `PARTICIPANT.TYPE.GUEST_MODERATOR`

src/components/ConversationSettings/LockingSettings.vue
```diff
@@ -25,26 +25,40 @@
...
-				aria-describedby="moderation_settings_lock_conversation_hint"
-				type="checkbox"
-				class="checkbox"
-				name="moderation_settings_lock_conversation_checkbox"
-				:checked="isReadOnly"
-				:disabled="isReadOnlyStateLoading"
-				@change="toggleReadOnly">
-			<label for="moderation_settings_lock_conversation_checkbox">{{ t('spreed', 'Lock conversation') }}</label>
+			<h3>{{ t('spreed', 'Limit writing to a conversation') }}</h3>
+			choosen: {{ writingConversationSelected }}
```
nickvergessen, line 29: debug?

src/components/ConversationSettings/LockingSettings.vue
```diff
@@ -25,26 +25,40 @@
...
 
 <script>
 import { showError, showSuccess } from '@nextcloud/dialogs'
 import { CONVERSATION } from '../../constants'
+import Multiselect from '@nextcloud/vue/dist/Components/Multiselect'
+
+const writingConversationOptions = [
+	{ value: 0, label: t('spreed', 'Everyone can read and write') },
+	{ value: 1, label: t('spreed', 'Lock conversation') },
+	{ value: 2, label: t('spreed', 'Only moderators can write') },
```
nickvergessen, line 52:
```suggestion
	{ value: 0, label: t('spreed', 'Everyone can read and write') },
	{ value: 2, label: t('spreed', 'Only moderators can write') },
	{ value: 1, label: t('spreed', 'Lock conversation') },
```

Let's change the order to something logical (not the value)

src/components/ConversationSettings/LockingSettings.vue
```diff
@@ -68,9 +85,16 @@ export default {
 		},
 	},
 
+	mounted() {
+		this.isReadOnlyStateLoading = true
+		const conversation = this.$store.getters.conversation(this.token)
```
nickvergessen, line 90: this needs to go into a computed otherwise it will never update when someone else changes the state?

jakobroehrl:
> @PVince81
> Thanks for looking over, now it's working! :)
> 
> 1. Should I change the status messages, too? https://github.com/nextcloud/spreed/blob/e44215e984ef1b9e7145243b2bbf1204fb968c75/src/components/ConversationSettings/LockingSettings.vue#L105
> 2. Should I try to show the options like here?
>    image

@PVince81 
Could you maybe look over this? Thanks! :)

PVince81:
- [ ] increase the field width

image

@ma12-co or replace it with something similar-looking like the notifications ?

- [ ] please adjust the status messages and matching tests
- [ ] please search for "readonly" on the server side, there are various places to extend like for example the occ commands to allow setting the new flag

PVince81:
- [ ] bug: the reply and delete options are still visible in the menu even when I can't post (could be a separate bug that existed before)
image

PVince81:
- [ ] adjust read-only state check in backend:
   - https://github.com/nextcloud/spreed/blob/enh/limitWritingToMods/lib/Middleware/InjectionMiddleware.php#L174 (see https://github.com/nextcloud/spreed/blob/enh/limitWritingToMods/lib/Middleware/InjectionMiddleware.php#L139 for code for checking moderator state). 
   - I'd also recommend renaming the method to adjust its updated semantics

PVince81:
> Could you maybe look over this? Thanks! :)

Nice work so far! There's still a bit more to be done, see comments.
Feel free to ask if you need more pointers.

jakobroehrl:
> * [ ]  please search for "readonly" on the server side, there are various places to extend like for example the occ commands to allow setting the new flag

How to change it here? @PVince81 
1. https://github.com/nextcloud/spreed/blob/master/lib/Command/Room/TRoomCommand.php#L149
2. https://github.com/nextcloud/spreed/blob/master/lib/Command/Room/Update.php#L102

PVince81:
> > * [ ]   please search for "readonly" on the server side, there are various places to extend like for example the occ commands to allow setting the new flag
> 
> How to change it here? @PVince81
> 
>     1. https://github.com/nextcloud/spreed/blob/master/lib/Command/Room/TRoomCommand.php#L149
> 
>     2. https://github.com/nextcloud/spreed/blob/master/lib/Command/Room/Update.php#L102

For 1. change the boolean to int and check where the function is called.

For 2, that line is for validating the occ command input. Just add the new constant you introduced into the array.

lib/Command/Room/TRoomCommand.php
```diff
@@ -137,11 +137,11 @@ protected function setRoomPublic(Room $room, bool $public): void {
 
 	/**
 	 * @param Room $room
-	 * @param bool $readOnly
+	 * @param int $readOnly
 	 *
 	 * @throws InvalidArgumentException
 	 */
-	protected function setRoomReadOnly(Room $room, bool $readOnly): void {
+	protected function setRoomReadOnly(Room $room, int $readOnly): void {
```
PVince81, line 144: please also change the next line, you need to compare the int value now

PVince81: please recheck the whole function, it needs to be rewritten to handle the int value as it's not a bool any more

PVince81:
Further locations I've found by grepping variations of "readOnly", "read-only", "read_only" that will need adjusting:

- [ ] system messages:
   - [ ] https://github.com/nextcloud/spreed/blob/enh/limitWritingToMods/lib/Chat/SystemMessage/Listener.php#L175
   - [ ] https://github.com/nextcloud/spreed/blob/enh/limitWritingToMods/lib/Chat/Parser/SystemMessage.php#L156
- [ ] let mods add shares to rooms when readOnly==2: https://github.com/nextcloud/spreed/blob/enh/limitWritingToMods/lib/Collaboration/Collaborators/RoomPlugin.php#L63
- [ ] fix occ flag update: https://github.com/nextcloud/spreed/blob/enh/limitWritingToMods/lib/Command/Room/Update.php#L143
- [ ] occ autocomplete: https://github.com/nextcloud/spreed/blob/enh/limitWritingToMods/lib/Command/Room/Update.php#L174
- [ ] adjust middleware to check for moderator: https://github.com/nextcloud/spreed/blob/enh/limitWritingToMods/lib/Middleware/InjectionMiddleware.php#L172
- [ ] add moderator check:
   - [ ] https://github.com/nextcloud/spreed/blob/enh/limitWritingToMods/lib/Share/RoomShareProvider.php#L171
- [ ] allow mods to have a call ? https://github.com/nextcloud/spreed/blob/enh/limitWritingToMods/src/components/TopBar/CallButton.vue#L160
- [ ] flow: allow mods' flow plugin to post to mod-restricted rooms: https://github.com/nextcloud/spreed/blob/enh/limitWritingToMods/src/views/FlowPostToConversation.vue#L82
- [ ] RoomSelector (available from Deck app): not sure if we can easily check for mod status here, probably ok to leave as is: https://github.com/nextcloud/spreed/blob/enh/limitWritingToMods/src/views/RoomSelector.vue#L99

- [ ] notify mobile devs so they can implement the same client side logic
- [ ] add new capability entry + docs?
- [ ] update docs, see docs/constants.md
- [ ] integration tests

nickvergessen:
Turning into a draft for the open points to have the review/to merge list cleaner

Issue #5214, opened 2021-02-19
Don't show call indicator when moderators are having call in lobby mode


PVince81:
## Steps to reproduce
1. Create a conversation as admin
2. Add user "bob" as non-moderator
3. Enable lobby mode
4. Start a call
5. Login as "bob" in a separate window
6. Check the left sidebar

### Expected behaviour
No call indicator for the conversation

### Actual behaviour
Call indicator is visible even though "bob" cannot join it.
<img width="333" alt="image" src="https://user-images.githubusercontent.com/277525/108508293-e288d080-72bb-11eb-92ad-c93315b7f46d.png">

## Talk app

12.0.0-dev.5

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/5265

PR #5265, opened 2021-02-25
Hide call icon when blocked by lobby

PVince81:
When a user is blocked by the lobby, they now don't see the call icon in
case moderators are having one.

Fixes https://github.com/nextcloud/spreed/issues/5214


nickvergessen (commented):
Should be filtered in the API instead?
So mobile apps and dashboard also work immediately.

PVince81:
> Should be filtered in the API instead?
> So mobile apps and dashboard also work immediately.

I considered this but was unsure how. Should the API lie and set the attribute to false or should we discard the attribute altogether ?

nickvergessen:
Make the API lie. It also does when you have no session and it's a public conversation

PVince81:
> Make the API lie. It also does when you have no session and it's a public conversation

adjusted, tested and squashed

nickvergessen (approved):

Issue #3304, opened 2020-04-09
Link share discussions design


skjnldsv:
https://github.com/nextcloud/spreed/blob/9451b8078850ada2e424a4f2ed7ff0904b0a3194/lib/Settings/Admin/AdminSettings.php#L74

I think this doesn't cover most of our cases.
I just want to share a picture or a file, not discuss it, I'm pretty sure most of our users doesn't care about this.
I would also add that there is currently no way to prevent spamming so maybe anyone having the link can pollute and fill the db with messages without the user knowing about it :thinking: 

cc @jancborchardt

nickvergessen:
Additional to the global switch, I think we should also have something like mode where users can discuss the file when the owner allowed it.

jancborchardt:
@karlitschek @jospoortvliet @rullzer your input here?

My biggest issue is not that we have it (others have it too) but the way it is presented:
- There’s a step to "activate" it, "Join conversation". It’s not easy to do, you don’t want to really open a whole thing on it, you just want to leave a comment, show appreciation, give quick feedback.
- It will create a conversation in Talk, which doesn’t really make a lot of sense – instead of e.g. using comments which would show up at the relevant place in Files, also on mobile.
- When you do the "Join conversation" thing, the "Join call" button is very present, which is not really something you are likely to do at this step cause you are probably viewing the file asynchronously from the person who sent it to you.

Nextcloud step 1 | Nextcloud step 2 | Dropbox
-|-|-
![discuss before](https://user-images.githubusercontent.com/925062/86802338-38826e00-c075-11ea-901d-4e0193384404.png) | ![discuss after](https://user-images.githubusercontent.com/925062/86802336-37e9d780-c075-11ea-9364-8236c7290fbb.png) | ![Dropbox comments](https://user-images.githubusercontent.com/925062/86802324-36b8aa80-c075-11ea-90e6-bdff9f2c1edc.png)

nickvergessen:
Well the "activation" step is currently needed as otherwise your instance get's spamed with conversations easily.
Maybe we combine both problems into a solution:

On sharing you can have a checkbox to attach a talk conversation to it, which then is created automatically. When it is there, we automatically show it ("Nextcloud step 2") in the sidebar. If not, the feature is just not available.

> instead of e.g. using comments which would show up at the relevant place in Files, also on mobile.

Feel free to discuss comments for guests in https://github.com/nextcloud/server/issues/2019

rullzer:
[...]

PR #5269, opened 2021-02-26
When reselecting the chat tab, reevaluate fileinfo

PVince81:
When reselecting the chat tab, some changes might have happened in other
tabs like sharing. This refreshes the view when this happens.

### Steps to reproduce the issue

1. Upload a file
2. Open the Chat tab and see the message about sharing the file first
3. Go to sharing tab
4. Share with another user
5. Go back to the sharing tab

### Before the fix
The tab still says that you need to share. Need to refresh the page to fix it.

### After the fix
The chat tab now says "Join this conversation".

I've also checked that switching back and forth while a chat is open keeps the room open and doesn't go back to the "Join the conversation" button.

Note: I think there was a bug report somewhere but couldn't find it any more.

### DIFF
--- a/src/FilesSidebarTabApp.vue
+++ b/src/FilesSidebarTabApp.vue
@@ -135,6 +135,8 @@ export default {
 			immediate: true,
 			handler(isChatTheActiveTab) {
 				this.forceTabsContentStyleWhenChatTabIsActive(isChatTheActiveTab)
+				// recheck the file info in case the sharing info was changed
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):
Nice papercut!

nickvergessen:
> I think there was a bug report somewhere but couldn't find it any more.

Maybe https://github.com/nextcloud/spreed/issues/3304
But it's not fully solved

Issue #5027, opened 2021-01-26
In fullscreen, neither the call actions menu nor the personal controls menu works, the viewer neither


ma12-co:
In fullscreen, neither the call actions menu nor the personal controls menu works, the viewer neither.

PVince81:
probably a containment issue:
<img width="592" alt="image" src="https://user-images.githubusercontent.com/277525/109305342-dfeb2580-783d-11eb-996a-7ec77f57433b.png">

the popover element is appended into the body but the element we picked for fullscreen is not in that hierarchy

PVince81:
also broken in Talk 10.0.5

nickvergessen:
Also broken in 9
Basically since the menus use the popover and that was moved to the body instead of being in place.
But i think in reset versions the popover menu has an option where to mount it:

https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/Actions/Actions.vue#L297

If we set that to the item we fullscreen it should work again?

PVince81:
yes, that or boundariesElement.

and find a good way to set the main container globally to all vues, like the store or mixins.

nickvergessen:
> yes, that or boundariesElement.

Well additionally maybe, but it needs to "container"ed in the fullscreened element because everything else is not rendered/visible.

> and find a good way to set the main container globally to all vues, like the store or mixins.

Please don't break everything again 🙈

PVince81:
I didn't mean reparenting, only exposing a reusable value to avoid repeating "document.getElementById('content-vue')" everywhere. I noticed that "this.$root.$el" contains that element too but the first time it's evaluated it contains the old element "#content" which gets replaced after Vue finishes it initialization. So not usable in a computed property.

I guess a helper method will do, then, so that this gets reevaluated every time.

PVince81:
- [ ] Actions menu => set `container="#content-vue"` in all relevant occurrences
- [ ] Emoji picker: missing support for container property
- [ ] Avatar: missing support for container property
- [ ] Modal: missing support for container property: https://github.com/nextcloud/nextcloud-vue/blob/v3.6.0/src/components/Modal/Modal.vue#L342

PVince81:
PR for the Actions bit: https://github.com/nextcloud/spreed/pull/5270

others will need more work as they require to extend the nextcloud-vue components...

PVince81:
- [ ] check if some tooltips also need a container

PVince81:
[...]

PR #5270, opened 2021-02-26
Fix Actions menus in fullscreen mode

PVince81:
Fullscreen mode is applied on the "#content-vue" container, so now all
relevant Actions components are using that as container too.

This makes the actions menu visible also in fullscreen mode.

Partial fix for https://github.com/nextcloud/spreed/issues/5027 only for the Actions components.
Others will need nextcloud-vue updates, see https://github.com/nextcloud/spreed/issues/5027#issuecomment-786675267

Note: the property explicitly wants a String selector

### DIFF
--- a/src/components/LeftSidebar/ConversationsList/AppContentListItem/AppContentListItem.vue
+++ b/src/components/LeftSidebar/ConversationsList/AppContentListItem/AppContentListItem.vue
@@ -56,6 +56,7 @@
 		</a>
 		<Actions
 			v-if="hasActions"
+			container="#content-vue"
 			menu-align="right"
 			:aria-label="conversationSettingsAriaLabel"
 			class="actions">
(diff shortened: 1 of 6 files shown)


src/components/LeftSidebar/ConversationsList/AppContentListItem/AppContentListItem.vue
```diff
@@ -56,6 +56,7 @@
 		</a>
 		<Actions
 			v-if="hasActions"
+			container="#content-vue"
```
nickvergessen, line 59: Is the navigation inside the content-vue?

PVince81: It is, but in fullscreen mode it's hidden by a property check. Probably by design.
I added this here just as a bonus.

nickvergessen (approved):

nickvergessen:
backporting to stable21.1 as this part doesn't require a dependency update.

Issue #5097, opened 2021-02-08
Copy direct link to message


PVince81:
**Is your feature request related to a problem? Please describe.**
Ability to copy a link to message to either:
- send it through other channels to mention it to someone
- link to it as a reference to a statement in some documents
- mention in another Talk channel

**Describe the solution you'd like**
For now, add a button or a link in the message row that allows to copy the already existing message link, the one with the hash that can be seen when using the unified search.

PVince81:
There could be technical limitations to the current link format: since it uses a URL hash, the latter might get lost during a redirection for example when logging in

nickvergessen:
Also currently it's not scrolling up and loading older chunks of messages.

PVince81:
solved in https://github.com/nextcloud/spreed/pull/5285

PR #5285, opened 2021-03-01
copy msg link to clipboard

jakobroehrl:
solves: #5097

image

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -135,7 +135,7 @@ the main body of the message as well as a quote.
 							{{ t('spreed', 'Reply privately') }}
 						</ActionButton>
 						<ActionButton
-							v-if="!isSystemMessage"
+							v-if="isReplyable"
 							icon="icon-external"
 							:close-after-click="true"
 							@click.stop.prevent="handleCopyMessageLink">
@@ -153,7 +153,7 @@ the main body of the message as well as a quote.
 							</ActionButton>
 						</template>
 						<template v-if="isDeleteable">
-							<ActionSeparator />
+							<ActionSeparator v-if="isReplyable || messageActions.length > 0" />
 							<ActionButton
 								icon="icon-delete"
 								:close-after-click="true"
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -145,6 +145,13 @@ the main body of the message as well as a quote.
 							</ActionButton>
 						</template>
 						<ActionButton
+							v-if="!isSystemMessage"
+							icon="icon-external"
+							:close-after-click="true"
+							@click.stop.prevent="handleCopyMessageLink">
+							{{ t('spreed', 'Copy message link') }}
+						</ActionButton>
+						<ActionButton
 							v-if="isDeleteable"
 							icon="icon-delete"
 							:close-after-click="true"
@@ -185,6 +192,7 @@ import {
 	TOAST_DEFAULT_TIMEOUT,
 } from '@nextcloud/dialogs'
 import { createOneToOneConversation } from '../../../../services/conversationsService'
+import { generateUrl } from '@nextcloud/router'
 
 export default {
 	name: 'Message',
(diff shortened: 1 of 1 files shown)


src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -628,6 +636,16 @@ export default {
 			this.$store.dispatch('addConversation', conversation)
 			this.$router.push({ name: 'conversation', params: { token: conversation.token } }).catch(err => console.debug(`Error while pushing the new conversation's route: ${err}`))
 		},
+
+		async handleCopyMsgLink() {
```
nickvergessen, line 640: No need to shortcut words

nickvergessen:
So with more and more actions (delete, deck-integration, reply-private, copy link) I think the most used action (reply) is getting moved to far away as you always need to open the ... first.
cc @ma12-co @jancborchardt any clever ideas how we can make replying quick again?

PVince81:
> So with more and more actions (delete, deck-integration, reply-private, copy link) I think the most used action (reply) is getting moved to far away as you always need to open the ... first.
> cc @ma12-co @jancborchardt any clever ideas how we can make replying quick again?

we probably need a second button just for replying.
and the three dots menu directly after that

jancborchardt:
> So with more and more actions (delete, deck-integration, reply-private, copy link) I think the most used action (reply) is getting moved to far away as you always need to open the ... first.

@nickvergessen  Thought exactly the same, yep.

> we probably need a second button just for replying.
> and the three dots menu directly after that

@PVince81 exactly that – just like we do in files too, or e.g. in the left navigation of calendar, in both cases where "Share" is a direct action. In this case "Reply" will be the direct action, and everything else is details.

ma12-co:
Agree, should we create a separate ticket for this?

nickvergessen:
yeah, it shouldn't block a "random" pr adding another action

nickvergessen:
Reported https://github.com/nextcloud/spreed/issues/5297

jakobroehrl:
added it to the new action menu:
image

PVince81:
nice!

I'm starting to wonder if we should start to add some separators in there, maybe to separate the hard-coded actions from the registered ones.

@ma12-co any thoughts on this ? (see above screenshot)

ma12-co:
Yep, we could have something like:

talk acitons
---
other apps actions 
---
delete

PVince81:
@jakobroehrl do you want to take care of the separators or is this something for another PR ?

note: there are like more actions coming

jakobroehrl:
> @jakobroehrl do you want to take care of the separators or is this something for another PR ?
> 
> note: there are like more actions coming

I will try the seperators here in this PR

Could you say me the actions, maybe I can help with the easy ones?

PVince81:
> Could you say me the actions, maybe I can help with the easy ones?

I have a PR that will bring "mark as unread" here later on: https://github.com/nextcloud/spreed/pull/3825.
I thought there were other already open PRs with more actions, not sure. In any case if we add separators as @ma12-co we can insert the new actions in the places that make sense between/after/before the separators.

jakobroehrl:
supernice :)
Let's merge it?
image

jakobroehrl:
> 
> I have a PR that will bring "mark as unread" here later on: #3825.

@PVince81 Maybe I can do this? Please ping me if it's easy and when I can start

PVince81:
> > I have a PR that will bring "mark as unread" here later on: #3825.
> 
> @PVince81 Maybe I can do this? Please ping me if it's easy and when I can start

@jakobroehrl I already created the action there, thanks :-)

The only thing you can do right now is add the separators as @ma12-co advised. I'd say put them in this PR here.
Future work can then decide where to put new actions.

PVince81:
okay, you added already. let me review!

PVince81 (approved):
:+1:

ma12-co (commented):
Nice work @jakobroehrl :)
One question, I'm not able to copy links of my own messages, I think it would be nice to be able to do so too. Is there a particular reason why this wasn't implemented? 

As a side note, I also think that we should use var(--color-primary-light) for the fading feedback on the target message to give some more emphasis and differentiate it from the messages hover/focus feedbacks.

jakobroehrl:
> Nice work @jakobroehrl :)
> One question, I'm not able to copy links of my own messages, I think it would be nice to be able to do so too. Is there a particular reason why this wasn't implemented?

Thanks! :)
That's a bug, fixed it. 

> As a side note, I also think that we should use var(--color-primary-light) for the fading feedback on the target message to give some more emphasis and differentiate it from the messages hover/focus feedbacks.

Sorry, what do you mean here? What exactly should I change?

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -144,13 +152,15 @@ the main body of the message as well as a quote.
...
 						</template>
-						<ActionButton
-							v-if="isDeleteable"
-							icon="icon-delete"
-							:close-after-click="true"
-							@click.stop="handleDelete">
-							{{ t('spreed', 'Delete') }}
-						</ActionButton>
+						<template v-if="isDeleteable">
+							<ActionSeparator />
```
nickvergessen, line 156:
```suggestion
							<ActionSeparator v-if="isReplyable || messageActions.length > 0" />
```

Otherwise you might start with a separator at some point :P

nickvergessen (approved):

ma12-co:
> That's a bug, fixed it.

Noice :) 

> Sorry, what do you mean here? What exactly should I change?


Basically those 3 variables in the animation into `var(--color-primary-light)`

https://github.com/nextcloud/spreed/blob/master/src/components/MessagesList/MessagesGroup/Message/Message.vue#L750

nickvergessen (approved):

PR #5291, opened 2021-03-02
Use container property to fix popups in fulllscreen mode

PVince81:
Many popup components are usually appending themselves to the document
body. This doesn't work in fullscreen mode where we use another
component as the root.

This fix sets the "container" property for all relevant popup components
to make sure they can be visible in fullscreen mode.

- [x] REQUIRES: https://github.com/nextcloud/nextcloud-vue/pull/1734
- [x] nextcloud-vue release + update

Partial fix of https://github.com/nextcloud/spreed/issues/5027


PVince81:
rebased to pick up library update

PVince81:
please review

ma12-co (approved):

Issue #5297, opened 2021-03-03
Give direct access to replying


nickvergessen:
As per https://github.com/nextcloud/spreed/pull/5285#issuecomment-788769422 and followup comments replying should be directly accessible.

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/5302

PR #5302, opened 2021-03-03
Direct reply button in message row

PVince81:
Moved reply button outside the actions menu in the message row.
This way it is quicker for the user to find and click.

Fixes https://github.com/nextcloud/spreed/issues/5297

Regular row:
image

Tall row:
image

As menu:
image

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -111,43 +111,47 @@ the main body of the message as well as a quote.
 						:size="16" />
 				</div>
 				<!-- Message Actions -->
-				<Actions
-					v-show="showActions && hasActions"
+				<div
+					v-show="showActions"
 					class="message__main__right__actions"
-					container="#content-vue"
 					:class="{ 'tall' : isTallEnough }">
-					<ActionButton
-						v-if="isReplyable"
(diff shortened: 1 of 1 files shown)


jakobroehrl (approved):
nice!

nickvergessen:
So when delete is the only option, can we still put it in a ... menu?
I think otherwise the immediately visible red icon is way to alerting

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -111,43 +111,47 @@ the main body of the message as well as a quote.
...
 						<ActionButton
-							:key="action.label"
-							:icon="action.icon"
+							icon="icon-reply"
+							@click.stop="handleReply">
+							{{ t('spreed', 'Reply') }}
+						</ActionButton>
+					</Actions>
+					<Actions
+						v-show="hasActionsMenu"
```
nickvergessen, line 127:
```suggestion
						v-show="hasActionsMenu"
						:force-menu="true"
```

PVince81:
> So when delete is the only option, can we still put it in a ... menu?
> I think otherwise the immediately visible red icon is way to alerting

Makes sense. That bit also disturbed me and I wasn't sure. At least it will prevent accidental clicks.

Committed, please review.

PR #5322, opened 2021-03-05
Don't redirect to not-found page if user switched away

PVince81:
Whenever a request is running in the background and might detect
that the current conversation or session is gone, it would prompt the
app to redirect to the not found page.

In some cases, the user already switched away to another page while the
request was running. The fix detects such situations and skips the not
found page redirect.

Fixes https://github.com/nextcloud/spreed/issues/4670

- [ ] BUG: switching back and forth quickly will confuse the new logic which will think that we're still in the old instance/session and would redirect to not-found

### DIFF
--- a/src/utils/signaling.js
+++ b/src/utils/signaling.js
@@ -448,7 +448,7 @@ Signaling.Internal.prototype._startPullingMessages = function() {
 			} else if (error.response && (error.response.status === 404 || error.response.status === 403)) {
 				// Conversation was deleted or the user was removed
 				console.error('Conversation was not found anymore')
-				EventBus.$emit('deletedSessionDetected')
+				EventBus.$emit('deletedSessionDetected', { token })
 			} else if (token) {
 				if (this.pullMessagesFails === 1) {
 					this.pullMessageErrorToast = showError(t('spreed', 'Lost connection to signaling server. Trying to reconnect.'), {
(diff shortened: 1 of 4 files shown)


src/utils/signaling.js
```diff
@@ -496,6 +516,23 @@ Signaling.Internal.prototype.sendPendingMessages = function() {
 	}.bind(this))
 }
 
+Signaling.Internal.prototype._leaveRoomBefore = function(token) {
+	console.log('Signaling.Internal._leaveRoomBefore() token=' + token +
+		' currentRoomToken=' + this.currentRoomToken + ' pullMessagesRequest.token=' + this.pullMessagesRequest?.token)
+	// note: we can't rely on this.currentRoomToken because in some cases it's already cleared
+	if (this.pullMessagesRequest?.token === token) {
+	//if (this.pullMessagesRequest) {
```
PVince81, line 524: using this one seems to make the 404 errors disappear completely in my tests

but there's a risk that we might cancel too eagerly and find ourselves in a situation where no more messages are getting pulled

nickvergessen:
What's the plan here?

PVince81:
I'm closing this for now since this hasn't been reported again and  https://github.com/nextcloud/spreed/issues/4670 is closed.

Now, there might still be a possibility to make this fail by switching conversations like crazy but this usually isn't an actual use case and any of the above fixes or proposed fixes are currently too risky. High risk vs low probability of occurrence => closing.

In case this pops up again we could reopen the case...

Also, it was said that we need to refactor/rework the signaling code anyway, so whatever comes next needs to be a bit more robust.

Issue #760, opened 2018-03-30
Sound notification for chat message


BloodyIron:
Can we get a sound to play when someone types a message in the text chat portion of Talk please? It's so easy to miss that someone has typed a message right now!

GoetheG:
This isna good idea. I really hope that you Nextcloud guys consider just adding Rocket.Chat into Nextcloud or donsomething together or whatever.

BloodyIron:
Please don't hijack this ticket/issue...

nickvergessen:
Sorry sound playing is not possible without user interaction, see https://github.com/nextcloud/notifications/pull/117

BloodyIron:
Hi @nickvergessen ,

I've found a bunch of resources that talk about Sound in Push Notifications as part of the standard:
* https://web-push-book.gauntface.com/demos/notification-examples/

(many more elsewhere)

However, I do agree it seems like current browsers don't support the sound notification. And I can't find why, or when this will change.

Is there any way we could keep this issue open waiting for browsers to gain support for this? Because I really want this to eventually be added, as having non-audible notifications for spreed messages and such is really easily ignored/accidentally missed.

nickvergessen:
Sadly it's the other way around,
Notification.sound was part of the standard and has been deprecated, that's why it's listed in "Obsolete handlers" :crying_cat_face: 
https://developer.mozilla.org/en-US/docs/Web/API/Notification#Obsolete_handlers

BloodyIron:
So what is the long-term way we (as a society, not even just the scope of this discussion) can address this functional need?
What about using WebRTC for "notification"-esque outcomes? Ala Discord/Riot

jookk:
sound notifications are missing on new messages. Now messaging and no new msg notification 👎  
possible someway?

nickvergessen:
There are text notifications and browser/os notifications.

Sound is not possbile because Firefox and Chrome (both supported browsers atm) prevent playing audio without direct user interaction.

alauzon:
FaceBook does it with Messenger. Why can't we do it also?

honzapatCZ:
Bump, How can facebook do it while we cant?

thecyberd3m0n:
there is no sound even if tab is opened, and loaded. I can leave opened nextcloud and allow permission, but please - just play some sound when something happens. Not from background - just normally.

nickvergessen:
From background means without the user pressing a play button.
[...]

Issue #5096, opened 2021-02-06
Web notifications, supposedly disallowed technique


honzapatCZ:
So the latest issue was closed, because "Evil browsers have some list of allowed sites or whatever", while I really tried, I searched thru the source of chromium and didnt find any, closest was this: https://github.com/chromium/chromium/blob/master/third_party/blink/renderer/core/html/media/autoplay_policy.cc
But thats normal policy that can be changed by the user the same as notifications at: chrome://settings/content/sound
And even that occurs only if the user didnt interact with the website at all, while if you have opened the messenger in some tab, you probably clicked thru some conversations as this is mainly to avoid spam websites that play crazy sounds and scams on load.

nickvergessen:
Closing as duplicate of #760 so the info is not distributed across multiple reports.

Issue #4651, opened 2020-11-19
Sound alarm on receiving new message - option: turn on/off etc.


pongraczi:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when [...]  
I have an open talk/spreed window on my browser and I just miss new messages, because only a little visual indicator appears in case of new incoming message.  

**Describe the solution you'd like**
* Audio alarm: I would like to hear an audible alarm (chime or something) which could alarm me to check the new message.  
  * Options: 
    * replace the default alarm sound from file in nextcloud
    * different alarm sound for different chatrooms
    * mute/unmute alarm globally (permanently, 10mins, 1hour, 1day, 1week,  every day between 20:00-08:00 etc.)
    * mute/unmute alarm for each chatroom (permanently, 10mins, 1hour, 1day, 1week,  every day between 20:00-08:00 etc.)
* Visual popup notice if possible from the browser, similar to the notification of the nextcloud desktop client.
* Optionally the desktop client could show a notification when new message arrives into a chatroom the user subscribed.

**Additional context**
So, the point is to get immediately a strong notification to do not miss important messages from spreed.

* Please note: I do not speak about android clients, the desktop client is optional as not every user install desktop clients.

nickvergessen:
> * Visual popup notice if possible from the browser, similar to the notification of the nextcloud desktop client.
> * Optionally the desktop client could show a notification when new message arrives into a chatroom the user subscribed.

You should already see popup notifications from the browser when you are mentioned in a non-active chat. For the currently active chat we are still unsure how to solve this in a good and still meaningful way: https://github.com/nextcloud/spreed/pull/1970
So what you are missing is to change the notification level to "All messages" in the left sidebar.
This would then also work for the desktop client.

As for audio alarm: it's on our todo to look into this once browsers are allowed to play sounds in background again without the need of user interaction. However I think we would start with sounds on calls then.

pongraczi:
> You should already see popup notifications from the browser when you are mentioned in a non-active chat. For the currently active chat we are still unsure how to solve this in a good and still meaningful way: #1970
> So what you are missing is to change the notification level to "All messages" in the left sidebar.
> This would then also work for the desktop client.

Oh boy, I feel stupid, you are absolutely right, the popup notification is working well in case (notification set to **All messages** as you mentioned):
* the message received into an inactive chat and I have an other browser tab active (like I edit this message, while I got a new message to an inactive chat)
[...]

PR #5335, opened 2021-03-06
Add basic sound notification

d0niek:
Sound file was downloaded from [here](https://notificationsounds.com/message-tones/that-was-quick-606).

It's a basic sound notification that I could make with my skills.
Probably it's not the perfect but it works and my users are happy.

I didn't know how to add new feature so I modified branche with my
installed version of Nextcloud Talk.

Resolves: #760, #5096, #4651


nickvergessen:
Thanks for the start.

Unluckily it's not easy like that. I just recently gave it another try in general:
https://github.com/nextcloud/notifications/issues/116#issuecomment-777279834

So to support all supported browsers and also work in background etc a bit more work is necessary.
Similarly it definetly needs a setting to disable it and also should only play the sound on one tab.

I have a local branch which brings in a audio setting for sounds in calls. I will push it on monday. We then can take your changes onto the branch controlled by the setting and continue from there.

d0niek:
As I follow sound notification topic in Nextcloud and understand it that you're going to implement sound play for Nextcloud notifications (top-right bell icon)? It would be nice but Facebook or Google don't have it and people can live with it. Advantage of Facebook or Google (Gmail) is that when someone text you then small window with chat is opening (sound is playing so you won't miss a message). Maybe one day I will browse my files in Nextcloud and same small window with chat will open? :-)

Nextcloud Talk (web) is like Facebook Messenger (www.messenger.com), Google Hangout (hangouts.google.com), Slack and many other chat applications which have sound notifications for all new messages, comming calls, etc. Nextcloud Talk is adding only * in tab title and that's why I added here a sound play. But it has a bug (?) because it happens only if window is not visible and it happens only once. If window is visible but browser hasn't focus, * won't be added to tab title and my sound won't play. If window is not visible and I receive more then one new message, event with adding * (playing sound) will happens only once.

Another thing is that when I have Nextcloud Talk tab open and someone start calling then all my devices are ringing (web browser, phone, tablet) and it's ok, I won't miss a call ;-) The same is with Facebook or Google. Now when I have Nextcloud Talk tab open but is not visible and someone text me, I miss my sound notification and someone text me again I will never know about it because my phone/tablet don't ring. Mobile devices only ring when Nextcloud Talk tab is not open or if a video call is comming.

Maybe you're right that I should used https://github.com/goldfire/howler.js to play a sound but I'm not a front-end guy and I don't want to make a mess. I prefer another PR from vue.js master guys which will upgrade my implementation. I just wanted to check how hard (easy) is to add sound notification. My users are so happy with this upgrade and I have a hope that more users can be happy too and sound notifications can be only better from this point :-)

nickvergessen:
Finally managed to push my changes. You can find them in https://github.com/nextcloud/spreed/pull/5410 and it comes with generic sound settings. I guess sounds for chat messages require a second setting as you might not want that while still having sounds in calls, but it should be possible to duplicate the changes for the setting easily.

PR #5355, opened 2021-03-09
Use actions for screensharing menu

ma12-co:
WIP, still to do:
- [x] Jumping container glitch;
- [x] Cleanup unused code;

### DIFF
--- a/src/components/CallView/shared/LocalMediaControls.vue
+++ b/src/components/CallView/shared/LocalMediaControls.vue
@@ -833,4 +833,14 @@ export default {
 		height: $clickable-area;
 	}
 }
+
+::v-deep button.action-item,
+::v-deep .action-item__menutoggle {
+	// Fix screensharing icon width
+	&:hover,
+	&:focus,
+	&:active {
+		background-color: transparent;
+	}
+}
 </style>
--- a/src/components/CallView/shared/LocalMediaControls.vue
+++ b/src/components/CallView/shared/LocalMediaControls.vue
@@ -68,55 +68,60 @@
 					fill-color="#ffffff"
 					decorative />
 			</button>
-			<button
+			<Actions
 				v-if="!screenSharingButtonHidden"
 				id="screensharing-button"
 				v-tooltip="screenSharingButtonTooltip"
 				:aria-label="screenSharingButtonAriaLabel"
 				:class="screenSharingButtonClass"
 				class="app-navigation-entry-utils-menu-button"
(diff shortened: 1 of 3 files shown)


PVince81:
the glitch is a known issue: https://github.com/nextcloud/nextcloud-vue/issues/1610

it already existed before but was barely noticeable.

ma12-co:
ready for review @PVince81

src/components/CallView/shared/LocalMediaControls.vue
```diff
@@ -68,55 +68,59 @@
...
-			<button
+			<Actions
 				v-if="!screenSharingButtonHidden"
 				id="screensharing-button"
 				v-tooltip="screenSharingButtonTooltip"
 				:aria-label="screenSharingButtonAriaLabel"
 				:class="screenSharingButtonClass"
 				class="app-navigation-entry-utils-menu-button"
-				@click.stop="toggleScreenSharingMenu">
+				boundaries-element="boundaryElement"
```
PVince81, line 78: you'll also need `container="#content-vue"` to make it work full screen, see other actions

src/components/CallView/shared/LocalMediaControls.vue
```diff
@@ -844,4 +832,14 @@ export default {
 		height: $clickable-area;
 	}
 }
+
+::v-deep button.action-item,
+::v-deep .action-item__menutoggle {
+	// Fix screensharing icon width
```
PVince81, line 838: is this an upstream issue to fix separately ?

on the other hand, once we switch to material design icons here this might disappear.

note that I had a quick look back then and I'm not sure we'll find icons that match all these menu actions

PVince81:
The icons look a bit brighter and almost makes them look disabled, not sure if intended or side effect.

Before:
image

After:
image

And here's the hand icon which is even darker, but also a MD icon:
image

src/components/CallView/shared/LocalMediaControls.vue
ma12-co: Once we'll the button component ready, covering all use cases, and actions will use it, we will be able to get rid of this I think

ma12-co:
> The icons and text look a bit brighter and almost makes them look disabled, not sure if intended or side effect.

This is the same problem we're seeing in actions in other apps, we'll have to figure out a global solution here

PVince81 (approved):
:+1: looks good now

Issue #5375, opened 2021-03-17
mark as unread: conversation list update


jakobroehrl:
## Steps to reproduce
1. set a message to unread (possible on the latest master)
2. Wait 24 sec (on my system) until the conversation is getting bold and shows the unread counter

### Expected behaviour
It should be bold within 1-2 secs
The conversation should go to the top of the list

PVince81:
Even though we update the conversations store, the left sidebar doesn't seem to react

I even tried to switch from `Vue.set()` with a full fledge `commit('addConversation', conversation)` like we do for other code paths, it still doesn't refresh the left sidebar.

Something seems broken with the reactivity on this side.

PVince81:
Aha! got it... we'd also need to update the value of `unreadMessages` and potentially also `unreadMention'.
For the former it means we need to count how many messages are present in the array starting from the current position.
Not sure if worth the effort, maybe we could just trigger `fetchConversation` again earlier for that entry so the server can compute more accurate values.

PVince81:
PR here https://github.com/nextcloud/spreed/pull/5378

PR #5378, opened 2021-03-18
Faster update of left sidebar after marking as unread

PVince81:
This fix re-fetches the current conversation to make sure that harder to compute values are re-fetched from the server like unreadMessages, unreadMention, etc.
    
This introduces a new store action "fetchConversation".
I've adjusted some of the service's "fetchConversation" usage and moved them to use this new store action instead.
(moving the non-store stuff step by step to the store to avoid repeating logic)

Fixes https://github.com/nextcloud/spreed/issues/5375

While testing, I noticed unrelated console errors for guests due to unread marker. I'll fix this separately

### DIFF
--- a/src/store/messagesStore.js
+++ b/src/store/messagesStore.js
@@ -401,7 +401,10 @@ const actions = {
 			context.commit('setVisualLastReadMessageId', { token, id })
 		}
 
-		await updateLastReadMessage(token, id)
+		if (context.getters.getUserId()) {
+			// only update on server side if there's an actual user, not guest
+			await updateLastReadMessage(token, id)
+		}
 
 		if (updateVisually) {
 			// refetch to get more accurate values for unreadMessages, unreadMention, etc
--- a/src/store/messagesStore.js
+++ b/src/store/messagesStore.js
@@ -386,7 +386,7 @@ const actions = {
 	 * @param {bool} updateVisually whether to also update the marker visually in the UI;
 	 */
 	async updateLastReadMessage(context, { token, id = 0, updateVisually = false }) {
-		const conversation = Object.assign({}, context.getters.conversations[token])
+		const conversation = context.getters.conversations[token]
 		if (!conversation || conversation.lastReadMessage === id) {
 			return
 		}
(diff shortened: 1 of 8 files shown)


PVince81:
since the guest part fix would cause conflicts, I've decided to push it here: https://github.com/nextcloud/spreed/pull/5378/commits/8037d495f73984d13f7ddec0357e2a9405210bdf

it simply skips the server-side update of read marker in guest mode

src/store/messagesStore.js
```diff
@@ -400,7 +400,16 @@ const actions = {
...
-		await updateLastReadMessage(token, id)
+
+		if (context.getters.getUserId()) {
+			// only update on server side if there's an actual user, not guest
+			await updateLastReadMessage(token, id)
+		}
+
+		if (updateVisually) {
+			// refetch to get more accurate values for unreadMessages, unreadMention, etc
+			context.dispatch('fetchConversation', { token })
```
nickvergessen, line 411: this will bring down our server in the company call
If someone writes a message everyone will mostly simultaniously pull the room endpoint to update, or when is `updateVisually` set to true?

PVince81: `updateVisually` in this code path is only true for direct forcing-actions like here "mark as unread".
the actual visual update is happening in a separate action

but you're right to be careful, I'll double check to make sure it's not called more than that

PVince81: I've checked and while it wouldn't be called too often, it still would be called when posting a new message because that's the moment where we update the indicator visually also.

I've now moved the `fetchConversation` to only happen when using "mark as unread", as it's the only situation where we'd have non-zero `unreadMessages` after an action.

jakobroehrl (dismissed):
It switches immediately to unread but the conversation doesn't got to the top

nickvergessen:
> It switches immediately to unread but the conversation doesn't got to the top

We don't sort by unread, so that's expected?

jakobroehrl:
> > It switches immediately to unread but the conversation doesn't got to the top
> 
> We don't sort by unread, so that's expected?

Hmm, you are right, I think. Let's merge it in.

nickvergessen (approved):
🐘

Issue #5372, opened 2021-03-16
Not sent message gets lost if call is ended


FlexW:
## Steps to reproduce
1. Write a message while in a call
2. Don't send that message
3. Someone ends the call
4. Message disappears

### Expected behaviour
The unsent message should not disappear.

### Actual behaviour
Message disappears.

## Talk app

Don't know. The issue appeared on cloud.nextcloud.com

## Browser

**Microphone available:** yes

**Camera available:** yes

**Operating system:** ArchLinux

**Browser name:** qutebrowser

**Browser version:** 2.1.0

## Server configuration

**Version:** 21.0.0.8

PVince81:
this is likely because the message form is situated in the sidebar and then gets destroyed when the call ends, and then a new message form gets rendered in the regular message list

as possible solution is storing the currently typed message in the vuex store instead of a local attribute, and reuse/restore it from there when recreating the field

nickvergessen:
But it shouldn't happen unless you are kicked out leave the call yourself

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/5379

PR #5379, opened 2021-03-18
Keep pending input message per conversation

PVince81:
### Description

Save the message from the current input field by conversation into the
store.

This makes it possible to keep the message when switching the
conversation.

Also fixes an issue where joining or leaving a call would clear the
message due to the field being destroyed and recreated between the
message list and the sidebar.

### Related issue

Fixes https://github.com/nextcloud/spreed/issues/5372

#### Cursor position

There's a small issue: when the field is populated, the cursor appears at the beginning.
Given [how complicated it is to set the cursor at the end](https://stackoverflow.com/questions/4233265/contenteditable-set-caret-at-the-end-of-the-text-cross-browser) we might want to pass on that one, considering that we'll also replace the field with a better version of it in https://github.com/nextcloud/spreed/pull/4333

#### Store module name

The store module is still called "quoteReplyStore", we could rename to "currentMessageStore" or something similar. Not sure if it's the right time now.

### Tests

- [x] TEST: switching to read-only channel like changelog doesn't log errors and the field stays empty
- [x] TEST: switch to conversation A, type text "abc", switch to B, type text "def", switch back to A => the field shows "abc"
- [x] TEST: type text "abc", start call => message stays in sidebar field
- [x] TEST: start call, type text "abc", leave call => message stays in message list field
- [x] TEST: quoted reply is also kept between conversation switch and starting/leaving a call
- [x] TEST: emojis also kept

### DIFF
--- a/src/components/NewMessageForm/NewMessageForm.vue
+++ b/src/components/NewMessageForm/NewMessageForm.vue
@@ -230,11 +230,24 @@ export default {
 		disabled(newValue) {
 			this.$refs.uploadMenu.$refs.menuButton.disabled = newValue
 		},
+
+		text(newValue) {
+			this.$store.dispatch('setCurrentMessageInput', { token: this.token, text: newValue })
+		},
+
+		token(token) {
+			if (token) {
+				this.text = this.$store.getters.currentMessageInput(token) || ''
+			} else {
+				this.text = ''
+			}
+		},
 	},
 
 	mounted() {
 		EventBus.$on('uploadStart', this.handleUploadStart)
 		EventBus.$on('retryMessage', this.handleRetryMessage)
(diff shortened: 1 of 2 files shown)


src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -230,11 +230,24 @@ export default {
 		disabled(newValue) {
 			this.$refs.uploadMenu.$refs.menuButton.disabled = newValue
 		},
+
+		text(newValue) {
```
PVince81, line 234: not sure if we should debounce. the current approach seems not laggy

ma12-co: No, I don't think it's needed here

ma12-co (approved):

PR #5396, opened 2021-03-22
Temporary OCS API for user avatar upload and delete

nickvergessen:
### Upload an avatar
```
curl -k -H "OCS-APIRequest: true" -u admin:admin 'https://nextcloud22.local/ocs/v2.php/apps/spreed/temp-user-avatar' -X POST -F 'files[]=@/home/nickv/avatar.jpg'
```

### Delete an avatar
```
curl -k -H "OCS-APIRequest: true" -u admin:admin 'https://nextcloud22.local/ocs/v2.php/apps/spreed/temp-user-avatar' -X DELETE
```

Not adding documentation because it's being wiped soon again and will be removed without any close done time once the one from server is there.

### DIFF
--- a/lib/Controller/TempAvatarController.php
+++ b/lib/Controller/TempAvatarController.php
@@ -1,0 +1,149 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2016 Lukas Reschke <lukas@statuscode.ch>
+ *
+ * @author Lukas Reschke <lukas@statuscode.ch>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Talk\Controller;
+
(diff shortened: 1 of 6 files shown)


PVince81:
for the record, the real deal will be https://github.com/nextcloud/server/pull/24579

PVince81:
Tested both endpoints and they worked fine with both jpg and png.

However, if you upload a non-square image there is no useful error message in the response:
```
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>failure</status>
  <statuscode>400</statuscode>
  <message/>
 </meta>
 <data>
  <message>An error occurred. Please contact your admin.</message>
 </data>
</ocs>
```

I had to look at the nextcloud.log to see what the problem was.
As far as I saw in the code we wanted to copy the messages there. (could also be a bug in the old controller)

PVince81 (approved):
:+1: looks fine, see my previous comment about error messages

lib/Controller/TempAvatarController.php
```diff
@@ -0,0 +1,149 @@
...
+					['data' => ['message' => $this->l->t('Unknown filetype')]],
+					Http::STATUS_BAD_REQUEST
+				);
+			}
+
+			$avatar = $this->avatarManager->getAvatar($this->userId);
+			$avatar->set($image);
+			return new DataResponse();
+		} catch (\Exception $e) {
+			$this->logger->error('Failed to delete avatar', [
```
nickvergessen, line 123: this is post not delete, copy paste error in here

lib/Controller/TempAvatarController.php
```diff
@@ -0,0 +1,149 @@
...
+				'exception' => $e,
+			]);
+
+			return new DataResponse(['message' => $this->l->t('An error occurred. Please contact your admin.')], Http::STATUS_BAD_REQUEST);
```
nickvergessen, line 127: I guess we can forward the error from the avatar manager as well? That could solve the issue @PVince81 mentioned

PVince81:
psalm not happy, might need to add to baseline as we're using the private `Filesystem` class for the blacklist check

nickvergessen:
yeah, I bumped the baseline.

PVince81 (approved):
:+1:

PR #5397, opened 2021-03-22
Fix auto-merge

PVince81:
Increase module version.

Inspired by https://github.com/nextcloud/nextcloud-vue/pull/1777

@nickvergessen but I heard you say we don't want auto-merge any more. If that's the case I can remove the whole block instead.

### DIFF
--- a/.github/workflows/dependabot-approve.yml
+++ b/.github/workflows/dependabot-approve.yml
@@ -12,7 +12,7 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
       # Nextcloud bot approve and merge request
-      - uses: ahmadnassri/action-dependabot-auto-merge@v1
+      - uses: ahmadnassri/action-dependabot-auto-merge@v2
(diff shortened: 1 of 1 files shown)


nickvergessen:
If the following libs can be excluded I'm fine:
* attachmediastream
* hark
* sdp-transform
* webrtc-adapter
* webrtcsupport

PVince81:
seems it might be possible, but it's more of an inclusion so would need to find the right pattern https://dependabot.com/docs/config-file/#automerged_updates

I personally dislike auto-merge as it's asking for trouble especially for potentially broken deps.
Would prefer to remove it completely and let use have a look at the deps, changelog, etc.

nickvergessen:
Fine by me

PVince81:
goodbye automerge https://github.com/nextcloud/spreed/pull/5397/commits/5a99c3435d4b6dbc9c16ba2112e96ee0c8ff838f

nickvergessen (approved):

PR #5410, opened 2021-03-24
Play sounds on call actions

### LINKED ISSUE #1072: Add acoustic feedback for a call actions
As @mario stated, that's a task for you: https://github.com/nextcloud/talk-android/issues/254

It would be great to have the following acoustic feedback as a normal phone or Skype does :)    

- Play a ringing sound when a call is initiated until the call receiver picks ups or joins the call
- When a user hangs up or a call ends, play an appropriate sound

### Actual behaviour
Currently when Alice calls Bob with NextCloud Talk, the screen simply stays black until Bob joins; or when Bob leaves, Alice won't be notified about it. That's a pretty bad UX as I figured out with friends.

### Naive suggestion

Could it not be solved by a simple status message system?

When Alice calls Bob and Bob's phone rings, Bob's phone could send periodically status messages to Alice, telling her that his phone rings.

If no status message is received by Alice, which indicate that the phone rang, the status is however unclear, but it could be assumed that Bob is offline.

During a call both could send keep alive messages to make clear that no one has gone offline (due to a bad mobil connection eg).

Those status messages could also be applied to the "hung up" action.

Or let's assume Alice and Bob are in a call and Charlie calls one of them. Then Charlie could be notified that Alice and Bob are currently not available, while Alice or Bob hears the knock sound that someone else is calling.

That's of course a very simple and naive idea; if you can think of an better solution I would very much welcome that of course :-)

### LINKED ISSUE #2345: acoustic feedback when a user joins/leaves a call
A few different use cases:

1. starting One-to-one call: I start a call and it it is silent until the recipient picks up and start talking. That can be really surprising/irritating. A "ringing" sound on the caller side like your know it from phones would be nice so that you have a indicator that the phone on the other side rings and a indicator when the recipient joined.

2. staring Group calls: A acoustic signal when someone joins the group call, like you know it from most other online meeting tools would be nice. This way you know someone joined, can pause the conversation and allow the new participant to say "hello".

3. ending group or one-to-one calls: If someone leaves a call the other people should get a acoustic feedback that someone left. It happens to me already a few times that I continued talking while my chat partner left (by accident) and we only found out when he re-joined. This could be avoided by a acoustic feedback.

@tobiasKaminsky just for your information, as we discussed this recently

nickvergessen:
Fix #2345
Fix #1072

(as discussed with @jancborchardt ):
- [x] Play a sound when someone joins a call (only if not played within the last 5 seconds)
- [x] Play a sound when someone leaves a call (only if not played within the last 5 seconds)
- [x] Play a waiting sound when you are alone in a call (only 3 times)
- [x] Have a setting to opt out of noises completely
- [ ] Check if we can play a sound when a call comes in, or are browsers too annoying in blocking it? (ref https://github.com/nextcloud/spreed/pull/5335#issuecomment-791932318 + https://github.com/nextcloud/notifications/issues/116#issuecomment-777279834 )
- [x] Add a note that sounds can't play on safari atm
- [x] Check CSP changes


ma12-co (commented):
> within the last 5 seconds 

Isn't it too much?

ma12-co:
> Check if we can play a sound when a call comes in, or are browsers too annoying in blocking it? 

We should probably leave it till there's a more reliable way to do this? 
I mean, If you know as a user that the you should expect a sound, you might start to rely on it as a notification, but if things are so precarious there we might just create confusion by doing this

nickvergessen:
> > within the last 5 seconds
> 
> Isn't it too much?

The sound file is already ~3 sec long, so if it starts again after 5 seconds, there were 2 seconds of break. If you think about bigger calls this already sounds problematic enough. But that is why @jancborchardt and me picked 5 seconds in the end

nickvergessen:
- [ ] Add a note that sounds can't play on safari atm

nickvergessen:
Ready for final testing from my side

ma12-co (approved):
In as is as discussed but I'd like to leave two comments here: 

* IMO sounds are too long and we should cut them, they should be snap like long, not several seconds long;
* I find the sound when starting the call a bit annoying, I'd remove it altogether;

nickvergessen:
- [ ] Sound also plays when a participant not in the call leaves the room

nickvergessen:
Fixed the problem with leaving as a chat-user only

PR #5421, opened 2021-03-26
Fix migration checking for unique constraint violation

nickvergessen:


ChristophWurst (approved):
Beautiful

PR #5453, opened 2021-04-07
Don't fail loading the store when some initial states are not there

nickvergessen:
Copied from https://github.com/nextcloud/server/blob/3e624dc39966f54c6d418eee2e8141fc5c21f751/core/src/login.js#L48-L54

Don't know when it started, but the current admin page doesn't load anymore due to combination of things. services now use the store and the store loads initial state, but that is only initialised on the talk controllers, not in settings.... going with a default value now solves this.


danxuliu (approved):
Tested and works :+1: 

For the record, it started with 82d0d83885 when the store was imported in `settingsService.js`, which is used both by the settings dialog in the regular Talk UI and the admin settings (specifically, the SIP bridge settings).

Issue #5349, opened 2021-03-09
"Raised hand" state not working properly in the sidebar with multiple sessions


nickvergessen:
## Steps to reproduce
1. Join a conversation with 2 sessions
2. In the session with the lower id raise hand
3. 

### Expected behaviour
Sidebar shows you raise your hand

### Actual behaviour
Sidebar doesn't show you raise your hand

PVince81:
When I tested it worked fine with internal signaling, but not with the HPB.
It was likely related to the way session ids are mapped.

PVince81:
I did some quick debugging and it looks like it's going to be tricky.
The results of the participant list only contain one session id (likely the one from the aggregated "max" participant result).

Maybe we should adjust the participant API and make 'sessionId` an array ? Or for compatibility, add a new separate array that contains a list of known session ids for that participant.

nickvergessen:
there is no compatibility needed atm as v4 is brand new

PVince81:
for now I'll go with an extra property as I'm a bit afraid of all the code paths to touch that could potentially break in our frontend.
most code paths only need to know if the session is "0" or not

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/5482

I was able to remove the "sessionId" attribute still as the affected frontend code was minimal.

PR #5482, opened 2021-04-14
Fix raised hand detection for multiple session, adjusted participant list response

PVince81:
### Description 

- Expose session ids as array in participants API
- The sessionId attribute is now removed from the participant list API response in favor of the sessionIds array.
- The raised hand state in the participant list is now retrieved based on the multiple sessions of a participant

As discussed in https://github.com/nextcloud/spreed/issues/5349#issuecomment-795624906, v4 of the API has not crystalized yet so it's ok to make changes in this PR.

### Related tickets

Fixes https://github.com/nextcloud/spreed/issues/5349

### Tests

Testing if raised hand appears when triggered from both sessions of a single user:

- [x] TEST: regular users with HPB
- [x] TEST: regular users without HPB
- [x] TEST: guests with HPB
- [x] TEST: guests without HPB

### DIFF
--- a/docs/participant.md
+++ b/docs/participant.md
@@ -35,7 +35,8 @@
         `lastPing` | int | * | Timestamp of the last ping of the user (should be used for sorting)
         `inCall` | int | * | Call flags the user joined with (see [constants list](constants.md#participant-in-call-flag))
         `sessionId` | string | * | `'0'` if not connected, otherwise a 512 character long string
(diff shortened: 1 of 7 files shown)


PVince81:
forgot to test this with guests

nickvergessen:
Looks good from a quick glance, but can't test atm

PVince81:
I've found additional minor locations that needed sessionId=>sessionIds adjustments, in the last commit.

Tests with guest user worked fine.

nickvergessen (approved):

PR #5484, opened 2021-04-15
Fix flickering of description overflow when switching between one-to-…

nickvergessen:
…one and a descripted room

### Steps
1. Create a conversation with a description that has 4 lines of text
2. Create a one-to-one conversation
3. Reload the page in the description room
4. Description is overlayed with shadow
5. Switch to one-to-one
6. Switch back
7. Description is still overlayed (was not before)


PVince81 (approved):
:+1:

Looks familiar: indeed, the watchers are not called initially

Issue #5399, opened 2021-03-22
Disable avatar menu for guests


PVince81:
## Steps to reproduce
1. Create a public conversation
2. Post a message as any logged in user
3. Copy link conversation
4. Log out
5. Open conversation link
6. Click on the three dots on previous the message's avatar

### Expected behaviour
No three dots to click, no avatar menu.

### Actual behaviour
Three dots visible and clickable. After clicking, an animation glitch quickly appears and disappears, then the menu is gone.
And the three dots are too.

## Talk app

**Talk app version:** Observed on stable21 640939a84fe271a91d3b4f5fdb2a27c216858d7d

nickvergessen:
Vue lib issue I guess

PVince81:
Yes and no, the vue lib will query the server and the server will return nothing and cause the glitch.

In our case we know we don't want the menu so we should set the property to disable the menu.
Unless there are use cases where we want (future) user actions for guests also ?

nickvergessen:
Ah, might then be from the switch of participants to actor type

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/5485

PR #5485, opened 2021-04-15
Disable avatar menu in messages as guest

PVince81:
When attending a conversation as guest, the menu of the avatars shown
before messages is now disabled. The menu would anyway not show any
entries.

Fixes https://github.com/nextcloud/spreed/issues/5399

- [x] TEST: menu is disabled as guest
- [x] TEST: menu is not disabled as guest
- [x] TEST: guest avatars never display a menu anyway, even for logged in users


nickvergessen (approved):
Currently okay, if for 22 we get a public profile we need to adjust. But future story

ma12-co (approved):

Issue #5489, opened 2021-04-15
Compilation errors with nextcloud-vue master post v3.9.0


PVince81:
### Steps

1. Checkout nextcloud-vue repo
2. Run `make dev-setup && make watch-js` and wait for completion
3. `npm link`
4. Checkout the [talk repo](https://github.com/nextcloud/spreed)
5. In the talk repo: `make dev-setup`
6. `npm link @nextcloud/vue` to get the above version
7. `make watch-js`

### Expected result
No compilation errors

### Actual result
With nextcloud-vue master, I get a lot of errors like:
```
ERROR in ./node_modules/@nextcloud/vue/node_modules/vue-material-design-icons/ArrowLeft.vue 1:0
Module parse failed: Unexpected token (1:0)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
```
seems related to material design.

If I then switch the nextcloud-vue checkout to v3.9.0 and recompile everything (redo `make dev-setup` for libs), the Talk repo compiles fine.

So there's a regression somewhere after the last released version.

I'll bisect...

PVince81:
Fix is here: https://github.com/nextcloud/spreed/pull/5490

PR #5490, opened 2021-04-15
Fix compatibility with nextcloud-vue v4.0.0 alpha

PVince81:
The material design icon loaders were affected by a recent change and we
need to stop excluding them for the compilation to work again.

Fixes https://github.com/nextcloud/spreed/issues/5489.

Thanks @juliushaertl for this solution.

I've tested with nextcloud-vue master as per steps in the ticket and it works fine with npm link.
Also retested with v3.9.0 from the original package.json and it works fine.

I've verified in both cases that the material design icons in the call view appear properly.

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -42,9 +42,6 @@ module.exports = {
 			{
 				test: /\.vue$/,
 				loader: 'vue-loader',
-				exclude: babelLoaderExcludeNodeModulesExcept([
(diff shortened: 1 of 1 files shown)


PVince81:
note: the library upgrade, once there's a release, will cause some styling glitches, see https://github.com/nextcloud/spreed/issues/5492

ma12-co (approved):

PR #5499, opened 2021-04-19
Added possibility to force asking guest displayname.

UspenskiyAN:
If "Ask guest for a displayname" settings in Talk global settings is on, guest can not enter to conversation until input some display name.
Signed-off-by: Uspenskiy Anatoliy <uspenskiyan@gmail.com>

Bildschirmfoto von 2021-04-26 17-02-14

### DIFF
--- a/lib/TInitialState.php
+++ b/lib/TInitialState.php
@@ -177,5 +177,10 @@ trait TInitialState {
 			'play_sounds',
 			false
 		);
+
+		$this->initialState->provideInitialState(
+			'ask_guest_username',
+			(int) $this->serverConfig->getAppValue('spreed', 'ask_guest_username', '0')
+		);
 	}
 }
--- a/lib/TInitialState.php
+++ b/lib/TInitialState.php
@@ -72,6 +72,11 @@ trait TInitialState {
 			'sip_dialin_info',
 			$this->talkConfig->getDialInInfo()
 		);
+
+		$this->initialState->provideInitialState(
+			'ask_guest_username',
+			(int) $this->serverConfig->getAppValue('spreed', 'ask_guest_username', '0')
+		);
 	}
 
 	protected function publishInitialStateForUser(IUser $user, IRootFolder $rootFolder, IAppManager $appManager): void {
@@ -176,11 +181,6 @@ trait TInitialState {
 		$this->initialState->provideInitialState(
 			'play_sounds',
 			false
-		);
-
-		$this->initialState->provideInitialState(
-			'ask_guest_username',
-			(int) $this->serverConfig->getAppValue('spreed', 'ask_guest_username', '0')
 		);
 	}
 }
--- a/src/components/AdminSettings/GeneralSettings.vue
+++ b/src/components/AdminSettings/GeneralSettings.vue
@@ -24,6 +24,17 @@
 	<div id="general_settings" class="videocalls section">
 		<h2>{{ t('spreed', 'General settings') }}</h2>
 
+		<p>
+			<input id="ask_guest_username"
+				v-model="askGuestUsername"
+				type="checkbox"
+				name="ask_guest_username"
+				class="checkbox"
+				:disabled="loading || loadingAskGuestUsername"
+				@change="saveAskGuestUsername">
+			<label for="ask_guest_username">{{ t('spreed', 'Ask guest for a displayname') }} <span class="icon-info" :title="t('spreed', 'If guest have no display name, nextcloud will ask it on guest joining to conversation')" /></label>
+		</p>
+
 		<h3>{{ t('spreed', 'Default notification settings') }}</h3>
 
 		<p>
@@ -83,9 +94,11 @@ export default {
 	data() {
 		return {
 			loading: true,
+			loadingAskGuestUsername: false,
 			loadingConversationsFiles: false,
 			loadingDefaultGroupNotification: false,
 
+			askGuestUsername: false,
 			defaultGroupNotificationOptions,
 			defaultGroupNotification: defaultGroupNotificationOptions[1],
 
@@ -96,6 +109,7 @@ export default {
 
 	mounted() {
 		this.loading = true
+		this.askGuestUsername = parseInt(loadState('spreed', 'ask_guest_username')) === 1
 		this.conversationsFiles = parseInt(loadState('spreed', 'conversations_files')) === 1
 		this.defaultGroupNotification = defaultGroupNotificationOptions[parseInt(loadState('spreed', 'default_group_notification')) - 1]
 		this.conversationsFilesPublicShares = parseInt(loadState('spreed', 'conversations_files_public_shares')) === 1
@@ -103,6 +117,15 @@ export default {
 	},
 
 	methods: {
+		saveAskGuestUsername() {
+			this.loadingAskGuestUsername = true
+
+			OCP.AppConfig.setValue('spreed', 'ask_guest_username', this.askGuestUsername ? '1' : '0', {
+				success: function() {
+					this.loadingAskGuestUsername = false
+				}.bind(this),
+			})
+		},
 		saveDefaultGroupNotification() {
 			this.loadingDefaultGroupNotification = true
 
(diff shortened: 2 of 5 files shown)


src/components/SetGuestUsername.vue
```diff
@@ -21,48 +21,79 @@
...
+		</form>
+		<Modal v-if="modal" :canClose="false">
+			<form
+				class="username-form"
+				@submit.prevent="handleChooseUserName2">
+				<div
+					class="username-form__wrapper"
+					style="display: block">
+					<h3>
+						{{ t('spreed', 'Display name: ') }}
```
nickvergessen, line 62: trailing spaces don't work in translations. I guess you need to figure this out with CSS instead.

src/components/AdminSettings/GeneralSettings.vue
```diff
@@ -96,13 +109,23 @@ export default {
...
 	},
 
 	methods: {
+		saveAskGuestUsername() {
+			this.loadingAskGuestUsername = true
+
+			OCP.AppConfig.setValue('spreed', 'ask_guest_username', this.askGuestUsername ? '1' : '0', {
+				success: function() {
+					this.loadingAskGuestUsername = false
+				}.bind(this),
```
nickvergessen, line 126:
```suggestion
				success: () => {
					this.loadingAskGuestUsername = false
				},
```

lib/TInitialState.php
```diff
@@ -177,5 +177,10 @@ protected function publishInitialStateForGuest(): void {
 			'play_sounds',
 			false
 		);
+
+		$this->initialState->provideInitialState(
+			'ask_guest_username',
```
nickvergessen, line 182: please also publish this for logged in users with any value, just to make sure it doesn't yield an error in the console at some time.

src/components/SetGuestUsername.vue
```diff
@@ -170,4 +209,10 @@ export default {
 	}
 }
 
+.modal__content {
+	width: 50vw;
+	margin: 10vw 0;
+	text-align: center;
+}
```
nickvergessen, line 216: Something missing from the Modal component?

nickvergessen:
cc @jancborchardt

src/components/SetGuestUsername.vue
UspAN: There's a space in translations:
https://github.com/nextcloud/spreed/blob/4030de49238322826dfbe2b1dc160943545c53c4/l10n/de.js#L918

nickvergessen: yeah but as you can see the german version is missing it already. They can not add trailing whitespace on transifex as they don't see it would need being there

ma12-co (commented):
Thanks for the contribution @UspAN :) 

I'm trying to test this and I'm not able to make the conversation public. Upon clicking "Allow guests I get this error". I cannot reproduce this on master

Screenshot from 2021-04-19 14-11-20

lib/TInitialState.php
UspAN: For logged in users SetGuestUsername.vue can not be used. Do you think, it is necessary?

src/components/SetGuestUsername.vue
UspAN: As I see it's in all languages:
https://github.com/nextcloud/spreed/blob/4030de49238322826dfbe2b1dc160943545c53c4/l10n/br.js#L684
https://github.com/nextcloud/spreed/blob/4030de49238322826dfbe2b1dc160943545c53c4/l10n/es.js#L901
https://github.com/nextcloud/spreed/blob/4030de49238322826dfbe2b1dc160943545c53c4/l10n/ru.js#L897
And if I delete a space - it doesn't translate.

nickvergessen: Translations will come afterwards automatically

nickvergessen: Ah, you only intend it, I will fix it then

nickvergessen: Raised https://github.com/nextcloud/spreed/issues/5500

lib/TInitialState.php
nickvergessen: well the main issue is when at some point the component is mounted it will log an error already.

src/components/AdminSettings/GeneralSettings.vue
UspAN: I think it makes sense to keep the uniformity

src/components/AdminSettings/GeneralSettings.vue
```diff
@@ -24,6 +24,17 @@
...
 
+		<p>
+			<input id="ask_guest_username"
+				v-model="askGuestUsername"
+				type="checkbox"
+				name="ask_guest_username"
+				class="checkbox"
+				:disabled="loading || loadingAskGuestUsername"
+				@change="saveAskGuestUsername">
+			<label for="ask_guest_username">{{ t('spreed', 'Ask guest for a displayname') }} <span class="icon-info" :title="t('spreed', 'If guest have no display name, nextcloud will ask it on guest joining to conversation')" /></label>
```
nickvergessen, line 35:
```suggestion
			<label for="ask_guest_username">{{ t('spreed', 'Ask guest for a displayname') }}</label>
			<em>{{ t('spreed', 'If guest have no display name, nextcloud will ask it on guest joining to conversation') }}</em>
```

UspAN: Is that how it should be?
image

src/components/SetGuestUsername.vue
UspAN: It was unnecessary. Thank you.

lib/TInitialState.php
UspAN: I moved code to `publishInitialStateShared()`

src/components/AdminSettings/GeneralSettings.vue
UspAN: I added paragraph/ Now it looks better.

UspAN:
Excuse me please. Is there something else I need to do?
I'm newbie here and I don't know what's next.

nickvergessen:
Currently waiting on feedback from our Design and Usability experts @jancborchardt and @nimisha-vijay 
They always where agains such a screen and settings are not the best idea for things like this. Either we have it for everyone or not, otherwise it's a bit confusing for users when they use 2 different installations and on one they are prompted for the name and  on the other one not.

UspAN:
Thanks for your answer. 
Well... I think, that it's a necessary feature. My colleague asked me to realize it because he tired to communicate with some "Guests" and it's useless to ask them to introduce themselves. May be there's no need in possibility to on and off this feature, but I'm not an author of this app, so I decided to do so. If that's right - I can do this feature required without possibility to disable it.

jancborchardt:
So @nimisha-vijay and I talked about this and we now agree that a modal is a nicer way for this to go forward. Reasoning and details:
- The modal will ideally only show once for a person on that instance since the name is remembered via browser storage – subsequent rooms you join should not show the modal.
- In the modal, we can show some more info on Talk, since this might well be the first time a person sees Talk. Sort of like a first-run-wizard but with only one page
- The input field will be empty with a placeholder of "Your name". If nothing is put in, we fall back to "Guest" (we could also not accept no input, but then we probably end up with a lot of people called "asdf"). The modal can also be directly dismissed, in which case the fallback "Guest" is used as well.
- If a guest wants to change their name after joining, we could show an avatar placeholder in the top right header bar, which on click would open a menu much like when you are logged in, with their current name stated and one entry called "Edit name". This would also move the naming controls out of the right sidebar, which is sort of reserved for conversation / Talk-specific settings.

What do you think @marcoambrosini @nickvergessen @UspAN

nickvergessen:
cc @UspenskiyAN as the mention above failed

nickvergessen:
Sorry, this was implemented via https://github.com/nextcloud/spreed/pull/10467 now.

So I'm closing this long open PR. Thanks for all the work!

PR #5506, opened 2021-04-22
Show an error when sharing fails

### LINKED ISSUE #5504: Upload with no sharing rights
## Steps to reproduce
1. enable "exclude groups from sharing"
2. use such an user
3. upload a file -> see no meaningful error, but it just "hangs"

nickvergessen:
Fix #5504

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -544,6 +544,9 @@ export default {
 			if (this.sendingFailure === 'quota') {
 				return t('spreed', 'Not enough free space to upload file')
 			}
+			if (this.sendingFailure === 'failed-share') {
+				return t('spreed', 'You are not allowed to share files')
+			}
 			return t('spreed', 'You cannot send messages to this conversation at the moment')
 		},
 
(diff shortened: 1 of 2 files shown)


src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -544,6 +544,9 @@ export default {
 			if (this.sendingFailure === 'quota') {
 				return t('spreed', 'Not enough free space to upload file')
 			}
+			if (this.sendingFailure === 'failed-share') {
+				return t('spreed', 'You cannot share files')
```
PVince81, line 548: why not "You are not allowed to share files" like below ? it's a bit clearer.

nickvergessen: Done

PVince81 (approved):
:+1:

tobiasKaminsky:
This is "only" about server side, or?
How are clients informed?

nickvergessen:
well there is a 403 error and the web UI client just didn't handle it but swallowed all errors and thought it worked `¯\_(ツ)_/¯`

tobiasKaminsky:
How do you handle those things? Shall I create an issue on Android and iOS then, or like files client one meta issue in this repo?

PR #5512, opened 2021-04-23
More JS tests for stores

PVince81:
Added more JS tests for stores.
Extracted global store config to a separate file to be able to include it in
tests.

Up to 56 tests now :tada:


PVince81:
need to find a way around this:
'Date prototype is read only, properties should not be added  no-extend-native`

mocking time is painful with Jest 24, which is what is bundled with the vue-cli plugin.
Ideally I'd prefer if it was possible to upgrade to Jest 26 which has better mocking of time, like `jest.setSystemTime` but I'm not sure it's possible.

PVince81:
I found a way to mock the date, should get rid of the warning now

nickvergessen (approved):

PR #5516, opened 2021-04-24
Fix missing screenshare icon

### LINKED ISSUE #5510: Screenshare icon not rendered
## Steps to reproduce
1. Start a call on master
2. Try to share your screen

### Expected behaviour
You see Monitor vue material design icon

### Actual behaviour
The icon slot is not rendered

There was a lot activity recently:
https://github.com/nextcloud/nextcloud-vue/commits/master/src/components/Actions/Actions.vue

Also the icon appears when you delete all actions but one:
https://github.com/nextcloud/spreed/blob/514405d556785ad9cf0d3ef925962b61e52a57d9/src/components/CallView/shared/LocalMediaControls.vue#L99-L123

So the talk code seems correct and something is broken in the hasMultipleActions part of the Actions component

cc @PVince81 @ma12-co @skjnldsv

nickvergessen:
Fix #5510

### DIFF
--- a/src/components/CallView/shared/LocalMediaControls.vue
+++ b/src/components/CallView/shared/LocalMediaControls.vue
@@ -96,29 +96,59 @@
 					decorative />
 				<!-- /Actions button icon -->
 				<!-- Actions -->
-				<ActionButton v-if="!screenSharingMenuOpen" @click.stop="toggleScreenSharingMenu" />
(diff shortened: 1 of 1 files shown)


skjnldsv (approved):
Makes much more sense.

ma12-co (approved):
Works

PR #5519, opened 2021-04-26
Fix loading the correct sound setting for users

nickvergessen:
### Steps
1. As a user
2. Open the settings
3. Disable call sounds
4. Reload the page
5. Check settings again.

### DIFF
--- a/src/store/soundsStore.js
+++ b/src/store/soundsStore.js
@@ -22,6 +22,7 @@
 
 import fromStateOr from './helper'
 import BrowserStorage from '../services/BrowserStorage'
+import { setPlaySounds } from '../services/settingsService'
 
 const state = {
 	userId: undefined,
@@ -72,7 +73,8 @@ const actions = {
 	 * @param {object} context default store context;
 	 * @param {boolean} enabled Whether sounds should be played
 	 */
-	setPlaySounds({ commit }, enabled) {
+	async setPlaySounds({ commit }, enabled) {
+		await setPlaySounds(status)
 		commit('setPlaySounds', enabled)
 	},
 }
--- a/src/store/soundsStore.js
+++ b/src/store/soundsStore.js
@@ -24,6 +24,7 @@ import fromStateOr from './helper'
 import BrowserStorage from '../services/BrowserStorage'
 
 const state = {
+	userId: undefined,
 	playSoundsUser: fromStateOr('spreed', 'play_sounds', false),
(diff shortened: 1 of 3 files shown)


src/store/soundsStore.js
```diff
@@ -29,8 +29,8 @@ const state = {
 }
 
 const getters = {
-	playSounds: (state) => {
-		if (state.userId) {
+	getPlaySounds: (state) => (userId) => {
+		if (userId) {
```
nickvergessen, line 33: Feels a bit weird to have to inject the current user (gives the wrong feeling you could read the setting of others), but seems like we can not access the state of other stores in here?

PVince81: you can: see https://vuex.vuejs.org/guide/modules.html#namespacing

then something like `rootGetters.actorStore.getUser()` or `rootGetters['actorStore/getUser']` might work

PVince81: since we didn't give modules a name, maybe even `rootGetters.getUser()` might work

src/store/soundsStore.js
```diff
@@ -48,16 +50,31 @@ const mutations = {
...
 
 const actions = {
+
+	/**
+	 * @param {object} context default store context;
+	 * @param {object} user A NextcloudUser object as returned by @nextcloud/auth
+	 * @param {string} user.uid The user id of the user
+	 */
+	setCurrentUser(context, user) {
+		context.commit('setUserId', user.uid)
```
PVince81, line 67: this will boom if user is null when dealing with guests ?

nickvergessen: https://github.com/nextcloud/spreed/blob/a661a156a0a6f2696db15cbad2b678807b5b779a/src/App.vue#L319-L324

Actor store also already depends on valid values: https://github.com/nextcloud/spreed/blob/6184127a50ad62e429600e2d6cc0310d2873af6d/src/store/actorStore.js#L124-L129 doc block is also rather clear it does not allow null

PVince81: yeah, while writing tests I remember that to set the user to null you need to call `setCurrentParticipant`.
it's a bit twisted, but explains it. let's move forward

PVince81 (approved):
:+1:

PVince81:
you might have missed that bit: https://github.com/nextcloud/spreed/pull/5524

PR #5522, opened 2021-04-27
More JS store tests featuring fileUploadStore and a few simple ones

PVince81:
Brings in the following:
- fileUploadStore tests
- guestNameStore tests
- newGroupConversationStore tests
- quoteReplyStore tests
- sidebarStore tests

For the other stores not covered:
- they are too trivial (just getters/setters)
- they are blocked by global store usage (requireshttps://github.com/nextcloud/spreed/pull/5526)
- they are not fully testable because of the initial loadState / BrowserStore.getItem: we might want to move to a more explicit way of loading state with a "loadState" action of sorts => to be considered separately

~~Currently blocked because "temporaryMessage" is accessing the global store and I didn't find a way to set the current mocked store to the global module.~~

~~Maybe need to refactor those usages and always receive the store injected in the function.
At least that's the quickest workaround I can think of.~~

~~The long term approach would be to refactor all those usages and move them into the store as store actions where applicable (like `createTemporaryMessage`)~~

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,11 +3,11 @@ module.exports = {
 		'@nextcloud'
 	],
 	overrides: [
-    {
-      'files': ['**/*.spec.js'],
-      'rules': {
-        'node/no-unpublished-import': 0
-      }
-    }
-  ]
+		{
+			'files': ['**/*.spec.js'],
+			'rules': {
+				'node/no-unpublished-import': 0
+			}
+		}
+	]
 }
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,4 +2,12 @@ module.exports = {
 	extends: [
 		'@nextcloud'
 	],
+	overrides: [
+    {
+      'files': ['**/*.spec.js'],
+      'rules': {
+        'node/no-unpublished-import': 0
+      }
+    }
+  ]
 }
--- a/jest.config.js
+++ b/jest.config.js
@@ -28,7 +28,7 @@ module.exports = {
 	setupFiles: ['jest-localstorage-mock'],
 	setupFilesAfterEnv: [
 		'<rootDir>/src/test-setup.js',
-	    'jest-mock-console/dist/setupTestFramework.js',
+		'jest-mock-console/dist/setupTestFramework.js',
 	],
 	transform: {
 		// process `*.js` files with `babel-jest`
--- a/jest.config.js
+++ b/jest.config.js
@@ -26,7 +26,10 @@ module.exports = {
 	testMatch: ['<rootDir>/src/**/*.(spec|test).(ts|js)'],
 	resetMocks: false,
 	setupFiles: ['jest-localstorage-mock'],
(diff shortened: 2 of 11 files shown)


src/store/fileUploadStore.spec.js
```diff
@@ -0,0 +1,78 @@
+import { createLocalVue } from '@vue/test-utils'
+import storeConfig from './storeConfig'
+import Vuex from 'vuex'
+import { cloneDeep } from 'lodash'
+
+import globalStore from '../store/index'
+
+// workaround for files using the store globally
+jest.mock('../store/index', () => jest.fn())
```
PVince81, line 9: we can't put an object in here without redefining it completely, so the closest we can mock is a function, but the users of the module expect an object, not a function

PVince81:
looks like there wouldn't be that much to adjust:
```
src/services/settingsService.js|26 col 23| import store from '../store/index'
src/services/conversationsService.js|27 col 23| import store from '../store/index'
src/services/messagesService.js|25 col 23| import store from '../store/index'
src/services/participantsService.js|36 col 23| import store from '../store/index'
src/utils/webrtc/models/LocalCallParticipantModel.js|22 col 29| import store from '../../../store/index.js'
src/utils/webrtc/models/LocalMediaModel.js|22 col 29| import store from '../../../store/index.js'
src/utils/webrtc/webrtc.js|31 col 26| import store from '../../store/index.js'
src/utils/fileUpload.js|23 col 23| import store from '../store/index'
src/utils/temporaryMessage.js|23 col 23| import store from '../store/index'
```

PVince81:
raised https://github.com/nextcloud/spreed/pull/5524 and https://github.com/nextcloud/spreed/pull/5526 to fix store usage

PVince81:
- [ ] separate the bug fix from fileUpload.js or is it trivial enough ?

PVince81:
I'll handle participantsStore and messagesStore separately as they are more complex.

Please review the above, the only complex one is the file upload store.

.eslintrc.js
```diff
@@ -2,4 +2,12 @@ module.exports = {
 	extends: [
 		'@nextcloud'
 	],
+	overrides: [
+    {
```
nickvergessen, line 6: tabs vs. spaces in this file

jest.config.js
```diff
@@ -26,7 +26,10 @@ module.exports = {
 	testMatch: ['<rootDir>/src/**/*.(spec|test).(ts|js)'],
 	resetMocks: false,
 	setupFiles: ['jest-localstorage-mock'],
-	setupFilesAfterEnv: ['<rootDir>/src/test-setup.js'],
+	setupFilesAfterEnv: [
+		'<rootDir>/src/test-setup.js',
+	    'jest-mock-console/dist/setupTestFramework.js',
```
nickvergessen, line 31: tabs  vs. spaces

nickvergessen (approved):
Looks okay otherwise

PVince81:
tabs fixed, waiting for CI now

PR #5526, opened 2021-04-28
Decouple conversations and messages services from the global store

PVince81:
Removes dodgy global store usage from `messagesService`.

Moves `fetchConversations`, `fetchMessages`, `lookForNewMessages` and `postNewMessages` to the `messagesStore`.

The logic was mostly copy pasted there and rewired to use the correct variable.

- [x] TEST: maintenance mode detection still working
- [x] TEST: conversation list refresh works
- [x] TEST: switching conversations still works and properly cancels fetch messages
- [x] TEST: posting messages still work

- BUG: cancelling the postNewMessage doesn't work after timeout (no exception thrown), but also did not work on master, to be investigated separately => raised as https://github.com/nextcloud/spreed/issues/5527

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -131,7 +131,6 @@ import debounce from 'debounce'
 import { EventBus } from '../../services/EventBus'
 import {
 	createOneToOneConversation,
-	fetchConversations,
 	searchPossibleConversations,
 	searchListedConversations,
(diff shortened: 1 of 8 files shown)


PVince81:
- [ ] streamline the timeout fix https://github.com/nextcloud/spreed/pull/5528

ma12-co (approved):
Code looks good and everything seems to work :+1:

PVince81:
rebased and fixed conflicts

Issue #5527, opened 2021-04-28
Temp message timeout not properly cancelling


PVince81:
## Steps to reproduce
1. Edit "ChatController#sendMessage" and add a `sleep(10);` at the top of the function
2. Edit NewMessageForm.vue and reduce the timeout to 1000 in `handleSubmit`
3. Compile and refresh the page
4. Open the network console
5. Send a temporary message

### Expected behaviour
Failed indicator appears and stays, request cancelled in network console.

### Actual behaviour
Failed indicator appears, request is not cancelled.
After 10 seconds, failed indicator disappears and considered a success.

## Talk app

**Talk app version:** master (d72448c5eb0b7273f40e1ce701f26110588cab0e)

PVince81:
clicking "resend" before the timeout has the button stuck, but no duplicates appear, so there's no harm, just slight inconvenience

PVince81:
I don't understand... the same construct seems to work for fetchMessages where the request gets cancelled when switching conversations

PVince81:
I've added a similar timeout logic for lookForNewMessages and cancelling the request does properly trigger the exception.

So now the only difference I found is that `postNewMessages` is a POST request while the other ones are GET requests.
So maybe POST requests are not cancellable or require a different way of setting it up.

PVince81:
Alright, found the issue: `postNewMessages` is not forwarding the options attribute.

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/5528

PR #5528, opened 2021-04-28
Fix cancelling postNewMessage

PVince81:
For timeout cancelling to work, the options need to be passed to the
request object.

Fixes https://github.com/nextcloud/spreed/issues/5527

We might want to backport.


PVince81:
JS test fixed. Am not used that we now have JS unit tests at all in the first place :sweat_smile:

nickvergessen (approved):

PR #5529, opened 2021-04-28
Decouple participantsService from the store

PVince81:
Moved joinConversation, leaveConversation and a few other dependencies
to the participantsStore.
Removes the dependency to the store from participantsService.

There's a slight change of behavior: joinConversation in the service now
also connects to signaling, and only when this is done the first
action dispatch are done. In the previous implementation the store
already received some updates before we even connected to signaling, see https://github.com/nextcloud/spreed/pull/5529/files#diff-e0f25089223beb2eb4703995c475f60ab752fa3a0053b63b848a03bff7cd5c16R46

Tests:
- [x] TEST: regular join and conversation switching
- [x] TEST: guest join and leave
- [x] TEST: file conversation
- [x] TEST: video verification

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -50,7 +50,6 @@ import { EventBus } from './services/EventBus'
 import BrowserStorage from './services/BrowserStorage'
 import { getCurrentUser } from '@nextcloud/auth'
 import {
-	joinConversation,
 	leaveConversationSync,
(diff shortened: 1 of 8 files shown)


nickvergessen (approved):
🐘 only tested without HPB version

PR #5530, opened 2021-04-28
Register visibility observer in all entry points

PVince81:
The observe-visibility extension was missing in many entry points
resulting in repeated console warnings.

The directive usage in the message list was introduced as part of #3825

### DIFF
--- a/src/mainFilesSidebar.js
+++ b/src/mainFilesSidebar.js
@@ -23,6 +23,7 @@
  */
 
 import Vue from 'vue'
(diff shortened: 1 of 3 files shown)


nickvergessen (approved):
Would this be something for the init.js file? Or is it not possible because of the orders of things?

PVince81:
> Would this be something for the init.js file? Or is it not possible because of the orders of things?

I wasn't sure because all other components are registered outside of it, so preferred to be safe in case there's an order thing.

PR #5531, opened 2021-04-28
Decouple fileupload and createTemporaryMessage from the global store module

PVince81:
1. Decouple fileUpload service from the global store
    
Moved the logic of "processFiles" into the existing "initialiseUpload" action of the fileUploadStore

2. Move createTemporaryMessage to a store action
    
Since createTemporaryMessage heavily depends on a store, moved it into a messagesStore action to decouple it from the global store module.    

Note: even though it's synchronous, the dispatch() function returns a promise so we need to treat it as if it was async.

This is prerequisite to be able to properly mock parts of the (non-global) store in https://github.com/nextcloud/spreed/pull/5522

Tested:
- [x] drag and drop upload of multiple files
- [x] upload dialog with multiple files
- [x] share existing file to the conversation from the file picker


src/store/messagesStore.js
```diff
@@ -286,6 +288,64 @@ const actions = {
...
+	createTemporaryMessage(context, { text, token, uploadId, index, file, localUrl }) {
+		const messageToBeReplied = context.getters.getMessageToBeReplied(token)
+		const date = new Date()
+		let tempId = 'temp-' + date.getTime()
+		const messageParameters = {}
+		if (file) {
+			tempId += '-' + uploadId + '-' + Math.random()
+			messageParameters.file = {
+				'type': 'file',
+				'file': file,
```
nickvergessen, line 313:
```suggestion
				file,
```

src/store/messagesStore.js
```diff
@@ -286,6 +288,64 @@ const actions = {
...
+			id: tempId,
+			actorId: context.getters.getActorId(),
+			actorType: context.getters.getActorType(),
+			actorDisplayName: context.getters.getDisplayName(),
+			timestamp: 0,
+			systemMessage: '',
+			messageType: '',
+			message: text,
+			messageParameters,
+			token: token,
```
nickvergessen, line 333:
```suggestion
			token,
```

src/store/messagesStore.js
PVince81: this is just copy pasted from the old copy, but probably a good idea to adjust. I'll adjust them all

src/store/messagesStore.js
```diff
@@ -286,6 +288,64 @@ const actions = {
...
+	createTemporaryMessage(context, { text, token, uploadId, index, file, localUrl }) {
+		const messageToBeReplied = context.getters.getMessageToBeReplied(token)
+		const date = new Date()
+		let tempId = 'temp-' + date.getTime()
+		const messageParameters = {}
+		if (file) {
+			tempId += '-' + uploadId + '-' + Math.random()
+			messageParameters.file = {
+				type: 'file',
+				file: file,
```
nickvergessen, line 313:
```suggestion
				file,
```

nickvergessen (approved):
Seems to still work

PR #4253, opened 2020-09-30
Show icon only with dnd status in the message

nickvergessen:
Ref https://github.com/nextcloud/nextcloud-vue/pull/1422 and https://github.com/nextcloud/server/pull/23114

PR #5533, opened 2021-04-29
Fix sorting of users that are online with user status offline

nickvergessen:
This was missed in https://github.com/nextcloud/spreed/pull/4253

### Steps
1. As a user with Z as first char go into a chat that has other users
2. Set your user status to offline
3. As another user go to the chat and check the sidebar

### Before
User is marked online but sorted at the bottom

Before | After
---|---
Bildschirmfoto von 2021-04-29 09-18-22 | Bildschirmfoto von 2021-04-29 09-18-27

### DIFF
--- a/src/components/RightSidebar/Participants/CurrentParticipants/CurrentParticipants.vue
+++ b/src/components/RightSidebar/Participants/CurrentParticipants/CurrentParticipants.vue
@@ -145,14 +145,19 @@ export default {
 				return p2IsGroup ? -1 : 1
 			}
 
-			let hasSessions1 = !!participant1.sessionIds.length
-			let hasSessions2 = !!participant2.sessionIds.length
+			const hasSessions1 = !!participant1.sessionIds.length
+			const hasSessions2 = !!participant2.sessionIds.length
+			/**
+			 * For now the user status is not overwriting the online-offline status anymore
+			 * It felt too weird having users appear as offline but they are in the call or chat actively
 			if (participant1.status === 'offline') {
 				hasSessions1 = false
 			}
 			if (participant2.status === 'offline') {
 				hasSessions2 = false
 			}
+			 */
+
 			if (!hasSessions1) {
 				if (hasSessions2) {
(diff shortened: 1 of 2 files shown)


src/components/RightSidebar/Participants/CurrentParticipants/CurrentParticipants.vue
```diff
@@ -145,14 +145,16 @@ export default {
 				return p2IsGroup ? -1 : 1
 			}
 
-			let hasSessions1 = !!participant1.sessionIds.length
-			let hasSessions2 = !!participant2.sessionIds.length
+			const hasSessions1 = !!participant1.sessionIds.length
+			const hasSessions2 = !!participant2.sessionIds.length
+			/*
```
PVince81, line 150: remove commented out block or are you still testing ?

nickvergessen: Keeping it as a comment similar to https://github.com/nextcloud/spreed/pull/4253/files#diff-4eb65f9626b81feabc22bb4e044ba305cbe75667d29909103aa935e97c081dcb so it's easy to see that it is missing intentionally

PVince81: in that case I'd suggest adding a text comment as well because the next developer will wonder if this was forgotten

tbh I don't really understand why commenting this out fixes the issue, so maybe the comment could also add a hint about that

nickvergessen: the problem is sorting accepted the user_status but the "show online/offline" ignored it 🤷

nickvergessen: Added a comment in both places

PVince81 (approved):
Thanks for the comment, makes it more clear :+1:

Issue #3110, opened 2020-03-20
new message indicator when sidebar is collapsed 


schiessle:
It would be nice if we would have a new message notification when the right sidebar is closed. Maybe the button to re-open the sidebar could have a unread message count? Maybe also change the color or blink twice if a new message comes in... just some ideas.

nickvergessen:
You mean while being in a call?

nickvergessen:
the main issue is, when the sidebar is closed, the message view is destroyed and we are actually not pulling for new messages anymore Oo

Otherwise the work is at https://github.com/nextcloud/spreed/pull/new/feature/3110/unread-indicator-on-folded-chat

schiessle:
> You mean while being in a call?

yes, would be nice to get somehow notified about chat messages while in a call. While writing this I noticed that this would be not only useful when the sidebar is closed but also when the user is in a different tab, e.g. the participants list.

PVince81:
also applies to when the participant list is open, for example when viewing who raised hands, if a message is posted in the chat tab, we should maybe post a number bubble there showing how many new messages are present there

nickvergessen:
Same the otherway around, there should be a raised hand bubble when you have the chat open

PVince81:
> the main issue is, when the sidebar is closed, the message view is destroyed and we are actually not pulling for new messages anymore 

as far as I could observe on master today, the message view is not destroyed while the sidebar is closed.
the calls to `getNewMessages()' are still pumping so we could then find out about unread messages once https://github.com/nextcloud/spreed/issues/1164 is done.

however I'm not sure if we should continue the above long pool as it might disturb call performance and use different techniques to find out about new messages, like through signaling: https://github.com/nextcloud/spreed/issues/3377

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/5534

Issue #5525, opened 2021-04-28
Show dot or number of unread messages in the list of Apps


q-wertz:
**Is your feature request related to a problem? Please describe.**
As already discussed in a lot of other issues/feature requests one currently misses a lot of notifications/new messages.
E.g. 
- playing sound https://github.com/nextcloud/spreed/issues/4651
- inactive Browser problem https://github.com/nextcloud/spreed/issues/1575 https://github.com/nextcloud/spreed/issues/5321

In my case I also do not see a red dot as prefix in the tab but an asterix * (maybe my browser/OS is the problem?)

**Describe the solution you'd like**
Another additional Notification would maybe be to show a red dot or the number of unread messages as an overlay of the Q in the Apps List

![dot](https://user-images.githubusercontent.com/20948931/116373562-fd684b80-a7fc-11eb-9c46-bf8f6a08dd37.png)
![num](https://user-images.githubusercontent.com/20948931/116373575-ffcaa580-a7fc-11eb-9205-bf2da85021da.png)

**Describe alternatives you've considered**
I am on Talk v10.0.7, in https://github.com/nextcloud/spreed/issues/5321#issuecomment-800190265 it is mentioned that with Talk 11.0.1 some things changed, so maybe notifications are more reliable now?

nickvergessen:
Can't find the ticket atm, but the main issue is that there is no API in the server for this.

The other problem is creating and updating the information. Like counting across all your chats on any page.

q-wertz:
Thanks for the reply

> Can't find the ticket atm, but the main issue is that there is no API in the server for this.

Ah ok. So this is an open issue on the server side? Or not even planned to be implemented there? Might also be useful for some other apps (Deck/Task showing today due cards, ...)

> The other problem is creating and updating the information. Like counting across all your chats on any page.

not a web developer, so I just trust you there on this :smile: 


Is it possible to get something similar into the notifications (bell) sign?
So red dot there and a single message with something like "5 new messages". Can think about some methods (no clue if they would really work technically) depending on what the Notification API is capable of.

q-wertz:
This one does only work inside the app environment? https://github.com/nextcloud/nextcloud-vue/tree/master/src/components/AppNavigationCounter

nickvergessen:
Yeah that is for the left sidebar, not the top one. and as said: populating the "dot/counter" with the information comes with a huge performance penalty.

jancborchardt:
[...]

PR #5534, opened 2021-04-29
New messages indicator while in call

PVince81:
Fixes https://github.com/nextcloud/spreed/issues/3110

- [x] requires https://github.com/nextcloud/nextcloud-vue/pull/1895

When the sidebar is closed while in call, there will be an unread messages counter on top of the "open sidebar" button:
image

It is blue for new mentions and grey if there are no mentions.

Additionally, the first time a new unread message arrives or a new mention, a notification will be displayed.

### DIFF
--- a/src/components/TopBar/TopBar.vue
+++ b/src/components/TopBar/TopBar.vue
@@ -130,7 +130,7 @@
 					@click="openSidebar">
 					<MessageText
 						slot="icon"
-						:size="24"
+						:size="16"
 						title=""
 						fill-color="#ffffff"
 						decorative />
@@ -320,11 +320,7 @@ export default {
 
 			// new messages arrived
 			if (newValue > 0 && oldValue === 0) {
-				if (this.hasUnreadMentions) {
-					showMessage(t('spreed', 'You have been mentioned in the chat.'))
-				} else {
-					showMessage(t('spreed', 'You have new unread messages in the chat.'))
-				}
+				showMessage(t('spreed', 'You have new unread messages in the chat.'))
 			}
 		},
 
@@ -333,6 +329,7 @@ export default {
 				return
 			}
 
+			// prevent duplicate notification caused by unreadMessagesCounter in case of mention
 			if (newValue && this.unreadMessagesCounter > 0) {
 				showMessage(t('spreed', 'You have been mentioned in the chat.'))
 			}
--- a/src/components/TopBar/TopBar.vue
+++ b/src/components/TopBar/TopBar.vue
@@ -198,6 +198,12 @@ export default {
 		},
 	},
 
+	data: () => {
+		return {
+			unreadNotificationHandle: null,
+		}
+	},
+
 	computed: {
 		isFullscreen() {
 			return this.$store.getters.isFullscreen()
@@ -319,8 +325,8 @@ export default {
 			}
 
 			// new messages arrived
-			if (newValue > 0 && oldValue === 0) {
-				showMessage(t('spreed', 'You have new unread messages in the chat.'))
+			if (newValue > 0 && oldValue === 0 && !this.hasUnreadMentions) {
+				this.notifyUnreadMessages(t('spreed', 'You have new unread messages in the chat.'))
 			}
 		},
 
@@ -329,9 +335,15 @@ export default {
 				return
 			}
 
-			// prevent duplicate notification caused by unreadMessagesCounter in case of mention
-			if (newValue && this.unreadMessagesCounter > 0) {
-				showMessage(t('spreed', 'You have been mentioned in the chat.'))
+			if (newValue) {
+				this.notifyUnreadMessages(t('spreed', 'You have been mentioned in the chat.'))
+			}
+		},
+
+		isInCall(newValue) {
+			if (!newValue) {
+				// discard notification if the call ends
+				this.notifyUnreadMessages(null)
 			}
 		},
 	},
@@ -344,6 +356,7 @@ export default {
 	},
 
 	beforeDestroy() {
+		this.notifyUnreadMessages(null)
 		document.removeEventListener('fullscreenchange', this.fullScreenChanged, false)
 		document.removeEventListener('mozfullscreenchange', this.fullScreenChanged, false)
 		document.removeEventListener('MSFullscreenChange', this.fullScreenChanged, false)
@@ -351,6 +364,16 @@ export default {
 	},
 
 	methods: {
+		notifyUnreadMessages(message) {
+			if (this.unreadNotificationHandle) {
+				this.unreadNotificationHandle.hideToast()
+				this.unreadNotificationHandle = null
+			}
+			if (message) {
+				this.unreadNotificationHandle = showMessage(message)
+			}
+		},
+
 		openSidebar() {
 			this.$store.dispatch('showSidebar')
 			BrowserStorage.setItem('sidebarOpen', 'true')
--- a/src/components/TopBar/TopBar.vue
+++ b/src/components/TopBar/TopBar.vue
@@ -125,6 +125,19 @@
 				close-after-click="true"
 				container="#content-vue">
 				<ActionButton
+					v-if="isInCall"
+					key="openSideBarButtonMessageText"
+					@click="openSidebar">
+					<MessageText
+						slot="icon"
+						:size="24"
+						title=""
+						fill-color="#ffffff"
+						decorative />
+				</ActionButton>
+				<ActionButton
+					v-else
+					key="openSideBarButtonMenuPeople"
 					:icon="iconMenuPeople"
 					@click="openSidebar" />
 			</Actions>
@@ -147,6 +160,7 @@ import CallButton from './CallButton'
 import BrowserStorage from '../../services/BrowserStorage'
 import ActionLink from '@nextcloud/vue/dist/Components/ActionLink'
 import ActionSeparator from '@nextcloud/vue/dist/Components/ActionSeparator'
+import MessageText from 'vue-material-design-icons/MessageText'
 import MicrophoneOff from 'vue-material-design-icons/MicrophoneOff'
 import { CONVERSATION, PARTICIPANT } from '../../constants'
 import { generateUrl } from '@nextcloud/router'
@@ -170,6 +184,7 @@ export default {
 		CounterBubble,
 		CallButton,
 		ActionSeparator,
+		MessageText,
 		MicrophoneOff,
 		ConversationIcon,
 	},
@@ -318,11 +333,7 @@ export default {
 				return
 			}
 
-			// in some cases the user already had unread messages and saw the other notification
-			// but then got mentionned
-			// the limit of "> 1" is here to avoid seeing two notifications, one for the first non-mention
-			// unread message and the second for the mention
-			if (newValue && this.unreadMessagesCounter > 1) {
+			if (newValue && this.unreadMessagesCounter > 0) {
 				showMessage(t('spreed', 'You have been mentioned in the chat.'))
 			}
 		},
@@ -476,6 +487,7 @@ export default {
 		position: absolute;
 		top: 40px;
 		right: 10px;
+		pointer-events: none;
 	}
 }
 
--- a/src/components/TopBar/TopBar.vue
+++ b/src/components/TopBar/TopBar.vue
@@ -129,13 +129,20 @@
 					@click="openSidebar" />
 			</Actions>
 		</div>
+		<CounterBubble
+			v-if="showOpenSidebarButton && isInCall && unreadMessagesCounter > 0"
+			class="unread-messages-counter"
+			:highlighted="hasUnreadMentions">
+			{{ unreadMessagesCounter }}
+		</CounterBubble>
 	</div>
 </template>
 
 <script>
-import { showError, showSuccess } from '@nextcloud/dialogs'
+import { showError, showSuccess, showMessage } from '@nextcloud/dialogs'
 import ActionButton from '@nextcloud/vue/dist/Components/ActionButton'
 import Actions from '@nextcloud/vue/dist/Components/Actions'
+import CounterBubble from '@nextcloud/vue/dist/Components/CounterBubble'
 import CallButton from './CallButton'
 import BrowserStorage from '../../services/BrowserStorage'
 import ActionLink from '@nextcloud/vue/dist/Components/ActionLink'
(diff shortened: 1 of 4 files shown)


PVince81:
@ma12-co I'm unsure about the bubble alignment. It needs to be positioned in a way that it still looks nice when there are more digits.

Also: I had to reuse the "AppNavigationCounter" outside the app navigation because somehow we didn't make that a generic numbers bubble component ?

src/components/TopBar/TopBar.vue
```diff
@@ -235,6 +248,37 @@ export default {
...
+			if (!this.isInCall || !this.showOpenSidebarButton) {
+				return
+			}
+
+			// new messages arrived
+			if (newValue > 0 && oldValue === 0) {
+				if (this.hasUnreadMentions) {
+					showMessage(t('spreed', 'You have been mentioned in the conversation sidebar.'))
+				} else {
+					showMessage(t('spreed', 'You have new unread messages in the conversation sidebar.'))
```
nickvergessen, line 262:
```suggestion
					showMessage(t('spreed', 'You have been mentioned in the chat.'))
				} else {
					showMessage(t('spreed', 'You have new unread messages in the chat.'))
```

src/components/TopBar/TopBar.vue
```diff
@@ -235,6 +248,37 @@ export default {
...
+			if (!this.isInCall || !this.showOpenSidebarButton) {
+				return
+			}
+
+			// in some cases the user already had unread messages and saw the other notification
+			// but then got mentionned
+			// the limit of "> 1" is here to avoid seeing two notifications, one for the first non-mention
+			// unread message and the second for the mention
+			if (newValue && this.unreadMessagesCounter > 1) {
+				showMessage(t('spreed', 'You have been mentioned in the conversation sidebar.'))
```
nickvergessen, line 277:
```suggestion
				showMessage(t('spreed', 'You have been mentioned in the chat.'))
```

src/components/TopBar/TopBar.vue
```diff
@@ -103,13 +103,19 @@
 				:icon="iconMenuPeople"
 				@click="openSidebar" />
 		</Actions>
+		<AppNavigationCounter
+			v-if="showOpenSidebarButton && isInCall && unreadMessagesCounter > 0"
+			class="unread-messages-counter"
+			:highlighted="hasUnreadMentions">
+			{{ unreadMessagesCounter }}</AppNavigationCounter>
```
nickvergessen, line 110:
```suggestion
			{{ unreadMessagesCounter }}
		</AppNavigationCounter>
```

src/components/TopBar/TopBar.vue
```diff
@@ -103,13 +103,19 @@
 				:icon="iconMenuPeople"
 				@click="openSidebar" />
 		</Actions>
+		<AppNavigationCounter
```
nickvergessen, line 106: this is abuse and might break without previous note as AppNavigation* should only be used in AppNavigation

nickvergessen (changes requested):

src/components/TopBar/TopBar.vue
PVince81: yeah, was afraid of that... so maybe we first adjust the nextcloud-vue component to be more generic.
might be better than just copy-pasting all the styles over here

src/components/TopBar/TopBar.vue
PVince81: I've been avoiding the term "chat" everywhere, believing that we had settled to only use the term "conversation" in visible texts ?

nickvergessen: No, we avoid "room" and use conversation instead.
But the chat is the chat. Also refering to the sidebar withit is weird, as when you leave the call and try to look at the sidebar where you have been mentioned you wont find anything as the chat is then in the main view

src/components/TopBar/TopBar.vue
PVince81: now thinking, a generic CounterBubble might not be good either because it would always need styling hacks to position it.

@ma12-co would it make sense to introduce a ActionButtonCounter, or a property for action buttons to display a bubble nearby ?

q-wertz: Sorry to jump in here but this would maybe also be a nice feature for other apps (see e.g. https://github.com/nextcloud/spreed/issues/5525 )

PVince81:
PR for CounterBubble: https://github.com/nextcloud/nextcloud-vue/pull/1895

src/components/TopBar/TopBar.vue
PVince81: yeah, so generic component it is: https://github.com/nextcloud/nextcloud-vue/pull/1895

nickvergessen:
Requirement is full filled, so this can be finished?

PVince81:
> Requirement is full filled, so this can be finished?

oh, yes, finally!!

src/components/TopBar/TopBar.vue
PVince81: adjusted to use CounterBubble now

PVince81:
rebased and retested, seems to work.

however can't test for too long due to the unrelated scrolling regression: https://github.com/nextcloud/spreed/issues/5747

marcoambrosini:
@PVince81 could we use a chat icon as default icon while in call, instead of the participants one?

src/components/TopBar/TopBar.vue
```diff
@@ -282,6 +297,37 @@ export default {
...
+
+		hasUnreadMentions(newValue, oldValue) {
+			if (!this.isInCall || !this.showOpenSidebarButton) {
+				return
+			}
+
+			// in some cases the user already had unread messages and saw the other notification
+			// but then got mentionned
+			// the limit of "> 1" is here to avoid seeing two notifications, one for the first non-mention
+			// unread message and the second for the mention
```
nickvergessen, line 324: This would be okay from my POV (second showMessage when you are now mentioned instead of unread chat)

PVince81:
> @PVince81 could we use a chat icon as default icon while in call, instead of the sidebar-participants hybrid one?
> 
> like `mdiMessageText `
> message-text

I've just tried this locally and it seems the Actions component is very glitchy.

When starting the call the icon doesn't appear even though the Vue inspector tells me the MessageTextIcon is there.
I have to open and then close the sidebar again to make it appear.

And then when I leave the call, the icon is still there even though `isInCall` should be false. Something weird going on... I'll try to debug this, maybe the reactivity has some trouble.

PVince81:
what the hell, ghost components...
image

PVince81:
I wonder if it's a bug in the vue transition and that it mistakenly fails to remove the DOM element even  though the component is not there any more. But I haven't found where the right sidebar transition is defined to disable it for testing.

Worst case I replace the material design icon with a regular one...

PVince81:
and here the glitch where the Actions render the single ActionButton hidden the first time:
image

image

PVince81:
alright, I managed to get rid of the glitches by setting "key" attributes on the buttons... I'll do more testing then push

src/components/TopBar/TopBar.vue
PVince81: done

PVince81:
@marcoambrosini please re-review

src/components/TopBar/TopBar.vue
```diff
@@ -125,21 +125,42 @@
 				close-after-click="true"
 				container="#content-vue">
 				<ActionButton
+					v-if="isInCall"
+					key="openSideBarButtonMessageText"
+					@click="openSidebar">
+					<MessageText
+						slot="icon"
+						:size="24"
```
nickvergessen, line 133:
```suggestion
						:size=16"
```

like all other icons in the UI?

nickvergessen:
### Duplicate notification

Something seems wrong with the condition as I see duplicate notifications
Bildschirmfoto von 2021-06-15 10-40-26


### Steps
1. Admin start a call
2. User 2 writes two messages in quick time:
     * Hello
     * `@admin`

### Delay

Also it seems there is a up to huge delay on the notification and an update of the counter, because you use the conversation values rather than the conversation store? Or is that not updated while the chat in the sidebar is hidden?

PVince81:
> Also it seems there is a up to huge delay on the notification and an update of the counter, because you use the conversation values rather than the message store? Or is that not updated while the chat in the sidebar is hidden?

As far as I know, the messages store always works independently from the conversation store. We don't infer mention count or new message count from the newly fetched messages. So the delay is when the conversation store reloads those values from the server. If there is a safe way to compute those we could add that separately.

The duplicate notification looks fishy though, I'll have a look.

nickvergessen:
> If there is a safe way to compute those we could add that separately.

the number of items in the store? For mentions you would have to check the parameters, so maybe you don't want to do that but we only have a indicator for new messages for now

PVince81:
> > If there is a safe way to compute those we could add that separately.
> 
> the number of items in the store?

Not enough, you need to know the message id of the last read message and then iterate over all the "next messages" after that in the array to find out how many are unread. If we're fine adding that logic I could do that and remove the "you have been mentioned" part. Depends what is more important: a quick notification or an accurate notification ?
I guess quick is better as "time spent in meetings is precious"

src/components/TopBar/TopBar.vue
```diff
@@ -282,6 +312,33 @@ export default {
...
+				}
+			}
+		},
+
+		hasUnreadMentions(newValue, oldValue) {
+			if (!this.isInCall || !this.showOpenSidebarButton) {
+				return
+			}
+
+			if (newValue && this.unreadMessagesCounter > 0) {
```
PVince81, line 336: this must be "> 1" to prevent the duplicated "you have been mentioned" notification.

but let's see if we decide to remove this message altogether in the main thread

nickvergessen:
> Depends what is more important: a quick notification or an accurate notification ?
I guess quick is better as "time spent in meetings is precious"

Well looking at what this is about, the instant is more important than accuracy. If you are unlucky and the poke is 29sec away... it's of no use anymore.

PVince81:
@nickvergessen I did the jump and also added notification parsing based on the logic I found in Mention.vue.

So now when you post a message concurrently the counter bubble and its highlight happen more quickly.
Please retest.

If this approach is acceptable I'll then move on to update the JS tests for the new "faster update" logic.

- [ ] update unit tests

PVince81:
@nickvergessen mind confirming whether the approach I used is acceptable or whether it's incomplete or risky ?

src/store/messagesStore.js
```diff
@@ -632,23 +672,47 @@ const actions = {
...
+						hasNewMention = true
+					}
+				}
 				lastMessage = message
 			}
+
+			// in case we encounter an already read message, reset the counter
+			// this is probably unlikely to happen unless one starts browsing from
+			// an earlier page and scrolls down
+			if (conversation.lastReadMessage.id === message.id) {
```
nickvergessen, line 690: conversation.lastReadMessage is an integer on the api?

src/store/conversationsStore.js
```diff
@@ -121,6 +121,15 @@ const mutations = {
 		Vue.set(state.conversations[token], 'lastMessage', lastMessage)
 	},
 
+	updateUnreadMessages(state, { token, unreadMessages, unreadMention }) {
```
nickvergessen, line 124: this will be wiped after 30 seconds by the room list update?

PVince81: Yes, exactly. It's just a shortcut to make it updater faster.

in my test I waited for 30 seconds to verify that the values were the same

nickvergessen: Okay seems to work somehow. We are not marking the chat read anymore when not open I guess.

PVince81: yes, this is new behavior on master since the introduction of the unread marker

PVince81:
> When the first message is a mention I get this:
> 
> Bildschirmfoto von 2021-06-17 16-00-12
> 

I thought this is what you meant when you commented on the line that had "unreadMessages > 1" instead of "> 0".
Sadly I force pushed back then so can't re-read it.

If that's a concern I can make the messages to be a single one again, with "you have been mentionned" with higher priority.

src/store/messagesStore.js
PVince81: indeed, I thought it was `conversation.lastMessage` which is a full object.
fixing...

nickvergessen:
What I ment was:

1. someone writes a message:
> You have new unread messages in the chat

2. (You still didn't check the chat) someone mentions you:
> You have been mentioned in the chat

What I didn't mean:
1. You close the chat
2. Someone writes a single message `@admin`
3. You see 2 notifications for 1 message

PVince81:
now thinking, I could simply store the notification handle and hide it to make sure that we only ever see one notification.

so in the first scenario you'd only see "you have been mentioned" at the end.

nickvergessen:
> so in the first scenario you'd only see "you have been mentioned" at the end.

Which is fine, but when there were like 5 mins between the normal message and a mention its good to reping I guess. That's why I ment that 2 notifications are prolly fine

PVince81:
> > so in the first scenario you'd only see "you have been mentioned" at the end.
> 
> Which is fine, but when there were like 5 mins between the normal message and a mention its good to reping I guess. That's why I ment that 2 notifications are prolly fine

Yes. I've just pushed after testing and this is how it behaves now.

PVince81:
unit test added for the messages store adjustments.

TopBar component tests have not been bootstrapped yet, maybe in a separate PR

marcoambrosini (commented):
I slightly modified the position of the icon and the counter.

One thing I'm not convinced is the toast for new messages. I think it's really invasive, especialle because it covers buttons making them unclickable. I'd much rather leave it as is or add a subtle animation to the counter each time the number increases. What do you think?

PVince81:
> One thing I'm not convinced is the toast for new messages.

This is a general problem, also when having connection issues or device issues.
Ideal would be to be able to define an offset for the toast container, or is it possible to specify a different container perhaps ?

To fix the messed up rebase, I force pushed my local work and then cherry-picked your last commit, then rebased again.
The commit history should be fine now.

PVince81:
for the toast notification, I found the ["selector" attribute](https://github.com/nextcloud/nextcloud-dialogs/blob/master/lib/toast.ts#L74), but setting it to "#call-container" didn't shift it.

PVince81:
@marcoambrosini I managed to shift the notification down, see https://github.com/nextcloud/spreed/pull/5794

nickvergessen (approved):
Nice and neat addition

PR #5542, opened 2021-04-30
Make sure the user we are testing with exists

nickvergessen:
Consequence from https://github.com/nextcloud/server/pull/26572


PVince81 (approved):
:+1:

PR #5565, opened 2021-05-03
Added JS tests for messages store

PVince81:


PVince81:
seems something slipped:
```
(node:31048) UnhandledPromiseRejectionWarning: [object Object]
(node:31048) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 7)
(node:31048) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

but jest doesn't fail when this happens

PVince81:
seems there are side effects, some of the mocked functions seem to expect resolving multiple times... I'll double check the execution order and cleanup approaches...

PVince81:
okay, found it: it's the cancellation test messing up.

lesson learned: if calling `dispatch()` without `await`, the mock still needs to have its promise resolved later before the end of the test, otherwise Node will complain

PVince81:
finally! I found the shortest solution: when calling `dispatch()` without an await to keep the request pending, either the promise needs to be resolved manually or add a catch block, like `dispatch(...).catch(() => {})`. This will prevent Node to complain that there's an unresolved Promise at the end of the tests.

nickvergessen (approved):

PR #5567, opened 2021-05-05
Added tests for RoomSelector

PVince81:
Might not be the most important component to test but I wanted to practice writing Vue tests first

### DIFF
--- a/src/test-setup.js
+++ b/src/test-setup.js
@@ -42,7 +42,9 @@ global.OCA = {
 	},
 }
 global.t = jest.fn().mockImplementation((app, text) => text)
+global.n = jest.fn().mockImplementation((app, text) => text)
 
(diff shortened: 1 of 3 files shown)


src/test-setup.js
```diff
@@ -36,4 +37,12 @@ global.OC = {
 		return 'en_GB'
 	},
 }
+global.OCA = {
+	Talk: {
+	},
+}
 global.t = jest.fn().mockImplementation((app, text) => text)
+
+Vue.prototype.t = global.t
```
nickvergessen, line 46: Also add n already

PVince81: I wasn't sure how to mock it in a good way without replicating the full logic. I'll just make it return the first text then.

nickvergessen (approved):

PR #5570, opened 2021-05-06
JS tests for Message component

PVince81:
- [x] TODO: more coverage
- [x] issue: seems it's not possible to detect the icon components as they are stateless => use `findComponent(Check)` instead of `findComponent({ name: 'Check' })` when dealing with icon components
- [x] bug: need better date mocking for the 6h deletion difference check (can't use a single mock here)
- [x] REQUIRES: https://github.com/nextcloud/spreed/pull/5586

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.spec.js
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.spec.js
@@ -1,0 +1,1060 @@
+import Vuex from 'vuex'
+import { createLocalVue, shallowMount } from '@vue/test-utils'
+import { cloneDeep } from 'lodash'
+import { EventBus } from '../../../../services/EventBus'
+import storeConfig from '../../../../store/storeConfig'
+import { CONVERSATION, PARTICIPANT, ATTENDEE } from '../../../../constants'
+import Check from 'vue-material-design-icons/Check'
+import CheckAll from 'vue-material-design-icons/CheckAll'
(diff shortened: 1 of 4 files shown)


PVince81:
almost there, it was a bumpy road especially for triggering clicks!!!
date issue is fixed as well

src/components/MessagesList/MessagesGroup/Message/Message.spec.js
```diff
@@ -0,0 +1,1060 @@
...
+				testDeleteMessageVisible(true, null, PARTICIPANT.TYPE.MODERATOR)
+			})
+
+			test('shows delete action on other people messages for owner', () => {
+				messageProps.actorId = 'another-user'
+				conversationProps.type = CONVERSATION.TYPE.PUBLIC
+				testDeleteMessageVisible(true, null, PARTICIPANT.TYPE.OWNER)
+			})
+
+			test('does not show delete action even for guest moderators', () => {
```
PVince81, line 798: @nickvergessen I assume this is intentional

nickvergessen: https://github.com/nextcloud/spreed/blob/2c3f80bb08793fec62a95478187ffafffb35bbff/lib/Controller/ChatController.php#L557

Allow guest moderators is false, so yeah only normal moderators/owners

nickvergessen (approved):

PR #5577, opened 2021-05-06
Update list of system messages

nickvergessen:
Fixing android atm, @Ivansss can you check iOS to show those as system messages properly?


PVince81 (approved):
:+1:

PR #5584, opened 2021-05-07
Remove the description from the system message

nickvergessen:
Looked a bit too weird in some clients with multiline descriptions and the quotes


ma12-co (approved):
source

PR #5598, opened 2021-05-10
JS tests for MessagesList

PVince81:
~~See skipped tests as TODO list~~

I didn't cover the bits with the read marker and scroller gymnastics as they require a real viewport and I'm not sure if testing individual conditions here with a faked viewport make sense.

So let's move forward with what I have already here.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -52,8 +52,10 @@ export const ATTENDEE = {
 		GUESTS: 'guests',
 		EMAILS: 'emails',
 		GROUPS: 'groups',
+		BOTS: 'bots',
 	},
 	BRIDGE_BOT_ID: 'bridge-bot',
+	CHANGELOG_BOT_ID: 'changelog',
 }
 export const PARTICIPANT = {
 	CALL_FLAG: {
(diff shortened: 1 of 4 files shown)


jest.config.js
```diff
@@ -24,6 +24,9 @@ module.exports = {
 	preset: '@vue/cli-plugin-unit-jest/presets/no-babel',
 	// Allow tests in the src and in tests/unit folders
 	testMatch: ['<rootDir>/src/**/*.(spec|test).(ts|js)'],
+	transformIgnorePatterns: [
+		'node_modules/(?!(vue-material-design-icons|@juliushaertl)/)',
```
nickvergessen, line 28: this is already there.. shouldn't there be a conflict?

PVince81: I added it to multiple PRs to be able to move forward. Interesting that it neither got discarded nor produced a conflict.
I'll rebase and it should disappear.

nickvergessen (approved):

src/constants.js
```diff
@@ -52,8 +52,10 @@ export const ATTENDEE = {
 		GUESTS: 'guests',
 		EMAILS: 'emails',
 		GROUPS: 'groups',
+		BOTS: 'bots',
```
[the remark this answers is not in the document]
PVince81: yeah, it feels a bit weird. but as it's been used in the `message.actorType` field anyway I thought it would be okay.

PR #5608, opened 2021-05-12
JS tests for LeftSidebar

PVince81:
- ~~move service calls to the store~~ maybe another time...
- [x] see skipped tests as TODOs

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -174,8 +174,9 @@ export default {
 			initialisedConversations: false,
 			cancelSearchPossibleConversations: () => {},
 			cancelSearchListedConversations: () => {},
-			// Keeps track of wheteher the conversation list is scrolled to the top or not
+			// Keeps track of whether the conversation list is scrolled to the top or not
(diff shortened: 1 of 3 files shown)


src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -29,7 +29,6 @@
 				class="conversations-search"
 				:is-searching="isSearching"
 				@input="debounceFetchSearchResults"
-				@submit="onInputEnter"
```
nickvergessen, line 32: https://github.com/nextcloud/spreed/commit/8c34922084f92b1cdfbe01352ae145ca7f441571

I just fixed this recently?
Did you solve a conflict wrong or why do you remove that hitting enter opens the first conversation?

PVince81: oh, this is something I found while testing: there was no handler for "onInputEnter" so I removed this line

I've double checked now and just realized that the handler is hidden in a mixin... okay, I'll revert this

PVince81: reverted

nickvergessen (approved):

Issue #5631, opened 2021-05-18
Error in log related to router and undefined handler symbol


PVince81:
## Steps to reproduce
1. Create two conversations A and B
2. Switch between A and B
3. Observe the JS console

### Expected behaviour
No errors

### Actual behaviour
<img width="408" alt="image" src="https://user-images.githubusercontent.com/277525/118639037-c374de80-b7d7-11eb-83c7-5fa9b49ef36c.png">

## Talk app

**Talk app version:** git master 23d1fc70c517b244848d0dd13aba9ad658bd03f9

PVince81:
bisect has narrowed it down to 24e5aad38 from #5605 

however I don't see anything in that commit that would have an effect on the router

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/5632

PR #5605, opened 2021-05-11
Move description editor to conversation settings dialog

ma12-co:
https://user-images.githubusercontent.com/26852655/117831559-94f18380-b26c-11eb-852f-24d466482f8c.mp4

Signed-off-by: Marco Ambrosini <marcoambrosini@pm.me>

PR #5632, opened 2021-05-18
Remove leftover routeChange handlers

PVince81:
Fixes a JS error due to missing handlers.

This was missing from https://github.com/nextcloud/spreed/pull/5605

Fixes https://github.com/nextcloud/spreed/issues/5631


danxuliu (approved):
Tested and works :+1: 

Also fixes the error in the following scenario:
- Open a conversation
- Reload the page
- Start a call

PR #5633, opened 2021-05-18
Reusable user ids one to one chats

nickvergessen:


lib/Manager.php
```diff
@@ -357,6 +357,34 @@ public function getRoomsForActor(string $actorType, string $actorId, array $sess
...
+		$helper = new SelectHelper();
+		$helper->selectRoomsTable($query);
+		$query->from('talk_rooms', 'r')
+			->where($query->expr()->eq('r.type', $query->createNamedParameter(Room::ONE_TO_ONE_CALL)))
+			->andWhere($query->expr()->like('r.name', $query->createNamedParameter('%' . $this->db->escapeLikeParameter(json_encode($userId)) . '%')));
+
+		$result = $query->execute();
+		$rooms = [];
+		while ($row = $result->fetch()) {
+			if ($row['token'] === null) {
```
PVince81, line 375: still relevant today or copy-paste issue ?

lib/Listener/UserDeletedListener.php
```diff
@@ -52,13 +52,18 @@ public function handle(Event $event): void {
...
 				$room->deleteRoom();
 			} else {
 				$this->participantService->removeUser($room, $user, Room::PARTICIPANT_REMOVED);
 			}
 		}
+
+		$leftRooms = $this->manager->getLeftOneToOneRoomsForUser($user->getUID());
+		foreach ($leftRooms as $room) {
+			$room->setType(Room::GROUP_CALL, true);
+			$room->setName($user->getDisplayName(), '');
```
PVince81, line 66: add a comment to explain what this implies: that the room won't be found any more by name when trying to recreate the same 1-1 conversation

lib/Listener/UserDeletedListener.php
```diff
@@ -52,13 +52,18 @@ public function handle(Event $event): void {
...
 			if ($this->participantService->getNumberOfUsers($room) === 1) {
 				$room->deleteRoom();
 			} else {
 				$this->participantService->removeUser($room, $user, Room::PARTICIPANT_REMOVED);
 			}
 		}
+
+		$leftRooms = $this->manager->getLeftOneToOneRoomsForUser($user->getUID());
+		foreach ($leftRooms as $room) {
+			$room->setType(Room::GROUP_CALL, true);
```
PVince81, line 65: this means other users could be invited to that conversation and then they could read past messages, feels like an unwanted side effect.
if we had a way to mark it as archived, like read-only mode but would need an external system user moderator then to prevent anyone to remove there read-only flag

lib/Manager.php
nickvergessen: copy pasta and will remain until we clean it up in all methods with one go

lib/Listener/UserDeletedListener.php
nickvergessen: No one cares? Don't want to make it overly complicated so that is good enough. Before you could do screenshares and share them already `¯\_(ツ)_/¯`

I'm fine with marking it archived by default as well. We could also "demote" the other user to be a normal user and they can then leave the conversation only. But simplicity is the key here I think.

lib/Listener/UserDeletedListener.php
nickvergessen: Added

lib/Listener/UserDeletedListener.php
PVince81: I've tested it and seems fine then.

A good thing is that now that it's a group conversation, the remaining user is able to delete it also.

PVince81 (approved):
:+1:

nickvergessen:
Failing sharing tests are fixed in https://github.com/nextcloud/spreed/pull/5634

PR #5652, opened 2021-05-21
Fix chatview

ma12-co:
Fix a regression introduced with #5596 


PVince81 (approved):
:+1:

PR #5667, opened 2021-05-27
Show name for bridged messages

gary-kim:
~~Requires https://github.com/42wim/matterbridge/pull/1506.~~ **Now merged!**

The way the current iteration works is that messages received from the `bridge-bot` user is labeled as Actor Type `bridged`. Actor Type `bridged` messages store the message author name in the Actor Id field in `oc_comments`. When serving to the frontend, this is slightly changed to put what is in Actor Id into Actor Display Name and set the Actor Id to `bridge-bot` (the reason it was done this way was to simplify the changes that would need to be made to clients).

**As far as clients have to care, the only difference is the addition of the Actor Type `bridged` and that for `bridged` messages, the Actor Display Name can be different for every message.**

### Screenshots
Screenshot 2021-05-28 at 09-52-29 Element  20  Test Room
Screenshot 2021-05-28 at 09-51-48 Test 2 - Talk - Nextcloud

### DIFF
--- a/lib/Controller/ChatController.php
+++ b/lib/Controller/ChatController.php
@@ -160,7 +160,7 @@ class ChatController extends AEnvironmentAwareController {
 			if ($actorDisplayName) {
 				$this->guestManager->updateName($this->room, $this->participant, $actorDisplayName);
 			}
-		} elseif ($this->userId === MatterbridgeManager::BRIDGE_BOT_USERID) {
+		} elseif ($this->userId === MatterbridgeManager::BRIDGE_BOT_USERID && $actorDisplayName) {
 			$actorType = Attendee::ACTOR_BRIDGED;
 			$actorId = str_replace(["/", "\""], "", $actorDisplayName);
 		} else {
--- a/lib/Controller/ChatController.php
+++ b/lib/Controller/ChatController.php
@@ -160,7 +160,7 @@ class ChatController extends AEnvironmentAwareController {
 			if ($actorDisplayName) {
 				$this->guestManager->updateName($this->room, $this->participant, $actorDisplayName);
 			}
-		} else if ($this->userId === MatterbridgeManager::BRIDGE_BOT_USERID) {
+		} elseif ($this->userId === MatterbridgeManager::BRIDGE_BOT_USERID) {
 			$actorType = Attendee::ACTOR_BRIDGED;
 			$actorId = str_replace(["/", "\""], "", $actorDisplayName);
 		} else {
--- a/lib/Controller/ChatController.php
+++ b/lib/Controller/ChatController.php
@@ -556,6 +556,9 @@ class ChatController extends AEnvironmentAwareController {
 		$attendee = $this->participant->getAttendee();
 		$isOwnMessage = $message->getActorType() === $attendee->getActorType()
 			&& $message->getActorId() === $attendee->getActorId();
+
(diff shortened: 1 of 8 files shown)


lib/MatterbridgeManager.php
```diff
@@ -383,6 +383,7 @@ private function generateConfig(array $bridge): string {
 					$serverUrl = $part['server'];
 				} else {
 					$serverUrl = preg_replace('/\/+$/', '', $this->urlGenerator->getAbsoluteURL(''));
+					$content .= "	SeparateDisplayName = true" ."\n";
```
nickvergessen, line 386: Does this mean all configs have to be recreated?

src/components/MessagesList/MessagesGroup/AuthorAvatar.vue
```diff
@@ -73,7 +74,7 @@ export default {
 			return this.authorType === 'bots' && this.authorId === 'changelog'
 		},
 		isUser() {
-			return this.authorType === 'users'
+			return this.authorType === 'users' || this.authorType === ATTENDEE.ACTOR_TYPE.BRIDGED
```
nickvergessen, line 77: Can you add this in `disableMenu` too, so the bridgebot has no avatar menu?

lib/MatterbridgeManager.php
gary-kim: Yes. Until the config is recreated, it will continue to function as before.

nickvergessen:
> Requires 42wim/matterbridge#1506

I guess we have to wait until that is merged?

gary-kim:
Yup. It'd be noop if merged right now

nickvergessen:
noop in terms of "works as before" or in terms of "breaks and doesn't do anything anymore"

gary-kim:
Works as before

lib/Controller/ChatController.php
```diff
@@ -160,6 +160,9 @@ protected function getActorInfo(string $actorDisplayName = ''): array {
 			if ($actorDisplayName) {
 				$this->guestManager->updateName($this->room, $this->participant, $actorDisplayName);
 			}
+		} elseif ($this->userId === MatterbridgeManager::BRIDGE_BOT_USERID && $actorDisplayName) {
+			$actorType = Attendee::ACTOR_BRIDGED;
```
nickvergessen, line 164: Missing from documentation:
https://github.com/nextcloud/spreed/blob/master/docs/constants.md#actor-types-of-chat-messages

gary-kim:
If everything seems fine on this side, we could merge now. It should function the same as before with a Matterbridge version that does not support the `SeperateDisplayName` config.

nickvergessen:
Maybe you can ping them again?

gary-kim:
https://github.com/42wim/matterbridge/pull/1506 is now merged so this feature should now work with the executables here: https://github.com/42wim/matterbridge/actions/runs/953056470.

nickvergessen (approved):
Good from my POV, but I guess we should wait for 22 to go out as the mobile apps need a patch too

gary-kim:
Would have been nice to get it into 22 but we're lacking in time, I suppose

nickvergessen:
Android patch at https://github.com/nextcloud/talk-android/pull/1355
iOS ticket raised in https://github.com/nextcloud/talk-ios/issues/579 but at least it shows something already.

nickvergessen:
Can you rebase @gary-kim

PR #5670, opened 2021-05-27
Feature/noid/circle tracking

nickvergessen:


lib/Listener/CircleMembershipListener.php
```diff
@@ -0,0 +1,144 @@
...
+		if ($event instanceof RemovingCircleMemberEvent) {
+			$this->removeFormerMemberFromRooms($event->getCircle(), $event->getMember());
+		}
+	}
+
+	protected function addingCircleMemberEvent(AddingCircleMemberEvent $event): void {
+		$roomsForTargetCircle = $this->manager->getRoomsForActor(Attendee::ACTOR_CIRCLES, $event->getCircle()->getSingleId());
+
+		if (empty($roomsForTargetCircle)) {
+			// The circle is not in any room => bye!
```
daita, line 81: Careful, the Member can be added to a Circle that is a member of another Circle that is itself an Actor of one of your Room.
In fact, at this point, you might want to check `$event->getCircle()->getMemberships()` (returns the list of Parent Circles the `$event->getCircle()` is members of.

nickvergessen: Done

lib/Listener/AMembershipListener.php
```diff
@@ -0,0 +1,119 @@
...
+
+	protected function filterRoomsWithOtherCircleMemberships(array $rooms, IUser $user): array {
+		if (!$this->appManager->isEnabledForUser('circles', $user)) {
+			\OC::$server->getLogger()->error('Circles not enabled', ['app' => 'nickv']);
+			return $rooms;
+		}
+
+		try {
+			/** @var Circles $circlesApi */
+			$circles = Circles::joinedCircles($user->getUID());
```
nickvergessen, line 100: @ArtificialOwl

1. Is there a better way to get all circles of a user as this method is deprecated?
2. I guess I also need the parent circles of each to see if any of those is a member in a room. What's the API call for that?

nickvergessen:
@danxuliu @ArtificialOwl ready for review. Would be nice to get it in, before we start working towards federation

danxuliu (commented):
I have made an initial pass on the code and it looks good, but I have not tested it and I am not familiar with circles either, so... you can treat this as a half approval :-P

Nevertheless I will try to do some tests to get a better grasp of all it.

lib/Listener/CircleDeletedListener.php
```diff
@@ -0,0 +1,62 @@
...
+	private $participantService;
+
+	public function __construct(Manager $manager,
+								ParticipantService $participantService) {
+		$this->manager = $manager;
+		$this->participantService = $participantService;
+	}
+
+	public function handle(Event $event): void {
+		if (!($event instanceof CircleDestroyedEvent)) {
```
ArtificialOwl, line 48: For better compatibility in globalscale, I would use `DestroyingCircleEvent` as `CircleDestroyedEvent` is only called on the master instance of the Circle (the instance to which the owner of the circle belongs to)

lib/Listener/CircleMembershipListener.php
```diff
@@ -0,0 +1,179 @@
...
+		$newMember = $event->getMember();
+		// Get the base circle of the membership
+		$basedOnCircle = $newMember->getBasedOn();
+		// Get all (nested) memberships in the added $newMember as a flat list
+		$userMembers = $basedOnCircle->getInheritedMembers();
+
+		$invitedBy = $newMember->getInvitedBy();
+		if ($invitedBy->getUserType() === Member::TYPE_USER && $invitedBy->getUserId() !== '') {
+			$this->session->set('talk-overwrite-actor', $invitedBy->getUserId());
+		} elseif ($invitedBy->getUserType() === Member::TYPE_APP && $invitedBy->getUserId() === 'occ') {
```
ArtificialOwl, line 109: while userId='occ' have no reason to change, `$invitedBy->getUserId() === 'occ'` can be replaced by `$invitedBy->getBasedOn()->getSource() === Member::APP_OCC`

lib/Listener/CircleMembershipListener.php
```diff
@@ -0,0 +1,179 @@
...
+	protected function addNewMemberToRooms(array $rooms, Member $member): void {
+		if ($member->getUserType() !== Member::TYPE_USER || $member->getUserId() === '') {
+			// Not a user?
+			return;
+		}
+
+		$user = $this->userManager->get($member->getUserId());
+		if (!$user instanceof IUser) {
+			return;
+		}
```
ArtificialOwl, line 129: overkill, I would just go with `if ($member->getUserType() !== Member::TYPE_USER || !$member->isLocal()) { return; }`

ArtificialOwl (commented):
Code looks good, I will do some live test as I am not sure how you handle the destroyed circle that contains circles-as-members

changes noted are optional

lib/Listener/CircleDeletedListener.php
nickvergessen: Since we don't have federated users yet, it shouldn't matter?

lib/Listener/CircleMembershipListener.php
nickvergessen: yields a psalm warning, so better save then sorry

lib/Listener/AMembershipListener.php
nickvergessen: Done

PR #5690, opened 2021-05-31
l10n: Improve styling, appearance and fix l10n

Valdnet:


src/components/SettingsDialog/SettingsDialog.vue
```diff
@@ -96,7 +96,7 @@
 				</div>
 			</dl>
 
-			<h3>{{ t('spreed', 'Shortcuts while in a call') }}</h3>
+			<em>{{ t('spreed', 'Shortcuts while in a call.') }}</em>
```
nickvergessen, line 99: I prefer h3 here. with the light grey it becomes less readable and feels then more like noise opposed to helping orientate in the long list.

src/components/SettingsDialog/SettingsDialog.vue
```diff
@@ -163,7 +163,7 @@ export default {
 		},
 
 		locationHint() {
-			return t('spreed', 'Choose in which folder attachments should be saved.')
+			return t('spreed', 'Choose folder in which attachments should be saved.')
```
nickvergessen, line 166:
```suggestion
			return t('spreed', 'Choose the folder in which attachments should be saved.')
```

src/components/SettingsDialog/SettingsDialog.vue
PVince81: for accessibility, if you remove h3 you'll need to add `role='heading' aria-level='3'` to keep the heading semantics for screen readers and others

in general I think it's a bad idea to rely on html markup for styling, we should rather use CSS for that

Valdnet:
Fixes made 😉.

nickvergessen (approved):

PR #5691, opened 2021-05-31
Deduplicate and bug-fix circles integration

nickvergessen:
As I'm not sure we will finish #5670 in time, here is a quick fix to fix the "Undefined method `getType`" error on circle members as the circles app evolved and renamed the method.

### DIFF
--- a/lib/Service/ParticipantService.php
+++ b/lib/Service/ParticipantService.php
@@ -23,6 +23,8 @@ declare(strict_types=1);
 
 namespace OCA\Talk\Service;
 
+use OCA\Circles\Model\Circle;
+use OCA\Circles\Model\Member;
 use OCA\Talk\Config;
 use OCA\Talk\Events\AddParticipantsEvent;
 use OCA\Talk\Events\JoinRoomGuestEvent;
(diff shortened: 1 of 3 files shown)


lib/Service/ParticipantService.php
```diff
@@ -361,6 +363,77 @@ public function addGroup(Room $room, IGroup $group, array $existingParticipants
...
+//			$this->attendeeMapper->findByActor($room->getId(), Attendee::ACTOR_CIRCLES, $circle->getSingleId());
+//		} catch (DoesNotExistException $e) {
+//			$attendee = new Attendee();
+//			$attendee->setRoomId($room->getId());
+//			$attendee->setActorType(Attendee::ACTOR_CIRCLES);
+//			$attendee->setActorId($circle->getSingleId());
+//			$attendee->setDisplayName($circle->getDisplayName());
+//			$attendee->setParticipantType(Participant::USER);
+//			$attendee->setReadPrivacy(Participant::PRIVACY_PRIVATE);
+//			$this->attendeeMapper->insert($attendee);
```
PVince81, line 431: I didn't find this code in the non-refactored version so I assume this is a "would-be in the future" that you've just added here.

to me, without doing a line by line compare, it looks like this PR only contains refactoring but no actual behavior change, since the above is commented out

can you point to the workaround/fix for the getType issue ? (or split the PR into one commit for refactoring and one for the fix)

nickvergessen: https://github.com/nextcloud/spreed/pull/5691/files#diff-d583983624ef8769460d4aa59e64b75ee54b6955e01e620e3eacedb7ba208786L717
and
https://github.com/nextcloud/spreed/pull/5691/files#diff-d583983624ef8769460d4aa59e64b75ee54b6955e01e620e3eacedb7ba208786L1062
both used getType

The fixed method uses getUserType:
https://github.com/nextcloud/spreed/pull/5691/files#diff-8d29629ca6909ff3b00eb3ff13094b704d2b10ce22eff12e6921dccf08a83100R388

The commented out code from here is basically from #5670 and because this function is basically a close of addGroup() right above.

PVince81: thanks for the explanation, makes sense

PVince81 (approved):
:+1:

nickvergessen:
Fixed psalm by ignoring the circles classes 🙈

Issue #5492, opened 2021-04-16
Upcoming issues with nextcloud-vue > 3.9.0


PVince81:
:warning: requires waiting for the next nextcloud-vue release to become effective :warning: 
Not sure if fixing these early would cause side effects. Backlogged for now.

## Steps to reproduce
1. Check out https://github.com/nextcloud/spreed/pull/5490 to avoid compilation issues
2. Run the steps from https://github.com/nextcloud/spreed/issues/5489 to get latest nextcloud-vue that comes after v3.9.0
3. Navigate through the Talk UI and notice what's broken

Note: without the npm link there are no styling issues, so those are definitely related to recent changes, very likely to https://github.com/nextcloud/nextcloud-vue/pull/1816

## Styling issues

- [x] misaligned icons: https://github.com/nextcloud/spreed/pull/5697

<img width="380" alt="image" src="https://user-images.githubusercontent.com/277525/115004790-784d7000-9ea7-11eb-8510-9c0e6dd9f835.png">

- [ ] spinner of avatar component is misaligned with the avatar component when opening the menu: https://github.com/nextcloud/nextcloud-vue/pull/1999

- [x] screen share icon is invisible in call view (not any more ?)
- ~~screen share menu icons are white on grey instead of black/dark~~ => already on master

I went to many places and didn't find anything else, so the impact is rather minimal.

PVince81:
fix for screen share icons here https://github.com/nextcloud/spreed/pull/5698

PVince81:
so the avatar component/menu has multiple glitches:
<img width="198" alt="image" src="https://user-images.githubusercontent.com/277525/120447910-52aef400-c38b-11eb-8788-20a00221f667.png">

possibly a bug in the lib itself

PVince81:
the lib update was done.

but brand new issues have appeared now: https://github.com/nextcloud/spreed/issues/5747

PR #5697, opened 2021-06-02
Upload icon is now material design "attachment"

PVince81:
Use attachment icon from material design for file uploads.
Also fixes an alignment issue with nextcloud-vue 4.0.0.

Part of https://github.com/nextcloud/spreed/issues/5492 to prepare for the 4.0.0 upgrade.

Before:
image

After:
image

Note: the icon is not visually identical and is rotated differently. If that's an issue we could use CSS to rotate it...

### DIFF
--- a/src/components/NewMessageForm/NewMessageForm.vue
+++ b/src/components/NewMessageForm/NewMessageForm.vue
@@ -45,7 +45,7 @@
 						:disabled="disabled"
 						:aria-label="t('spreed', 'Share files to the conversation')"
 						:aria-haspopup="true">
-						<Attachment
+						<Paperclip
 							slot="icon"
 							:size="16"
 							decorative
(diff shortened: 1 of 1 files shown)


Ivansss:
Could we use the "vertical" clip icon for the attachments?
That's whe one we are using in the iOS app.
https://fonts.google.com/icons?selected=Material%20Icons%20Two%20Tone%3Aattach_file

PVince81:
> Could we use the "vertical" clip icon for the attachments?
> That's whe one we are using in the iOS app.
> https://fonts.google.com/icons?selected=Material%20Icons%20Two%20Tone%3Aattach_file

ah! when browsing https://materialdesignicons.com/ I overlooked that there's also a vertical one called "paperclip"
I'll use that one, then. thanks

Ivansss:
> ah! when browsing https://materialdesignicons.com/ I overlooked that there's also a vertical one called "paperclip"
> I'll use that one, then. thanks

Great! :) I think the vertical one looks better.

PVince81:
here w go:

image

nickvergessen (approved):

PR #5698, opened 2021-06-02
Fix color of screen share menu icons

PVince81:
Remove fill color to avoid white on grey.

Before:
image

After:
image


nickvergessen (approved):
Opsy

PR #5716, opened 2021-06-07
Update API docs index.md links + extra constants

PVince81:


nickvergessen (approved):

docs/constants.md
```diff
@@ -52,12 +52,15 @@ title: Constants
...
 * `users` - Logged-in users
+* `groups` - Groups
 * `guests` - Guest without a login
 * `emails` - A guest invited by email address
 
 ### Actor types of chat messages
 * `guests` - guest users
 * `users` - logged-in users
+* `emails` - users invited by email
+* `groups` - groups
```
nickvergessen, line 63: actually groups can't be author of chat messages.

PVince81: was this a mistake ? https://github.com/nextcloud/spreed/blob/master/src/constants.js#L54

nickvergessen: Attendees can have such actor types, but not chat messages.

PVince81: ah, got it. I never realized that actor type constants were used in two places, I'll remove that line

PVince81: https://github.com/nextcloud/spreed/pull/5722

PR #5717, opened 2021-06-07
Document circles too

nickvergessen:
* Circles are currently not stored and returned due to #5670 
* But by adding them already the mobile clients are future proof and we don't have to update the API with a v5 after the PR is merged


PVince81 (approved):
:+1:

PR #5718, opened 2021-06-07
Fix casing of exported initWebRTC function

PVince81:
In webrtc/index.js we import the function as WebRtc.
The name of the function at export time is semantically irrelevant but
having it correct might help when grepping.

In theory we could also remove that name altogether.

Tested with HPB to make sure this doesn't break.


nickvergessen (approved):
¯\_(ツ)_/¯

Issue #5733, opened 2021-06-09
Random 404 when deleting room with session


PVince81:
## Steps to reproduce

1. Create a room
2. Join the room
3. In the left sidebar delete the room
4. Be unlucky :four_leaf_clover: 

### Expected behaviour
Redirect to the main page, not "not-found" page and the room is deleted.

### Actual behaviour
Sometimes, the room is not deleted and the console returns a 404 error for the DELETE operation.

## Talk app

**Talk app version:** 11.2.1

**Custom Signaling server configured:** yes

## Server configuration

**Nextcloud Version:** 21.0.2

### Server log (data/nextcloud.log)

Nothing relevant.

## Analysis

This was reported to me internally, I was not able to reproduce this locally.

I've dug into the code and it looks like there are two requests happening in parallel when deleting a conversation:
1) [switching to the default route](https://github.com/nextcloud/spreed/blob/v11.2.1/src/components/LeftSidebar/ConversationsList/Conversation.vue#L329) triggers a DELETE to /room/$token/participant/active leading to https://github.com/nextcloud/spreed/blob/v11.2.1/lib/Controller/RoomController.php#L1659
2) [deleteConversation call](https://github.com/nextcloud/spreed/blob/v11.2.1/src/components/LeftSidebar/ConversationsList/Conversation.vue#L334) triggers a DELETE to /room/$token

The code does not wait for 1) to finish and directly triggers 2).

After adding more extensive logging on the target env where the problem occurs, I noticed that the 404 for the deletion actually does find the room, but fails to find the participant here: https://github.com/nextcloud/spreed/blob/v11.2.1/lib/Middleware/InjectionMiddleware.php#L156

It looks like depending whether the session is still there, which correlated with request 1), it might enter the if or else statement, using a different approach for finding the participant. A race condition could cause the session id to not exist any more in the database while it still existed at the time of retrieval from the PHP session.

I did more local tests with logging in both code paths and noticed that **the one or the other branch was randomly picked**, which confirms that there is some race condition in place. However in my local setup it never resulted in a 404.

From my understanding, it should never happen that the session id can be deleted in one PHP request while the other one is still reading it. Unless maybe the PHP session handling was configured differently on the target system...

### Approach 1: wait for request 1) to finish

Currently not possible because of the way the routing + onRouteChange events are done, these are not async/await-compatible on the event bus, so we can't wait for the "leaveConversation" to finish.

### Approach 2: discard request 1)

[...]

PR #5736, opened 2021-06-10
Don't trigger leave event when deleting conversation

PVince81:
When deleting a conversation, the route change would implicitly trigger
a leave request which ran in parallel to the conversation deletion one,
causing random issues in some systems like the latter returning a 404
instead of proceeding with the deletion.

This fix removes the route change on deletion. The new behavior will
implicitly redirect to the "not-found" page since the session does not
exist any more. The latter behavior is now aligned with the "Leave
conversation" action which also redirects to "not-found" page.

Fixes https://github.com/nextcloud/spreed/issues/5733
See ticket for details about the race condition and also alternative approaches that were considered but discarded.

### DIFF
--- a/src/components/LeftSidebar/ConversationsList/Conversation.spec.js
+++ b/src/components/LeftSidebar/ConversationsList/Conversation.spec.js
@@ -475,30 +475,6 @@ describe('Conversation.vue', () => {
 				expect(updateTokenAction).not.toHaveBeenCalled()
 			})
 
-			test('redirects when deleting current conversation', async() => {
-				const actionHandler = jest.fn().mockResolvedValueOnce()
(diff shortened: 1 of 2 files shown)


PVince81:
Tests:
- [x] TEST: with HPB, redirects to not-found page after deletion
- [ ] TEST: without HPB

PVince81:
aha, unit tests! need to practice that new habit

nickvergessen (approved):
Works just as fine, but over all it feels weird to read the following after a delete:
> ### The conversation does not exist
> 
> Join a conversation or start a new one!

Maybe we adjust this at some time to show a "Successfully deleted" instead of even redirect to the welcome view with a showSuccess()

PR #5743, opened 2021-06-11
Add JS test for AverageStatValue

PVince81:
Note: the values were taken from the existing calculation, and I haven't verified whether the weighted average values are mathematically correct, but at first glance they looked fine.

@danxuliu would be good if you could review and tell me whether the chosen test values make sense for testing this


nickvergessen (approved):

danxuliu (commented):
> would be good if you could review and tell me whether the chosen test values make sense for testing this

Yes, in my opinion the values make sense :+1:

In the real stats the cumulative values should be monotonically increasing, so the following value will never be lower than the previous value. But... that is the theory; in practice sometimes (corner cases) they decrease, so it is good that the cumulative stats are also tested with decreasing values.

Thanks for the test :-)

PR #5780, opened 2021-06-16
Don't refresh the room when the participant list changes

nickvergessen:
Currently when a call with 40 people is started and they join one by one
with 3.1s difference, each of the participant will fetch the conversation
40 times resulting in 1.6k requests.
At the same time the info doesn't really change anymore as the participant
details of other participants have been removed from the room API with
version 3. A refreshing of the room therefor only makes sense once
to update the hasCall and callFlag attributes.

The conversations callFlag however is not used at the moment and the worst
that could happen without the refreshing is the values updating on the next
refreshing of the complete room list which is done every 30 seconds.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -186,7 +186,6 @@ export default {
 		if (!getCurrentUser()) {
 			EventBus.$off('shouldRefreshConversations', this.debounceRefreshCurrentConversation)
 		}
-		EventBus.$off('Signaling::participantListChanged', this.debounceRefreshCurrentConversation)
 		document.removeEventListener('visibilitychange', this.changeWindowVisibility)
 	},
 
@@ -197,7 +196,6 @@ export default {
 			})
 			EventBus.$on('shouldRefreshConversations', this.debounceRefreshCurrentConversation)
 		}
-		EventBus.$on('Signaling::participantListChanged', this.debounceRefreshCurrentConversation)
 
 		if (this.$route.name === 'conversation') {
 			// Update current token in the token store


danxuliu (approved):
From [all the properties returned in the room data](https://github.com/nextcloud/spreed/blob/d9555a2ee0c6a0d8ba7fd7ace96b5a49efed01c3/lib/Controller/RoomController.php#L347-L385) as far as I can tell these are the ones that depend on the participants:
- participantType
- participantFlags
- hasCall
- canStartCall
- canDeleteConversation
- publishingPermissions
- canEnableSIP
- callFlag

The following properties depend only on the current participant type:
- participantType
- canStartCall
- canDeleteConversation
- canEnableSIP

Note that `hasCall` depends on the current participant type, as it ([will be reset if the lobby is active and the participant is demoted](https://github.com/nextcloud/spreed/blob/d9555a2ee0c6a0d8ba7fd7ace96b5a49efed01c3/lib/Controller/RoomController.php#L494)), but it also depends on the other participants.

The following properties depend only on the current participant call flags and publishing permissions:
- participantFlags (it depends on call flags but [it is also limited by the publishing permissions](https://github.com/nextcloud/spreed/blob/ad1de57c4cebb01268b9445fca48c8e6a4747ba7/lib/Service/ParticipantService.php#L718-L724))

And the following properties depend only on the current participant publishing permissions:
- publishingPermissions

Currently turns out that [participant type changes also trigger a room modified event](https://github.com/nextcloud/spreed/blob/4009de75a20434a484e2bdb1c85153e6fc364b45/lib/Signaling/Listener.php#L133), so not fetching the room again when the participant list changed will probably affect only to:
- participantFlags
- hasCall
- publishingPermissions
- callFlag

For now I think it is OK to not fetch the room again when the participant list changed to prevent the excessive number of requests :+1:

For a later stage we could try to do something like sending a `roomModified` event to everyone only when `hasCall` and `callFlag` have actually changed, and send a `roomModified` event only to the participant whose call flags and publishing permissions have changed. That should be enough to keep the responsiveness without causing too much requests (and if we could read the data from the events themselves rather than fetching the data again that would be even better).

The above would be for the external signaling server. For the internal signaling server I guess that we will need to add some additional events to cover all that.

PR #5787, opened 2021-06-16
Display voice message preview before uploading

marcoambrosini:
preview

### DIFF
--- a/src/components/UploadEditor.vue
+++ b/src/components/UploadEditor.vue
@@ -24,38 +24,45 @@
 		class="upload-editor"
 		container="#content-vue"
 		@close="handleDismiss">
-		<!--native file picker, hidden -->
-		<input id="file-upload"
-			ref="fileUploadInput"
-			multiple
-			type="file"
-			class="hidden-visually"
-			@change="handleFileInput">
-		<transition-group
-			class="upload-editor__previews"
-			name="fade"
-			tag="div">
-			<template v-for="file in files">
-				<FilePreview
-					:key="file.temporaryMessage.id"
-					v-bind="file.temporaryMessage.messageParameters.file"
-					:is-upload-editor="true"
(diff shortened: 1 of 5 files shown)


src/store/fileUploadStore.js
```diff
@@ -211,6 +211,8 @@ const actions = {
 			let localUrl = ''
 			if (file.type === 'image/png' || file.type === 'image/gif' || file.type === 'image/jpeg') {
 				localUrl = URL.createObjectURL(file)
+			} else if (isVoiceMessage) {
```
PVince81, line 214: add this arg to the JSDoc and extend the unit test fileUploadStore.spec.js ?

src/components/UploadEditor.vue
```diff
@@ -24,38 +24,45 @@
...
+				tag="div">
+				<template v-for="file in files">
+					<FilePreview
+						:key="file.temporaryMessage.id"
+						v-bind="file.temporaryMessage.messageParameters.file"
+						:is-upload-editor="true"
+						@remove-file="handleRemoveFileFromSelection" />
+				</template>
+				<div
+					:key="'addMore'"
```
nickvergessen, line 47: on a div? what does it do?

nickvergessen (approved):
Works,but see request from vincent

src/components/UploadEditor.vue
marcoambrosini: It's because it's inside a loop so it needs a key

src/store/fileUploadStore.js
marcoambrosini: @PVince81 I'll create an issue for the tests and I'll look into it once I have finished all the urgent tasks

src/components/UploadEditor.vue
nickvergessen: but its not a component and div html elements don't have a key attribute?

PR #5794, opened 2021-06-18
Offset toastify when top bar is visible

PVince81:
To avoid overlapping with the top bar, the toastify notification is
shifted down.

image

Doesn't look that fantastic though, wondering if we should move the notifications to the bottom instead. The latter might need more involved changes as it would require support for global toastify options in nextcloud-dialogs.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -451,6 +451,13 @@ export default {
 }
 </script>
 
+<style lang="scss">
+/** override toastify position due to top bar */
+body.has-topbar .toastify-top {
+	top: 65px !important;
+}
+</style>
+
 <style lang="scss" scoped>
 .content {
 	height: 100%;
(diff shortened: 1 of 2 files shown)


nickvergessen (approved):
Feels hacky, but looks nice and does the job.

Approving but lets wait for a frontender to confirm

PVince81:
a less hacky alternative would be to adjust nextcloud-dialogs to also pass the "offset" attribute through to Toastify, but then we also need to set it in every call to showMessage, maybe setup a wrapper function in Talk, or introduce global options for that.

PVince81:
tickets raised in nextcloud-dialogs:
- forward offset option: https://github.com/nextcloud/nextcloud-dialogs/issues/381
- implement global options: https://github.com/nextcloud/nextcloud-dialogs/issues/380

marcoambrosini (approved):

nickvergessen:
Bildschirmfoto von 2021-06-21 14-55-13

@PVince81 need to adjust something or unluckily revert this.

Issue #5764, opened 2021-06-14
When leaving a call the chat requests are canceled


nickvergessen:
## Steps to reproduce
1. Create a conversation
2. Start a call
3. Leave the call

### Expected behaviour
![Bildschirmfoto von 2021-06-14 13-14-23](https://user-images.githubusercontent.com/213943/121883827-81758480-cd12-11eb-9981-636df7f1924c.png)

### Actual behaviour

![Bildschirmfoto von 2021-06-14 13-14-12](https://user-images.githubusercontent.com/213943/121883875-8df9dd00-cd12-11eb-8ad5-a66e5d8d830c.png)

It seems both, the chat request in the right sidebar as well as the one in the main view get canceled, while th emain one is required to persist:
![Bildschirmfoto von 2021-06-14 12-58-05](https://user-images.githubusercontent.com/213943/121883904-95b98180-cd12-11eb-9bcd-953b4a195a0e.png)

PVince81:
I think I know: when I moved the cancellation to the store, I'm reusing the same attribute for any request.
This is similar to https://github.com/nextcloud/spreed/pull/5699.

Might need to map the attribute to a specific request.

On the other hand, here it's a bit special because we have two component instances based on the same list, so in theory only a single request is needed and no cancellation. It's the MessagesList that triggers cancellation when being destroyed, maybe that needs to happen elsewhere.

PVince81:
Ok, I tried splitting the cancel functions by token and it doesn't help. Because this is not reason.

The first cancelled:
- call is triggered from the "chatIdentifier" watcher, immediately when creating the view, and leads to `handleStartGettingMessagesPreCondition`
- cancellation also triggered from there

The second cancelled:
- call comes from the "setTimeout()" in `getNewMessages()`
- cancellation triggered by MessagesList's `beforeDestroy`

So the old view cancels while being destroyed, and the new MessagesList cancels again while setting itself up.
But it seems that the main view gets created before the sidebar vue gets destroyed, so there's an overlap.

Now if the `chatIdentifier` didn't really change, I don't see why an initial fetch would need to cancel an old call. If there was one pending it could just reuse it. To detect an initial `chatIdentifier` one can check that there's no `oldValue` set.

I'll give this a try.

PVince81:
seems this is more complex than I thought.

the expectation seems to be that we want to cancel and refetch whenever one of "token", "isParticipant" and "isInLobby" changes, but ideally not cancel and refetch the same thing

[...]

PR #5795, opened 2021-06-18
Fix concurrency when cancelling lookForNewMessages

PVince81:
When switching between regular mode and call mode, the two MessagesList
view instances are concurrently disturbing each other's
lookForNewMessages call.

To fix this, we now introduce a requestId for those requests to which we
pass in the chatIdentifier, extended by a generated view id. This brings
back the old behavior for now.

In the future we should look into keeping the polling request and
sharing it between the two Vues, which requires some refactoring of the
polling logic and timeouts.

Fixes https://github.com/nextcloud/spreed/issues/5764

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -79,6 +79,7 @@ import debounce from 'debounce'
 import { EventBus } from '../../services/EventBus'
 import LoadingPlaceholder from '../LoadingPlaceholder'
 import ChevronDown from 'vue-material-design-icons/ChevronDown'
+import { uniqueId } from 'lodash'
 
 export default {
 	name: 'MessagesList',
@@ -114,6 +115,8 @@ export default {
 
 	data() {
 		return {
+			viewId: null,
+
 			/**
 			 * When scrolling to the top of the div .scroller we start loading previous
 			 * messages. This boolean allows us to show/hide the loader.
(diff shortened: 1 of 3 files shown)


nickvergessen (approved):
Works now, but the console.log is still there twice:

> The request has been canceled Object { message: "canceled" }

src/components/MessagesList/MessagesList.vue
```diff
@@ -259,13 +262,17 @@ export default {
...
+				if (oldValue) {
+					this.$store.dispatch('cancelLookForNewMessages', { requestId: oldValue })
+				}
 				this.handleStartGettingMessagesPreconditions()
 			},
 		},
 	},
 
 	mounted() {
+		this.viewId = uniqueId('messagesList')
```
nickvergessen, line 275: This breaks uploading files, creating folder and deleting either in the files app

src/components/MessagesList/MessagesList.vue
```diff
@@ -79,6 +79,7 @@ import debounce from 'debounce'
 import { EventBus } from '../../services/EventBus'
 import LoadingPlaceholder from '../LoadingPlaceholder'
 import ChevronDown from 'vue-material-design-icons/ChevronDown'
+import { uniqueId } from 'lodash'
```
nickvergessen, line 82: This breaks uploading files, creating folder and deleting either in the files app

Issue #5287, opened 2021-03-02
Forward a chat message to another chat


jakobroehrl:
It would be nice if the user could select a message and forward it to another chat

nickvergessen:
Yeah we can use the talk message actions option and the conversation picking on the deck card action that was introduced for 11.1 and just copy the text and send it as a message to the new conversation.

jakobroehrl:
@nickvergessen 
Could you sketch out a litte bit how to solve this?

nickvergessen:
When clicking on the action in the actions menu, copy the text and sent it as a message to the other chat.

PR #5800, opened 2021-06-18
forward message

jakobroehrl:
Fixes: #5287

image
image

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -204,19 +204,13 @@ the main body of the message as well as a quote.
 				<span>{{ t('spreed', 'Unread messages') }}</span>
 			</div>
 		</div>
-
-		<Modal
+		<RoomSelector
 			v-if="modal"
 			container="#content-vue"
-			@close="modal=false">
-			<div>
-				<RoomSelector
-					:show-postable-only="true"
-					:dialog-title="setDialogTitle"
-					@select="setSelectedRoom"
-					@close="modal=false" />
-			</div>
-		</Modal>
+			:show-postable-only="true"
+			:dialog-title="setDialogTitle"
+			@select="setSelectedRoom"
+			@close="modal=false" />
 	</li>
 </template>
 
@@ -254,7 +248,6 @@ import {
 import { generateUrl } from '@nextcloud/router'
 import Location from './MessagePart/Location'
 import Contact from './MessagePart/Contact.vue'
-import Modal from '@nextcloud/vue/dist/Components/Modal'
 import RoomSelector from '../../../../views/RoomSelector.vue'
 
 export default {
@@ -278,7 +271,6 @@ export default {
 		Reload,
 		Share,
 		ActionSeparator,
-		Modal,
 		RoomSelector,
 	},
 
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -211,6 +211,20 @@ the main body of the message as well as a quote.
 			:dialog-title="setDialogTitle"
 			@select="setSelectedRoom"
 			@close="modal=false" />
+
+		<Modal v-if="forwared">
+			<EmptyContent icon="icon-checkmark" class="forwardModal">
+				{{ t('spreed', '"{msg}" was forwarded to "{user}"', { msg: message, user: selectedRoom }, undefined, { sanitize: false, escape: false }) }}
+				<template #desc>
+					<button class="primary" @click="openConversation(room)">
+						{{ t('spreed', 'Go to conversation') }}
+					</button>
+					<button @click="forwared=false">
+						{{ t('spreed', 'Close') }}
+					</button>
+				</template>
+			</EmptyContent>
+		</Modal>
 	</li>
 </template>
 
@@ -249,6 +263,8 @@ import { generateUrl } from '@nextcloud/router'
 import Location from './MessagePart/Location'
 import Contact from './MessagePart/Contact.vue'
 import RoomSelector from '../../../../views/RoomSelector.vue'
+import EmptyContent from '@nextcloud/vue/dist/Components/EmptyContent'
+import Modal from '@nextcloud/vue/dist/Components/Modal'
 
 export default {
 	name: 'Message',
@@ -272,6 +288,8 @@ export default {
 		Share,
 		ActionSeparator,
 		RoomSelector,
+		EmptyContent,
+		Modal,
 	},
 
 	mixins: [
@@ -420,6 +438,7 @@ export default {
 			seen: false,
 			modal: false,
 			selectedRoom: null,
+			forwared: false,
 		}
 	},
 
@@ -805,16 +824,19 @@ export default {
 			this.modal = true
 		},
 
-		setSelectedRoom(room) {
+		async setSelectedRoom(room) {
 			this.selectedRoom = room
 			this.modal = false
-			this.$store.dispatch('forwardMessage', {
+			await this.$store.dispatch('forwardMessage', {
 				token: room,
 				message: this.messageObject,
 			})
+			this.forwared = true
 
-			// Display a third dialog and push the new route only if the user requires it
-			// this.$router.push({ name: 'conversation', params: { token: room } }).catch(err => console.debug(`Error while pushing the new conversation's route: ${err}`))
+		},
+		openConversation() {
+			this.$router.push({ name: 'conversation', params: { token: this.selectedRoom } }).catch(err => console.debug(`Error while pushing the new conversation's route: ${err}`))
+			this.forwared = false
 		},
 	},
 }
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -783,7 +783,7 @@ export default {
 		},
 
 		async handleMarkAsUnread() {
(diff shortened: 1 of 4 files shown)


src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -188,6 +194,13 @@ the main body of the message as well as a quote.
 				<span>{{ t('spreed', 'Unread messages') }}</span>
 			</div>
 		</div>
+
+		<Modal
+			v-if="modal"
+			container="#content-vue"
+			@close="modal=false">
+			<div><LeftSidebar /></div>
```
nickvergessen, line 202: This sounds wrong :D 
Can't we reuse the src/views/RoomSelector.vue

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -157,6 +157,12 @@ the main body of the message as well as a quote.
 								:href="linkToFile">
 								{{ t('spreed', 'Go to file') }}
 							</ActionLink>
+							<ActionLink
+								icon="icon-share"
```
nickvergessen, line 161: Should use Material design icon `Share` instead

jakobroehrl: How to use it within an ActionLink componente?

nickvergessen: Well first don't use ActionLink when you want a buttom instead of a link, and then the docs have some samples:
https://nextcloud-vue-components.netlify.app/#/Components/Actions?id=actionbutton

```vue
		<ActionButton>
			<HandLeft
				slot="icon"
				:size="24"
				decorative
				title="" />
			Raise left hand
		</ActionButton>
```

jakobroehrl:
Why this @select is not working?
https://github.com/nextcloud/spreed/blob/1e8abdc890b0960297bc0cd4c1a79761d3ecc15f/src/components/MessagesList/MessagesGroup/Message/Message.vue#L209

I get no log here:
https://github.com/nextcloud/spreed/blob/1e8abdc890b0960297bc0cd4c1a79761d3ecc15f/src/components/MessagesList/MessagesGroup/Message/Message.vue#L802

But this code line is reached:
https://github.com/nextcloud/spreed/blob/1e8abdc890b0960297bc0cd4c1a79761d3ecc15f/src/views/RoomSelector.vue#L132

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -188,6 +198,19 @@ the main body of the message as well as a quote.
...
+
+		<Modal
+			v-if="modal"
+			container="#content-vue"
+			@close="modal=false">
+			<div>
+				<RoomSelector
+					:show-postable-only="true"
+					:dialog-title="setDialogTitle"
+					@select="setSelectedRoom(room)"
```
marcoambrosini, line 210: This doesn't work because you're calling the function here, you should remove the `(room)` part and just accept the room as a parameter where you define the event handler

jakobroehrl: @marcoambrosini  Thanks for your answer. But the button has no effect:
image

Tried @select="setSelectedRoom()" and @select="setSelectedRoom"

I still get no log here:
https://github.com/nextcloud/spreed/blob/1e8abdc890b0960297bc0cd4c1a79761d3ecc15f/src/components/MessagesList/MessagesGroup/Message/Message.vue#L801-L805

jakobroehrl:
This is working so far.
I think the user would expect to change the room after forwarding. 
And file forwards doesn't work.

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -188,6 +198,32 @@ the main body of the message as well as a quote.
...
+			</EmptyContent>
+			<div class="forwarded-confirmation__navigation">
+				<button @click="showForwardedConfirmation=false">
+					{{ t('spreed', 'Dismiss') }}
+				</button>
+				<button class="primary" @click="openConversation(room)">
+					{{ t('spreed', 'Go to conversation') }}
+				</button>
+			</div>
+		</Modal>
```
nickvergessen, line 226: I would accept this if you ask for confirmation. But otherwise I guess showSuccess should be enough?

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -627,6 +674,14 @@ export default {
 				apiVersion: 'v3',
 			}
 		},
+
+		dialogTitle() {
+			return t('spreed', 'Forward message')
+		},
+
+		dialogSubtitle() {
+			return t('spreed', 'Choose a conversation to which forward the selected message.')
```
nickvergessen, line 683:
```suggestion
			return t('spreed', 'Choose a conversation to forward the selected message to.')
```

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -749,7 +804,7 @@ export default {
 		},
 
 		async handleMarkAsUnread() {
-			// update in backend + visually
+			// update in backend showError(t('spreed', 'No permission to post messages in this conversation'))+ visually
```
nickvergessen, line 807: ??

marcoambrosini:
@PVince81 In the last commit I'm moving the logic from the `Message` component to the `Forwarder` component. I'm introducing a bug that provokes an error upon clicking select in the roomselector but I'm having trouble finding out where it comes from.

marcoambrosini:
- [ ] link directly to the message when pushing new route

marcoambrosini:
>  link directly to the message when pushing new route
@PVince81 this seems tricky.. I can not come up with and easy way to make the forwarder aware of the ID of the newly posted message in the conversation

PVince81:
> > link directly to the message when pushing new route
> 
> @PVince81 this seems tricky.. I can not come up with and easy way to make the forwarder aware of the ID of the newly posted message in the conversation

did you try returning the id from the 'forwardMessage' action ?
the `postNewMessage` call returns a response from the server which usually contains at least the message id IIRC

marcoambrosini (approved):

marcoambrosini:
@PVince81 please take a look when you have some time

marcoambrosini (changes requested):
Blocking so nobody gets tempted to merge before @PVince81  takes a look

src/components/MessagesList/MessagesGroup/Message/MessagePart/Forwarder.vue
```diff
@@ -0,0 +1,153 @@
...
+			return t('spreed', 'Choose a conversation to forward the selected message.')
+		},
+
+		selectedConversationName() {
+			return this.$store.getters?.conversation(this.selectedConversationToken).name
+		},
+	},
+
+	methods: {
+		async setselectedConversationToken(token) {
```
PVince81, line 104: typo in capitalization: s/setselected/setSelected/

src/components/MessagesList/MessagesGroup/Message/MessagePart/Forwarder.vue
```diff
@@ -0,0 +1,153 @@
...
+			this.selectedConversationToken = token
+			const response = await this.$store.dispatch('forwardMessage', {
+				token,
+				message: this.messageObject,
+			})
+			this.showForwardedConfirmation = true
+			this.forwardedMessageID = response.data.ocs.data.id
```
PVince81, line 111: what happens if the call fails ? might be good to catch the exception here

src/components/MessagesList/MessagesGroup/Message/MessagePart/Forwarder.vue
```diff
@@ -0,0 +1,153 @@
...
+		},
+
+		openConversation() {
+
+			this.$router.push({
+				name: 'conversation',
+				hash: `#message_${this.forwardedMessageID}`,
+				params: {
+					token: `${this.selectedConversationToken}`,
+			 },
```
PVince81, line 121: weird alignment

src/store/messagesStore.js
```diff
@@ -35,6 +35,7 @@ import { showError } from '@nextcloud/dialogs'
 import {
 	ATTENDEE,
 } from '../constants'
+import { cloneDeep } from 'lodash'
```
PVince81, line 38: please import with `import cloneDeep from 'lodash/cloneDeep'` of whatever works.

last time we used this kind of import there were weird unexplained side effects in the files app. Ref: https://github.com/nextcloud/spreed/issues/5858

src/store/messagesStore.js
```diff
@@ -875,6 +876,25 @@ const actions = {
...
+	 */
+	async forwardMessage(context, { token, message }) {
+		const messageToBeForwarded = cloneDeep(message)
+		messageToBeForwarded.token = token
+		try {
+			const response = await postNewMessage(messageToBeForwarded)
+			return response
+		} catch (error) {
+			console.error('Error while forwarding message')
+			showError(t('spreed', 'Error while forwarding message'))
```
PVince81, line 895: Let's not catch exceptions inside the store but outside, this way we don't have UI logic in here like showError.
UI logic belongs in the Vue that calls it.

PVince81 (commented):
See comments.
But it seems you already merged :-(

Gibran2075:
Hii, I have a problem with forwarding messages, the problem is with the files feature, in the files sidebar I can chat but if I forward it I can't open go to conversation, how do I fix it? i am now using spreed 22.2.3, please help me
Screenshot (452)

nickvergessen:
You reported this already in https://github.com/nextcloud/spreed/issues/11604

PR #5823, opened 2021-06-21
Set focus to message only when anchor is set

PVince81:
Don't set focus on message element when scrolling to the first unread
message as this conflicts with focusInput() from the message form.

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -779,7 +779,7 @@ export default {
 		 *
 		 * @param {string} messageId message id
 		 * @param {boolean} smooth true to smooth scroll, false to jump directly
-		 * @param {boolean} highlightAnimation true to highlight the focussed message
+		 * @param {boolean} highlightAnimation true to highlight and set focus to the message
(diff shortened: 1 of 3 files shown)


nickvergessen:
Doesn't seem to work here. For whatever recent reason the chat input is disabled for a second or two and so the focusing doesnt work

PVince81:
> Doesn't seem to work here. For whatever recent reason the chat input is disabled for a second or two and so the focusing doesnt work

Not sure what is disabling the field now.

The current/old code for focussing the field is a bit naive and is applied on mount regardless of state, and hopes that after a delay everything is fine.

PVince81:
Got it: we have new logic that disables the field with "Joining conversation..." here: https://github.com/nextcloud/spreed/blob/master/src/components/NewMessageForm/NewMessageForm.vue#L210

Now we could wait for the joining to finish and then move the focus, but depending how long it takes it might take the user out of context if they were typing elsewhere. I guess it's not too bad though, at least to fix the regression.

I'll try the latter approach.

PVince81:
@nickvergessen I've pushed a fix, please retry

nickvergessen (approved):

PR #5828, opened 2021-06-21
Populate user status for one to one conversations via the backend call already

nickvergessen:
### Steps
1. Open Talk
2. Create one-to-one conversations with X users
3. Reload talk front page

### Before
X requests for user statuses (most with 404)

### After
User status is preloaded via conversations API and no additional requests are needed

### DIFF
--- a/docs/conversation.md
+++ b/docs/conversation.md
@@ -9,6 +9,12 @@
 
 * Method: `GET`
 * Endpoint: `/room`
+* Data:
+
+    field | type | Description
+    ---|---|---
+    `noStatusUpdate` | int | Whether the "online" user status of the current user should be "kept-alive" (`1`) or not (`0`) (defaults to `0`)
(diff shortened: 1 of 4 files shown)


docs/conversation.md
```diff
@@ -9,6 +9,12 @@
 
 * Method: `GET`
 * Endpoint: `/room`
+* Data:
+
+    field | type | Description
+    ---|---|---
+    `noStatusUpdate` | int | Whether the "online" user status of the current user should be "kept-alive" or not (default 0)
```
blizzz, line 16: what are the possible values here? Apart from  0, but what does it stand for?

nickvergessen: 1, but yeah will add it.

blizzz (approved):
Looks good, one remark

PVince81 (approved):
:+1: code and approach looks good

nickvergessen:
Fixed the doc issue and did the fixup

PR #5831, opened 2021-06-21
Fix asterix conflict with notifications

nickvergessen:
Reverse of https://github.com/nextcloud/notifications/pull/1028

### Steps
1. Reload page on talk and wait until it finished loading
2. Open a new tab so the page is in the background
3. Trigger a notification: `occ notification:generate`
4. Write a message in a channel where this user is a member of

### Before
`* * …`

### After
`* …`

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -391,11 +391,14 @@ export default {
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


PVince81 (approved):
:+1:

Issue #5826, opened 2021-06-21
User status in conversation title is cut by a few pixels


ChristophWurst:
## Steps to reproduce
1. Start a one to one conversation
2. Make sure the other user is online

### Expected behaviour

User status icon renders fully

### Actual behaviour

![Bildschirmfoto von 2021-06-21 16-11-51](https://user-images.githubusercontent.com/1374172/122776608-89ea3400-d2ab-11eb-99b7-2c31c0eefc40.png)

## Talk app

**Talk app version:** c.nc.com

PVince81:
confirmed

PVince81:
seems the top bar is not designed to fit:
<img width="169" alt="image" src="https://user-images.githubusercontent.com/277525/122782585-22cf7e00-d2b1-11eb-8568-f87f5f47ef00.png">

multiple possibilities:
- change overflow from "clip" to "visible" to make it appear
- or reduce the avatar size to 42px instead of 44px (but then it's less than the standard we use in other places)
- or increase top bar size
- ...

nickvergessen:
We don't reduce the avatar size

PVince81:
PR here https://github.com/nextcloud/spreed/pull/5836 which moves the ConversationIcon out of the truncating anchor tag.

Issue #5790, opened 2021-06-17
Clicking on the avatar in the conversation list tries to leave the call


nickvergessen:
## Steps to reproduce
1. Have a one-to-one conversation
2. Be in anothern conversation or even call
3. Click on the avatar of a one-to-one conversation in the conversation list to check

### Expected behaviour
You see the profile menu

### Actual behaviour
* If in call: Attempted to leave the call
* If not in call: Conversation is switched
![Bildschirmfoto von 2021-06-17 11-21-41](https://user-images.githubusercontent.com/213943/122369750-c2fa6f80-cf5e-11eb-91cc-04f87ced4d2b.png)

PVince81:
I've seen a related but different issue: if you click on the avatar menu in the top bar, it will click on the parent "a" and display the settings dialog.

I think both can be prevented by having the Avatar component stop the events with `.prevent`.

PVince81:
Seems to be another trouble with v-tooltip as that's where the click handler is happening.
The click handler from Avatar is never called apparently.

I've checked the latest version of v-tooltip and it also doesn't call `evt.preventDefault()` or something.
So probably this is not fixable on nextcloud-vue side...

which means that consumers of Avatar need to make sure that if they wrap them into an `a` element or something, they need to detect whether the click landed on the avatar or outside before proceeding...

PVince81:
another idea: disable the menu in those places

PVince81:
I have a fix for the top bar case here: https://github.com/nextcloud/spreed/pull/5836

the fix might not be suitable for the left sidebar because we do want the ConversationIcon to be clickable when switching conversations, but only when it's not one to one, otherwise we'd want the avatar menu.

PVince81:
PR here for disabling the menu for now: https://github.com/nextcloud/spreed/pull/5837

PR #5836, opened 2021-06-22
Move conversation icon out of conversation header

PVince81:
Move the conversation icon outside of the conversation header so that it
doesn't get cut.

Fixes issue where the status icon was cut.
Fixes issue where clicking the avatar would open the settings dialog
instead of the avatar menu. (mentioned in https://github.com/nextcloud/spreed/issues/5790#issuecomment-865280442)

Fixes https://github.com/nextcloud/spreed/issues/5826

Tests:
- [x] description is still truncated properly, no layout issue
- [x] avatar menu working again
- [x] status icon is not cut
- [x] hamburger menu working correctly in narrow window or wide window

### DIFF
--- a/src/components/TopBar/TopBar.vue
+++ b/src/components/TopBar/TopBar.vue
@@ -21,14 +21,15 @@
 
 <template>
 	<div class="top-bar" :class="{ 'in-call': isInCall }">
+		<ConversationIcon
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
Makes sense, didn't test

PR #5839, opened 2021-06-22
Fix scrolling to bottom in already read conversations

nickvergessen:
1. Have a one-to-one conversation with some messages
2. Read them
3. Go to another conversation
4. Come back
5. Reload the page

### Before
You are in the middle of the chat

### After
You are at the bottom of the chat


PVince81 (approved):
:+1: as discussed

Issue #5848, opened 2021-06-22
Voice record as guest displays prompt for basic auth


PVince81:
## Steps to reproduce
1. Create a conversation
2. Allow guests
3. Log out
4. Open conversation
5. Click the microphone to create a recording
6. Post the recording after listening to its preview

### Expected behaviour
Recording is posted

### Actual behaviour
Basic auth dialog
<img width="565" alt="image" src="https://user-images.githubusercontent.com/277525/122956030-f2571500-d380-11eb-8d9d-d8bb7ff6f833.png">

We need to use the "public.php" endpoint instead, might be possible to get it through some of the OC APIs.

## Talk app

**Talk app version:** 12.0.0 git master 40ec034568336675c9d3230ee2df6afb26a0ea87

PVince81:
now thinking, we should probably disable this button for guests as guests can't upload attachments anyway.

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/5851

PR #5851, opened 2021-06-22
Remove voice recording for guests

PVince81:
Guests currently have no way of uploading files, pictures, so voice
recordings fall under the same category.

Fixes https://github.com/nextcloud/spreed/issues/5848

### DIFF
--- a/src/components/NewMessageForm/NewMessageForm.vue
+++ b/src/components/NewMessageForm/NewMessageForm.vue
@@ -105,12 +105,12 @@
 				</div>
 
 				<AudioRecorder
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):

nickvergessen:
Actually the following button to send the message should be `v-else` instead of `v-if="!hasText"`

nickvergessen (changes requested):
See comment

nickvergessen (approved):

nickvergessen:
Stylelint is a temp failure

PR #5437, opened 2021-03-29
Use ListItem component

marcoambrosini:
This removes `AppContentListItem` and adds the upstreamed `ListItem` component.
requres https://github.com/nextcloud/nextcloud-vue/pull/616

Signed-off-by: Marco Ambrosini <marcoambrosini@pm.me>

PR #5855, opened 2021-06-23
Show the conversation subtitle on hover

ChristophWurst:
# Before

Bildschirmfoto vom 2021-06-23 09-57-23

# After

Bildschirmfoto vom 2021-06-23 09-56-40

### DIFF
--- a/src/components/LeftSidebar/ConversationsList/AppContentListItem/AppContentListItem.vue
+++ b/src/components/LeftSidebar/ConversationsList/AppContentListItem/AppContentListItem.vue
@@ -44,7 +44,7 @@
 					</span>
 				</div>
 				<div class="acli__content__line-two">
-					<span class="acli__content__line-two__subtitle">
+					<span class="acli__content__line-two__subtitle" :title="subtitleText">
 						<slot name="subtitle" />
 					</span>
 					<span class="acli__content__line-two__counter">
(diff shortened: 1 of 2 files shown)


src/components/LeftSidebar/ConversationsList/AppContentListItem/AppContentListItem.vue
```diff
@@ -112,6 +112,11 @@ export default {
 			type: String,
 			default: '',
 		},
+
+		subtitleText: {
```
ChristophWurst, line 116: ideally we could get the text from the slot contents but this turned out to be a non-trivial problem especially with Vue 3

nickvergessen:
I think we need to limit is somehow.

Bildschirmfoto von 2021-06-23 10-29-16

nickvergessen:
Conflicts with #5437

ChristophWurst:
> I think we need to limit is somehow.

Looks good to me actually :)

marcoambrosini (commented):
I think that shomething like this should be implemented on the listitem component in the nextcloud vue library

skjnldsv:
> I think we need to limit is somehow.

Make it scroll :D

ChristophWurst:
Guess this one won't make it.

PR #5860, opened 2021-06-23
Fix screenshare icon not crossing out when a screen is shared

### LINKED ISSUE #5859: Screenshare icon does not indicate/change at all when screensharing
## Steps to reproduce
1. Start a call
2. Start a screenshare
3. Forgot you were screensharing cause it’s not obvious

### Expected behaviour
The screenshare icon communicates the state, just like audio and video on/off do. (This is also something which was the case before, although the strikethrough display might not have been the best default icon. :)

### Actual behaviour
The icon stays the exact same, not even the opacity changes.

As an example, here’s how it looks like in Discord, very present and obvious:
image

nickvergessen:
Fix #5859

### DIFF
--- a/src/components/CallView/shared/LocalMediaControls.vue
+++ b/src/components/CallView/shared/LocalMediaControls.vue
@@ -81,15 +81,15 @@
 				@update:open="screenSharingMenuOpen = true"
 				@update:close="screenSharingMenuOpen = false">
 				<!-- Actions button icon -->
-				<Monitor
+				<MonitorOff
 					v-if="model.attributes.localScreen"
 					slot="icon"
 					:size="24"
 					title=""
 					fill-color="#ffffff"
 					decorative />
-				<MonitorOff
-					v-if="!model.attributes.localScreen"
+				<Monitor
+					v-else
 					slot="icon"
 					:size="24"
 					title=""
--- a/src/components/CallView/shared/LocalMediaControls.vue
+++ b/src/components/CallView/shared/LocalMediaControls.vue
@@ -81,14 +81,14 @@
 				@update:open="screenSharingMenuOpen = true"
 				@update:close="screenSharingMenuOpen = false">
 				<!-- Actions button icon -->
-				<MonitorOff
+				<CancelPresentation
 					v-if="model.attributes.localScreen"
 					slot="icon"
 					:size="24"
 					title=""
 					fill-color="#ffffff"
 					decorative />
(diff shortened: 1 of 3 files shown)


src/components/CallView/shared/LocalMediaControls.vue
```diff
@@ -81,15 +81,15 @@
 				@update:open="screenSharingMenuOpen = true"
 				@update:close="screenSharingMenuOpen = false">
 				<!-- Actions button icon -->
-				<Monitor
+				<MonitorOff
```
nickvergessen, line 84: Ah, now I remember why this was changed.

With this patch when:
* audio is on => icon
* audio is off => icon crossed
* video is on => icon
* video is off => icon crossed
* screenshare is on => icon crossed 😕  *nickvergessen hurt itself in confussion*
* screenshare is off => icon 😕  *nickvergessen hurt itself in confussion*

So for screensharing the icon is invers to the mic/video icon and it feels odd if they are next to each others like that.
Maybe inverting the color of the background and icon is a better state representation?

PVince81: or invert screenshare on vs off to align with the others ?

not sure what @nextcloud/designers-talk think then...

nickvergessen: This is currently the case but not when your screen is not shared. We wouod have to use a crossed out icon for that and it feels weird...

jancborchardt: So yeah, for audio and video the icon reflects the current state. For screensharing the logic is inverted, but I’d not necessarily say that’s a bad thing.

Here’s possibly better icons, also from Material design:
- "Present to all" https://fonts.google.com/icons?selected=Material+Icons:present_to_all
- "Cancel presentation" https://fonts.google.com/icons?selected=Material+Icons:cancel_presentation

nickvergessen: Done

PVince81 (approved):
:+1: looks alright

sadly it seems the new icons are not in the latest vue-material-design-icons either

PVince81:
image

image

@jancborchardt ^

nickvergessen:
> sadly it seems the new icons are not in the latest vue-material-design-icons either

Yeah skj dived a bit into it, the lib uses another lib which wasn't updatd since january

PR #5863, opened 2021-06-23
Only scroll to focussed message once

PVince81:
We should scroll as soon as we have the old messages / context loaded.
For newer messages, don't scroll up again to the read marker in case
there is one.

It happened to me on c.nc.com that even though I scrolled down after seeing the read marker, the chat scrolled up again to focus on the read marker. I suspect that it's related to the call I've removed in this PR.

@nickvergessen this is what we discussed yesterday and weren't sure if we wanted to keep. Let's remove it for now.

I'll retest:

- [x] initial scroll to read marker still works
- [x] initial scroll to search result still works
- [x] initial scroll to bottom works when no unread messages

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -459,10 +459,7 @@ export default {
 					})
 
 					// get history + new messages
-					await this.getMessages(true)
-				} else {
-					// get only new messages
-					await this.getMessages(false)
+					await this.getOldMessages(true)
 				}
 
 				// focus on next tick to make sure the DOM elements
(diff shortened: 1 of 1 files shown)


nickvergessen (changes requested):
This is in case of a room change. It breaks for me when switching between conversations.
But it is correct when I focus another tab and go back
Peek 2021-06-24 08-22

PVince81:
Hmmm, so maybe we need to add this in yet another code path. My guest is that it seemed to work before just by chance since we always call lookForNewMessages after switching conversations. I guess we probably don't always call getOldMessages when switching and keep them cached ? I'll recheck this

nickvergessen:
> I guess we probably don't always call getOldMessages when switching and keep them cached ? I'll recheck this

yeah why should we, we still have them all loaded

PVince81:
> > I guess we probably don't always call getOldMessages when switching and keep them cached ? I'll recheck this
> 
> yeah why should we, we still have them all loaded

alright, that was my misunderstanding but now from the code I see it's obvious.

I've fixed the focus logic to always be:
1) get old messages (if necessary)
2) set focus
3) poll new messages

I've adjusted the function namings to avoid future misunderstandings.

nickvergessen (approved):

Issue #5870, opened 2021-06-25
Reply quote glitches + shifted emoji button when replying


PVince81:
## Steps to reproduce
1. Send a message
2. Reply to the message

### Expected behaviour
Nice looking UI

### Actual behaviour
Emoji button shifts up.
Reply quote box is outside the input field now ?? Was this a new design choice or a bug ?
<img width="545" alt="image" src="https://user-images.githubusercontent.com/277525/123381724-623bea00-d591-11eb-8593-b978387e7583.png">

## Talk app

**Talk app version:** 12.0.0 git master af58b3100dcde68c9c4affab371a99aec5a78a9d

PVince81:
PR for the emoji button: https://github.com/nextcloud/spreed/pull/5871

for the quote box position, will need input from @marcoambrosini whether this design change was intended. The quote box used to be inside the input frame.

nickvergessen:
> The quote box used to be inside the input frame.

No, or well maybe technically but not visually: (screenshot from 21)
![Bildschirmfoto von 2021-06-25 10-55-35](https://user-images.githubusercontent.com/213943/123398518-edbe7680-d5a3-11eb-879e-87ed06d1ce9e.png)

PR #5871, opened 2021-06-25
Fix emoji picker vertical position

PVince81:
Fix CSS syntax error that prevented the emoji picker to stick to the
bottom.

Before:
image

After:
image

Please note that if you type multiple lines it will stick at the bottom. Since the "bottom" rule was already there I'll assume that this was already by design, just not activately properly:
image

Fixes https://github.com/nextcloud/spreed/issues/5870


nickvergessen:
PRs against master are now "23"

nickvergessen (approved):

Issue #5872, opened 2021-06-25
Unread messages marker always at bottom of viewport when initially switching


PVince81:
## Steps to reproduce
1. Open talk
2. Open a conversation that contains unread messages
3. Wait

### Expected behaviour
Scrolls down automatically to the unread messages marker, and the marker appears mostly at the top of the viewport.

### Actual behaviour
Unread marker appears at the bottom of the viewport and the user must scroll down.

## Talk app

**Talk app version:** 12.0.0 af58b3100dcde68c9c4affab371a99aec5a78a9d

## Analysis

This issue is revealed by the change from yesterday: https://github.com/nextcloud/spreed/pull/5863

The order is now as follows:

1. get old messages
2. scroll to read message marker
3. get new messages / look for new messages long poll

It appears that the "old" messages are not from the last message ever in the conversation but they start at the read marker: https://github.com/nextcloud/spreed/blob/master/src/components/MessagesList/MessagesList.vue#L454

This logic might have worked in the past when we did not have the marker. One thing to try is to change the "first known message" to be the `lastMessage.id` from the conversation. This way `getOldMessages` would simply load "all" the messages before polling for new ones.

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/5873

PR #5873, opened 2021-06-25
Fix initial scroll of unread messages marker

PVince81:
For the initial scroll to work, the first call to getOldMessages must
fetch messages beyond the read marker itself. Otherwise the marker will
appear at the bottom of the screen as there aren't any new messages yet
until `lookForNewMessages` runs.

Fixes https://github.com/nextcloud/spreed/issues/5872

Tests with conversation that has a lot of new unread messages:

- [x] marker at top of viewport when switching to conversation after an **initial page load**
- [x] marker at top of viewport when switching back to a conversation after having been in the conversation before

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -447,6 +447,7 @@ export default {
 					token: this.token,
 					id: this.conversation.lastReadMessage,
 				})
+
 				if (this.$store.getters.getFirstKnownMessageId(this.token) === null) {
 					// first time load, initialize important properties
 					this.$store.dispatch('setFirstKnownMessageId', {
@@ -458,18 +459,31 @@ export default {
 						id: this.conversation.lastReadMessage,
 					})
 
-					// get history + new messages
+					// get history before last read message
 					await this.getOldMessages(true)
+					// at this stage, the read marker will appear at the bottom of the view port since
+					// we haven't fetched the messages that come after it yet
+					// TODO: should we still show a spinner at this stage ?
 				}
 
-				// focus on next tick to make sure the DOM elements
-				// for known messages are already rendered
-				this.$nextTick(() => {
-					this.scrollToFocussedMessage()
-				})
+				let hasScrolled = false
+				// if lookForNewMessages will long poll instead of returning existing messages,
+				// scroll right away to avoid delays
+				if (this.$store.getters.getLastKnownMessageId(this.token) === this.conversation.lastMessage.id) {
+					hasScrolled = true
+					await this.$nextTick(() => {
+						this.scrollToFocussedMessage()
+					})
+				}
 
 				// get new messages
 				await this.lookForNewMessages()
+
+				// don't scroll if lookForNewMessages was polling as we don't want
+				// to scroll back to the read marker after receiving new messages later
+				if (!hasScrolled) {
+					this.scrollToFocussedMessage()
+				}
 			} else {
 				this.$store.dispatch('cancelLookForNewMessages', { requestId: this.chatIdentifier })
 			}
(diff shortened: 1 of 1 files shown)


PVince81:
with a bit of luck this change will also make conversation switching feel even faster because now the new unread messages are loaded right away in the first call instead of waiting for the second one.
it also means one less server call: instead of "getOldMessages", "lookForNewMessages->returns directly", "lookForNewMessages->long poll" it becomes "getOldMessages", "lookForNewMessages->long poll"

nickvergessen:
PRs against master are now "23"

src/components/MessagesList/MessagesList.vue
```diff
@@ -451,14 +451,14 @@ export default {
 					// first time load, initialize important properties
 					this.$store.dispatch('setFirstKnownMessageId', {
 						token: this.token,
-						id: this.conversation.lastReadMessage,
+						id: this.conversation.lastMessage.id,
```
nickvergessen, line 454: this feels very wrong. the lastMessage does not have to do anything with the reading status?

So when there are "more" messages you don't start at the readmarker anymore?

PVince81: yes.

imagine a conversation that exists on the server:
- message A
- message B
- unread marker
- message C
- message D

Previously, when calling getOldMessages it would only retrieve message A and message B.
then lookForNewMessages would right away fetch message C and message D, as they already exist.

With the new behavior, getOldMessages will fetch all messages A to D, so lookForNewMessages can start polling right away. There's no point in having to wait for lookForNewMessages to fetch the "old" messages C and D.

Or did you have something else in mind ?
-

nickvergessen: imagine a conversation that exists on the server:

* message A
* message B
* unread marker
* message 1
* …
* message 300

getOldMessages only gets 200 messages. so you will get message 100-300 you are far of your readmarker and the first read message is not even loaded, so you don't know

PVince81: hmmm, should we then set firstKnown to be the unread marker but keep lastKnown to be the last message ? (or the other way around, not sure about the order)

nickvergessen: That will break everything.

the "known" are the loaded ones. if you set a gap in there it will not load those messages ever. If you make them overlap you will load the messages between twice.

PVince81:
temp test plan while still debugging, I've pushed a WIP with log statements.

1. Create two conversations "first" and "second" as Alice and invite Bob in both
2. Open converstion "second" as Alice
2. Open a second browser for Bob
3. As Bob, open "first"
4. Quickly type messages: "---", 1, 2, 3, 4, and count up to 20 or more, one number per message
5. As Alice while still on "second", refresh the page
6. As Alice, open "first": **this tests the initial load and scroll**
7. Wait for unread marker to appear above the "---" message: the viewport should be scrolled so that the marker is visible
8. As Alice, Switch to "second" again
9. As Bob, type another set of messages "---", 1, 2, 3, 4, ... 20
10. As Alice, wait for the conversation list to refresh and display the unread messages counter
11. As Alice, switch to "first" again: **this tests the scroll when we already had old messages in the store**
12. Wait for unread marker to appear above the "---" message: the viewport should be scrolled so that the marker is visible
13. Scroll down a tiny bit but not to the bottom
14. Create and switch to a new tab in Alice's browser to trigger the blur/focus event, wait for two seconds
15. Go back to the former tab and wait two seconds
16. Read marker should have moved down to the top of the viewport a few messages further down from the previous marker position: **this tests the read marker movement when switching away**
17. Scroll down to bottom
18. Go back to the empty second tab and wait two seconds
19. Go back to the former tab from Alice and wait two seconds
20. Read marker should have disappeared, the conversation is marked as read: **this tests read marker disappearance when scrolled to bottom**

src/components/MessagesList/MessagesList.vue
PVince81: I've reverted this back and have pushed an alternative approach

PVince81:
The new approach seems to work but has a tiny glitch: the initial conversation load has the read marker at the bottom and it stays there until `lookForNewMessages` finishes. I think we had that already in previous iterations and might not be that bad.
If bad we could either try to:
- add a loading spinner below
- or add a spinner on the whole viewport
- and/or smooth scroll to the new marker position after its update

nickvergessen (approved):

PR #5881, opened 2021-06-25
Add a debug command to generate an arbitrary number of chat message

nickvergessen:


PVince81 (approved):
:+1:

danxuliu (approved):
Tested and works :+1: 

I would prefer not to add debug code along with the normal code. But... this looks handy and adding it through a separate app or something like that would be probably overkill, so :+1:

PR #5889, opened 2021-06-28
🗑️ 💬 Clear chat history - Backend

### LINKED ISSUE #1243: Allow to clear chat history
> > ~I often used chat and there were a lot of messages. Now as soon as I open a chat, the application starts opening all messages from the very first. Opening this message history can take 20-30 seconds, during which I cannot write anything to the chat or even close the tab in the browser. Is it possible to hide the history of the chat as it is made in Facebook messenger?~
> 
> The chat is loaded newest to oldest as a temporary workaround in Talk 7.
The optimal solution is to start at the read marker as described in #1164 
Which will be covered by #1788

Since emptying all chat might still be a useful feature, eg in webinaries before opening the lobby, I will keep this issue open. - @nickvergessen

nickvergessen:
- [x] API to clear the history
- [x] Unshare all files
- [x] Document it
- [x] System message "A has removed the history of the conversation"
- [x] Integration test

Fix #1243

### DIFF
--- a/docs/chat.md
+++ b/docs/chat.md
@@ -153,6 +153,30 @@ See [OCP\RichObjectStrings\Definitions](https://github.com/nextcloud/server/blob
 
 * Response: [See official OCS Share API docs](https://docs.nextcloud.com/server/latest/developer_manual/client_apis/OCS/ocs-share-api.html?highlight=sharing#create-a-new-share)
 
+## Clear chat history
+
+* Required capability: `clear-history`
+* Method: `DELETE`
+* Endpoint: `/chat/{token}`
+
+* Response:
+    - Status code:
+        + `200 OK` - When deleting was successful
+        + `202 Accepted` - When deleting was successful but Matterbridge is enabled so the message was leaked to other services
+        + `403 Forbidden` When the user is not a moderator
+        + `404 Not Found` When the conversation could not be found for the participant
+
+    - Header:
+
+        field | type | Description
+        ---|---|---
+        `X-Chat-Last-Common-Read` | int | ID of the last message read by every user that has read privacy set to public. When the user themself has it set to private the value the header is not set (only available with `chat-read-status` capability)
+
+    - Data:
+        The full message array of the new system message "You cleared the history of the conversation", as defined in [Receive chat messages of a conversation](#receive-chat-messages-of-a-conversation)
+        When rendering this message the client should also remove all messages from any cache/storage of the device.
+
+
 ## Deleting a chat message
 
 * Required capability: `delete-messages`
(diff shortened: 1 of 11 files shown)


docs/chat.md
```diff
@@ -153,6 +153,30 @@ See [OCP\RichObjectStrings\Definitions](https://github.com/nextcloud/server/blob
...
+        + `200 OK` - When deleting was successful
+        + `202 Accepted` - When deleting was successful but Matterbridge is enabled so the message was leaked to other services
+        + `403 Forbidden` When the user is not a moderator
+        + `404 Not Found` When the conversation could not be found for the participant
+
+    - Header:
+
+        field | type | Description
+        ---|---|---
+        `X-Chat-Last-Common-Read` | int | ID of the last message read by every user that has read privacy set to public. When the user themself has it set to private the value the header is not set (only available with `chat-read-status` capability)
```
PVince81, line 173: I guess this will return the id of the system message that triggers the deletion ?

docs/chat.md
```diff
@@ -268,6 +292,7 @@ See [OCP\RichObjectStrings\Definitions](https://github.com/nextcloud/server/blob
 * `guest_moderator_promoted` - {actor} promoted {user} to moderator
 * `guest_moderator_demoted` - {actor} demoted {user} from moderator
 * `message_deleted` - Message deleted by {actor} (Should not be shown to the user)
+* `cleared_history` - {actor} cleared the history of the conversation
```
PVince81, line 295: this key isn't using the same pattern like the others

all the others are "$object_$verb" but here you used "$verb_$object".

</picky_mode>

lib/Controller/ChatController.php
```diff
@@ -592,7 +592,43 @@ public function deleteMessage(int $messageId): DataResponse {
 
 		$bridge = $this->matterbridgeManager->getBridgeOfRoom($this->room);
 
-		$response = new DataResponse($data, $bridge['enabled'] ? Http::STATUS_ACCEPTED: Http::STATUS_OK);
+		$response = new DataResponse($data, $bridge['enabled'] ? Http::STATUS_ACCEPTED : Http::STATUS_OK);
+		if ($this->participant->getAttendee()->getReadPrivacy() === Participant::PRIVACY_PUBLIC) {
```
PVince81, line 596: I'm guessing this is an additional fix not directly related to clearing the chat history ?

if yes you might want to make that a separate PR/commit so we can backport, if applicable

PVince81 (approved):
:+1: some minor comments

docs/chat.md
nickvergessen: Actually not, as the chat message is not "read" by everyone, so the value will be 0 basically all the time.

docs/chat.md
nickvergessen: makes sense

lib/Controller/ChatController.php
nickvergessen: Yeah we could/should. The header is documented, but the clients are forgiving
I will manually backport it together with the `deleteMessages` fix and the broken integration test

PVince81 (approved):
:+1:

Ivansss:
Capability `clear-history` is missing :/

PVince81:
> Capability `clear-history` is missing :/

added with https://github.com/nextcloud/spreed/pull/6023

PR #5896, opened 2021-06-29
Catch the correct exception in case of UniqueConstraintViolationExcep…

nickvergessen:
…tion


PVince81 (approved):
:+1:

PR #5905, opened 2021-07-01
Listen to token changes instead of conversation object

PVince81:
The converstion watcher is triggered even for attribute changes, so we
now use the token watcher instead for scrolling up so it only happens
when meaningful.

Note: the conversation display name watcher still needs to watch for "conversation" as the name might change while the token doesn't.

Fixes https://github.com/nextcloud/spreed/issues/5904


danxuliu (approved):
Tested and works :+1:

nickvergessen:
Backport in https://github.com/nextcloud/spreed/pull/5906

Issue #3151, opened 2020-03-25
Background blur / modify background of video


jancborchardt:
This is not something for soon as it’s a bunch of work and there’ other things to be worked on first. But just for reference.

--------------

From https://news.ycombinator.com/item?id=22670896

>  I can’t begin to tell you how many times we have had people that need extra time to prepare for a meeting because of the background. As a manager, I have had lots of personal conversations with employees who legitimately stress out about getting the computer into a place with a good background.

- Zoom has custom backgrounds: https://support.zoom.us/hc/en-us/articles/210707503-Virtual-Background
- Microsoft Teams has background blur: https://support.office.com/en-us/article/blur-your-background-in-a-teams-meeting-f77a2381-443a-499d-825e-509a140f4780
- Skype has background blur: https://support.skype.com/en/faq/FA34896/what-is-background-blur-in-skype
- Jitsi has "Blur my background" in beta now as well: https://github.com/jitsi/jitsi-meet/blob/master/react/features/stream-effects/blur/JitsiStreamBlurEffect.js

PVince81:
### Research technologies

The goal is mostly to find one that performs best and where the license is compatible.

- [ ] tensorflow/body-pix
   - [ ] license ok?
   - [ ] prototype: https://github.com/nextcloud/spreed/pull/5918 by @ipasanec 
- [ ] mediapipe
   - [ ] license ok?
   - [ ] prototype
- [ ] tflite ? (as pointed at by @LukasReschke internally)
   - [ ] license ok?
   - [ ] prototype ?

- [ ] how big is the package to download, size of model ? should we load it on-demand after turning it on the first time ?
- [ ] is it also possible to replace the blur with an image or video ?

### Integration

- [ ] implement plugin endpoint to allow registering video filter plugins to avoid the need to touch any of the internals (see https://github.com/nextcloud/spreed/pull/5918 for locations that need touching)
- [ ] add button to toggle blur background on and off, possibly in the three dots menu with a keyboard shortcut "B"
- [ ] remember last blur setting per conversation/token like we do for video/audio state
- [ ] to test:
    - [ ] make sure that disabling/enabling video still keeps the blur
    - [ ] make sure the effect can survive disconnections/reconnections and the new "null stream" approach when idle
    - [ ] effect should not apply to screen shares, only videos
    - [ ] must continue working after [video constraints are applied automatically](https://github.com/nextcloud/spreed/blob/master/src/utils/webrtc/VideoConstrainer.js#L56) 
[...]

PR #5918, opened 2021-07-02
Background blur effect demo

ipasanec:
Hi Joas,
hi Vincent,

these are my code changes for the blur effect.
The blur stream is inserted in MediaDevicesManager. 
In order to end all streams and turn off the camera device, I had to dispatch an event in localmedia.

Sincerely yours
Immanuel


PVince81:
the code looks good so far, would be good to solve the following:
- [ ] fix eslint issues
- [ ] add the library to package.json with npm install, so it also appears in package-lock.json

I've added next steps for integrating this in the related ticket: https://github.com/nextcloud/spreed/issues/3151

package.json
```diff
@@ -17,37 +17,63 @@
...
 		"attachmediastream": "^2.1.0",
+		"babel-eslint": "^10.1.0",
+		"babel-loader": "^8.2.2",
 		"color.js": "^1.2.0",
 		"crypto-js": "^4.0.0",
+		"css-loader": "^4.3.0",
 		"debounce": "^1.2.1",
 		"emoji-regex": "^9.2.2",
 		"escape-html": "^1.0.3",
+		"eslint": "^7.30.0",
```
PVince81, line 48: You don't need to include eslint and webpack, these are implicitly imported from the existing dev dependencies `@nextcloud/webpack-vue-config` and others. Not sure if this happened on its own while you updated your branch with master ?

You only need to add the deps needed for video processing.

PVince81 (commented):
Thanks for the update, see comment about deps

package.json
ipasanec: Thanks for the advice!
The dependencies got accidentally while trying to get the dev installation to work. Should have used `--no-save`.
I will commit a fix.

PVince81:
obsoleted by https://github.com/nextcloud/spreed/pull/6036

PR #5923, opened 2021-07-02
Add mark as read function to dashboard

jakobroehrl:
image

### DIFF
--- a/src/dashboard.js
+++ b/src/dashboard.js
@@ -25,7 +25,6 @@ import { generateFilePath } from '@nextcloud/router'
 import { getRequestToken } from '@nextcloud/auth'
 import { translate, translatePlural } from '@nextcloud/l10n'
 import Dashboard from './views/Dashboard'
-import store from './store'
 
 // CSP config for webpack dynamic chunk loading
 // eslint-disable-next-line
@@ -49,7 +48,6 @@ document.addEventListener('DOMContentLoaded', function() {
 		const View = Vue.extend(Dashboard)
 		new View({
 			propsData: {},
-			store,
 		}).$mount(el)
 	})
 
--- a/src/dashboard.js
+++ b/src/dashboard.js
@@ -25,6 +25,7 @@ import { generateFilePath } from '@nextcloud/router'
 import { getRequestToken } from '@nextcloud/auth'
 import { translate, translatePlural } from '@nextcloud/l10n'
 import Dashboard from './views/Dashboard'
+import store from './store'
 
 // CSP config for webpack dynamic chunk loading
 // eslint-disable-next-line
@@ -48,6 +49,7 @@ document.addEventListener('DOMContentLoaded', function() {
 		const View = Vue.extend(Dashboard)
 		new View({
 			propsData: {},
+			store,
 		}).$mount(el)
 	})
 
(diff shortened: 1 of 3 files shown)


nickvergessen:
> But I see no change on the dashboard, what did I do wrong?

The dashboard is not using the store, so the "action" you are calling does nothing.

jakobroehrl:
> > But I see no change on the dashboard, what did I do wrong?
> 
> The dashboard is not using the store, so the "action" you are calling does nothing.

This is right. I get  Cannot read property 'dispatch' of undefined. 

Do you have any idea how I can import the store into https://github.com/nextcloud/spreed/blob/master/src/dashboard.js
This 2 tries doesn't work:
1 import store from './store/messagesStore.js'

2 const store = new Vuex.Store({
	modules: {
		messageActionsStore,
	},
})

gary-kim:
Have you tried importing it then making it part of the Vue init?

```javascript
import store from `./store/messageStore.js`
```

```javascript
OCA.Dashboard.register('spreed', (el) => {
		const View = Vue.extend(Dashboard)
		new View({
			propsData: {},
			store,
		}).$mount(el)
	})
```

nickvergessen:
Need to use the root storage, see other entry point files

jakobroehrl:
Thank you, it's working without errors now, but it has no effect. 
After triggering the mark read button, the conversation is not read.

jakobroehrl:
The problem is here: https://github.com/nextcloud/spreed/blob/master/src/store/messagesStore.js#L517-L520
I have a token and a context, but
context.getters.conversations[token] is undefined

PVince81:
@jakobroehrl this action expects that the conversation has already been fetched and exists.

Maybe in this context for the dashboard you'll need to add a flag to skip updating the "conversation" object and only have the call to the service to do the server side update.

Ideal solution (for later) would be to refactor the dashboard code to use the store to fetch the list of recent conversations and store them in the same location. But probably something for another time as it would quickly become too complex.

jakobroehrl:
> @jakobroehrl this action expects that the conversation has already been fetched and exists.
> 
> Maybe in this context for the dashboard you'll need to add a flag to skip updating the "conversation" object and only have the call to the service to do the server side update.
> 

@PVince81 
Thanks for you reply. Now I call https://github.com/nextcloud/spreed/blob/master/src/store/conversationsStore.js#L326-L336
and conversation is set. 
conversation.unreadMessages = 0
conversation.unreadMention = false

is working, too. But it has no effect. If I open the talk app after I mark as read in the dashbaord the conversation is still unread.

nickvergessen:
Rebased the branch to get rid of the 250 commits that were merged by accident.

And added a fixup  commit to fix using the service directly ignoring the store (which is not yet part of the dashboard)

Ready for review

jakobroehrl:
@nickvergessen 
Thanks for your work. 
I get this in the console if I click on the Dashboard Button:
Property oder Method "handlers" is not defined

nickvergessen:
did you npm ci && npm run dev?

nickvergessen:
> Property oder Method "handlers" is not defined

Found it now. It was also printed on master. Removed it therefor

jakobroehrl:
Working very good!

nickvergessen:
I noticed it's a bit confusing when the conversation is already read. But the WidgetItem only allows to specify the same actions for all items.
So you could have an item there in the list, then you click "Mark read" it reloads and nothing changes. You get the same items with the same options and can mark the same conversation read again.

marcoambrosini:
I'd say we should wait to merge this until it's possible to have different actions for different actions depending on the widgetitem

nickvergessen:
Reported at https://github.com/nextcloud/nextcloud-vue-dashboard/issues/286

PR #5925, opened 2021-07-02
Store the last "comment" message in the rooms table to compare it with the unread counter

nickvergessen:
Before a chat that was "read" but had a system message afterwards (like a user leaving)
would trigger another query to the oc_comments table to check for the unread message count,
although the "comment" counter could still be 0 as only "unreadable" messages were posted.


nickvergessen:
Not doing it for now in favor of #5926 
We can see how bad the performance still is afterwards.

PR #5926, opened 2021-07-02
Use memcache to save the query for unread message count

nickvergessen:

### DIFF
--- a/lib/Chat/ChatManager.php
+++ b/lib/Chat/ChatManager.php
@@ -77,6 +77,8 @@ class ChatManager {
 	protected $timeFactory;
 	/** @var ICache */
 	protected $cache;
+	/** @var ICache */
+	protected $unreadCountCache;
 
 	public function __construct(CommentsManager $commentsManager,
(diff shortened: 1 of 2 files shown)


blizzz (approved):

lib/Chat/ChatManager.php
```diff
@@ -315,7 +321,13 @@ public function getLastReadMessageFromLegacy(Room $chat, IUser $user): int {
 	}
 
 	public function getUnreadCount(Room $chat, int $lastReadMessage): int {
-		return $this->commentsManager->getNumberOfCommentsForObjectSinceComment('chat', (string) $chat->getId(), $lastReadMessage, 'comment');
+		$key = $chat->getId() . '-' . $lastReadMessage;
```
PVince81, line 324: add comment like:
"for a given message id $lastReadMessage we cache the number of messages that exist past that message, which happen to also be the number of unread messages, because this is expensive to query/compute."

PVince81 (approved):
:+1: looks good now

Issue #5937, opened 2021-07-04
Partial text is sent but not all after I press enter using an input method to compose words


jimlinntu:
## Steps to reproduce

https://user-images.githubusercontent.com/31609554/124377035-7774ef00-dcdc-11eb-9fa4-0f5b4e39f98e.mov

### Expected behaviour
I think all text should be output.

### Actual behaviour
Only partial text is output.

## Talk app

**Talk app version:** (see apps admin page: `/index.php/settings/apps`)
9.0.10

**Custom Signaling server configured:** yes/no and version (see additional admin settings: `/index.php/index.php/settings/admin/talk#signaling_server`)
unrelated to this issue

**Custom TURN server configured:** yes/no (see additional admin settings: `/index.php/settings/admin/talk#turn_server`)
unrelated to this issue
**Custom STUN server configured:** yes/no (see additional admin settings: `/index.php/settings/admin/talk#stun_server`)
unrelated to this issue

## Browser
Version 91.0.4472.114 (Official Build) (x86_64)
**Microphone available:** yes/no
yes
**Camera available:** yes/no
yes
**Operating system:** Windows/Ubuntu/...
macOS
**Browser name:** Firefox/Chrome/...
Chrome
**Browser version:** 50.1/55/...

jimlinntu:
I believe this issue is related to:

https://github.com/nextcloud/spreed/blob/master/src/components/NewMessageForm/AdvancedInput/AdvancedInput.vue

of not handling `compositionstart`, `compositionupdate` and `compositionend` events

PR #5939, opened 2021-07-04
Prevent submitting the message when the user is composing

jimlinntu:
Resolved https://github.com/nextcloud/spreed/issues/5937

I am not sure if this will cause incompatible between different browsers,
but at least I tested it on Chrome it seems great!

Signed-off-by: Jim Lin <b04705003@ntu.edu.tw>

Related information: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/isComposing

### DIFF
--- a/src/components/NewMessageForm/AdvancedInput/AdvancedInput.vue
+++ b/src/components/NewMessageForm/AdvancedInput/AdvancedInput.vue
@@ -321,6 +321,12 @@ export default {
 				return
 			}
 
+			// If the user is stil composing by an input method,
+			// we should not submit the message
+			if (event.isComposing) {
+				return
+			}
+
 			// TODO: add support for CTRL+ENTER new line
 			if (!(event.shiftKey)) {
 				event.preventDefault()


jimlinntu:
https://user-images.githubusercontent.com/31609554/124380188-f888b200-dced-11eb-978e-558b74a5c6c8.mov

PVince81:
@jimlinntu 谢谢. I wasn't aware that there was an attribute `isComposing`. :smile: 

For me there was never a problem with input method and hitting enter while compositing.

I've still tested with fcitx on Linux with Chromium 91 and Firefox 89 and there was no side effect.

The PR seems fine and low-risk, merging.

PVince81:
@jimlinntu mind telling us what input method you used and on what platform, for the record ? Thanks

PVince81 (approved):
:+1:

nickvergessen:
Also needs to be copied to the vue lib, otherwise #4333 will bring it back as a regression?

PVince81:
I'll add it to the checklist there to keep it in mind, then we can check how to put it into the lib

jimlinntu:
> @jimlinntu mind telling us what input method you used and on what platform, for the record ? Thanks

I am using macOS Mojave 10.14.6

jimlinntu:
Just curious. I am currently using Nextcloud 19, is it possible to backport this version to NC19?

That will be great because I will not need to be aware that I can only press one Enter during my typing 😅

PVince81:
@jimlinntu Nextcloud 22 just got announced yesterday and Nextcloud 19 is already end of life, and Nextcloud 20 will be end of life in October. I'd suggest you look into upgrading very soon :-)

PR #5942, opened 2021-07-05
Add missing import

nickvergessen:
Why does something like this not yield ESLint or fail the node build? -.-


PVince81 (approved):
:+1:

PVince81:
> Why does something like this not yield ESLint or fail the node build? -.-

maybe the version updates / streamlining is missing something, I'll check if it worked in the past

PVince81:
I've jumped to a past commit b4e07df4477d4ed53c42552f7b8321f0e937851d where we had the old eslint config version, but it still doesn't detect it.

Issue #5960, opened 2021-07-06
Unread marker glitches with unread pages of messages


PVince81:
## Steps to reproduce
1. Have 300+ unread messages in a conversation without being active in it yet, but also have the marker at the beginning of the chat on the first page.
2. Open the browser console
3. Open the conversation
4. Wait for load
5. When everything has stabilized, switch to another window then come back.

### Expected behaviour
Unread marker is in focus above the first unread message.
Counter in left sidebar stays stable.
No "clear" message in console.

### Actual behaviour
Unread marker stays visible.
Message `clearLastReadMessage because of isSticky` which means the conversation was marked as fully read due to "sticky" aka "being scrolled to bottom" even though we weren't.
Unread messages counter goes wild and increments by 100 for every loaded page.
After coming back to the chat, the counter is removed (0 unread messages)

## Talk app

**Talk app version:** 13.0.0 git master 72b592b17dd9d90b0aaf94cba58ebdc4bcc9bfe7

### Early clear

The problem only appears with a lot of messages. It is likely related to the logic that loads multiple pages of 100 messages below the marker. Before the first page loads the number of visible messages is not enough for a scrollbar to appear, so the code believes that we are "scrolled to bottom" and sets "isSticky". However, as new message appear, the sticky mode should be discarded when the scrollbar appears. The sticky mode is what makes the read marker get cleared too early, and the logic for the read marker is triggered by scroll events, indirectly triggered by the new messages being appended.

We need to delay any kind of scroll handler logic until all of the `lookForNewMessages` call have finished, and before we reach the blocking long-polling one.

### Counter issue

As for the counter going up, not sure yet, need to examine the matching code.

PVince81:
counter going up is because lookForNewMessages always counts up... it needs to only count if the new messages were from a long-poll call, basically if they are "brand new" and were not known before

PVince81:
fix is here: https://github.com/nextcloud/spreed/pull/5963

PR #5963, opened 2021-07-07
Fix read marker and counter issues with pending pages

PVince81:
Whenever there are several pages of unread messages, the
"lookForNewMessages()" call will not return all the results at once.

In case there are results left to load, the read marker is not reset by
the false when "isSticky" (aka scrolled to bottom) happens.

When loading new messages, we now only increase the conversation counter
when the latter was out of sync with the current messages list.

Fixes https://github.com/nextcloud/spreed/issues/5960

Make sure to check the steps there when testing as it requires 300+ messages as lookForNewMessages only returns 100 results at a time.

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -469,7 +469,7 @@ export default {
 				let hasScrolled = false
 				// if lookForNewMessages will long poll instead of returning existing messages,
 				// scroll right away to avoid delays
-				if (this.$store.getters.getLastKnownMessageId(this.token) === this.conversation.lastMessage.id) {
+				if (!this.$store.getters.hasMoreMessagesToLoad(this.token)) {
 					hasScrolled = true
 					await this.$nextTick(() => {
 						this.scrollToFocussedMessage()
@@ -658,6 +658,9 @@ export default {
 		 * Also see updateReadMarkerPosition() for the backend update.
 		 */
 		refreshReadMarkerPosition() {
+			if (!this.conversation) {
+				return
+			}
 			console.debug('setVisualLastReadMessageId token=' + this.token + ' id=' + this.conversation.lastReadMessage)
 			this.$store.dispatch('setVisualLastReadMessageId', {
 				token: this.token,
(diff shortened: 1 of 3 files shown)


PVince81:
I forgot to add unit tests. Sadly nothing failed so far... hmmm.

A review for the approach would still be nice.

- [ ] add JS tests

PVince81:
I've added tests but noticed a logic error, so will need to retest the isSticky fix

PVince81:
doesn't work correctly. there's something fishy with the logic, it seems to be reversed.
semantically it seems to make sense, so not sure what I mixed up there

PVince81:
=> seems to work now, probably a refresh issue

src/components/MessagesList/MessagesList.vue
```diff
@@ -699,8 +702,8 @@ export default {
 				return
 			}
 
-			// if we're at bottom of the chat, then simply clear the marker
-			if (this.isSticky) {
+			// if we're at bottom of the chat with no more new messages to load, then simply clear the marker
+			if (this.isSticky && !this.$store.getters.hasMoreMessagesToLoad()) {
```
nickvergessen, line 706: In line 472 you inject the token?

nickvergessen:
```suggestion
			if (this.isSticky && !this.$store.getters.hasMoreMessagesToLoad(this.token)) {
```

src/store/messagesStore.js
```diff
@@ -111,6 +111,24 @@ const state = {
 }
 
 const getters = {
+	/**
+	 * Returns whether more messages can be loaded, which means that the current
+	 * message list doesn't yet contain all future messages.
+	 * If false, the next call to "lookForNewMessages" will be blocking/long-polling.
+
```
nickvergessen, line 118:
```suggestion
	 *
```

src/store/messagesStore.js
```diff
@@ -111,6 +111,24 @@ const state = {
...
+	 * @param {object} getters the getters object.
+	 * @returns {bool} true if more messages exist that needs loading, false otherwise
+	 */
+	hasMoreMessagesToLoad: (state, getters) => (token) => {
+		const conversation = getters.conversation(token)
+		if (!conversation) {
+			return false
+		}
+
+		return getters.getLastKnownMessageId(token) < conversation.lastMessage.id
```
nickvergessen, line 129: So this can be inaccurate, as conversation.lastMessage is not always updated.
But I guess you only notice this when there are only such system messages on the "new page" of the chat. So let's ignore that for now.

nickvergessen (approved):

Issue #5959, opened 2021-07-06
Unread marker appearing at bottom when system messages are involved


PVince81:
## Steps to reproduce
1. Post a few messages
2. Delete the last message
or
1. Post a command, like "//me test"
or
1. Post an invalid command, like "/invalid"

then scroll to bottom

### Expected behaviour
Unread marker disappears as we're at the bottom.

### Actual behaviour
Unread marker is visible at the bottom of the chat without any further messages below.

## Talk app

**Talk app version:** 13.0.0 git master 72b592b17dd9d90b0aaf94cba58ebdc4bcc9bfe7

## Analysis

In Message.vue we're deciding whether to display the unread marker or not when at the bottom.
To detect if the message in question is the last one, we use `conversation.lastMessage.id`.
In the case of system messages, that id is not accurate as it points at the last readable message, not visible one.
We need to change this to use the last message present in the array inside of the messagesStore: let's add a getVisualLastMessageId() getter to retrieve it and then make Message.vue use that there.

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/5964

PR #5964, opened 2021-07-07
Never show read marker on the very last message

PVince81:
Added condition to detect the last message when deciding whether to
render the read marker.

Fixes issues when the last message is a deleted one or a command and
where the marker was displayed.

Fixes https://github.com/nextcloud/spreed/issues/5959

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -402,6 +402,10 @@ export default {
 
 	computed: {
 		isLastReadMessage() {
+			if (!this.nextMessageId) {
+				// never display indicator on the very last message
+				return false
+			}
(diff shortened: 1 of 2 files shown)


PVince81:
added test and squashed

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -402,6 +402,10 @@ export default {
 
 	computed: {
 		isLastReadMessage() {
+			if (!this.nextMessageId) {
+				// never display indicator on the very last message
+				return
```
nickvergessen, line 407:
```suggestion
				return false
```

nickvergessen (approved):

PVince81:
adjusted, thanks for spotting

Issue #5957, opened 2021-07-06
Error: There is already an encoder stored which handles exactly the same mime types.


nickvergessen:
## Steps to reproduce
1. Have messages in the chat

### Expected behaviour
✨ 

### Actual behaviour
```
 [Vue warn]: Error in mounted hook (Promise/async): "Error: There is already an encoder stored which handles exactly the same mime types."

found in

---> <AudioRecorder> at src/components/NewMessageForm/AudioRecorder/AudioRecorder.vue
       <NewMessageForm> at src/components/NewMessageForm/NewMessageForm.vue
         <ChatView> at src/components/ChatView.vue
           <MainView> at src/views/MainView.vue
             <AppContent>
               <Content>
                 <App> at src/App.vue
                   <Root> vue.runtime.esm.js:610
```

PVince81:
@marcoambrosini mind having a look ? this is blocking other work: it seems that the error is aborting the regular flow of calls that are usually done when posting, like clearing the read marker for example

PR #5966, opened 2021-07-07
 Initialize the wav encoder in the entry points instead of in the audiorecorder component

marcoambrosini:
fix: #5957


nickvergessen (approved):
xD now I see it

PR #5970, opened 2021-07-07
Wait for the wav encoder to be initialized before allowing recordings

marcoambrosini:

### DIFF
--- a/src/store/audioRecorderStore.js
+++ b/src/store/audioRecorderStore.js
@@ -1,0 +1,51 @@
+/**
+ * @copyright Copyright (c) 2021 Marco Ambrosini <marcoambrosini@pm.me>
+ *
+ * @author Marco Ambrosini <marcoambrosini@pm.me>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+import { register } from 'extendable-media-recorder'
+import { connect } from 'extendable-media-recorder-wav-encoder'
+
+const state = () => ({
+	encoderReady: false,
+})
+
+const getters = {
+	encoderReady: state => {
+		return state.encoderReady
+	},
+}
+
+const mutations = {
+	encoderReady: (state) => {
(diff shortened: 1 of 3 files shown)


src/App.vue
```diff
@@ -338,6 +338,7 @@ export default {
 		}
 
 		register(await connect())
+		emit('talk:audioEncoder:ready')
```
PVince81, line 341: wouldn't it be possible to check and wait for encoder state only in NewMessageForm or deeper instead of globally ?

marcoambrosini: The only 2 solutions that came up to me are a store or an event. I wouldn´t know how else to do it. And store seemed overkill. Do you have any ideas?

PVince81: can't you move register/connect to the NewMessageForm's in its mounted function ?

PVince81: and if it can be registered only once, store the fact that it was already registered.

one way would be also to wrap it in a service JS file that only initializes once ever.
you could all its initialize function then in NewMessageForm's mounted or even closer to the place where it's needed

marcoambrosini: The newmessageform gets thrown away and recreated when you start a call, that is why I moved the initialization of the encoder up at "entry point" level. Even if we would use a service file, the `encoderReady` value would need to be stored globally.

PVince81: encoderService.js
```
import ...

let encoderReady = false
async function initialize() {
    if (encoderReady) return
    await register(connect())
}

export default initialize
```

marcoambrosini: Actually that's probably what we should do, I just realised that my code has a bug. Audiorecordings won't work for the talk chat. I'll write the value in a store, sounds good?

PVince81: for a flag that only changes once from false to true and never again, I think it's ok to have it stored inside that service file

the benefit is greater as we only need to call initialize() in the one place we need it and not spread across the code with even extra events

PVince81: > Actually that's probably what we should do, I just realised that my code has a bug. Audiorecordings won't work for the talk chat. I'll write the value in a store, sounds good?

in that case transform the proposed "initialize" function into a store action, that would do.
the action could then be called multiple times but it would only initialize once

src/store/audioRecorderStore.js
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+import { register } from 'extendable-media-recorder'
+import { connect } from 'extendable-media-recorder-wav-encoder'
+
+const state = () => ({
```
PVince81, line 26: am wondering if we need a generic store category for that.
having a specific store submodule just for this feels a bit overkill

PVince81 (approved):
:+1: approach looks good, see minor comment

PVince81:
some weird unrelated errors:
```
Previous: PDOException: SQLSTATE[HY000]: General error: 4047 InnoDB refuses to write tables with ROW_FORMAT=COMPRESSED or KEY_BLOCK_SIZE.
```

src/store/audioRecorderStore.js
marcoambrosini: What do you have in mind?

PR #5974, opened 2021-07-08
Use async import for vue component on collections entrypoint

juliushaertl:
Use async loading of the roomlist component for the collections endpoint as otherwise the imports of dependent components from `@nextcloud/vue` takes up quite some time when loading the files app initially without triggering the display of the room selector.

image

| Before | After | 
| ------ | ----- |
| talk-collections-old | talk-collections-new |

### DIFF
--- a/src/collections.js
+++ b/src/collections.js
@@ -21,7 +21,6 @@
  */
 
 import Vue from 'vue'
-import RoomSelector from './views/RoomSelector'
 
 // eslint-disable-next-line no-unexpected-multiline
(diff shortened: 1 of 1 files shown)


juliushaertl:
Not sure if feasable but something similar might be useful for the files sidebar script, see [flamegraph](https://user-images.githubusercontent.com/3404133/124889665-983c8d80-dfd7-11eb-83cc-03cf051b3b41.png)

PVince81 (approved):
:+1: makes sense

PR #5975, opened 2021-07-08
[stable22] Unshare all items directly when deleting a room

nickvergessen:
Partial backport of #5889


PVince81 (approved):
:+1:

PVince81:
hmm, should also be backported to stable22.1 then ?

nickvergessen:
> should also be backported to stable22.1 then?

its backported by #5971

Issue #5395, opened 2021-03-22
End meeting for all


wiswedel:
**Is your feature request related to a problem? Please describe.**
As a moderator, I want to be able to end a meeting once and for all, so participants cannot misuse ongoing meeting rooms after the official meeting time is over.

**Describe alternatives you've considered**

- Remove every remaining participant one by one and then lock the room
- Remove every remaining participant one by one and then delete the room (both hard when there's a whole lecture hall full of uncontrollable users remaining in the room)
- Leave the call and delete the conversation while others are still in it --> not effective. call session remains open even after room is deleted

**Additional context**
Here's how it looks in Zoom:
![end meeting](https://user-images.githubusercontent.com/1804221/111981983-b35abe80-8b08-11eb-830c-ce15e4c46586.gif)

PVince81:
@wiswedel for the workaround, in an upcoming 11.1.x you don't need to remove participants from the room. Just lock the room or enable lobby and it will kill the call. (ref https://github.com/nextcloud/spreed/pull/5215)

nickvergessen:
> call session remains open even after room is deleted

That should not be the case.

PVince81:
@ma12-co where to put such button ? should the "leave call" button be made into a dropdown of some sorts ?

nickvergessen:
I think that is too trouble some.

Just for the record this is only for moderators.


cc @jancborchardt

PVince81:
On the API side:

The leave call is a `DELETE on /call/{token}`.
I suggest we either add an argument `?all=true` or make it `/call/{token}/all` to end the call for all.

PVince81:
@nextcloud/designers-talk for a design concept for where to put the button, as this is not trivial

nickvergessen:
From my perspective this is something for a managed call view which should be at the tip section of the sidebar.
There is currently a button to end the lobby and I think this is were the call moderation should be completely.

marcoambrosini:
I think this should go in the topbar. A split button seems to me like the most appropriate solution. Teams does exactly that.

![split teams](https://user-images.githubusercontent.com/26852655/124246547-5025f280-db21-11eb-8ae3-65d05dbf99e3.PNG)

PVince81:
PR here: https://github.com/nextcloud/spreed/pull/5981

PR #5981, opened 2021-07-08
📅 End meeting for all

PVince81:
- [x] backend impl (based on read-only mode kicking out users)
- [x] PHP tests
- [x] frontend button
- [x] vuex store action + backend call
- [x] only moderators or guest moderators can end for all
- [x] also send signaling message to kick people out more quickly ?
- [x] system message for "X has ended the call" instead of "you left the call"
- [x] Jest tests
- [x] Test with Android
- [x] Test with iOS

---

- [x] @marcoambrosini for the button design
- [x] @danxuliu for terminating the call screen as another user in `src/utils/webrtc/webrtc.js` (ref https://github.com/nextcloud/spreed/pull/5981#issuecomment-907206782 )
- [x] @nickvergessen for changing the system message and call summary

---

- [x] TEST: without HPB: end meeting for all :no_entry_sign:  does not kick out user from another browser
- [x] TEST: HPB: end meeting for all :no_entry_sign:  does not kick out user from another browser

Fixes https://github.com/nextcloud/spreed/issues/5395

### DIFF
--- a/lib/Controller/CallController.php
+++ b/lib/Controller/CallController.php
@@ -167,12 +167,7 @@ class CallController extends AEnvironmentAwareController {
 		}
 
 		if ($all && $this->participant->hasModeratorPermissions()) {
-			$participants = $this->participantService->getParticipantsInCall($this->room);
-
-			// kick out all participants out of the call
-			foreach ($participants as $participant) {
-				$this->participantService->changeInCall($this->room, $participant, Participant::FLAG_DISCONNECTED);
-			}
+			$this->participantService->endCallForEveryone($this->room, $this->participant);
 		} else {
 			$this->participantService->changeInCall($this->room, $this->participant, Participant::FLAG_DISCONNECTED);
 		}
--- a/lib/Controller/CallController.php
+++ b/lib/Controller/CallController.php
@@ -157,15 +157,25 @@ class CallController extends AEnvironmentAwareController {
 	 * @PublicPage
 	 * @RequireParticipant
 	 *
+	 * @param bool $all whether to also terminate the call for all participants
 	 * @return DataResponse
 	 */
-	public function leaveCall(): DataResponse {
+	public function leaveCall(bool $all = false): DataResponse {
 		$session = $this->participant->getSession();
 		if (!$session instanceof Session) {
 			return new DataResponse([], Http::STATUS_NOT_FOUND);
 		}
 
-		$this->participantService->changeInCall($this->room, $this->participant, Participant::FLAG_DISCONNECTED);
+		if ($all && $this->participant->hasModeratorPermissions()) {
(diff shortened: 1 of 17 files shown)


PVince81:
note: the current button is a hack of an Actions menu:
image

we might need to create a new component for that later, or extend the Actions component to allow for a label and also changing the default colors ?

nickvergessen:
From my pov we should go for a dedicated meeting setting section in the right sidebar first.
So there is quick access to mute all, end call, end lobby, ...

lib/Controller/CallController.php
```diff
@@ -152,15 +152,25 @@ public function updateCallFlags(int $flags): DataResponse {
...
 	 */
-	public function leaveCall(): DataResponse {
+	public function leaveCall(bool $all): DataResponse {
 		$session = $this->participant->getSession();
 		if (!$session instanceof Session) {
 			return new DataResponse([], Http::STATUS_NOT_FOUND);
 		}
 
-		$this->participantService->changeInCall($this->room, $this->participant, Participant::FLAG_DISCONNECTED);
+		if ($all) {
```
nickvergessen, line 164: Check for moderator permissions 🙈

PVince81:
> From my pov we should go for a dedicated meeting setting section in the right sidebar first.
> So there is quick access to mute all, end call, end lobby, ...

Ah yes, I remember now that you said it before. Got distracted by @marcoambrosini's latest proposal.

nickvergessen:
Assigning:
* @marcoambrosini for the button design
* @danxuliu for terminating the call screen as another user in `src/utils/webrtc/webrtc.js` (ref https://github.com/nextcloud/spreed/pull/5981#issuecomment-907206782 )
* @nickvergessen for changing the system message and call summary

nickvergessen:
activity is fixed

marcoambrosini:
So I discussed this with @jancborchardt and @nimisha-vijay and for now we're going to go with the button design that's currently implemented.

nickvergessen:
> Test with Android

Stays on the call screen: https://github.com/nextcloud/talk-android/issues/1660

> Test with iOS

Stays on the call screen: https://github.com/nextcloud/talk-ios/issues/658

nickvergessen:
Time to get it in then?

Rebased to fix up

Ivansss:
Could we add a capability here for the clients?

danxuliu (commented):
Is the last commit right? `Sounds.playLeave(true)` is still called both when the user leaves the call locally and when the meeting is ended remotely.

nickvergessen:
didn't ment to have that pushed. sorry, will remove again

nickvergessen:
> Could we add a capability here for the clients?

Available with `publishing-permissions` capability and only works for moderators

PR #5986, opened 2021-07-09
Fix constant names

nickvergessen:


PVince81 (approved):
:+1:

why not also change the usages also right away ? your magic IDE certainly is able to do so quickly ? :wink:

nickvergessen:
For now laziness

PR #5997, opened 2021-07-14
Add skip TLS option for XMPP and Rocketchat

### LINKED ISSUE #5568: Certificate is Invalid - Can't add SkipTLSVerify=true without .toml file being overwritten
Matterbridge version : 1.22.1

Hello,

I'm attempting to connect my Matterbridge to my ejabberd.  My ejabberd does not have a valid certificate and I would like to exclude or ignore the cert check from matterbridge.  

https://github.com/42wim/matterbridge/blob/master/matterbridge.toml.sample

I can see in the matterbridge.toml.example file how to disable or skip this check:

#Enable to not verify the certificate on your xmpp server.
#e.g. when using selfsigned certificates
#OPTIONAL (default false)
SkipTLSVerify=true

I've found my .toml file in my /tmp folder.  But when I try and update that file with this SkipTLSVerify=True...as soon as I start the matterbridge service or save the config on the matterbridge GUI in Nextcloud my .toml file is overwritten and the cert check is removed from my .toml file.

Please assist with how I can skip the check for TLS cert for matterbridge to connect to XMPP?  Could a checkbox be added to the matterbridge config to not check for TLS?

Thank you.

gary-kim:
Closes #5568

### DIFF
--- a/lib/MatterbridgeManager.php
+++ b/lib/MatterbridgeManager.php
@@ -432,6 +432,9 @@ class MatterbridgeManager {
 				$content .= sprintf('	Server = "%s"', $part['server']) . "\n";
 				$content .= sprintf('	Login = "%s"', $part['login']) . "\n";
 				$content .= sprintf('	Password = "%s"', $part['password']) . "\n";
+				if ($part['skiptls']) {
+					$content .= sprintf('	SkipTLSVerify = true') . "\n";
+				}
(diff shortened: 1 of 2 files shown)


eneiluj (approved):
Tested the config file generation :heavy_check_mark: 
I didn't check if Matterbridge actually uses this option.

nickvergessen (approved):

PR #6024, opened 2021-07-15
Check circles membership when trying to add

nickvergessen:

### DIFF
--- a/lib/Service/ParticipantService.php
+++ b/lib/Service/ParticipantService.php
@@ -413,7 +413,7 @@ class ParticipantService {
 		// FIXME use \OCA\Circles\Manager::getLink() in the future
 		$membersInCircle = $circle->getInheritedMembers();
 		foreach ($membersInCircle as $member) {
-			if ($member->getUserType() === Member::TYPE_USER && $member->getUserId() === $userId) {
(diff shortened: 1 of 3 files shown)


lib/Service/ParticipantService.php
```diff
@@ -396,6 +397,30 @@ public function addGroup(Room $room, IGroup $group, array $existingParticipants
...
+			$circle = Circles::detailsCircle($circleId);
+		} catch (\Exception $e) {
+			throw new ParticipantNotFoundException('Circle not found');
+		}
+
+		// FIXME use \OCA\Circles\Manager::getLink() in the future
+		$membersInCircle = $circle->getInheritedMembers();
+		foreach ($membersInCircle as $member) {
+			if ($member->getUserType() === Member::TYPE_USER && $member->getUserId() === $userId) {
+				return $circle;
```
ArtificialOwl, line 417: Overkill: check also that `$member->isLocal()`

ArtificialOwl (approved):

lib/Service/ParticipantService.php
nickvergessen: getUserType not covering this?

ArtificialOwl: in GlobalScale setup, userId might be identical while having different `getIinstance()`. isLocal() confirm the user is local

Issue #6044, opened 2021-07-20
Disable voice recording in read-only conversations


PVince81:
## Steps to reproduce
1. Create a conversation
2. Lock it
3. Click on the voice recording icon
or
1. Join the "Talk updates" conversation

### Expected behaviour
Button is disabled or absent.

### Actual behaviour
Button is there and triggers the voice recorder.

## Talk app

**Talk app version:** 12.0.1

@marcoambrosini something for you ? It's likely just a missing check for the read-only attribute

PVince81:
![image](https://user-images.githubusercontent.com/277525/126383775-0b0979b7-800b-44c9-a1ff-c6ad39d36333.png)


seems there's also another regression: the placeholder should say something like "you cannot post messages here" or something

PR #6051, opened 2021-07-23
Bugfix/6044/disable voice recording on readonly conversations

marcoambrosini:
fix: #6044

### DIFF
--- a/src/components/NewMessageForm/AudioRecorder/AudioRecorder.vue
+++ b/src/components/NewMessageForm/AudioRecorder/AudioRecorder.vue
@@ -28,8 +28,9 @@
 				content: startRecordingTooltip,
 				delay: tooltipDelay,
 			}"
-			:disabled="!encoderReady"
+			:disabled="!canStartRecording"
 			class="audio-recorder__trigger nc-button nc-button__main"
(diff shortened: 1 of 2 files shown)


src/components/NewMessageForm/AudioRecorder/AudioRecorder.vue
```diff
@@ -327,6 +346,18 @@ export default {
 				opacity: 1;
 			}
 		}
+		&--disabled {
```
PVince81, line 349: any chance to make this global in some way ?
how are we doing it for the other buttons like the attachment one ?

marcoambrosini: It'll be included in the button component, which I will pick up again soon!

PVince81:
I think it's not transparent enough, see if you compare it with the attachment icon:
image

Now it seems the attachment button might be an action icon while the voice recording one isn't:
image
(the hover is there even when disabled, it's a separate issue)

marcoambrosini:
@PVince81 I darkened the attachment Icon which was way too light. The emoji button behaves differently as its opacity increases on hover. I will standardise everything in the button component.

PVince81 (approved):
:+1:

Issue #6056, opened 2021-07-23
Forwarded message loses mention


PVince81:
## Steps to reproduce
1. Create two conversations "abc" and "def" with a user "Alice" and "Bob" in both
2. Enter "abc" as Alice
3. Send a message `Hello @Bob, what's up?`
4. Login as Bob
5. Forward the message to the room "def"
6. Open the room "def"

### Expected behaviour
The mention is kept

### Actual behaviour
The mention placeholder is seen and not kept.
<img width="485" alt="image" src="https://user-images.githubusercontent.com/277525/126783809-aaad5639-f6be-4c5b-9b4c-ac43d44658b9.png">

## Talk app

**Talk app version:** master 8190320b564cd7fb4dbdf8cd126fed66b071b008

I haven't tested yet with files but I suspect that the problem will be the same.
We can't just forward the text and need to somehow also forward the message parameters.

@marcoambrosini FYI

PVince81:
okay, seems that files cannot be forwarded

but objects will exist still in case of mentions

PR #6065, opened 2021-07-26
Fix forwarding mentions

marcoambrosini:
fix: #6056 

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/Forwarder.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/Forwarder.vue
@@ -59,6 +59,7 @@ import RoomSelector from '../../../../../views/RoomSelector.vue'
 import EmptyContent from '@nextcloud/vue/dist/Components/EmptyContent'
 import Modal from '@nextcloud/vue/dist/Components/Modal'
 import { showError } from '@nextcloud/dialogs'
+import cloneDeep from 'lodash/cloneDeep'
 
 export default {
 	name: 'Forwarder',
@@ -99,16 +100,37 @@ export default {
 		selectedConversationName() {
 			return this.$store.getters?.conversation(this.selectedConversationToken).name
 		},
+
+		/**
+		 * Object containing all the mentions in the message that will be forwarded
+		 * @returns {Object} mentions.
+		 */
+		mentions() {
+			const mentions = {}
+			for (const key in this.messageObject.messageParameters) {
+				if (key.startsWith('mention')) {
+					mentions[key] = this.messageObject.messageParameters[key]
+				}
+			}
+			return mentions
+		},
 	},
 
 	methods: {
 		async setSelectedConversationToken(token) {
 			this.selectedConversationToken = token
+			const messageToBeForwarded = cloneDeep(this.messageObject)
+			// Overwrite the selected conversation token
+			messageToBeForwarded.token = token
+			// If there are mentions in the message to be forwarded, replace them in the message
+			// text.
+			if (this.mentions !== {}) {
+				for (const mention in this.mentions) {
+					messageToBeForwarded.message = messageToBeForwarded.message.replace(`{${mention}}`, '@' + this.mentions[mention].name)
+				}
+			}
 			try {
-				const response = await this.$store.dispatch('forwardMessage', {
-					token,
-					message: this.messageObject,
-				})
+				const response = await this.$store.dispatch('forwardMessage', { messageToBeForwarded })
 				this.showForwardedConfirmation = true
 				this.forwardedMessageID = response.data.ocs.data.id
 			} catch (error) {
--- a/src/store/messagesStore.js
+++ b/src/store/messagesStore.js
@@ -35,7 +35,6 @@ import { showError } from '@nextcloud/dialogs'
 import {
 	ATTENDEE,
 } from '../constants'
-import cloneDeep from 'lodash/cloneDeep'
 
 /**
  * Returns whether the given message contains a mention to self, directly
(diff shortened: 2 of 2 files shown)


src/components/MessagesList/MessagesGroup/Message/MessagePart/Forwarder.vue
```diff
@@ -99,16 +100,37 @@ export default {
...
 		},
+
+		/**
+		 * Object containing all the mentions in the message that will be forwarded
+		 * @returns {Object} mentions.
+		 */
+		mentions() {
+			const mentions = {}
+			for (const key in this.messageObject.messageParameters) {
+				if (key.startsWith('mention')) {
```
PVince81, line 111: please use the param type to see if it's a "user", "guest" or "call" mention

there's some example code here: https://github.com/nextcloud/spreed/blob/bugfix/noid/forward-message-mention/src/store/messagesStore.js#L59

PVince81 (commented):
See comment.

Also, this doesn't work when mentioning the whole room. In the latter case, maybe we simply replace with the room name

src/components/MessagesList/MessagesGroup/Message/MessagePart/Forwarder.vue
marcoambrosini: I noticed that if I replace with .name instead of with .id, both room and user mentions work. I just pushed the change

src/components/MessagesList/MessagesGroup/Message/MessagePart/Forwarder.vue
```diff
@@ -99,16 +100,37 @@ export default {
...
 		async setSelectedConversationToken(token) {
 			this.selectedConversationToken = token
+			const messageToBeForwarded = cloneDeep(this.messageObject)
+			// Overwrite the selected conversation token
+			messageToBeForwarded.token = token
+			// If there are mentions in the message to be forwarded, replace them in the message
+			// text.
+			if (this.mentions !== {}) {
+				for (const mention in this.mentions) {
+					messageToBeForwarded.message = messageToBeForwarded.message.replace(`{${mention}}`, '@' + this.mentions[mention].name)
```
nickvergessen, line 129:
```suggestion
					messageToBeForwarded.message = messageToBeForwarded.message.replace(`{${mention}}`, '@' + this.mentions[mention].id)
```
Id or really name?

src/components/MessagesList/MessagesGroup/Message/MessagePart/Forwarder.vue
nickvergessen: Only if displayname === id, you have to use id.

nickvergessen:
> Also, this doesn't work when mentioning the whole room. In the latter case, maybe we simply replace with the room name

That is a huge privacy fail. Either we just do at all again or we replace it with soemthing generic which doesn't reveal the origin of the message

src/components/MessagesList/MessagesGroup/Message/MessagePart/Forwarder.vue
marcoambrosini: Always name, so that mentions for users will work and for rooms will display the room name instead of the token

marcoambrosini:
> That is a huge privacy fail.

Is it @nickvergessen ? It's just plain text. Anyone can make this up. There's no sign that this is a forwarded message. And there would be no way to tell that that was a room mention.

src/components/MessagesList/MessagesGroup/Message/MessagePart/Forwarder.vue
nickvergessen: But mentioning with display name does not work if you name is not your id (mostly the case).

nickvergessen: Since you can not mention a room but only at-all we can replace it

But same issue is with guests. A guest mention in another room does not resolve

PVince81: it doesn't harm if the mention doesn't resolve, unless the user is expecting some different behavior, like auto-inviting ? but I doubt it

PVince81:
> > That is a huge privacy fail.
> 
> Is it @nickvergessen ? It's just plain text. Anyone can make this up. There's no sign that this is a forwarded message. And even if there was, there would be no way to tell that that was a room mention.

I'd say it's the responsibility of the user to decide whether they want to forward the whole message, assuming that the whole content will be visible. At least as a user I wouldn't expect the software to decide for me to hide/censor content.

marcoambrosini:
so, @PVince81 ? Should we leave it as is? Approve? Reject?

PVince81:
> so, @PVince81 ? Should we leave it as is? Approve? Reject?

I'm still concerned about this comment: https://github.com/nextcloud/spreed/pull/6065#discussion_r678296382

maybe can you check if there is a difference between displayname, name and id in all the scenarios we considered ?
make sure to test with users who have the same display name like id and some where it's different
would be nice to list all the tested scenarios in bullet points

if the current code works already in all those scenarios I think we can then move forward with this.
and when Joas is back if there are further concerns we could adjust later on

src/components/MessagesList/MessagesGroup/Message/MessagePart/Forwarder.vue
marcoambrosini: I don't understand this either, If in that scenario `displayName === id`, why can't I just use displayName always as I'm doing here?

marcoambrosini:
Tested when:
- User has the same id and display name;
- User has different id;
- Room mention;
Works!

Issue #5723, opened 2021-06-07
🌏 Federated Conversations Overview


gary-kim:
# See [🌏 Testing federation while there is no frontend yet](https://github.com/nextcloud/spreed/issues/10686) for testing steps

Progress on implementing federated Nextcloud Talk rooms.

This implementation will have each room on a single server so each participating server will only communicate with the server hosting the room (This does mean that the room will be inaccessible regardless of the server the user is on if the server hosting the room is not available).

### Required

* [x] API compatibility checker
  * [x] Capability handling https://github.com/nextcloud/spreed/issues/10680
  * [ ] Publish OCM resource https://github.com/nextcloud/spreed/pull/10692
* [x] Remote message access API
* [x] Remote message send API
* [x] Remote user list + presence API
* [x] Remote API authentication
* [x] Add messages from other instances in the client receive messages API
* [x] Send messages to instance hosting the room
* [ ] Remote call attendance
  * [ ] Remote signaling
  * [ ] Remote Talk configuration info
  * [ ] Handling CSP issues
* [x] Keeping track of remote servers (current through is to use server_url but how to handle if that changes?) #5775 
* [ ] Request join system?
* [ ] Remote moderators?
* [ ] Token shares (for various ROS shares)
	* Probably public link shares at first. Can add proxying through remote server later.
* [x] Sending Shares #5775
* [x] Receiving Shares #6105 
* [x] Validation by remoteAttendeeId + accessToken might be ambiguous, maybe we should keep the invites in the DB with an accepted flag and add the remoteServer there? https://github.com/nextcloud/spreed/pull/10766
* [x] Send `sendRemoteUnShare` when removing participant https://github.com/nextcloud/spreed/pull/11464
* [x] Federation is not checking if the target user is allowed to use talk

### Integration tests

- [x] Test that it doesn't work when federation is not enabled
- [x] Test remote participant is added to participant list
- [x] Test invite is sent
- [x] Test accepting an invite (any change for the inviter?)
- [x] Test declining an invite (is the remote participant removed?)
- [x] Test requesting the room as a remote participant
- [x] Test if remote participants can see the participant list

### Bugs

- [x] Database has unique index on token which needs to be extended to token+server_url

@nickvergessen

nickvergessen:
> Remote message send API

I guess we use this for it:
[...]

PR #6105, opened 2021-08-11
Add sending room shares support

gary-kim:
For #5723 

This adds the backend for sending room shares.

Also includes tests for some of the sending and receiving functions.

### DIFF
--- a/lib/BackgroundJob/RetryJob.php
+++ b/lib/BackgroundJob/RetryJob.php
@@ -1,0 +1,97 @@
+<?php
+/**
+ * @copyright Copyright (c) 2016, ownCloud, Inc.
+ * @copyright Copyright (c) 2021 Gary Kim <gary@garykim.dev>
+ *
+ * @author Bjoern Schiessle <bjoern@schiessle.org>
+ * @author Björn Schießle <bjoern@schiessle.org>
+ * @author Joas Schilling <coding@schilljs.com>
+ * @author Lukas Reschke <lukas@statuscode.ch>
+ * @author Morris Jobke <hey@morrisjobke.de>
+ * @author Roeland Jago Douma <roeland@famdouma.nl>
+ * @author Gary Kim <gary@garykim.dev>
+ *
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program. If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+namespace OCA\Talk\BackgroundJob;
+
+use OCA\Talk\Federation\Notifications;
+use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\BackgroundJob\IJobList;
+use OCP\BackgroundJob\Job;
+use OCP\ILogger;
+
+/**
+ * Class RetryJob
+ *
+ * Background job to re-send update of federated re-shares to the remote server in
+ * case the server was not available on the first try
+ *
+ * @package OCA\Talk\BackgroundJob
+ */
+class RetryJob extends Job {
+	/** @var Notifications */
+	private $notifications;
+
+	/** @var int max number of attempts to send the request */
+	private $maxTry = 20;
+
+	/** @var int how much time should be between two tries (10 minutes) */
+	private $interval = 600;
+
+
+	public function __construct(Notifications $notifications,
+								ITimeFactory $timeFactory) {
+		parent::__construct($timeFactory);
+		$this->notifications = $notifications;
+	}
+
+	/**
+	 * run the job, then remove it from the jobList
+	 *
+	 * @param IJobList $jobList
+	 * @param ILogger|null $logger
+	 */
+	public function execute(IJobList $jobList, ILogger $logger = null) {
+		if (((int)$this->argument['try']) > $this->maxTry) {
+			$jobList->remove($this, $this->argument);
+			return;
+		}
+		if ($this->shouldRun($this->argument)) {
+			parent::execute($jobList, $logger);
+			$jobList->remove($this, $this->argument);
+		}
+	}
+
+	protected function run($argument) {
+		$remote = $argument['remote'];
+		$data = json_decode($argument['data'], true);
+		$try = (int)$argument['try'] + 1;
+
+		$this->notifications->sendUpdateDataToRemote($remote, $data, $try);
+	}
+
+	/**
+	 * test if it is time for the next run
+	 *
+	 * @param array $argument
+	 * @return bool
+	 */
+	protected function shouldRun(array $argument) {
+		$lastRun = (int)$argument['lastRun'];
+		return (($this->time->getTime() - $lastRun) > $this->interval);
+	}
+}
--- a/lib/BackgroundJob/RetryJob.php
+++ b/lib/BackgroundJob/RetryJob.php
@@ -67,7 +67,7 @@ class RetryJob extends Job {
 	 * @param IJobList $jobList
 	 * @param ILogger|null $logger
 	 */
-	public function execute(IJobList $jobList, ILogger $logger = null) {
+	public function execute(IJobList $jobList, ?ILogger $logger = null) {
 		if (((int)$this->argument['try']) > $this->maxTry) {
 			$jobList->remove($this, $this->argument);
 			return;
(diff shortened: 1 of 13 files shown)


gary-kim:
Ran into a slight issue while trying to write the reshare system.
Since rooms don't have an owner, for now, the "owner" field is just whoever sent the invitation but the federated id is also used for determining the room's home server, this won't work for reshares (invitations) from another server.
Maybe the `bridge-bot` account could be used as the default room owner if the room doesn't have a set owner?

cc @nickvergessen

gary-kim:
Just pushed a commit that will try to find a room owner but if it cannot find one, `bridge-bot` will be used instead.

nickvergessen:
> Maybe the bridge-bot account could be used as the default room owner if the room doesn't have a set owner?

The user might not exist on the other server and I would like to not require a user to be created.

What are you using the owner for?

lib/Controller/RoomController.php
```diff
@@ -1079,6 +1087,18 @@ public function addParticipantToRoom(string $newParticipant, string $source = 'u
...
+		} elseif ($source === 'remote') {
+			try {
+				$newUser = $this->cloudIdManager->resolveCloudId($newParticipant);
+			} catch (\InvalidArgumentException $e) {
+				return new DataResponse([], Http::STATUS_BAD_REQUEST);
+			}
+
+			$participantsToAdd[] = [
+				'actorType' => Attendee::ACTOR_FEDERATED_USERS,
+				'actorId' => $newUser->getId() . '@' . $newUser->getRemote(),
```
nickvergessen, line 1099: getId already contains the remote

lib/Federation/CloudFederationProviderTalk.php
```diff
@@ -256,10 +302,18 @@ private function notifyAboutNewShare(IUser $shareWith, string $shareId, string $
 		$this->notificationManager->notify($notification);
 	}
 
+	private function validSharedRoomTypes(): array {
+		return [
+			Room::TYPE_ONE_TO_ONE,
```
nickvergessen, line 307: We don't federate one-to-one chats, do we?

lib/Federation/Notifications.php
```diff
@@ -0,0 +1,254 @@
...
+
+		/** @var IUser|null $roomOwner */
+		$roomOwner = null;
+		try {
+			$roomOwners = $this->attendeeMapper->getActorsByParticipantTypes($room->getId(), [Participant::OWNER]);
+			if (!empty($roomOwners) && $roomOwners[0]->getActorType() === Attendee::ACTOR_USERS) {
+				$roomOwner = $this->userManager->get($roomOwners[0]->getActorId());
+			}
+		} catch (Exception $e) {
+			$roomOwner = $this->matterbridgeManager->getBotUser($room);
```
nickvergessen, line 114: I would just pick the first moderator if there is no owner. A local moderator is always needed, the last one leaving the chat is blocked and needs to delete the conversation instead of leaving it.

lib/Service/ParticipantService.php
```diff
@@ -317,6 +324,14 @@ public function addUsers(Room $room, array $participants): void {
 			$readPrivacy = Participant::PRIVACY_PUBLIC;
 			if ($participant['actorType'] === Attendee::ACTOR_USERS) {
 				$readPrivacy = $this->talkConfig->getUserReadPrivacy($participant['actorId']);
+			} elseif ($participant['actorType'] === Attendee::ACTOR_FEDERATED_USERS) {
+				if ($addedBy === null) {
+					continue;
```
nickvergessen, line 329: Please add to docs that `$addedBy` is required for federated users.

lib/BackgroundJob/RetryJob.php
```diff
@@ -0,0 +1,97 @@
+<?php
+/**
```
nickvergessen, line 2: declare strict

lib/BackgroundJob/RetryJob.php
```diff
@@ -0,0 +1,97 @@
...
+ * @copyright Copyright (c) 2016, ownCloud, Inc.
+ * @copyright Copyright (c) 2021 Gary Kim <gary@garykim.dev>
+ *
+ * @author Bjoern Schiessle <bjoern@schiessle.org>
+ * @author Björn Schießle <bjoern@schiessle.org>
+ * @author Joas Schilling <coding@schilljs.com>
+ * @author Lukas Reschke <lukas@statuscode.ch>
+ * @author Morris Jobke <hey@morrisjobke.de>
+ * @author Roeland Jago Douma <roeland@famdouma.nl>
+ * @author Gary Kim <gary@garykim.dev>
```
nickvergessen, line 12: is it so much copy paste?

lib/Model/AttendeeMapper.php
```diff
@@ -73,7 +73,25 @@ public function getById(int $id): Attendee {
...
+	 * @return Attendee
+	 * @throws DBException
+	 * @throws DoesNotExistException
+	 * @throws MultipleObjectsReturnedException
+	 */
+	public function getByRemoteIdAndToken(int $id, string $token): Attendee {
+		$query = $this->db->getQueryBuilder();
+		$query->select('*')
+			->from($this->getTableName())
+			->where($query->expr()->eq('remote_id', $query->createNamedParameter($id, IQueryBuilder::PARAM_INT)))
```
nickvergessen, line 93: remote_id is string in the database?

lib/BackgroundJob/RetryJob.php
gary-kim: It's mostly the same as https://github.com/nextcloud/server/blob/master/apps/federatedfilesharing/lib/BackgroundJob/RetryJob.php so just kept all the author names.

lib/Model/AttendeeMapper.php
gary-kim: Whoops, nice catch.

lib/Service/ParticipantService.php
```diff
@@ -350,6 +371,36 @@ public function addUsers(Room $room, array $participants): void {
...
 
+	public function getHighestPermissionAttendee(Room $room): ?Attendee {
+		$attendee = null;
+		try {
+			$roomOwners = $this->attendeeMapper->getActorsByParticipantTypes($room->getId(), [Participant::OWNER]);
+
+			if (!empty($roomOwners)) {
+				foreach ($roomOwners as $owner) {
+					if ($owner->getActorType() === Attendee::ACTOR_USERS) {
+						$attendee = $roomOwners[0];
```
nickvergessen, line 382:
```suggestion
						$attendee = $owner;
```

nickvergessen: But what you actually want to do is:
```suggestion
						return $owner;
```

lib/Service/ParticipantService.php
```diff
@@ -350,6 +371,36 @@ public function addUsers(Room $room, array $participants): void {
...
+				$roomModerators = $this->attendeeMapper->getActorsByParticipantTypes($room->getId(), [Participant::MODERATOR]);
+				if (!empty($roomOwners)) {
+					foreach ($roomModerators as $moderator) {
+						if ($moderator->getActorType() === Attendee::ACTOR_USERS) {
+							$attendee = $moderator;
+							break;
+						}
+					}
+				}
+			}
```
nickvergessen, line 398:
```suggestion
			$roomModerators = $this->attendeeMapper->getActorsByParticipantTypes($room->getId(), [Participant::MODERATOR]);
			if (!empty($roomModerators)) {
				foreach ($roomModerators as $moderator) {
					if ($moderator->getActorType() === Attendee::ACTOR_USERS) {
						return $moderator;
					}
				}
			}
```

lib/Service/ParticipantService.php
```diff
@@ -350,6 +371,36 @@ public function addUsers(Room $room, array $participants): void {
...
+		} catch (DBException $e) {
+		}
+		return $attendee;
```
nickvergessen, line 401:
```suggestion
		return null;
```

nickvergessen (commented):
Minor last things

gary-kim:
:+1: Let me fix those up

gary-kim:
Moving the highest permission function into `ParticipantService` has made a circular dependency. That needs to be fixed

nickvergessen:
I guess we should add a config which blocks this for now, so people don't spam content to the databases before we implemented everything.

lib/Controller/RoomController.php
```diff
@@ -1084,6 +1095,21 @@ public function addParticipantToRoom(string $newParticipant, string $source = 'u
 			$this->guestManager->sendEmailInvitation($this->room, $participant);
 
 			return new DataResponse($data);
+		} elseif ($source === 'remote') {
+			if (!$this->federationManager->isEnabled()) {
+				return new DataResponse([]. Http::STATUS_BAD_REQUEST);
```
nickvergessen, line 1100:
```suggestion
				return new DataResponse([], Http::STATUS_BAD_REQUEST);
```

nickvergessen (approved):

nickvergessen:
Wuhu, finally one step further :D

nickvergessen:
THanks for all the hard work and sorry for the slow reviews @gary-kim

PR #6110, opened 2021-08-12
Add automatic status change feature

CarlSchwan:
See for the API in server: https://github.com/nextcloud/server/pull/28403

Signed-off-by: Carl Schwan <carl@carlschwan.eu>

### DIFF
--- a/lib/Status/Listener.php
+++ b/lib/Status/Listener.php
@@ -1,0 +1,68 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2019 Joas Schilling <coding@schilljs.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Talk\Status;
+
+use OCA\Talk\Events\ModifyParticipantEvent;
+use OCA\Talk\Room;
+use OCA\Talk\Signaling\Messages;
+use OCA\UserStatus\Service\PredefinedStatusService;
+use OCP\EventDispatcher\IEventDispatcher;
+use OCP\IUserSession;
+use OCP\UserStatus\IManager;
+use OCP\UserStatus\IUserStatus;
+
+class Listener {
+	/** @var IUserSession */
+	public $userSession;
+
+	public function __construct(IUserSession $userSession) {
+		$this->userSession = $userSession;
+	}
+
+	public static function register(IEventDispatcher $dispatcher): void {
+		$dispatcher->addListener(Room::EVENT_BEFORE_SESSION_JOIN_CALL, static function (ModifyParticipantEvent $event) {
+			$room = $event->getRoom();
+			/** @var \OCA\Talk\Chat\SystemMessage\Listener $listener */
+			$listener = \OC::$server->get(self::class);
+			/** @var IManager $statusManager */
+			$statusManager = \OC::$server->get(IManager::class);
+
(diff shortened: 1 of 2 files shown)


lib/Status/Listener.php
```diff
@@ -0,0 +1,68 @@
...
+		$dispatcher->addListener(Room::EVENT_BEFORE_SESSION_JOIN_CALL, static function (ModifyParticipantEvent $event) {
+			$room = $event->getRoom();
+			/** @var \OCA\Talk\Chat\SystemMessage\Listener $listener */
+			$listener = \OC::$server->get(self::class);
+			/** @var IManager $statusManager */
+			$statusManager = \OC::$server->get(IManager::class);
+
+			$statusManager->setUserStatus($listener->userSession->getUser()->getUID(), PredefinedStatusService::CALL, IUserStatus::DND, true);
+			$messages = \OC::$server->get(Messages::class);
+			$messages->addMessageForAllParticipants($event->getRoom(), 'refresh-participant-list');
```
nickvergessen, line 53: This should not be done, it will kill the server performance. The next auto refresh is good enough (which will also bring the talks mic/video icon on the participants list

lib/Status/Listener.php
```diff
@@ -0,0 +1,68 @@
...
+
+	public static function register(IEventDispatcher $dispatcher): void {
+		$dispatcher->addListener(Room::EVENT_BEFORE_SESSION_JOIN_CALL, static function (ModifyParticipantEvent $event) {
+			$room = $event->getRoom();
+			/** @var \OCA\Talk\Chat\SystemMessage\Listener $listener */
+			$listener = \OC::$server->get(self::class);
+			/** @var IManager $statusManager */
+			$statusManager = \OC::$server->get(IManager::class);
+
+			$statusManager->setUserStatus($listener->userSession->getUser()->getUID(), PredefinedStatusService::CALL, IUserStatus::DND, true);
```
nickvergessen, line 51: As said in the past this is a very crucial and fatal automation which at least needs an opt-out.

If you are on dnd, you will not receive push notifications to your mobile apps anymore. Since now i would be automatically on dnd all the time I basically would never receive push notifications.

Also not every call is like a meeting.

lib/Status/Listener.php
```diff
@@ -0,0 +1,68 @@
...
+		$dispatcher->addListener(Room::EVENT_AFTER_SESSION_LEAVE_CALL, static function (ModifyParticipantEvent $event) {
+			$room = $event->getRoom();
+			/** @var \OCA\Talk\Chat\SystemMessage\Listener $listener */
+			$listener = \OC::$server->get(self::class);
+			/** @var IManager $statusManager */
+			$statusManager = \OC::$server->get(IManager::class);
+
+			$statusManager->revertUserStatus($listener->userSession->getUser()->getUID(), PredefinedStatusService::CALL, IUserStatus::DND);
+			$messages = \OC::$server->get(Messages::class);
+			$messages->addMessageForAllParticipants($event->getRoom(), 'refresh-participant-list');
```
nickvergessen, line 65: Same here, this needs to be dropped

nickvergessen (changes requested):

lib/Status/Listener.php
CarlSchwan: When discussing the feature with @jancborchardt and @nimisha-vijay we decided to keep it simple and not make it configurable.

Maybe something we could change instead of setting a DND status is to not change the current status and just change the message. DND is anyway often already enabled at the OS level when sharing a screen for example and I could see how not receiving notifications during every meeting could be annoying.

> Also not every call is like a meeting.

Trying to detect that type of call is currently happening is hard and probably not a good idea in terms of privacy since every user can see it.

nickvergessen: > Trying to detect that type of call is currently happening is hard and probably not a good idea in terms of privacy since every user can see it.

We will soon have such a distinction (see the roadmap project) to pick certain other settings too, e.g. if everyone can unmute or moderators have to give permissions etc.

Maybe we can discuss it next week when I'm back

nickvergessen: @jancborchardt @CarlSchwan Shall we have a call about it on monday?

CarlSchwan: Yes, I'm free for a call from 13h to 17h on Monday

jancborchardt:
So @nickvergessen @CarlSchwan and I had a call and we will do:
- We use "Away" instead of "Do not disturb" as dnd will prevent any push notifications. This would not be nice if important messages come in or you are in a longer coworking call where you actually are available.
- As mentioned in the spec: For proper feedback that your status changed, the avatar in the top right of the header should have the status icon/emoji overlay just like it does elsewhere too. This way you are more likely to see that your status has changed and can reset if you like.
- In the Status modal, between the "What’s your status" input and the 5 suggestions, we put a block like this (with the second line being a text link / button:
> Automatically set during calls and events.
> Reset to **📅 Vacation next week**

PR #6152, opened 2021-08-27
Add some logging to the device selection

nickvergessen:
Before this only the deviceId was used to pick a device.
If not it fell back to the first device in the list.
Now we first try to find the previous device based on the label.


marcoambrosini (approved):
Looks good

danxuliu (approved):
Code makes sense :+1: 

I have split the commit in two for better future reference, one to set the fallback device once all the devices were processed and another one for the logging.

PR #6167, opened 2021-08-30
Stop localmediacontrols propagation

marcoambrosini:

### DIFF
--- a/src/components/CallView/shared/LocalMediaControls.vue
+++ b/src/components/CallView/shared/LocalMediaControls.vue
@@ -60,13 +60,15 @@
 					:size="24"
 					title=""
 					fill-color="#ffffff"
-					decorative />
+					decorative
+					@click.stop />
 				<VideoOff
 					v-else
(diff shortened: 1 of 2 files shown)


danxuliu (changes requested):
Adding `@click.stop` in the icons (`VideoIcon`, `VideoOff` and `Hand`) is not needed, probably because the propagation is already stopped by the `@click.stop` in their parent button.

Adding `@click.stop` in the `ActionButton` does not seem to be needed either (except if the `ActionButton` is shown directly instead of in a popover); maybe the click is stopped by the popover that the button belongs to, or it does not reach the parent video component due to the popover being a descendant of the #content-vue or body elements rather than the video component element.

Finally, unfortunately adding `@click.stop` in the `Actions` component does not seem to work, as the click is still handled by the video mixin when the button/icon that shows the menu is clicked :shrug:

marcoambrosini:
Ouch! Let's then keep the current approach

PR #6171, opened 2021-08-30
Improve ESLint part 2

nickvergessen:


nickvergessen:
> /src/store/messagesStore.js
>   148:0  warning  @param "token" does not match an existing function parameter  jsdoc/check-param-names
>   161:0  warning  @param "token" does not match an existing function parameter  jsdoc/check-param-names

src/components/AdminSettings/Commands.vue
```diff
@@ -30,7 +30,9 @@
 			</small>
 		</h2>
 
-		<p class="settings-hint" v-html="commandHint" />
+		<p class="settings-hint">
+			{{ commandHint }}
+		</p>
```
nickvergessen, line 35: I think this breaks stuff, v-html warning was always there

marcoambrosini (approved):

PR #6178, opened 2021-08-31
Fix audio recordings in conversations with special chars and long names

nickvergessen:


marcoambrosini (approved):

Ivansss:
I'm getting errors when the file name is longer then 228 chars.
We could set the max length to 150 chars just to be safe.

PR #6190, opened 2021-09-01
Allow setting publishing permissions for all users

### LINKED ISSUE #931: Improve webinars
For creating webinars I need the possibility for the owner/moderator to disable microfon, camera and screensharing for all participants. 
I found the following line in the PR Announcement from January 2018
- Fullcontrol - the moderator can invite, mute and remove participants as needed.

But I didn't found any possibility to do that in Version 3.2.

nickvergessen:
- [x] Add API endpoint
- [ ] Add docs
- [x] Add unit tests
- [ ] Add integration tests
- [ ] Add events so updates trigger via HPB and internal signaling
- [ ] 
Fix #931


nickvergessen:
Questions:
- [ ] Should it be possible to remove the "in_call" flag from the permissions for moderators? I see it makes sense for users, to allowed individually in a lobby case, but a moderator could accidentally be locking themselves out, so I guess for moderators we would ignore it and always allow them to join.
- [ ] What should happen with people that already joined and get "in_call" removed, are they kicked, or allowed to stay?
- [ ] I think we should do a similar enhancement to setAttendeePublishingPermissions which currently only allows to "set" instead of +/- a single permission.

lib/Room.php
```diff
@@ -1030,6 +1049,21 @@ public function setSIPEnabled(int $newSipEnabled): bool {
...
 
+		return true;
+	}
+
+	public function setPermissions(string $level, int $newPermissions): bool {
+		$query = $this->db->getQueryBuilder();
+		$query->update('talk_rooms')
+			->set($level . '_permissions', $query->createNamedParameter($newPermissions, IQueryBuilder::PARAM_INT))
+			->where($query->expr()->eq('id', $query->createNamedParameter($this->getId(), IQueryBuilder::PARAM_INT)));
+		$query->execute();
```
vitormattos, line 1060:
```suggestion
		$query->executeQuery();
```

Issue #4623, opened 2020-11-17
Differenciate unread counters for group mentions and direct mentions


marcoambrosini:
Group mentions and direct mention chips should have different colors.

nickvergessen:
I still totally disagree.

Also in the subject you talk about unread counters and in the description about the mention chips?

nickvergessen:
Also with the upcoming introduction of groups where more a small group you are part of will be mentioned instead of `@all` all the time this should become less and less a problem.

ma12-co:
I don't have super strong opinions on this, I think it could be nice though. cc @jancborchardt

jancborchardt:
Right now, there are several different ways you can get mentioned, and thus a blue notification bubble or not:
1. Direct message in a 1on1 chat: This is obviously important for you as only you will see it
2. @mention in a group chat message: Also important since it’s directed at you, but others might also see it and/or answer, depending on the nature of the message
3. @mention among others in a group chat: Basically same as before, but in this case others are mentioned together with you in the message
4. @all / group mention in a group chat: For everyone in the conversation, with an extra nudge.
5. Just a simple message in a group chat: For everyone in the conversation, nothing extra. This is the only case where the unread bubble will be grey.

By mixing in **4** as visually having the same importance as 1–3 (which are definitely important for you _specifically_), we muddy what it means to be an "important" notification. I’m not saying to just make it grey too, but e.g. just give a blue border instead of making it indistinguishable from a mention specifically to you, creating information overload. (@jenniferpiperek concurred with this when we discussed it at the conference.)

E.g. in GitHub notifications, group mentions are split from mentions which are directed specifically at you, because they are differently important. (Further research would have to be done if we want to know how others do it.)


----

> Also with the upcoming introduction of groups where more a small group you are part of will be mentioned instead of @all all the time this should become less and less a problem.

What do you mean by this actually @nickvergessen?

nickvergessen:
#4324 restructured what a "attendee" is, its not long only users or guests, but (currently) also emails.
Now when you add a group, we can add the group also as an attendee, allowing us to track them and automatically add users to groups based on that, as well as we could do something similar to the guest mentions (`@"guest/sha1(session)`) to mention groups (`@group/engineering`) which would then allow to mention all members of the engineering group with one mention.

nickvergessen:
Backend is in progress at https://github.com/nextcloud/spreed/pull/6193

[...]

PR #6193, opened 2021-09-01
Add a flag for the direct mentions

nickvergessen:
This allows the UI can show the user bubble depending on user vs. all mentions

Backend for #4623

- [x] Expose whether there are personal mentions
- [x] Update the counter when a user was mentioned directly
- [x] Add docs
- [ ] Add unit tests
- [x] Add integration tests

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -16,7 +16,7 @@ And in the works for the [coming versions](https://github.com/nextcloud/spreed/m
 
 	]]></description>
 
-	<version>13.0.0-dev.1</version>
+	<version>13.0.0-dev.2</version>
 	<licence>agpl</licence>
 
 	<author>Daniel Calviño Sánchez</author>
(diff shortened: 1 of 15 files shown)


nickvergessen:
Since we can not reset the counter partly when the readmarker is changed, I will convert the API result to a boolean

nickvergessen:
Ready for review

danxuliu (changes requested):
The direct mention counter is reset when the last mention is read, so it is not properly reset if the last direct mention is read but there is another not direct mention after it.

I have added an integration test that shows it.

nickvergessen:
We don't have that data available `¯\_(ツ)_/¯`
To fix that we would need to retain a second real id. But I guess we could so that too.
Not sure why I didn't do it like that from the beginning

nickvergessen:
Fixed

danxuliu (commented):
Should the database field (and thus the attribute in the model) be named `last_mention_direct_message` rather than `last_mention_direct` for consistency with `last_mention_message`?

Other than that, :+1:

nickvergessen:
Would be `last_direct_mention_message` but I find this getting too long and unreadable (and already coming very close to the oracle size limit)

PR #6205, opened 2021-09-02
Unify handling of deleted users

### LINKED ISSUE #3598: Inconsistent handling of deleted users

nickvergessen:
- [x] System messages show "Deleted user" without any avatar and without user mention chip `#4CFF00`
- [x] Messages show "Deleted user" as author  `#4CFF00` with X as avatar (unchanged)
- [x] 1-to-1 conversations retain the name of the deleted user to give the other participant a chance to decide on deletion etc `#0094FF`

Notes:
* Mentions still show `@old-user-id` as otherwise all strings starting with `@` would be replaced with "Deleted user", as there is no record which user id ever existed. `#FF0000`
* As mentioned in the ticket we know the former user id sometimes (not all the times) and don't know a display name anymore, therefore can not show "Deleted user (xyz)" as requested by @jancborchardt so I went for "Deleted user". It would also not be sufficient for GDPR compliance

Fix #3598 

Bildschirmfoto von 2021-09-02 16-01-39


marcoambrosini (approved):

nickvergessen:
Fixed unit test by adding the expectedException call

PR #6237, opened 2021-09-16
Move unreadMessageElement from computed to a method

nickvergessen:
Depending on invocation order otherwise the computed
could be first used before the element was mounted to the DOM.
That would cache the computed and cause the element
to never exist although it was added to the DOM afterwards.

The problem with a cached "null" is that
"first unread message has not been seen yet, so don't move it"
then always exited the updating of the readmarker.

---

Might not solve the read-marker topic completely. I still have the issue on the company instance where I wonder why the following code would exit on the first already seen message instead of the message that is last fully displayed on the screen:
https://github.com/nextcloud/spreed/blob/050b0901e49fe4071b4ea66c9fcf740cf7e4439d/src/components/MessagesList/MessagesList.vue#L634-L644

But after talking with @PVince81 about it, we should not end up in this case anyway, as the following part should have executed. However because the company instance is not in debug mode and not compiled with vue debug assets, I can not find out which of the conditions doesn't match:
https://github.com/nextcloud/spreed/blob/050b0901e49fe4071b4ea66c9fcf740cf7e4439d/src/components/MessagesList/MessagesList.vue#L716-L721

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -145,20 +145,6 @@ export default {
 		},
 
 		/**
-		 * Finds the first visual unread message element
-		 *
-		 * @return {object} DOM element of the first unread message
-		 */
-		unreadMessageElement() {
-			let el = document.getElementById('message_' + this.visualLastReadMessageId)
-			if (el) {
-				el = el.closest('.message')
-			}
-
-			return el
-		},
-
-		/**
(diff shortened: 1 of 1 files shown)


PVince81 (approved):
:+1: since it doesn't really need to be cached and won't be called repeatedly

nickvergessen:
Doing the fixup

marcoambrosini (approved):

PR #6256, opened 2021-09-17
Don't error hard when adding users to room

nickvergessen:
Happened to me multiple times with the iOS app which sends the add requests in parallel
Can really write a test for it, but the flow is basically the following:

Request to add group with user "test1" | Request to add user "test1"
---|---
Gets list of participants | Gets list of participants
Gets group members | …
… | Adds user
Adds each user 💥 | ✅ 

Before:
```
An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 53-users-admin for key oc_talk_attendees.ta_ident
```


Ivansss (approved):
🐘

vitormattos (commented):
looks ok for me.

vitormattos (approved):

PR #5415, opened 2021-03-25
Split conversation integration tests into 2 folders

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>

PR #6259, opened 2021-09-17
[stable21] Split integration tests to avoid timeouts

nickvergessen:
Backport of #5415


vitormattos (commented):
What do you think about separating by context (test suite) and not by a sequential number (1, 2, 3,...)? Maybe it gets more semantic and easier to understand.

nickvergessen:
This is what we did already. The problem is some contexts are too big and run into a timeout. We can also rethink it for the master branch afterwards. but in the backport I want to keep it for now like we did it in the master already. I moved the same files to the `-2` folders which give the best time splitting (even between both)

vitormattos:
... step-2 or collection-2 instead of just -2?

PR #6263, opened 2021-09-19
l10n: Change word

Valdnet:
Changed word "malformed" to "inaccurate" because the location cannot be garbled, only inaccurate.


nickvergessen:
No, it's not about inaccurate, it's a denial error when you throw garbage against the API. So malformed data is what it refers to.

Valdnet:
It is not better, then, that they are "damaged"?

The audio may be distorted, not the data.

nickvergessen:
Its location data, not audio. If it's not a valid geo-id its someone trying to do dirty things they shouldn't be doing.

Valdnet:
> The audio may be distorted,

> Its location data, not audio.

It was just an example :wink:.

PR #6290, opened 2021-09-25
Profile action

Pytal:
[WHIP] Workable Hack In Progress

### DIFF
--- a/lib/Profile/TalkAction.php
+++ b/lib/Profile/TalkAction.php
@@ -1,0 +1,112 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright 2021 Christopher Ng <chrng8@gmail.com>
+ *
+ * @author Christopher Ng <chrng8@gmail.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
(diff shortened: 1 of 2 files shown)


lib/Profile/TalkAction.php
```diff
@@ -0,0 +1,112 @@
...
+		$targetUser = $this->userManager->get($this->userId);
+		if ($visitingUser === $targetUser) {
+			return $this->l10nFactory->get(Application::APP_ID)->t('Open Talk');
+		}
+		return $this->l10nFactory->get(Application::APP_ID)->t('Talk to %s', [$targetUser->getDisplayName()]);
+	}
+
+	public function getLabel(): string {
+		$visitingUser = $this->userSession->getUser();
+		$targetUser = $this->userManager->get($this->userId);
```
nickvergessen, line 89: Profile only works for logged in users? If it also works for guests, actions need to be able to opt out.

nickvergessen: Ah, now I got it, maybe the member can be renamed to `targetUserId`, because in controllers `userId` is always the currently active user

lib/Profile/TalkAction.php
```diff
@@ -0,0 +1,112 @@
...
+		return $this->urlGenerator->getAbsoluteURL($this->urlGenerator->imagePath('spreed', 'app-dark.svg'));
+	}
+
+	public function getTarget(): string {
+		$visitingUser = $this->userSession->getUser();
+		$targetUser = $this->userManager->get($this->userId);
+		if ($visitingUser === $targetUser) {
+			return $this->urlGenerator->linkToRouteAbsolute('spreed.Page.index');
+		}
+		return $this->urlGenerator->linkToRouteAbsolute('spreed.Page.index') . '?callUser=' . $this->userId;
```
nickvergessen, line 110: `$visitingUser` is null for guests and that will make the action redirect to a broken page.
If the actions are also used for guests it shoulduse:
```suggestion
		if (!$visitingUser || $visitingUser === $targetUser) {
			return $this->urlGenerator->linkToRouteAbsolute('spreed.Page.index');
		}
		return $this->urlGenerator->linkToRouteAbsolute('spreed.Page.index') . '?callUser=' . $this->userId;
```

nickvergessen: Also similar to the call action in the contacts menu we need to check if the talk app is enabled for the target users:
https://github.com/nextcloud/spreed/blob/84d42604e60aa1c31ecf698dcc069f36c494148a/lib/ContactsMenu/Providers/CallProvider.php#L81-L84

nickvergessen (changes requested):

PR #6294, opened 2021-09-27
add origin username in forward message

jakobroehrl:

image


nickvergessen:
It was an intentional decision to not do this and I would like to stick with that for now

PR #6297, opened 2021-09-28
Moved code to specialist method and add fallback to all

### LINKED ISSUE #5357: Add fallback for `@all` of other chat solutions
## Steps to reproduce
1. Start typing `@here` like in Slack

### Expected behaviour
Should get an autocomplete suggestion for `@all` which is the Talk way

### Actual behaviour
Plain text is posted

Code to adjust:
https://github.com/nextcloud/spreed/blob/2c3f80bb08793fec62a95478187ffafffb35bbff/lib/Controller/ChatController.php#L665

vitormattos:
## explaining changes
I moved the logic of add `@all` on the mentions array to a specialist method and refactored the logic to make more clean to add new mentions types and of course, I added `@here`

## To do
- [x] resolve #5357
- [x] tests

## Acceptance criteria
When I do a GET to `/api/{apiVersion}/chat/{token}/mentions` with query string `?search=here`
The response object will need a element on data property with `id: all`

### DIFF
--- a/lib/Chat/ChatManager.php
+++ b/lib/Chat/ChatManager.php
@@ -546,7 +546,7 @@ class ChatManager {
 			array_unshift($results, [
 				'id' => 'all',
 				'label' => $roomDisplayName,
-				'source' => 'calls',
+				'source' => 'calls'
 			]);
 		}
 		return $results;
--- a/lib/Chat/ChatManager.php
+++ b/lib/Chat/ChatManager.php
@@ -531,4 +531,37 @@ class ChatManager {
 	public function searchForObjects(string $search,  array $objectIds, string $verb = '', int $offset = 0, int $limit = 50): array {
 		return $this->commentsManager->searchForObjects($search, 'chat', $objectIds, $verb, $offset, $limit);
 	}
+
+	public function addConversationNotify(array $results, string $search, Room $room, Participant $participant): array {
+		if ($room->getType() === Room::TYPE_ONE_TO_ONE) {
+			return $results;
+		}
+		$attendee = $participant->getAttendee();
+		if ($attendee->getActorType() === Attendee::ACTOR_USERS) {
+			$roomDisplayName = $room->getDisplayName($attendee->getActorId());
+		} else {
+			$roomDisplayName = $room->getDisplayName('');
+		}
+		if ($search === '' || $this->searchIsPartOfConversationNameOrAtAll($search, $roomDisplayName)) {
+			array_unshift($results, [
+				'id' => 'all',
+				'label' => $roomDisplayName,
+				'source' => 'calls',
+			]);
+		}
+		return $results;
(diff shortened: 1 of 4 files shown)


lib/Chat/ChatManager.php
```diff
@@ -531,4 +531,37 @@ public function deleteMessages(Room $chat): void {
 	public function searchForObjects(string $search,  array $objectIds, string $verb = '', int $offset = 0, int $limit = 50): array {
 		return $this->commentsManager->searchForObjects($search, 'chat', $objectIds, $verb, $offset, $limit);
 	}
+
+	public function addGroupNotify(array $results, string $search, Room $room, Participant $participant): array {
```
nickvergessen, line 535: Let's call this:
```suggestion
	public function addConversationNotify(array $results, string $search, Room $room, Participant $participant): array {
```

as group notifications could refer to real Nextcloud groups in the future (now that groups can be added to conversations).

tests/php/Chat/ChatManagerTest.php
```diff
@@ -433,4 +434,125 @@ public function testClearHistory(): void {
...
+
+	/**
+	 * @dataProvider dataSearchIsPartOfConversationNameOrAtAll
+	 */
+	public function testSearchIsPartOfConversationNameOrAtAll(string $search, string $roomDisplayName, bool $expected): void {
+		$reflector = new \ReflectionClass('OCA\Talk\Chat\ChatManager');
+		$method = $reflector->getMethod('searchIsPartOfConversationNameOrAtAll');
+		$method->setAccessible(true);
+
+		$actual = $method->invokeArgs($this->chatManager, [$search, $roomDisplayName]);
```
nickvergessen, line 460: There is a method in the base class for that:

```suggestion
		$actual = self::invokePrivate($this->chatManager, 'searchIsPartOfConversationNameOrAtAll', [$search, $roomDisplayName]);
```

tests/php/Chat/ChatManagerTest.php
```diff
@@ -433,4 +434,125 @@ public function testClearHistory(): void {
...
+		return [
+			[
+				'',
+				function () {
+					$room = $this->createMock(Room::class);
+					$room->expects($this->once())
+						->method('getType')
+						->willReturn(Room::TYPE_ONE_TO_ONE);
+					return $room;
+				},
```
nickvergessen, line 474: Please move the `createMock` to the tests. Makes the data sets better forseeable what they try to test:
e.g. https://github.com/nextcloud/spreed/blob/acf6e50e3d6d22d8328bb831631e307c5cedbcd2/tests/php/Chat/Parser/SystemMessageTest.php#L135

nickvergessen (changes requested):

nickvergessen (approved):

PR #6300, opened 2021-09-29
Ignore sessions older than 30 seconds on notifications again

nickvergessen:
## Steps to reproduce with iOS
1. As user 1 create a chat with user 2
2. Set the setting to be notified for all messages
2. Join with the android/ios app into a chat
3. Join the call and kill the app
4. Optional: Make sure there is still a session entry in oc_talk_sessions
5. Wait 30 seconds
5. As user 2 open the chat 
6. Write a message into the chat (without mentioning the user)

## Steps to reproduce with hacking DB;
1. As user 1 create a chat with user 2
2. Set the setting to be notified for all messages
2. Switch to a different conversation
3. Manually create a valid session entry in oc_talk_sessions for user 1 with a last_ping far in the past
5. As user 2 open the chat 
6. Write a message into the chat (without mentioning the user)

## Without this PR
No notification for user 1

## With this PR
Notification for user 1

### DIFF
--- a/lib/Chat/Notifier.php
+++ b/lib/Chat/Notifier.php
@@ -33,6 +33,7 @@ use OCA\Talk\Model\Session;
 use OCA\Talk\Participant;
 use OCA\Talk\Room;
 use OCA\Talk\Service\ParticipantService;
+use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\Comments\IComment;
 use OCP\IConfig;
 use OCP\Notification\IManager as INotificationManager;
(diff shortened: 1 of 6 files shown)


lib/Chat/Notifier.php
```diff
@@ -385,7 +390,7 @@ protected function shouldParticipantBeNotified(Participant $participant, ICommen
 
 		if ($participant->getSession() instanceof Session) {
 			// User is online
-			return false;
+			return $participant->getSession()->getLastPing() > $this->timeFactory->getTime() - Session::SESSION_TIMEOUT;
```
vitormattos, line 393: I think is a good idea implement test to cover this scenario

nickvergessen: Added

Ivansss (approved):
Tested and works 👍

vitormattos (approved):

PR #6308, opened 2021-10-01
Implement remove file room of removed file

### LINKED ISSUE #1328: Add background job to remove file rooms of removed files
## Steps to reproduce
1. Upload a file
2. Start to chat about it
3. Delete the file
4. Also delete the file from trashbin if enabled

### Expected behaviour
The room should be deleted after some time

### Actual behaviour
The room exists until the end of the instance

vitormattos:
## Explaning changes
* Refactored the code to reduce the quantity of query's
* Fixed a bug when get `federationManager` property and property don't exists

## To do
* [x] resolve #1328
* [x] Resolve performance problem.
  The code do a lot of queries on database. Is good reduce the numbers of queries maintaining with a lower memory usage.
* [x] unit tests

### DIFF
--- a/lib/BackgroundJob/RemoveEmptyRooms.php
+++ b/lib/BackgroundJob/RemoveEmptyRooms.php
@@ -23,9 +23,6 @@ declare(strict_types=1);
 
 namespace OCA\Talk\BackgroundJob;
 
-use OC\Files\Filesystem;
-use OC\Files\View;
-use OCA\Talk\Federation\FederationManager;
 use OCA\Talk\Service\ParticipantService;
 use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\BackgroundJob\TimedJob;
@@ -53,9 +50,6 @@ class RemoveEmptyRooms extends TimedJob {
 	/** @var IUserMountCache */
 	protected $userMountCache;
 
-	/** @var FederationManager */
-	protected $federationManager;
-
 	protected $numDeletedRooms = 0;
 
 	/**
@@ -69,8 +63,7 @@ class RemoveEmptyRooms extends TimedJob {
 								Manager $manager,
 								ParticipantService $participantService,
 								LoggerInterface $logger,
-								IUserMountCache $userMountCache,
-								FederationManager $federationManager) {
+								IUserMountCache $userMountCache) {
 		parent::__construct($timeFactory);
 
 		// Every 5 minutes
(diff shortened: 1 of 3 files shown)


lib/BackgroundJob/RemoveEmptyRooms.php
```diff
@@ -47,15 +47,23 @@ class RemoveEmptyRooms extends TimedJob {
...
+	protected $userMountCache;
 
 	protected $numDeletedRooms = 0;
 
+	/**
+	 * The room to proccess
+	 *
+	 * @var Room|null
+	 */
+	protected $room = null;
```
nickvergessen, line 60: Made some bad experiences lately as cron jobs keep the state across multiple executions within the same `cron.php` call. It's not a problem here, but I would avoid the pattern. Can we instead pass around the room object?

lib/BackgroundJob/RemoveEmptyRooms.php
```diff
@@ -77,13 +86,47 @@ protected function run($argument): void {
...
 	public function callback(Room $room): void {
-		if ($room->getType() === Room::CHANGELOG_CONVERSATION) {
+		$this->defineRoomToProcces($room);
+		$this->deleteIfIsEmpty();
+		$this->deleteIfFileIsRemoved();
+	}
+
+	private function defineRoomToProcces(Room $room) {
+		if ($room->getType() === Room::TYPE_CHANGELOG) {
+			return;
```
nickvergessen, line 96: if this is true, `deleteIfIsEmpty` and `deleteIfFileIsRemoved` will again be called with the old room object

lib/BackgroundJob/RemoveEmptyRooms.php
vitormattos: The room object is changed on all executions. We can pass the object between the methods but will make the code biggest.

lib/BackgroundJob/RemoveEmptyRooms.php
vitormattos: You found a :bug: 
I will write a test to cover this scenario.

lib/BackgroundJob/RemoveEmptyRooms.php
vitormattos: Do you want that I thing on another approach?

nickvergessen: You could just pass in $room into the other methods?

nickvergessen (approved):

PR #6335, opened 2021-10-06
Reduce psalm infos: return types

vitormattos:
Steps:
* run `psalm --alter --issues=MissingClosureReturnType,MismatchingDocblockReturnType,MissingReturnType,InvalidReturnType,MissingParamType` and fixes manually

### DIFF
--- a/lib/Controller/SettingsController.php
+++ b/lib/Controller/SettingsController.php
@@ -96,7 +96,9 @@ class SettingsController extends OCSController {
 	}
 
 	/**
+	 * @param string $setting
 	 * @param int|null|string $value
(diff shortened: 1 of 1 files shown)


lib/Controller/SettingsController.php
```diff
@@ -95,6 +95,9 @@ public function setUserSetting(string $key, $value): DataResponse {
 		return new DataResponse();
 	}
 
+	/**
+	 * @param int|null|string $value
```
nickvergessen, line 99:
```suggestion
	 * @param string $setting
	 * @param int|null|string $value
 	 * @return bool
```

PR #6394, opened 2021-10-23
accept SHA256 hash sums for reference id as documented

blizzz:
- also use them for chat
- cf. https://github.com/nextcloud/spreed/blob/40caf331092c9dd5d23f1365d7fb4f0768d06b50/docs/chat.md

(Greetings from off-duty me, found it while wondering why the reference ids didn't match while hacking on https://codeberg.org/blizzz/harbour-nextcloud-talk/pulls/34)


nickvergessen:
21 in https://github.com/nextcloud/spreed/pull/6529
20 is EOL

PR #6427, opened 2021-10-28
Start with integration tests for invites

nickvergessen:
- [x] Test that it doesn't work when federation is not enabled
- [x] Test remote participant is added to participant list
- [ ] Test invite is sent
- [ ] Test accepting an invite (any change for the inviter?)
- [ ] Test declining an invite (is the remote participant removed?)
- [ ] Test requesting the room as a remote participant
- [ ] Test if remote participants can see the participant list


gary-kim:
This is just proving the usefulness of tests, isn't it.

I will get started on the messaging sharing after this is done.

nickvergessen:
Not there yet, but let's merge this in between step as it fixes some issues and brings us closer without having merge conflicts at the end

nickvergessen:
Open tasks where moved to https://github.com/nextcloud/spreed/issues/5723

vitormattos (approved):

PR #6437, opened 2021-11-02
Use a more descriptive displayId

Pytal:
Use a more descriptive displayId for the profile action, "Reach me via Talk" instead of just "Talk", so that it's clearer when users are configuring their profile settings, as suggested by @nimishavijay

- [x] Requires https://github.com/nextcloud/server/pull/29482 to allow for the displayId to be changed dynamically and not just written to the database

i.e.
image

### DIFF
--- a/lib/Profile/TalkAction.php
+++ b/lib/Profile/TalkAction.php
@@ -76,7 +76,7 @@ class TalkAction implements ILinkAction {
 	}
 
 	public function getDisplayId(): string {
-		return $this->l->t('Talk');
+		return $this->l->t('Contact via Talk');
 	}
 
 	public function getTitle(): string {


nickvergessen:
IIRC we avoid using me/you as much as possible in the UI.

nickvergessen:
Also it's the reverse of the Avatar menu action `Talk to X`, so the wording should somehow align from my POV.

nickvergessen:
Ah wait, it's only the label of the settings option?

Pytal:
> IIRC we avoid using me/you as much as possible in the UI.

> Also it's the reverse of the Avatar menu action Talk to X, so the wording should somehow align from my POV.

Yep, @nextcloud/designers feel free to adjust :)

> Ah wait, it's only the label of the settings option?

Currently yes, just used when the user configures visibility

lib/Profile/TalkAction.php
```diff
@@ -76,7 +76,7 @@ public function getId(): string {
 	}
 
 	public function getDisplayId(): string {
-		return $this->l->t('Talk');
+		return $this->l->t('Reach me via Talk');
```
jancborchardt, line 79: How about this? :)

```suggestion
		return $this->l->t('Contact via Talk');
```

nickvergessen (approved):

PR #6440, opened 2021-11-02
Implement integration tests and skeleton to new endpoints

### LINKED ISSUE #2513: Blocking
Similar to blocking in Social: https://github.com/nextcloud/social/issues/38

To deal with workplace harassment or generally ignoring certain accounts, we should allow people to block other accounts.

This would then result in them not being visible at all to you. No 1on1 conversations, not in groups, not in search, they would not be able to mention you, etc.

Slack for example does _not_ allow this and it’s quite a problem especially in larger companies: https://qz.com/1641708/slack-doesnt-care-that-you-cant-block-a-workplace-harasser/

vitormattos:
- [ ] API endpoints to
  - [ ] block user `POST /settings/block-user`
    - [ ] create table to save block relation
      | blocked_users      |
      | ------------ |
      | actor_type   |
      | actor_id     |
      | blocked_type |
      | blocked_id   |
      | datetime     |
    - [ ] integration tests flow to use on GET and DELETE endpoints
    - [ ] unit tests flow to use on GET and DELETE endpoints
  - [ ] list all users blocked `GET /settings/block-user`
  - [ ] unblock user `DELETE /settings/block-user`
- [ ] block user 2 to send private message to user 1 if user 1 blocked user 2
- [ ] user 1 don't receive notifications on groups with user 2 if user 1 blocked user 2 and user 2 user @all or @user1
- [ ] cover all with integration and unit tests

close #2513

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -648,6 +648,30 @@ return [
 				'apiVersion' => 'v1',
 			],
 		],
+		[
+			'name' => 'Settings#blockUser',
+			'url' => '/api/{apiVersion}/settings/block-user',
+			'verb' => 'POST',
+			'requirements' => [
+				'apiVersion' => 'v(4)'
+			]
+		],
+		[
+			'name' => 'Settings#unblockUser',
+			'url' => '/api/{apiVersion}/settings/block-user',
+			'verb' => 'DELETE',
+			'requirements' => [
+				'apiVersion' => 'v(4)'
+			]
+		],
+		[
+			'name' => 'Settings#listBlockedUsers',
+			'url' => '/api/{apiVersion}/settings/block-user',
+			'verb' => 'GET',
+			'requirements' => [
+				'apiVersion' => 'v(4)'
+			]
+		],
 
 		/**
 		 * HostedSignalingServer
--- a/lib/Controller/SettingsController.php
+++ b/lib/Controller/SettingsController.php
@@ -154,4 +154,25 @@ class SettingsController extends OCSController {
 
 		return new DataResponse();
 	}
+
+	/**
+	 * @NoAdminRequired
+	 */
+	public function blockUser($type, $blocked): DataResponse {
+		return new DataResponse([], Http::STATUS_OK);
+	}
+
+	/**
+	 * @NoAdminRequired
+	 */
+	public function unblockUser(): DataResponse {
+		return new DataResponse([], Http::STATUS_OK);
+	}
+
+	/**
+	 * @NoAdminRequired
+	 */
+	public function listBlockedUsers(): DataResponse {
+		return new DataResponse([], Http::STATUS_OK);
+	}
 }
--- a/tests/integration/features/bootstrap/FeatureContext.php
+++ b/tests/integration/features/bootstrap/FeatureContext.php
@@ -2126,4 +2126,24 @@ class FeatureContext implements Context, SnippetAcceptingContext {
 	protected function assertStatusCode(ResponseInterface $response, int $statusCode, string $message = '') {
 		Assert::assertEquals($statusCode, $response->getStatusCode(), $message);
 	}
+
+	/**
+	 * @When /^user "([^"]*)" block (user|group|email|circle) "([^"]*)" with (\d+) \((v4)\)$/
+	 *
+	 * @param string $user
+	 * @param string $type
+	 * @param string $blocked
+	 * @param int $statusCode
+	 * @param string $apiVersion
+	 */
+	public function userBlockUserWithV(string $user, string $type, string $blocked, int $statusCode, string $apiVersion) {
+		$this->setCurrentUser($user);
+		$this->sendRequest(
+			'POST', '/apps/spreed/api/' . $apiVersion . '/settings/block-user', [
+				'type' => $type,
+				'blocked' => $blocked
+			]
+		);
+		$this->assertStatusCode($this->response, $statusCode);
+	}
 }
--- a/tests/integration/features/settings/block-user.feature
+++ b/tests/integration/features/settings/block-user.feature
@@ -1,0 +1,19 @@
+Feature: settings/block-user
+
+  Background:
+    Given user "participant1" exists
+    Given user "participant2" exists
+    Given user "participant3" exists
+
+  Scenario: user 1 block user 2
+    When user "participant1" creates room "group room" (v4)
+      | roomType | 2 |
+      | roomName | room |
+    And user "participant1" adds user "participant2" to room "group room" with 200 (v4)
+    And user "participant1" adds user "participant3" to room "group room" with 200 (v4)
+    And user "participant1" gets the following candidate mentions in room "group room" for "" with 200
+      | id           | label                    | source |
+      | all          | room                     | calls  |
+      | participant2 | participant2-displayname | users  |
+      | participant3 | participant3-displayname | users  |
+    And user "participant1" block user "participant2" with 200 (v4)


appinfo/routes.php
```diff
@@ -648,6 +648,30 @@
 				'apiVersion' => 'v1',
 			],
 		],
+		[
+			'name' => 'Settings#blockUser',
+			'url' => '/api/{apiVersion}/settings/block-user',
+			'verb' => 'POST',
+			'requirements' => [
+				'apiVersion' => 'v(4)'
```
nickvergessen, line 656:
```suggestion
				'apiVersion' => 'v2'
```

Settings is on v1 at the moment, so we can go with v2?

lib/Controller/BlockController.php
```diff
@@ -0,0 +1,74 @@
...
+								) {
+		parent::__construct($appName, $request);
+		$this->blockAgentService = $blockAgentService;
+		$this->userSession = $userSession;
+	}
+
+	/**
+	 * @NoAdminRequired
+	 */
+	public function block(string $type = Attendee::ACTOR_USERS, string $blockedId): DataResponse {
```
nickvergessen, line 54: No default before requires parameters

lib/Controller/BlockController.php
```diff
@@ -0,0 +1,74 @@
...
+	public function unblock(string $type = Attendee::ACTOR_USERS, string $blockedId): DataResponse {
+		$this->blockAgentService->unblock(Attendee::ACTOR_USERS, $this->userSession->getUser()->getUID(), $type, $blockedId);
+		return new DataResponse([], Http::STATUS_OK);
+	}
+
+	/**
+	 * @NoAdminRequired
+	 */
+	public function listBlocked(): DataResponse {
+		$list = $this->blockAgentService->listBlocked($this->userSession->getUser()->getUID());
```
nickvergessen, line 71: Please insert actorType here too, so in the future we can also get the block feature for guests, etc

lib/Migration/Version13000Date20211102171628.php
```diff
@@ -0,0 +1,78 @@
...
+	 * @param IOutput $output
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if (!$schema->hasTable('talk_block_agent')) {
```
nickvergessen, line 48:
```suggestion
		if (!$schema->hasTable('talk_blocked')) {
```
Lets not use agent as a new word for people

lib/Model/BlockAgentMapper.php
```diff
@@ -0,0 +1,73 @@
...
+		$qb = $this->db->getQueryBuilder();
+
+		$qb->delete($this->tableName)
+			->andWhere(
+				$qb->expr()->eq('actor_type', $qb->createNamedParameter($entity->getActorType(), Types::STRING)),
+				$qb->expr()->eq('actor_id', $qb->createNamedParameter($entity->getActorId(), Types::STRING)),
+				$qb->expr()->eq('blocked_type', $qb->createNamedParameter($entity->getBlockedType(), Types::STRING)),
+				$qb->expr()->eq('blocked_id', $qb->createNamedParameter($entity->getBlockedId(), Types::STRING))
+			);
+		$qb->executeStatement();
```
nickvergessen, line 70: Is this needed? We just always delete based on Id and done?

lib/Service/BlockAgentService.php
```diff
@@ -0,0 +1,103 @@
...
+		}
+	}
+
+	public function unblock(string $actorType, string $actorId, string $blockedType, string $blockedId) {
+		$blockAgent = new BlockAgent();
+		$blockAgent->setActorType($actorType);
+		$blockAgent->setActorId($actorId);
+		$blockAgent->setBlockedType($blockedType);
+		$blockAgent->setBlockedId($blockedId);
+		$this->blockAgentMapper->delete($blockAgent);
```
nickvergessen, line 77: Just get from blockAgentMapper with a function that takes the parameters

lib/Service/BlockAgentService.php
```diff
@@ -0,0 +1,103 @@
...
+		$blockAgent->setActorType($actorType);
+		$blockAgent->setActorId($actorId);
+		$blockAgent->setBlockedType($blockedType);
+		$blockAgent->setBlockedId($blockedId);
+		$blockAgent->setDatetime($this->timeFactory->getDateTime());
+		try {
+			$this->blockAgentMapper->insert($blockAgent);
+			$blockedList = $this->cache->get($actorId) ?? [];
+			$blockedList[] = $blockAgent;
+			$this->cache->set($actorId, $blockedList);
```
nickvergessen, line 63: This is assuming only ever users would block, but it might be a guest feature too at some point.
I would also not cache it for now, until we know where we use it and which caching layout makes most sense.

lib/Model/BlockAgentMapper.php
```diff
@@ -0,0 +1,73 @@
...
+
+class BlockAgentMapper extends QBMapper {
+	/**
+	 * @param IDBConnection $db
+	 */
+	public function __construct(IDBConnection $db) {
+		parent::__construct($db, 'talk_block_agent', BlockAgent::class);
+	}
+
+	public function getBlockListByBlocker($blocker) {
```
nickvergessen, line 41: Should take the actorType and actorId
```suggestion
	public function getBlockListByActor(string $actorType, string $actorId): array {
```

lib/Model/BlockAgentMapper.php
```diff
@@ -0,0 +1,73 @@
...
+			->where($query->expr()->eq('actor_id', $query->createNamedParameter($blocker)));
+
+		$return = [];
+		$result = $query->executeQuery();
+		while ($row = $result->fetch()) {
+			/** @var BlockAgent */
+			$blockedAgent = $this->mapRowToEntity($row);
+			$return[$blockedAgent->getBlockedId()] = $blockedAgent;
+		}
+		return $return;
```
nickvergessen, line 54:
```suggestion
		return $this->findEntities($query);
```

especially since the "getBlockedId" may not be unique at some point

lib/Model/BlockAgentMapper.php
vitormattos: I don't implemented method to delete by ID, only implemented method to delete using blocked id and blocked type.
Do you think it will be necessary to delete by Id and by blocked id and blocked type?

lib/Service/BlockAgentService.php
vitormattos: Cache removed at this time

lib/Model/BlockAgentMapper.php
vitormattos: I implemented now two methods, to get all block list without making distinction by the type and other method defining the type.

lib/Migration/Version13000Date20211102171628.php
vitormattos: I renamed to talk_block_actor, this solve?

lib/Controller/BlockController.php
vitormattos: I implemented now two methods, to get all block list without making distinction by the type and other method defining the type.

lib/Migration/Version13000Date20211102171628.php
nickvergessen: yes

PR #6442, opened 2021-11-03
Show an error/warning when the HPB is incompatible

nickvergessen:
- [x] Show an error in the HPB configuration section

Bildschirmfoto von 2021-11-03 15-27-32

- [x] Show an error/warning for users

Bildschirmfoto von 2021-11-03 15-57-38

### DIFF
--- a/lib/Controller/SignalingController.php
+++ b/lib/Controller/SignalingController.php
@@ -216,6 +216,10 @@ class SignalingController extends OCSController {
 				],
 			]);
 
+			if (!$this->signalingManager->isCompatibleSignalingServer($response)) {
+				return new DataResponse(['error' => 'UPDATE_REQUIRED'], Http::STATUS_INTERNAL_SERVER_ERROR);
+			}
+
 			$body = $response->getBody();
 
 			$data = json_decode($body, true);
--- a/src/utils/signaling.js
+++ b/src/utils/signaling.js
@@ -962,7 +962,7 @@ Signaling.Standalone.prototype.helloResponseReceived = function(data) {
 		}
 	}
 
-	if (!this.features['audio-video-permissions']) {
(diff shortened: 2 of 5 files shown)


lib/Controller/SignalingController.php
```diff
@@ -216,6 +216,11 @@ public function getWelcomeMessage(int $serverId): DataResponse {
 				],
 			]);
 
+			$featureHeader = $response->getHeader('X-Spreed-Signaling-Features');
+			if (strpos($featureHeader, 'audio-video-permissions') === false) {
```
fancycode, line 220: While this works for now, better would be to split the comma-separated string and check the individual items.

fancycode (approved):
See comment above.

lib/Controller/SignalingController.php
nickvergessen: it's `, ` (comma + space) seperated, right?
I will split by command then then trim to be save

nickvergessen:
Added screenshots

danxuliu (approved):
Tested and works :+1: 

It would be nice if the message in the admin section mentions the minimum compatible version, but I think it is also fine not to mention it so the admin tries with the latest one ;-)

src/utils/signaling.js
```diff
@@ -962,6 +962,16 @@ Signaling.Standalone.prototype.helloResponseReceived = function(data) {
 		}
 	}
 
+	if (!this.features['audio-video-permissions']) {
```
fancycode, line 965: Better:
```
if (!this.hasFeature('audio-video-permissions')) {
```
https://github.com/nextcloud/spreed/blob/0e1b20723f6882a04a326e565a22abd17c3e25f1/src/utils/signaling.js#L161

nickvergessen: ah nice

PR #6450, opened 2021-11-05
Reconnect when permissions change with internal signaling

### LINKED ISSUE #6445: Granting permissions while being in a call requires the user to rejoin?
## Steps to reproduce
1. Create a conversation
2. Set conversation permissions to restricted
3. Add a user
4. Set users permissions to start+join call
5. As moderator start the call
6. As user join the call
7. As moderator set custom permissions for the user allowing them audio+video
8. As user try to enable mic/cam

### Expected behaviour
First you see "no permissions" then "no camera"

### Actual behaviour
First you see "no permissions" then "Enable camera"

## Talk app

**Talk app version:** master

### LINKED ISSUE #6446: Removing permissions while being in a call does not reflect
Reverse to #6445 so might be the same root cause/patch

## Steps to reproduce
1. Create a conversation
2. Set conversation permissions to **all**
3. Add a user
4. As moderator start the call
5. As user join the call
6. As moderator remove all permissions for the user

### Expected behaviour
Camera and audio for the participant stop

### Actual behaviour
Camera and audio for the participant continue

## Talk app

**Talk app version:** master

nickvergessen:
Fix #6445 
Fix #6446 

Otherwise it exists here:
https://github.com/nextcloud/spreed/blob/1e1cd5e25f785c32f33fc8af63b9d887d9d2a408/src/utils/webrtc/webrtc.js#L966-L968

### DIFF
--- a/lib/Controller/SignalingController.php
+++ b/lib/Controller/SignalingController.php
@@ -408,7 +408,7 @@ class SignalingController extends OCSController {
 				'lastPing' => $session->getLastPing(),
 				'sessionId' => $session->getSessionId(),
 				'inCall' => $session->getInCall(),
-				'permissions' => $participant->getPermissions(),
+				'participantPermissions' => $participant->getPermissions(),
 			];
 		}
 


danxuliu (commented):
The property in the participant was originally called `publishingPermissions`. In 390ae657a18523bb0b39f6531fa27de9bb075410 it was renamed, but it seems that the name was mixed: [the value returned by the HPB](https://github.com/nextcloud/spreed/commit/390ae657a18523bb0b39f6531fa27de9bb075410#diff-7ed631e9f04a0445d27e26f716dee988bd363d56cbfa4453ed74bd3605433f9aR299) (and what is checked in the WebRTC code) was renamed to `participantPermission`, while [the value returned by the internal signaling server](https://github.com/nextcloud/spreed/commit/390ae657a18523bb0b39f6531fa27de9bb075410#diff-9fc0995666c1b2714b4c68a64ca745e909c57584e059f40e3aec6ba76e10e408L407) is just `permissions`.

Maybe it would be enough to rename `permissions` to `participantPermissions` in the SignalingController rather than adding a new value? Or is `permissions` used somewhere else?

nickvergessen:
> Maybe it would be enough to rename permissions to participantPermissions in the SignalingController rather than adding a new value? Or is permissions used somewhere else?

I wondered the same, didn't see anything that reuses the signaling permissions.
But I also didn't know it's that new. So let me drop it.

danxuliu (approved):
Tested and works :+1:

PR #6459, opened 2021-11-08
Add ability to set default participant permissions

marcoambrosini:
Default | Non-default
---|---
Bildschirmfoto von 2021-11-10 09-35-29 | Bildschirmfoto von 2021-11-10 09-32-01

### DIFF
--- a/src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
+++ b/src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
@@ -159,6 +159,18 @@
 			<template v-if="selfIsModerator && !isModerator">
 				<ActionSeparator />
 				<ActionButton
+					v-if="hasNonDefaultPermissions"
+					:close-after-click="true"
+					@click="applyDefaultPermissions">
+					<template #icon>
+						<LockReset
+							:size="20"
+							title=""
+							decorative />
+					</template>
+					{{ t('spreed', 'Reset custom permissions') }}
+				</ActionButton>
+				<ActionButton
 					:close-after-click="true"
 					@click="grantAllPermissions">
 					<template #icon>
@@ -168,17 +180,6 @@
 							decorative />
 					</template>
 					{{ t('spreed', 'Grant all permissions') }}
-				</ActionButton>
-				<ActionButton
-					:close-after-click="true"
-					@click="applyDefaultPermissions">
-					<template #icon>
-						<LockReset
-							:size="20"
-							title=""
-							decorative />
-					</template>
-					{{ t('spreed', 'Default permissions') }}
 				</ActionButton>
 				<ActionButton
 					:close-after-click="true"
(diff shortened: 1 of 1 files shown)


src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
```diff
@@ -169,6 +169,17 @@
 					</template>
 					{{ t('spreed', 'Grant all permissions') }}
 				</ActionButton>
+				<ActionButton
```
nickvergessen, line 172: I would put it bottom (or top), not make a conditional button appear in the middle of the list and you always will have to search

src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
```diff
@@ -169,6 +169,17 @@
...
+					:close-after-click="true"
+					@click="applyDefaultPermissions">
+					<template #icon>
+						<LockReset
+							:size="20"
+							title=""
+							decorative />
+					</template>
+					{{ t('spreed', 'Default permissions') }}
```
nickvergessen, line 181: I would call it "Reset permissions"

src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
marcoambrosini: It's not conditional atm, but ok will put it on top

src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
marcoambrosini: fine by me

src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
marcoambrosini: Actually this will be a radio soon and it probably makes sense if it stays in the middle, because on the permissions scale it is in between

nickvergessen: So, I would not show this, if the attendee does not have custom permissions set for themselves but is currently following "call/conversation" permissions. And that's why I would then call it "reset" and have it at the end of the list.
It might or might not be bigger/smaller than Grant all/none, bout could also be very much one of that. But it's more about the fact that the user will now follow global rules again and their custom permission icon will vanish.

nickvergessen: So what do we do? as long as it's not a radio I guess it makes more sense to have the condition on the button, show it at the end of the list and use the wording "reset".
If it would be a radio default in the middle might make more sense, but we are running out of time.

Therefor my proposal would be the button approach for now. Have the commit available locally already:
Default | Non-default
---|---
Bildschirmfoto von 2021-11-10 09-35-29 | Bildschirmfoto von 2021-11-10 09-32-01

nickvergessen (approved):

PR #6460, opened 2021-11-09
Call view improvements

### LINKED ISSUE #6191: Design improvements in the call view
* Add some pixels breathing room between the videos
* Make sure the background between the videos is not black but dark grey
* Place icons/buttons of own video always at the same position (independent of grid vs. promoted vs. screenshare view, or when the grid gets another column/row)
* Sort people with videos to the front
* Fixed size of the grid (Teams has 9 max?) to avoid performance issues

marcoambrosini:
Fix #6191

### DIFF
--- a/src/components/CallView/shared/LocalMediaControls.vue
+++ b/src/components/CallView/shared/LocalMediaControls.vue
@@ -154,7 +154,7 @@
 					@click="toggleHandRaised">
 					<HandBackLeft
 						slot="icon"
-						:size="16"
+						:size="20"
 						decorative
 						title="" />
 					{{ raiseHandButtonLabel }}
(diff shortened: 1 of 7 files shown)


nickvergessen:
Text | Screenshot
---|---
Okay for now due to timeline, but must fix before final: Buttons are overflowing on phone screen size |  Bildschirmfoto am 2021-11-10 um 14 50 10

src/components/CallView/shared/LocalMediaControls.vue
```diff
@@ -166,23 +166,30 @@
...
+					@click="changeView">
+					{{ changeViewText }}
+				</ActionButton>
 				<ActionSeparator />
 				<ActionButton
 					icon="icon-settings"
 					:close-after-click="true"
 					@click="showSettings">
-					{{ t('spreed', 'Settings') }}
+					{{ t('spreed', 'Devices Settings') }}
```
nickvergessen, line 192:
```suggestion
					{{ t('spreed', 'Devices settings') }}
```

nickvergessen (approved):

PR #6501, opened 2021-11-15
Only run unit tests when php is affected

nickvergessen:
Reduces CI time with dependabot JS updates


nickvergessen:
See https://github.com/nextcloud/notifications/pull/1113 for demonstration

danxuliu (approved):

PR #6507, opened 2021-11-16
Overwrite hasCall based on chat messages

### LINKED ISSUE #6447: Delay between seeing a "X started a call" and being able to join
## Steps to reproduce
1. Be a participant without "Start a call" permissions
2. As a user join the room
3. As a moderator start the call
4. As user try to join the call

### Expected behaviour
Button turns green soon

### Actual behaviour
Button stays disabled until the 30s conversation list refresh kicks in which is rather random to the user.
Similar to the lobby screen, but there it is more hidden to the user that they would already be able to join.

## Talk app

**Talk app version:** master

nickvergessen:
Fix #6447

### DIFF
--- a/lib/Activity/Listener.php
+++ b/lib/Activity/Listener.php
@@ -133,15 +133,10 @@ class Listener {
 		$numGuests = $this->participantService->getGuestCount($room, $activeSince);
 
 		$message = 'call_ended';
-		if ((\count($userIds) + $numGuests) === 1) {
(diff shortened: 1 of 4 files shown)


nickvergessen:
Currently it prints messages like:
> Call with Adam and 0 guests (Duration 0:05)

We can have a look in removing "0 guests" after the string freeze.

marcoambrosini (approved):
Works!

PR #6521, opened 2021-11-17
Add breathing room to stripe

nickvergessen:
- [x] Add breathing room in the stripe as well
- [x] Round placeholder of promoted-placeholder
- [x] Round promoted video
- [x] Round own video?
- [x] Round corners of avatar placeholder, the transform is making this too complicated for me:
https://github.com/nextcloud/spreed/blob/0f69fef5c48e6d7f59af6f365aa17faa02e0c824/src/components/CallView/shared/VideoBackground.vue#L180

Bildschirmfoto von 2021-11-17 10-10-41

### DIFF
--- a/src/components/CallView/shared/LocalVideo.vue
+++ b/src/components/CallView/shared/LocalVideo.vue
@@ -341,6 +341,7 @@ export default {
 	overflow: hidden;
 	display: flex;
 	flex-direction: column;
+	margin-top: 8px;
 }
 
 .videoWrapper,
--- a/src/components/CallView/shared/LocalVideo.vue
+++ b/src/components/CallView/shared/LocalVideo.vue
@@ -336,12 +336,12 @@ export default {
 
 .video-container-stripe {
 	position:relative;
-	height: 100%;
 	flex: 0 0 300px;
 	overflow: hidden;
 	display: flex;
 	flex-direction: column;
 	margin-top: 8px;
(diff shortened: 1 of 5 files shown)


src/components/CallView/shared/LocalVideo.vue
```diff
@@ -336,11 +336,12 @@ export default {
 
 .video-container-stripe {
 	position:relative;
-	height: 100%;
 	flex: 0 0 300px;
 	overflow: hidden;
 	display: flex;
 	flex-direction: column;
+	margin-top: 8px;
+	height: 242px !important;
```
nickvergessen, line 344: with 3/4 cameras this skips the rounded corners.
the with is set to 300px (in flex) => /4*3 => 225 => 17 pixels of empty space around it, vs 2*10px of the radius is hardly noticable

Bildschirmfoto von 2021-11-18 08-48-28

Could we maybe make the flex grow wider, so the video always fills the height but instead grows by width in the stripe?

nickvergessen: Should be even worse when someone has a fullHD cam with 16/9 ratio

nickvergessen: Moved to https://github.com/nextcloud/spreed/issues/6540

marcoambrosini (approved):

PR #6552, opened 2021-11-18
Fix buttons in the topbar

marcoambrosini:
Screenshot 2021-11-18 at 09 29 35


nickvergessen (approved):
Looks good, but the screenshot is borked...

Also there is still no hover feedback on 3 of those buttons, right?

PR #6563, opened 2021-11-19
Allow to enable SIP without individual pin requirement

### LINKED ISSUE #4957: Dial-in without PIN
As a moderator, I would like to allow SIP dial-in without PIN, so I can invite an unspecified group of users to a public meeting and allow them to dial-in.

Non-PIN dial-in participants would appear as unnamed "Guest" in the participants list.

All registered invited users and email invitees shall remain receiving PINs.

Currently, only registered invited users and email invitees receive a PIN to dial-in. Guest participants with only the link to the conversation cannot dial-in.

Affected situations:
- Large meetings that are invited to via a mailing list
- forwarded invitations
- delegated participations

nickvergessen:
Fix #4957 

- [x] Allow setting SIP without PIN on the API
- [x] Add an option for the moderator to disable individual pin requirements

### Idea
When `sipEnabled` on the getRoom request is set to:
* `0` SIP is disabled
* `1` a user pin is required (current)
* `2` no pin is required, just the conversation token

### DIFF
--- a/appinfo/routes/routesRoomController.php
+++ b/appinfo/routes/routesRoomController.php
@@ -92,6 +92,8 @@ return [
 		['name' => 'Room#getParticipantByDialInPin', 'url' => '/api/{apiVersion}/room/{token}/pin/{pin}', 'verb' => 'GET', 'requirements' => array_merge($requirementsWithToken, [
 			'pin' => '^\d{7,32}$',
 		])],
+		/** @see \OCA\Talk\Controller\RoomController::createGuestByDialIn() */
+		['name' => 'Room#createGuestByDialIn', 'url' => '/api/{apiVersion}/room/{token}/open-dial-in', 'verb' => 'POST', 'requirements' => $requirementsWithToken],
 		/** @see \OCA\Talk\Controller\RoomController::setNotificationLevel() */
 		['name' => 'Room#setNotificationLevel', 'url' => '/api/{apiVersion}/room/{token}/notify', 'verb' => 'POST', 'requirements' => $requirementsWithToken],
 		/** @see \OCA\Talk\Controller\RoomController::setNotificationCalls() */
--- a/docs/capabilities.md
+++ b/docs/capabilities.md
@@ -95,3 +95,4 @@ title: Capabilities
 
 ## 15
 * `chat-permission` - When permission 128 is required to post chat messages, reaction or share items to the conversation
+* `sip-support-nopin` - Whether SIP can be configured to not require a custom attendee PIN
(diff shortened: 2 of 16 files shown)


fancycode:
While this works, the anonymous phone users are not shown in the participant list. For regular users Talk will receive the `joined` / `left` events (with `actorId` / `actorType`), but not for the anonymous users - this is something that's missing.

For public meetings, the SIP bridge currently accepts an empty PIN directly and joins the call to the conference without an additional API call to Talk. Regular users are checked with a call to `/room/<token>/pin/<pin>` and the resulting actor information is passed to Talk.

nickvergessen:
Does this already work with master of SIP+HPB so we can test locally?

nickvergessen:
@fancycode ^

fancycode:
Sorry for the late reply, I sent you an email. This needs a modification of the backend API of the SIP bridge to support the new value (`SIP_ENABLED_NO_PIN`).

nickvergessen:
Rebased to fix the conflict

fancycode:
> cc @fancycode the SIP bridge was updated?

Not yet, as currently the anonymous SIP users would not be visible in the participant list (see https://github.com/nextcloud/spreed/pull/6563#issuecomment-984657053). If that behaviour is fine, I can do the release but I assumed you want to somehow show these users in the participant list.

nickvergessen:
Hmm so what is the way forward? Needs more work on your side? Or an API point on our side first?

nickvergessen:
`POST /api/{apiVersion}/room/{token}/open-dial-in` endpoint added.

It will create a guest actor with a session and return the full room info with the participant

fancycode:
Works as expected, I can now join users with an empty pin. These users show up as "Guest" in the participant list with a phone icon. Should we show the (anonymized) number instead or keep it as-is?

nickvergessen:
We can add a display name option to the endpoint. The number would be shown to everyone then? Not sure that is too good?

fancycode:
> We can add a display name option to the endpoint. The number would be shown to everyone then? Not sure that is too good?

Yeah, that would be visible to all. Probably fine as a first step to keep `Guest` and change it later if users want this changed.

fancycode (approved):
Works for the SIP bridge, didn't check the Talk integration itself. I'll release the corresponding SIP bridge backend APIs so this is supported there.

nickvergessen:
Realised iOS has the option displayed in the settings, so solving https://github.com/nextcloud/talk-ios/issues/772 is required to get resolved before this can be released. So no backport I guess

Ivansss:
@nickvergessen I guess docs about [Enabled or disable SIP dial-in](https://nextcloud-talk.readthedocs.io/en/latest/webinar/#enabled-or-disable-sip-dial-in) need to be updated? :)

PR #6574, opened 2021-11-19
Fix topbar height

marcoambrosini:
Before:
Screenshot 2021-11-19 at 14 51 34

After:
Screenshot 2021-11-19 at 14 51 39

### DIFF
--- a/src/components/TopBar/TopBar.vue
+++ b/src/components/TopBar/TopBar.vue
@@ -560,6 +560,7 @@ export default {
 		justify-content: center;
 		width: 100%;
 		overflow: hidden;
+		height: $clickable-area;
 	}
 	.title {
 		font-weight: bold;


src/components/TopBar/TopBar.vue
```diff
@@ -560,6 +560,7 @@ export default {
 		justify-content: center;
 		width: 100%;
 		overflow: hidden;
+		height: $clickable-area;
```
nickvergessen, line 563: Could we instead change the height in the left sidebar? I don't want to risk breaking any layout in the callviews of sidebars, etc just because we change the height somewhere

marcoambrosini: No, this is a regression introduced with the wrapping of the controls in the callview. And here I'm touching only the text, which is not shown anywhere the callview

nickvergessen (approved):

PR #6583, opened 2021-11-22
Always open the conversation settings with the right data

nickvergessen:
1. Create 2 conversations and add their name as description
2. While being in conversation A open the settings of conversation B via the actions menu in the conversation list
3. Settings of conversation A are opened and modified


marcoambrosini (approved):
🙈

PR #6602, opened 2021-11-24
Check share restrictions on one to one conversation

vitormattos:
How test:

Background: user1 and user2 exists

Scenario:
- Login with User1
- Open Talk app
- Search by user2
- Define this app settings values:
  | appid | configkey                                     | configvalue |
  | ----- | --------------------------------------------- | ----------- |
  | core  | shareapi_restrict_user_enumeration_full_match | no          |
  | core  | shareapi_allow_share_dialog_user_enumeration  | yes         |
  | core  | shareapi_restrict_user_enumeration_to_group   | yes         |
  | core  | shareapi_restrict_user_enumeration_to_phone   | yes         |
- Click on user1 on search result
- You will need see 4xx error on Network tab on response of request to create conversation

> **PS**: If you define the app settings before do a search for user2 you will never see the user2

### DIFF
--- a/lib/Service/RoomService.php
+++ b/lib/Service/RoomService.php
@@ -32,6 +32,7 @@ use OCA\Talk\Participant;
 use OCA\Talk\Room;
 use OCP\EventDispatcher\IEventDispatcher;
 use OCP\IUser;
+use OCP\Share\IManager as IShareManager;
 
 class RoomService {
 
@@ -39,14 +40,18 @@ class RoomService {
 	protected $manager;
 	/** @var ParticipantService */
 	protected $participantService;
+	/** @var IShareManager */
+	protected $shareManager;
 	/** @var IEventDispatcher */
 	private $dispatcher;
 
 	public function __construct(Manager $manager,
 								ParticipantService $participantService,
+								IShareManager $shareManager,
 								IEventDispatcher $dispatcher) {
 		$this->manager = $manager;
 		$this->participantService = $participantService;
+		$this->shareManager = $shareManager;
 		$this->dispatcher = $dispatcher;
 	}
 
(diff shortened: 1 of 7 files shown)


lib/Service/RoomService.php
```diff
@@ -61,6 +66,10 @@ public function createOneToOneConversation(IUser $actor, IUser $targetUser): Roo
 			throw new InvalidArgumentException('invalid_invitee');
 		}
 
+		if (!$this->shareManager->currentUserCanEnumerateTargetUser($actor, $targetUser)) {
+			throw new RoomNotFoundException();
+		};
```
nickvergessen, line 71: should be inside the catch below, so that existing conversations can still be reopened

nickvergessen (approved):

nickvergessen:
Could have an integration test?

nickvergessen:
Integration test place is https://github.com/nextcloud/spreed/blob/master/tests/integration/features/conversation-2/one-to-one.feature#L13
Setting the configs can be done via something like https://github.com/nextcloud/spreed/blob/f4208eff934795b0cad9586293ff78ebb5d958d9/tests/integration/features/bootstrap/FeatureContext.php#L1748-L1763 (but with adjusted appid in the url

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -1808,8 +1809,10 @@ public function resetSpreedAppData() {
 		$currentUser = $this->currentUser;
 		$this->setCurrentUser('admin');
 		$this->sendRequest('DELETE', '/apps/spreedcheats/');
-		foreach ($this->changedConfigs as $config) {
-			$this->sendRequest('DELETE', '/apps/provisioning_api/api/v1/config/apps/spreed/' . $config);
+		foreach ($this->changedConfigs as $appId => $configs) {
+			foreach ($configs as $config) {
+				$this->sendRequest('DELETE', '/apps/provisioning_api/api/v1/config/apps/' . $appId . '/' . $config);
+			}
```
vitormattos, line 1815: @nickvergessen do you think that will be necessary preserve previous state of config if is a config of other app? Example: Make GET to store the current value before do a POST to define the config value. If the current value is null, will make a DELETE, if not, will make a POST

nickvergessen: From my POV this is a nice gimmic, but you pay with time for it. On each set you make 2 requests instead of one. Not very helpful.

PR #6603, opened 2021-11-24
Return room token

### LINKED ISSUE #4544: Return room token in occ talk:room:create
Write a shell script that creates a room with occ then adds participants.
1. Create a room with occ: `occ talk:room:create --owner admin room1`
2. Now try to add participants with `occ talk:room:add` but you don't have the token.

`occ talk:room:create` should output the resulting room token so that the script can grab it for further use.

vitormattos:
close #4544

### DIFF
--- a/lib/Command/Room/Create.php
+++ b/lib/Command/Room/Create.php
@@ -156,6 +156,7 @@ class Create extends Base {
 			$output->writeln(sprintf('<error>%s</error>', $e->getMessage()));
 			return 1;
 		}
+		$output->writeln('Room token: ' . $room->getToken());
 
 		$output->writeln('<info>Room successfully created.</info>');
 		return 0;


PVince81 (approved):
:+1:

vitormattos:
Implementing tests and small fixes on this command on other PR

lib/Command/Room/Create.php
```diff
@@ -156,6 +156,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 			$output->writeln(sprintf('<error>%s</error>', $e->getMessage()));
```
nickvergessen, line 156: Not an error :) use info or question?

vitormattos: I don't did this. I can change but is an error, the methods addRoom* trigger errors and this row is to show the error.

nickvergessen: Nevermind wrong line

PR #6613, opened 2021-11-25
Implement tests on command create room

vitormattos:
* Increase test coverage on command `talk:room:create`.
* Remove unnecessary rows.

### DIFF
--- a/lib/Command/Room/Create.php
+++ b/lib/Command/Room/Create.php
@@ -105,11 +105,6 @@ class Create extends Base {
 		$owner = $input->getOption('owner');
 		$moderators = $input->getOption('moderator');
 
-		if (!in_array($readOnly, [null, (string)Room::READ_WRITE, (string)Room::READ_ONLY], true)) {
-			$output->writeln('<error>Invalid value for option "--readonly" given.</error>');
-			return 1;
-		}
-
 		if (!in_array($listable, [
 			null,
 			(string)Room::LISTABLE_NONE,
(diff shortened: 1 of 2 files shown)


lib/Command/Room/Create.php
```diff
@@ -105,11 +105,6 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 		$owner = $input->getOption('owner');
 		$moderators = $input->getOption('moderator');
 
-		if (!in_array($readOnly, [null, (string)Room::READ_WRITE, (string)Room::READ_ONLY], true)) {
```
nickvergessen, line 108: Why is this?

vitormattos: On interaction on cli, this value never will be different of boolean because this option have the mode `InputOption::VALUE_NONE` and when the interaction is from cli, the value will be boolean.
Search by `InputOption::VALUE_NONE` on this page: https://symfony.com/doc/current/console/input.html

I believe that this conditional  is implemented because when we run unit tests is possible define a value different of boolean because the unit tests don't pass by validation types of options.

vitormattos: And the variable name is wrote using O (uppercase), different of the name defined on argument of method. This will generate a `Undefined variable` warning and never will enter on `if`.

nickvergessen: 🙈

nickvergessen (approved):

nickvergessen:
/backport 2ab6fa13b1a14306e94e4f369e4113b4721237ad to stable23

PR #6620, opened 2021-11-25
Allow to opt out of voice optimisations

nickvergessen:
- [x] Allow to select between full optimization and custom bits
- [ ] Decide whether to store in the **browser** or server
    - [ ] Load default value from there
- [ ] Decide whether to store **globally** or per conversation
- [ ] Actually use values in WebRTC getUserMedia

Peek 2021-11-25 17-13


nickvergessen:
Closing for now due to the lack of feedback on the implementation idea and the effort necessary in the end for such a niche feature.

PR #6640, opened 2021-12-01
Add command to remove users from all rooms

### LINKED ISSUE #5798: Add a command to remove a user from all their rooms
E.g. when a employee is going to be disabled undeploying them first might make sense for the visibility in the team

```
occ talk:user:remove USERID
```

vitormattos:
Close #5798

### DIFF
--- a/lib/Command/User/Remove.php
+++ b/lib/Command/User/Remove.php
@@ -1,0 +1,80 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2021, Vitor Mattos <vitor@php.rio>
+ *
+ * @author Vitor Mattos <vitor@php.rio>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Talk\Command\User;
+
+use OC\Core\Command\Base;
+use OCA\Talk\Manager;
+use OCP\IUserManager;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class Remove extends Base {
+	/** @var IUserManager */
+	private $userManager;
+	/** @var Manager */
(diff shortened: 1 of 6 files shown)


lib/Command/Room/TRoomCommand.php
```diff
@@ -314,6 +314,20 @@ protected function removeRoomParticipants(Room $room, array $userIds): void {
...
+			$user = $this->userManager->get($userId);
+			if (!$user) {
+				throw new InvalidArgumentException(sprintf("User '%s' not found.", $userId));
+			}
+			$users[] = $user;
+		}
+		foreach ($users as $user) {
+			$this->manager->removeUserFromAllRooms($user);
+		}
+	}
```
nickvergessen, line 329: if you don't need anything else from the TRoomCommand, I would say don't use it and just put the code directly into your Command class

lib/Command/User/Remove.php
```diff
@@ -0,0 +1,54 @@
...
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class Remove extends Base {
+	use TRoomCommand;
+
+	protected function configure(): void {
+		$this
+			->setName('talk:user:remove')
+			->setDescription('Remove a user from all their rooms')->addOption(
```
nickvergessen, line 40:
```suggestion
			->setDescription('Remove a user from all their rooms')
			->addOption(
```

tests/integration/features/command/user-remove.feature
```diff
@@ -0,0 +1,15 @@
...
+
+  Scenario: Remove a user from all their rooms
+    Given user "participant1" creates room "room" (v4)
+      | roomType | 3 |
+      | roomName | room |
+    And user "participant1" adds user "participant2" to room "room" with 200 (v4)
+    And invoking occ with "talk:user:remove --user participant2"
+    And the command output contains the text "Users successfully removed from all rooms"
+    Then the command was successful
+    And user "participant2" is participant of the following rooms (v4)
```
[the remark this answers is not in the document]
vitormattos: I changed the room type to one-to-one room only to cover the end of method but I think don't have api method to test this case.

Do you have any suggestion to test the last rows of method?

PS: I only moved this method from `lib/Listener/UserDeletedListener.php` to don't duplicate code.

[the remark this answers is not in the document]
nickvergessen: Pushed

lib/Manager.php
```diff
@@ -389,6 +389,25 @@ public function getLeftOneToOneRoomsForUser(string $userId): array {
 		return $rooms;
 	}
 
+	public function removeUserFromAllRooms(IUser $user) {
```
nickvergessen, line 392: From my POV this should go into the ParticipantService. The manager is "legacy" and should be replaced by a proper RoomService and ParticipantService in the future, so new stuff should go into those services directly

nickvergessen (approved):

lib/Manager.php
vitormattos: Remove UserFromAllRooms depend of Manager and when I put the Manager class on constructor of ParticipantService did recursive loop on dependency injection because the Manager require the ParticipantService class. To solve will be necessary move the dependency methods from Manager to other classes. This will be a big change.

nickvergessen: yeah got it already. So ignore for now

PR #6649, opened 2021-12-02
Improve grid pagination buttons

marcoambrosini:
- More opacity
- better positioning
- Shadow removed 
- Borders removed

Screenshot 2021-12-02 at 18 34 04
Screenshot 2021-12-02 at 18 54 40

### DIFF
--- a/src/components/CallView/Grid/Grid.vue
+++ b/src/components/CallView/Grid/Grid.vue
@@ -42,11 +42,13 @@
 			<div v-if="!isStripe || stripeOpen" class="wrapper" :style="wrapperStyle">
 				<div :class="{'stripe-wrapper': isStripe, 'wrapper': !isStripe}">
 					<button v-if="hasPreviousPage && gridWidth > 0 && showVideoOverlay"
+						:class="{'stripe': isStripe}"
 						class="grid-navigation grid-navigation__previous"
 						:aria-label="t('spreed', 'Previous page of videos')"
 						@click="handleClickPrevious">
 						<ChevronLeft
 							decorative
+							fill-color="#ffffff"
 							title=""
 							:size="24" />
 					</button>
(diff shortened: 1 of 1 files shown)


jancborchardt:
Much better! 2 details:
- The icon doesnt look to be centered in the button (the right one is 1-2 pixels off to the left visually, and vice versa). It's similar with the "bottom row collapse" button where the icon looks a bit further up than middle.
- Would it look better to have the buttons use a more transparent but dark background, with white icon? Just like the "bottom row collapse" button and the other actions at the top.

nickvergessen:
> Grey and the centering is off too

Grey is guest, centering is off becuase it's a lower case character

marcoambrosini:
>  but dark background, with white icon?

The main concern here was visibility, and if we do that visibility will decrease considerably IMO, especially when the videos are turned on

nickvergessen:
> especially when the videos are turned on.

That would exactly be my counter argument. Most people have brighter walls/backgrounds or even white walls. So when it is on top of a video a black button would stand out a bit more.

marcoambrosini:
This is like the stripe button, and I think we're going to regret it

Screenshot 2021-12-10 at 09 20 48

marcoambrosini:
A more opaque version could work, although it feels a bit strange, but maybe it's because of habit. I still think that the white one would be more visible. 

https://user-images.githubusercontent.com/26852655/145541455-83799491-2f96-4906-8c8a-b5c6ccd2a953.mov

nickvergessen:
Could also make it non opaque when hovering the video call area?

jancborchardt:
> A more opaque version could work, although it feels a bit strange, but maybe it's because of habit. I still think that the white one would be more visible. 
> 
> https://user-images.githubusercontent.com/26852655/145541455-83799491-2f96-4906-8c8a-b5c6ccd2a953.mov
> 
> 

I'd say that one is quite good, especially on videos as Joas mentions. But I'm fine either way – most important is that the button is not faded out. :)

marcoambrosini:
Ok all done, ready for review again :)

nickvergessen (approved):

PR #6701, opened 2021-12-14
Allow joining open conversations with passwords

nickvergessen:
See integration test for steps

### DIFF
--- a/tests/integration/features/conversation/join-listable.feature
+++ b/tests/integration/features/conversation/join-listable.feature
@@ -95,3 +95,18 @@ Feature: conversation/join-listable
       | regular-user           | USER              | users     |
       | user-guest@example.com | USER              | users     |
       | "guest"                | GUEST             | guests    |
+
+  # -----------------------------------------------------------------------------
+  # Join listed conversation which has a password
+  # -----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


tests/integration/features/conversation/join-listable.feature
```diff
@@ -95,3 +95,18 @@ Feature: conversation/join-listable
...
+  # -----------------------------------------------------------------------------
+  # Join listed conversation which has a password
+  # -----------------------------------------------------------------------------
+  Scenario: Only users with accounts can join an all-listed group room
+    Given user "creator" creates room "room" (v4)
+      | roomType | 3    |
+      | roomName | room |
+    When user "creator" sets password "foobar" for room "room" with 200 (v4)
+    And user "creator" allows listing room "room" for "all" with 200 (v4)
+    When user "regular-user" joins room "room" with 200 (v4)
```
vitormattos, line 108: Password isn't mandatory to join room?

nickvergessen: So if it's not listable that makes totally sense to be mandatory.
But when it's listable and you find it via search having to enter the password sounds like a very awkward flow.

vitormattos (approved):

PR #6721, opened 2021-12-15
FIx path to files

nickvergessen:
Regression from #6713 

> /apps/spreed/img/join_call.wav:1 Failed to load resource: the server responded with a status of 404 (Not Found)
> Uncaught (in promise) DOMException: Failed to load because no supported source was found.


danxuliu (approved):
Tested and works :+1:

PR #6727, opened 2021-12-20
l10n: Change "Video" to "Camera"

Valdnet:
Changed the word "Video" to "Camera" for the keyboard shortcut in the application options.

obraz

### DIFF
--- a/src/components/SettingsDialog/SettingsDialog.vue
+++ b/src/components/SettingsDialog/SettingsDialog.vue
@@ -101,7 +101,7 @@
 				<div>
 					<dt><kbd>V</kbd></dt>
 					<dd class="shortcut-description">
-						{{ t('spreed', 'Video on and off') }}
+						{{ t('spreed', 'Camera on and off') }}
 					</dd>
 				</div>
 				<div>


rakekniven:
Before changing this single string I would vote for digging through complete spreed (talk app).
Shortcut "V" needs to aligned as well.

Keep in mind that camera is the device for video. IMHO video is ok.

nickvergessen:
Yeah we took video to align with other solutions and the widespread v-shortcut

Valdnet:
In my opinion, it should be "Camera" because we are talking about a device here as is "Microphone". We don't have "Audio on and off", but "Microphone on and off" 😁.

rakekniven:
"Audio" is the output (sound) for me.

Valdnet:
> "Audio" is the output (sound) for me.

Same as "Video" is the output of the camera image. In my opinion it should be "Camera on and off".
It is about switching the device on and off, not the source of the output.
Examples of other text strings from Talk that refer to the device, not the output source:

- **Camera** and microphone check
- You can now change your **camera** and microphone while being in a call
- No **camera**
- Access to **camera** was denied
- Enable the **camera**
e.t.c.

nickvergessen:
So I'm fine changing this as it explains the details, but where we have less space (in the tooltips) we should stay with video to make the link to the V shortcut more obvious

nickvergessen (approved):

Issue #1920, opened 2019-06-25
Reactions to chat messages


georgehrke:
Something similar to Slack:
![](https://cdn0.tnwcdn.com/wp-content/blogs.dir/1/files/2015/07/reaction-examples.png)

[Source](https://thenextweb.com/apps/2015/07/09/how-did-we-ever-live-without-this/)

fsteinbauer:
This and threads are the crucial features we are missing to finally ditch Slack.

nickvergessen:
API in progress: https://github.com/nextcloud/spreed/pull/6730

nickvergessen:
Just for the record, there is a technical part to it. The summary of the emoji list is limited to 4k chars.

In telegram the list is also limited to 10 emojis:
👍👎❤️🔥🎉🤩😱😄😢💩🤢

juliushaertl:
Just for completeness, slack also allows any emoji (even custom ones):
![image](https://user-images.githubusercontent.com/3404133/149380898-0e62b7cf-dad7-4c65-ac83-5ac3c417105c.png)

vitormattos:
Complementing, follow reactions on rocket.chat, is very similar to Slack.

![Screenshot_20220113_161014](https://user-images.githubusercontent.com/1079143/149394005-4b8768b0-c7db-46af-a031-310ac8177f85.png)

![Screenshot_20220113_161101](https://user-images.githubusercontent.com/1079143/149394089-a38c1181-edc5-4d61-bb47-6d6caa7581e2.png)

![Screenshot_20220113_161256](https://user-images.githubusercontent.com/1079143/149394331-eb7dc11b-05d0-44ef-aeed-621c231044a0.png)

nickvergessen:
Seeing this, maybe on the summary of the reactions we sort by count and only summary the first X entries, to make sure the summary is less than 4k chars @vitormattos

vitormattos:
For now, the API don't restrict the emojis list to use, this rule is client side.

To display the reaction summary also is a client side implementation. Exists an endpoint to return a data dictionary with reactions emojis by type and the total of reactions using the emoji. I will change the return of API to return ordered by quantity to make more easy to display on client.

marcoambrosini:
It would be nice to have both approaches: the ability to quickly select one emoji in a small set and also the possibility to expand it to the full emoji picker

jancborchardt:
[...]

PR #6730, opened 2021-12-22
Api reactions to message

vitormattos:
Implement #1920

- [x] blocked by https://github.com/nextcloud/server/pull/30379

## Compatibility with old versions of clients
* When get messages from API, the API will return one more property on message (reaction)
* The reaction message is returned as system message and marked to don't reply
* When the author of reaction delete the reaction, the message will be displayed on on clients same as deleted message

## TODO on new clients
* Don't display messages of type `reaction` and `reaction_deleted`
* When receive the response of endpoint `GET /api/{apiVersion}/chat/{token}` (`receiveMessages`) with `reaction` or `reaction_deleted`, get the id of parent message on reaction and update this.
  > **P.S.**: Use the `delete` message type as example to change the reactions counter:
  > https://github.com/nextcloud/spreed/blob/091575237c17945b37c7c72ba53c5ad85b1961ae/src/store/messagesStore.js#L255-L258

## How client's developers can do review on this PR
* Look the tests scenarios on [tests/integration/features/reaction/react.feature](https://github.com/nextcloud/spreed/blob/542ec3d1f7811564b9d8af96555cf9144db2303c/tests/integration/features/reaction/react.feature)
* Search on [Files changed](https://github.com/nextcloud/spreed/pull/6730/files) by `.md` and look the changes on current endpoints and the new endpoints documentation.

### DIFF
--- a/docs/chat.md
+++ b/docs/chat.md
@@ -50,6 +50,7 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
         `message` | string | Message string with placeholders (see [Rich Object String](https://github.com/nextcloud/server/issues/1706))
         `messageParameters` | array | Message parameters for `message` (see [Rich Object String](https://github.com/nextcloud/server/issues/1706))
         `parent` | array | **Optional:** See `Parent data` below
+        `reactions` | array | **Optional:** An array map with relation between reaction emoji and total of reactions with this emoji
 
 #### Parent data
 
@@ -324,3 +325,4 @@ See [OCP\RichObjectStrings\Definitions](https://github.com/nextcloud/server/blob
 * `matterbridge_config_removed` - {actor} removed the Matterbridge configuration
 * `matterbridge_config_enabled` - {actor} started Matterbridge
 * `matterbridge_config_disabled` - {actor} stopped Matterbridge
+
--- a/docs/chat.md
+++ b/docs/chat.md
@@ -50,7 +50,7 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
         `message` | string | Message string with placeholders (see [Rich Object String](https://github.com/nextcloud/server/issues/1706))
         `messageParameters` | array | Message parameters for `message` (see [Rich Object String](https://github.com/nextcloud/server/issues/1706))
         `parent` | array | **Optional:** See `Parent data` below
-        `reactions` | array | **Optional:** An array map with relation between reaction emoji and total of reactions with this emoji
+        `reactions` | array | **Optional:** An array map with relation between reaction emoji and total count of reactions with this emoji
 
 #### Parent data
 
--- a/docs/reaction.md
+++ b/docs/reaction.md
@@ -22,7 +22,7 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
 ## Delete a reaction
 
 * Method: `DELETE`
-* Endpoint: `/chat/{token}/{messageId}`
+* Endpoint: `/reaction/{token}/{messageId}`
 * Data:
 
     field | type | Description
--- a/docs/reaction.md
+++ b/docs/reaction.md
@@ -17,7 +17,7 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
     - Status code:
         + `200 OK` Reaction already exists
         + `201 Created`
-        + `400 Bad Request` In case of any other error
+        + `400 Bad Request` In case of no reaction support, message out of reactions context or any other error
         + `404 Not Found` When the conversation or message to react could not be found for the participant
         + `409 Conflict` User already did this reaction to this message
 
@@ -35,7 +35,7 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
 * Response:
     - Status code:
         + `201 Created`
-        + `400 Bad Request` In case of any other error
+        + `400 Bad Request` In case of no reaction support, message out of reactions context or any other error
         + `404 Not Found` When the conversation or message to react or reaction could not be found for the participant
 
 ## Retrieve reactions of a message by type
@@ -52,6 +52,7 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
 * Response:
     - Status code:
         + `200 OK`
+        + `400 Bad Request` In case of no reaction support, message out of reactions context or any other error
         + `404 Not Found` When the conversation or message to react could not be found for the participant
 
     - Data:
--- a/docs/reaction.md
+++ b/docs/reaction.md
@@ -5,7 +5,7 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
 ## React to a message
 
 * Method: `POST`
-* Endpoint: `/chat/{token}/{messageId}`
+* Endpoint: `/reaction/{token}/{messageId}`
 * Data:
 
     field | type | Description
--- a/docs/reaction.md
+++ b/docs/reaction.md
@@ -14,6 +14,7 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
 
 * Response:
     - Status code:
+        + `200 OK` Reaction already exists
         + `201 Created`
         + `400 Bad Request` In case of any other error
         + `404 Not Found` When the conversation or message to react could not be found for the participant
--- a/docs/reaction.md
+++ b/docs/reaction.md
@@ -38,7 +38,7 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
 ## Retrieve reactions of a message by type
 
 * Method: `GET`
-* Endpoint: `/chat/{token}/{messageId}/?reaction=`
+* Endpoint: `/reaction/{token}/{messageId}`
 * Data:
 
     field | type | Description
--- a/docs/reaction.md
+++ b/docs/reaction.md
@@ -18,3 +18,19 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
         + `400 Bad Request` In case of any other error
         + `404 Not Found` When the conversation or message to react could not be found for the participant
         + `409 Conflict` User already did this reaction to this message
+
+## Delete a reaction
+
+* Method: `DELETE`
+* Endpoint: `/chat/{token}/{messageId}`
+* Data:
+
+    field | type | Description
+    ---|---|---
+    `reaction` | string | the reaction emoji
+
+* Response:
+    - Status code:
+        + `201 Created`
+        + `400 Bad Request` In case of any other error
+        + `404 Not Found` When the conversation or message to react or reaction could not be found for the participant
--- a/docs/reaction.md
+++ b/docs/reaction.md
@@ -4,6 +4,7 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
 
 ## React to a message
 
+* Required capability: `reactions`
 * Method: `POST`
 * Endpoint: `/reaction/{token}/{messageId}`
 * Data:
@@ -22,6 +23,7 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
 
 ## Delete a reaction
 
+* Required capability: `reactions`
 * Method: `DELETE`
 * Endpoint: `/reaction/{token}/{messageId}`
 * Data:
@@ -38,6 +40,7 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
 
 ## Retrieve reactions of a message by type
 
+* Required capability: `reactions`
 * Method: `GET`
 * Endpoint: `/reaction/{token}/{messageId}`
 * Data:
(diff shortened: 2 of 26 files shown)


docs/chat.md
```diff
@@ -50,6 +50,7 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
         `message` | string | Message string with placeholders (see [Rich Object String](https://github.com/nextcloud/server/issues/1706))
         `messageParameters` | array | Message parameters for `message` (see [Rich Object String](https://github.com/nextcloud/server/issues/1706))
         `parent` | array | **Optional:** See `Parent data` below
+        `reactions` | array | **Optional:** An array map with relation between reaction emoji and total of reactions with this emoji
```
nickvergessen, line 53:
```suggestion
        `reactions` | array | **Optional:** An array map with relation between reaction emoji and total count of reactions with this emoji
```

docs/reaction.md
```diff
@@ -0,0 +1,61 @@
+# Reaction API
+
+Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
+
+## React to a message
+
+* Method: `POST`
+* Endpoint: `/chat/{token}/{messageId}`
```
nickvergessen, line 8: This is not what your route file says

```suggestion
* Endpoint: `/reaction/{token}/{messageId}`
```

docs/reaction.md
```diff
@@ -0,0 +1,61 @@
...
+    - Status code:
+        + `201 Created`
+        + `400 Bad Request` In case of any other error
+        + `404 Not Found` When the conversation or message to react could not be found for the participant
+        + `409 Conflict` User already did this reaction to this message
+
+## Delete a reaction
+
+* Method: `DELETE`
+* Endpoint: `/chat/{token}/{messageId}`
```
nickvergessen, line 25:
```suggestion
* Endpoint: `/reaction/{token}/{messageId}`
```

docs/reaction.md
```diff
@@ -0,0 +1,61 @@
...
+* Response:
+    - Status code:
+        + `201 Created`
+        + `400 Bad Request` In case of any other error
+        + `404 Not Found` When the conversation or message to react or reaction could not be found for the participant
+
+## Retrieve reactions of a message by type
+
+* Method: `GET`
+* Endpoint: `/chat/{token}/{messageId}/?reaction=`
```
nickvergessen, line 41:
```suggestion
* Endpoint: `/reaction/{token}/{messageId}`
```

lib/Chat/ChatManager.php
```diff
@@ -334,22 +334,6 @@ public function getParentComment(Room $chat, string $parentId): IComment {
...
-	 */
-	public function getComment(Room $chat, string $messageId): IComment {
-		$comment = $this->commentsManager->get($messageId);
-
-		if ($comment->getObjectType() !== 'chat' || $comment->getObjectId() !== (string) $chat->getId()) {
-			throw new NotFoundException('Message not found in the right context');
-		}
-
-		return $comment;
-	}
```
nickvergessen, line 351: This is a safety thing, please do not remove.

lib/Chat/CommentsManager.php
```diff
@@ -25,10 +25,28 @@
...
 use OCP\DB\QueryBuilder\IQueryBuilder;
 
 class CommentsManager extends Manager {
+	/**
+	 * @param Room $chat
+	 * @param string $messageId
+	 * @return IComment
+	 * @throws NotFoundException
+	 */
+	public function getComment(Room $chat, string $messageId): IComment {
```
nickvergessen, line 40: This class is only there so we can have things before they are merged upstream. Anything that is being added here should be upstreamed. This method is talk related and should therefore be in the ChatManager

lib/Controller/ReactionController.php
```diff
@@ -0,0 +1,146 @@
...
+	 *
+	 * @param int $messageId for reaction
+	 * @param string $reaction the reaction emoji
+	 * @return DataResponse
+	 */
+	public function react(int $messageId, string $reaction): DataResponse {
+		$participant = $this->getParticipant();
+		try {
+			// Verify if messageId is of room
+			$this->commentsManager->getComment($this->getRoom(), (string) $messageId);
```
nickvergessen, line 64: We also need to validate that the message is actually a reactionable message?
System messages should not be reactable? Neither deleted messages nor reactions themselves?

nickvergessen: Also as written above the method needs to be moved back to the chatManager and then this controller should depend on the ChatController, or even better you move all this to the reactionManager and just handle a the feedback of the reactionManager in the controller?

lib/Controller/ReactionController.php
```diff
@@ -0,0 +1,146 @@
...
+
+		try {
+			// Verify already reacted whith the same reaction
+			$this->commentsManager->getReactionComment(
+				$messageId,
+				$participant->getAttendee()->getActorType(),
+				$participant->getAttendee()->getActorId(),
+				$reaction
+			);
+			return new DataResponse([], Http::STATUS_CONFLICT);
```
nickvergessen, line 77:
```suggestion
			return new DataResponse([], Http::STATUS_OK);
```
I would do okay instead of conflict, because it is not a problem. Clients can still distinct the cases, but shouldn't show a message to users anyway

lib/Controller/ReactionController.php
```diff
@@ -0,0 +1,146 @@
...
+				$messageId,
+				$reaction
+			);
+		} catch (NotFoundException $e) {
+			return new DataResponse([], Http::STATUS_NOT_FOUND);
+		} catch (\Exception $e) {
+			return new DataResponse([], Http::STATUS_BAD_REQUEST);
+		}
+
+		return new DataResponse([], Http::STATUS_CREATED);
```
nickvergessen, line 121:
```suggestion
		return new DataResponse([], Http::STATUS_OK);
```

nickvergessen (changes requested):

lib/Controller/ChatController.php
```diff
@@ -141,6 +146,7 @@ public function __construct(string $appName,
 		$this->userId = $UserId;
 		$this->userManager = $userManager;
 		$this->appManager = $appManager;
+		$this->commentsManager = $commentsManager;
```
nickvergessen, line 149: unused?

lib/Chat/ReactionManager.php
```diff
@@ -0,0 +1,123 @@
...
+		$comment = $this->commentsManager->create(
+			$participant->getAttendee()->getActorType(),
+			$participant->getAttendee()->getActorId(),
+			'chat',
+			(string) $chat->getId()
+		);
+		$comment->setParentId((string) $messageId);
+		$comment->setMessage($reaction);
+		$comment->setVerb('reaction');
+		$this->commentsManager->save($comment);
```
nickvergessen, line 66: Since this is all bypassing all the chat manager code, there are no notifications, is that intended?
I think it's kind of important (at least in one-to-one conversations) to be notified so you know when people voted.

lib/Controller/ReactionController.php
```diff
@@ -0,0 +1,151 @@
...
+	 * @RequireModeratorOrNoLobby
+	 *
+	 * @param int $messageId for reaction
+	 * @param string $reaction the reaction emoji
+	 * @return DataResponse
+	 */
+	public function react(int $messageId, string $reaction): DataResponse {
+		$participant = $this->getParticipant();
+		try {
+			// Verify if messageId is of room
```
nickvergessen, line 68:
```suggestion
			// Verify that messageId is part of the room
```

lib/Controller/ReactionController.php
```diff
@@ -0,0 +1,151 @@
...
+			$this->reactionManager->getCommentToReact($this->getRoom(), (string) $messageId);
+		} catch (NotFoundException $e) {
+			return new DataResponse([], Http::STATUS_NOT_FOUND);
+		}
+
+		try {
+			// Verify already reacted whith the same reaction
```
nickvergessen, line 75:
```suggestion
			// Check if the user already reacted with the same reaction
```

lib/Controller/ReactionController.php
```diff
@@ -0,0 +1,151 @@
...
+	 * @RequireReadWriteConversation
+	 * @RequireModeratorOrNoLobby
+	 *
+	 * @param int $messageId for reaction
+	 * @param string|null $reaction the reaction emoji
+	 * @return DataResponse
+	 */
+	public function getReactions(int $messageId, ?string $reaction): DataResponse {
+		try {
+			// Verify if messageId is of room
```
nickvergessen, line 141: See above

nickvergessen (changes requested):

nickvergessen:
Server PR is in

lib/Chat/ReactionManager.php
nickvergessen: As per call let's notify users based on the "notify for every chat message" setting

docs/reaction.md
```diff
@@ -0,0 +1,62 @@
+# Reaction API
+
+Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`
+
+## React to a message
+
+* Method: `POST`
```
nickvergessen, line 7:
```suggestion
* Required capability: `reactions`
* Method: `POST`
```
Same for all other methods in this doc

docs/reaction.md
```diff
@@ -0,0 +1,62 @@
...
+        + `200 OK`
+        + `404 Not Found` When the conversation or message to react could not be found for the participant
+
+    - Data:
+        Array with data of reactions:
+
+        field | type | Description
+        ---|---|---
+        `actorType` | string | `guests` or `users`
+        `actorId` | string | User id of the reaction author
```
nickvergessen, line 60:
```suggestion
        `actorId` | string | Actor id of the reacting participant
```

lib/Controller/ReactionController.php
```diff
@@ -0,0 +1,138 @@
...
+	/** @var ReactionManager */
+	private $reactionManager;
+
+	public function __construct(string $appName,
+								IRequest $request,
+								CommentsManager $commentsManager,
+								ChatManager $chatManager,
+								ReactionManager $reactionManager) {
+		parent::__construct($appName, $request);
+		$this->commentsManager = $commentsManager;
```
nickvergessen, line 51: phpstorm says this is assigned but never used.
Drop it?

lib/Chat/ReactionManager.php
```diff
@@ -0,0 +1,172 @@
...
+
+		$this->notifyIfNecessary($chat, $comment, $parentMessage);
+		return $comment;
+	}
+
+	private function notifyIfNecessary(Room $chat, IComment $reactionInstance, IComment $parentMessage): void {
+		if ($chat->getType() !== Room::TYPE_ONE_TO_ONE) {
+			return;
+		}
```
nickvergessen, line 103: this was a misunderstanding.

We also want notifications, if users are:
* on `NOTIFY_ALWAYS` or
* on `NOTIFY_DEFAULT` and
    * room is `TYPE_ONE_TO_ONE` or
    * getDefaultGroupNotification is `NOTIFY_ALWAYS`

Basically similar to `OCA\Talk\Chat\Notifier::notifyOtherParticipant`

nickvergessen (changes requested):

nickvergessen (changes requested):
Pushed some commits to:
1. Reduce errors in log when running integration tests
2. Take notification preference into account
3. Render the notification

Please fix the tests accordingly

nickvergessen:
Another regression with this PR:

1. Create a room
2. Set notification level to "Always notify"
3. Post a message
4. Invite user 2
5. As user 2: Reply with a message to the admin message from 3\.
6. As admin: you have 2 notifications

This works fine on master

lib/Chat/ReactionManager.php
```diff
@@ -0,0 +1,145 @@
...
+				'timestamp' => $comment->getCreationDateTime()->getTimestamp(),
+				'reaction' => $comment->getMessage(),
+			];
+		}
+		return $reactions;
+	}
+
+	public function getCommentToReact(Room $chat, string $messageId): IComment {
+		if (!$this->commentsManager->supportReactions()) {
+			throw new NotFoundException('Reactions unsuported');
```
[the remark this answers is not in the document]
vitormattos: Now we have 3 scenarios that throw exception:

* Reactions unsupported
* Message not found
* Invalid message type to react

You want specific error code to each exception?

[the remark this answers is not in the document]
nickvergessen: > Reactions unsupported

400

> Message not found

404

> Invalid message type to react

400

nickvergessen:
Quoting @marcoambrosini from chat:

```
any reason why the reaction details endpoint doesn't return an object like in the message endpoint?

in the message endpoint the number of reactions is indexed "by emoji", like this:

reatcions: {
    ❤️ : 1 
}

While if I want to get the reactions details, the endpoint returns an array of object where each reaction is a separate object 

[{...}, {...}, {...}, ...]

----
My question is wouldn't it be better to have it like in the message endpoint? Meaning the reaction as a key and then an array of users that reacted, like:

{
    ❤️ : [{...}, {...} ...]
}
```

nickvergessen (approved):

PR #6762, opened 2022-01-11
6438: updated margin for rightbar in 1v1 channel

### LINKED ISSUE #6438: Add margin to the right of the chat the same size of the sidebar in 1on1s
to avoid the jump of the chat messages and `newMessageForm`

nikola-gladovic:
Fix #6438


src/components/RightSidebar/RightSidebar.vue
```diff
@@ -141,6 +142,13 @@ export default {
 		opened() {
 			return !!this.token && !this.isInLobby && this.show
 		},
+		showLargeBox() {
```
marcoambrosini, line 145: Maybe we can come up with a better name here? Like showSidebarPlaceholder or something similar.
Also there should be some documentation on top of this computed that describes what this does since it's a bit complicated to figure out

src/components/RightSidebar/RightSidebar.vue
```diff
@@ -232,8 +240,17 @@ export default {
...
+	mounted() {
+		this.resize()
+		window.addEventListener('resize', this.resize)
+	},
+	destroyed() {
+		window.removeEventListener('resize', this.resize)
+	},
 	methods: {
+		resize() {
+			this.largeBox = window.innerWidth > 1400
```
marcoambrosini, line 252: This evaluates as a boolean here we want a width, right? like `sidebarPlaceholderWidth`

marcoambrosini:
It feels weird because it should kick in only on much bigger screens, and never compress the chat

nikola-gladovic:
> It feels weird because it should kick in only on much bigger screens, and never compress the chat, newMessageForm nor TopBar

Could you please let me know about solution for it?

marcoambrosini (commented):
The logic I think  we should use is this https://github.com/nextcloud/spreed/issues/6438#issue-1042347393

nikola-gladovic:
I did rebase + force pushing for this PR. :)

marcoambrosini (commented):
This is the current state

https://user-images.githubusercontent.com/26852655/155706745-c4026586-5504-4212-be50-b6c607215f58.mov

Some points:
* Both the `TopBar` and the `NewMessageForm` components should span for the whole width of the window, and only their content should be centered with the `MessagesList``
* I think that, if anything, this should happen on very large screens only, I shouldn't get this on my laptop. cc @jancborchardt 
* The margin is applied in group conversation when the sidebar is collapsed, see video below

https://user-images.githubusercontent.com/26852655/155707342-35208c01-394b-426d-b3c5-3b71c4a11964.mov

nickvergessen:
> Both the TopBar and the NewMessageForm components should span for the whole width of the window, and only their content should be centered with the MessagesList

just to avoid misunderstandings. The message form should not jump to a different position when you change between a group and a one-to-one chat.

marcoambrosini:
> The message form should not jump to a different position when you change between a group and a one-to-one chat.

This is not true all the time @nickvergessen. Only on particularly big screens.

nikola-gladovic:
Could you please review updated version ?

marcoambrosini (commented):
This is still not working for me.
Apart from that, @jancborchardt at which widths do you think that this should kick in?

nickvergessen:
Closing for now.
Needs a somewhat different approach I think

PR #6763, opened 2022-01-11
6417: added search feature to forward modal

nikola-gladovic:


src/views/RoomSelector.vue
```diff
@@ -106,18 +113,24 @@ export default {
...
 				return room.type !== CONVERSATION.TYPE.CHANGELOG
 					&& (!this.currentRoom || this.currentRoom !== room.token)
 					&& (!this.showPostableOnly || room.readOnly === CONVERSATION.STATE.READ_WRITE)
 					&& room.objectType !== 'file'
 					&& room.objectType !== 'share:password'
 			})
+			if (!this.searchText) {
+				return roomsTemp
+			} else {
+				return roomsTemp.filter(x => x.name.includes(this.searchText))
```
nickvergessen, line 132:
```suggestion
				return roomsTemp.filter(x => x.displayName.includes(this.searchText))
```

Should be display name otherwise it's the user id for one-to-one conversations

nikola-gladovic: Ok, thanks

nickvergessen:
Pushed a commit to make the search case-insensitive

nickvergessen (approved):

Issue #6154, opened 2021-08-27
Following own video while being alone in a public call shows the "Copy link" button on top


nickvergessen:
## Steps to reproduce
1. Create a public conversation
2. Start a call
3. Follow your own video

### Expected behaviour
You see your own video

### Actual behaviour
"Copy link" button on top
![Bildschirmfoto von 2021-08-27 13-24-27](https://user-images.githubusercontent.com/213943/131120652-3fdf7dc8-dff0-4a5d-859c-beaa6c25286f.png)

## Talk app

**Talk app version:** master 44ad976b56198a3456a01fe261639ec6bb10cc7d

**Custom Signaling server configured:** yes

**Custom TURN server configured:** yes

**Custom STUN server configured:** no

## Browser

**Microphone available:** yes

**Camera available:** yes

**Operating system:** Ubuntu

**Browser name:** Firefox

**Browser version:** 92beta2

nickvergessen:
Okay, so the problematic z-index was added in:
https://github.com/nextcloud/spreed/pull/4641

Seems the problem is that the call view is overlapping the full view and you can see the background (with button) but not click it otherwise.

nickvergessen:
Seems fixed by https://github.com/nextcloud/spreed/pull/6768

PR #6768, opened 2022-01-12
6154: fixed copy link issue for public call

nikola-gladovic:
#6154


nickvergessen (approved):
Seems to work. I wonder why the z-index was there in first place?!

marcoambrosini (approved):

Issue #6607, opened 2021-11-25
Clicking a link scrolls to the latest message


juliushaertl:
## Steps to reproduce
1. Scroll back in chat history
2. Click a link

### Expected behaviour
The scroll position remains at the current position

### Actual behaviour
The scroll position jumps back to the most recent message

### Additional note

This does not happen when using the middle-mouse button to open in a new tab.

## Talk app

**Talk app version:** v13.0.0-rc.3 

## Browser

**Operating system:** Arch Linux

**Browser name:** Firefox

**Browser version:** 93

nickvergessen:
Fixed with https://github.com/nextcloud/spreed/pull/6769

PR #6769, opened 2022-01-12
6607: fixed scrolling issue to latest message

nikola-gladovic:
#6607 

I think it is unused code.

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -241,9 +241,6 @@ export default {
 		isWindowVisible(visible) {
 			if (visible) {
 				this.onWindowFocus()
-				// FIXME: the sidebar chat takes much longer to open, this is why we need a higher value here
-				// need to investigate why the sidebar takes that long to open and is not even animated
-				window.setTimeout(() => this.scrollToFocussedMessage(), 100)
 			}
 		},
 		chatIdentifier: {


marcoambrosini (commented):
Why unused? The content of that watcher runs every time the tab is focused in theory

marcoambrosini:
@PVince81 do you remember what was the reasoning behind scrolling to the first unread every time a tab is refocused btw?

PVince81:
> @PVince81 do you remember what was the reasoning behind scrolling to the first unread every time a tab is refocused btw?

not exactly.
probably was a UX thing so that when you return to the hidden tab you start at the first unread message
an improvement could be to add a link somewhere "jump to the first unread message"

marcoambrosini:
yes we could have that, so let's remove this then

src/components/MessagesList/MessagesList.vue
```diff
@@ -238,14 +238,14 @@ export default {
...
-				// need to investigate why the sidebar takes that long to open and is not even animated
-				window.setTimeout(() => this.scrollToFocussedMessage(), 100)
-			}
-		},
+		// isWindowVisible(visible) {
+		// 	if (visible) {
+		// 		this.onWindowFocus()
+		// 		// FIXME: the sidebar chat takes much longer to open, this is why we need a higher value here
+		// 		// need to investigate why the sidebar takes that long to open and is not even animated
+		// 		window.setTimeout(() => this.scrollToFocussedMessage(), 100)
```
marcoambrosini, line 246: This is the part we need to remove @nikola-gladovic

src/components/MessagesList/MessagesList.vue
```diff
@@ -238,14 +238,14 @@ export default {
...
-			if (visible) {
-				this.onWindowFocus()
-				// FIXME: the sidebar chat takes much longer to open, this is why we need a higher value here
-				// need to investigate why the sidebar takes that long to open and is not even animated
-				window.setTimeout(() => this.scrollToFocussedMessage(), 100)
-			}
-		},
+		// isWindowVisible(visible) {
+		// 	if (visible) {
+		// 		this.onWindowFocus()
```
marcoambrosini, line 243: This we should keep

src/components/MessagesList/MessagesList.vue
nikola-gladovic: Ok, sure

src/components/MessagesList/MessagesList.vue
nikola-gladovic: Ok

marcoambrosini (commented):
Can you rebase so that there is only one commit please? Looking good otherwise

marcoambrosini (approved):

Issue #5615, opened 2021-05-17
Message menu bugos on scrolling


Valdnet:
## Steps to reproduce

1. Open a conversation.
2. Write multiple messages.
3. Click on the Meatballs menu of the messages for some message.
4. Scroll through the messages with the mouse wheel.

The contents of the Meatballs menu appear in the upper left corner of the page.

When you click the mouse:

![obraz](https://user-images.githubusercontent.com/47037905/118473962-d79eda00-b70a-11eb-8305-048d75a78a90.png)

After scrolling the mouse wheel:
![obraz](https://user-images.githubusercontent.com/47037905/118474160-0cab2c80-b70b-11eb-817b-938c655e3148.png)

## Talk app

**Talk app version:** (see apps admin page: `/index.php/settings/apps`)
11.2.1

## Browser

**Browser name:** Firefox/Chrome/...
Firefox

**Browser version:** 50.1/55/...
88.0.1

## Server configuration

**Nextcloud Version:** (see admin page)
21.0.1

nickvergessen:
@marcoambrosini this feels quite weird and I can still reproduce it. Maybe you can have a look?

marcoambrosini:
How should we handle this? Is it ok if we close the menu upon scrolling?

nickvergessen:
Sounds brutal but better than a menu going wild.
But maybe you can first find out what the problem is? It sounds like a missing/wrong boundaries element or something alike.

szaimen:
Sounds like https://github.com/nextcloud/server/issues/23872 to me... (So needs to be fixed in the vue components probably)

nickvergessen:
Shouldn't happen anymore since quite a while

PR #6775, opened 2022-01-13
5615: fixed message menu error in scrolling

nikola-gladovic:
#5615

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -32,7 +32,7 @@ the main body of the message as well as a quote.
 		:data-next-message-id="nextMessageId"
 		:data-previous-message-id="previousMessageId"
 		class="message">
-		<div :class="{'hover': showActions && !isSystemMessage && !isDeletedMessage, 'system' : isSystemMessage}"
+		<div :class="{'normal-message-body': !isSystemMessage && !isDeletedMessage, 'system' : isSystemMessage}"
 			class="message-body"
 			@mouseover="handleMouseover"
 			@mouseleave="handleMouseleave">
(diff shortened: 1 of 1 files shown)


src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -119,9 +119,8 @@ the main body of the message as well as a quote.
 					<!-- Message Actions -->
 					<div
 						v-if="hasActions"
-						v-show="showActions"
 						class="message-body__main__right__actions"
-						:class="{ 'tall' : isTallEnough }">
+						:class="{ 'tall' : isTallEnough, 'action-show': showActions, 'action-hide': !showActions }">
```
marcoambrosini, line 123: This doesn't solve the problem as the whole block is hidden if showActions is false

nikola-gladovic: The root cause of this bug is that block is display none, so the menu can't find exactly position of that block.
So I changed some logic here, when showActions false, this block will not show in UI, but not display none. ;)

nickvergessen: I can confirm that it works.
On that note, when the menu is open, it would be okay to not hide the action menu icon.

marcoambrosini: I think that that's the way to go

marcoambrosini: Even better, the actions emits an `update:open` event, so we could have

`<Actions  @update:open="handleActionMenuUpdate" />`

and then in the methods

```
handleActionMenuUpdate(velue) {
    this.isActionMenuOpen = value //This value being stored in the data
}
```

nikola-gladovic: Fixed

nickvergessen: You forgot to push?

marcoambrosini: Seems like it

nickvergessen (approved):

nickvergessen:
Rebased, tested, works nicely

PR #6777, opened 2022-01-13
Don't update the participant list when only you joined

nickvergessen:

### DIFF
--- a/src/utils/signaling.js
+++ b/src/utils/signaling.js
@@ -1166,6 +1166,8 @@ Signaling.Standalone.prototype.processRoomEvent = function(data) {
 				// may now no longer exist.
 				leftUsers = Object.assign({}, this.joinedUsers)
 			}
+
+			let userListIsDirty = false
 			for (i = 0; i < joinedUsers.length; i++) {
 				this.joinedUsers[joinedUsers[i].sessionid] = true
 				delete leftUsers[joinedUsers[i].sessionid]
@@ -1176,6 +1178,8 @@ Signaling.Standalone.prototype.processRoomEvent = function(data) {
 						// as otherwise you get the warning for your own old session immediately
 						this.ownSessionJoined = true
 					}
+				} else {
+					userListIsDirty = true
 				}
 			}
 			leftUsers = Object.keys(leftUsers)
@@ -1184,10 +1188,16 @@ Signaling.Standalone.prototype.processRoomEvent = function(data) {
 
 				for (i = 0; i < leftUsers.length; i++) {
 					delete this.joinedUsers[leftUsers[i]]
+
+					if (!this.settings.userId || leftUsers[i].userid !== this.settings.userId) {
+						userListIsDirty = true
+					}
 				}
 			}
 			this._trigger('usersJoined', [joinedUsers])
-			this._trigger('participantListChanged')
+			if (userListIsDirty) {
+				this._trigger('participantListChanged')
+			}
 		}
 		break
 	case 'leave':


src/utils/signaling.js
```diff
@@ -1184,10 +1188,16 @@ Signaling.Standalone.prototype.processRoomEvent = function(data) {
 
 				for (i = 0; i < leftUsers.length; i++) {
 					delete this.joinedUsers[leftUsers[i]]
+
+					if (!this.settings.userId || joinedUsers[i].userid !== this.settings.userId) {
```
nickvergessen, line 1192:
```suggestion
					if (!this.settings.userId || leftUsers[i].userid !== this.settings.userId) {
```

danxuliu (commented):
When you receive the `join` event from the external signaling server after you joined a room I would still emit the `participantListChanged`, as the participant list indeed changed (participant list as the concept, not the Vue component), but with an extra parameter to know if it is a "self-join" (which could be just the list of joined users) so it can be ignored if not needed. That is, I would move the deciding logic from the signaling to the `getParticipants` mixin, as it is the one that fetches the participants when joining a conversation and thus the one that knows that it is not needed to fetch them again.

Nevertheless, the approach in this pull request (or the extra parameter) only prevents an extra request to the server after joining if no other user is in the room. If you join a room in which there are other users the external signaling server will send [a `join` event about your own participant to all the participants in the room (including you)](https://github.com/strukturag/nextcloud-spreed-signaling/blob/76ad24ca62b188a6ac3575790a92e499bc5b2fa0/room.go#L445) and [another `join` event sent just to you with all the other participants (that is, excluding you)](https://github.com/strukturag/nextcloud-spreed-signaling/blob/76ad24ca62b188a6ac3575790a92e499bc5b2fa0/hub.go#L1209). Although the events are published in that order they arrive in reverse order (the first event goes through the NAT and the second is sent directly, I guess it is what causes that), although I am not sure if that would always happen.

Due to this a different approach could be to pass the list of joined sessions in the `participantListChanged`* (with a parameter to know that they were added and not changed or removed) and check in the `getParticipants` mixin if all the joined sessions are already in the participants data to fetch it again or not. Although this might require to queue the checks if the participant data is being fetched to prevent deciding based on stalled data... (for example, if a participant leaves and quickly joins again with different parameters, like the status). So maybe it is better to improve all this at a later point and for now just remove the request for your own join :shrug:

*Or listen to `usersJoined`, `usersChanged` and `usersLeft` rather than `participantListChanged` when the external signaling server is used.

nickvergessen:
> So maybe it is better to improve all this at a later point and for now just remove the request for your own join shrug

Which is exactly what I was trying to do. Simplicity and still removing a refresh.

danxuliu (changes requested):
> Which is exactly what I was trying to do. Simplicity and still removing a refresh.

Fine by me :+1: Nevertheless, my comment about moving the logic from the signaling object to the `getParticipant` mixin still stands :-)

nickvergessen:
> my comment about moving the logic from the signaling object to the getParticipant mixin still stands :-)

Well the participant list didn't change, we updated it already with the response of our own join. But sure, let's continue emitting a false event and modify all listeners to ignore it instead of not emitting it

nickvergessen:
Because the other listeners also reload the room information I don't dare to touch them just yet. So ready for review

danxuliu (approved):
I thought that the `joinedUsers` array provided the Nextcloud session id, so what I meant was just calling `this._trigger('participantListChanged', { joined: joinedUsers })` from _signaling.js_ and then ignoring the update in _getParticipants.js_ if `participants?.joined?.find(joinedParticipant => joinedParticipant.nextcloudSessionId === this.participant.sessionId`. But `joinedUsers` only provides the signaling session ID, not the Nextcloud session ID, so the current participant can not be matched :shrug:

Due to this I agree to use just the original commit, so I dropped the second commit that emitted whether the participant list was dirty or not, as the event logic is still coupled to the _getParticipants_ mixin (and others) and moving the check does not provide a benefit; the other listeners of `participantListChanged` event (_FilesSidebarTabApp_, _PublicShareAuthSidebar_ and _PublicShareSidebar_) do not need to listen to the event when self-joined either.

Tested and works :+1:

fancycode:
>  But joinedUsers only provides the signaling session ID, not the Nextcloud session ID, so the current participant can not be matched

FYI, this will change with the next release of the signaling server: https://github.com/strukturag/nextcloud-spreed-signaling/pull/178

PR #6785, opened 2022-01-17
Create message buttons bar

### LINKED ISSUE #5486: Reply button overlap on two row messages
## Steps to reproduce
1. Type a message like in the screenshot and send it
2. Open the participants panel and conversations panel
3. Hover the reply button

### Expected behaviour
Buttons appear below the date

### Actual behaviour
Reply button overlaps the text

image

## Talk app

**Talk app version:** Observed on c.nc.com and locally on master (c5015f53d1faeb097ed9b251037954ab850779f0)

Screen size is 1920x1080 with "Force font DPI" set to 144 on my Plasma workspace.

@ma12-co I guess the button should move to another row ?

marcoambrosini:
Fix #5486

Adds a buttons bar for message actions in prep for message reactions

https://user-images.githubusercontent.com/26852655/149757685-6e12f719-648d-4108-b896-b47a354c1b81.mov

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.spec.js
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.spec.js
@@ -511,7 +511,7 @@ describe('Message.vue', () => {
 	})
 
 	describe('actions', () => {
-		const ACTIONS_SELECTOR = '.message-body__main__right__actions'
+		const ACTIONS_SELECTOR = '.message__buttons-bar'
 
 		beforeEach(() => {
 			store = new Vuex.Store(testStoreConfig)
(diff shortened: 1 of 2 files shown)


nickvergessen:
Since it's now only on hover/active, can we have it on the same place (bottom + right) all the time and not have an offset based on whether read marker ticks are shown and other things?

marcoambrosini:
Sound's good but it's always been hover/active

marcoambrosini:
Done, also fine tuned messages spacing

https://user-images.githubusercontent.com/26852655/150321685-d6fe4f8e-575d-4526-a0bc-18a0ca1acb91.mov

nikola-gladovic:
Looks good :) 
Just one confirmation about the buttons bar, do we have design data about this ?
Otherwise, may I suggest smaller size? ;)

marcoambrosini:
> Otherwise, may I suggest smaller size? ;)

Funny that you mention this, we would also like to make it a bit smaller but for now we cannot go below 44px diameter in buttons. But we will have this discussion soon in our design channel :)

nickvergessen (approved):

Issue #5094, opened 2021-02-06
& is displayed as &amp; in the Delete Conversation modal


jaller94:
## Steps to reproduce
1. Create a room with an & in the name.
2. Delete it.

### Expected behaviour
![Screenshot_2021-02-06_09-44-13](https://user-images.githubusercontent.com/10872136/107113547-e51d0c00-685f-11eb-9339-e505bfa03c38.png)

### Actual behaviour
![Screenshot_2021-02-06_09-43-50](https://user-images.githubusercontent.com/10872136/107113550-ea7a5680-685f-11eb-91d8-ce8c7bef744a.png)

## Talk app

**Talk app version:** 9.0.8

PVince81:
still happening on master

PVince81:
we could switch from `.confirm` to `.confirmHtml` but need to make sure to properly sanitize the conversation name

nickvergessen:
https://github.com/nextcloud/spreed/blob/master/src/components/LeftSidebar/ConversationsList/Conversation.vue#L321

or you adjust this similar to
https://github.com/nextcloud/spreed/blob/master/src/components/LeftSidebar/ConversationsList/Conversation.vue#L215-L222

See https://github.com/nextcloud/server/blob/master/core/src/OC/l10n.js#L88-L89

nickvergessen:
Fixed with https://github.com/nextcloud/spreed/pull/6791

PR #6791, opened 2022-01-18
5094: fixed special symbols for translation

nikola-gladovic:

#5094


marcoambrosini (approved):
👍

PR #6798, opened 2022-01-19
Don't use the avatar for the callview background, to save O(n^2) requests

nickvergessen:
WIP, Marco wants to get the average color from an existing image we might have loaded already.
Otherwise we go with this for 24

Before | After
---|---
Bildschirmfoto von 2022-01-19 12-27-34 | Bildschirmfoto von 2022-01-19 12-20-14


nickvergessen:
Approaching feature freeze, so let's go with this for now

marcoambrosini (approved):

PR #6812, opened 2022-01-21
Don't force signaling mode when HPB trial is used

### LINKED ISSUE #6811: "signaling_mode=internal" entry on oc_appconfig bypass signaling server config
## Steps to reproduce
1. Configure and deploy a HPB server
2. Configure Nextcloud talk to use this HPB => all is green
3. The HPB is not used at all

### Expected behaviour
The talk config page should say something is wrong

### Actual behaviour
All seem to be OK on the Talk config page, but when starting a call, there is no noticeable difference with a call without HPB. No network traffic can reach the HPB.

## Talk app

**Talk app version:** 13.0.1

**Custom Signaling server configured:** yes ! 

**Custom TURN server configured:** yes !

**Custom STUN server configured:** yes !

**Solution :** there was an (old ?) entry in the database on table oc_appconfig named signaling_mode, with the value "internal". This hidden entry bypass all the signaling server config without warning the administrator. I think it's related with this line of code in public function getSignalingMode of spreed Config.php : 

`$mode = $this->config->getAppValue('spreed', 'signaling_mode', null);`

But we couldn't find an interface to see or change the value of this entry in the database. We had to compare with other Nextcloud instances to see it (after weeks of incomprehensible failed tests :/ ) and simply remove the database entry.

nickvergessen:
Fix #6811

### DIFF
--- a/lib/BackgroundJob/CheckHostedSignalingServer.php
+++ b/lib/BackgroundJob/CheckHostedSignalingServer.php
@@ -93,7 +93,7 @@ class CheckHostedSignalingServer extends TimedJob {
 		if ($oldStatus !== $newStatus) {
 			if ($oldStatus === 'active') {
 				// remove signaling servers if account is not active anymore
-				$this->config->setAppValue('spreed', 'signaling_mode', 'internal');
+				$this->config->deleteAppValue('spreed', 'signaling_mode');
 				$this->config->deleteAppValue('spreed', 'signaling_servers');
 
 				$notificationSubject = 'removed';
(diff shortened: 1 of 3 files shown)


danxuliu (commented):
I guess https://github.com/nextcloud/spreed/blob/b6829acbaabb2e323fb0b0c12d0251485d2e8781/lib/Controller/HostedSignalingServerController.php#L124 should be removed too.

danxuliu (approved):

PR #6851, opened 2022-01-31
l10n: Add to translation

Valdnet:
Add text string to be translated and change to uppercase.


nickvergessen:
This is not a text string, it's the protocol. you also don't type `translated(HTTPS)://github.com` into your browser.

Valdnet:
@nickvergessen So where to change this text string?

obraz

I don't think it displays correctly (lowercase).

nickvergessen:
its a protocol. you don't translate it and we don't make it uppercase.

Issue #6697, opened 2021-12-13
Sort participants in the callview based on more attributes


nickvergessen:
### Current
FIFO

### Expected
* Permissions
* Joined with media

### Additional
To prevent constant resorting, we should add people at the front if the conversation is restricted and the new user has permissions and otherwise give it some delay after a resorting happened.

nikola-gladovic:
I think we have 3 media: mobile, video, audio. how to sort those ? which media is top priority ?

nickvergessen:
The most important bit would be to have people with any media/permission at the beginning

nickvergessen:
Rightsidebar fixed in https://github.com/nextcloud/spreed/pull/6872
Actual call view is a bit more complicated

nikola-gladovic:
@marcoambrosini 
Could you please explain about this ticket? 
What should we update more? :)

marcoambrosini:
@nickvergessen could you explain exactly how you'd want it done?

nickvergessen:
E.g. in the company call it's a bit weird when you have 3 pages of participants and people with permissions that could enable the mic/video are spread over all pages. It would make more sense if they are "at the beginning" so you have to paginate as little as possible. But then again it might also be okay to recommend the speaker view more

PR #6872, opened 2022-02-07
Better participants sorting in sidebar tab

nikola-gladovic:

### DIFF
--- a/src/components/RightSidebar/Participants/CurrentParticipants/CurrentParticipants.vue
+++ b/src/components/RightSidebar/Participants/CurrentParticipants/CurrentParticipants.vue
@@ -21,8 +21,8 @@
 
 <template>
 	<div>
-		<ParticipantsList v-if="participants.length"
-			:items="participants"
+		<ParticipantsList v-if="participantsList.length"
+			:items="participantsList"
 			:loading="!participantsInitialised" />
 		<Hint v-else :hint="t('spreed', 'No search results')" />
 	</div>
@@ -58,11 +58,7 @@ export default {
 			default: true,
 		},
 	},
-	data() {
-		return {
-			isCurrentUserIsModerator: false,
-		}
-	},
+
 	computed: {
 		token() {
 			return this.$store.getters.getToken()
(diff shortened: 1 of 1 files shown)


nickvergessen:
You are modifying the right sidebar. While I think it also makes sense to sort participants there, the ticket was especially about the callview, but not sure that is task that is easy. Biggest issue is that it needs lot of work in the non-vue-d webrtc lib parts and collections.

src/components/RightSidebar/Participants/CurrentParticipants/CurrentParticipants.vue
```diff
@@ -203,6 +203,14 @@ export default {
 				return moderator1 ? -1 : 1
 			}
 
+			if (p1inCall) {
+				return participant1.attendeePermissions < participant2.attendeePermissions ? 1 : -1
+			}
+
+			if (p1inCall && participant1.attendeePermissions === participant2.attendeePermissions) {
+				return participant1.inCall < participant2.inCall ? 1 : -1
+			}
```
nickvergessen, line 212: The inCall attribute is already sorted by in line ~179

I think I would do it all the time (independent from inCall or not)

```suggestion
			if (participant1.attendeePermissions !== participant2.attendeePermissions) {
				return participant1.attendeePermissions < participant2.attendeePermissions ? 1 : -1
			}
```

 but only for moderators (other users dont see permissions):
* So include the participant mixin
* check with code like this if the current user is a moderator:
https://github.com/nextcloud/spreed/blob/b06c7f24886e6773219aa60d93e27ee1bb9f59d8/src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue#L614-L616

nikola-gladovic: I think already edit in the correct place (right sidebar), but I need to update new logic for admin view, other guys will not change. right?

nickvergessen: > but I need to update new logic for admin view, other guys will not change. right?

yes

nikola-gladovic:
I updated this PR :)

marcoambrosini (commented):
@nikola-gladovic as Joas commented, this doesn't affect the call-view, so I removed that from the title and unlinked the issue from the pr. Could you please describe what your pr does and what needs to be tested? 

Btw I'm having guests before regular users in the sidebar tab, is that intended?

Screenshot 2022-02-25 at 12 10 36

nickvergessen:
> Btw I'm having guests before regular users in the sidebar tab, is that intended?

Also happens on master. seems it was dropped somewhere else by accident

nickvergessen:
Moved to new issue in https://github.com/nextcloud/spreed/issues/6930

nikola-gladovic:
Ok, thanks.
So, it will be related with #6930, not a #6697 . right?

nickvergessen:
> So, it will be related with https://github.com/nextcloud/spreed/issues/6930, not a https://github.com/nextcloud/spreed/issues/6697 . right?

yes

nikola-gladovic:
Done

PR #6892, opened 2022-02-10
Remove API documentation of standalone signaling server.

fancycode:
Probably makes more sense to maintain this in the repository of the signaling server where the API is defined and implemented (see https://github.com/strukturag/nextcloud-spreed-signaling/pull/194).

Online available at https://nextcloud-spreed-signaling.readthedocs.io/en/latest/


docs/index.md
```diff
@@ -18,7 +18,6 @@
 * [Chat API](chat.md)
 * [Webinar API](webinar.md)
 * [Internal Signaling API](internal-signaling.md)
-* [Standalone Signaling API](standalone-signaling-api-v1.md)
```
nickvergessen, line 21: Would something like a link to `https://nextcloud-spreed-signaling.readthedocs.io/en/latest/` work?
Otherwise I would keep the entry and replace the content of the `standalone-signaling-api-v1.md` with a link to https://nextcloud-spreed-signaling.readthedocs.io/en/latest/

fancycode: Good point, changed and squashed.

PR #6895, opened 2022-02-10
Handle generate the end_call_for_all activity intentionally

nickvergessen:

### DIFF
--- a/lib/Activity/Listener.php
+++ b/lib/Activity/Listener.php
@@ -25,6 +25,7 @@ namespace OCA\Talk\Activity;
 
 use OCA\Talk\Chat\ChatManager;
 use OCA\Talk\Events\AddParticipantsEvent;
+use OCA\Talk\Events\ModifyEveryoneEvent;
 use OCA\Talk\Events\ModifyParticipantEvent;
 use OCA\Talk\Events\ModifyRoomEvent;
 use OCA\Talk\Events\RoomEvent;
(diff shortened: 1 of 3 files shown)


lib/Activity/Listener.php
```diff
@@ -89,6 +90,11 @@ public static function register(IEventDispatcher $dispatcher): void {
 		$dispatcher->addListener(Room::EVENT_BEFORE_END_CALL_FOR_EVERYONE, $listener);
```
nickvergessen, line 90: it's already done in ^

danxuliu (commented):
Before this change there should be already just a single system message when ending the call for everyone. When [`generateCallActivity` is first called](https://github.com/nextcloud/spreed/blob/5ac20349c6ff3c16a26bbdd1ecb3d6e9c773b9db/lib/Activity/Listener.php#L87) by the handler of `EVENT_BEFORE_END_CALL_FOR_EVERYONE` [`Room::activeSince` will be reset](https://github.com/nextcloud/spreed/blob/5ac20349c6ff3c16a26bbdd1ecb3d6e9c773b9db/lib/Activity/Listener.php#L142). When `generateCallActivity` is then called after each participant leaves the call nothing will be done in `generateCallActivity`, as [`activeSince` will be null now](https://github.com/nextcloud/spreed/blob/5ac20349c6ff3c16a26bbdd1ecb3d6e9c773b9db/lib/Activity/Listener.php#L127).

Is this addresing something else? What needs to be tested?

nickvergessen:
No, that is actually correct I missed it that it uses the before event on my first observation.
SO yeah there is no performance impact here, just making it more explicit that it is not intended to be called for the users in case of the "end for all"

vitormattos (approved):

danxuliu (approved):
Code is fine :+1: 

I have added an integration test for the system messages when ending a call (even if nothing changed in this pull request in that regard I think that it is good to have an explicit test for that).

I have also added a fixup to reword the comment to something, I think, clearer. But if it is not feel free to drop it :-)

PR #6907, opened 2022-02-15
Add see-links to route files

nickvergessen:
Makes it a bit easier while debugging to jump from a known URL to the actual controller/code

### DIFF
--- a/appinfo/routes/routesRoomController.php
+++ b/appinfo/routes/routesRoomController.php
@@ -34,40 +34,71 @@ $requirementsWithToken = [
 
 return [
 	'ocs' => [
+		/** @see \OCA\Talk\Controller\RoomController::getRooms() */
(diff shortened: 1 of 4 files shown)


vitormattos (commented):
I tested locally, verified the settings and documentation about annotations of my IDE but don't worked to me. I think that is a problem on my IDE.

I checked the documentation of phpdoc about inline tag [@see](https://docs.phpdoc.org/guide/references/phpdoc/inline-tags/see.html) and looking it seems correct.

If work on your IDE, sounds a good idea.

And other routes files?

nickvergessen:
> And other routes files?

Didn't want to create conflicts with other PRs, so step by step

nickvergessen:
Bildschirmfoto von 2022-02-16 11-56-50

vitormattos:
I think is a bug on extension of VSCodium, `@see` only work to me with class name, without method, when I put the method don't work more.

vitormattos (approved):

PR #6912, opened 2022-02-17
Remove unused app config

vitormattos:


lib/Migration/Version13000Date20220217115327.php
```diff
@@ -0,0 +1,49 @@
...
+ *
+ */
+namespace OCA\Talk\Migration;
+
+use Closure;
+use OCP\IConfig;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version13000Date20220217115327 extends SimpleMigrationStep {
```
nickvergessen, line 32: 14000 because it's master?

Also a version change in info.xml is needed to trigger it

nickvergessen:
Also a version change in info.xml is needed to trigger it

nickvergessen (approved):

PR #6929, opened 2022-02-24
Wrap notification generation in a transaction

nickvergessen:
### Steps
1. Create 70 users:
```sh
i=10
while [ $i -ne 70 ]
do
        i=$(($i+1))
        echo "$i"

	USERID=$(uuidgen)

done
```
2. Create a conversation as admin with group "dummies"
3. Start a call
4. Measure time around the section of the diff:
```php
			// Add this before the begin of the transaction
			$start = microtime(true);
```
```php
			// Add this after the commit of the transaction
			$end = microtime(true);
			\OC::$server->getLogger()->error('notify: ' . ($end - $start));
```
5. Measure the time again with the beginTransaction and commit being commented out

With transaction | Without transaction
---|---
~0.11s | ~0.65s

This is at least backportable. I think in the end It would be better if the notification app would provide a way to to create a notification for multiple users in one call and then do the transaction itself

### DIFF
--- a/lib/Notification/Listener.php
+++ b/lib/Notification/Listener.php
@@ -31,6 +31,7 @@ use OCA\Talk\Room;
 use OCA\Talk\Service\ParticipantService;
 use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\EventDispatcher\IEventDispatcher;
+use OCP\IDBConnection;
 use OCP\Notification\IManager;
 use OCP\IUser;
 use OCP\IUserSession;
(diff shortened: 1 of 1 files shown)


lib/Notification/Listener.php
```diff
@@ -260,6 +266,7 @@ public function sendCallNotifications(Room $room): void {
 				$this->logger->error($e->getMessage(), ['exception' => $e]);
 			}
 		}
```
nickvergessen, line 268: need to wrap with a try catch and rollback in case of exception

vitormattos (approved):
Excellent results

PR #6941, opened 2022-02-26
Reset all in_call flags in one go

nickvergessen:
* Replaces O(n) queries with a single query in the "End meeting for all" case

### DIFF
--- a/lib/Model/SessionMapper.php
+++ b/lib/Model/SessionMapper.php
@@ -23,6 +23,7 @@ declare(strict_types=1);
 
 namespace OCA\Talk\Model;
 
+use OCA\Talk\Participant;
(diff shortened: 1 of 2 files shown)


danxuliu (changes requested):
When the HPB is used now the call is not properly ended; the `inCall` flag of the participants in the `participants->update` event sent by the external signaling server has its previous value rather than `0` (so the clients of other participants are not aware that they are no longer in the call).

nickvergessen:
Lemme retry with less users manually (I always ran into #6948 with my automated test)

vitormattos (approved):

vitormattos:
Only need solve the fixup and DCO tests before merge.

PR #6943, opened 2022-02-28
Combine multiple system messages updates

### LINKED ISSUE #6944: Creating multiple system messages in one request sends multiple requests to the HPB
## Steps to reproduce
1. Create a conversation
2. Setup HPB
3. Add a group with 70 users to the conversation

### Expected behaviour
1 request to the HPB so admin user reloads the chat and sees the system messages

### Actual behaviour
70 requests to the HPB happen

Ref https://github.com/nextcloud/spreed/pull/6943

The problem of origin is that the HPB listens to the `EVENT_AFTER_SYSTEM_MESSAGE_SEND` event.
That should be adjusted to honor `$shouldSkipLastMessageUpdate` and then the new update introduced in #6943 should trigger a final event for the HPB to listen to.

nickvergessen:
- [x] Only update last_message once, when creating multiple system chat messages
- [x] Combine the update of last message and last activity
- [x] Only push one update to the HPB

Fix #6944

### DIFF
--- a/lib/Chat/SystemMessage/Listener.php
+++ b/lib/Chat/SystemMessage/Listener.php
@@ -43,6 +43,7 @@ use OCA\Talk\Share\RoomShareProvider;
 use OCA\Talk\TalkSession;
 use OCA\Talk\Webinary;
 use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\Comments\IComment;
 use OCP\EventDispatcher\Event;
 use OCP\EventDispatcher\IEventDispatcher;
 use OCP\EventDispatcher\IEventListener;
@@ -263,7 +264,15 @@ class Listener implements IEventListener {
 					|| $listener->getUserId() !== $participant['actorId']
 					// - has joined a listable room on their own
 					|| $participantType === Participant::USER) {
-					$listener->sendSystemMessage($room, 'user_added', ['user' => $participant['actorId']]);
+					$comment = $listener->sendSystemMessage(
+						$room,
+						'user_added',
+						['user' => $participant['actorId']],
+						null,
+						$event->shouldSkipLastMessageUpdate()
+					);
+
+					$event->setLastMessage($comment);
 				}
 			}
 		});
@@ -379,7 +388,7 @@ class Listener implements IEventListener {
 		}
 	}
 
-	protected function sendSystemMessage(Room $room, string $message, array $parameters = [], Participant $participant = null): void {
+	protected function sendSystemMessage(Room $room, string $message, array $parameters = [], Participant $participant = null, bool $shouldSkipLastMessageUpdate = false): IComment {
 		if ($participant instanceof Participant) {
 			$actorType = $participant->getAttendee()->getActorType();
 			$actorId = $participant->getAttendee()->getActorId();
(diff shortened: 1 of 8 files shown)


lib/Service/ParticipantService.php
```diff
@@ -426,9 +433,25 @@ public function addUsers(Room $room, array $participants, ?IUser $addedBy = null
...
+			$lastMessage = $event->getLastMessage();
+			if ($lastMessage instanceof IComment) {
+				$this->updateRoomLastMessage($room, $lastMessage);
+			}
 		}
 	}
 
+	protected function updateRoomLastMessage(Room $room, IComment $message): void {
+		$room->setLastMessage($message);
+		$lastMessageCache = $this->cacheFactory->createDistributed('talk/lastmsgid');
```
vitormattos, line 446: Don't will me more readable?
```suggestion
		$lastMessageCache = $this->cacheFactory->createDistributed('talk/last_message_id');
```

lib/Service/ParticipantService.php
```diff
@@ -426,9 +433,25 @@ public function addUsers(Room $room, array $participants, ?IUser $addedBy = null
...
 			$this->dispatcher->dispatch(Room::EVENT_AFTER_USERS_ADD, $event);
+
+			$lastMessage = $event->getLastMessage();
+			if ($lastMessage instanceof IComment) {
+				$this->updateRoomLastMessage($room, $lastMessage);
+			}
 		}
 	}
 
+	protected function updateRoomLastMessage(Room $room, IComment $message): void {
```
vitormattos, line 444: Is this method only for update cache? If true, maybe will be good expose this at name of method
```suggestion
	protected function updateCacheOfRoomLastMessage(Room $room, IComment $message): void {
```

lib/Chat/SystemMessage/Listener.php
```diff
@@ -408,11 +417,13 @@ protected function sendSystemMessage(Room $room, string $message, array $paramet
 			$referenceId = (string) $referenceId;
 		}
 
-		$this->chatManager->addSystemMessage(
+		return $this->chatManager->addSystemMessage(
```
vitormattos, line 420: The method `addSystemMessage` haven't docblock explaining what `shouldSkipLastMessageUpdate` do.
The argument name explain what he does but, what is the impact of skip last message update?
Maybe will be good use docblock in `addSystemMessage` or also in `sendSystemMessage`

The `event->shouldSkipLastMessageUpdate` also haven't explanation about what he does.

lib/Service/ParticipantService.php
nickvergessen: No, this is already in use in other places and we need to purge the correct one

lib/Service/ParticipantService.php
nickvergessen: It's not only about the cache but also about the database

lib/Chat/SystemMessage/Listener.php
nickvergessen: > what is the impact of skip last message update?

It skips the update of the last message field in the database

vitormattos: What is the impact when we skip this update? When isn't necessary?

nickvergessen: Added

vitormattos (approved):

PR #6945, opened 2022-02-28
Only calculate the room name once instead of loading all participants

nickvergessen:
### Steps
1. Create 70 users:
```sh
i=10
while [ $i -ne 70 ]
do
        i=$(($i+1))
        echo "$i"

	USERID=$(uuidgen)

done
```
2. Create a conversation as admin with group "dummies" (and check the database queries of this request)

Left is before, right is with this patch
Bildschirmfoto von 2022-02-28 12-09-44

### DIFF
--- a/lib/Activity/Listener.php
+++ b/lib/Activity/Listener.php
@@ -232,6 +232,11 @@ class Listener {
 			return;
 		}
 
+		// We know the new participant is in the room,
+		// so skip loading them just to make sure they can read it.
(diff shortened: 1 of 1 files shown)


danxuliu (approved):
Code looks good based on the use of `userId` in [`resolveRoomDisplayName()`](https://github.com/nextcloud/spreed/blob/55121e46170af29d752eece2b99cc72bb8b6f1e2/lib/Manager.php#L978) (although I have not tested it).

Issue #6927, opened 2022-02-22
Vue buttons


nickvergessen:
Since we updated to `@nextcloud/vue` 5.0 with #6899

We should now get rid of all normal buttons and move to the Button component, similarly to https://github.com/nextcloud/spreed/pull/6834

nickvergessen:
cc @nikola-gladovic this could be something for you as you have to look into several places and can learn about their interactions.

PR #6950, opened 2022-03-01
6927: updated button for dashboard

nikola-gladovic:

#6927

### DIFF
--- a/src/assets/buttons.scss
+++ b/src/assets/buttons.scss
@@ -22,6 +22,11 @@
 
 @import 'variables';
 
+.button-vue--vue-default {
+	background-color: var(--color-background-dark);
+	color: var(--color-main-text);
+}
+
 .nc-button {
 	width: $clickable-area;
 	height: $clickable-area;
--- a/src/views/Dashboard.vue
+++ b/src/views/Dashboard.vue
@@ -45,9 +45,11 @@
 			<EmptyContent icon="icon-talk">
 				<template #desc>
 					{{ t('spreed', 'Say hi to your friends and colleagues!') }}
-					<button @click="clickStartNew">
+					<Button class="button-start-conversation"
+						type="default"
+						@click="clickStartNew">
 						{{ t('spreed', 'Start a conversation') }}
-					</button>
+					</Button>
 				</template>
 			</EmptyContent>
 		</template>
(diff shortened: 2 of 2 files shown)


src/assets/buttons.scss
```diff
@@ -22,6 +22,11 @@
 
 @import 'variables';
 
+.button-vue--vue-default {
+	background-color: var(--color-background-dark);
+	color: var(--color-main-text);
+}
```
nickvergessen, line 28: This looks wrong. If we need a mode of the button that does not exist yet, we should fix it in the vue library

nikola-gladovic: Should I wait the updated Vue library? ;)

marcoambrosini: No need, I would just use the secondary button here. Whenever you see that gray-ish button, you should replace with the current type="secondary" from the button component. 

Screenshot 2022-03-02 at 18 03 48

nikola-gladovic: Ok

src/views/Dashboard.vue
```diff
@@ -203,6 +206,7 @@ export default {
 </script>
 
 <style lang="scss" scoped>
+	@import '../assets/buttons';
```
nickvergessen, line 209: this is not necessary anymore. The button component takes care of all this.

```suggestion
```

src/views/Dashboard.vue
```diff
@@ -216,4 +220,8 @@ export default {
 			margin-bottom: 2vh;
 		}
 	}
+	.button-start-conversation {
```
nickvergessen, line 223: Linting warning:

223:2  ✖  Expected empty line before rule  rule-empty-line-before


```suggestion
	}

	.button-start-conversation {
```

nickvergessen:
Dark mode still has issues with `secondary`

Bildschirmfoto von 2022-03-04 10-31-26
Bildschirmfoto von 2022-03-04 10-32-02

But this needs fixing in the lib.
If we merge this, we have to make sure that the lib gets fixed before the final.

Your take @marcoambrosini

src/views/Dashboard.vue
nikola-gladovic: Ok

jancborchardt:
Yep @marcoambrosini we need to use primary-element and primary-element-light here, cause those have limiters to not be too light?

marcoambrosini:
I created an issue for this in the vue lib, but yes I think we can merge and fix this in the vue library soon

PR #6953, opened 2022-03-03
Return empty array when comment haven't reactions

vitormattos:
Integration tests changed to cover this scenario


nickvergessen (approved):
🙈

PR #7005, opened 2022-03-15
6927: updated button for allowed groups

nikola-gladovic:

#6927


marcoambrosini (approved):
Tested and works

PR #7011, opened 2022-03-16
Reduce integration tests queries

vitormattos:
Goals: Reduce the quantity of endpoints and consequently the quantity of queries.

Before:
* GET to check if group exists. One problem: this method do a query to get all users of group (unnecessary).
* POST to create if not exists

After:
* POST to create group. If group already exists, the endpoint will return this on the status, getting the same result.

Reduces of queries on integration tests:

Screenshot_20220316_124906

### DIFF
--- a/tests/integration/features/bootstrap/FeatureContext.php
+++ b/tests/integration/features/bootstrap/FeatureContext.php
@@ -2083,28 +2083,25 @@ class FeatureContext implements Context, SnippetAcceptingContext {
 	 * @param string $group
 	 */
 	public function assureGroupExists($group) {
-		$response = $this->groupExists($group);
-		if ($response->getStatusCode() !== 200) {
-			$this->createGroup($group);
-			$response = $this->groupExists($group);
-			$this->assertStatusCode($response, 200);
-		}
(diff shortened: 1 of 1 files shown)


nickvergessen:
Nice one.
But the user's are not returned on the API call, so this looks like a bug in the API endpoint. It should not get each and every single user when not returning them?

So can we create a report in server about this as well?

nickvergessen:
https://github.com/nextcloud/server/blob/719dbafd1339702a170f04ebbc4f20e80d45e8c9/apps/provisioning_api/lib/Controller/GroupsController.php#L135

Nevermind, the endpoint is deprecated and it seems to be expected.

nickvergessen (approved):

vitormattos:
Isn't a problem from API. The endpoint used before is very deprecated:

https://github.com/nextcloud/server/blob/master/apps/provisioning_api/lib/Controller/GroupsController.php#L133

vitormattos:
My comment come very late.

vitormattos:
@nickvergessen I can't merge because has broken tests

PR #7022, opened 2022-03-18
Add an endpoint for a media tab

nickvergessen:
- [x] Performance improvement in https://github.com/nextcloud/server/pull/31614
- [x] Write integration test
- [x] Add docs
- [x] Add capability

### DIFF
--- a/docs/chat.md
+++ b/docs/chat.md
@@ -163,7 +163,7 @@ See [OCP\RichObjectStrings\Definitions](https://github.com/nextcloud/server/blob
 
   field | type | Description
   ---|---|---
-  `offset` | int | Offset parameter to get another chunk of chat messages
+  `lastKnownMessageId` | int | Serves as an offset for the query. The lastKnownMessageId for the next page is available in the `X-Chat-Last-Given` header.
   `limit` | int | Number of chat messages with shares you want to get
 
 * Response:
@@ -172,6 +172,12 @@ See [OCP\RichObjectStrings\Definitions](https://github.com/nextcloud/server/blob
         + `200 OK`
         + `404 Not Found` When the conversation could not be found for the participant
         + `412 Precondition Failed` When the lobby is active and the user is not a moderator
+
+    - Header:
+
+      field | type | Description
+      ---|---|---
+      `X-Chat-Last-Given` | int | Offset (lastKnownMessageId) for the next page.
 
     - Data:
       Array of messages as defined in [Receive chat messages of a conversation](#receive-chat-messages-of-a-conversation)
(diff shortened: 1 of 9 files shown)


tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -1553,6 +1553,22 @@ public function userSeesTheFollowingMessagesInRoom($user, $identifier, $statusCo
...
 
+	/**
+	 * @Then /^user "([^"]*)" sees the following shared media in room "([^"]*)" with (\d+)(?: \((v1)\))?$/
+	 *
+	 * @param string $user
+	 * @param string $identifier
+	 * @param string $statusCode
+	 * @param string $apiVersion
+	 */
+	public function userSeesTheFollowingSharedMediaInRoom($user, $identifier, $statusCode, $apiVersion = 'v1', TableNode $formData = null) {
```
vitormattos, line 1564:
```suggestion
	public function userSeesTheFollowingSharedMediaInRoom($user, $identifier, $statusCode, $apiVersion = 'v1', TableNode $formData = null): void {
```

lib/Controller/ChatController.php
```diff
@@ -700,6 +700,37 @@ public function markUnread(): DataResponse {
...
+			$this->messageParser->parseMessage($message);
+
+			if (!$message->getVisibility()) {
+				continue;
+			}
+
+			$messages[] = $message->toArray();
+		}
+
+		return new DataResponse($messages);
```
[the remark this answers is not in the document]
nickvergessen: We don't know up front how many pages there will be. Also currently the response is the same as with normal chat messages (as they are just chat messages with a filter applied).

Also the subsequent comments API does not allow id based offset but only counter offset.
Let me see if we can bring that in, and then we add the same headers we use currently in the chat endpoint.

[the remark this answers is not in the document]
nickvergessen: Fixed

docs/chat.md
```diff
@@ -154,6 +154,28 @@ See [OCP\RichObjectStrings\Definitions](https://github.com/nextcloud/server/blob
...
+
+* Response:
+    - Note: if a file was shared multiple times it will be returned multiple times
+    - Status code:
+        + `200 OK`
+        + `404 Not Found` When the conversation could not be found for the participant
+        + `412 Precondition Failed` When the lobby is active and the user is not a moderator
+
+    - Data:
+      Array of messages as defined in [Receive chat messages of a conversation](#receive-chat-messages-of-a-conversation)
```
vitormattos, line 177: Changes:

* Add headers
* Fix indent

```suggestion

    - Header:
        field | type | Description
        ---|---|---
        `X-Chat-Last-Given` | int | Offset (lastKnownMessageId) for the next page.
        `X-Chat-Last-Common-Read` | int | ID of the last message read by every user that has read privacy set to public. When the user themself has it set to private the value the header is not set (only available with `chat-read-status` capability and when lastCommonReadId was sent)

    - Data:
        Array of messages as defined in [Receive chat messages of a conversation](#receive-chat-messages-of-a-conversation)
```

nickvergessen: added

vitormattos (approved):

PR #7024, opened 2022-03-18
👍 Reactions part two

marcoambrosini:


nickvergessen:
- [ ] Reaction option shows in read-only conversations like the Talk Update?

marcoambrosini:
- [ ] only allow reactions for message type comment (no system messages)

nickvergessen:
- [ ] Left sidebar shows reaction as last message, but does not happen when you reload while being in another chat or no chat at all.

nickvergessen (changes requested):
.DS_Store file exists

nickvergessen:
Tests spam an error:
```
    console.error node_modules/vuex/dist/vuex.common.js:878
      [vuex] duplicate getter key: userHasReacted
    console.debug src/components/MessagesList/MessagesGroup/Message/Message.vue:664
      adding reaction
    console.error node_modules/vuex/dist/vuex.common.js:878
      [vuex] duplicate getter key: userHasReacted
    console.debug src/components/MessagesList/MessagesGroup/Message/Message.vue:672
      user has already reacted, removing reaction
```

nickvergessen:
- [ ] Emoji list hides too soon
    I don't know, I think it's a papercut but I find it a bit sensitive:
    - Click on the emoji icon to add a reaction
    - See your mouse overlaying something
    - Move your mouse a bit down
    - Next message is hovered and therefor the pop up which you wanted to look at is closed

    I think similarly to the actions menu and the actual emoji picker, it would make sense to keep the state visible when you switched the the emoji list already.

- [ ] Half of the emojis are multiple characters and are not accepted by the reactions api which only accepts single character emojis atm. This was solved in the draft because we only had a short list of predefined emojis, but now this is disturbing.
Can also be fixed in a follow up.

nickvergessen (approved):

PR #7030, opened 2022-03-19
Fix the secondary buttons with the style of old buttons

nickvergessen:
### Before
Theme `#BA0013` | Before 
---|---
Dark | Bildschirmfoto von 2022-03-19 22-23-42
Bright | Bildschirmfoto von 2022-03-19 22-23-26

### After - Looking like a normal `<button>` in 23
Theme `#BA0013` | Normal | Hover
---|---|---
Dark | Bildschirmfoto von 2022-03-19 22-17-50 | Bildschirmfoto von 2022-03-19 22-17-53
Bright | Bildschirmfoto von 2022-03-19 22-21-00 | Bildschirmfoto von 2022-03-19 22-21-03

 
Maybe we can apply this in the vue lib even as a temp fix for 24? It would allow to migrate all buttons to the component already and later with an update we can make it shiny?

### DIFF
--- a/src/assets/buttons.scss
+++ b/src/assets/buttons.scss
@@ -54,5 +54,16 @@
 			background-color: var(--color-background-darker);
 		}
 	}
+}
 
+.button-vue--vue-secondary {
+	color: var(--color-main-text);
+	background-color: var(--color-background-dark);
+	border: 1px solid var(--color-border-dark);
+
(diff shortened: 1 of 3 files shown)


nickvergessen:
Obsoleted by https://github.com/nextcloud/nextcloud-vue/pull/2562

PR #7031, opened 2022-03-21
Group on PHP level as MariaDB can not handle this correctly

nickvergessen:
MySQL docs:
> Reject queries for which the select list, HAVING condition, or ORDER BY
list refer to nonaggregated columns that are neither named in the GROUP BY
clause **nor are functionally dependent on (uniquely determined by) GROUP BY columns**.

MariaDB docs:
> For SELECT ... GROUP BY queries, disallow SELECTing columns which are
not referred to in the GROUP BY clause, unless they are passed to an
aggregate function like COUNT() or MAX(). Produce a 1055 error.

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -103,6 +103,11 @@ jobs:
         options: --health-cmd="mysqladmin ping" --health-interval 5s --health-timeout 2s --health-retries 5
 
     steps:
+      - name: Enable ONLY_FULL_GROUP_BY MySQL option
+        run: |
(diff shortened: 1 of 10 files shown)


nickvergessen:
More problems found in https://github.com/nextcloud/spreed/pull/7032
And we need to port them to this backportable branch.
Due to the drone config change #7032 can not be backported automatically

nickvergessen:
Please review very soon @vitormattos

nickvergessen:
Backporting in #7036

PR #7047, opened 2022-03-24
Allow to delete shared files and objects

nickvergessen:
- [x] Implement on API
- [x] Add/adjust unit test
- [x] Add/adjust integration test
- [x] Implement in web UI

Remarks:
- Moderators and owner of the share will unshare the share completely, others can only remove the message
- Files currently always remain available for the owner as per long discussions in the past, because we can not know if the file is used in any other way in parallel.

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
@@ -304,14 +304,9 @@ export default {
 				return false
 			}
 
-			const isObjectShare = this.message === '{object}'
-				&& this.messageParameters?.object
-
 			return (moment(this.timestamp * 1000).add(6, 'h')) > moment()
-				&& this.messageType === 'comment'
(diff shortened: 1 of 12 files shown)


nickvergessen:
I guess it needs another capability for the mobile clients to do the same adjustment as the web did? cc @Ivansss

Ivansss:
Yep, it will be nice to have a capability for this :)

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
```diff
@@ -302,14 +302,9 @@ export default {
 				return false
 			}
 
-			const isObjectShare = this.message === '{object}'
-				&& this.messageParameters?.object
-
 			return (moment(this.timestamp * 1000).add(6, 'h')) > moment()
-				&& this.messageType === 'comment'
+				&& (this.messageType === 'comment' || this.messageType === 'voice-message')
```
marcoambrosini, line 306: Missing something here?

nickvergessen:
Rebased to solve conflict as I fixed something that was fixed in #7024

PR #7077, opened 2022-03-31
Improve API for media tab to be able to filter results

nickvergessen:


marcoambrosini (approved):

nickvergessen:
- [ ] Missing deleting of attachments when a message is deleted
- [ ] Missing deleting of attachments when a conversation is deleted

PR #7081, opened 2022-04-01
6697: Sort participants in the callview

### LINKED ISSUE #6697: Sort participants in the callview based on more attributes

nikola-gladovic:

Fix #6697

### DIFF
--- a/src/components/RightSidebar/Participants/CurrentParticipants/CurrentParticipants.vue
+++ b/src/components/RightSidebar/Participants/CurrentParticipants/CurrentParticipants.vue
@@ -219,8 +219,8 @@ export default {
 			}
 
 			if (this.currentParticipantIsModerator) {
-				if (participant1.attendeePermissions !== participant2.attendeePermissions) {
-					return participant1.attendeePermissions < participant2.attendeePermissions ? 1 : -1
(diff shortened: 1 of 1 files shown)


src/components/RightSidebar/Participants/CurrentParticipants/CurrentParticipants.vue
```diff
@@ -219,8 +219,8 @@ export default {
 			}
 
 			if (this.currentParticipantIsModerator) {
-				if (participant1.attendeePermissions !== participant2.attendeePermissions) {
-					return participant1.attendeePermissions < participant2.attendeePermissions ? 1 : -1
+				if (participant1.permissions !== participant2.permissions) {
```
nickvergessen, line 222: From my pov this is intentional so custom permissions are sorted over the users that fall back to default permissions

nickvergessen, line 222: From my pov this is intentional so custom permissions are sorted over the users that fall back to default permissions

PR #7114, opened 2022-04-07
Add constants for the verbs in the oc_comments table

### LINKED ISSUE #7078: Replace message verbs with constants
We use way too many messages verbs (rendered and in DB) so magic strings are not okay anymore.
we should define them as constants so they are all defined in one place and typos can be prevented.

nickvergessen:
Fix #7078


vitormattos (approved):
Good work

Issue #3299, opened 2020-04-08
Make people aware of unread mentions out of the viewport of the conversation list


tobiasKaminsky:
Having lots of favorited chats might lead to missing unread one, as the user has to scroll down a bit.
If app chooser a litte unread chat counter would be shown, the user can directly see that they have to scroll down a bit.

nickvergessen:
The biggest issue is people would then expect this is there all the time so we would need to get your room list on all pages, etc.
Huge performance break hitting in, so backlog for now

tobiasKaminsky:
True.
Maybe to make this more clear, we could show a 0 (or x for unread chats) when we are in talk and only the talk icon if we are in any other app.

Dubidubiduu:
I have a lot users only logging in for our monthly conference. Would be a big plus to inform them about unread messages. Since I prefer "sidemenu" over regular one I would prefer that hint on profile icon or so. Something like an additional custom field (e.x. Use Nextcloud Talk App for smartphone) and never miss a message is a big plus.

Also a "use smartphone app to never miss a message" overlay once a month to users who aren't using that, propably nc-app based, would help.

tobiasKaminsky:
I had this today again.
I have multiple 1:1 chats and one of those chats I delayed for ~1-2 days to answer it, but then all new chats were higher in list and thus the unread and important chat scrolled out of the visible list and thus I never answered it.
I just found it now by "accident".

nickvergessen:
Mabye we add a red dot similar to the notifications app one, but only when you are on talk? But that might be too confusing already again, as people might then think "Talk doesnt have the red dot, so no need to go there".

cc @nextcloud/designers-talk for input?

jancborchardt:
Let’s step back and look at the problem:

> Having lots of favorited chats might lead to missing unread one, as the user has to scroll down a bit.

The solution to that would be to show an indicator on the bottom of the conversation list. E.g. like this, similar to Twitter and Instagram do it with the "new unread posts" on the top of their view:
![image](https://user-images.githubusercontent.com/925062/88677366-795a2980-d0ed-11ea-9cbe-2ee3d074274b.png)

This doesn’t have the drawback of the global indicator either. :)

tobiasKaminsky:
Another, and maybe more simple solution would be to have sorting like this:
- favorites
- unread
- rest, sorted by date

This way one can always see their unread conversations (except if one has too many favorites and too small window :see_no_evil:)

[...]

PR #7115, opened 2022-04-07
3299: updated unread mentions viewport of conversation list

nikola-gladovic:

#3299

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -258,12 +258,14 @@ export default {
 		}, 30000)
 
 		EventBus.$on('should-refresh-conversations', this.debounceFetchConversations)
+		EventBus.$once('conversations-received', this.handleUnreadMention)
 
 		this.mountArrowNavigation()
 	},
 
 	beforeDestroy() {
 		EventBus.$off('should-refresh-conversations', this.debounceFetchConversations)
+		EventBus.$off('conversations-received', this.handleUnreadMention)
 
 		this.cancelSearchPossibleConversations()
 		this.cancelSearchPossibleConversations = null
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -32,6 +32,12 @@
 			<NewGroupConversation v-if="canStartConversations" />
 		</div>
 		<template #list>
+			<Button v-if="!preventFindingUnread && unreadTop > 0"
+				class="unread-mention-button unread-mention-button__top"
+				type="primary"
+				@click="scrollTopUnread">
+				{{ unreadTop }} more mentions
+			</Button>
 			<div ref="scroller"
 				class="left-sidebar__list"
 				@scroll="debounceHandleScroll">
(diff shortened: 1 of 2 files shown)


src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -32,6 +32,12 @@
 			<NewGroupConversation v-if="canStartConversations" />
 		</div>
 		<template #list>
+			<Button v-if="!preventFindingUnread && unreadTop > 0"
+				class="unread-mention-button unread-mention-button__top"
+				type="primary"
+				@click="scrollTopUnread">
+				{{ unreadTop }} more mentions
+			</Button>
```
nickvergessen, line 40: In the mobile apps we only show it for the bottom and I think that is okay.

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -91,14 +97,20 @@
 					<Hint v-else :hint="t('spreed', 'No search results')" />
 				</template>
 			</div>
+			<Button v-if="!preventFindingUnread && unreadBottom > 0"
+				class="unread-mention-button unread-mention-button__bottom"
+				type="primary"
+				@click="scrollBottomUnread">
+				{{ unreadBottom }} more mentions
```
nickvergessen, line 104: Needs to be translated.
Best is to do that in a computed:
```js
		unreadBottomLabel() {
			return n('spreed', '%n unread mention', '%n unread mentions', this.unreadBottom)
		},
```

n is for plurals, maybe you read https://docs.nextcloud.com/server/latest/developer_manual/basics/front-end/l10n.html#javascript quickly :)

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -276,7 +295,28 @@ export default {
...
+			})
+			setTimeout(() => {
+				this.handleUnreadMention()
+				this.preventFindingUnread = false
+			}, 500)
+		},
+		scrollBottomUnread() {
+			this.preventFindingUnread = true
+			this.$refs.scroller.scrollTo({
+				top: this.lastUnreadPos - this.$refs.scroller.clientHeight + 150,
```
nickvergessen, line 312: in the mobile apps we always scroll to the first unread mention outside of the viewport, so would be nice to do that here aswell. After clicking it then still shows the button if there are more below the viewport, so that is fine.

src/components/LeftSidebar/LeftSidebar.vue
nickvergessen: After checking the mobile apps, we don't show a count there, so if that makes it a bit more performant or something we can also skip it and just go with:

```suggestion
				{{ t('spreed', 'Unread mentions') }}
```

nickvergessen (changes requested):
Few comments but looks and works good otherwise

src/components/LeftSidebar/LeftSidebar.vue
nikola-gladovic: Ok, for now,  desktop has 2 buttons, and mobile has 1 button :)

nickvergessen (changes requested):

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -437,6 +458,10 @@ export default {
 					singleConversation: false,
 				})
 				this.isFetchingConversations = false
+				this.handleUnreadMention()
+				if (!this.$route.params.token) {
+					this.scrollBottomUnread()
```
nickvergessen, line 463: SHould not do this

nikola-gladovic: Currently, when you visit page, it will scroll down to unread mentions.
If we remove those code, it will not auto scroll anymore.
So which is the right behavior ?

nickvergessen: Yeah no scrolling is expected. We just show the button in that case.

nickvergessen (approved):

PR #7135, opened 2022-04-11
Expose an attribute "self" in the reaction summary with the list of o…

nickvergessen:
…wn reactions

cc @Ivansss

### DIFF
--- a/lib/Chat/ReactionManager.php
+++ b/lib/Chat/ReactionManager.php
@@ -175,7 +175,7 @@ class ReactionManager {
 	 * @return array[]
 	 * @psalm-return array<int, string[]>
 	 */
-	public function getReactionsForMessages(Participant $participant, array $messageIds): array {
+	public function getReactionsByActorForMessages(Participant $participant, array $messageIds): array {
 		return $this->commentsManager->retrieveReactionsByActor(
 			$participant->getAttendee()->getActorType(),
 			$participant->getAttendee()->getActorId(),
--- a/lib/Chat/ReactionManager.php
+++ b/lib/Chat/ReactionManager.php
@@ -170,6 +170,20 @@ class ReactionManager {
 	}
 
 	/**
+	 * @param Participant $participant
+	 * @param array $messageIds
+	 * @return array[]
+	 * @psalm-return array<int, string[]>
+	 */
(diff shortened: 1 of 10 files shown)


lib/Controller/ChatController.php
```diff
@@ -516,6 +501,45 @@ public function receiveMessages(int $lookIntoFuture,
...
+			// Add the self part also to potential parent elements
+			if (isset($parentMap[$messageId])) {
+				foreach ($parentMap[$messageId] as $mid) {
+					if (isset($messages[$commentIdToIndex[$mid]])) {
+						$messages[$commentIdToIndex[$mid]]['parent']['reactions']['self'] = $reactions;
+					}
+				}
+			}
+		}
+
```
vitormattos, line 542: Will be good put this in a method and reduce the responsibility of controller. Maybe, a good place to put the new method is in ReactionManager.

args: $message, $messages
return: $messages

lib/Chat/ReactionManager.php
```diff
@@ -169,6 +169,20 @@ public function retrieveReactionMessages(Room $chat, Participant $participant, i
 		return $reactions;
 	}
 
+	/**
+	 * @param Participant $participant
+	 * @param array $messageIds
+	 * @return array[]
+	 * @psalm-return array<int, string[]>
+	 */
+	public function getReactionsForMessages(Participant $participant, array $messageIds): array {
```
vitormattos, line 178:
```suggestion
	public function getReactionsByActorForMessages(Participant $participant, array $messageIds): array {
```

vitormattos (changes requested):

lib/Controller/ChatController.php
nickvergessen: I dislike this pattern of modifying local content in submethods in 3rd level code.
I'm fine with moving it to a method in the controller, but it's too unique to this place (like the parent loading just before) to put it somewhere else.

vitormattos: Move is good to make more readable the code and understood the main purpose of methods.
I suggested to move to ReactionManager only to maintain reactions code in the same place, but, no problem to maintain in a private method of ChatController

vitormattos (approved):

Issue #7152, opened 2022-04-14
Move anonymous functions in listeners to static methods


vitormattos:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

**Problem description**
Listeners methods actually add new listeners functions using anonymous functions. This approach generate a big register methods and hard codes to test.

**Solution description**
Move anonymous functions to public methods in listeners class.

After moving all to static methods, will be good cover the new methods with unit tests.

**Additional context**
Example of solution: https://github.com/nextcloud/spreed/commit/b78ce0709a8493d61746b6690b43f07b5f564d3b

Affected classes:
- [x] lib/Chat/Changelog/Listener.php https://github.com/nextcloud/spreed/pull/7170
- [x] lib/Chat/Command/Listener.php https://github.com/nextcloud/spreed/pull/7172
- [x] lib/Chat/Parser/Listener.php https://github.com/nextcloud/spreed/pull/7166
- [x] lib/Chat/SystemMessage/Listener.php https://github.com/nextcloud/spreed/pull/7183
- [x] lib/Collaboration/Collaborators/Listener.php https://github.com/nextcloud/spreed/pull/7171
- [x] lib/Listener/RestrictStartingCalls.php https://github.com/nextcloud/spreed/pull/7215
- [x] lib/Signaling/Listener.php https://github.com/nextcloud/spreed/pull/7213
- [x] lib/Status/Listener.php https://github.com/nextcloud/spreed/pull/7214

PR #7183, opened 2022-04-20
Move system message listeners to static methods

vitormattos:
related: https://github.com/nextcloud/spreed/issues/7152

### DIFF
--- a/lib/Chat/SystemMessage/Listener.php
+++ b/lib/Chat/SystemMessage/Listener.php
@@ -77,281 +77,303 @@ class Listener implements IEventListener {
 	}
 
 	public static function register(IEventDispatcher $dispatcher): void {
-		$dispatcher->addListener(Room::EVENT_BEFORE_SESSION_JOIN_CALL, static function (ModifyParticipantEvent $event) {
-			$room = $event->getRoom();
-			/** @var self $listener */
-			$listener = \OC::$server->get(self::class);
-			/** @var ParticipantService $participantService */
-			$participantService = \OC::$server->get(ParticipantService::class);
+		$dispatcher->addListener(Room::EVENT_BEFORE_SESSION_JOIN_CALL, self::class . '::sendSystemMessageAboutBeginOfCall');
+		$dispatcher->addListener(Room::EVENT_AFTER_SESSION_LEAVE_CALL, self::class . '::sendSystemMessageAboutCallLeft');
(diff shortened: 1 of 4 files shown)


lib/Chat/SystemMessage/Listener.php
```diff
@@ -77,281 +77,303 @@ public function __construct(IRequest $request,
...
+			$listener->sendSystemMessage($room, 'lobby_non_moderators');
+		}
+	}
 
-		$dispatcher->addListener(Room::EVENT_AFTER_USERS_ADD, static function (AddParticipantsEvent $event) {
-			$room = $event->getRoom();
-			if ($room->getType() === Room::TYPE_ONE_TO_ONE) {
-				return;
-			}
+	public static function addXJoinedTheConversationWeneverUserId(AddParticipantsEvent $event): void {
```
nickvergessen, line 254: Typo:
```suggestion
	public static function addXJoinedTheConversationWheneverUserId(AddParticipantsEvent $event): void {
```
But actually lets go simple:
```suggestion
	public static function addSystemMessageUserAdded(AddParticipantsEvent $event): void {
```

lib/Share/RoomShareProvider.php
```diff
@@ -155,7 +155,7 @@ public function create(IShare $share): IShare {
 		foreach ($existingShares as $existingShare) {
 			if ($existingShare->getSharedWith() === $share->getSharedWith()) {
 				// FIXME Should be moved away from GenericEvent as soon as OCP\Share20\IManager did move too
-				$this->dispatcher->dispatch(self::EVENT_SHARE_FILE_AGAIN, new GenericEvent($existingShare));
+				$this->dispatcher->dispatch(new GenericEvent($existingShare), self::EVENT_SHARE_FILE_AGAIN);
```
vitormattos, line 158: @nickvergessen I think this solved a bug.

nickvergessen: https://github.com/nextcloud/server/blob/215aef3cbdc1963be1bb6bca5218ee0a4b7f1665/lib/private/EventDispatcher/EventDispatcher.php#L86-L87

I think it explodes :P

vitormattos: Crazy.
Psalm point to this file:
https://github.com/symfony/event-dispatcher-contracts/blob/main/EventDispatcherInterface.php#L32

I changed now the property to `OCP\EventDispatcher\IEventDispatcher` and solved the psalm issue.

nickvergessen (approved):

PR #7186, opened 2022-04-20
Fix access to undefined key

vitormattos:


lib/Controller/ChatController.php
```diff
@@ -554,7 +554,7 @@ protected function loadSelfReactions(array $messages, array $commentIdToIndex):
 
 		// Inject the reactions self into the $messages array
 		foreach ($reactionsById as $messageId => $reactions) {
-			if (isset($messages[$commentIdToIndex[$messageId]])) {
+			if (isset($commentIdToIndex[$messageId]) && isset($messages[$commentIdToIndex[$messageId]])) {
```
nickvergessen, line 557:
```suggestion
			if (isset($commentIdToIndex[$messageId], $messages[$commentIdToIndex[$messageId]])) {
```

nickvergessen: can you also check how this can happen? because from my assumption all comments should have an index assigned when they end up in this array?

nickvergessen (approved):

PR #7218, opened 2022-04-25
Enable helper app with force

nickvergessen:
This allows to not bump the version in the helper app anymore.
Will save me a roundtrip of CI every time we branch off as I always forget it.

Also notifications and guests server requirement are of no interest to us and instead of having troubles during the transition period with red CI we just force enable them too.

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -25,8 +25,7 @@ local Pipeline(test_set, database, services) = {
 				"bash ./before_install.sh $APP_NAME $CORE_BRANCH $DATABASEHOST",
 				"cd ../server",
 				"./occ app:enable $APP_NAME",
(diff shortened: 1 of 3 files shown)


nickvergessen:
Just for the record, the only thing `--force` does is ignoring the nextcloud version requirement.

nickvergessen:
@danxuliu thoughts?

danxuliu (commented):
For the helper app I guess it is fine. For the other apps I am not a fan, as it could lead to forgetting to set the right branch on the stableXX branches due to forcing the installation of the master branch. Also in the master branch, if the apps are not actually compatible yet, we could still have a red CI, but with stranger errors rather than a failure when enabling the app.

PR #7307, opened 2022-05-04
Allow to "Send call notifications"

nickvergessen:

### DIFF
--- a/docs/call.md
+++ b/docs/call.md
@@ -47,6 +47,28 @@
         + `404 Not Found` When the user did not join the conversation before
         + `412 Precondition Failed` When the lobby is active and the user is not a moderator
 
+## Send call notification
+
+* Required capability: `send-call-notification`
+* Method: `POST`
+* Endpoint: `/call/{token}/ring/{attendeeId}`
+* Data:
+
+    field | type | Description
+    ---|---|---
+    `attendeeId` | int | The participant to notify
+
+* Response:
+    - Status code:
+        + `200 OK`
+        + `400 Bad Request` When the target participant is not a user (Guest, group, etc.)
+        + `400 Bad Request` When the target participant is already in the call
+        + `400 Bad Request` When the room has no call in process
+        + `400 Bad Request` When the actor is not in the call
+        + `403 Forbidden` When the current user is not a moderator
+        + `404 Not Found` When the conversation could not be found for the participant
+        + `412 Precondition Failed` When the lobby is active and the user is not a moderator
+
 ## Update call flags
 
 * Method: `PUT`
--- a/lib/Controller/CallController.php
+++ b/lib/Controller/CallController.php
@@ -137,6 +137,30 @@ class CallController extends AEnvironmentAwareController {
 	 * @PublicPage
 	 * @RequireCallEnabled
 	 * @RequireParticipant
+	 * @RequirePermissions(permissions=call-start)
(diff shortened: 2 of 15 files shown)


lib/Service/ParticipantService.php
```diff
@@ -988,6 +989,28 @@ public function changeInCall(Room $room, Participant $participant, int $flags, b
 		}
 	}
 
+	public function refreshCallNotificationForAttendee(Room $room, Participant $currentParticipant, int $targetAttendeeId) {
```
vitormattos, line 992:
```suggestion
	public function refreshCallNotificationForAttendee(Room $room, Participant $currentParticipant, int $targetAttendeeId): boolean {
```

vitormattos (changes requested):

lib/Controller/CallController.php
```diff
@@ -127,6 +127,29 @@ public function joinCall(?int $flags = null, ?int $forcePermissions = null): Dat
...
+	/**
+	 * @PublicPage
+	 * @RequireModeratorParticipant
+	 *
+	 * @param int $attendeeId
+	 * @return DataResponse
+	 */
+	public function ringAttendee(int $attendeeId): DataResponse {
+		if ($this->room->getCallFlag() === Participant::FLAG_DISCONNECTED) {
+			return new DataResponse([], Http::STATUS_BAD_REQUEST);
```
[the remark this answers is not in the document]
nickvergessen: I don't think it matters much, as in all the cases the UI option should not be displayed anymore. It can basically only happen in parallel executions when the UI is going to hide the option anyway soon.

lib/Controller/CallController.php
```diff
@@ -131,6 +131,29 @@ public function joinCall(?int $flags = null, ?int $forcePermissions = null): Dat
 		return new DataResponse();
 	}
 
+	/**
+	 * @PublicPage
+	 * @RequireModeratorParticipant
```
nickvergessen, line 136: Should be possible with Start-call permissions

docs/call.md
```diff
@@ -47,6 +47,28 @@
         + `404 Not Found` When the user did not join the conversation before
         + `412 Precondition Failed` When the lobby is active and the user is not a moderator
 
+## Resend call notification
```
marcoambrosini, line 50: Resend -> Send

nickvergessen: Done

lib/Controller/CallController.php
nickvergessen: Implemented

lib/Notification/Listener.php
```diff
@@ -288,4 +291,30 @@ public function markCallNotificationsRead(Room $room): void {
...
+		if ($event instanceof SendCallNotificationEvent) {
+			$this->sendCallNotification($event->getRoom(), $event->getActor()->getAttendee(), $event->getTarget()->getAttendee());
+		}
+	}
+
+	public function sendCallNotification(Room $room, Attendee $actor, Attendee $target): void {
+		try {
+			// Remove previous call notifications
+			$notification = $this->notificationManager->createNotification();
+			$notification->setApp('spreed')
```
vitormattos, line 305:
```suggestion
			$notification->setApp(Application::APP_ID)
```

vitormattos (approved):
got conflicts.

PR #7308, opened 2022-05-04
TS!

marcoambrosini:
Typescript_logo_2020 svg


src/types/vendor.d.ts
```diff
@@ -0,0 +1,4 @@
+declare function t(app: string, string: string, vars?: { [key: string]: string }, count?: number, options?: EscapeOptions): string;
+
+declare module 'vue-material-design-icons/VideoOff'
+declare module 'vue-material-design-icons/MenuDown'
```
nickvergessen, line 4: Here it is without `.vue` in the next file it's with?

src/types/vendor.d.ts
```diff
@@ -0,0 +1,4 @@
+declare function t(app: string, string: string, vars?: { [key: string]: string }, count?: number, options?: EscapeOptions): string;
+
+declare module 'vue-material-design-icons/VideoOff'
```
juliushaertl, line 3: I think it makes sense to split this file into one for each dependency, and maybe one Nextcloud.d.ts for the core ones.

nickvergessen (approved):
`¯\_(ツ)_/¯`

PR #7322, opened 2022-05-05
6389: updated the list of device name of too narrow style

### LINKED ISSUE #6389: List of device names is too narrow and might trim away important parts
## Steps to reproduce
1. Use firefox so devices have weird names
2. Start a call and open the device checker
3. Try to select a device

### Expected behaviour
Device names are fully visible

### Actual behaviour
Parts are cut off

Bildschirmfoto von 2021-10-22 16-14-12

nikola-gladovic:

Fixed #6389

### DIFF
--- a/src/components/DeviceChecker/DeviceChecker.vue
+++ b/src/components/DeviceChecker/DeviceChecker.vue
@@ -362,6 +362,7 @@ export default {
 @include avatar-mixin(128px);
 
 .device-checker {
+	width: 100%;
 	padding: 20px;
 	background-color: var(--color-main-background);
 	overflow-y: auto;
@@ -437,5 +438,6 @@ export default {
 
 ::v-deep .modal-container {
 	display: flex !important;
+	width: 750px !important;
 }
 </style>


src/components/DeviceChecker/DeviceChecker.vue
```diff
@@ -437,5 +438,6 @@ export default {
 
 ::v-deep .modal-container {
 	display: flex !important;
+	width: 750px !important;
```
nickvergessen, line 441: instead of making the container very wide which again shows weird grey space around the video could we show a tooltip with the full name of the device on hover?

nickvergessen, line 441: instead of making the container very wide which again shows weird grey space around the video could we show a tooltip with the full name of the device on hover?

nikola-gladovic: Regarding this ticket, I think we need to customize from the 3rd library.
https://nextcloud-vue-components.netlify.app/#/Components/Multiselect

marcoambrosini (commented):
Maybe for now we can just make the modal size `normal` @nikola-gladovic, without altering the component's css with v-deep

nickvergessen:
Closing for now.
Needs a somewhat different approach I think

PR #7323, opened 2022-05-05
6889: reduced the action menu

### LINKED ISSUE #6889: When in a call, reduce action menus from 2 to 1
### How to use GitHub

---

When in a call, there are 2 menus, the "More actions" 3-dot icon and the settings gearwheel. While the logic is that "More actions" only shows the actions relevant for inside the call, it creates confusion as to where to find what:
Current in call menu

It would be no issue to combine both menus:
Combined menu
(Note that "Rename conversation" should be done via conversation settings and doesn’t need to be in that action menu.)

nikola-gladovic:

Fixed #6889


nickvergessen:
Works in a call, but when you are not in a call the menu is not shown anymore.

Also I think the menu should have the ⚙️  Icon instead of the ... icon?

nikola-gladovic:
It means that we must show the menu bar, when we are not in a call ?

nickvergessen:
I think that was the idea, yes

Issue #6235, opened 2021-09-16
Backend long term tasks


nickvergessen:
This is a list of long term tasks. We might not work on them actively for now, but new code should respect the tasks and not repeat old behaviour:

* `OCA\Talk\Room` should be an ordinary `OCP\AppFramework\Db\Entity` class in the `OCA\Talk\Model` namespace
* Database interacting functions from the Room class should be moved/migrated to the `OCA\Talk\Service\RoomService`
* `$this->dispatcher->dispatch()` is deprecated, so we need to move to dedicated events

PR #7327, opened 2022-05-06
Implement time to live

vitormattos:
Ref #7359

To implement
- [x] Documentation
- [x] Capability
- [x] Routes
- [x] Integration tests
- [x] Controller
- [x] Changes on Service
- [x] Run event before and after define the TTL
- [x] Job to trigger delete message
~- [ ] Run event before and after trigger the job~
- [x] Remove TTL capability if the systemcron is disabled
- [x] Notify clients by system message with the ids of deleted messages to make possible clients remove the messages from message list without wait for page reload.
- [x] Delete orphans reactions.
    Scenario:
      set ttl to 10 min, 10:30 am
      send message X, 10:35 am
      react to message X in , 10:45 am
      apply expire TTL job at 10:46 am
      the message X will be deleted and the orphan reaction also need be deleted
~- [ ] Delete shared files and unshare files~ replaced by https://github.com/nextcloud/spreed/issues/7508
~- [ ] Unit tests (if necessary)~ replaced by https://github.com/nextcloud/spreed/issues/7508


docs/conversation.md
```diff
@@ -330,6 +331,36 @@
         + `401 Unauthorized` When the participant is a guest
         + `404 Not Found` When the conversation could not be found for the participant
 
+## Set time to live of messages for a conversation
```
nickvergessen, line 334:
```suggestion
## Set time to live for messages in a conversation
```

docs/conversation.md
```diff
@@ -330,6 +331,36 @@
...
+
+* Required capability: `time-to-live`
+* Method: `DELETE`
+* Endpoint: `/room/{token}/ttl`
+
+* Response:
+    - Status code:
+        + `200 OK`
+        + `403 Forbidden` When the current user is not a moderator, owner or guest moderator
+        + `404 Not Found` When the conversation could not be found for the participant
```
nickvergessen, line 362: Since 0 means disabled, we can also just do a post with 0? Less endpoints less breaking?

appinfo/routes/routesRoomController.php
```diff
@@ -102,5 +102,9 @@
 		['name' => 'Room#setLobby', 'url' => '/api/{apiVersion}/room/{token}/webinar/lobby', 'verb' => 'PUT', 'requirements' => $requirementsWithToken],
 		/** @see \OCA\Talk\Controller\RoomController::setSIPEnabled() */
 		['name' => 'Room#setSIPEnabled', 'url' => '/api/{apiVersion}/room/{token}/webinar/sip', 'verb' => 'PUT', 'requirements' => $requirementsWithToken],
+		/** @see \OCA\Talk\Controller\RoomController::setTimeToLive() */
+		['name' => 'Room#setTimeToLive', 'url' => '/api/{apiVersion}/room/{token}/ttl', 'verb' => 'POST', 'requirements' => $requirements],
+		/** @see \OCA\Talk\Controller\RoomController::removeTimeToLive() */
+		['name' => 'Room#setTimeToLive', 'url' => '/api/{apiVersion}/room/{token}/ttl', 'verb' => 'DELETE', 'requirements' => $requirements],
```
nickvergessen, line 108:
```suggestion
		['name' => 'Room#removeTimeToLive', 'url' => '/api/{apiVersion}/room/{token}/ttl', 'verb' => 'DELETE', 'requirements' => $requirements],
```

docs/conversation.md
```diff
@@ -81,6 +81,7 @@
         `statusMessage` | string | v4 | | Optional: Only available for one-to-one conversations and when  `includeStatus=true` is set
         `participants` | array | v1 | v2 | **Removed**
         `guestList` | string | v1 | v2 | **Removed**
+        `timeToLive` | int | v1 | The time to live of messages in this chat. Zero if disabled.
```
nickvergessen, line 84:
```suggestion
        `timeToLive` | int | v4 | | The time to live of messages in this chat. Zero if disabled. (only available with `time-to-live` capability)
```

docs/conversation.md
```diff
@@ -330,6 +331,24 @@
...
+## Set time to live for messages in a conversation
+
+* Required capability: `time-to-live`
+* Method: `POST`
+* Endpoint: `/room/{token}/ttl`
+* Data:
+
+    field | type | Description
+    ---|---|---
+    `ttl` | int | The time to live for messages in seconds. If is zero, will disable the ttl.
```
nickvergessen, line 343:
```suggestion
    `ttl` | int | The time to live for messages in seconds. If is zero, messages will not be deleted automatically.
```

lib/Controller/RoomController.php
```diff
@@ -1697,4 +1699,22 @@ public function resendInvitations(?int $attendeeId): DataResponse {
...
+	public function setTimeToLive(string $token, int $ttl): DataResponse {
+		if ($ttl < 0) {
+			return new DataResponse([], Http::STATUS_BAD_REQUEST);
+		}
+		try {
+			$sessionId = $this->session->getSessionForRoom($token);
+			$participant = $this->room->getParticipant($this->userId, $sessionId);
+		} catch (ParticipantNotFoundException $e) {
+			return new DataResponse([], Http::STATUS_FORBIDDEN);
+		}
```
nickvergessen, line 1716: The `@RequireModeratorParticipant` annotation ensures moderator status already.

docs/conversation.md
```diff
@@ -81,6 +81,7 @@
         `statusMessage` | string | v4 | | Optional: Only available for one-to-one conversations and when  `includeStatus=true` is set
         `participants` | array | v1 | v2 | **Removed**
         `guestList` | string | v1 | v2 | **Removed**
+        `timeToLive` | int | v4 | The time to live of messages in this chat. Zero if disabled. (only available with `time-to-live` capability)
```
nickvergessen, line 84:
```suggestion
        `timeToLive` | int | v4 | | The time to live of messages in this chat. Zero if disabled. (only available with `time-to-live` capability)
```
Empty column is needed, so the Description is in the right column:

2022-05-09 19_50_44-Implement time to live by vitormattos · Pull Request #7327 · nextcloud_spreed

Also it should be above `participants` so removed items are at the end.

lib/Controller/RoomController.php
```diff
@@ -1698,4 +1700,16 @@ public function resendInvitations(?int $attendeeId): DataResponse {
...
+
+	/**
+	 * @PublicPage
+	 * @RequireModeratorParticipant
+	 */
+	public function setTimeToLive(int $ttl): DataResponse {
+		if ($ttl < 0) {
+			return new DataResponse([], Http::STATUS_BAD_REQUEST);
+		}
+		$this->roomService->setTimeToLive($this->room, $this->userId, $ttl);
```
nickvergessen, line 1712:
```suggestion
		$this->roomService->setTimeToLive($this->room, $this->participant, $ttl);
```

Otherwise it can not be used by guest moderators

lib/Service/RoomService.php
```diff
@@ -391,4 +404,121 @@ public function verifyPassword(Room $room, string $password): array {
...
+		$result = $query->executeQuery();
+		$lastCheckedEpoch = $result->fetchOne();
+		$lastChechedDateTime = $this->timeFactory->getDateTime('@' . $lastCheckedEpoch);
+		return $lastChechedDateTime;
+	}
+
+	private function getMessageIdsByRoomIdInDateInterval(int $roomId, \DateTime $min, \DateTime $max): array {
+		$query = $this->db->getQueryBuilder();
+		$query->select('id')
+			->from('comments')
```
nickvergessen, line 479: needs to go into `Chat/CommentsManager.php` and should be upstreamed for the release.

tests/integration/features/ttl/ttl.feature
```diff
@@ -0,0 +1,23 @@
+Feature: room/ttp
```
nickvergessen, line 1:
```suggestion
Feature: ttl/ttl
```
But I would suggest to move it into the chat/ folder
```suggestion
Feature: chat/ttl
```

A new folder means a new drone job, means more CI time is spent on cloning and setting up.
And since it's about chat messages only, chat/ sounds good enough?

appinfo/info.xml
```diff
@@ -62,6 +62,7 @@ And in the works for the [coming versions](https://github.com/nextcloud/spreed/m
 		<job>OCA\Talk\BackgroundJob\CheckReferenceIdColumn</job>
 		<job>OCA\Talk\BackgroundJob\CheckHostedSignalingServer</job>
 		<job>OCA\Talk\BackgroundJob\CheckMatterbridges</job>
+		<job>OCA\Talk\BackgroundJob\ApplyTtl</job>
```
nickvergessen, line 65: Remove this, it's not needed as each room will get their own job added on the fly

docs/conversation.md
```diff
@@ -330,6 +331,24 @@
         + `401 Unauthorized` When the participant is a guest
         + `404 Not Found` When the conversation could not be found for the participant
 
+## Set time to live for messages in a conversation
```
nickvergessen, line 334: Jan doesn't like "ttl" and "time to live"
Maybe we rename the database field to "chat_expiration"
and also rename the value on the APIs and the capabilities?

lib/BackgroundJob/ApplyTtl.php
```diff
@@ -0,0 +1,52 @@
...
+		// Every 5 minutes
+		$this->setInterval(5 * 60);
+		$this->setTimeSensitivity(IJob::TIME_SENSITIVE);
+	}
+
+	/**
+	 * @param array $argument
+	 */
+	protected function run($argument): void {
+		$this->roomService->deleteExpiredTtl($argument['room_id'], $this->getId());
```
nickvergessen, line 50: I would expect this method on OCA\Talk\Chat\ChatManager?

lib/Migration/Version15000Date20220506005058.php
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Talk\Migration;
+
+use Closure;
+use Doctrine\DBAL\Types\Types;
```
nickvergessen, line 30:
```suggestion
use OCP\DB\Types;
```

lib/Room.php
```diff
@@ -279,6 +284,10 @@ public function setListable(int $newState): void {
 		$this->listable = $newState;
 	}
 
+	public function getTimeToLive(): int {
+		return $this->timeToLive;
+	}
```
nickvergessen, line 289: Needs a `setTimeToLive` method as well (see new way of doing things from https://github.com/nextcloud/spreed/issues/6235

tests/integration/features/ttl/ttl.feature
vitormattos: typo

appinfo/info.xml
vitormattos: Good to known. I will remove.

lib/Service/RoomService.php
```diff
@@ -392,4 +409,101 @@ public function verifyPassword(Room $room, string $password): array {
...
+	public function setTimeToLive(Room $room, Participant $participant, int $ttl): void {
+		$event = new ChangeTtlEvent($room, $ttl);
+		$this->dispatcher->dispatch(Room::EVENT_BEFORE_SET_TIME_TO_LIVE, $event);
+
+		$update = $this->db->getQueryBuilder();
+		$update->update('talk_rooms')
+			->set('time_to_live', $update->createNamedParameter($ttl, IQueryBuilder::PARAM_INT))
+			->where($update->expr()->eq('id', $update->createNamedParameter($room->getId(), IQueryBuilder::PARAM_INT)));
+		$update->executeStatement();
+		$jobList = Server::get(IJobList::class);
```
nickvergessen, line 422: Should use DI in the constructor

lib/Service/RoomService.php
```diff
@@ -392,4 +409,101 @@ public function verifyPassword(Room $room, string $password): array {
...
+
+	public function setTimeToLive(Room $room, Participant $participant, int $ttl): void {
+		$event = new ChangeTtlEvent($room, $ttl);
+		$this->dispatcher->dispatch(Room::EVENT_BEFORE_SET_TIME_TO_LIVE, $event);
+
+		$update = $this->db->getQueryBuilder();
+		$update->update('talk_rooms')
+			->set('time_to_live', $update->createNamedParameter($ttl, IQueryBuilder::PARAM_INT))
+			->where($update->expr()->eq('id', $update->createNamedParameter($room->getId(), IQueryBuilder::PARAM_INT)));
+		$update->executeStatement();
```
nickvergessen, line 421: After the update the value should be set on the object:
```php
$room->setTimeToLive($ttl);
```

lib/Service/RoomService.php
```diff
@@ -392,4 +409,101 @@ public function verifyPassword(Room $room, string $password): array {
...
+		$jobList = Server::get(IJobList::class);
+		if ($ttl > 0) {
+			$this->ttlSystemMessage($room, $participant, $ttl, 'ttl_enabled');
+			$jobList->add(ApplyTtl::class, ['room_id' => $room->getId()]);
+		} else {
+			$this->ttlSystemMessage($room, $participant, $ttl, 'ttl_disabled');
+			$jobList->remove(ApplyTtl::class, ['room_id' => $room->getId()]);
+		}
```
nickvergessen, line 429:
```suggestion
		if ($ttl > 0) {
			$jobList->add(ApplyTtl::class, ['room_id' => $room->getId()]);
			$this->ttlSystemMessage($room, $participant, $ttl, 'ttl_enabled');
		} else {
			$jobList->remove(ApplyTtl::class, ['room_id' => $room->getId()]);
			$this->ttlSystemMessage($room, $participant, $ttl, 'ttl_disabled');
		}
```

The background job is more important so let's do that first, so in case it fails we didn't add a chat message yet

lib/Service/RoomService.php
```diff
@@ -392,4 +409,101 @@ public function verifyPassword(Room $room, string $password): array {
...
+
+
+	public function deleteExpiredTtl(int $roomId, int $jobId): array {
+		$room = $this->manager->getRoomById($roomId);
+
+		$max = $this->getMaxDateTtl($room->getTimeToLive());
+		$min = $this->getMinDateTtl($jobId);
+
+		$ids = $this->commentsManager->getMessageIdsByRoomIdInDateInterval($roomId, $min, $max);
+		if (count($ids)) {
```
nickvergessen, line 456: If you are not interested in the value, don't count:
```suggestion
		if (!empty($ids)) {
```

lib/Service/RoomService.php
```diff
@@ -392,4 +409,101 @@ public function verifyPassword(Room $room, string $password): array {
...
+		$query->select('last_checked')
+			->from('jobs')
+			->where(
+				$query->expr()->eq('id', $query->createNamedParameter($jobId, IQueryBuilder::PARAM_INT))
+			);
+		$result = $query->executeQuery();
+		$lastCheckedEpoch = $result->fetchOne();
+		$lastChechedDateTime = $this->timeFactory->getDateTime('@' . $lastCheckedEpoch);
+		return $lastChechedDateTime;
+	}
```
nickvergessen, line 479: Illegal access of private table.

Also we shouldn't need it.

As discussed we will change the logic and save the ttl while writing a message and then we can simply get the messages by just checking if their expiration timestamp is != 0 and lower than "NOW"

lib/Service/RoomService.php
```diff
@@ -392,4 +409,101 @@ public function verifyPassword(Room $room, string $password): array {
...
+		$delete->delete('comments')
+			->where(
+				$delete->expr()->orX(
+					$delete->expr()->in('id', $delete->createNamedParameter($ids, IQueryBuilder::PARAM_INT_ARRAY)),
+					$delete->expr()->in('parent_id', $delete->createNamedParameter($ids, IQueryBuilder::PARAM_INT_ARRAY)),
+					$delete->expr()->in('topmost_parent_id', $delete->createNamedParameter($ids, IQueryBuilder::PARAM_INT_ARRAY))
+				)
+			);
+		$delete->executeStatement();
+	}
```
nickvergessen, line 492: Illegal access of private table.

Please move to comments manager and send upstream pull request to the server repo.

lib/Service/RoomService.php
```diff
@@ -392,4 +409,101 @@ public function verifyPassword(Room $room, string $password): array {
...
+				'ttl_expired',
+				json_encode(['message' => 'message_deleted', 'parameters' => ['message' => $id]]),
+				$this->timeFactory->getDateTime(),
+				false,
+				null,
+				$id,
+				true
+			);
+		}
+	}
```
nickvergessen, line 508: We shouldn't create a system message that a message has been deleted...

lib/Chat/CommentsManager.php
```diff
@@ -92,4 +92,30 @@ public function retrieveReactionsByActor(string $actorType, string $actorId, arr
...
+		$query = $this->dbConn->getQueryBuilder();
+		$query->select('id')
+			->from('comments')
+			->where(
+				$query->expr()->andX(
+					$query->expr()->eq('object_id', $query->createNamedParameter($roomId, IQueryBuilder::PARAM_INT)),
+					$query->expr()->eq('object_type', $query->createNamedParameter('chat')),
+					$query->expr()->gte('creation_timestamp', $query->createNamedParameter($min, IQueryBuilder::PARAM_DATE)),
+					$query->expr()->lte('creation_timestamp', $query->createNamedParameter($max, IQueryBuilder::PARAM_DATE))
+				)
```
nickvergessen, line 112: Should also limit to some messages only?
Would be nice to keep that a user was added?

Then again, if we turn around the logic and add the ttl time when the messages is being created we can control it for each message and just not set a TTL for those.

tests/integration/spreedcheats/lib/Controller/ApiController.php
```diff
@@ -96,4 +97,41 @@ public function resetSpreed(): DataResponse {
...
 		return new DataResponse();
 	}
+
+	/**
+	 * @NoCSRFRequired
+	 *
+	 * @return JSONResponse
+	 */
+	public function getTtlJob($token): JSONResponse {
+		$class = 'OCA\Talk\BackgroundJob\ApplyTtl';
```
nickvergessen, line 107:
```suggestion
		$class = \OCA\Talk\BackgroundJob\ApplyTtl::class;
```

tests/integration/spreedcheats/lib/Controller/ApiController.php
```diff
@@ -27,6 +27,7 @@
 
 use OCP\AppFramework\OCSController;
 use OCP\AppFramework\Http\DataResponse;
+use OCP\AppFramework\Http\JSONResponse;
```
nickvergessen, line 30: Should use DataResponse in all cases

tests/integration/spreedcheats/lib/Controller/ApiController.php
```diff
@@ -96,4 +97,41 @@ public function resetSpreed(): DataResponse {
...
+	/**
+	 * @NoCSRFRequired
+	 *
+	 * @return JSONResponse
+	 */
+	public function getTtlJob($token): JSONResponse {
+		$class = 'OCA\Talk\BackgroundJob\ApplyTtl';
+		$roomId = $this->getRoomIdByToken($token);
+		if (!$roomId) {
+			return new JSONResponse();
```
nickvergessen, line 110: Return 404 status, so errors in integration tests are debuggable

tests/integration/spreedcheats/lib/Controller/ApiController.php
```diff
@@ -96,4 +97,41 @@ public function resetSpreed(): DataResponse {
...
+					$query->expr()->eq('class', $query->createNamedParameter($class)),
+					$query->expr()->eq('argument', $query->createNamedParameter(json_encode(['room_id' => (int) $roomId])))
+				)
+			);
+		$result = $query->executeQuery();
+		$job = $result->fetchOne();
+		if ($job) {
+			return new JSONResponse(['id' => (int) $job]);
+		}
+		return new JSONResponse();
```
nickvergessen, line 126: Also here 404

lib/Service/RoomService.php
vitormattos: How we will notify the clients to remove the deleted messages from message list? I did the same that when the user delete a message.

lib/Migration/Version15000Date20220506005058.php
nickvergessen: 3

lib/Service/RoomService.php
nickvergessen: They will have to respect the expiration timestamp we give when rendering the message.
The problem is this message ^ will again expire? :P 

But it seems you forgot about the change o approach we discussed last week or 2 weeks ago.
We need to bring a dedicated column to the oc_comments table with a PR in server.

.drone.jsonnet
```diff
@@ -123,6 +123,7 @@ local PipelinePostgreSQL(test_set) = Pipeline(
 	PipelineSQLite("conversation-2"),
 	PipelineSQLite("federation"),
 	PipelineSQLite("reaction"),
+	PipelineSQLite("message_expire"),
```
nickvergessen, line 126: As mentioned before, please move your file to chat/ so we don't need a new pipeline (saves ~2mins of overhead

lib/Service/RoomService.php
vitormattos: A problem to save the time interval or a date to expire the message is when the room manager change or disable the interval. In this case, if the room manager disable the time to expire and have a message with time to expire in the future, this message will be deleted in the future. If this isn't a problem, we can set by message the time to expire.

nickvergessen: we will label it like this and it's how other messengers handle it as well. only the TTL while posting is accepted. Whatever applies at that time will take care of this message

lib/Service/RoomService.php
```diff
@@ -524,4 +542,43 @@ public function verifyPassword(Room $room, string $password): array {
 			'url' => '',
 		];
 	}
+
+	public function setExpireInterval(Room $room, Participant $participant, int $seconds): void {
+		$event = new ChangeExpireDateEvent($room, $seconds);
+		$this->dispatcher->dispatch(Room::EVENT_BEFORE_SET_EXPIRE_DATE, $event);
```
nickvergessen, line 548: Event name sounds like the room would be expired instead of the chat messages.
(And room expiration might come in the future, so we should keep the event name clear)

lib/Service/RoomService.php
```diff
@@ -524,4 +542,43 @@ public function verifyPassword(Room $room, string $password): array {
...
+		$update->update('talk_rooms')
+			->set('message_expire', $update->createNamedParameter($seconds, IQueryBuilder::PARAM_INT))
+			->where($update->expr()->eq('id', $update->createNamedParameter($room->getId(), IQueryBuilder::PARAM_INT)));
+		$update->executeStatement();
+		$room->setExpireInterval($seconds);
+		if ($seconds > 0) {
+			$this->jobList->add(ApplyExpireDate::class, ['room_id' => $room->getId()]);
+			$this->expireDateSystemMessage($room, $participant, $seconds, 'message_expire_enabled');
+		} else {
+			$this->jobList->remove(ApplyExpireDate::class, ['room_id' => $room->getId()]);
```
nickvergessen, line 560: There could still be messages which need to be expired.
So we need to keep the job?

Wasn't that also the reason why we wanted to move to a single background job instead of "per conversation"?

tests/integration/features/chat/expire-date.feature
```diff
@@ -0,0 +1,23 @@
+Feature: room/expire_date
```
nickvergessen, line 1: also here should revert to "expire messages"

nickvergessen (changes requested):

docs/conversation.md
```diff
@@ -51,6 +51,7 @@
         `participantFlags` | int | v1 | | "In call" flags of the user's session making the request (only available with `in-call-flags` capability)
         `readOnly` | int | v1 | | Read-only state for the current user (only available with `read-only-rooms` capability)
         `listable` | int | v3 | | Listable scope for the room (only available with `listable-rooms` capability)
+        `expireDate` | int | v4 | | The expire date time in seconds in this chat. Zero if disabled. (only available with `message-expire` capability)
```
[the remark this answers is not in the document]
vitormattos: We already approved as expireDate in server. Do you preffer to change all, including the server methods, to expireMessage?
https://github.com/nextcloud/server/blob/master/lib/public/Comments/IComment.php#L310

lib/Chat/Parser/SystemMessage.php
```diff
@@ -456,6 +456,17 @@ public function parseMessage(Message $chatMessage): void {
...
+			$parsedMessage = $this->l->t('{actor} set the self-destruction timer to {seconds} seconds');
+			if ($currentUserIsActor) {
+				$parsedMessage = $this->l->t('You set the self-destruction timer to {seconds} seconds');
+			}
+			$parsedParameters['seconds'] = $parameters['seconds'];
+		} elseif ($message === 'message_expire_disabled') {
+			$parsedMessage = $this->l->t('{actor} disabled the self-destruction timer');
+			if ($currentUserIsActor) {
+				$parsedMessage = $this->l->t('You disabled the self-destruction timer');
+			}
```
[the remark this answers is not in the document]
vitormattos: I think the same as you when I wrote this message, but I don't know what will the best to translation.

We have the problem about singular and plural and I don't know if the sequence is {timeInterval}{timeUnit} or {timeUnit}{timeInterval} in all languages.
```
{actor} set the message expiration to 1 second
{actor} set the message expiration to 1 seconds
{actor} set the message expiration to 1 day
{actor} set the message expiration to 1 days
{actor} set the message expiration to 1 week
{actor} set the message expiration to 1 weeks
```
If the order is the same in all languages, maybe we can implement this:
{actor} set the message expiration to {timeInterval} {timeUnit}

lib/Service/RoomService.php
vitormattos: Probably we don't need to maintain this. We have here two scenarios:

Expire job by rom:
- We only will have a job if we have a room with expire message time.

Global expire message:
- A query running every 5 minutes (or other interval defined by cron)

Thinking about performance, I believe that if we have a job running all times that the cron is executed, won't be a problem.

What way you preffer?

But is working now, maybe we can merge this PR for the developers of clients go ahead and modify this in a second step in other PR to don't make this PR more big.

lib/Chat/Parser/SystemMessage.php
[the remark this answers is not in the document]
nickvergessen: Plurals are handled by using `$l->n()` so that's not an issue.

And then it's basically something like:
```php
<?php
$weeks = $parameters['seconds'] > (86400 * 7) ? (int) round($parameters['seconds'] / (86400 * 7)) : 0;
$days = $parameters['seconds'] > 86400 ? (int) round($parameters['seconds'] / 86400) : 0;
$hours = $parameters['seconds']  ? (int) round($parameters['seconds'] / 3600) : 0;

if ($weeks > 0) {
    $parsedMessage = $this->l->t('{actor} set the message expiration to %n week', '{actor} set the message expiration to %n weeks', $weeks);
} elseif ($days > 0) {
    $parsedMessage = $this->l->t('{actor} set the message expiration to %n day', '{actor} set the message expiration to %n days', $days);
} else {
    $parsedMessage = $this->l->t('{actor} set the message expiration to %n hour', '{actor} set the message expiration to %n hours', $hours);
}
```

docs/conversation.md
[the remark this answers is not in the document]
nickvergessen: Server methods are fine as they are on the Comments object and table so there it is clear that it expires the comment only.

docs/chat.md
```diff
@@ -384,4 +384,4 @@ See [OCP\RichObjectStrings\Definitions](https://github.com/nextcloud/server/blob
 * Creating a poll is an `object_shared` with a poll object
 * `poll_voted` - {actor} voted on the poll {poll}
 * `poll_closed` - {actor} closed the poll {poll}
-
+* `message_expire_enabled` - {actor} set the self-destruction timer to {seconds} sec
```
nickvergessen, line 387: missing disabled message

lib/Service/RoomService.php
nickvergessen: > We only will have a job if we have a room with expire message time.

That is what I'm trying to say. It is not true. Expiration could be turned off now, but messages that where posted before still need to expire like the setting was back then. so you can not immediatelly remove the background job.

nickvergessen: > But is working now, maybe we can merge this PR for the developers of clients go ahead and modify this in a second step in other PR to don't make this PR more big.

yeah

nickvergessen:
Follow ups moved to https://github.com/nextcloud/spreed/issues/7508

lib/Service/RoomService.php
```diff
@@ -524,4 +541,42 @@ public function verifyPassword(Room $room, string $password): array {
...
+				'parameters' => ['seconds' => $seconds]
+			]),
+			$this->timeFactory->getDateTime(),
+			false
+		);
+	}
+
+	public function getMessageExpiration(Room $room): int {
+		return $room->getMessageExpiration();
+	}
```
nickvergessen, line 581: Remove?

```suggestion
```

nickvergessen (approved):

PR #7338, opened 2022-05-06
Allow the HPB to group session pinging across rooms

### LINKED ISSUE #7335: Combine session updates of multiple rooms into 1 request from the HPB
### How to use GitHub

---
Background numbers of a company call:
```
$ cat nextcloud.log | grep "nextcloud-spreed-signaling" | wc -l
7510
$ cat nextcloud.log | grep "nextcloud-spreed-signaling" | grep " sessions in room " | wc -l
6279
$ cat nextcloud.log | grep "nextcloud-spreed-signaling" | grep -v " sessions in room " | wc -l
1231
```

about 6.2k ping requests happen. Those could mostly be combined to ~6 per minute => 360 per hour (if the server can handle ~200 sessions in one request).

The idea is to add a capability based on a config which signals the HPB how many sessions can be sent per request and then drop the room check from the controller as we don't need to know it afterwards to update the sessions:
https://github.com/nextcloud/spreed/blob/c07251f0c67d3c02debe1f6dbe0e12ee3fe9f99a/lib/Controller/SignalingController.php#L732-L746

Once implemented a ticket in https://github.com/strukturag/nextcloud-spreed-signaling/issues should be created

nickvergessen:
- [x] Requires https://github.com/strukturag/nextcloud-spreed-signaling/issues/246

Default enabled with 200 sessions per request:
```
occ config:app:set spreed "session-ping-limit" --value="200"
```

Fix #7335

### DIFF
--- a/lib/Capabilities.php
+++ b/lib/Capabilities.php
@@ -119,7 +119,10 @@ class Capabilities implements IPublicCapability {
 				],
 				'conversations' => [],
 				'previews' => [
-					'max-gif-size' => (int)$this->serverConfig->getAppValue('spreed', 'max-gif-size', '3145728')
+					'max-gif-size' => (int)$this->serverConfig->getAppValue('spreed', 'max-gif-size', '3145728'),
+				],
+				'signaling' => [
+					'session-ping-limit' => max(0, (int)$this->serverConfig->getAppValue('spreed', 'session-ping-limit', '200')),
 				],
 			],
 		];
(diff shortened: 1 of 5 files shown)


lib/Capabilities.php
```diff
@@ -114,7 +114,10 @@ public function getCapabilities(): array {
 				],
 				'conversations' => [],
 				'previews' => [
-					'max-gif-size' => (int)$this->serverConfig->getAppValue('spreed', 'max-gif-size', '3145728')
+					'max-gif-size' => (int)$this->serverConfig->getAppValue('spreed', 'max-gif-size', '3145728'),
+				],
+				'signaling' => [
+					'session-ping-limit' => (int)$this->serverConfig->getAppValue('spreed', 'session-ping-limit', '0'),
```
nickvergessen, line 120:
```suggestion
					'session-ping-limit' => (int)$this->serverConfig->getAppValue('spreed', 'session-ping-limit', '200'),
```

Should we default enable with 200? @fancycode

fancycode: Yeah, 200 is probably a good start. We can tweak this later once you have some performance numbers.

fancycode (approved):
lgtm

PR #7369, opened 2022-05-13
Retry sending signaling messages up to 3 times

nickvergessen:

### DIFF
--- a/lib/Signaling/BackendNotifier.php
+++ b/lib/Signaling/BackendNotifier.php
@@ -25,6 +25,7 @@ declare(strict_types=1);
 
 namespace OCA\Talk\Signaling;
 
+use GuzzleHttp\Exception\ConnectException;
 use GuzzleHttp\Exception\ServerException;
 use OCA\Talk\Config;
 use OCA\Talk\Model\Attendee;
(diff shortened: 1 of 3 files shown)


lib/Signaling/BackendNotifier.php
```diff
@@ -82,6 +84,13 @@ protected function doRequest(string $url, array $params): void {
 			if (!$this->signalingManager->isCompatibleSignalingServer($response)) {
 				throw new \RuntimeException('Signaling server needs to be updated to be compatible with this version of Talk');
 			}
+		} catch (ServerException $e) {
+			if ($retries > 1) {
+				$this->logger->error('Failed to send message to signaling server, ' . $retries . ' retries left!', ['exception' => $e]);
```
fancycode, line 89: I don't think all requests should be retried in all cases. E.g. if the shared secret is wrong and an authentication error is returned, it doesn't make much sense to retry the request as the retries will fail, too.

nickvergessen: ServerException is 5xx status code. 4xx are in ClientException:
https://docs.guzzlephp.org/en/stable/quickstart.html#exceptions

That being said, I guess we should also retry on ConnectException?

vitormattos (approved):

vitormattos:
Do we have a place to document this?

PR #7381, opened 2022-05-17
Move setType to room service

nickvergessen:


lib/Manager.php
```diff
@@ -391,7 +391,7 @@ public function removeUserFromAllRooms(IUser $user): void {
 		foreach ($leftRooms as $room) {
 			// We are changing the room type and name so a potential follow up
 			// user with the same user-id can not reopen the one-to-one conversation.
-			$room->setType(Room::TYPE_GROUP, true);
+			Server::get(RoomService::class)->setType($room, Room::TYPE_GROUP, true);
```
vitormattos, line 394: Do got problem when use as a property defined with dependency injection?

vitormattos: Then, is good put a `@todo`.

nickvergessen: The todo is `Server::get()`

vitormattos (approved):

Issue #7373, opened 2022-05-16
Compile SCSS manually


nickvergessen:
The SCSS cacher has been removed with https://github.com/nextcloud/server/pull/32261

So we need to manually compile our SCSS files.
Would of course be nice to get rid of them if possible and move the stuff to the actual components, but e.g. with the At.scss and unified search as well as the icons it might not be suitable.
There we should fix them to the new variables manually and then compile them like in the server

- [x] app https://github.com/nextcloud/spreed/pull/7576
- [x] At https://github.com/nextcloud/spreed/pull/7576
- [x] icons https://github.com/nextcloud/spreed/pull/7576
- [x] publicshare https://github.com/nextcloud/spreed/pull/7576
- [x] publicshareauth https://github.com/nextcloud/spreed/pull/7576
- [x] unified-search https://github.com/nextcloud/spreed/pull/7576
- [x] icon-folder-multiple-image is undefined used in RightSidebar.vue for "shared items" https://github.com/nextcloud/spreed/pull/7576
- [ ] settings-admin https://github.com/nextcloud/spreed/pull/7577
- [ ] #7578

https://github.com/nextcloud/spreed/blob/9e9197cc56fd8374c4916ba1fd029721cd1a5486/src/mainPublicShareSidebar.js#L98-L115

skjnldsv:
> Would of course be nice to get rid of them if possible and move the stuff to the actual components, but e.g. with the At.scss and unified search as well as the icons it might not be suitable.

just importing them in your main.js should work out of the box.
For the icons mixin thing though, we can chat about it tomorrow to find out what is missing :)

PR #7386, opened 2022-05-17
Use material icons everywhere in CallButton

juliushaertl:
Replace old icon classes with the fitting material design icon to have it visible with server/talk master.

Small bit for https://github.com/nextcloud/spreed/issues/7373

Found when looking into typescript stuff with @marcoambrosini

### DIFF
--- a/src/components/TopBar/CallButton.vue
+++ b/src/components/TopBar/CallButton.vue
@@ -33,10 +33,9 @@
 			:disabled="startCallButtonDisabled || loading || blockCalls"
 			:type="startCallButtonType"
 			@click="handleClick">
-			<template #icon>
-				<span class="icon"
-					:class="startCallIcon" />
-			</template>
+			<Video slot="icon"
+				:size="20"
+				decorative />
 			{{ startCallLabel }}
 		</Button>
 		<Button v-else-if="showLeaveCallButton && !canEndForAll"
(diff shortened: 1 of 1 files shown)


marcoambrosini (approved):

src/components/TopBar/CallButton.vue
```diff
@@ -33,21 +33,19 @@
 			:disabled="startCallButtonDisabled || loading || blockCalls"
 			:type="startCallButtonType"
 			@click="handleClick">
-			<template #icon>
-				<span class="icon"
-					:class="startCallIcon" />
-			</template>
+			<Video slot="icon"
```
nickvergessen, line 36: Bildschirmfoto vom 2022-05-18 08-14-03

@juliushaertl what's up with this phpstorm note?

nickvergessen (approved):
Works

src/components/TopBar/CallButton.vue
juliushaertl: Makes sense to migrate then I guess, but so far it does not seem to be part of our eslint config, so maybe it is a local check of PHPstorm or some plugin there?

https://eslint.vuejs.org/rules/no-deprecated-slot-attribute.html
> This rule is included in all of "plugin:vue/vue3-essential", "plugin:vue/vue3-strongly-recommended" and "plugin:vue/vue3-recommended". 

And we don't pull in any of those https://github.com/nextcloud/eslint-config/blob/master/index.js#L20

raimund-schluessler: The recommended syntax is to use a `template` component in conjunction with with the `v-slot` directive (or its shorthand `# `, see https://v2.vuejs.org/v2/guide/components-slots.html#Named-Slots

So like
```vue
<template #icon>
	<Video />
</template>
```
That saves a bit of effort if you ever migrate to vue3, because the old syntax is not supported there anymore.

PR #7394, opened 2022-05-19
Add phpunit as dependency like in other apps

nickvergessen:


marcoambrosini (approved):

nickvergessen:
truck.... that doesn't seem to trigger the run. So let me push again

PR #7331, opened 2022-05-06
Add an option for "Silent send" to the API

nickvergessen:
Fix #6257 

- [ ] ~~Check if it would be possible to make the "silent" persistent so it could be displayed later on~~ => https://github.com/nextcloud/spreed/issues/6257#issuecomment-1125827646

Issue #3950, opened 2020-07-28
📴 "Silent call" for group/public calls


Bolli84:
When you are using group chats, and want to make a short call for a few persons, you need to set a new room for it.
It would be good to have the option to "JOIN" a call silently also, when yu are the first caller. The first one, sends a call message to every other registered person for the room. 

E.G You reuse an existing room for weekly meetings. the first one, wh enters the room sends call informations to all others. This is pretty inconvenient for group call or appointment calls.
For direct immediate one-on-X calls, this is pretty cool, but not for scheduled calls.

Feature-wish:
Add a checkbox: Call all participants now? with yes/no. with "no", the room shall behave as no other registered person is connected to the room.

- [x] API https://github.com/nextcloud/spreed/pull/7416
- [x] Web #7445
- [x] iOS https://github.com/nextcloud/talk-ios/issues/786
- [x] Android https://github.com/nextcloud/talk-android/issues/1991
   - [x] please ping @mahibi whenever API & Web is merged so the android implementation can be tested

nickvergessen:
We are currently discussing an option to not be notified about calls in a conversation as addition to the chat notifications we have currently. I guess that would solve your issue too.

Bolli84:
It is part of the solution.

Being not informed about a call needs the receipient to take the action.

It would be good if the caller would have the option during opening a group call

jospoortvliet:
> It is part of the solution.
> 
> Being not informed about a call needs the receipient to take the action.
> 
> It would be good if the caller would have the option during opening a group call

I feel like recipients/participants should be able to decide what matters to them, not the sender.

Perhaps it makes sense to create an 'important'urgent' room where you start a call if you want everyone in, and let ppl disable this call notification on the other rooms.

Bolli84:
If you look at Skype for business, Teams, Zoom, .... you enter a "GROUP"-room, without notifing all participants by ringing them. 

For group-rooms, talk should do that equally and the SENDER/INVITER can decide, ring all or not.
OR the ROOM-OWNER/opener/inviter should do that.

nickvergessen:
Yeah and if you look at all mobile messengers which were our main comparisons before corona, they all do ring.

We will look into the issue, but we have many usecases to cover and we are yet unsure how to do it in a way that is not to troubleslme or disruptive

Bolli84:
Cool.

A nice way would be a little Box for the First caller/opener in a group: "Do you want to call the participants or open the room?" 
[...]

PR #7416, opened 2022-05-25
Support "Silent call" on API

vitormattos:
Related with #3950

- [x] Implement changes on API to accept boolean `silent` flag
- [x] Implement capability `silent-call`
- [x] Document api changes
- [x] Should add an integration test in features/callapi/ checking the notifications, like we did for https://github.com/nextcloud/spreed/pull/7331

### DIFF
--- a/docs/capabilities.md
+++ b/docs/capabilities.md
@@ -98,4 +98,5 @@ title: Capabilities
 * `silent-send` - Whether the chat API allows to send chat messages without triggering notifications
 * `sip-support-nopin` - Whether SIP can be configured to not require a custom attendee PIN
 * `send-call-notification` - When the API allows to resend call notifications for individual users that did not join yet
+* `silent-call` - Allow to start calls without sending notification
 * `config => call => enabled` - Whether calling is enabled on the instance or not
--- a/lib/Service/ParticipantService.php
+++ b/lib/Service/ParticipantService.php
@@ -43,6 +43,7 @@ use OCA\Talk\Events\RemoveParticipantEvent;
 use OCA\Talk\Events\RemoveUserEvent;
 use OCA\Talk\Events\RoomEvent;
 use OCA\Talk\Events\SendCallNotificationEvent;
+use OCA\Talk\Events\SilentModifyParticipantEvent;
 use OCA\Talk\Exceptions\ForbiddenException;
 use OCA\Talk\Exceptions\InvalidPasswordException;
 use OCA\Talk\Exceptions\ParticipantNotFoundException;
@@ -947,7 +948,7 @@ class ParticipantService {
 		$this->dispatcher->dispatch(Room::EVENT_AFTER_END_CALL_FOR_EVERYONE, $event);
 	}
 
-	public function changeInCall(Room $room, Participant $participant, int $flags, bool $endCallForEveryone = false): void {
+	public function changeInCall(Room $room, Participant $participant, int $flags, bool $endCallForEveryone = false, bool $silent = false): void {
 		$session = $participant->getSession();
 		if (!$session instanceof Session) {
 			return;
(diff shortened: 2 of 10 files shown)


docs/capabilities.md
```diff
@@ -99,3 +99,4 @@ title: Capabilities
 * `sip-support-nopin` - Whether SIP can be configured to not require a custom attendee PIN
 * `send-call-notification` - When the API allows to resend call notifications for individual users that did not join yet
 * `config => call => enabled` - Whether calling is enabled on the instance or not
+* `silent-call` - Allow to start call for group/public without notification
```
nickvergessen, line 102: move before `config => …` please

lib/Service/ParticipantService.php
```diff
@@ -961,10 +961,10 @@ public function changeInCall(Room $room, Participant $participant, int $flags, b
 			$flags &= ~Participant::FLAG_WITH_VIDEO;
 		}
 
-		if ($flags !== Participant::FLAG_DISCONNECTED) {
+		if ($flags !== Participant::FLAG_DISCONNECTED && !$silent) {
 			$event = new ModifyParticipantEvent($room, $participant, 'inCall', $flags, $session->getInCall());
 			$this->dispatcher->dispatch(Room::EVENT_BEFORE_SESSION_JOIN_CALL, $event);
-		} else {
+		} elseif (!$silent) {
```
nickvergessen, line 967: We still need the Events to trigger so the HPB notifies users about the participant, etc.
So extend the event with the parameter and inside the listener for the call notification return early when the silent parameter is true

lib/Service/ParticipantService.php
```diff
@@ -961,10 +961,10 @@ public function changeInCall(Room $room, Participant $participant, int $flags, b
 			$flags &= ~Participant::FLAG_WITH_VIDEO;
 		}
 
-		if ($flags !== Participant::FLAG_DISCONNECTED) {
+		if ($flags !== Participant::FLAG_DISCONNECTED && !$silent) {
```
nickvergessen, line 964: We still need the Events to trigger so the HPB notifies users about the participant, etc.
So extend the event with the parameter and inside the listener for the call notification return early when the silent parameter is true

lib/Service/ParticipantService.php
nickvergessen, line 967: end call can never be silent, so this is not needed
```suggestion
		} else {
```

nickvergessen (changes requested):
- [ ] Should add an integration test in features/callapi/ checking the notifications, like we did for https://github.com/nextcloud/spreed/pull/7331

lib/Signaling/Listener.php
```diff
@@ -262,6 +263,9 @@ public static function notifyAfterRoomDisconected(ParticipantEvent $event): void
 	}
 
 	public static function notifyAfterJoinUpdateAndLeave(ModifyParticipantEvent $event): void {
+		if ($event instanceof SilentModifyParticipantEvent) {
```
nickvergessen, line 266: this is also not the correct listener...

Actually what you want to opt out is this one:
https://github.com/nextcloud/spreed/blob/master/lib/Notification/Listener.php#L97-L101

vitormattos: Every when we create a room, this event is called and define the attribute room.last_activity with the current date.

https://github.com/nextcloud/spreed/blob/master/lib/Chat/SystemMessage/Listener.php#L83

The event:
https://github.com/nextcloud/spreed/blob/master/lib/Notification/Listener.php#L97-L101
Check this:
https://github.com/nextcloud/spreed/blob/master/lib/Notification/Listener.php#L195
And if the last_activity have a date, will never send the call notification.

docs/capabilities.md
```diff
@@ -98,4 +98,5 @@ title: Capabilities
 * `silent-send` - Whether the chat API allows to send chat messages without triggering notifications
 * `sip-support-nopin` - Whether SIP can be configured to not require a custom attendee PIN
 * `send-call-notification` - When the API allows to resend call notifications for individual users that did not join yet
+* `silent-call` - Allow to start call for group/public without notification
```
nickvergessen, line 101:
```suggestion
* `silent-call` - Allow to start calls without sending notification
```

nickvergessen (approved):
I think it's good now.
Probably we should go over the commits together and squash everything at the end, then it should be good to merge

vitormattos:
squashed

PR #7434, opened 2022-05-31
Route to the first search result upon pressing enter

marcoambrosini:
Sneaky lil change, been annoyed by this for too long

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -28,6 +28,7 @@
 				:is-searching="isSearching"
 				@input="debounceFetchSearchResults"
 				@submit="onInputEnter"
+				@keydown.enter.native="handleEnter"
 				@abort-search="abortSearch" />
 			<NewGroupConversation v-if="canStartConversations" />
 		</div>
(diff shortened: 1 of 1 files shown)


src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -490,6 +491,14 @@ export default {
 			this.handleScroll()
 			this.handleUnreadMention()
 		}, 50),
+
+		// Route to the first search result
+		handleEnter() {
+			if (this?.conversationsList[0]?.token) {
+				this.$router.push({ name: 'conversation', params: { token: this.conversationsList[0].token } })
```
nickvergessen, line 498: return after this?

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -490,6 +491,14 @@ export default {
...
+			}
+			this.handleClickSearchResult()
```
nickvergessen, line 500: This should start a 1-1 if the first shown entry is a user?
Doesn't work for me

nickvergessen (changes requested):

src/components/LeftSidebar/LeftSidebar.vue
marcoambrosini: Why? There's no need to return anything there

nickvergessen: Okay, seems I misunderstood the part after the if

nickvergessen (approved):

PR #7445, opened 2022-06-01
Silent calls front end

marcoambrosini:
Screenshot 2022-06-03 at 12 44 58

partially solves #3950

### DIFF
--- a/src/components/DeviceChecker/DeviceChecker.vue
+++ b/src/components/DeviceChecker/DeviceChecker.vue
@@ -149,9 +149,39 @@
 				{{ t('spreed', 'Always show this dialog before joining a call in this conversation.') }}
 			</CheckboxRadioSwitch>
 
-			<!-- Join call -->
-			<CallButton class="call-button"
-				:force-join-call="true" />
+			<div class="device-checker__call-buttons">
+				<!-- Silent call -->
+				<Actions :force-menu="true">
+					<template v-if="!silentCall">
+						<ActionButton :close-after-click="true"
+							icon="icon-upload"
+							:title="t('spreed', 'Silent call')"
+							@click="silentCall= true">
+							{{ t('spreed', 'The conversation participants won\'t be notifyed about this call starting') }}
+							<BellOff slot="icon"
+								:size="16"
+								decorative
+								title="" />
+						</ActionButton>
+					</template>
+					<template v-else>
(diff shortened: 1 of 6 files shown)


src/components/DeviceChecker/DeviceChecker.vue
```diff
@@ -149,9 +149,39 @@
...
-				:force-join-call="true" />
+			<div class="device-checker__call-buttons">
+				<!-- Silent call -->
+				<Actions :force-menu="true">
+					<template v-if="!silentCall">
+						<ActionButton :close-after-click="true"
+							icon="icon-upload"
+							:title="t('spreed', 'Silent call')"
+							@click="silentCall= true">
+							{{ t('spreed', 'The conversation participants won\'t be notifyed about this call starting') }}
```
nickvergessen, line 160:
```suggestion
							{{ t('spreed', 'The conversation participants will not be notified about this call') }}
```

src/components/DeviceChecker/DeviceChecker.vue
```diff
@@ -149,9 +149,39 @@
...
+								decorative
+								title="" />
+						</ActionButton>
+					</template>
+					<template v-else>
+						<ActionButton :close-after-click="true"
+							icon="icon-upload"
+							:title="t('spreed', 'Normal call')"
+							@click="silentCall= false">
+							{{ t('spreed', 'The conversation participants will be notifyed about this call starting') }}
```
nickvergessen, line 172:
```suggestion
							{{ t('spreed', 'The conversation participants will be notified about this call') }}
```

src/services/callsService.js
```diff
@@ -40,11 +40,13 @@ import {
  *
  * @param {string} token The token of the call to be joined.
  * @param {number} flags The available PARTICIPANT.CALL_FLAG for this participants
+ * @param {boolean }silent Whether the call should trigger a notifications and
```
nickvergessen, line 43:
```suggestion
 * @param {boolean} silent Whether the call should trigger a notifications and
```

nickvergessen (commented):
- [ ] `…` menu should be hidden if there is already a call and the button says "Join call" only
- [ ] Having said that, in this case we should always "Silent" join the call. This means that when the caller leaves in the meantime it won't ring everyone's phone again like in an endless call loop we sometimes experience?

Issue #2874, opened 2020-01-30
Allow setting conversation password on conversation creation via API


mario:
Currently to set a password on room creation you need two API calls:

- create a conversation
- set a password

I'd like an optional password field on the creation flow to avoid the extra network call

nickvergessen:
Same for the invites. I guess it is time for an api/v2/ at least on the creation flow it could be improved a lot.

Antreesy:
Fixed with #14497

PR #7473, opened 2022-06-10
Implement password policy checker

### LINKED ISSUE #3089: Enforce password policy for shared links
The password policy under Administration/Security is not applied on shared links in talk.
image

Since it is a general security setting (and not only for sharing files), it should apply to every shared link including talk rooms. 

1) The password field in talk should respect the settings in Administration/Security/Password policy
image

2) There should be an option (like in Administration/Sharing) to enforce password protection for talk conversations. This would prevent users from creating rooms without a password.
image

3)  People invited by mail should get a second e-mail with a custom password for this talk conversation (like sharing via mail). 
image

nickvergessen:
Fix #3089

### DIFF
--- a/docs/conversation.md
+++ b/docs/conversation.md
@@ -247,9 +247,15 @@
 * Response:
     - Status code:
         + `200 OK`
+        + `400 Bad Request` When the password does not match the password policy. Show `ocs.data.message` to the user in this case
         + `403 Forbidden` When the current user is not a moderator or owner
         + `403 Forbidden` When the conversation is not a public conversation
         + `404 Not Found` When the conversation could not be found for the participant
+
+    - Data:
+        field | type | Description
+        ---|---|---
+        `message` | string | Only available on `400 Bad Request`, translated error with the violated password policy rules
 
 ## Set default or call permissions for a conversation
 
--- a/lib/Command/Room/TRoomCommand.php
+++ b/lib/Command/Room/TRoomCommand.php
@@ -35,6 +35,7 @@ use OCA\Talk\Participant;
 use OCA\Talk\Room;
 use OCA\Talk\Service\ParticipantService;
 use OCA\Talk\Service\RoomService;
+use OCP\HintException;
 use OCP\IGroup;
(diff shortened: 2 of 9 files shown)


SystemKeeper:
Noticed two things

1. I can't remove a password which was correctly set anymore (GUI reflects a removed password):

https://user-images.githubusercontent.com/1580193/173200095-8626bb2f-d5f0-4f6c-998f-93876f412c44.mp4

3. When the password validation fails, the GUI still reflects a successful change:

https://user-images.githubusercontent.com/1580193/173200147-61f93d43-3eb4-46a6-9470-6f421e3a020b.mp4

SystemKeeper:
3. Conversation is still created, although an error occured:


https://user-images.githubusercontent.com/1580193/173200683-4efa80d9-e246-4ebb-9ae5-c46a8217deeb.mp4

nickvergessen:
Yeah noticed too that there are many issues with the field.
Also when you open the settings and the conversation has a password already, the input is not visible 🙈

nickvergessen:
> Conversation is still created, although an error occured:

That is actually a sad issue, as we currently set the password after the creation. Basically boils down to:
https://github.com/nextcloud/spreed/issues/2874

SystemKeeper:
Oh I see. Then this problem is out of scope for this PR and currently expected.

nickvergessen:
Or we use this moment to implement it

vitormattos (approved):

vitormattos (changes requested):
Only need fix cs

nickvergessen:
> Also when you open the settings and the conversation has a password already, the input is not visible :see_no_evil:

@marcoambrosini maybe you can help with the missing input field?

SystemKeeper:
Haven’t tested, but isn’t there special handling for removing the password (eg setting the password to „“) needed?

nickvergessen:
I had wrapped the event dispatch in an if checking for non empty string.... But seems it got removed by a rebase or something

vitormattos (approved):

Issue #7480, opened 2022-06-13
📊 Simple polls API - part 2


nickvergessen:
Follow ups to https://github.com/nextcloud/spreed/pull/7306

- [x] Add new media tab
- [x] Add capability for mobile clients
- [x] Question that came up:
Should we have an option that prevents changing the vote after you voted once? The goal is to prevent Band-Wagon on the second level via the editing option, but e.g. Telegram still allows this and we think it could be useful e.g. when you use it to agree on an appointment and duties change or similar stuff.

mahibi:
For the system message that someone voted on a poll, please don't use "talk-poll" as attribute text. Let's use "talk-poll-voted" instead.
Could you do that @vitormattos ?

This would make it much easier for the android app to differentiate between the messages.

vitormattos:
It sounds like a good idea.

mahibi:
i vote for 
`{"option0":1,"option1":1}`
as it's the structure the android app expects atm

nickvergessen:
Fix in https://github.com/nextcloud/spreed/pull/7500

mahibi:
@nickvergessen closed unintentionally? or is there already capability / mediatab?

nickvergessen:
Follow ups in https://github.com/nextcloud/spreed/pull/7512

Now only the editing/revoting question is open.
To be clarified with @marcoambrosini next week.

nickvergessen:
- [ ] Editing shall be allowed in both modes
- [ ] We should show the number of voters to the author and moderators on hidden polls so they know when to close

PR #7487, opened 2022-06-21
Add voted type

vitormattos:
Related: #7480
cc @mahibi


marcoambrosini (commented):
shouldn't we call it `talk-poll-vote`?

mahibi:
uhm sorry i think i caused some confusion here 🙈
actually i don't need this change!
it was even not about the parameters that it thought which was wrong but the 
'message' => 'poll_voted'
is just enough.

so ignore my comment in https://github.com/nextcloud/spreed/issues/7480#issuecomment-1160380493
not sure what came to my mind :D

PR #7489, opened 2022-06-21
Add app version to capabilities

SystemKeeper:
To make troubleshooting problems with Talk iOS easier I've submitted a PR which adds some troubleshooting infos to talk-ios, see https://github.com/nextcloud/talk-ios/pull/829
For this it would be really helpful to also show the version of the server version, but currently this version is not exposed. This PR adds the version as part of the capabilities.


vitormattos (approved):

nickvergessen (approved):
I'm fine with this in general.
We just have to make sure that this is never used to assume on features.
We have feature flags, the version is just used as diagnostic information

PR #7512, opened 2022-07-04
📊 Poll follow ups

### LINKED ISSUE #7480: 📊 Simple polls API - part 2

nickvergessen:
- [x] Add new media tab
- [x] Add capability for mobile clients
- [x] Allow editing/updating of votes, TBD with @mahibi @Ivansss @marcoambrosini 

Fix #7480

### DIFF
--- a/lib/Controller/PollController.php
+++ b/lib/Controller/PollController.php
@@ -248,7 +248,14 @@ class PollController extends AEnvironmentAwareController {
 
 		if (!$canSeeSummary && $poll->getStatus() === Poll::STATUS_OPEN) {
 			$data['votes'] = [];
-			$data['numVoters'] = 0;
+			if ($this->participant->hasModeratorPermissions(false)
+				|| ($poll->getActorType() === $this->participant->getAttendee()->getActorType()
+					&& $poll->getActorId() === $this->participant->getAttendee()->getActorId())) {
+				// Allow moderators and the author to see the number of voters,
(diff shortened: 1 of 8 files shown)


vitormattos (approved):

lib/Controller/PollController.php
```diff
@@ -248,7 +248,14 @@ protected function renderPoll(Poll $poll, array $votedSelf = [], array $detailed
 
 		if (!$canSeeSummary && $poll->getStatus() === Poll::STATUS_OPEN) {
 			$data['votes'] = [];
-			$data['numVoters'] = 0;
+			if ($this->participant->hasModeratorPermissions(false)
+				|| ($poll->getActorType() === $this->participant->getAttendee()->getActorType()
+					&& $poll->getActorId() === $this->participant->getAttendee()->getActorId())) {
+				// Allow moderators and the author to see the number of voters,
+				// So they know when to close the poll.
```
mahibi, line 255: @nickvergessen update doc for numVoters?

PR #7573, opened 2022-07-12
Move At.scss to At.css - Fixes missing mention popups

nickvergessen:
Ref #7373


nickvergessen:
Combined in https://github.com/nextcloud/spreed/pull/7576

PR #7574, opened 2022-07-12
Move share files to css

nickvergessen:
- [ ] Based on #7573


nickvergessen:
Combined in https://github.com/nextcloud/spreed/pull/7576

Issue #7566, opened 2022-07-11
Parsing issues of HTML elements


nickvergessen:
### Problem description

We could see that there are a number of validation errors that can cause accessibility issues. Among other things, there are several elements with identical ID-values.

There are also validation errors regarding code semantics that doesn’t follow the standard.

We had limited means to validate the code but we can affirm that there are issues that fail the WCAG-standard.

The following validation errors aren’t allowed in the WCAG-standard:

* Elements have complete start and end tags
* Elements are nested according to their specifications
* Elements do not contain duplicate attributes
* Any IDs are unique

It is worth keeping in mind that other validation errors can also cause accessibility problems. But these four are very likely to do so.

It’s generally best to keep a website’s code as close to standard as possible. This makes it more robust and minimizes risks for errors when assistive technologies interact with it.

We list some serious issues below, but there are more. We recommend that you validate the code, control that it’s semantically correct, and fix as many issues as possible.

### Talk app

Some examples from the Talk app:

* Elements without end tags
* Button-elements nested in other buttons or link-elements
* Div-elements nested in list-elements
* li-elements nested in span-elements
* Use of negative tab-index on interactive elements
* Elements with identical id-values

### Suggested solution

* Validate the code and fix as many issues as you can, but especially any of the four that fail this WCAG-criterion
* Follow standard as much as you can and validate anytime you build something new

---

(The DIV inside list I complained already about many times, mostly in the left sidebar)

nickvergessen:
> Error: Attribute delay not allowed on element [div](https://html.spec.whatwg.org/multipage/#the-div-element) at this point.
> Error: Attribute trigger not allowed on element [div](https://html.spec.whatwg.org/multipage/#the-div-element) at this point.

caused by
https://github.com/nextcloud/spreed/blob/762ccb66eeb7a95761e7bf9d1789b2d0db9783cf/src/components/MessagesList/MessagesGroup/Message/Message.vue#L126-L127

But the components seem to still have these props, so not sure where it goes wrong. Maybe it's because of the loop?

nickvergessen:
`<a data-v-7dc9c69e="" data-v-4bbf400e="" tabindex="" tag="a" href="https://nextcloud.local/index.php/f/82" target="_blank" rel="noopener noreferrer" class="file-preview file-preview--viewer-available rich-text--component" data-v-252db2c7="">`

Causes the following issues
[...]

PR #7586, opened 2022-07-13
Fix HTML validation of meta element

nickvergessen:
Ref https://github.com/nextcloud/spreed/issues/7566


danxuliu (approved):
I have not tried the extension, but the markup is fixed :+1:

PR #7595, opened 2022-07-13
Listener to send message expiration message

vitormattos:
PS: Branch created from #7594. Necessary to review and merge first the parent branch.

### DIFF
--- a/lib/Service/RoomService.php
+++ b/lib/Service/RoomService.php
@@ -538,9 +538,7 @@ class RoomService {
 	}
 
 	public function setMessageExpiration(Room $room, int $seconds): void {
-		$oldValue = $room->getMessageExpiration();
-		$room->setMessageExpiration($seconds);
-		$event = new ModifyRoomEvent($room, 'messageExpiration', $seconds, $oldValue);
+		$event = new ModifyRoomEvent($room, 'messageExpiration', $seconds);
 		$this->dispatcher->dispatch(Room::EVENT_BEFORE_SET_MESSAGE_EXPIRATION, $event);
 
 		$update = $this->db->getQueryBuilder();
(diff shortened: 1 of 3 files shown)


lib/Service/RoomService.php
```diff
@@ -547,29 +549,6 @@ public function setMessageExpiration(Room $room, Participant $participant, int $
 			->where($update->expr()->eq('id', $update->createNamedParameter($room->getId(), IQueryBuilder::PARAM_INT)));
 		$update->executeStatement();
 
-		$room->setMessageExpiration($seconds);
```
nickvergessen, line 550: Should be kept

lib/Service/RoomService.php
```diff
@@ -537,8 +537,10 @@ public function verifyPassword(Room $room, string $password): array {
 		];
 	}
 
-	public function setMessageExpiration(Room $room, Participant $participant, int $seconds): void {
-		$event = new ModifyRoomEvent($room, 'messageExpiration', $seconds, null, $participant);
+	public function setMessageExpiration(Room $room, int $seconds): void {
+		$oldValue = $room->getMessageExpiration();
+		$room->setMessageExpiration($seconds);
```
nickvergessen, line 542: Should be done later so it's the same as in all other methods (and the before event still has the old value in the room object

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -2588,6 +2588,20 @@ public function waitForXSeconds($seconds): void {
 		sleep($seconds);
 	}
 
+	/**
+	 * @When apply message expiration job manually
+	 */
+	public function applyMessageExpirationJob(): void {
```
nickvergessen, line 2594: needs a rebase

nickvergessen (changes requested):

lib/Service/RoomService.php
vitormattos: kept when rebased from master

nickvergessen (approved):

PR #7596, opened 2022-07-13
occ command to set expire message

vitormattos:
Before, review and merge this PR: https://github.com/nextcloud/spreed/pull/7595


lib/Command/Room/MessageExpiration.php
```diff
@@ -0,0 +1,90 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class MessageExpiration extends Base {
+	use TRoomCommand;
+
+	protected function configure(): void {
+		$this
+			->setName('talk:room:message-expiration')
```
nickvergessen, line 42: Please merge into the talk:room:update command as suggested yesterday
Also talk:room:create should have this option

lib/Service/RoomService.php
```diff
@@ -537,8 +537,10 @@ public function verifyPassword(Room $room, string $password): array {
 		];
 	}
 
-	public function setMessageExpiration(Room $room, Participant $participant, int $seconds): void {
-		$event = new ModifyRoomEvent($room, 'messageExpiration', $seconds, null, $participant);
+	public function setMessageExpiration(Room $room, int $seconds): void {
+		$oldValue = $room->getMessageExpiration();
+		$room->setMessageExpiration($seconds);
```
nickvergessen, line 542: This was okay at the old place. That's also how it's done in all the other events of this kind

nickvergessen (changes requested):

lib/Service/RoomService.php
```diff
@@ -547,29 +547,6 @@ public function setMessageExpiration(Room $room, Participant $participant, int $
 			->where($update->expr()->eq('id', $update->createNamedParameter($room->getId(), IQueryBuilder::PARAM_INT)));
 		$update->executeStatement();
 
-		$room->setMessageExpiration($seconds);
```
nickvergessen, line 550: This one line should be kept?!🤔

nickvergessen (changes requested):

lib/Service/RoomService.php
vitormattos: No more.

vitormattos: Not to context of this method, but, we can maintain  if we will need the newest value in other methods.

nickvergessen: Well the server's response should have the new value, so yeah the object should be updated.

nickvergessen: Ah this should go into #7595 
Or should we close that and just merge this one?

tests/integration/features/command/message-expiration.feature
```diff
@@ -0,0 +1,27 @@
...
+  Background:
+    Given user "participant1" exists
+
+  Scenario: Enable message expiration and check after expire
+    Given user "participant1" creates room "room1" (v4)
+      | roomType | 3     |
+      | roomName | room1 |
+    And user "participant1" sends message "Message 1" to room "room1" with 201
+    And user "participant1" set the message expiration to 3 of room "room1" with 200 (v4)
+    And invoking occ with "talk:room:message-expiration room-name:room1 --seconds=3"
```
nickvergessen, line 11: does not exist anymore

tests/integration/features/command/message-expiration.feature
```diff
@@ -0,0 +1,27 @@
...
+    And the command output contains the text "Message expiration enabled successful as 3 seconds."
+    And user "participant1" sends message "Message 2" to room "room1" with 201
+    And wait for 3 seconds
+    And apply message expiration job manually
+    Then user "participant1" sees the following messages in room "room1" with 200
+      | room  | actorType | actorId      | actorDisplayName         | message     | messageParameters | parentMessage |
+      | room1 | users     | participant1 | participant1-displayname | Message 1   | []                |               |
+    And invoking occ with "talk:room:message-expiration room-name:room1 --seconds=0"
```
nickvergessen, line 19: does not exist anymore

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -253,6 +253,22 @@ public function userCanFindListedRoomsWithTerm(string $user, string $term, strin
 		$this->assertRooms($rooms, $formData);
 	}
 
+	/**
+	 * @When /^user "([^"]*)" get token of room "([^"]*)" \((v4)\)$/
```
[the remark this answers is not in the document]
vitormattos: Good!

tests/integration/features/command/update.feature
```diff
@@ -0,0 +1,25 @@
+Feature: update
```
nickvergessen, line 1:
```suggestion
Feature: command/update
```

tests/integration/features/command/update.feature
```diff
@@ -0,0 +1,25 @@
+Feature: command/update
+
+  Background:
+    Given user "participant1" exists
+
+  Scenario: Create a public room with message expiration time and update removing the expiration time
+    Given invoking occ with "talk:room:create room1 --user participant1 --owner participant1 --public --message-expiration=3"
+    And user "participant1" is participant of the following rooms (v4)
+      | name  |
+      | room1 |
```
nickvergessen, line 10: Maybe we check here that setting the expiration worked?
Same after the update.

Because otherwise if the commands ever break, you start deebugging the failing expiration instead of checking if expiration it self actually works.

Also since the expiration is tested elsewhere (is it? Should be in chat/something, right?) We can skip it here after checking that the value is correctly set in the conversations data returned by this request

nickvergessen (changes requested):

tests/integration/features/command/update.feature
vitormattos: > We can skip it here after checking that the value is correctly set in the conversations data returned by this request
Good!

nickvergessen (approved):

PR #7623, opened 2022-07-21
Add headline to conversation list for screenreaders

### LINKED ISSUE #7103: Add proper h2 elements for accessibility
The interface needs proper h2 elements, e.g.:
- [x] Conversation list
- [x] [Conversation title] as it shows in header
- [x] some title for the right sidebar (which is ideally not the same as the Conversation title since it would duplicate, could be "Conversation details")
- [x] Requires a new release with
    - https://github.com/nextcloud/nextcloud-vue/pull/2865
    - https://github.com/nextcloud/nextcloud-vue/pull/2845
- [ ] https://github.com/nextcloud/spreed/pull/7623 ❓ 

Extracted from https://github.com/nextcloud/spreed/issues/6801

nickvergessen:
Fix #7103 

Or is this not needed anymore as the AppNavigation already has `Conversation list` as aria label and there is only the search box in between? @jancborchardt any clue?

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -37,7 +37,7 @@
 				<ul ref="scroller"
 					class="scroller"
 					@scroll="debounceHandleScroll">
-					<AppNavigationCaption v-if="isSearching"
+					<AppNavigationCaption :class="{'hidden-visually': !isSearching}"
 						:title="t('spreed', 'Conversations')" />
 					<li role="presentation">
 						<ConversationsList ref="conversationsList"


src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -36,7 +36,7 @@
 			<div ref="scroller"
 				class="left-sidebar__list"
 				@scroll="debounceHandleScroll">
-				<AppNavigationCaption v-if="isSearching"
+				<AppNavigationCaption :class="{'hidden-visually': !isSearching}"
```
marcoambrosini, line 39:
```suggestion
				<AppNavigationCaption v-show="isSearching"
```

nickvergessen: This will not make it appear on screenreaders in case the user is not search which is the whole point of the PR?

PR #7629, opened 2022-07-22
Expire shared link when expire message

vitormattos:


lib/Chat/ChatManager.php
```diff
@@ -176,6 +176,8 @@ public function addSystemMessage(
 			$comment->setVerb(self::VERB_SYSTEM);
 		}
 
+		$this->setMessageExpiration($chat, $comment);
```
nickvergessen, line 179: This now expires all system messages, including "X added Y", not sure that is expected.
But for expiring call logs it definetly is expected, so I guess it's fine.

nickvergessen (approved):

PR #7642, opened 2022-07-25
Announce new setting sections with aria-live

### LINKED ISSUE #7622: Info text in settings dialog must be announced with aria-live
For accessibility with screen readers, texts like the following must be announced with aria-live after appearing:

image

nickvergessen:
Fix #7622


nickvergessen:
I wonder if that also applies to sections that bring it accessible stuff (like I did now), or only stuff that is not interactable?

Also there is a showSuccess which is now read out. We could add it to that as well, sounds like a solid approach with less impact if above question is answered with "only stuff that has no actions"

marcoambrosini (approved):

nickvergessen:
Not needed as it's the reading order

PR #7647, opened 2022-07-26
Move signaling to vue buttons

nickvergessen:


nickvergessen:
Removed the checkbox change (is now tracked in #7707 ) so all buttons are done with this PR

src/components/AdminSettings/SignalingServer.vue
```diff
@@ -38,17 +38,23 @@
...
-		<a v-show="!loading"
-			v-tooltip.auto="t('spreed', 'Delete this server')"
-			class="icon icon-delete"
-			@click="removeServer" />
+		<Button v-show="!loading"
+			type="tertiary-no-background"
+			:aria-label="t('spreed', 'Delete this server')"
+			@click="removeServer">
+			<template #icon>
+				<Delete />
```
raimund-schluessler, line 46: Unless you really want the icon to be the default size (`24`), you should probably set the size we agreed on kind of globally, which is `20`:
```suggestion
				<Delete :size="20"/>
```
see e.g. https://github.com/nextcloud/nextcloud-vue/blob/71468d3a42bc0e53899e31e1df6fdfb093ec3840/src/components/AppNavigationToggle/AppNavigationToggle.vue#L33

src/components/AdminSettings/SignalingServers.vue
```diff
@@ -21,17 +21,19 @@
...
+			<Button v-if="!loading && showAddServerButton"
+				class="signaling-server__add-icon"
+				type="tertiary-no-background"
+				:aria-label="t('spreed', 'Add a new high-performance backend server')"
 				@click="newServer">
-				<span class="hidden-visually">{{ t('spreed', 'Add a new server') }}</span>
-			</a>
-			<span v-else-if="loading" class="icon icon-loading-small" />
+				<template #icon>
+					<Plus />
```
raimund-schluessler, line 34: Same here, probably.

src/components/AdminSettings/SignalingServer.vue
nickvergessen: Right, also fixed stun and turn icons

src/components/AdminSettings/SignalingServer.vue
```diff
@@ -38,17 +38,23 @@
 			@change="updateVerify">
 		<label :for="'verify' + index">{{ t('spreed', 'Validate SSL certificate') }}</label>
 
-		<a v-show="!loading"
-			v-tooltip.auto="t('spreed', 'Delete this server')"
```
raimund-schluessler, line 42: Just a hint: `ButtonVue` has no tooltip by default, so if you want to keep it, you need to add it manually.

src/components/AdminSettings/SignalingServer.vue
```diff
@@ -38,17 +38,23 @@
...
+			</template>
+		</Button>
 
 		<span v-if="server">{{ connectionState }}</span>
 	</div>
 </template>
 
 <script>
+import Button from '@nextcloud/vue/dist/Components/Button'
 import Tooltip from '@nextcloud/vue/dist/Directives/Tooltip'
```
raimund-schluessler, line 56: Import can go, unless you want to keep the tooltip for the Button, in which case you need to add it to the button manually, see above.

src/components/AdminSettings/SignalingServers.vue
```diff
@@ -84,7 +86,10 @@
 
 <script>
 import SignalingServer from '../../components/AdminSettings/SignalingServer.vue'
+import Button from '@nextcloud/vue/dist/Components/Button'
 import Tooltip from '@nextcloud/vue/dist/Directives/Tooltip'
```
raimund-schluessler, line 90: Same here, and actually various other places where you switched to `Button`.

src/components/AdminSettings/SignalingServer.vue
nickvergessen: Removed

src/components/AdminSettings/SignalingServer.vue
nickvergessen: Should be fine for now if it is the accepted default behaviour. The icon should indicate it pretty obvious and screenreaders have the aria-label

src/components/AdminSettings/SignalingServers.vue
```diff
@@ -184,6 +188,21 @@ export default {
...
+		display: flex;
+		align-items: center;
+	}
+
+	&__add-icon {
+		display: inline-block;
+		width: 44px;
+		height: 44px;
+		vertical-align: middle;
+	}
```
raimund-schluessler, line 203: Is this necessary? Should be handled by the `Button`, no?

raimund-schluessler (approved):
Code looks good, I didn't test it though.

src/components/AdminSettings/SignalingServers.vue
nickvergessen: No, that's right. Not sure where it is coming from.

PR #7648, opened 2022-07-26
Add `$shallow` arg to `RoomShareProvider::getSharesInFolder`

artonge:
- [ ] https://github.com/nextcloud/server/pull/31728

A new `$shallow` argument was introduced to `IShareProvider::getSharesInFolder` in NC25 by https://github.com/nextcloud/server/pull/31728.

This PR adds this new argument and the necessary changes to make it work based on this example https://github.com/nextcloud/server/pull/31728/files#diff-dee1351696d7eae959761edf66c8e5a9052e28be73319733d857a9ab2d9fc967=

Both PRs will have failing CI as long as the other one is not merged. I am unsure which PR to merge first to limit the failing CI on both sides.

### DIFF
--- a/lib/Share/RoomShareProvider.php
+++ b/lib/Share/RoomShareProvider.php
@@ -520,10 +520,11 @@ class RoomShareProvider implements IShareProvider {
 	 * @param string $userId
 	 * @param Folder $node
 	 * @param bool $reshares Also get the shares where $user is the owner instead of just the shares where $user is the initiator
+	 * @param bool $shallow Whether the method should stop at the first level, or look into sub-folders.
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):
🐘

nickvergessen:
> Both PRs will have failing CI as long as the other one is not merged. I am unsure which PR to merge first to limit the failing CI on both sides.

Since it's an optional parameter why should it fail on this repo?

artonge:
> Since it's an optional parameter why should it fail on this repo?

Hmmm, I though linting would be stricter, perfect if it does not fail :)

Issue #7508, opened 2022-07-01
⏳ Expiring messages - API part 2


nickvergessen:
- [x] Adjust background job logic
    - [x] Adjust method name after merge of:
      https://github.com/nextcloud/server/pull/33086
    - [x] Transform the background job to run it for all rooms in one go: 
      https://github.com/nextcloud/spreed/pull/7571
- [x] Unshare attachments
  https://github.com/nextcloud/spreed/pull/7629
- [x] Allow to set expiration via console/occ
  https://github.com/nextcloud/spreed/pull/7596
- [x] Move send system message after change the expiring time to a Listener
  https://github.com/nextcloud/spreed/pull/7595
- [ ] Unit tests (if necessary)

vitormattos:
I think that all is done here.

Issue #7359, opened 2022-05-11
⏳ Expiring messages


nickvergessen:
- [x] API
    - [x] Part 1: #7327
    - [x] Part 2:
      - #7508 
      - #7571
- [x] Web https://github.com/nextcloud/spreed/pull/7689
- [ ] iOS https://github.com/nextcloud/talk-ios/issues/853
- [ ] Android https://github.com/nextcloud/talk-android/issues/2281

marcoambrosini:
Time to live -> expiring messages
This is done on a per conversation basis by moderators can set this

Options:
- 1hr minimum (due to technical limitation)
- Allow different options beyond that (maybe some quick settings and then a custom, fine tune option)

"Older messages were deleted" or some sort of info as first message in the message list

--
- [ ] Hide the feature when systemcron is not used 
- [ ] Show info about module-depending features in the admin settings

marcoambrosini:
Can this be changed once a conversation is created? 
If so let's say there's already a message history in the conversation and the moderator toggles expiring messages on. What would happen to the already present messages? cc @nickvergessen

nickvergessen:
It would affect all existing messages and basically the next cron job would expire messages that are older than the set time?
We would also document the TTL change with a system message so it's clear to all participants.

Being able to change it after creation is a must have as otherwise it's basically impossible to get it for one-to-one conversations (1. due to the flow, 2. due to the fact that you would need to get the other participant to also leave it and only then you could create it with TTL).
Also I don't think we should artificially restirct the TTL

marcoambrosini:
Would it be possible to preserve the history before the TTL has been set? Like this

<img width="708" alt="Screenshot 2022-05-16 at 12 43 38" src="https://user-images.githubusercontent.com/26852655/168576162-4e5dc48f-8baa-4790-ae0c-ff1d9f821b11.png">

nickvergessen:
That would require storing the TTL on a per message basis. Since it's not our database table and also the database table is quite big I'm not sure we can easily add a new DB column, but that would be possible in theory.

Is it like this in Signal and Telegram too?

marcoambrosini:
It's like this in WA and Signal, telegram has secret chats and regular chats

nickvergessen:
Hmm while adding a column for this we could then also add a column for meta data (to store the "silent sent" info).
The question is how happy the server/comments people are if we bloat it up for talk specifics, but we can see

vitormattos:
[...]

PR #7653, opened 2022-07-26
Improve parse system message when message expiration is enabled

vitormattos:
**Problem**: Running tests directly by API, I defined the expiration time to 3 seconds and received the message "`You set the message expiration to 0 hour`".

**Solution**: Implemented minutes and seconds in response messages.

**PS**: For now, the messages are expired using a cron job process that run in usually for each 5 minutes. Expire in X seconds only will work if the cron runs less than a minute or manually triggered.

Related with #7508


nickvergessen:
yeah seconds makes little sense and as per https://github.com/nextcloud/spreed/issues/7359#issuecomment-1123400708 the minimum we need to support is 1h.
So I would say the 0 hours is expected and we don't need to add more translation strings.

PR #7655, opened 2022-07-28
Change the page title only after the route changed

### LINKED ISSUE #6211: Browser suggests multiple (wrong) Talk conversations for same title
## Steps to reproduce
1. Use Talk a lot with different rooms
2. Open a new tab and type the name of the room you want to join
3. Be confused which to pick

### Expected behaviour
The browser suggestions should be correct.

### Actual behaviour
The same title is shown wrongly for a whole lot of conversation IDs, and only one of those is the correct one. This makes the browser suggestions essentially useless.
This is probably due to loading timing issues, guessing something with Vue setting the title? @marcoambrosini 
talk url issues

## Browser

**Microphone available:** yes

**Camera available:** yes

**Operating system:** Ubuntu 20.04

**Browser name:** Firefox

**Browser version:** 91.0.2

nickvergessen:
This is an attempt to fix the problem that multiple links are
suggested by the browser when searching for a room by the name.
The assumption is that while the URL is being changed the page
title is recorded but we currently change the title to the new
name already.

Fix #6211

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -289,7 +289,6 @@ export default {
 		})
 
 		const beforeRouteChangeListener = (to, from, next) => {
-
 			if (this.isNextcloudTalkHashDirty) {
 				// Nextcloud Talk configuration changed, reload the page when changing configuration
 				window.location = generateUrl('call/' + to.params.token)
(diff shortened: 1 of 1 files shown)


marcoambrosini (approved):
Should solve the problem

PR #7665, opened 2022-08-02
Return 400 on empty questions and options

### LINKED ISSUE #7664: Create poll endpoint should fail when there is no question attribute
### Expected behaviour
Polls without or with an empty `question` attribute should not be possible to be created.

### Actual behaviour
At the moment, the create poll endpoint allows to create polls without `question` attribute.

nickvergessen:
Fix #7664


Ivansss (approved):
Tested and works 👍

PR #7683, opened 2022-08-07
Update npm and node engines versions

nextcloud-command:
Automated update of the npm and node engines versions


nickvergessen:
Jest seems to break installation

PR #7684, opened 2022-08-08
Run PHPUnit on composer change

nickvergessen:
E.g. https://github.com/nextcloud/spreed/pull/7682 didn't run it (kind of fine, but if the php dependency we use in no-dev is updated they should run)


PVince81 (approved):
:+1:

Issue #7462, opened 2022-06-07
Limit chat history to timestamp of joining?


nickvergessen:

marcoambrosini:
I'd say it's ok if the behaviour is hardcoded for all users. And hopefully in the future it could/should be a togglable setting on a per conversation basis

vitormattos:
Telegram has this setting by room.
![Screenshot_20220722_110942](https://user-images.githubusercontent.com/1079143/180457054-6dc133b3-8c0c-4645-8d48-0692ba744a60.png)

vitormattos:
Telegram has this setting by room.
![Screenshot_20220722_110942](https://user-images.githubusercontent.com/1079143/180457054-6dc133b3-8c0c-4645-8d48-0692ba744a60.png)
@nickvergessen maybe could be possible to implement a new setting by room to define this. We only will need to define the default value.

nickvergessen:
I think default should be hidden, but I guess we can clarify that in todays call

vitormattos:
@nickvergessen looking this issue I found that I already started to implement this solution here #7686 but this PR was closed. Did you have other idea or suggestion to implement the changes to solve this issue?

vitormattos:
Hi @nickvergessen will we follow with the previous PR or do you have other idea to implement this feature?

nickvergessen:
The previous PR has a too negative impact on the performance. We need to find another way somehow.

PR #7686, opened 2022-08-09
[WIP] Feature/limit chat history

vitormattos:
API to https://github.com/nextcloud/spreed/issues/7462

- [x] Capability 
- [x] New fields and endpoints
- [x] Changes in create and update commands
- [x] Logic to limit chat history
- [x] Tests
    - [x] Shares from before the timestamp must not be accessible directly
    - [x] Shares in a public room must not be accessible either
    - [x] Shares should not be mounted for the user
    - [x] A request for older history must return 204 like when there is actually no history at all
    - [x] Getting future messages from a valid message id onwards which is before the timestamp should not reveal that the message id is valid and fail hard instead with a 404 or something like that
    - [x] Quoting a message from before via direct API call must not work
    - [x] Search must not return older messages
    - [x] Media tab must not return older items


docs/capabilities.md
```diff
@@ -104,3 +104,4 @@ title: Capabilities
 * `talk-polls` - Polls feature is available to use in the chat
 * `config => call => enabled` - Whether calling is enabled on the instance or not
 * `config => signaling => session-ping-limit` - Number of sessions the HPB is allowed to ping in the same request
+* `limit-chat-history` - Limit chat history to timestamp of joining
```
nickvergessen, line 107:
```suggestion
* `limit-show-history` - Limit chat history to timestamp of joining
```

appinfo/routes/routesRoomController.php
```diff
@@ -104,5 +104,7 @@
 		['name' => 'Room#setSIPEnabled', 'url' => '/api/{apiVersion}/room/{token}/webinar/sip', 'verb' => 'PUT', 'requirements' => $requirementsWithToken],
 		/** @see \OCA\Talk\Controller\RoomController::setMessageExpiration() */
 		['name' => 'Room#setMessageExpiration', 'url' => '/api/{apiVersion}/room/{token}/message-expiration', 'verb' => 'POST', 'requirements' => $requirements],
+		/** @see \OCA\Talk\Controller\RoomController::setPreHistory() */
+		['name' => 'Room#setPreHistory', 'url' => '/api/{apiVersion}/room/{token}/pre-history', 'verb' => 'POST', 'requirements' => $requirements],
```
nickvergessen, line 108:
```suggestion
		['name' => 'Room#setShowHistory', 'url' => '/api/{apiVersion}/room/{token}/pre-history', 'verb' => 'POST', 'requirements' => $requirements],
```

As general wording I would go with "Show history" for the feature, from the settings to API and database field.

lib/Migration/Version15000Date20220809000040.php
```diff
@@ -0,0 +1,91 @@
...
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('talk_rooms');
+		if (!$table->hasColumn('pre_history')) {
+			// As per call the default for NEW conversations should be "Hidden".
+			$table->addColumn('pre_history', Types::BOOLEAN, [
+				'notnull' => false,
+				'default' => false
+			]);
+		}
```
nickvergessen, line 61: I would vote to go with a SMALLINT
This allows us e.g. to extend the feature so that logged in/added users see the full history and guests only the stuff while they joined.

lib/Migration/Version15000Date20220809000040.php
```diff
@@ -0,0 +1,91 @@
...
+
+		$table = $schema->getTable('talk_attendees');
+		if (!$table->hasColumn('history_since')) {
+			$table->addColumn('history_since', Types::DATETIME_MUTABLE, [
```
nickvergessen, line 65: Would name this something like "joined_since" or something alike.

I wondered if we want to store the time or the message id. Don't know what would be more handy later on.

nickvergessen (changes requested):

lib/Migration/Version15000Date20220809000040.php
vitormattos: > I wondered if we want to store the time or the message id. Don't know what would be more handy later on.

About using ID:
* If the group have "expire message" enabled, the join message will be deleted, and we will have an ID pointing to an invalid message.

About using time:
* Is clearer to identify the date limit in attendee table/instance and if we have a date time, is because this user can't will get messages before this date, if we don't have, is because we won't stop to delivery old messages.

I implemented the logic to cut the message in a loop that we do to organize the message list and when we found a message greater than join date, I break the loop and follow to delivery the previous parsed messages.

I implemented in the loop because to cut in query will be nessary to change https://github.com/nextcloud/server/blob/master/lib/private/Comments/Manager.php#L471 to define the filter as greater than the ID (or date) of join message when we have this information. This will put more complexity in this method, and I think that this maybe is unnecessary.

lib/Migration/Version15000Date20220809000040.php
vitormattos: I like to think in the future and make the code easier to be upgraded.
Yes, for now, numeric value solve the problem and make possible to be like an enumerate with other possible values different from 1 and null.

lib/Controller/ChatController.php
```diff
@@ -434,11 +434,18 @@ public function receiveMessages(int $lookIntoFuture,
...
 		$i = 0;
 		$messages = $commentIdToIndex = $parentIds = [];
+		$historySince = $attendee->getHistorySince();
 		foreach ($comments as $comment) {
 			$id = (int) $comment->getId();
 			$message = $this->messageParser->createMessage($this->room, $this->participant, $comment, $this->l);
 			$this->messageParser->parseMessage($message);
 
+			if ($historySince) {
+				if ($message->getComment()->getCreationDateTime() < $historySince) {
```
nickvergessen, line 444: This makes an infinite request loop when scrolling too far up. In case the end is reached, we need to abort with the same return code as if there are no messages at all. See lines ~420+ with the `if (empty($comments)) {`

nickvergessen: Maybe instead of doing this here, you should add a array_filter method before that if statement.

Also we need to validate that `$lastKnownMessageId` is not older than the timestamp.

vitormattos: I think that was solved in last commit.

lib/Controller/ChatController.php
```diff
@@ -432,20 +432,15 @@ public function receiveMessages(int $lookIntoFuture,
 			return $response;
 		}
 
+		$comments = $this->filterHistorySince($comments, $attendee);
```
nickvergessen, line 435: NEeds to go before `if (empty($comments)) {` so you return if there is nothing left

vitormattos: right

vitormattos:
@nickvergessen when I defined that the default behavior is to don't see the history before join date, all integration tests that have the join date message in return when list the history broke.
Example:
https://github.com/nextcloud/spreed/blob/master/tests/integration/features/reaction/react.feature#L7-L19
In this case, the message `conversation_created` won't return.

What you suggest fixing this?
 1. Change all broke tests removing the `conversation_created`
 2. Change the create room command in all broke tests and set to make the history visible

vitormattos:
> Getting future messages from a valid message id onwards which is before the timestamp should not reveal that the message id is valid and fail hard instead with a 404 or something like that

@nickvergessen about this topic, what you think to only don't return the messages that the user haven't permission to see?

Example:

user1 created conversation
Message 1
message 2
user2 added
user2 retrieve messages in the future starting from message "user1 created conversation"
Then, the user2 only will see "user2 added"

lib/Chat/ChatManager.php
```diff
@@ -716,4 +721,37 @@ private function searchIsPartOfConversationNameOrAtAll(string $search, string $r
...
 	}
+
+	/**
+	 * If the attendee only can see the history after join date,
+	 * will remove all comments before join date.
+	 *
+	 * @param Room $room The room of comments
+	 * @param array IComment[] $comments
+	 * @param Attendee $attendee
+	 * @return array
```
nickvergessen, line 732: **Required:** 
```suggestion
	 * @return IComment[]
```

lib/Chat/ChatManager.php
```diff
@@ -716,4 +721,37 @@ private function searchIsPartOfConversationNameOrAtAll(string $search, string $r
...
+	 */
+	public function filterHistorySince(Room $room, array $comments, Attendee $attendee): array {
+		// Filter nothing if the room setting is to show history to all
+		if ($room->getShowHistory() === 1) {
+			return $comments;
+		}
+		// Filter nothing if the user haven't the history since.
```
nickvergessen, line 739: **Comment:**
```suggestion
		// Filter nothing if the attendee doesn't have the history since.
```

lib/Chat/ChatManager.php
```diff
@@ -716,4 +721,37 @@ private function searchIsPartOfConversationNameOrAtAll(string $search, string $r
...
+		// This will occur for all user added before the manager limit the history to new participants
```
nickvergessen, line 740: **Comment:**
```suggestion
		// This will occur for all attendees added before the history was limited
```

lib/Chat/ChatManager.php
```diff
@@ -716,4 +721,37 @@ private function searchIsPartOfConversationNameOrAtAll(string $search, string $r
...
+		$historySince = $attendee->getHistorySince();
+		if (!$historySince) {
```
nickvergessen, line 742: **Required:** 
Does this work for \DateTime objects? Or is it null?

lib/Model/Attendee.php
```diff
@@ -60,6 +60,8 @@
  * @method null|string getAccessToken()
  * @method void setRemoteId(string $remoteId)
  * @method string getRemoteId()
+ * @method null|int getHistorySince()
```
nickvergessen, line 63: **Required:** 

Settings is `\DateTime` and getting is int?

lib/Chat/ChatManager.php
```diff
@@ -716,4 +721,37 @@ private function searchIsPartOfConversationNameOrAtAll(string $search, string $r
...
+			return $comments;
+		}
+		foreach ($comments as $key => $comment) {
+			// Ignore if the comment isn't of current room
+			if ($comment->getObjectId() != $room->getId()) {
+				continue;
+			}
```
nickvergessen, line 749: **Comment:**
This is for search? Feels very weird, but maybe we can add the explanation why this could be the case

lib/Chat/ChatManager.php
```diff
@@ -716,4 +721,37 @@ private function searchIsPartOfConversationNameOrAtAll(string $search, string $r
 	public function deleteExpiredMessages(): void {
 		$this->commentsManager->deleteCommentsExpiredAtObject('chat', '');
 	}
+
+	/**
+	 * If the attendee only can see the history after join date,
+	 * will remove all comments before join date.
```
nickvergessen, line 727: **Comment:**
```suggestion
	 * Filter the history if the attendee joined the room later and the history is restricted
```

lib/Chat/ChatManager.php
```diff
@@ -716,4 +721,37 @@ private function searchIsPartOfConversationNameOrAtAll(string $search, string $r
...
+	 *
+	 * @param Room $room The room of comments
+	 * @param array IComment[] $comments
+	 * @param Attendee $attendee
+	 * @return array
+	 */
+	public function filterHistorySince(Room $room, array $comments, Attendee $attendee): array {
+		// Filter nothing if the room setting is to show history to all
+		if ($room->getShowHistory() === 1) {
```
nickvergessen, line 736: **Required:** 
should use a constant

lib/Controller/RoomController.php
```diff
@@ -412,6 +412,7 @@ protected function formatRoomV4(Room $room, ?Participant $currentParticipant, ?a
 			'listable' => Room::LISTABLE_NONE,
 			'callFlag' => Participant::FLAG_DISCONNECTED,
 			'messageExpiration' => 0,
+			'showHistory' => false,
```
nickvergessen, line 415: **Required:** 
should be an constant int of the acceptable settings.

lib/Service/RoomService.php
```diff
@@ -551,4 +551,13 @@ public function setMessageExpiration(Room $room, int $seconds): void {
 
 		$this->dispatcher->dispatch(Room::EVENT_AFTER_SET_MESSAGE_EXPIRATION, $event);
 	}
+
+	public function setShowHistory(Room $room, int $showHistory): void {
```
nickvergessen, line 555: **Required:** 

Needs to throw if the value is not 0 or 1 for now.
Should dispatch evens like with all other room attributes as well

lib/Service/RoomService.php
```diff
@@ -551,4 +551,13 @@ public function setMessageExpiration(Room $room, int $seconds): void {
...
+		$update = $this->db->getQueryBuilder();
+		$update->update('talk_rooms')
+			->set('show_history', $update->createNamedParameter($showHistory, IQueryBuilder::PARAM_INT));
```
nickvergessen, line 558: **Required:** 💥 

missing limitation for the room id?!

lib/Controller/RoomController.php
```diff
@@ -1746,4 +1748,13 @@ public function setMessageExpiration(int $seconds): DataResponse {
 		$this->roomService->setMessageExpiration($this->room, $seconds);
 		return new DataResponse();
 	}
+
+	/**
+	 * @PublicPage
+	 * @RequireModeratorParticipant
+	 */
+	public function setShowHistory(int $showHistory): DataResponse {
+		$this->roomService->setShowHistory($this->room, $showHistory);
```
nickvergessen, line 1757: **Required:** 
Since the service can throw after the comment there was resolved, it should be catched here and return a 400 Bad Request

docs/conversation.md
```diff
@@ -361,6 +362,23 @@
...
+* Data:
+
+| field        | type | Description                                       |
+|--------------|------|---------------------------------------------------|
+| `showHistory` | bool | Hide/unhide message history for new participants |
+
+* Response:
+    - Status code:
+        + `200 OK`
+        + `403 Forbidden` When the current user is not a moderator, owner or guest moderator
```
nickvergessen, line 379: **Required:** 

```suggestion
        + `200 OK`
        + `400 Bad Request` When the provided showHistory parameter is invalid.
        + `403 Forbidden` When the current user is not a moderator, owner or guest moderator
```

After resolving the comments on the RoomController

docs/conversation.md
```diff
@@ -361,6 +362,23 @@
...
+## Toggle Pre history
+
+* Required capability: `limit-show-history`
+* Method: `POST`
+* Endpoint: `/room/{token}/show-history`
+* Data:
+
+| field        | type | Description                                       |
+|--------------|------|---------------------------------------------------|
+| `showHistory` | bool | Hide/unhide message history for new participants |
```
nickvergessen, line 374: **Required:** 
```suggestion
| `showHistory` | int | Limit message history for new participants |
```

lib/Controller/ChatController.php
```diff
@@ -417,6 +417,8 @@ public function receiveMessages(int $lookIntoFuture,
 			$comments = $this->chatManager->getHistory($this->room, $lastKnownMessageId, $limit, (bool) $includeLastKnown);
 		}
 
+		$comments = $this->chatManager->filterHistorySince($this->room, $comments, $attendee);
```
nickvergessen, line 420: **Comment:**
The filtering like this is a bit sad. It means we potentially load a lot of data from the database which we then throw away.
But I guess adding it to the comments API then creates issues with index usage, etc. So let's keep it for now.

docs/capabilities.md
```diff
@@ -105,3 +105,5 @@ title: Capabilities
 * `config => call => enabled` - Whether calling is enabled on the instance or not
 * `config => signaling => session-ping-limit` - Number of sessions the HPB is allowed to ping in the same request
 * `config => signaling => hello-v2-token-key` - Public key to use when verifying JWT auth tokens for hello V2.
+* `limit-chat-history` - Limit chat history to timestamp of joining
+* `limit-show-history` - Limit chat history to timestamp of joining
```
nickvergessen, line 109: **Required:** 
```suggestion
* `limit-show-history` - Chat history can be limited for guests/participants to the moment of joining.
```

lib/Migration/Version15000Date20220809000040.php
```diff
@@ -0,0 +1,90 @@
...
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('talk_rooms');
+		if (!$table->hasColumn('show_history')) {
+			// As per call the default for NEW conversations should be "Hidden".
```
nickvergessen, line 56: **Comment:**
What call? I know you mean the feature discussion call on Monday, but that is not relevant for the source code:
```suggestion
			// Defaulting to "Hidden" for existing conversations.
```

But then again, the column is null able, so it will be null instead of 0 which would be your constant.

lib/Migration/Version15000Date20220809000040.php
```diff
@@ -0,0 +1,90 @@
...
+			$table->addColumn('show_history', Types::SMALLINT, [
+				'notnull' => false
```
nickvergessen, line 58: **Required:** 

```suggestion
				'notnull' => true,
				'default' => 1,
```

lib/Migration/Version15000Date20220809000040.php
nickvergessen: Okay, so we use the time point. Should be okay.


Any reason to use DateTime instead of bigint like we use for other time points across talk?

lib/Migration/Version15000Date20220809000040.php
```diff
@@ -0,0 +1,90 @@
...
+
+		$table = $schema->getTable('talk_rooms');
+		if ($table->hasColumn('show_history')) {
+			// For existing conversations we should fall back to "Visible".
+			$update = $this->connection->getQueryBuilder();
+			$update->update('talk_rooms')
+				->set('show_history', $update->createNamedParameter(1, IQueryBuilder::PARAM_INT));
+			$update->executeStatement();
+		}
+	}
```
nickvergessen, line 89: **Required:**
Can be dropped after the column definition above is adjusted

```suggestion
```

lib/Search/MessageSearch.php
```diff
@@ -139,8 +139,18 @@ public function search(IUser $user, ISearchQuery $query): SearchResult {
...
 
+		$filteredComments = [];
+		foreach ($roomMap as $room) {
+			$participant = $room->getParticipant($user->getUID(), false);
+			$attendee = $participant->getAttendee();
+			$filteredComments = array_merge(
+				$filteredComments,
+				$this->chatManager->filterHistorySince($room, $comments, $attendee)
+			);
+		}
```
nickvergessen, line 150: **Required:** 

With array_merged you readd them all the time?

```suggestion
		$filteredComments = $comments;
		foreach ($roomMap as $room) {
			$participant = $room->getParticipant($user->getUID(), false);
			$attendee = $participant->getAttendee();
			$filteredComments = $this->chatManager->filterHistorySince($room, $filteredComments, $attendee);
		}
```

lib/Share/RoomShareProvider.php
```diff
@@ -582,46 +582,70 @@ public function getSharesInFolder($userId, Folder $node, $reshares, $shallow = t
 	 */
 	public function getSharesBy($userId, $shareType, $node, $reshares, $limit, $offset): array {
 		$qb = $this->dbConnection->getQueryBuilder();
-		$qb->select('*')
-			->from('share');
+		$qb->select('share.*', 'ta.history_since', 'tr.show_history')
+			->from('share', 'share')
+			->leftJoin('share', 'talk_rooms', 'tr', $qb->expr()->eq('tr.token', 'share.share_with'))
+			->leftJoin('tr', 'talk_attendees', 'ta',
```
nickvergessen, line 588: **Comment:**

This query is executed very often and we are now left joining 2 other tables.
I wonder how big the performance impact is. Maybe we can simulate the query on our instance to measure it a bit.

lib/Share/RoomShareProvider.php
```diff
@@ -582,46 +582,70 @@ public function getSharesInFolder($userId, Folder $node, $reshares, $shallow = t
...
+			$shares[] = $share;
 		}
 		$cursor->closeCursor();
 
 		return $shares;
 	}
 
+	private function participantCanSeeShareSince(IShare $share, array $data): bool {
+		if (!empty($data['show_history']) && !empty($data['history_since'])) {
+			if ($data['show_history'] === null && $data['history_since']) {
```
nickvergessen, line 638: **Comment:**
this should be catched by the empty-line above already?

lib/Share/RoomShareProvider.php
```diff
@@ -779,14 +803,23 @@ public function getSharedWith($userId, $shareType, $node, $limit, $offset): arra
...
 				'f.fileid', 'f.path', 'f.permissions AS f_permissions', 'f.storage', 'f.path_hash',
 				'f.parent AS f_parent', 'f.name', 'f.mimetype', 'f.mimepart', 'f.size', 'f.mtime', 'f.storage_mtime',
-				'f.encrypted', 'f.unencrypted_size', 'f.etag', 'f.checksum'
+				'f.encrypted', 'f.unencrypted_size', 'f.etag', 'f.checksum',
+				'ta.history_since', 'tr.show_history'
 			)
 				->selectAlias('st.id', 'storage_string_id')
 				->from('share', 's')
+				->leftJoin('s', 'talk_rooms', 'tr', $qb->expr()->eq('tr.token', 's.share_with'))
+				->leftJoin('tr', 'talk_attendees', 'ta',
```
nickvergessen, line 812: **Comment:**
joining 5 tables in this query 🙈

lib/Share/RoomShareProvider.php
```diff
@@ -779,14 +803,23 @@ public function getSharedWith($userId, $shareType, $node, $limit, $offset): arra
...
+					$qb->expr()->andX(
+						$qb->expr()->eq('ta.room_id', 'tr.id'),
+						$qb->expr()->eq('ta.actor_id', $qb->createNamedParameter($userId))
```
nickvergessen, line 815: **Required:**
That's not enough. You need to check the actor type as well

lib/Share/RoomShareProvider.php
```diff
@@ -582,46 +582,70 @@ public function getSharesInFolder($userId, Folder $node, $reshares, $shallow = t
...
 		$qb = $this->dbConnection->getQueryBuilder();
-		$qb->select('*')
-			->from('share');
+		$qb->select('share.*', 'ta.history_since', 'tr.show_history')
+			->from('share', 'share')
+			->leftJoin('share', 'talk_rooms', 'tr', $qb->expr()->eq('tr.token', 'share.share_with'))
+			->leftJoin('tr', 'talk_attendees', 'ta',
+				$qb->expr()->andX(
+					$qb->expr()->eq('ta.room_id', 'tr.id'),
+					$qb->expr()->eq('ta.actor_id', $qb->createNamedParameter($userId))
```
nickvergessen, line 591: **Required:** 

Not enough, needs to check the actor_type too

nickvergessen (changes requested):

lib/Controller/ChatController.php
vitormattos: Will be a bit complicated to refactor the comments API.
This was the simplest way that I found. I filtered the comments before all the others processing, I think that this won't impact the performance because only will occur in the end of history.

lib/Migration/Version15000Date20220809000040.php
vitormattos: More human-readable only looking in the database without necessity to do conversions and because I identified other points that use DateTime instead of bigint.

vitormattos: Looking the result of tests I think that we will need to use the ID.

https://drone.nextcloud.com/nextcloud/spreed/9390/2/4
https://drone.nextcloud.com/nextcloud/spreed/9388/2/4
https://drone.nextcloud.com/nextcloud/spreed/9387/2/4

This isn't different failures in different tests, is the same failure:
- I used >= to filter the history using time. But, if we have two actions in the same second of join date, the new user will see the two messages.
If we only use >, the solution to this problem, the join message will never be displayed.
I don't like to use ID in this case, as I explained before, but I think is the solution at this moment.

vitormattos: Will be good to think more. I think that the solution will be to store both data.

PR #7689, opened 2022-08-09
Add UI option to set the message expiration

nickvergessen:
Bildschirmfoto vom 2022-08-09 14-47-59

Ref https://github.com/nextcloud/spreed/issues/7359

### DIFF
--- a/lib/Chat/Parser/SystemMessage.php
+++ b/lib/Chat/Parser/SystemMessage.php
@@ -457,9 +457,10 @@ class SystemMessage {
 				$parsedMessage = $this->l->t('You deleted a reaction');
 			}
 		} elseif ($message === 'message_expiration_enabled') {
-			$weeks = $parameters['seconds'] > (86400 * 7) ? (int) round($parameters['seconds'] / (86400 * 7)) : 0;
-			$days = $parameters['seconds'] > 86400 ? (int) round($parameters['seconds'] / 86400) : 0;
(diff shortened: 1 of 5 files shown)


danxuliu (approved):
Tested and works :+1:

As far as I can see using `@update:value` rather than `@input` (which, if I understand it correctly, should avoid the need to use the `overwriteExpiration` value) is needed to avoid the UI jumping between values, as the store is not updated until the server value is updated. Maybe the store should be updated as soon as the value is set and then reverted if it failed to solve that, but I am fine with the current approach.

Besides that, I have added fixup commits to move an import to its own line, rename `ConversationSettings` to `ExpirationSettings`, rename `e` to `expiration` and `error`, remove console logs, get the last value from the array rather than use a magic constant, rename `expirationOperation` to `expirationOptions`, and show an explicit option when a custom time is set instead of showing _1 Hour_. If any of the fixups is not need feel free to drop them before squashing.

nickvergessen:
> rename ConversationSettings to ExpirationSettings

I did this intentionally as that is the only section having multiple subcomponents and I think we should plan to combine them going forward. but whatever...

nickvergessen:
Fixed eslint and squashed

> /home/runner/work/spreed/spreed/src/components/ConversationSettings/ExpirationSettings.vue
Error:   42:8  error  Component name `ConversationSettings` should match file name `ExpirationSettings`  vue/match-component-file-name

PR #7700, opened 2022-08-10
Move reactions to vue buttons

nickvergessen:
Before | After
---|---
Bildschirmfoto vom 2022-08-10 14-21-57 | Bildschirmfoto vom 2022-08-10 14-21-35


nickvergessen:
Tests fixed

danxuliu (approved):
Tested and works :+1:

I have added a fixup to remove some unneeded rules. The ones applied to the internal elements are leftovers from a previous version according to @nickvergessen, and the one for the border color is not needed due to the border color being set as a box-shadow in the Vue button.

PR #7704, opened 2022-08-10
Remove the unused nc-button class

nickvergessen:
Before | After
---|---
Bildschirmfoto vom 2022-08-10 16-31-44 | Bildschirmfoto vom 2022-08-10 16-37-11


src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -561,7 +561,7 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-@import '../../assets/buttons';
+@import '../../assets/variables';
```
nickvergessen, line 564: Was indirectly imported before and is needed to avoid:
```
Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
SassError: Undefined variable.
    ╷
609 │             opacity: $opacity_full;
    │                      ^^^^^^^^^^^^^
```

danxuliu (approved):
Tested and works :+1:

Note that, once you hover on the alert icon, the button to send the message again will never be hidden. However this issue was there before this pull request; it is probably caused by [how the `showReloadButton` is set](https://github.com/nextcloud/spreed/blob/700315785144ad3661177d7212e42431e7edd94c/src/components/MessagesList/MessagesGroup/Message/Message.vue#L78-L82) (maybe the button swallows the blur event and it does not bubble up to the parent div or something like that).

PR #7719, opened 2022-08-11
Fix voting for option 0

nickvergessen:
The problem is the entity had 0 as a default it would skip writing
it into the changeset of the query for the insert.
But the DB had a default of 1 so it would store a 1 afterwards in the db.

Having null as default on the Vote model fixes this, as then any number
is a change from the default, the mark-dirty happens and we add it to
the query.

The problem is integration tests always send integers due to the form body usage.
The `0` from the integration tests therefor hits the server as `'0'`, === does not match
and the mark dirty happens:
https://github.com/nextcloud/server/blob/1d30fb7852e61cf9a2fcd03e83fa193c6aa5c827/lib/public/AppFramework/Db/Entity.php#L106-L109
Therefor the option was always specified from integration tests and it set a 0.

### DIFF
--- a/lib/Migration/Version15000Date20220503121308.php
+++ b/lib/Migration/Version15000Date20220503121308.php
@@ -132,7 +132,6 @@ class Version15000Date20220503121308 extends SimpleMigrationStep {
 			$table->addColumn('option_id', Types::INTEGER, [
 				'notnull' => true,
 				'length' => 6,
-				'default' => 1,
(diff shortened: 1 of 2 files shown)


marcoambrosini (approved):
many thank yous

vitormattos (approved):
Isn't a big problem, but developers of clients will need to do this change in database handmade.

PR #3399, opened 2020-04-27
Handle pasting screenshots to the chat message form

guryanovev:
Fix #2172

PR #7743, opened 2022-08-13
Reduce number of queries required to read shares

starypatyk:
When diagnosing [slow loading of chat messages](https://github.com/nextcloud/talk-android/issues/2228) in the Talk Android app, I have noticed that information about files (usually images) shared in the chat is read using a separate query for each file.

This is an initial proof-of-concept that reduces it to a single query.

Instead of executing a separate query in **RoomShareProvider::getShareById** for each share, we gather all share ids and execute a single query in a new method **RoomShareProvider::getSharesByIds**.

While performance improvements in the development environment are difficult to notice, on my family "production" environment this gives around 15% improvement of the **ChatController::receiveMessages** execution time (**280 ms** down from **330 ms**).

This code is certainly **not** production-ready and requires refactoring/clean-up. However I would like to get some feedback, if this looks like a reasonable way to go forward.

Trying to read info about all shared items together also seems to be a pre-requisite to eventually optimize **SystemMessage::getFileFromShare**. Currently additional 3-4 database queries are performed for each shared file (by the Files app if I understand correctly).

Thoughts?

### DIFF
--- a/lib/Controller/ChatController.php
+++ b/lib/Controller/ChatController.php
@@ -40,6 +40,7 @@ use OCA\Talk\Room;
 use OCA\Talk\Service\AttachmentService;
 use OCA\Talk\Service\ParticipantService;
 use OCA\Talk\Service\SessionService;
+use OCA\Talk\Share\RoomShareProvider;
 use OCP\App\IAppManager;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\DataResponse;
@@ -74,6 +75,7 @@ class ChatController extends AEnvironmentAwareController {
 	/** @var string[] */
 	protected array $guestNames;
 	private MessageParser $messageParser;
+	protected RoomShareProvider $shareProvider;
 	private IManager $autoCompleteManager;
 	private IUserStatusManager $statusManager;
 	protected MatterbridgeManager $matterbridgeManager;
@@ -97,6 +99,7 @@ class ChatController extends AEnvironmentAwareController {
 								AttachmentService $attachmentService,
 								GuestManager $guestManager,
 								MessageParser $messageParser,
+								RoomShareProvider $shareProvider,
 								IManager $autoCompleteManager,
 								IUserStatusManager $statusManager,
 								MatterbridgeManager $matterbridgeManager,
@@ -119,6 +122,7 @@ class ChatController extends AEnvironmentAwareController {
 		$this->attachmentService = $attachmentService;
 		$this->guestManager = $guestManager;
 		$this->messageParser = $messageParser;
+		$this->shareProvider = $shareProvider;
 		$this->autoCompleteManager = $autoCompleteManager;
 		$this->statusManager = $statusManager;
 		$this->matterbridgeManager = $matterbridgeManager;
@@ -431,6 +435,22 @@ class ChatController extends AEnvironmentAwareController {
 			}
 			return $response;
 		}
+
+		// Quickly scan messages for share IDs
+		$share_ids = [];
+		foreach ($comments as $comment) {
+			$verb = $comment->getVerb();
+			$message = $comment->getMessage();
+			if ($verb === 'object_shared') {
+				$data = json_decode($message, true);
+				$parameters = $data['parameters'];
+				$share_ids[] = $parameters['share'];
+			}
+		}
+		// Ignore the result for now. Retrieved Share objects will be cached by
+		// the RoomShareProvider and returned from the cache to
+		// the MessageParser without additional database queries.
+		$this->shareProvider->getSharesByIds($share_ids);
 
 		$i = 0;
 		$messages = $commentIdToIndex = $parentIds = [];
--- a/lib/Controller/ChatController.php
+++ b/lib/Controller/ChatController.php
@@ -312,8 +312,10 @@ class ChatController extends AEnvironmentAwareController {
 	/*
 	 * Gather share IDs from the comments and preload share definitions
 	 * to avoid separate database query for each individual share.
+	 *
+	 * @param IComment[] $comments
 	 */
-	protected function preloadShares(array $comments) {
+	protected function preloadShares(array $comments): void {
 		// Scan messages for share IDs
 		$shareIds = [];
 		foreach ($comments as $comment) {
--- a/lib/Controller/ChatController.php
+++ b/lib/Controller/ChatController.php
@@ -437,20 +437,23 @@ class ChatController extends AEnvironmentAwareController {
 		}
 
 		// Quickly scan messages for share IDs
-		$share_ids = [];
+		$shareIds = [];
 		foreach ($comments as $comment) {
 			$verb = $comment->getVerb();
-			$message = $comment->getMessage();
 			if ($verb === 'object_shared') {
+				$message = $comment->getMessage();
 				$data = json_decode($message, true);
-				$parameters = $data['parameters'];
-				$share_ids[] = $parameters['share'];
+				if (isset($data['parameters']['share'])) {
+					$shareIds[] = $data['parameters']['share'];
+				}
 			}
 		}
 		// Ignore the result for now. Retrieved Share objects will be cached by
 		// the RoomShareProvider and returned from the cache to
 		// the MessageParser without additional database queries.
-		$this->shareProvider->getSharesByIds($share_ids);
+		if (!empty($shareIds)) {
+			$this->shareProvider->getSharesByIds($shareIds);
+		}
 
 		$i = 0;
 		$messages = $commentIdToIndex = $parentIds = [];
--- a/lib/Controller/ChatController.php
+++ b/lib/Controller/ChatController.php
@@ -309,6 +309,31 @@ class ChatController extends AEnvironmentAwareController {
 		return $this->parseCommentToResponse($comment);
 	}
 
+	/*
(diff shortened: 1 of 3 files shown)


nickvergessen:
Can you give some rough numbers about your chats? Are they mostly files only? On our company instance there are roughly 7k files shared between 2m chat messages. So the cases where so many files are in a single chunk that duplicating the method is worth the improvements are rather limited

starypatyk:
Hi @nickvergessen,

My "production" Nextcloud instance is a family server used primarily to share photos, so images are very frequently shared in chats. The server runs for over 8 years now (it started as Owncloud), but we use Talk for only about 1.5 year.

The statistics are:

- **~2200** total messages.
- **~420** files shared (`SELECT COUNT(*) FROM oc_share WHERE share_type = 10 -- SHARE_TYPE_ROOM`) - but please note that many of these are shared in group chats.

In two sample chats I have **33** and **40** images in the last 100 messages.

On the other hand, in my day job (where we use Teams), I see that we also send a lot of images (primarily screenshots). As I understand the [Handle pasting screenshots to the chat message form](https://github.com/nextcloud/spreed/pull/3399) feature also uses files sharing. Same applies to [Image keyboard support in chat](https://github.com/nextcloud/talk-android/pull/1749) that I had implemented at the beginning of this year in the Android client - to enliven the conversations. :wink:

Maybe I am heavily biased, but for me images are an important part of chat conversations.

nickvergessen:
> Maybe I am heavily biased, but for me images are an important part of chat conversations.

Totally fine, we only have a Meme channel with a that high rate.
Since we have a performance related workgroup this week your changes came just in time.
I'll have a look at it during the week I guess.

starypatyk:
Thanks @nickvergessen!

Please note that the code is currently of "proof-of-concept" quality.

I have not been able to find a way to gather all share IDs using the existing infrastructure (MessageParser::parseMessage -> Listener::parseSystemMessage -> SystemMessage::parseMessage -> SystemMessage::getFileFromShare). That's why there is a hack at the top level method (ChatController::receiveMessages) going directly to the RoomShareProvider.

nickvergessen:
I deployed it on our testing copy for now

Before | After
---|---
170 | 158

so yeah at the moment it is not helping too much.
Will look if we can bring the info collecting on a further level

lib/Share/RoomShareProvider.php
```diff
@@ -631,6 +635,11 @@ public function getSharesBy($userId, $shareType, $node, $reshares, $limit, $offs
 	 * @throws ShareNotFound
 	 */
 	public function getShareById($id, $recipientId = null): IShare {
+
+		if (isset($this->sharesByIdCache[$id])) {
```
nickvergessen, line 639: Cache needs to be invalidated on delete/update function calls

lib/Share/RoomShareProvider.php
```diff
@@ -665,6 +674,51 @@ public function getShareById($id, $recipientId = null): IShare {
...
+	 * Get shares by ids
+	 *
+	 * Not part of IShareProvider API, but needed by OCA\Talk\Controller\ChatController.
+	 *
+	 * @param int[] $id
+	 * @param string|null $recipientId
+	 * @return IShare[]
+	 * @throws ShareNotFound
+	 */
+	public function getSharesByIds($ids, $recipientId = null): array {
```
nickvergessen, line 687:
```suggestion
	public function getSharesByIds(array $ids, ?string $recipientId = null): array {
```

nickvergessen (changes requested):

nickvergessen:
> That's why there is a hack at the top level method (ChatController::receiveMessages) going directly to the RoomShareProvider.

I would say that is even okay, could even make the function name reflect that: `preloadSharesByIds` or something alike

lib/Controller/ChatController.php
```diff
@@ -432,6 +436,22 @@ public function receiveMessages(int $lookIntoFuture,
...
 
+		// Quickly scan messages for share IDs
+		$share_ids = [];
+		foreach ($comments as $comment) {
+			$verb = $comment->getVerb();
+			$message = $comment->getMessage();
+			if ($verb === 'object_shared') {
+				$data = json_decode($message, true);
+				$parameters = $data['parameters'];
+				$share_ids[] = $parameters['share'];
```
nickvergessen, line 447: Need to check with isset, so other shares like deck cards, locations, etc don't break the response.

```suggestion
		$shareIds = [];
		foreach ($comments as $comment) {
			$verb = $comment->getVerb();
			if ($verb === 'object_shared') {
				$message = $comment->getMessage();
				$data = json_decode($message, true);
				if (isset($data['parameters']['share'])) {
					$shareIds[] = $data['parameters']['share'];
				}
```

lib/Share/RoomShareProvider.php
```diff
@@ -28,6 +28,7 @@
 
 namespace OCA\Talk\Share;
 
+use OC\Cache\CappedMemoryCache;
```
nickvergessen, line 31:
```suggestion
use OCP\Cache\CappedMemoryCache;
```

Should use the public one

lib/Share/RoomShareProvider.php
```diff
@@ -83,6 +84,8 @@ class RoomShareProvider implements IShareProvider {
 	private IL10N $l;
 	private IMimeTypeLoader $mimeTypeLoader;
 
+	private $sharesByIdCache;
```
nickvergessen, line 87:
```suggestion
	private CappedMemoryCache $sharesByIdCache;
```

nickvergessen:
With those 2 PRs in addition our Meme channel is down to 69 queries:
- https://github.com/nextcloud/spreed/pull/7762
- https://github.com/nextcloud/server/pull/33578

lib/Share/RoomShareProvider.php
starypatyk: I have added cache invalidation (cleaning actually :wink:) in all update/delete calls that seemed related in https://github.com/nextcloud/spreed/pull/7743/commits/03f64768cc8219e7718980d7a6eef4d227b83de7.

starypatyk:
@nickvergessen - thanks for your comments. I have already applied most of your suggestions.

Regarding the remaining suggestion:
> > That's why there is a hack at the top level method (ChatController::receiveMessages) going directly to the RoomShareProvider.
> 
> I would say that is even okay, could even make the function name reflect that: `preloadSharesByIds` or something alike

I would like to reduce code duplication between current `getShareById` and `getSharesByIds` methods. When looking at these more carefully, I have noticed that my code has two errors:

- does not check for shares referring to deleted files - `isAccessibleResult`
- may skip `resolveSharesForRecipient`

I am still working on this.

starypatyk:
@nickvergessen - I have refactored getShareById/getSharesByIds in https://github.com/nextcloud/spreed/pull/7743/commits/e8bd174ddccdad989e087f8f9428e718b4bf500b.

I hope that I have fixed the issues with inaccessible shares and `resolveSharesForRecipient`.

Could you check this please? The code became a bit convoluted and I am far from saying that I have checked all possible scenarios.

nickvergessen:
Looks good.

If you have troubles satisfying the CI jobs we can also take over at the end.

starypatyk:
I have managed to fix all failing tests.

I am wondering, if it makes sense to write an additional test that would check that the `oc_share` table is only read once during execution of the **ChatController::receiveMessages** method. For sure, I do not understand the test/CI infrastructure well enough to do it myself. :wink:

lib/Share/RoomShareProvider.php
```diff
@@ -631,6 +652,35 @@ public function getSharesBy($userId, $shareType, $node, $reshares, $limit, $offs
...
+	}
+
+	/**
+	 * Get shares by ids
+	 *
+	 * Not part of IShareProvider API, but needed by OCA\Talk\Controller\ChatController.
+	 *
+	 * @param int[] $id
+	 * @param string|null $recipientId
+	 * @return (IShare|false)[]
```
nickvergessen, line 681: You don't really return false?

nickvergessen:
> I am wondering, if it makes sense to write an additional test that would check that the oc_share table is only read once during execution of the ChatController::receiveMessages method. For sure, I do not understand the test/CI infrastructure well enough to do it myself. :wink:

I kind of dislike unit tests on that level.
I would leave this as a todo for a follow up. My idea would be to add it to https://github.com/nextcloud/spreed/tree/master/tests/integration/features/scaling

Where we have some tests were we run locally with total query counting enabled and that would then reveal an overall picture.

lib/Share/RoomShareProvider.php
starypatyk: Indeed! :+1: Thanks for pointing this out. Corrected by https://github.com/nextcloud/spreed/pull/7743/commits/26497b415680e2f7330c665a85142a076aab1f21 and https://github.com/nextcloud/spreed/pull/7743/commits/aa9921d461ef89d42b3da763c08747dcd99e207e.

starypatyk:
@nickvergessen - I do not plan any more changes - unless you have additional comments. Please feel free to merge.

nickvergessen (approved):
Added some last doc fixes, now lets wait for CI and then merge

nickvergessen:
Thanks a lot @starypatyk for the patch and the good analysis!

PR #7745, opened 2022-08-14
Fix app settings dialog style

raimund-schluessler:
Requires https://github.com/nextcloud/nextcloud-vue/pull/3035.

@nickvergessen Could you check whether this fixes the settings modal scroll issue for you?


nickvergessen (approved):
Works

PR #7747, opened 2022-08-16
Fix the talk dashboard icon on dark mode

nickvergessen:
Before | After
---|---
Bildschirmfoto vom 2022-08-16 10-01-04 | Bildschirmfoto vom 2022-08-16 09-59-55

The old icon is defined in server which toggles the icon based on the a11y theming, but the dashboard additionally has a `filter: var(--background-invert-if-dark)` so the icon was dark again.

### DIFF
--- a/css/icons.css
+++ b/css/icons.css
@@ -79,3 +79,7 @@
 .app-files .app-sidebar__close.forced-white {
 	color: #ffffff;
 }
+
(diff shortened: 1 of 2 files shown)


skjnldsv (approved):
Ah right, it is white by default

marcoambrosini (approved):

PR #7748, opened 2022-08-16
Make postgres integration tests reliable

nickvergessen:
Ref https://drone.nextcloud.com/nextcloud/spreed/9142/21/4

### DIFF
--- a/tests/integration/features/command/user-remove.feature
+++ b/tests/integration/features/command/user-remove.feature
@@ -16,7 +16,7 @@ Feature: User remove
     And the command output contains the text "Users successfully removed from all rooms"
     Then the command was successful
     And user "participant2" is participant of the following rooms (v4)
-    And user "participant1" is participant of the following rooms (v4)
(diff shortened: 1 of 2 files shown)


tests/integration/features/command/user-remove.feature
```diff
@@ -8,6 +8,7 @@ Feature: User remove
     Given user "participant1" creates room "room1" (v4)
       | roomType | 1 |
       | invite   | participant2 |
+    And sleep 1
```
nickvergessen, line 11: Okay, after this test is seems like the problem is not creating the rooms, but the room list is sorted by last chat message and that order is mixed up for postgres as there is no sleep between their generation.

danxuliu (approved):
I have not run the tests, but code looks good :+1: 

Just a little nitpick, at first I was a bit confused about `$shouldOrder`; maybe it would match better the `(unordered )?` part if it was named `$isResponseUnordered` or something like that? Anyway now that I got it I am fine too with `$shouldOrder`.

PR #7750, opened 2022-08-16
Use DisplayNameCache

CarlSchwan:
Reduce query by 1/3 when fetching the list of rooms on the production
instance.

TODO:
- [x] Use dependency injections
- [x] Adapt tests
- [x] Make DisplayNameCache part of OCP https://github.com/nextcloud/server/pull/32461

### DIFF
--- a/lib/Chat/MessageParser.php
+++ b/lib/Chat/MessageParser.php
@@ -34,7 +34,6 @@ use OCA\Talk\Room;
 use OCP\Comments\IComment;
 use OCP\EventDispatcher\IEventDispatcher;
 use OCP\IL10N;
-use OCP\IUser;
 use OCP\IUserManager;
 
 /**
@@ -79,7 +78,7 @@ class MessageParser {
 		$actorId = $comment->getActorId();
 		$displayName = '';
 		if ($comment->getActorType() === Attendee::ACTOR_USERS) {
-			$displayName = $this->userManager->getDisplayName($comment->getActorId());
+			$displayName = $this->userManager->getDisplayName($comment->getActorId()) ?? $comment->getActorId();
 		} elseif ($comment->getActorType() === Attendee::ACTOR_BRIDGED) {
 			$displayName = $comment->getActorId();
 			$actorId = MatterbridgeManager::BRIDGE_BOT_USERID;
--- a/lib/Chat/MessageParser.php
+++ b/lib/Chat/MessageParser.php
@@ -79,8 +79,8 @@ class MessageParser {
 		$actorId = $comment->getActorId();
 		$displayName = '';
 		if ($comment->getActorType() === Attendee::ACTOR_USERS) {
-			$user = $this->userManager->get($comment->getActorId());
-			$displayName = $user instanceof IUser ? $user->getDisplayName() : $comment->getActorId();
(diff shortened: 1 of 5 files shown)


lib/Manager.php
```diff
@@ -990,8 +990,7 @@ public function resolveRoomDisplayName(Room $room, string $userId): string {
 
 			foreach ($users as $participantId) {
 				if ($participantId !== $userId) {
-					$user = $this->userManager->get($participantId);
-					$otherParticipant = $user instanceof IUser ? $user->getDisplayName() : $participantId;
+					$otherParticipant = $$this->userManager->getDisplayName($participantId);
```
nickvergessen, line 993:
```suggestion
					$otherParticipant = $this->userManager->getDisplayName($participantId);
```

nickvergessen (changes requested):

lib/Chat/Parser/SystemMessage.php
```diff
@@ -698,11 +698,16 @@ protected function getRemoteUser(string $federationId): array {
 	}
 
 	protected function getDisplayName(string $uid): string {
+		$userName = $this->userManager->getDisplayName($uid);
+		if ($userName !== $uid) {
+			return $userName;
+		}
+
+		// Verify that the user actually exists
```
icewind1991, line 706: Do we actually need this? for users without a display name set specifically this will always hit

nickvergessen: Well we need to know if the user exists.
Talked already with Carl if the `getDisplayName` method could throw or return null in such cases instead of returning the uid

CarlSchwan: Yeah fixed now

lib/Chat/MessageParser.php
```diff
@@ -79,8 +79,7 @@ protected function setActor(Message $message): void {
 		$actorId = $comment->getActorId();
 		$displayName = '';
 		if ($comment->getActorType() === Attendee::ACTOR_USERS) {
-			$user = $this->userManager->get($comment->getActorId());
-			$displayName = $user instanceof IUser ? $user->getDisplayName() : $comment->getActorId();
+			$displayName = $this->userManager->getDisplayName($comment->getActorId());
```
nickvergessen, line 82:
```suggestion
			$displayName = $this->userManager->getDisplayName($comment->getActorId()) ?? $comment->getActorId();
```

lib/Manager.php
```diff
@@ -990,8 +990,7 @@ public function resolveRoomDisplayName(Room $room, string $userId): string {
 
 			foreach ($users as $participantId) {
 				if ($participantId !== $userId) {
-					$user = $this->userManager->get($participantId);
-					$otherParticipant = $user instanceof IUser ? $user->getDisplayName() : $participantId;
+					$otherParticipant = $this->userManager->getDisplayName($participantId);
```
nickvergessen, line 993:
```suggestion
					$otherParticipant = $this->userManager->getDisplayName($participantId) ?? participantId;
```

nickvergessen (approved):

PVince81 (approved):
:+1:

PR #7752, opened 2022-08-16
Ensure that the reactions array is always a JSON object

### LINKED ISSUE #7751: reactions in Talk API sometimes delivered as array, sometimes map
If there are no reactions to a message, that message will be delivered with `reactions: []`, which is an array of some kind. In the case where there is a reaction, the value is delivered as `reactions: {"\u2764\ufe0f":1}` which is an object/map, not an array.

PHP might parse this just fine, but in Go it's trickier (and i imagine same thing for all statically typed languages). Go expects to always get the same data type. I suppose it's possible to work around it (haven't tried and I don't use reactions right now anyway), but I think it is an example of bad API design to use different types depending on the value (empty or non-empty).

### How to use GitHub

---

## Steps to reproduce
1. Send 2 messages in Talk
2. React to one of the messages with an emoji
3. Get the message using Talk API
4. Notice the different types of the `reactions` field for the 2 messages

### Expected behaviour
The `reactions` field should have the same type always.

### Actual behaviour
When `reactions` is empty, it's delivered as an array. When it's non-empty, it's delivered as a json object/map.

## Talk app

14 (Nextcloud 24)

nickvergessen:
Fix #7751 

- [x] Confirmation from @mahibi that it does not break Android
- [x] Confirmation from @Ivansss that it does not break iOS

### DIFF
--- a/lib/Model/Message.php
+++ b/lib/Model/Message.php
@@ -172,6 +172,13 @@ class Message {
 	public function toArray(): array {
 		$expireDate = $this->getComment()->getExpireDate();
 
+		$reactions = $this->getComment()->getReactions();
+		if (empty($reactions)) {
+			// Cheating here to make sure the reactions array is always a
+			// JSON object on the API, even when there is no reaction at all.
+			$reactions = new \StdClass();
+		}
+
 		$data = [
 			'id' => (int) $this->getComment()->getId(),
 			'token' => $this->getRoom()->getToken(),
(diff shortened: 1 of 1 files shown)


mahibi:
> * Confirmation from @mahibi that it does not break Android

tested and works

mahibi:
>     * [ ]  Confirmation from @mahibi that it does not break Android

tested and works

Ivansss (approved):
Tested and works 👍
(needs rebase)

nickvergessen:
Rebased to solve the conflict with #7771

vitormattos (approved):

nickvergessen:
Manual backport in https://github.com/nextcloud/spreed/pull/7899

PR #7761, opened 2022-08-17
Polls cleanup

marcoambrosini:
- [x] Add some feedback about what the user voted
- [x] Disallow submit empty vote
- [x] Display user's vote when going back to the voting page if user has voted

Ref #7795

### DIFF
--- a/lib/Chat/Parser/SystemMessage.php
+++ b/lib/Chat/Parser/SystemMessage.php
@@ -410,6 +410,7 @@ class SystemMessage {
 			}
 		} elseif ($message === 'object_shared') {
 			$parsedParameters['object'] = $parameters['metaData'];
+			$parsedParameters['object']['id'] = (string) $parsedParameters['object']['id'];
 			$parsedMessage = '{object}';
 
 			if (isset($parsedParameters['object']['type'])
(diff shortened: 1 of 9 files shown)


nickvergessen:
- [ ] I'm not sure anymore what the discussion about forcing focus was, but after "Add answer" should we focus the new input field? Or how likely is it that someone wants to use enter multiple times to create the answer fields first. Only matters for keyboard navigation as the mouse needs to be repositioned anyway.
- [ ] After you just voted the modal has no closing X and 3 confusing options:
    - Dismiss => Will it revert my vote? Maybe it can be migrated to a closing X in the top right?
    - Change your vote
    - Close poll => It will close the modal, right? 😛 => Should be "End poll" as per feature review meeting we had with Android back then
    - Also the buttons have issues with translation length. Missing padding on first button and 3 lines on the second one just as samples from German: 
Bildschirmfoto vom 2022-08-25 16-59-53
- [ ] When you voted there is no indication what you voted for. You have to use the "Edit" screen to see if you want to edit. See next screen, I voted **3** only. But all options with votes are primary color and presented the same way
- [ ] Percentages should be rounded to 2 digits?
Bildschirmfoto vom 2022-08-25 17-06-31
- [ ] When a poll was ended, the system message should be visible
- [ ] When a poll was ended it should say "Poll ended" not "You voted"
Bildschirmfoto vom 2022-08-25 17-11-08
- [ ] When someone else ends the poll and you then open it, you see the voting options, etc. Submitting yields an error in the top right which is okay, but there is no indication why that is. So the poll details need to be refreshed when the system message is received (and in case of ending it needs to be immediately not after 5 secs), so editing is aborted and not visible anymore.
- [ ] For non-moderators instead of "0 votes" we should say "You voted" in the summary line of private polls when the user voted already

PR #7762, opened 2022-08-17
Delay getting the node until we know that we use the data

nickvergessen:
Removes 2 queries for each file that is shared from another user

Meme channel:
Before | After
---|---
142 | 86


PVince81 (approved):
:+1: makes sense

PVince81:
some related tests are failing and might need adjusting

skjnldsv (approved):

PR #7767, opened 2022-08-18
Hide deprecated projects in shared items tab by default

Pytal:
- [x] Deprecate projects which is superseded by Related resources https://github.com/nextcloud/server/issues/28320
- [x] Add empty content to Media tab


nickvergessen (dismissed):
Works technically, but now the tab is completely empty.
I think we can continue from here onwards and the talk team has to add an empty content thing for the media view.

nickvergessen:
Rebased and added the empty content.

marcoambrosini (approved):

PR #7790, opened 2022-08-25
l10n: Change style of sentence

Valdnet:

### DIFF
--- a/src/components/ConversationSettings/ExpirationSettings.vue
+++ b/src/components/ConversationSettings/ExpirationSettings.vue
@@ -22,7 +22,7 @@
 <template>
 	<div>
 		<div class="app-settings-section__hint">
-			{{ t('spreed', 'Chat messages expire after a certain time. Note: Files shared in chat will not be deleted for the owner, but will no longer be shared in the conversation.') }}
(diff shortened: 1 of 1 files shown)


Valdnet:
@rakekniven Thanks, my oversight 😉.

src/components/ConversationSettings/ExpirationSettings.vue
```diff
@@ -22,7 +22,7 @@
 <template>
 	<div>
 		<div class="app-settings-section__hint">
-			{{ t('spreed', 'Expire chat messages after a certain time. Note: files shared into the chat will only be unshared from the conversation but are not deleted for the owner.') }}
+			{{ t('spreed', 'Chat messages expire after a certain time. Note: Files shared in chat will not be deleted for the owner, but will no longer be shared in the conversation.') }}
```
nickvergessen, line 25: Well they only do when you configure it. Default is Off. Maybe we go with:

```suggestion
			{{ t('spreed', 'Chat messages can be expired after a certain time. Note: Files shared in chat will not be deleted for the owner, but will no longer be shared in the conversation.') }}
```

marcoambrosini (approved):

PR #7788, opened 2022-08-23
Fix notifications when lobby is active

nickvergessen:
Fix #7775 

- [x] Call notifications
- [x] Chat notifications
- [x] Remove existing notifications when lobby is enabled afterwards or permissions are removed

PR #7794, opened 2022-08-25
[stable24] Fix notifications when lobby is active

nickvergessen:
Backport of #7788 

Auto backport failed because the callapi/notifications.feature file was only introduced with silent call which is not in 24


danxuliu (approved):
Tested and works :+1:

PR #7802, opened 2022-08-26
Fix email invite SIP Dial-In system message users

### LINKED ISSUE #7691: Wrong name in system message when guest starts call through dialin
## Steps to reproduce
1. Create room with SIP enabled and invite an external user by email.
2. Dialin as invited user with meeting id and pin.

### Expected behaviour
There should be a system message showing that the invited user that started a call.

### Actual behaviour
There is a system message "**Deleted user** has started a call" ("**Gelöschter Benutzer** hat einen Anruf begonnen"). The entry in the participant list gets a phone icon, so the mapping works.

This happens with Talk 14.0.3 and the latest version of the SIP bridge.

nickvergessen:
Fix #7691

### DIFF
--- a/lib/Chat/Parser/SystemMessage.php
+++ b/lib/Chat/Parser/SystemMessage.php
@@ -367,7 +367,7 @@ class SystemMessage {
 				$parsedMessage = $this->l->t('An administrator demoted {user} from moderator');
 			}
 		} elseif ($message === 'guest_moderator_promoted') {
-			$parsedParameters['user'] = $this->getGuest($room, $parameters['session']);
+			$parsedParameters['user'] = $this->getGuest($room, Attendee::ACTOR_GUESTS, $parameters['session']);
 			$parsedMessage = $this->l->t('{actor} promoted {user} to moderator');
 			if ($currentUserIsActor) {
 				$parsedMessage = $this->l->t('You promoted {user} to moderator');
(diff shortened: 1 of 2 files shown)


fancycode:
Tested, now shows `Guest started a call` (as is if dialed in without a pin). Is this expected or should it be the email address? Imho should be the email address as the user was invited.

nickvergessen:
Fixed unit tests and made emails return "Email (guest)" as default

fancycode:
Still shows only `Guest` for me. I updated to your latest commit, restarted php-fpm and rebuilt the JS. Browser cache is cleared. Anything else I'm missing?

nickvergessen:
Maybe git add on my site solves it 🤦🏼 
```
$ git status
On branch bugfix/7691/fix-email-invite-sipdial-in-systemmessages
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   lib/Chat/Parser/SystemMessage.php

```

fancycode (approved):
Tested and the email is now shown.

However when the call ends, the message says `Call with 0 guests` or `Call with Username and 0 guests`, i.e. the phone participant is not counted. Should I open a new issue for this or do you want to fix it here?

nickvergessen:
> However when the call ends, the message says Call with 0 guests or Call with Username and 0 guests, i.e. the phone participant is not counted. Should I open a new issue for this or do you want to fix it here?

New issue please, thanks :)

fancycode:
> > However when the call ends, the message says Call with 0 guests or Call with Username and 0 guests, i.e. the phone participant is not counted. Should I open a new issue for this or do you want to fix it here?
> 
> New issue please, thanks :)

https://github.com/nextcloud/spreed/issues/7825

PR #7820, opened 2022-08-29
Show voters details in polls

marcoambrosini:

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/Poll.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/Poll.vue
@@ -532,6 +532,7 @@ export default {
 
 	&__details {
 		display: flex;
+		height: 32px;
 	}
 
 	&-subtitle {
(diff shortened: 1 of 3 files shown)


src/components/MessagesList/MessagesGroup/Message/MessagePart/Poll.vue
```diff
@@ -130,9 +130,14 @@
...
 							</div>
-							<p v-if="selfHasVotedOption(index)" class="results__option-subtitle">
-								{{ t('spreed','You voted') }}
-							</p>
+							<div v-if="getFilteredDetails(index).length > 0 || selfHasVotedOption(index)"
+								class="results__option__details">
+								<PollVotersDetails v-if="details"
+									:details="getFilteredDetails(index)" />
+								<p v-if="selfHasVotedOption(index)" class="results__option-subtitle">
+									{{ t('spreed','You voted') }}
```
nickvergessen, line 138: Need to use a different string here I guess. Translation is the same as when the poll is open but you voted and e.g. the german text is unfitting to match both cases.
Maybe we need to say "You voted for this option"?

nickvergessen (approved):
Looks good in general

nickvergessen:
Aborted drone as there was no PHP change

Merging

PR #7834, opened 2022-08-31
Fix voted guests handling

nickvergessen:
* src/components/AvatarWrapper/AvatarWrapper.vue Reverting the changes from #7820 and move it to AvatarWrapperSmall
* Fix guest avatars not showing in the small size

Before | After
---|---
Bildschirmfoto vom 2022-08-31 17-07-12 | Bildschirmfoto vom 2022-08-31 17-04-55

### DIFF
--- a/src/components/AvatarWrapper/AvatarWrapper.vue
+++ b/src/components/AvatarWrapper/AvatarWrapper.vue
@@ -21,10 +21,7 @@
 
 <template>
 	<div class="avatar-wrapper"
-		:class="{
-			'offline': offline,
-			'avatar-wrapper--condensed': condensed,
-		}">
+		:class="{'offline': offline}">
 		<div v-if="iconClass"
 			class="icon"
 			:class="[`avatar-${sizeToString}px`, iconClass]" />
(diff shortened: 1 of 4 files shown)


danxuliu (commented):
Tested and works :+1: 

The avatar size [was 24px](https://github.com/nextcloud/spreed/pull/7834/files#diff-dbeee8b80b700c0dfefc71fba0145247a86cb1a566706d8220ccfece2457e2acL36) and now it [is 22px](https://github.com/nextcloud/spreed/pull/7834/files#diff-2a2cfb998f4e4232f95f4a13ba3f38b65f6fea3ee4f54403069e2b5341a7e868R132). Probably not important, but just in case.

Independently of that, the avatar of unnamed guests does not show `?`, and the border colour does not separate between two avatars, like the white border (or black border in dark mode, see original screenshot) of user avatars does:
Talk-Poll-Result-Guest-Avatars

nickvergessen:
Both fixed now

Bildschirmfoto vom 2022-09-01 06-44-22

danxuliu (approved):
Tested and works :+1:

PR #7868, opened 2022-09-07
Adjust videos border radius

marcoambrosini:


danxuliu (approved):
Tested and works :+1: 

For consistency I have added a fixup commit to also adjust the border of the "videos" shown in the development mode of the grid.

Independently of that, I have noticed that the followed participant is not "framed" like the current speaker in the speaker mode. Is that known/expected or should I open a new issue for that?
Talk-Call-Speaker-Mode-Default-Framed
Talk-Call-Speaker-Mode-Following-Not-Framed

PR #7869, opened 2022-09-07
Fix sidebar overlap with callview

marcoambrosini:


danxuliu (approved):
Tested and works :+1:

PR #7871, opened 2022-09-08
Fix visibility of sidebar toggle

### LINKED ISSUE #7872: Sidebar button not visible
## Steps to reproduce
1. Open a conversation
2. Try to open the sidebar

### Expected behaviour
You see a sidebar button

### Actual behaviour
It’s nowhere to be found. @marcoambrosini as discussed in the design call. :) Happens with tech-preview and latest master.
image

marcoambrosini:
Using the new syntax for slots

fixes #7872

### DIFF
--- a/src/components/TopBar/TopBar.vue
+++ b/src/components/TopBar/TopBar.vue
@@ -75,10 +75,11 @@
 				:aria-label="t('spreed', 'Conversation actions')"
 				:container="container"
 				@shortkey.native="toggleFullscreen">
-				<span slot="icon"
-					:class="{'top-bar__button__force-white': isInCall}">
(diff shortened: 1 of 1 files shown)


danxuliu (approved):
Tested and works :+1: 

From what I could see the only missing icon was the one to open the sidebar, the others were fine before this pull request. I guess they were changed too just for consistency, but I mention it just in case.

And is all that expected (that the new syntax is needed only for the icon to open the sidebar and the rest are fine with the old syntax)? Or something else needs to be fixed, for example in _nextcloud-vue_?

marcoambrosini:
Not sure why it's needed only there, maybe because of some nested slot magic in the actionbuttons?

PR #7873, opened 2022-09-08
Update callview backdrop

marcoambrosini:
Screenshot 2022-09-08 at 14 31 17

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -510,6 +510,10 @@ export default {
 
 .content {
 	&.in-call {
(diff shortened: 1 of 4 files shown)


danxuliu (commented):
The blur filter adds a noticeable lag in Chromium. This is the same issue that we had in the past that we tried to [work around](https://github.com/nextcloud/spreed/pull/4678) but ended having to [completly remove the blur filter](https://github.com/nextcloud/spreed/pull/4985).

PR #7876, opened 2022-09-08
Fix action menus being cut

marcoambrosini:
Use default container (body)


marcoambrosini:
Not the right fix.. The popovers are appended to the body and the tabbing order is messed up

nickvergessen:
Keep the container and just remove the boundary element?

nickvergessen (approved):

PR #7895, opened 2022-09-12
Fix username overlap with localvideo

marcoambrosini:

Screenshot 2022-09-12 at 09 41 31


nickvergessen:
Can we fix it without z-index? Why is it needed in first place? :P

nickvergessen:
Okay we need it because of 
https://github.com/nextcloud/spreed/blob/e8d2e2767aaffd48c07199b410559f09f3e50705/src/components/CallView/shared/VideoBottomBar.vue#L283
...

PR #7912, opened 2022-09-14
Fix deleted users in polls

### LINKED ISSUE #7839: Handling poll votes of users after they got deleted
### How to use GitHub

---

## Steps to reproduce
1. Create a poll as admin A
2. As user U vote in the poll
3. As admin A delete user U

### Expected behaviour
Experience and displaying of remaining bits of user U should be consistent

### Actual behaviour
Display name is still there, actor type is still user so we try to render the avatar which does not exist
Bildschirmfoto vom 2022-09-01 08-18-40

nickvergessen:
Fix #7839 

Before | After
---|---
 | Bildschirmfoto vom 2022-09-14 08-36-14

### DIFF
--- a/lib/Listener/UserDeletedListener.php
+++ b/lib/Listener/UserDeletedListener.php
@@ -24,15 +24,19 @@ declare(strict_types=1);
 namespace OCA\Talk\Listener;
 
 use OCA\Talk\Manager;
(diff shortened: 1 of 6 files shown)


marcoambrosini (approved):
🐘

PR #7914, opened 2022-09-14
Allow to "Send call notification" for owners

### LINKED ISSUE #7907: Moderators can not "Recall" owner
### How to use GitHub

---

## Steps to reproduce
1. As user A create a room
2. Invite user B and promote to moderator
3. As user B start a call and try to "Resend call notification"

### Expected behaviour
Works

### Actual behaviour
Actions menu is not there

nickvergessen:
Fix #7907

### DIFF
--- a/src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
+++ b/src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
@@ -49,7 +49,7 @@
 		<div class="participant-row__user-wrapper"
 			:class="{
 				'has-call-icon': callIcon,
-				'has-menu-icon': canBeModerated && !isSearched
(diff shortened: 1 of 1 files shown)


danxuliu (approved):
Tested and works :+1: 

One detail, though; [the top separator is added only when the permissions are shown](https://github.com/nextcloud/spreed/blob/02e83a61dc246c3130a6ed8bd2e969996b96d0b8/src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue#L130-L131), so there is no separator between `Demote from moderator` and `Send call notification`. Should the separator be added in that case too (and for resending mail invitations)? In any case it can be done in a follow up, as it is not caused by this fix.
Talk-Participant-Menu-Separator-Moderator
Talk-Participant-Menu-Separator-User

PR #7917, opened 2022-09-14
Fix flow and chat permissions

### LINKED ISSUE #7476: Flow should not write messages when the user has no chat permission
### How to use GitHub

---

## Steps to reproduce
1. Configure a flow with a room where you are not the owner
2. As the owner remove chat permission for the user or make the room read only

### Expected behaviour
Flow does not post anymore

### Actual behaviour
From reading the code the flow would still post, e.g. allowing pupils to spam a class room

nickvergessen:
Fix #7476

### DIFF
--- a/lib/Flow/Operation.php
+++ b/lib/Flow/Operation.php
@@ -29,6 +29,7 @@ use OCA\Talk\Chat\ChatManager;
 use OCA\Talk\Exceptions\ParticipantNotFoundException;
 use OCA\Talk\Exceptions\RoomNotFoundException;
 use OCA\Talk\Manager as TalkManager;
+use OCA\Talk\Model\Attendee;
 use OCA\Talk\Participant;
 use OCA\Talk\Room;
 use OCP\EventDispatcher\Event;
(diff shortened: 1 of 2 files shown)


lib/Flow/Operation.php
```diff
@@ -116,7 +117,17 @@ public function onEvent(string $eventName, Event $event, IRuleMatcher $ruleMatch
 				}
 
 				$room = $this->getRoom($token, $uid);
+				if (!($room->getReadOnly() !== Room::READ_WRITE)) {
```
come-nc, line 120:
```suggestion
				if ($room->getReadOnly() === Room::READ_WRITE) {
```

come-nc (changes requested):

lib/Flow/Operation.php
nickvergessen: yeah, it was a bug before. Didn't mean to invert it. read-write is okay, read-only is what we want to filter out.

AndyScherzinger (approved):
🐘

come-nc (approved):

PR #7939, opened 2022-09-15
Don't return deleted files

### LINKED ISSUE #7238: Deleted shared items should not be returned
## Steps to reproduce
1. Share a picture in a chat
2. Remove that picture
3. Fetch share items for that conversation

### Expected behaviour
Share items API should not return deleted items

### Actual behaviour
Share items API returns deleted items

vitormattos:
Close #7238


lib/Chat/Parser/SystemMessage.php
```diff
@@ -403,6 +403,7 @@ public function parseMessage(Message $chatMessage): void {
 					$chatMessage->setMessageType(ChatManager::VERB_MESSAGE);
 				}
 			} catch (\Exception $e) {
+				$chatMessage->setVisibility(false);
```
nickvergessen, line 406: This is too generic. We still want it in the chat. Just not in the right sidebar.

My idea would be to filter them out in `getObjectsSharedInRoomOverview` and `getObjectsSharedInRoom`

We can detect it by checking the system message attribute for `file_shared` and then not having a `file` entry in the message parameter list

lib/Controller/ChatController.php
```diff
@@ -468,6 +468,11 @@ public function receiveMessages(int $lookIntoFuture,
 		$i = 0;
 		$messages = $commentIdToIndex = $parentIds = [];
 		foreach ($comments as $comment) {
+			$message = $comment->getMessage();
+			if ($this->messageParser->isSharedFile($message) && !$this->messageParser->fileOfMessageExists($message)) {
+				continue;
+			}
```
nickvergessen, line 474:
```suggestion
```

This endpoint should work as before.

Only those 2:
* getObjectsSharedInRoomOverview
* getObjectsSharedInRoom

Should filter them out

lib/Controller/ChatController.php
```diff
@@ -790,6 +795,10 @@ public function getObjectsSharedInRoomOverview(int $limit = 7): DataResponse {
 		$comments = $this->chatManager->getMessagesById($this->room, array_merge(...array_values($messageIdsByType)));
 
 		foreach ($comments as $comment) {
+			if (!$this->messageParser->fileOfMessageExists($comment->getMessage())) {
```
nickvergessen, line 798:
```suggestion
			if ($this->messageParser->isSharedFile($comment->getMessage()) && !$this->messageParser->fileOfMessageExists($comment->getMessage())) {
```

lib/Controller/ChatController.php
```diff
@@ -841,12 +850,17 @@ public function getObjectsSharedInRoom(string $objectType, int $lastKnownMessage
 		return $response;
 	}
 
-	protected function getMessagesForRoom(Room $room, array $messageIds): array {
+	private function getMessagesForRoom(Room $room, array $messageIds): array {
 		$comments = $this->chatManager->getMessagesById($room, $messageIds);
 
 		$messages = [];
 		foreach ($comments as $comment) {
+			if (!$this->messageParser->fileOfMessageExists($comment->getMessage())) {
```
nickvergessen, line 858:
```suggestion
			if ($this->messageParser->isSharedFile($comment->getMessage()) && !$this->messageParser->fileOfMessageExists($comment->getMessage())) {
```

lib/Controller/ChatController.php
```diff
@@ -841,12 +850,17 @@ public function getObjectsSharedInRoom(string $objectType, int $lastKnownMessage
 		return $response;
 	}
 
-	protected function getMessagesForRoom(Room $room, array $messageIds): array {
+	private function getMessagesForRoom(Room $room, array $messageIds): array {
```
nickvergessen, line 853:
```suggestion
	protected function getMessagesForRoom(Room $room, array $messageIds): array {
```
Please leave all methods and members always protected unless it is really necessary to make something private.

lib/Chat/MessageParser.php
```diff
@@ -103,4 +107,30 @@ protected function setActor(Message $message): void {
 			$displayName
 		);
 	}
+
+	public function isSharedFile(string $message): bool {
+		$parameters = $this->getParametersFromMessage($message);
+		return !empty($parameters['share']);
+	}
+
+	private function getParametersFromMessage(string $message): array {
```
nickvergessen, line 116:
```suggestion
	protected function getParametersFromMessage(string $message): array {
```

lib/Chat/MessageParser.php
```diff
@@ -103,4 +107,30 @@ protected function setActor(Message $message): void {
...
+		return $data['parameters'];
+	}
+
+	public function fileOfMessageExists(string $message): bool {
+		$parameters = $this->getParametersFromMessage($message);
+		if (empty($parameters['share'])) {
+			return false;
+		}
+		try {
+			$this->shareProvider->getShareById($parameters['share']);
```
nickvergessen, line 130: This is not scaling.

Please call this method in the controllers entry points before the loops:
```php
$this->preloadShares($comments);
```
Similar to this:
https://github.com/nextcloud/spreed/blob/8a4bd006212beafd6e1c5593b46a7bffd0bdcf17/lib/Controller/ChatController.php#L466

nickvergessen (changes requested):

lib/Controller/ChatController.php
vitormattos: Won't we have only messages as shared file here?

lib/Controller/ChatController.php
vitormattos: Won't we have only messages as shared file here?

lib/Controller/ChatController.php
nickvergessen: `receiveMessages` which is the normal chat view. And we have intergration tests for it and they all fail now.

lib/Chat/MessageParser.php
nickvergessen: We use protected basically everywhere in talk to make it easier to extend the code and inherit it.

> $ find lib/ -iname "*.php" | xargs grep "protected function" | wc -l
> 247
> $ find lib/ -iname "*.php" | xargs grep "private function" | wc -l
> 57

so unless really necessary (and it's not here) please use protected and never use private

lib/Controller/ChatController.php
nickvergessen: No, it's also contacts, polls, locations, ....

lib/Controller/ChatController.php
nickvergessen: see above

nickvergessen (changes requested):

lib/Chat/MessageParser.php
vitormattos: This is a good command to list the files that could be outside this pattern and a good start point to define a rule, put in the linter and create a GitHub issue to check if all private methods are really necessary to be private.

lib/Chat/ChatManager.php
```diff
@@ -716,4 +716,45 @@ private function searchIsPartOfConversationNameOrAtAll(string $search, string $r
...
+	}
+
+	/**
+	 * When receive a list of comments, filter the comments,
+	 * removing all that have shares of file that no more exists
+	 *
+	 * @param IComment[] $comments
+	 * @return IComment[]
+	 */
+	public function removeFileNotExists(array $comments): array {
```
nickvergessen, line 750:
```suggestion
	public function filterCommentsWithNonExistingFiles(array $comments): array {
```

lib/Controller/ChatController.php
```diff
@@ -783,35 +783,52 @@ public function getObjectsSharedInRoomOverview(int $limit = 7): DataResponse {
...
-			$message = $this->messageParser->createMessage($this->room, $this->participant, $comment, $this->l);
-			$this->messageParser->parseMessage($message);
-
-			if (!$message->getVisibility()) {
-				continue;
-			}
-
-			$messages[(int) $comment->getId()] = $message->toArray($this->getResponseFormat());
-		}
+		$messages = $this->getFilteredMessages(array_merge(...array_values($messageIdsByType)));
```
nickvergessen, line 792:
```suggestion
		$messages = $this->getMessagesForRoom($room, array_merge(...array_values($messageIdsByType)));
```

lib/Controller/ChatController.php
```diff
@@ -783,35 +783,52 @@ public function getObjectsSharedInRoomOverview(int $limit = 7): DataResponse {
...
+
+			if (!$message->getVisibility()) {
+				continue;
+			}
+
+			$messages[(int) $comment->getId()] = $message->toArray($this->getResponseFormat());
+		}
+		return $messages;
+	}
+
```
nickvergessen, line 831:
```suggestion
```

nickvergessen (changes requested):

nickvergessen:
> Conflicting files
> lib/Chat/MessageParser.php

nickvergessen (approved):

PR #7940, opened 2022-09-16
Fix poll voter details

### LINKED ISSUE #7836: Only 8 voters are shown per option even when "expanding" the list
### How to use GitHub

---

Similar issue as with reactions back then. Because a tooltip alike popover is used, it is not possible to see the full result list and who voted for what.
* Currently only 8 avatars are shown 🆗 
* When expanding by clicking them you can see 8 avatars and displaynames 🚯 

In order to use it for recorded votings in bigger rooms there needs to be a way to see the full list of who voted what.

marcoambrosini:
Signed-off-by: Marco <marcoambrosini@icloud.com>

fix #7836 

Screenshot 2022-09-16 at 09 40 06

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/PollVotersDetails.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/PollVotersDetails.vue
@@ -24,7 +24,7 @@
 		<button slot="trigger"
 			tabindex="0"
 			class="poll-voters-details">
-			<AvatarWrapperSmall v-for="(item, index) in details"
+			<AvatarWrapperSmall v-for="(item, index) in details.slice(0, 8)"
 				:id="item.actorId"
 				:key="index"
 				:source="item.actorType"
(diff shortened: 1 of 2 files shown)


src/components/MessagesList/MessagesGroup/Message/MessagePart/PollVotersDetails.vue
```diff
@@ -24,7 +24,7 @@
 		<button slot="trigger"
 			tabindex="0"
 			class="poll-voters-details">
-			<AvatarWrapperSmall v-for="(item, index) in details"
+			<AvatarWrapperSmall v-for="(item, index) in details.slice(0, 8)"
```
nickvergessen, line 27: Should we have a guest avatar with `…` in case there where more than the slice?

marcoambrosini: yes we should, separate issue tho

nickvergessen (approved):

PR #7942, opened 2022-09-16
Implement dashboard IAPIWidget

### LINKED ISSUE #7867: Adapt talk mentions widget for mobile clients
### Is your feature request related to a problem? Please describe.

With https://github.com/nextcloud/server/pull/33658 mobile clients can access app widget data.

### Describe the solution you'd like

Adapt the *upcoming cards* widget to https://github.com/nextcloud/server/pull/33658 once it's done.

### Blocked by

* https://github.com/nextcloud/server/pull/33658

nickvergessen:
Fix #7867

- [x] Requires https://github.com/nextcloud/server/pull/33658
- [x] Requires OCP update after tonight

---

- [x] Implement IAPIWidget
- [x] Implement IIconWidget
- [x] Should we also implement IButtonWidget ?


vitormattos (commented):
Isn't good to implement test scenarios to cover this code and check if to specific room list will return the specific items?

nickvergessen:
> Isn't good to implement test scenarios to cover this code

I was thinking about it as well.
My main issue is that integration tests go against API of the dashboard app. Also the result is polluted heavily depending on the enabled apps. But sure let me check

tests/integration/features/integration/dashboard.feature
```diff
@@ -0,0 +1,35 @@
+Feature: integration/dashboard
+  Background:
+    Given user "participant1" exists
+    Given user "participant2" exists
+
+  Scenario: User gets the available dashboard widgets
+    When user "participant1" sees the following entry when loading the list of dashboard widgets (v1)
+      | id     | title         | icon_class          | icon_url         | widget_url                                   | item_icons_round | order | buttons |
+      | spreed | Talk mentions | dashboard-talk-icon | img/app-dark.svg | http://localhost:8080/index.php/apps/spreed/ | true             | 10    | [{"type":"more","text":"More unread mentions","link":"http://localhost:8080/index.php/apps/spreed/"}] |
```
vitormattos, line 9: I think that is good to replace the URL by a placeholder representing the place of the base URL. Today the URL is fixed, but this can change in other environments.
For example, using `<baseUrl>` instead of `http://localhost:8080` and  in the method of this sentence do the magic handling the key "link".

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -1694,6 +1694,79 @@ protected function preparePollExpectedData(array $expected): array {
...
+
+			Assert::assertEquals($widget, $data[$widgetId][$key], 'Wrong details for item #' . $key);
+		}
+	}
+
+	/**
+	 * @Then /^sleep 1 second$/
+	 *
+	 */
+	public function sleepOneSecond(): void {
```
vitormattos, line 1766: We already have a method like this:

`wait for :seconds seconds`

lib/Dashboard/TalkWidget.php
```diff
@@ -84,4 +124,89 @@ public function load(): void {
...
+				$subtitle = str_replace($placeholders, $replacements, $message->getMessage());
+			}
+		}
+
+		return new WidgetItem(
+			$room->getDisplayName($userId),
+			$subtitle,
+			$this->url->linkToRouteAbsolute('spreed.Page.showCall', ['token' => $room->getToken()]),
+			$this->getRoomIconUrl($room, $userId)
+		);
```
tobiasKaminsky, line 178: Can you add here roundIcon: true

https://github.com/nextcloud/server/blob/master/lib/public/Dashboard/Model/WidgetOptions.php#L32
it is in
Widget Options:
https://github.com/nextcloud/server/blob/master/lib/public/Dashboard/IOptionWidget.php#L37

nickvergessen: Items itself don't have the options and the widget itself already has the "roundIcon" option set.
https://github.com/nextcloud/spreed/pull/7942/files#diff-9d0ac71d126947c6f952b00aba3afc1df1e75b99431cb53086b0c43c2f677c29R90

vitormattos (approved):

PR #7954, opened 2022-09-19
Fix chat endpoints with empty reactions in XML format

### LINKED ISSUE #7952: error posting chat message via api - TypeError: XMLWriter::writeElement(): Argument #2 ($content) must be of type ?string, stdClass given at /var/www/html/lib/private/AppFramework/OCS/BaseResponse.php#148
I post messages from a Python script to a chat on my Nextcloud server using [the api](https://nextcloud-talk.readthedocs.io/en/turn_doc/chat/#sending-a-new-chat-message).

I'm seeing a new error every time I post a message. The error shows up in my main Nextcloud server log, and it seems to have started when I upgraded the Talk app to version 14.0.5 on Thu Sep 15.

The error is: `TypeError: XMLWriter::writeElement(): Argument #2 ($content) must be of type ?string, stdClass given at /var/www/html/lib/private/AppFramework/OCS/BaseResponse.php#148`.

Here's the full error message from my `nextcloud.log` file (redacted and formatted for readability):

```json
{
    "reqId": "redacted",
    "level": 3,
    "time": "2022-09-15T20:20:04+00:00",
    "remoteAddr": "redacted",
    "user": "redacted",
    "app": "PHP",
    "method": "POST",
    "url": "/ocs/v2.php/apps/spreed/api/v1/chat/redacted",
    "message": "TypeError: XMLWriter::writeElement(): Argument #2 ($content) must be of type ?string, stdClass given at /var/www/html/lib/private/AppFramework/OCS/BaseResponse.php#148",
    "userAgent": "python-requests/2.20.0",
    "version": "24.0.5.1",
    "data": {
        "app": "PHP"
    }
}
```

The message still shows up in the chat room despite the logged error.

My Nextcloud server is running in a Docker container from a `24.0.5-apache` image.

On the client side, I get an HTTP 200 response with nothing in the body. I'm using code [almost exactly like this](https://github.com/nextcloud/spreed/issues/1879#issuecomment-584407332) to send messages and my code hasn't changed since Dec 2021.

Hopefully this is enough info to point to the culprit. I'm happy to provide more debugging information or a stand-alone repro if necessary.

Note the response status code is 200 but [the docs say it should be 201](https://nextcloud-talk.readthedocs.io/en/turn_doc/chat/#sending-a-new-chat-message).

nickvergessen:
Fix #7952 

Check result of both curl commands before and after:

```sh
# json
curl -k -u admin:admin \
    'https://nextcloud25.local/ocs/v2.php/apps/spreed/api/v1/chat/7d85h4a9?setReadMarker=0&lookIntoFuture=0&lastKnownMessageId=201&includeLastKnown=1' \
    -H 'OCS-APIRequest: true' \
    -H 'Accept: application/json, text/plain, */*' \
    | jq .


# xml
curl -k -u admin:admin \
    'https://nextcloud25.local/ocs/v2.php/apps/spreed/api/v1/chat/7d85h4a9?setReadMarker=0&lookIntoFuture=0&lastKnownMessageId=201&includeLastKnown=1' \
    -H 'OCS-APIRequest: true'
```

(Adjust domain and token and make sure the json one still returns an empty object for reactions

### DIFF
--- a/lib/Controller/AEnvironmentAwareController.php
+++ b/lib/Controller/AEnvironmentAwareController.php
@@ -27,6 +27,7 @@ declare(strict_types=1);
 
 namespace OCA\Talk\Controller;
 
+use OC\AppFramework\Http\Dispatcher;
 use OCA\Talk\Participant;
 use OCA\Talk\Room;
 use OCP\AppFramework\OCSController;
@@ -58,5 +59,26 @@ abstract class AEnvironmentAwareController extends OCSController {
 
 	public function getParticipant(): ?Participant {
 		return $this->participant;
+	}
+
+	/**
+	 * Following the logic of {@see Dispatcher::executeController}
+	 * @return string Either 'json' or 'xml'
+	 */
+	public function getResponseFormat(): string {
+		// get format from the url format or request format parameter
+		$format = $this->request->getParam('format');
+
+		// if none is given try the first Accept header
+		if ($format === null) {
+			$headers = $this->request->getHeader('Accept');
+			/**
+			 * Default value of
+			 * @see OCSController::buildResponse()
+			 */
+			$format = $this->getResponderByHTTPHeader($headers, 'xml');
+		}
+
+		return $format;
 	}
 }
--- a/lib/Controller/ChatController.php
+++ b/lib/Controller/ChatController.php
@@ -168,9 +168,9 @@ class ChatController extends AEnvironmentAwareController {
 
 		$this->participantService->updateLastReadMessage($this->participant, (int) $comment->getId());
 
-		$data = $chatMessage->toArray();
+		$data = $chatMessage->toArray($this->getResponseFormat());
 		if ($parentMessage instanceof Message) {
-			$data['parent'] = $parentMessage->toArray();
+			$data['parent'] = $parentMessage->toArray($this->getResponseFormat());
 		}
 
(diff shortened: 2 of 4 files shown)


nickvergessen:
cc @meonkeys

lib/Controller/AEnvironmentAwareController.php
```diff
@@ -59,4 +60,24 @@ public function setParticipant(Participant $participant): void {
 	public function getParticipant(): ?Participant {
 		return $this->participant;
 	}
+
+	/**
+	 * @return string Either 'json' or 'xml'
+	 */
+	public function getResponseFormat(): string {
+		// get format from the url format or request format parameter
+		$format = $this->request->getParam('format');
```
vitormattos, line 69: If in format have blablabla will pass, not?!
Isn't good to check if format is in a list of allowed formats?

nickvergessen: Server also does not care:
https://github.com/nextcloud/server/blob/368f83095d443c28b0d90d2308549349cc28dd05/lib/private/AppFramework/Http/Dispatcher.php#L231-L244

Will add that link as a comment

vitormattos: In the server maybe make sense because the API of apps could deliver specifics formats different from json and xml, but in this case only is valid json and xml.

vitormattos (approved):

PR #7986, opened 2022-09-22
Allow to count the participants per call

### LINKED ISSUE #7451: Allow to protocol participant counts for individual calls
Currently we run `occ talk:active-calls` to monitor the participant count of our webinars. But when there is more than 1 call it does not work anymore.
So we need a similar command that can be restricted to an individual room.

### LINKED ISSUE #8014: occ talk:active-calls: wrong return code when there is one or more active calls
### How to use GitHub

---

## Steps to reproduce
1. Run `occ talk:active-calls` when at least a call is going on
2. Check the return code (from a shell command line: `echo $?`)

### Expected behaviour
The return code of the command should be 0, if the command is successful.

### Actual behaviour
The return code is 0 only if there are no active calls. Note that a wrong return code may break external programs that run the program in batch mode, for example, for monitoring purposes. In my case it was Python's subprocess.check_output.

## Talk app

**Talk app version:** 14.0.5

## Server configuration

**PHP version:** 8.1

**Nextcloud Version:** 24.0.5

**List of activated apps:**

nickvergessen:
Fix #7451 
Fix #8014

```sh
occ  talk:active-calls -d --output json_pretty
{
    "9537209724": 1
}
```

To get the count of a single conversation:
```sh
occ talk:active-calls -d --output json | jq -r '.["9537209724"]'
1
```

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -83,7 +83,9 @@ And in the works for the [coming versions](https://github.com/nextcloud/spreed/m
 		<command>OCA\Talk\Command\Command\Delete</command>
 		<command>OCA\Talk\Command\Command\ListCommand</command>
 		<command>OCA\Talk\Command\Command\Update</command>
+		<command>OCA\Talk\Command\Monitor\Calls</command>
 		<command>OCA\Talk\Command\Monitor\HasActiveCalls</command>
+		<command>OCA\Talk\Command\Monitor\Room</command>
 		<command>OCA\Talk\Command\Stun\Add</command>
 		<command>OCA\Talk\Command\Stun\Delete</command>
 		<command>OCA\Talk\Command\Stun\ListCommand</command>
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -78,12 +78,12 @@ And in the works for the [coming versions](https://github.com/nextcloud/spreed/m
 	</repair-steps>
 
 	<commands>
-		<command>OCA\Talk\Command\ActiveCalls</command>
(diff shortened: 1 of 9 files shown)


lib/Command/ActiveCalls.php
```diff
@@ -84,6 +106,48 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 			$data = ['calls' => $numCalls, 'participants' => $numParticipants];
 			$this->writeArrayInOutputFormat($input, $output, $data);
 		}
-		return 1;
+		return $input->getOption('non-zero') ? 1 : 0;
+	}
+
+	protected function executeDetails(InputInterface $input, OutputInterface $output): int {
+		$query = $this->connection->getQueryBuilder();
```
vitormattos, line 113: Wouldn't it be better in a Service to make this query usable by other parts of code?

vitormattos (commented):
Maybe will be good to implement integration tests to cover this new command.

lib/Command/ActiveCalls.php
nickvergessen: I had the same thought and then i wondered if I even want this code to be available for reuse as it somewhat undermins the privacy.
For now I would leave it out and only move it once needed

vitormattos: Will be easier to found this query, maybe in the `SessionService`, to reuse in other places.

Also, if we isolate this query, it's makes easier to write unit tests because now we only have integration tests covering Talk commands. Should be good to have more unit tests covering more scenarios than integration tests.

PR #8001, opened 2022-09-23
🏷️ /call/ reference provider

### LINKED ISSUE #7988: Implement link widget thingy for /call/ urls
Following the sample of https://github.com/nextcloud/server/pull/33755

nickvergessen:
Fix #7988

### /call/…

* If the user is a participant
    * Title is the conversation name
    * Description the conversation description
* Open conversation joinable for the user
    * Title is the conversation name
    * Description the conversation description
* Public link (with and without password)
    * Title is the "Private conversation"
    * Description is empty

### /call/…#message_…

To get a closer preview the user needs to be a participant (open conversation or public link are not enough, in those cases the above information is shown)

* Title is "Message of {user} in {conversation}"
* Description is the plain text chat message

Participant | Stranger
---|---
Bildschirmfoto vom 2022-09-23 14-47-20 | Bildschirmfoto vom 2022-09-23 14-47-47

### DIFF
--- a/lib/Collaboration/Reference/TalkReferenceProvider.php
+++ b/lib/Collaboration/Reference/TalkReferenceProvider.php
@@ -255,19 +255,13 @@ class TalkReferenceProvider implements IReferenceProvider {
 
 	protected function getRoomIconUrl(Room $room, string $userId): string {
 		if ($room->getType() === Room::TYPE_ONE_TO_ONE) {
-			$participants = json_decode($room->getName(), true);
-
-			foreach ($participants as $p) {
-				if ($p !== $userId) {
-					return $this->urlGenerator->linkToRouteAbsolute(
-						'core.avatar.getAvatar',
-						[
-							'userId' => $p,
-							'size' => 64,
-						]
-					);
-				}
-			}
+			return $this->urlGenerator->linkToRouteAbsolute(
+				'core.avatar.getAvatar',
+				[
+					'userId' => $room->getSecondParticipant($userId),
+					'size' => 64,
+				]
+			);
 		}
 
 		return $this->urlGenerator->getAbsoluteURL($this->urlGenerator->imagePath('spreed', 'changelog.svg'));
--- a/lib/Collaboration/Reference/TalkReferenceProvider.php
+++ b/lib/Collaboration/Reference/TalkReferenceProvider.php
@@ -31,6 +31,7 @@ use OCA\Talk\Exceptions\ParticipantNotFoundException;
 use OCA\Talk\Exceptions\RoomNotFoundException;
 use OCA\Talk\Manager;
 use OCA\Talk\Model\Attendee;
+use OCA\Talk\Participant;
 use OCA\Talk\Room;
 use OCP\Collaboration\Reference\IReference;
 use OCP\Collaboration\Reference\IReferenceProvider;
@@ -150,6 +151,11 @@ class TalkReferenceProvider implements IReferenceProvider {
 		}
 
 		$room = $this->roomManager->getRoomForUserByToken($referenceMatch['token'], $this->userId);
+		try {
+			$participant = $room->getParticipant($this->userId);
+		} catch (ParticipantNotFoundException $e) {
+			$participant = null;
+		}
 
 		/**
 		 * Default handling:
@@ -158,15 +164,13 @@ class TalkReferenceProvider implements IReferenceProvider {
 		 */
 		$roomName = $room->getDisplayName($this->userId);
 		$title = $roomName;
-		$description = $room->getDescription();
+		$description = '';
 
-		$participant = null;
-		if (!empty($referenceMatch['message'])) {
-			try {
-				$participant = $room->getParticipant($this->userId);
-			} catch (ParticipantNotFoundException $e) {
-			}
+		if ($participant instanceof Participant
+			|| $this->roomManager->isRoomListableByUser($room, $this->userId)) {
+			$description = $room->getDescription();
 		}
+
 
 		/**
 		 * If linking to a comment and the user is already a participant
--- a/lib/Collaboration/Reference/TalkReferenceProvider.php
+++ b/lib/Collaboration/Reference/TalkReferenceProvider.php
@@ -1,6 +1,7 @@
 <?php
 
-declare(strict_types = 1);
+declare(strict_types=1);
+
 /**
  * @copyright Copyright (c) 2022 Joas Schilling <coding@schilljs.com>
  *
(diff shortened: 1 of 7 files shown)


lib/Collaboration/Reference/TalkReferenceProvider.php
```diff
@@ -0,0 +1,289 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+namespace OCA\Talk\Collaboration\Reference;
+
+use OC\Collaboration\Reference\Reference;
```
nickvergessen, line 28: @juliushaertl I saw in some PR that you work on a public class for this?

lib/Collaboration/Reference/TalkReferenceProvider.php
```diff
@@ -0,0 +1,289 @@
...
+		$reference->setDescription($description);
+		$reference->setUrl($this->urlGenerator->linkToRouteAbsolute('spreed.Page.showCall', ['token' => $room->getToken()]));
+		$reference->setImageUrl($this->getRoomIconUrl($room, $this->userId));
+
+		$reference->setRichObject('call', [
+			'id' => $room->getToken(),
+			'name' => $roomName,
+			'link' => $reference->getUrl(),
+			'call-type' => $this->getRoomType($room),
+		]);
```
eneiluj, line 230: This is useless as there is no 'call' reference widget registered in the frontend.
Vue-Richtext falls back to the OpenGraph widget so the richObject is ignored.

nickvergessen: But it also doesn't hurt and allows easier migration in the future 😅
And I'm too lazy to throw away stuff the future self might use

eneiluj: :+1:

eneiluj (approved):

nickvergessen:
- [ ] Wait for decision on OCP class for Reference
    - [ ] Use new OCP class or
    - [ ] Ignore current OC class via psalm.xml

lib/Collaboration/Reference/TalkReferenceProvider.php
juliushaertl: Here we go: https://github.com/nextcloud/server/pull/34249

lib/Collaboration/Reference/TalkReferenceProvider.php
```diff
@@ -0,0 +1,288 @@
...
+	public function matchReference(string $referenceText): bool {
+		return $this->getTalkAppLinkToken($referenceText) !== null;
+	}
+
+	/**
+	 * @param string $referenceText
+	 * @return array|null
+	 * @psalm-return ReferenceMatch|null
+	 */
+	private function getTalkAppLinkToken(string $referenceText): ?array {
```
vitormattos, line 77: I think that is good to put the rule to prevent private usage in the linter, not?!

This is a good method to cover with unit test because is very important to this class.

lib/Collaboration/Reference/TalkReferenceProvider.php
```diff
@@ -0,0 +1,288 @@
...
+		$indexPhpUrl = $this->urlGenerator->getAbsoluteURL('/index.php/call/');
+		$rewriteUrl = $this->urlGenerator->getAbsoluteURL('/call/');
+
+		if (str_starts_with($referenceText, $indexPhpUrl)) {
+			$urlOfInterest = substr($referenceText, strlen($indexPhpUrl)) ?: null;
+		} elseif (str_starts_with($referenceText, $rewriteUrl)) {
+			$urlOfInterest = substr($referenceText, strlen($rewriteUrl)) ?: null;
+		} else {
+			return null;
+		}
```
vitormattos, line 87: Maybe will be good to put this in a separated method to make this method more simple.
Suggestion: `extractUrlOfInterest`

lib/Collaboration/Reference/TalkReferenceProvider.php
```diff
@@ -0,0 +1,288 @@
...
+
+		$token = substr($urlOfInterest, 0, $cutPosition);
+		$messageId = null;
+		if ($hashPosition !== false) {
+			$afterHash = substr($urlOfInterest, $hashPosition + 1);
+			if (preg_match('/^message_(\d+)$/', $afterHash, $matches)) {
+				$messageId = $matches[1];
+			}
+		}
```
vitormattos, line 114: What you think to create a method like `extractTokenFromUrl` to make this method more simple and cover the new method with unit test?

lib/Collaboration/Reference/TalkReferenceProvider.php
```diff
@@ -0,0 +1,288 @@
...
+						]
+					);
+				}
+			}
+		}
+
+		return $this->urlGenerator->getAbsoluteURL($this->urlGenerator->imagePath('spreed', 'changelog.svg'));
+	}
+
+	protected function getRoomType(Room $room): string {
```
vitormattos, line 276: We already have other method very similar. Isn't the moment to move this to a more generic place and cover with unit test?
https://github.com/nextcloud/spreed/blob/758a5a767e61e2504152d86b7d051f89396ff882/lib/Notification/Notifier.php#L548-L559

lib/Collaboration/Reference/TalkReferenceProvider.php
```diff
@@ -0,0 +1,288 @@
...
+	public function getCacheKey(string $referenceId): ?string {
+		$referenceMatch = $this->getTalkAppLinkToken($referenceId);
+		if ($referenceMatch === null) {
+			return '';
+		}
+
+		return ($this->userId ?? '') . '#' . ($referenceMatch['message'] ?? 0);
+	}
+
+	protected function getRoomIconUrl(Room $room, string $userId): string {
```
vitormattos, line 256: This method return two types of icon URL, is a good point to cover with unit test.

lib/Collaboration/Reference/TalkReferenceProvider.php
```diff
@@ -0,0 +1,288 @@
...
+			return $reference;
+		}
+
+		return null;
+	}
+
+	/**
+	 * @throws RoomNotFoundException
+	 */
+	private function fetchReference(Reference $reference): void {
```
vitormattos, line 143: What you think to use a tool to check the cyclomatic complexity in Talk? Like this, we have a lot of other big methods.

marcoambrosini (approved):

lib/Collaboration/Reference/TalkReferenceProvider.php
nickvergessen: Have tests prepared locally, will commit tomorrow

lib/Collaboration/Reference/TalkReferenceProvider.php
nickvergessen: I'm not really a fan of these micro functions where you have to jump around to see what the bigger picture does

lib/Collaboration/Reference/TalkReferenceProvider.php
nickvergessen: The function is now only chained 3rdparty calls, so not sure tests are too useful.

lib/Collaboration/Reference/TalkReferenceProvider.php
nickvergessen: The function is now only chained 3rdparty calls, so not sure tests are too useful.

vitormattos (approved):

PR #8026, opened 2022-09-27
Align search field with rest of the page

CarlSchwan:
And add some right padding

Before:

image

After:

image

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -516,7 +516,7 @@ export default {
 
 .new-conversation {
 	display: flex;
-	padding: 8px 0 8px 12px;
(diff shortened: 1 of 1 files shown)


nimishavijay (approved):
👌🏽

CarlSchwan:
hmm actually this is wrong, the close sidebar button is the one that is not alligned with the rest

nickvergessen:
Seems to be intentional as @marcoambrosini just changed it the other way around 5 days ago https://github.com/nextcloud/spreed/pull/7977

jancborchardt:
Hm, but the change at https://github.com/nextcloud/spreed/pull/7977 seems off, looked better before with the border-radius and position relative to the menu toggle as well. @marcoambrosini?

PR #8031, opened 2022-09-27
Fix missing "Leave call" button for moderators

### LINKED ISSUE #8030: Leave call button missing for moderators in "Can start call"-limited conversations
That being said, also the leave call button seems to break sometimes:
Bildschirmfoto vom 2022-09-27 15-42-00

_Originally posted by @nickvergessen in https://github.com/nextcloud/spreed/issues/8008#issuecomment-1259530142_

nickvergessen:
Fix #8030 

Before | After
---|---
Bildschirmfoto vom 2022-09-27 15-42-00 | Bildschirmfoto vom 2022-09-27 17-00-32

### DIFF
--- a/src/components/TopBar/CallButton.vue
+++ b/src/components/TopBar/CallButton.vue
@@ -49,11 +49,11 @@
 			{{ leaveCallLabel }}
 		</NcButton>
 		<NcActions v-else-if="showLeaveCallButton && canEndForAll"
-			:disabled="loading">
+			:disabled="loading"
+			:menu-title="leaveCallCombinedLabel"
+			type="error">
 			<template #icon>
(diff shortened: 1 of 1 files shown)


src/components/TopBar/CallButton.vue
```diff
@@ -181,6 +179,10 @@ export default {
 			return t('spreed', 'Leave call')
 		},
 
+		leaveCallCombinedLabel() {
+			return this.leaveCallLabel + ' ▼'
```
nickvergessen, line 183: We can not have a second icon on the button, so for now I went with this plain text symbol.

marcoambrosini (approved):

marcoambrosini (approved):

PR #8057, opened 2022-09-30
Don't search in lobbied conversations

### LINKED ISSUE #8054: Messages from a room with a lobby can still be found when searching for them
## Steps to reproduce
1. Create room with user A
2. Add user B to the room
3. Enable lobby for the room using user A
4. Send a message using user A into the room
5. Search for the message text using user B

### Expected behaviour
The message should not show up because it's in a room with a lobby, so the user can't see the messages

### Actual behaviour
The message appears in the search results

## Talk app

**Talk app version:** v15.0.0-rc.2

_Note: This bug was first found on the Android client, but also happens in the web interface._

nickvergessen:
Fix #8054 

cc @jld3103


lib/Search/MessageSearch.php
```diff
@@ -120,6 +121,13 @@ public function search(IUser $user, ISearchQuery $query): SearchResult {
 				continue;
 			}
 
+			if ($room->getLobbyState() !== Webinary::LOBBY_NONE) {
+				$participant = $room->getParticipant($user->getUID(), false);
```
jld3103, line 125: Why is the order of different to the if clause above and split into two parts?

nickvergessen: because in the `Current` it's a one-time thing, so it does not matter as much, while here we would potentially set up your user in hundreds of conversations even if they don't have a lobby in the end.

jld3103 (approved):

vitormattos (approved):
I think that is important to create an issue to implement integration tests covering this scenario

vitormattos (approved):

nickvergessen:
/backport 0710aa01e079ecb07a49b0f712cb3336d3364e69 to stable25

nickvergessen:
/backport 0710aa01e079ecb07a49b0f712cb3336d3364e69 to stable23

Issue #521, opened 2017-12-10
Disable "Public calls"


azuledu:
In our organization we use Nextcloud only in our intranet. 
In this use case, the feature "Public calls" has no sense because the generated URL is not reachable from outside.
It would be good to have a configuration option to remove the "New public call" button for this scenario. 

Thanks.

nickvergessen:
Yeah, makes sense.

Nevertheless I could still see a usecase for "public calls" in your organisation:
You can create a public call and send the link via email to everyone, and they can join the call. You do not have to invite all members to the call individually.

zguithues:
"New Public Call" (only as public as your nextcloud server)

PR #8082, opened 2022-10-02
Added parameter to disallow creating of rooms with public access

pboguslawski:
Added spreed application config parameter `public_rooms_allowed` to disallow
creating of rooms with public (guest via link) access (i.e. for internal
Nextcloud setups). Use

```
occ config:app:set spreed public_rooms_allowed --value 'yes'
```

to allow (default if not set) and

```
occ config:app:set spreed public_rooms_allowed --value 'no'
```

to disallow creating of rooms with public access.

Related: https://github.com/nextcloud/spreed/issues/521
Author-Change-Id: IB#1126215
Signed-off-by: Pawel Boguslawski <pawel.boguslawski@ib.pl>


lib/Manager.php
```diff
@@ -908,6 +910,12 @@ public function getChangelogRoom(string $userId): Room {
 	 * @return Room
 	 */
 	public function createRoom(int $type, string $name = '', string $objectType = '', string $objectId = ''): Room {
+
+		// Force group room if public room was requested and public rooms are disallowed.
+		if (($type === Room::TYPE_PUBLIC) && ($this->config->getAppValue('spreed', 'public_rooms_allowed', 'yes') === 'no')) {
+			$type = Room::TYPE_GROUP;
```
nickvergessen, line 916: Not sure if we should simply convert it. Or just error with a 400 Bad Request and a message the clients can display.

lib/Room.php
```diff
@@ -257,6 +262,11 @@ public function getType(): int {
 	}
 
 	public function setType(int $type): void {
+		// Force group room if public room was requested and public rooms are disallowed.
+		if (($type === self::TYPE_PUBLIC) && ($this->config->getAppValue('spreed', 'public_rooms_allowed', 'yes') === 'no')) {
+			$type = self::TYPE_GROUP;
+		}
```
nickvergessen, line 268: This is not correct. You want to do this change on the RoomService, not the model.
But see comment above.

lib/Settings/Admin/AdminSettings.php
```diff
@@ -146,6 +147,11 @@ protected function initMatterbridge(): void {
 			'matterbridge_enable',
 			$this->serverConfig->getAppValue('spreed', 'enable_matterbridge', '0') === '1'
 		);
+
+		$this->initialState->provideInitialState(
+			'public_rooms_allowed',
+			$this->serverConfig->getAppValue('spreed', 'public_rooms_allowed', 'yes') === 'yes'
```
nickvergessen, line 153: We need to expose this to the mobile clients as a capability so they can also hide the features accordingly.

nickvergessen (changes requested):

marcoambrosini (approved):
tested and works 👌

lib/Manager.php
pboguslawski: Conversion was choosen because spreed code ignores errors returned by modified functions and this seems to be safer. Conversion should never happen because function should be not available in UI also (and if bad guys will try to modify POST, will get group conversation which should not hurt regular users).

lib/Room.php
pboguslawski: Modified places where choosen to avoid problems with lack of return errors handling. If you still recommend moving it elswhere - please suggest detailed change.

lib/Manager.php
nickvergessen: That assumes all clients got updated in the meantime :)

lib/Settings/Admin/AdminSettings.php
pboguslawski: This mod was created in web UI in mind. Not experienced in nextcloud devel nor using mobile app so please suggest required changes if necessary.

lib/Manager.php
pboguslawski: Sorry, we use web UI only - feel free to extend it for other requires areas if necessarry (not experienced enough in nc devel to propose it now).

nickvergessen: You can fix it with one step by moving it to `RoomService::createConversation` and throwing an exception or basically adjust this list here:
https://github.com/nextcloud/spreed/blob/220089657318501ff9b08941349b27a79b0f7a1b/lib/Service/RoomService.php#L141-L147

lib/Room.php
nickvergessen: As per above, you should do it only in the RoomService

lib/Settings/Admin/AdminSettings.php
nickvergessen: Basically add a line like this which in addition also checks your config:
https://github.com/nextcloud/spreed/blob/91d3bd8df8a5288c2abb3badde00c248d974dae8/lib/Capabilities.php#L151

nickvergessen (changes requested):
- [ ] Also RoomService::setType() should check the config.
- [ ] RoomController::addParticipantToRoom should not allow to add email addresses to a conversation (which also makes the room public)
- [ ] Since emails are then not allowed anymore, we should also not autocomplete them here:
https://github.com/nextcloud/spreed/blob/d01450f87a68454a7acf6fa8e51ad8f0bec40ce6/src/services/conversationsService.js#L83

lib/Manager.php
pboguslawski: `type` is property of `Room` not `RoomService`  so verification should take place in `Room` class to avoid future problems (i.e. code outside of `RoomService` will create/change room of disabled type). Seems that now `Room::setType` and `Room` constructor are the best places for this verification but code using it should properly handle errors when disabled type is passed. Please let me know if its possible to just throw exception in `Room` methods mentioned above without messing with room type conversions (there should not be any other trash left after such exception - not sure if it would be safe now).

lib/Settings/Admin/AdminSettings.php
pboguslawski: Not experienced with capabilities in Nextcloud - please create change suggestion and will just accept it to this PR or make separate PR to adjust it.

pboguslawski:
> * [ ]  Also RoomService::setType() should check the config.
> 
>     * [ ]  RoomController::addParticipantToRoom should not allow to add email addresses to a conversation (which also makes the room public)
> 
>     * [ ]  Since emails are then not allowed anymore, we should also not autocomplete them here:
>       https://github.com/nextcloud/spreed/blob/d01450f87a68454a7acf6fa8e51ad8f0bec40ce6/src/services/conversationsService.js#L83

Verification should be placed in Room class in all methods capable of setting/changing `type` (now constructor and setType probably) but code using Room must be prepared for errors/exceptions (not sure if already is) so please propose better solution than conversion proposed in this change (as suggestions or separate PR).

pboguslawski:
a38fdfd0c2faada01a90b4dbdc31838bb2f8d694 disables adding e-mails to room when public rooms are disabled.

nickvergessen:
Sorry for the lack of responsiveness lately.
I still have your PR on my todo list and hopefully will reach it this week.

nickvergessen:
Hi there, I copied your branch into our organisation and sent you an invite.
This allows us to better collaborate on the feature.

I rebased it on latest master and did some fixes and changes.
New pull request is here: https://github.com/nextcloud/spreed/pull/8357

PR #8084, opened 2022-10-03
Unify duplicated code

vitormattos:


lib/Share/RoomShareProvider.php
```diff
@@ -403,36 +417,12 @@ public function deleteFromSelf(IShare $share, $recipient): void {
...
-					'share_type' => $qb->createNamedParameter(self::SHARE_TYPE_USERROOM),
-					'share_with' => $qb->createNamedParameter($recipient),
-					'uid_owner' => $qb->createNamedParameter($share->getShareOwner()),
-					'uid_initiator' => $qb->createNamedParameter($share->getSharedBy()),
-					'parent' => $qb->createNamedParameter($share->getId()),
-					'item_type' => $qb->createNamedParameter($share->getNodeType()),
-					'item_source' => $qb->createNamedParameter($share->getNodeId()),
-					'file_source' => $qb->createNamedParameter($share->getNodeId()),
-					'file_target' => $qb->createNamedParameter($share->getTarget()),
-					'permissions' => $qb->createNamedParameter(0),
```
nickvergessen, line 425: Here the permission was hardcoded to 0 and now it is not anymore. The code was not 100% the same and the details matter here

nickvergessen:
Any progress, or shall we close this?
Seems to risky at the moment

PR #8087, opened 2022-10-03
Move setActiveSince to RoomService

vitormattos:
Ref: #6235

### DIFF
--- a/lib/Manager.php
+++ b/lib/Manager.php
@@ -175,6 +175,7 @@ class Manager {
 			(string) $row['password'],
 			(string) $row['remote_server'],
 			(string) $row['remote_token'],
+			(int) $row['active_guests'],
 			(int) $row['default_permissions'],
 			(int) $row['call_permissions'],
 			(int) $row['call_flag'],
--- a/lib/Manager.php
+++ b/lib/Manager.php
@@ -175,7 +175,6 @@ class Manager {
 			(string) $row['password'],
 			(string) $row['remote_server'],
 			(string) $row['remote_token'],
-			(int) $row['active_guests'],
 			(int) $row['default_permissions'],
 			(int) $row['call_permissions'],
 			(int) $row['call_flag'],
--- a/lib/Room.php
+++ b/lib/Room.php
@@ -170,6 +170,7 @@ class Room {
 	private string $password;
 	private string $remoteServer;
 	private string $remoteToken;
+	private int $activeGuests;
 	private int $defaultPermissions;
 	private int $callPermissions;
 	private int $callFlag;
@@ -202,6 +203,7 @@ class Room {
 								string $password,
 								string $remoteServer,
 								string $remoteToken,
+								int $activeGuests,
 								int $defaultPermissions,
 								int $callPermissions,
 								int $callFlag,
@@ -231,6 +233,7 @@ class Room {
 		$this->password = $password;
 		$this->remoteServer = $remoteServer;
 		$this->remoteToken = $remoteToken;
+		$this->activeGuests = $activeGuests;
 		$this->defaultPermissions = $defaultPermissions;
 		$this->callPermissions = $callPermissions;
 		$this->callFlag = $callFlag;
@@ -391,7 +394,16 @@ class Room {
 		$this->description = $description;
 	}
 
+	/**
+	 * @deprecated Use ParticipantService::getGuestCount() instead
+	 * @return int
+	 */
+	public function getActiveGuests(): int {
+		return $this->activeGuests;
+	}
+
 	public function resetActiveSince(): void {
+		$this->activeGuests = 0;
 		$this->activeSince = null;
 	}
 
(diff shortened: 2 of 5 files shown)


lib/Manager.php
```diff
@@ -175,7 +175,6 @@ public function createRoomObject(array $row): Room {
 			(string) $row['password'],
 			(string) $row['remote_server'],
 			(string) $row['remote_token'],
-			(int) $row['active_guests'],
```
nickvergessen, line 178: This is a value on the model, please keep it injected and also keep the getter + setter to modify the value on the model (only the DB update part is basically moved to the RoomService

lib/Room.php
```diff
@@ -203,7 +202,6 @@ public function __construct(Manager $manager,
 								string $password,
 								string $remoteServer,
 								string $remoteToken,
-								int $activeGuests,
```
nickvergessen, line 206: This is a value on the model, please keep it injected and also keep the getter + setter to modify the value on the model (only the DB update part is basically moved to the RoomService

lib/Room.php
```diff
@@ -394,16 +391,7 @@ public function setDescription(string $description): void {
 		$this->description = $description;
 	}
 
-	/**
-	 * @deprecated Use ParticipantService::getGuestCount() instead
-	 * @return int
-	 */
-	public function getActiveGuests(): int {
```
nickvergessen, line 401: This is a value on the model, please keep it injected and also keep the getter + setter to modify the value on the model (only the DB update part is basically moved to the RoomService

lib/Room.php
```diff
@@ -726,48 +714,10 @@ public function setLastActivity(\DateTime $now): bool {
...
-		}
-
-		$update = $this->db->getQueryBuilder();
-		$update->update('talk_rooms')
-			->set('active_since', $update->createNamedParameter($since, IQueryBuilder::PARAM_DATE))
-			->where($update->expr()->eq('id', $update->createNamedParameter($this->getId(), IQueryBuilder::PARAM_INT)))
-			->andWhere($update->expr()->isNull('active_since'));
-		$update->executeStatement();
-
+	public function setActiveSince(\DateTime $since): void {
```
nickvergessen, line 719: should handle the callFlag and guestFlag in here

nickvergessen (changes requested):

lib/Service/RoomService.php
```diff
@@ -569,6 +569,50 @@ public function resetActiveSince(Room $room): bool {
...
+		}
+
+		$update = $this->db->getQueryBuilder();
+		$update->update('talk_rooms')
+			->set('active_since', $update->createNamedParameter($since, IQueryBuilder::PARAM_DATE))
+			->where($update->expr()->eq('id', $update->createNamedParameter($room->getId(), IQueryBuilder::PARAM_INT)))
+			->andWhere($update->expr()->isNull('active_since'));
+		$update->executeStatement();
+
+		$room->setActiveSince($since, $callFlag, $isGuest);
```
nickvergessen, line 611: This is not called, if `if ($room->getActiveSince() instanceof \DateTime) {` exited above.

In that case you need to update the callFlag + guest count but not the activeSince (so maybe it makes more sense to have a method per value)

lib/Room.php
```diff
@@ -726,48 +726,14 @@ public function setLastActivity(\DateTime $now): bool {
...
-				->set(
-					'call_flag',
-					$update->expr()->bitwiseOr('call_flag', $callFlag)
-				)
-				->where($update->expr()->eq('id', $update->createNamedParameter($this->getId(), IQueryBuilder::PARAM_INT)));
-			$update->executeStatement();
-
+	public function setActiveSince(\DateTime $since, int $callFlag, bool $isGuest): void {
+		$this->activeSince = $since;
+		$this->callFlag = $callFlag;
```
nickvergessen, line 733: The database does:
```php
$update->expr()->bitwiseOr('call_flag', $callFlag)
```

So you also need to bitwise combine it here:

```suggestion
		$this->callFlag = $this->callFlag | $callFlag;
```

lib/Room.php
```diff
@@ -726,48 +726,14 @@ public function setLastActivity(\DateTime $now): bool {
...
-				->set('active_guests', $update->createFunction($update->getColumnName('active_guests') . ' + 1'))
-				->set(
-					'call_flag',
-					$update->expr()->bitwiseOr('call_flag', $callFlag)
-				)
-				->where($update->expr()->eq('id', $update->createNamedParameter($this->getId(), IQueryBuilder::PARAM_INT)));
-			$update->executeStatement();
-
+	public function setActiveSince(\DateTime $since, int $callFlag, bool $isGuest): void {
+		$this->activeSince = $since;
```
nickvergessen, line 732:
```suggestion
		if (!$this->activeSince) {
			$this->activeSince = $since;
		}
```

nickvergessen (changes requested):

lib/Room.php
```diff
@@ -726,48 +726,16 @@ public function setLastActivity(\DateTime $now): bool {
 
 	/**
 	 * @param \DateTime $since
```
nickvergessen, line 728: Remove the complete doc as it's typed?

nickvergessen (approved):

PR #8151, opened 2022-10-12
Adjust filtering and sorting to the same way the dashboard frontend does

nickvergessen:
There is still a minor bug now on the web.
1-1 chats with unread messages are not sorted together with group chats where you have been pinged.
But that is something for a follow up and the dashboard api for the clients is the expected way.


src/views/Dashboard.vue
```diff
@@ -190,6 +190,7 @@ export default {
 				})
 
 				if (importantRooms.length) {
+					// FIXME unread 1-1 conversations are not sorted like unread mentions in group chats
```
vitormattos, line 193:
```suggestion
					/** @todo FIXME unread 1-1 conversations are not sorted like unread mentions in group chats **/
```

What you think? Is best use the annotation and in a future we can search only by @toto to check all that we neet to pay attention and fix.

vitormattos (approved):

PR #8195, opened 2022-10-17
Fix screenshot mode for grid view

nickvergessen:


danxuliu (approved):
Tested and works :+1:

PR #8196, opened 2022-10-17
Fix broken size of raise hand icon

nickvergessen:
Found while fixing #8195


danxuliu (approved):
Tested and works :+1:

PR #8209, opened 2022-10-20
Reduce two menus to one

### LINKED ISSUE #6889: When in a call, reduce action menus from 2 to 1

marcoambrosini:
fix #6889

Signed-off-by: Marco <marcoambrosini@icloud.com>

### DIFF
--- a/src/components/TopBar/TopBarMenu.vue
+++ b/src/components/TopBar/TopBarMenu.vue
@@ -20,54 +20,105 @@
 -->
 
 <template>
-	<NcActions v-if="showActions && isInCall"
-		v-tooltip="t('spreed', 'More actions')"
+	<NcActions v-if="!isSidebar"
+		v-shortkey.once="disableKeyboardShortcuts ? null : ['f']"
+		v-tooltip="t('spreed', 'Conversation actions')"
+		class="top-bar__button"
+		:aria-label="t('spreed', 'Conversation actions')"
 		:container="container"
-		:aria-label="t('spreed', 'More actions')">
-		<template #icon>
+		@shortkey.native="toggleFullscreen">
+		<!-- White icon if in call -->
+		<template v-if="isInCall" #icon>
 			<DotsHorizontal :size="20"
 				fill-color="#ffffff" />
 		</template>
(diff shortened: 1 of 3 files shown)


marcoambrosini:
Current status: 
Screenshot 2022-10-25 at 14 26 12

nickvergessen:
When the right sidebar is open the seperator needs to go:
Bildschirmfoto vom 2022-10-26 14-28-53


While opening the menu the icon is invisible (but this is unchanged from #8019 )

grafik

nickvergessen (changes requested):

src/components/TopBar/TopBarMenu.vue
```diff
@@ -0,0 +1,396 @@
...
+	<NcActions v-if="!isSidebar"
+		v-shortkey.once="disableKeyboardShortcuts ? null : ['f']"
+		v-tooltip="t('spreed', 'Conversation actions')"
+		class="top-bar__button"
+		:aria-label="t('spreed', 'Conversation actions')"
+		:container="container"
+		@shortkey.native="toggleFullscreen">
+		<!-- White icon if in call -->
+		<template v-if="isInCall" #icon>
+			<DotsHorizontal :size="20"
```
nickvergessen, line 32: Should we should the settings wheel as before in case it's not a call?

15.0.0 | This PR
---|---
Bildschirmfoto vom 2022-10-31 10-05-12 | Bildschirmfoto vom 2022-10-31 10-05-05

marcoambrosini: I don't think so, as it contains (and will likely contain more) actions that are not settings

nickvergessen (approved):

szaimen:
🎉🎉🎉

PR #8220, opened 2022-10-22
Query in sqlite returning wrong result.

### LINKED ISSUE #8186: Integration features/command/monitor-calls.feature:26 test fails on SQLite
### How to use GitHub

---

## Steps to reproduce
1. https://drone.nextcloud.com/nextcloud/spreed/10239/4/3

Failed asserting that 'There are currently 1 calls in progress with 1 participants\n
… ' contains "There are currently 2 calls in progress with 3 participants".

vitormattos:
The function `in` already put the parenthesis, duplicated parenthesis return wrong result with sqlite.

close #8186


nickvergessen (approved):

nickvergessen:
Now I wonder in how many places we do this wrong

nickvergessen:
And seems there are at least 2 others in server (middle one is not an IN, so it might still work?):
image

So we should try to find a programmatic way to prevent this.
Can you raise a tech debt ticket in the server about it?

PR #8262, opened 2022-10-31
Add conversation delete support via OCP API

ChristophWurst:
For https://github.com/nextcloud/server/pull/34894. Use case and description is also found in the other PR.

### DIFF
--- a/lib/OCP/TalkBackend.php
+++ b/lib/OCP/TalkBackend.php
@@ -7,6 +7,7 @@ use OCA\Talk\Model\Attendee;
 use OCA\Talk\Participant;
 use OCA\Talk\Room;
 use OCA\Talk\Service\ParticipantService;
+use OCA\Talk\Service\RoomService;
 use OCP\IURLGenerator;
 use OCP\Talk\IConversation;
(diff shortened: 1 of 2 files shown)


lib/OCP/TalkBackend.php
```diff
@@ -15,13 +16,16 @@
 class TalkBackend implements ITalkBackend {
 	protected Manager $manager;
 	protected ParticipantService $participantService;
+	private RoomService $roomService;
```
nickvergessen, line 19:
```suggestion
	protected RoomService $roomService;
```

nickvergessen (approved):

lib/OCP/TalkBackend.php
nickvergessen: As a app developer since ages and phpbb extension developer before that I really dislike the use of private. It really often killed possible extensions and had to wait for an upstream fix (which you can barely version depend on) instead of just extending an existing implementation.

ChristophWurst:
https://github.com/nextcloud/server/pull/34894 is in

PR #8270, opened 2022-11-03
Show voters count on voting screen for moderators

### LINKED ISSUE #8267: Number of votes not visible before voting myself
### How to use GitHub

---

## Steps to reproduce
1. User A starts a poll (public, multiple answers)
2. User B votes

### Expected behaviour
User A (or a moderator) should be able to see how many votes are there already.

### Actual behaviour

Talk-Web:

Talk-iOS:
image

Talk-Web after voting myself (User A):
image

## Talk app

**Talk app version:** Sermo

nickvergessen:
Fix #8267 

Currently using this string because it's translated already and we want to release today.

Will make a follow up to be backported after the release to remove the "Poll results" bit as its the voting screen not the result screen?

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/Poll.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/Poll.vue
@@ -62,6 +62,11 @@
 							{{ pollName }}
 						</h2>
 					</div>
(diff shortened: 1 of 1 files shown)


Ivansss (approved):
👍

PR #8271, opened 2022-11-03
Improve wording of votes count on voting screen

nickvergessen:
Followup #8270

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/Poll.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/Poll.vue
@@ -62,6 +62,11 @@
 							{{ pollName }}
 						</h2>
 					</div>
+					<div class="poll__summary">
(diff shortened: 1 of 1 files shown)


src/components/MessagesList/MessagesGroup/Message/MessagePart/Poll.vue
```diff
@@ -62,6 +62,11 @@
 							{{ pollName }}
 						</h2>
 					</div>
+					<div class="poll__summary">
+						<template v-if="currentUserIsPollCreator || currentUserIsModerator || pollIsPublic">
+							{{ n('spreed', 'Poll intermediate results • %n vote', 'Poll intermediate results • %n votes', votersNumber) }}
```
marcoambrosini, line 67: I would remove the "intermediate", if the poll is open we just simply don't append "ended"

nickvergessen: Okay, then https://github.com/nextcloud/spreed/pull/8270 is enough.

PR #8291, opened 2022-11-04
Fix chat "not loading" and performance issue with many reactions or v…

### LINKED ISSUE #8289: Many reactions in a row can break the chat
### How to use GitHub

---

## Steps to reproduce
Sample conversation was created on sermo

1. Create a set of message `/spam 50 messages`
2. Write a test message (`This one has the reactions`)
3. Open the reaction emoji picker on the previous message and run this code in your JS console:
```js
[...document.querySelectorAll('.emoji-mart-emoji')].forEach(input => input.click())
```
4. Waiting some time for the 1.868 requests to finish
5. Write another chat message (`Empty chat?`)
6. Reload the web UI

### Expected behaviour
Chat should be readable

### Actual behaviour
- [ ] Web UI chat does not load further and browser becomes unresponsive for the author
- [ ] Android chat does not load further
- [ ] iOS app freezes

nickvergessen:
…otes

Fix #8289

### DIFF
--- a/src/store/messagesStore.js
+++ b/src/store/messagesStore.js
@@ -37,6 +37,7 @@ import CancelableRequest from '../utils/cancelableRequest.js'
 import { showError } from '@nextcloud/dialogs'
 import {
 	ATTENDEE,
+	CHAT,
 } from '../constants.js'
 
 /**
@@ -732,9 +733,12 @@ const actions = {
 	 * @param {string} data.token the conversation token;
 	 * @param {object} data.requestOptions request options;
 	 * @param {string} data.lastKnownMessageId last known message id;
+	 * @param {number} data.minimumVisible Minimum number of chat messages we want to load
 	 * @param {boolean} data.includeLastKnown whether to include the last known message in the response;
 	 */
-	async fetchMessages(context, { token, lastKnownMessageId, includeLastKnown, requestOptions }) {
+	async fetchMessages(context, { token, lastKnownMessageId, includeLastKnown, requestOptions, minimumVisible }) {
+		minimumVisible = minimumVisible || CHAT.MINIMUM_VISIBLE
+
 		context.dispatch('cancelFetchMessages')
 
 		// Get a new cancelable request function and cancel function pair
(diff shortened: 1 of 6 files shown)


src/store/messagesStore.js
```diff
@@ -796,6 +798,15 @@ const actions = {
 
 		context.commit('loadedMessagesOfConversation', { token })
 
+		if (invisibleMessages > 95) {
+			// If there are more than 95 invisible messages we load another chunk
```
ShGKme, line 802: `95` seems to be a magic number for me here. If I understand correctly, that is because default limit of chunk is 100 messages and the problem occurs when a whole chunk consists only of "invisible messages". Currently web client never sets the limit and always uses default, so limit does not occur on client side.

What about adding a constant like `DEFAULT_FETCH_MESSAGES_LIMIT`?

nickvergessen: > and the problem occurs when a whole chunk consists only of "invisible messages"

Not only, but yeah then you get the "numb" feeling. But also if it loads for a long time and then only one message shows and you have to scroll down and up again it is weird.

ShGKme: Oh, I see now. So then I'd prefer to add 2 constants for this place to be clean.
Something like `DEFAULT_FETCH_MESSAGE_LIMIT = 100` and `MIN_PREFERED_FETCH_VISIBLE_MESSAGES = 5`.
Or at least a comment with description / link to issue, so it will be no magic in 95...

> if it loads for a long time and then only one message shows and you have to scroll down and up again it is weird

Maybe then require 20-25 messages instead of 5? It is approximate amount of messages takes up the whole 4K screen forcing user to scroll up for previous messages.

ShGKme: There is also a hard but strict way - fetch messages and count visible messages continuously until  there will be 100 visible messages or messages ends. (Or even do this part on server side). But not sure it's worth it =D

nickvergessen: Made a change that should improve the situation

marcoambrosini (approved):
looks good!

ShGKme (approved):
Thanks for adding constants :)

PR #8292, opened 2022-11-04
REQ Stable24: hide details, keep messages

punkyard:
could you add an option to remove/hide manually details one by one
or hide them all with a one-click button?

cause details are really polluting the chatt
thanks

talk activities in chat


nickvergessen:
Please create issues for your requests. You are not trying to make a code change/pull request, but a feature/bug/enhancement suggestion which belongs into the issues.

And your request already has one, so you don't even need to do that.

PR #8309, opened 2022-11-07
Bump min PHP version in CI to 8.0

nickvergessen:
- [x] Ref https://github.com/nextcloud/server/issues/35025

Until 7.4 is dropped from server we just upgrade general CI to 8.0 already:
- [x] https://github.com/nextcloud/spreed/pull/8313


composer.json
```diff
@@ -17,7 +17,7 @@
 		"classmap-authoritative": true,
 		"optimize-autoloader": true,
 		"platform": {
-			"php": "7.4"
+			"php": "8.0"
 		},
```
vitormattos, line 21: When the server bumped the minimum version to >=8, its make necessary to lock the minimum version in all `composer.json` too?

vitormattos (approved):

vitormattos (approved):

composer.json
vitormattos: I verified the status of this bump in server and don't finished. I think that is good to create an issue to remove the platform section in Talk after the PHP version bump on server will be finished.

nickvergessen: We can never remove the platform. Each composer file needs to have the platform specified.
But yeah, server is still in the process, as long as 7.4 is not removed, we will not bump here.

PR #8333, opened 2022-11-10
🖼️ Conversation avatars API

### LINKED ISSUE #927: 🖼️ Conversation avatars

vitormattos:
Fix #927

### 🚧 TODO

- [x] We need OCS API, suggestion: `/api/{apiVersion}/room/{token}/image` - Similar to https://github.com/nextcloud/spreed/blob/master/lib/Controller/TempAvatarController.php
  - [x] POST a new image
    - [x] Needs to validate the file is an image
    - [x] Needs to validate the image is a square and of a supported type (check User avatar code for what we should allow)
    - [x] File should be stored with `OCP\Files\IAppData` with conversation token + file type as name
  - [x] DELETE to remove an image
  - [x] GET which serves the image as FileDisplayResponse
    - [x] Not sure if that works as OCSController, if it does not we need another Controller with another route.
    - [x] If AppData does not have a file with the conversation name, we fall back to returning an icon based on the conversation type
    - [x] If the conversation is a one-to-one conversation we fall back to the user avatar?
    - [x] `/dark` URL for dark avatars
- [x] Add capability
- [x] Add integration tests
- [ ] Adjust "icon url" usage in:
  - [x] activities
  - [x] notifications
  - [x] search results
  - [x] references
  - [ ] etc.

### 🏁 Checklist


lib/Controller/GroupAvatarController.php
```diff
@@ -0,0 +1,176 @@
...
+			);
+		}
+
+		try {
+			$image = new \OC_Image();
+			$image->loadFromData($content);
+			$image->readExif($content);
+			$image->fixOrientation();
+			if (!($image->height() === $image->width())) {
+				throw new NotSquareException($this->l->t('Avatar image is not square'));
```
nickvergessen, line 96: Please throw a normal `InvalidArgumentException` or something instead of using `OC` private classes

nickvergessen (changes requested):

nickvergessen:
Added some todos in the PR description

SystemKeeper:
If this in renamed to a more general "AvatarController", would that supersede the "TempAvatarController" (`temp-user-avatar-api` capability) ?

nickvergessen:
> would that supersede the "TempAvatarController" (temp-user-avatar-api capability) ?

No, as that is still an API that should not be in talk but in server

appinfo/routes/routesGroupAvatarController.php
```diff
@@ -0,0 +1,40 @@
...
+];
+
+return [
+	'ocs' => [
+		/** @see \OCA\Talk\Controller\GroupAvatarController::postAvatar() */
+		['name' => 'GroupAvatar#postAvatar', 'url' => '/api/{apiVersion}/group-avatar/{token}', 'verb' => 'POST', 'requirements' => $requirements],
+		/** @see \OCA\Talk\Controller\GroupAvatarController::getAvatar() */
+		['name' => 'GroupAvatar#getAvatar', 'url' => '/api/{apiVersion}/group-avatar/{token}', 'verb' => 'GET', 'requirements' => $requirements],
+		/** @see \OCA\Talk\Controller\GroupAvatarController::deleteAvatar() */
+		['name' => 'GroupAvatar#deleteAvatar', 'url' => '/api/{apiVersion}/group-avatar/{token}', 'verb' => 'DELETE', 'requirements' => $requirements],
```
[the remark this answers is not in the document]
nickvergessen: Talk does not have user avatars, so this is fine.

lib/Controller/GroupAvatarController.php
```diff
@@ -0,0 +1,176 @@
...
+		} elseif ($folder->fileExists($token . '.jpeg')) {
+			$file = $folder->getFile($token . '.jpeg');
+		} else {
+			// @todo Implement fallback images
+			return new DataResponse([], Http::STATUS_NOT_FOUND);
+		}
+
+		$resp = new FileDisplayResponse($file);
+		$resp->addHeader('Content-Type', $file->getMimeType());
+		return $resp;
```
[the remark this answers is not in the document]
vitormattos: Isn't one day a long time?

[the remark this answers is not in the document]
nickvergessen: User avatars are also cached 1 day. How often would you change a conversation avatar? Most of the chats I have basically set the image once and then never change it.

docs/conversation.md
```diff
@@ -378,3 +378,50 @@
...
+        + `404 Not Found` When the conversation could not be found for the participant
+
+* Required capability: `avatar`
+* Method: `GET`
+* Endpoint: `/room/{token}/avatar/dark`
+
+* Response:
+    - Status code:
+        + `200 OK`
+        + `400 Bad Request` When the conversation haven't an avatar
```
nickvergessen, line 426: dark and normal should be the same. also we don't want a 400 here. only 404 when the user is not a participant

lib/Controller/AvatarController.php
```diff
@@ -0,0 +1,129 @@
...
+				Filesystem::isFileBlacklisted($file['tmp_name'])
+			) {
+				throw new InvalidArgumentException($this->l->t('Invalid file provided'));
+			}
+			if ($file['size'] > 20 * 1024 * 1024) {
+				throw new InvalidArgumentException($this->l->t('File is too big'));
+			}
+
+			$content = file_get_contents($file['tmp_name']);
+			unlink($file['tmp_name']);
```
nickvergessen, line 82: should move all this to the service?
I kind of dislike reading the image content and passing it to a method

lib/Service/AvatarService.php
```diff
@@ -0,0 +1,130 @@
...
+			$file = $folder->getFile($token . '.jpeg');
+		} else {
+			if ($room->getType() === Room::TYPE_ONE_TO_ONE) {
+				$users = json_decode($room->getName(), true);
+				foreach ($users as $participantId) {
+					if ($participantId !== $user->getUID()) {
+						$avatar = $this->avatarManager->getAvatar($participantId);
+						$file = $avatar->getFile(512);
+					}
+				}
```
nickvergessen, line 105: This should be first thing in the method? There is no use to do any of the above when it's the avatar of the other user anyway?

lib/Service/AvatarService.php
```diff
@@ -0,0 +1,130 @@
...
+			} elseif ($room->getObjectType() === 'file') {
+				$file = new InMemoryFile($token, file_get_contents(__DIR__ . '/../../img/icon-file-white.svg'));
```
nickvergessen, line 107: Use `$dark` ?

vitormattos: I think that we don't will use the dark variable. I only maintained to don't mix the context of commits. Probability I will remove the endpoint /dark.

nickvergessen: But we need the /dark endpoint?!

lib/Service/AvatarService.php
vitormattos: This is the fallback, only is necessary if the room haven't an avatar.

This part I copied from here:
https://github.com/nextcloud/spreed/blob/master/lib/Search/ConversationSearch.php#L118-L138
And probability I will unify those codes after do more small adjusts in the code to cover the pending requirements.

docs/conversation.md
```diff
@@ -378,3 +378,50 @@
...
+* Data:
+
+| field  | type   | Description                                                                                                                         |
+| ------ | ------ | ----------------------------------------------------------------------------------------------------------------------------------- |
+| `file` | string | Blob of image in a multipart/form-data request. Only accept images with mimetype equal to PNG or JPEG and need to be squared image. |
+
+* Response:
+    - Status code:
+        + `200 OK`
+        + `400 Bad Request` When: no image, file is too big, invalid mimetype or resource, isn't square, unknown error
```
nickvergessen, line 396: Also 400 when it's a 1-1 conversation. We shall always show the other users avatar in that case?

lib/Service/AvatarService.php
nickvergessen: Unification of the code is not needed.
As per todo list you can remove all those iconClass checking there and just always return the icon URL with `.../avatar/token`

lib/Service/AvatarService.php
vitormattos: For now, not. And I think that we don't will need. I will check more and maybe remove in the next push.

nickvergessen: We need this, as we need to Mimik the user avatars for one-to-one conversations and they also have /dark when the client is using dark mode?!

lib/Controller/GroupAvatarController.php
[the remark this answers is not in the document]
vitormattos: I think that we will need to store the latest version of avatar in the room table.

When the participant A as a room moderator change the avatar
And the participant B retrieve the Avatar
And the participant A change the avatar a more time
The participant B need to receive the newest version of avatar.

If we set the cache without a query string defining the version of avatar, the participant B don't will see the newest avatar of room by a day.

@nickvergessen what you think to put the avatar version in room table? I suggest this table because avatar is an attribute of room.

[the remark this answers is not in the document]
nickvergessen: But it's the same for user avatars.
A day is totally fine. We can cachr it shorter if it is the fall back icon avatar, but i would not do effort to store a version of the avatar.
This is a very rare scenario and it is the same on basically every service.

lib/Controller/AvatarController.php
vitormattos: The first part of this code is too much specific of handling file from global var `$_FILE` and I think that is best to maintain in the controller.

A good point is about send the content of image or the path of image to `setImage` method

I think that the current way is the best way sending the content of image to method because we need do delete the uploaded file. The method `service->setImage` throw exceptions and this can broke the flow and don't delete the file.

Do you have any suggestion?

lib/Service/AvatarService.php
vitormattos: I tried to remove the classes but I think won't be possible because the fallback images are in svg format and is used CSS to make with 22px in the center of circle with a border. This is a point but have other.

The search component use an `img` tag to display the avatar (maybe will be necessary replace by `NcAvatar`) and I don't know why but SVG files give an error when loading.

lib/Service/AvatarService.php
vitormattos: I will check this case.

docs/conversation.md
vitormattos: Done

lib/Controller/AvatarController.php
nickvergessen: You could either create the OC_Image here and forward that on the message call, or you create a `setAvatarFromRequest` which then does all the meta handling and then internally calls setAvatar with "okay" parameters?

lib/Search/ConversationSearch.php
```diff
@@ -113,20 +117,18 @@ public function search(IUser $user, ISearchQuery $query): SearchResult {
 				}
 			}
 
-			$icon = '';
+			if ($this->avatarService->roomHasAvatar($room)) {
+				$icon = $this->url->linkToRouteAbsolute('ocs.spreed.Avatar.getAvatar', [
+					'token' => $room->getToken(),
+					'apiVersion' => 'v1',
+					'v' => $this->avatarService->getAvatarVersion($room, $user->getUID()),
```
nickvergessen, line 124: We said we don't need an avatar version

lib/Service/AvatarService.php
nickvergessen: Then keep the SVG stuff here, but the one-to-one part can be moved to the beginning of the method. No need to check the filesystem with it.

lib/Service/AvatarService.php
```diff
@@ -0,0 +1,174 @@
...
+		if ($room->getType() === Room::TYPE_ONE_TO_ONE && $userId) {
+			return (string) $this->config->getUserValue($userId, 'avatar', 'version', '0');
+		}
+		return (string) ($this->cache->get($room->getToken() . '.avatarVersion') ?? 0);
+	}
+
+	public function getAvatarUrl(Room $room, ?string $userId = null): string {
+		if (!$this->roomHasAvatar($room)) {
+			return '';
+		}
```
nickvergessen, line 167: We want to have a fallback of some sorts. If this does not always return a working image, clients need to have multiple logic paths again to decide which URL to load, etc.

src/components/ConversationIcon.vue
```diff
@@ -31,6 +31,7 @@
 			:disable-menu="disableMenu"
 			:display-name="item.displayName"
 			:preloaded-user-status="preloadedUserStatus"
+			:url="item.avatar"
```
nickvergessen, line 34: I would recommend to just skip all the frontend work for now and focus on the backend path.
We can create a ticket and get a frontender to work on it afterwards.

tests/integration/features/chat/avatar.feature
```diff
@@ -0,0 +1,47 @@
+Feature: chat/avatar
```
nickvergessen, line 1: Should be in conversation

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -2812,6 +2817,40 @@ public function userLogsIn(string $user) {
 		$this->assertStatusCode($this->response, 200);
 	}
 
+	/**
+	 * @When /^user "([^"]*)" send the file "([^"]*)" as avatar of room "([^"]*)" with (\d+)(?: \((v1)\))?$/
```
nickvergessen, line 2821:
```suggestion
	 * @When /^user "([^"]*)" uploads file "([^"]*)" as avatar of room "([^"]*)" with (\d+)(?: \((v1)\))?$/
```

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -2812,6 +2817,40 @@ public function userLogsIn(string $user) {
...
+					'contents' => $file !== 'invalid' ? fopen(__DIR__ . '/../../../..' . $file, 'r') : '',
+				],
+			],
+		];
+		$this->sendRequest('POST', '/apps/spreed/api/' . $apiVersion . '/room/' . self::$identifierToToken[$identifier] . '/avatar', null, [], $options);
+		$this->assertStatusCode($this->response, $statusCode);
+	}
+
+	/**
+	 * @When /^the room "([^"]*)" need to have an avatar with (\d+)(?: \((v1)\))?$/
```
nickvergessen, line 2838:
```suggestion
	 * @When /^the room "([^"]*)" has an avatar with (\d+)(?: \((v1)\))?$/
```

lib/Service/AvatarService.php
```diff
@@ -0,0 +1,174 @@
...
+		return $room->getType() === Room::TYPE_ONE_TO_ONE;
+	}
+
+	public function getAvatarVersion(Room $room, ?string $userId): string {
+		if ($room->getType() === Room::TYPE_ONE_TO_ONE && $userId) {
+			return (string) $this->config->getUserValue($userId, 'avatar', 'version', '0');
+		}
+		return (string) ($this->cache->get($room->getToken() . '.avatarVersion') ?? 0);
+	}
```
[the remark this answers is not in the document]
vitormattos: The version is only a string to expire the avatar URL if the image will be updated.
At this moment, every time the room only will have an avatar, not multiple. If the redis is restarted, will expire all avatar URL and do a new GET to the same avatar.

appinfo/info.xml
```diff
@@ -16,7 +16,7 @@ And in the works for the [coming versions](https://github.com/nextcloud/spreed/m
 
 	]]></description>
 
-	<version>16.0.0-beta.4</version>
+	<version>16.0.0-beta.5</version>
```
nickvergessen, line 19: Should be dev. I forgot to reset it on the bump:
```suggestion
	<version>16.0.0-dev.1</version>
```

docs/conversation.md
nickvergessen: but it's missing from the docs?!

lib/Activity/Listener.php
```diff
@@ -175,6 +175,7 @@ public function generateCallActivity(Room $room, bool $endForEveryone = false, ?
 				->setTimestamp($this->timeFactory->getTime())
 				->setSubject('call', [
 					'room' => $room->getId(),
+					'avatar' => $room->getAvatar(),
```
nickvergessen, line 178: This should not be written into the database, but that is what you are doing here.

If you want to add it to the parameter (which I would **not** do), you actually have to do that here:
https://github.com/nextcloud/spreed/blob/659c2ba7ff5bad286f9ed3f456b5759b0906c8c6/lib/Activity/Provider/Base.php#L115-L121
But then you would also need to add it here to the docs:
https://github.com/nextcloud/server/blob/master/lib/public/RichObjectStrings/Definitions.php#L189-L194

---

But I think you should do the change here instead:
https://github.com/nextcloud/spreed/blob/659c2ba7ff5bad286f9ed3f456b5759b0906c8c6/lib/Activity/Provider/Base.php#L75-L79

lib/Chat/ChatManager.php
```diff
@@ -701,7 +705,8 @@ public function addConversationNotify(array $results, string $search, Room $room
 			array_unshift($results, [
 				'id' => 'all',
 				'label' => $roomDisplayName,
-				'source' => 'calls'
+				'source' => 'calls',
+				'avatar' => $this->avatarService->getAvatarUrl($room),
```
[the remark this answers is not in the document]
nickvergessen: Still needs to be removed

lib/Activity/Listener.php
```diff
@@ -221,6 +222,7 @@ public function generateInvitationActivity(Room $room, array $participants): voi
 				->setSubject('invitation', [
 					'user' => $actor->getUID(),
 					'room' => $room->getId(),
+					'avatar' => $room->getAvatar(),
```
nickvergessen, line 225: Also here, this is not the right place, please remove

lib/Activity/Listener.php
```diff
@@ -253,6 +255,7 @@ public function generateInvitationActivity(Room $room, array $participants): voi
 					->setSubject('invitation', [
 						'user' => $actor->getUID(),
 						'room' => $room->getId(),
+						'avatar' => $room->getAvatar(),
```
nickvergessen, line 258: Also here, this is not the right place, please remove

lib/Collaboration/Reference/TalkReferenceProvider.php
```diff
@@ -223,7 +227,7 @@ protected function fetchReference(Reference $reference): void {
 		$reference->setTitle($title);
 		$reference->setDescription($description);
 		$reference->setUrl($this->urlGenerator->linkToRouteAbsolute('spreed.Page.showCall', ['token' => $room->getToken()]));
-		$reference->setImageUrl($this->getRoomIconUrl($room, $this->userId));
+		$reference->setImageUrl($this->getRoomIconUrl($room));
```
nickvergessen, line 230:
```suggestion
		$reference->setImageUrl($this->avatarService->getAvatarUrl($room));
```

lib/Collaboration/Reference/TalkReferenceProvider.php
```diff
@@ -257,18 +261,8 @@ public function getCacheKey(string $referenceId): ?string {
...
-					'userId' => $room->getSecondParticipant($userId),
-					'size' => 64,
-				]
-			);
-		}
-
-		return $this->urlGenerator->getAbsoluteURL($this->urlGenerator->imagePath('spreed', 'changelog.svg'));
+	protected function getRoomIconUrl(Room $room): string {
+		return $this->avatarService->getAvatarUrl($room);
 	}
```
nickvergessen, line 266:
```suggestion
```

lib/Controller/RoomController.php
```diff
@@ -479,6 +483,7 @@ protected function formatRoomV4(Room $room, ?Participant $currentParticipant, ?a
 			'description' => $room->getDescription(),
 			'listable' => $room->getListable(),
 			'messageExpiration' => $room->getMessageExpiration(),
+			'avatar' => $this->avatarService->getAvatarUrl($room),
```
nickvergessen, line 486: Should always be added. Please move to the array in line 375:
```php
$roomData = [
```

nickvergessen: Also it's missing from the documentation in conversation.md

lib/Dashboard/TalkWidget.php
```diff
@@ -205,36 +209,12 @@ protected function prepareRoom(Room $room, string $userId): WidgetItem {
 			$room->getDisplayName($userId),
 			$subtitle,
 			$this->url->linkToRouteAbsolute('spreed.Page.showCall', ['token' => $room->getToken()]),
-			$this->getRoomIconUrl($room, $userId)
+			$this->getRoomIconUrl($room)
```
nickvergessen, line 212:
```suggestion
			$this->avatarService->getAvatarUrl($room)
```

lib/Dashboard/TalkWidget.php
```diff
@@ -205,36 +209,12 @@ protected function prepareRoom(Room $room, string $userId): WidgetItem {
...
-		} elseif ($room->getObjectType() === 'emails') {
-			return $this->url->getAbsoluteURL($this->url->imagePath('core', 'actions/mail.svg'));
-		} elseif ($room->getType() === Room::TYPE_PUBLIC) {
-			return $this->url->getAbsoluteURL($this->url->imagePath('core', 'actions/public.svg'));
-		}
-
-		return $this->url->getAbsoluteURL($this->url->imagePath('core', 'actions/group.svg'));
+	protected function getRoomIconUrl(Room $room): string {
+		return $this->avatarService->getAvatarUrl($room);
 	}
```
nickvergessen, line 218:
```suggestion
```

lib/Service/AvatarService.php
```diff
@@ -0,0 +1,214 @@
...
+			'image/jpeg',
+			'image/png',
+		];
+		if (!in_array($mimeType, $allowedMimeTypes)) {
+			throw new InvalidArgumentException($this->l->t('Unknown filetype'));
+		}
+
+		$token = $room->getToken();
+		$avatarFolder = $this->getAvatarFolder($token);
+		$avatarName = UUIDUtil::getUUID();
```
nickvergessen, line 120: This is a function of 3rdparty tool SabreDav.
I would recommend to store with the token only?
If you want to use random, please use `OCP\Security\ISecureRandom::generate` and make sure to limit the characters to `OCP\Security\ISecureRandom::CHAR_HUMAN_READABLE` or something so slash is not included

lib/Migration/Version16000Date20221116163301.php
```diff
@@ -0,0 +1,56 @@
...
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('talk_rooms');
+		if (!$table->hasColumn('avatar')) {
+			$table->addColumn('avatar', Types::STRING, [
+				'notnull' => false,
+				'length' => 255,
```
nickvergessen, line 50: that's a lot of data. 64 should be more than enough, especially since it's in a subfolder per conversation token at the moment?

lib/Search/ConversationSearch.php
```diff
@@ -113,20 +117,21 @@ public function search(IUser $user, ISearchQuery $query): SearchResult {
...
+				$arguments = [
+					'token' => $room->getToken(),
+					'apiVersion' => 'v1',
+				];
+				if ($avatar = $room->getAvatar()) {
+					$arguments['v'] = $avatar;
 				}
-			} elseif ($room->getObjectType() === 'file') {
-				$iconClass = 'conversation-icon icon-file-white';
+				$icon = $this->url->linkToRouteAbsolute('ocs.spreed.Avatar.getAvatar', $arguments);
```
nickvergessen, line 128:
```suggestion
				$icon = $this->url->linkToOCSRouteAbsolute('spreed.Avatar.getAvatar', $arguments);
```

lib/Service/AvatarService.php
```diff
@@ -0,0 +1,214 @@
...
+		$avatarFolder->newFile($avatarName, $image->data());
```
nickvergessen, line 121: You only generate new avatars, but never delete the previous one?!

```suggestion
		$avatarFolder->newFile($avatarName, $image->data());
		
		try {
			$oldAvatar = $avatarFolder->getFile($room->getAvatar());
			$oldAvatar->delete();
		} catch (NotFoundException $e) {
		}
```

lib/Service/AvatarService.php
```diff
@@ -0,0 +1,214 @@
...
+		$room->setAvatar($avatarName);
```
nickvergessen, line 122: This should be done by the room service, as it does that already for all the other methods.

lib/Service/RoomService.php
```diff
@@ -336,6 +336,17 @@ public function setLobby(Room $room, int $newState, ?\DateTime $dateTime, bool $
 		return true;
 	}
 
+	public function setAvatar(Room $room, $avatarName): bool {
+		if ($room->getType() === Room::TYPE_ONE_TO_ONE) {
+			return false;
+		}
+		$update = $this->db->getQueryBuilder();
+		$update->update('talk_rooms')
```
nickvergessen, line 344: query is not executed

lib/Service/RoomService.php
```diff
@@ -336,6 +336,17 @@ public function setLobby(Room $room, int $newState, ?\DateTime $dateTime, bool $
...
+			->set('avatar', $avatarName)
+			->where($update->expr()->eq('id', $update->createNamedParameter($room->getId(), IQueryBuilder::PARAM_INT)));
+		return true;
```
nickvergessen, line 347: room object is not modified

nickvergessen: Needs to dispatch an event, so we can later on create a system message, just compare with all the other methods around this one.

nickvergessen (changes requested):

lib/Activity/Listener.php
vitormattos: We don't store two formats of image. What we will do when need PNG and the image is SVG? We will drop support to SVG?

nickvergessen: Conversation avatars can not be SVG, so yeah the safety net can be removed and we just always return the PNG conversation avatar

nickvergessen, line 178: Remove this, we shouldn't write it into the DB

lib/Activity/Provider/Base.php
```diff
@@ -67,15 +72,18 @@ public function preParse(IEvent $event): IEvent {
...
 			throw new \InvalidArgumentException('User can not user Talk');
 		}
 
-		if ($this->activityManager->getRequirePNG()) {
+		try {
+			$room = $this->manager->getRoomForUser($event->getObjectId(), $uid);
+			$event->setIcon($this->avatarService->getAvatarUrl($room, $uid));
+		} catch (RoomNotFoundException $th) {
+			// When the roomId wont exists and to prevent an exception
 			$event->setIcon($this->url->getAbsoluteURL($this->url->imagePath('spreed', 'app-dark.png')));
```
nickvergessen, line 86:
```suggestion
			if ($this->activityManager->getRequirePNG()) {
				$event->setIcon($this->url->getAbsoluteURL($this->url->imagePath('spreed', 'app-dark.png')));
			} else {
				$event->setIcon($this->url->getAbsoluteURL($this->url->imagePath('spreed', 'app-dark.svg')));
			}
```

lib/Activity/Provider/Base.php
```diff
@@ -117,6 +125,7 @@ protected function getRoom(Room $room, string $userId): array {
 			'id' => $room->getId(),
 			'name' => $room->getDisplayName($userId),
 			'link' => $this->url->linkToRouteAbsolute('spreed.Page.showCall', ['token' => $room->getToken()]),
+			'iconUrl' => $this->avatarService->getAvatarUrl($room, $userId),
```
nickvergessen, line 128:
```suggestion
			'icon-url' => $this->avatarService->getAvatarUrl($room, $userId),
```

lib/Chat/ChatManager.php
```diff
@@ -123,6 +126,7 @@ public function __construct(CommentsManager $commentsManager,
 		$this->roomService = $roomService;
 		$this->pollService = $pollService;
 		$this->notifier = $notifier;
+		$this->avatarService = $avatarService;
```
nickvergessen, line 129: Unused, remove

lib/Controller/RoomController.php
```diff
@@ -412,6 +416,7 @@ protected function formatRoomV4(Room $room, ?Participant $currentParticipant, ?a
 			'listable' => Room::LISTABLE_NONE,
 			'callFlag' => Participant::FLAG_DISCONNECTED,
 			'messageExpiration' => 0,
+			'avatar' => $this->avatarService->getAvatarUrl($room),
```
nickvergessen, line 419: This is the full avatar url, but the documentation more sounds like it's only the `v=` part?

lib/Search/ConversationSearch.php
```diff
@@ -113,36 +117,21 @@ public function search(IUser $user, ISearchQuery $query): SearchResult {
...
-			} else {
-				$iconClass = 'conversation-icon icon-contacts-white';
+			$arguments = [
+				'token' => $room->getToken(),
+				'apiVersion' => 'v1',
+			];
+			if ($avatar = $room->getAvatar()) {
+				$arguments['v'] = $avatar;
 			}
+			$icon = $this->url->linkToOCSRouteAbsolute('spreed.Avatar.getAvatar', $arguments);
```
nickvergessen, line 127: Why does this search here have so much special handling.

Calling `$this->avatarService->getAvatarUrl($room)` like in dashboard, activity and notifications should be good enough?!

lib/Service/AvatarService.php
```diff
@@ -0,0 +1,236 @@
...
+		$token = $room->getToken();
+		$avatarFolder = $this->getAvatarFolder($token);
+		$avatarName = $this->getRandomAvatarName($room);
+		$avatarFolder->newFile($avatarName, $image->data());
+		$room->setAvatar($avatarName);
+		$this->roomService->setAvatar($room, $avatarName);
+	}
+
+	private function getRandomAvatarName(Room $room): string {
+		$name = $this->random->generate(10, ISecureRandom::CHAR_HUMAN_READABLE);
```
nickvergessen, line 132: That's the size of the db, but we need 4 chars for the file suffix like `.png` ?

lib/Service/AvatarService.php
```diff
@@ -0,0 +1,236 @@
...
+		$mimeType = $image->mimeType();
+		$allowedMimeTypes = [
+			'image/jpeg',
+			'image/png',
+		];
+		if (!in_array($mimeType, $allowedMimeTypes)) {
+			throw new InvalidArgumentException($this->l->t('Unknown filetype'));
+		}
+
+		$this->deleteAvatarFromFilesystem($room);
```
nickvergessen, line 121: Would do this only after the `newFile` was written successfully?

lib/Service/AvatarService.php
nickvergessen: You added it in the server, but didn't remove it here?

lib/Service/AvatarService.php
```diff
@@ -0,0 +1,236 @@
...
+
+	public function getAvatarUrl(Room $room): string {
+		$arguments = [
+			'token' => $room->getToken(),
+			'apiVersion' => 'v1',
+		];
+		if ($avatar = $room->getAvatar()) {
+			$arguments['v'] = $avatar;
+		}
+		return $this->url->linkToRouteAbsolute('ocs.spreed.Avatar.getAvatar', $arguments);
```
nickvergessen, line 234:
```suggestion
		return $this->url->linkToOCSRouteAbsolute('spreed.Avatar.getAvatar', $arguments);
```

lib/Service/RoomService.php
```diff
@@ -336,6 +336,26 @@ public function setLobby(Room $room, int $newState, ?\DateTime $dateTime, bool $
 		return true;
 	}
 
+	public function setAvatar(Room $room, $avatarName): bool {
```
nickvergessen, line 339:
```suggestion
	public function setAvatar(Room $room, string $avatarName): bool {
```

lib/Service/RoomService.php
```diff
@@ -336,6 +336,26 @@ public function setLobby(Room $room, int $newState, ?\DateTime $dateTime, bool $
...
+		if ($room->getType() === Room::TYPE_ONE_TO_ONE) {
+			return false;
+		}
+
+		$event = new ModifyRoomEvent($room, 'avatar', $avatarName, $room->getAvatar());
+		$this->dispatcher->dispatch(Room::EVENT_BEFORE_AVATAR_SET, $event);
+
+		$room->setAvatar($avatarName);
```
nickvergessen, line 347: Should be after the DB update as in other methods.

nickvergessen (changes requested):

lib/Manager.php
```diff
@@ -173,6 +173,7 @@ public function createRoomObject(array $row): Room {
 			(string) $row['name'],
 			(string) $row['description'],
 			(string) $row['password'],
+			(string) $row['avatar'],
```
nickvergessen, line 176: you cast to string here, but the setter allows null, so just allow it here too:
```suggestion
			$row['avatar'],
```

lib/Room.php
```diff
@@ -199,6 +202,7 @@ public function __construct(Manager $manager,
 								string $name,
 								string $description,
 								string $password,
+								string $avatar,
```
nickvergessen, line 205:
```suggestion
								?string $avatar,
```

lib/Service/AvatarService.php
```diff
@@ -0,0 +1,235 @@
...
+			}
+		}
+		return $file;
+	}
+
+	private function deleteAvatarFromFilesystem(Room $room): void {
+		$token = $room->getToken();
+		$avatarFolder = $this->getAvatarFolder($token);
+		if (count($avatarFolder->getDirectoryListing()) >= 1) {
+			$avatarFolder->delete();
```
nickvergessen, line 193: There is no need to remove the complete folder.
Could just loop over `getDirectoryListing` and delete all images which are not the new one?

lib/Service/AvatarService.php
```diff
@@ -0,0 +1,235 @@
...
+		}
+		return $room->getType() === Room::TYPE_ONE_TO_ONE;
+	}
+
+	public function getAvatarVersion(Room $room, ?string $userId): string {
+		if ($room->getType() === Room::TYPE_ONE_TO_ONE && $userId) {
+			return (string) $this->config->getUserValue($userId, 'avatar', 'version', '0');
+		}
+		return (string) ($this->cache->get($room->getToken() . '.avatarVersion') ?? 0);
+	}
```
nickvergessen, line 223:
```suggestion
```

both functions not used (also `$this->cache` is undefined)

lib/Service/AvatarService.php
```diff
@@ -0,0 +1,235 @@
...
+
+	public function getAvatarUrl(Room $room): string {
+		$arguments = [
+			'token' => $room->getToken(),
+			'apiVersion' => 'v1',
+		];
+		if ($avatar = $room->getAvatar()) {
```
nickvergessen, line 230:
```suggestion
		$avatar = $room->getAvatar()
		if ($avatar) {
```

We don't do inline assignments in Nextcloud

docs/conversation.md
```diff
@@ -88,6 +88,7 @@
...
 | `guestList`             | string  | v1    | v2      | **Removed**                                                                                                                                                                                                                                                                                                                                                                                       |
+| `avatarUrl`             | string  | v4    |         | Avatar URL of the conversation including a version flag `v=…` for easier expiration of the avatar in case a moderator updates it, since the avatar endpoint should be cached for 24 hours.                                                                                                                                                                                                        |
```
nickvergessen, line 91: I think this parameter is a bit problematic. It will make the avatar look updated on the conversation list, but e.g. search results, notifications, etc still show the old one. So basically we would need to tell clients to track this value and when it is different, they should invalidate their avatar image cache.

nickvergessen: cc @Ivansss @mahibi 

Any idea what the best way would be? I'd like to have it similar to user avatars. They work without any additional information, you just need the userid (conversation token) and can make your URL. It's cached so can take some time until others see it. Should be fine, unless you did the change yourself.
Would you think it's okay to have it cached/old on the mobile apps when you changed it in the web? Or what would be a good way to handle this?

lib/Service/AvatarService.php
```diff
@@ -0,0 +1,208 @@
...
+
+		$token = $room->getToken();
+		$avatarFolder = $this->getAvatarFolder($token);
+
+		// Delete previous avatars
+		foreach ($avatarFolder->getDirectoryListing() as $file) {
+			$file->delete();
+		}
+
+		$avatarName = $this->random->generate(10, ISecureRandom::CHAR_HUMAN_READABLE);
```
nickvergessen, line 130: Should add the file extension?

```suggestion
		$avatarName = $this->random->generate(6, ISecureRandom::CHAR_HUMAN_READABLE) . ($mimeType === 'image/jpeg' ? '.jpg' : '.png');
```

vitormattos: Is possible, but we will need change the current implementation a bit to consider the extension.
For now, the name that we stored on database is only to have a token to expire the previous image. Now, if we send a request without the version, will get the room image too.

lib/Service/AvatarService.php
```diff
@@ -200,7 +200,7 @@ public function getAvatarUrl(Room $room): string {
 		];
 
 		$avatarVersion = $room->getAvatar();
-		if ($avatarVersion) {
+		if ($avatarVersion !== '') {
```
vitormattos, line 203: To do this we will need set a default value in database to empty, for now, the default value is null.

nickvergessen: I fixed that. And also the Manager was already casting to string anyway.

docs/conversation.md
nickvergessen: > The frontend every time get the version directly from userconfig:

This is only for your very own avatar. It is not done for the avatar of others, see:
https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/NcAvatar/NcAvatar.vue#L627-L647

But that is exactly the problem here. It's basically always the avatar of others 😬

vitormattos: I think that this wont be a problem to Talk.

I already implemented to send the version of avatar in Room object, then, in all places that we need to display the current room avatar, will be possible.

In the places that we haven't the Room object, we will send the URL of avatar and we already have a contract to do this, have places in server that need a property called icon, other use iconUrl and we can do the same in Talk.

nickvergessen: That is exactly why I started the discussion here. We will NOT be able to add the avatar URL on all things. But anyway.

We can move this to a follow up issue so we can merge this PR

nickvergessen:
Rebased to resolve conflicts from breakout rooms part1 merge

nickvergessen (approved):
Added a commit so the file can be viewed in the browser and does not trigger a download.

Other than that only the system messages are missing, but we can do that as a follow up.

nickvergessen:
Follow up in https://github.com/nextcloud/spreed/issues/8443

PR #8340, opened 2022-11-10
🖖 Breakout rooms 🛠️ API - Part 1

### LINKED ISSUE #8338: 🖖 Breakout rooms 🛠️ API
Meta ticket at https://github.com/nextcloud/spreed/issues/8337

## Eta 18.11.

- [x] Capability for version
- [x] Configuring breakout rooms before the call
  - [x] Automatic distribution of equal size
  - [x] Manual distribution
  - [x] Integration tests
  - [x] API Documentation
- [x] Make moderators a moderator in all breakout rooms
  - [x] Remove from all breakout rooms on demotion
  - [x] Integration tests
- [x] Don't allow breakout rooms in:
  - [x] One-to-one
  - [x] Breakout-rooms
  - [x] Should we limit further? Amount: 1 <= X <= 20
- [x] https://github.com/nextcloud/spreed/pull/8340

## Eta 25.11.
- [x] Capability for yes/no
  - [x] Integration tests
  - [x] API Documentation
- [x] Create API for moderator to post a message to all chat rooms
  - [x] Integration tests
  - [x] API Documentation
- [x] Create API to start and stop the break out rooms
  - [x] Set/remove the lobby
  - [x] Need a flag on the conversation object/api
  - [x] Integration tests
  - [x] API Documentation
- [x] https://github.com/nextcloud/spreed/pull/8370

## Eta 07.12.
- [x] Based on #8370 
- [x] Create API for participants to "raise hand" across the rooms
  - [x] Integration tests
  - [x] API Documentation
- [x] Create API for participants to abort/reset assistance request
  - [x] Integration tests
  - [x] API Documentation
- [x] Configuring breakout rooms before the call
  - [x] Free distribution (?)
    - [x] Expose the list of potential breakout rooms (⚠️ listed convos would expose to everyone on the instance)
    - [x] Integration tests
    - [x] API Documentation
- [x] Allow students to switch between breakout rooms in case of free distribution
  - [x] Integration tests
  - [x] API Documentation
- [x] https://github.com/nextcloud/spreed/pull/8442

## 💬 Follow ups / Discussions: Part 1
- [x] Delete breakout rooms when the parent room is being deleted
- [x] Allow to create breakout rooms by providing objectType + objectId
  - [x] Lobby state is correctly set
  - [x] Moderators are added and promoted
- [x] Should prevent direct deletion of individual rooms when they are breakout rooms (as it breaks the mapping)?
  - [x] Allow this and use it as mechanism to manually rearrange the breakout room situation
- [x] When a user is added directly to a breakout room, we should add them to the parent room too or prevent adding
  - [x] Allow this and remove from other breakout rooms and use it as mechanism to manually rearrange the breakout room situation
- [x] ~~Moderators should be able to move single participants between breakout rooms~~ See above point
- [x] Moderators should be able to provide the map afterwards again
- [x] Remove participants from breakout rooms when removed from parent
- [x] Prevent changes directly on breakout rooms
  - [x] Guests allow/disallow
  - [x] Lobby
- [x] #8462 

## 💬 Follow ups / Discussions: Part 2
- [x] ~~Allow timers for the breakout rooms which automatically stops them~~
- [ ] Signaling messages for transfering users to the breakout rooms and back. https://github.com/nextcloud/spreed/pull/8477
- [ ] #8545

nickvergessen:
Ref #8338 

### 🚧 TODO

- [x] Capability for version
- [x] Configuring breakout rooms before the call
  - [x] Automatic distribution of equal size
  - [x] Manual distribution
  - [x] Integration tests
  - [x] API Documentation
- [x] Make moderators a moderator in all breakout rooms
  - [x] Remove from all breakout rooms on demotion
  - [x] Integration tests
- [x] Don't allow breakout rooms in:
  - [x] One-to-one
  - [x] Breakout-rooms
  - [x] Should we limit further? Amount: 1 <= X <= 20
  - [x] Integration tests

### 🏁 Checklist

### DIFF
--- a/docs/breakout-rooms.md
+++ b/docs/breakout-rooms.md
@@ -1,0 +1,48 @@
+# Breakout rooms management
+
+Group and public conversations can be used to host breakout rooms.
+
+* Only moderators can configure and remove breakout out rooms
+* Only moderators can start and stop breakout out rooms
+* Moderators in the parent conversation are added as moderators to all breakout rooms and remove from all on demotion
+
+## Base endpoint
+
+* API v1: Base endpoint `/ocs/v2.php/apps/spreed/api/v1`: since Nextcloud 26
+
+## Configure breakout rooms
+
+* Required capability: `breakout-rooms-v1`
+* Method: `POST`
+* Endpoint: `/breakout-rooms/{token}`
+* Data:
+
+| field         | type   | Description                                                                                          |
+|---------------|--------|------------------------------------------------------------------------------------------------------|
+| `mode`        | int    | Participant assignment mode (see [constants list](constants.md#breakout-room-modes))                 |
+| `amount`      | int    | Number of breakout rooms to create (Minimum `1`, maximum `20`)                                       |
+| `attendeeMap` | string | A json encoded Map of attendeeId => room number (0 based) (Only considered when the mode is "manual" |
+
+* Response:
+    - Status code:
+        + `200 OK`
+        + `400 Bad Request` When breakout rooms are already configured
+        + `400 Bad Request` When the conversation is not a group or public conversation
+        + `400 Bad Request` When the conversation is a breakout room itself
+        + `400 Bad Request` When the mode is invalid
+        + `400 Bad Request` When the amount is below the minimum or above the maximum
+        + `400 Bad Request` When the attendee map contains an invalid room number
+        + `403 Forbidden` When the current user is not a moderator/owner
+        + `404 Not Found` When the conversation could not be found for the participant
+
+## Remove breakout rooms
+
+* Required capability: `breakout-rooms-v1`
+* Method: `DELETE`
+* Endpoint: `/breakout-rooms/{token}`
+
+* Response:
+    - Status code:
+        + `200 OK`
+        + `403 Forbidden` When the current user is not a moderator/owner
+        + `404 Not Found` When the conversation could not be found for the participant
--- a/docs/breakout-rooms.md
+++ b/docs/breakout-rooms.md
@@ -2,8 +2,8 @@
 
 Group and public conversations can be used to host breakout rooms.
 
-* Only moderators can configure and remove breakout out rooms
-* Only moderators can start and stop breakout out rooms
+* Only moderators can configure and remove breakout rooms
+* Only moderators can start and stop breakout rooms
 * Moderators in the parent conversation are added as moderators to all breakout rooms and remove from all on demotion
 
 ## Base endpoint
--- a/lib/Migration/Version16000Date20221110151714.php
+++ b/lib/Migration/Version16000Date20221110151714.php
@@ -52,6 +52,9 @@ class Version16000Date20221110151714 extends SimpleMigrationStep {
 			$table->addColumn('breakout_room_mode', Types::INTEGER, [
 				'default' => 0,
 			]);
+			$table->addColumn('breakout_room_status', Types::INTEGER, [
+				'default' => 0,
+			]);
 			return $schema;
 		}
 
--- a/lib/Migration/Version16000Date20221110151714.php
+++ b/lib/Migration/Version16000Date20221110151714.php
@@ -1,0 +1,60 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (c) 2022 Joas Schilling <coding@schilljs.com>
+ *
+ * @author Joas Schilling <coding@schilljs.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Talk\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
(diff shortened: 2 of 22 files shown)


lib/Service/BreakoutRoomService.php
```diff
@@ -0,0 +1,237 @@
...
+
+				if (min($attendeeMap) < 0) {
+					throw new InvalidArgumentException('attendeeMap');
+				}
+			}
+		}
+
+		$breakoutRooms = $this->createBreakoutRooms($parent, $amount);
+
+		$participants = $this->participantService->getParticipantsForRoom($parent);
```
vitormattos, line 114: I think that will be good to check if all members of `$attendeeMap` are participants.

lib/Service/BreakoutRoomService.php
```diff
@@ -0,0 +1,237 @@
...
+	/**
+	 * @param Room $parent
+	 * @param int $mode
+	 * @psalm-param 0|1|2|3 $mode
+	 * @param int $amount
+	 * @param string $attendeeMap
+	 * @return Room[]
+	 * @throws InvalidArgumentException When the breakout rooms are configured already
+	 */
+	public function setupBreakoutRooms(Room $parent, int $mode, int $amount, string $attendeeMap): array {
```
vitormattos, line 66: This var name will make the code more readable in all places of this class that use `$parent` as Room object.
```suggestion
	public function setupBreakoutRooms(Room $parentRoom, int $mode, int $amount, string $attendeeMap): array {
```

lib/Service/BreakoutRoomService.php
```diff
@@ -0,0 +1,237 @@
...
+
+	public function createBreakoutRooms(Room $parent, int $amount): array {
+		// Safety caution cleaning up potential orphan rooms
+		$this->deleteBreakoutRooms($parent);
+
+		$rooms = [];
+		for ($i = 1; $i <= $amount; $i++) {
+			$breakoutRoom = $this->roomService->createConversation(
+				$parent->getType(),
+				str_replace('{number}', (string) $i, $this->l->t('Room {number}')),
```
vitormattos, line 212: Is good to put a hint to translators to make more easy to translators identify the place of this phrase will occur and get a best understanding about what is this phrase.

https://docs.nextcloud.com/server/latest/developer_manual/basics/front-end/l10n.html#provide-context-hints-for-translators

```suggestion
				// TRANSLATORS The room name followed with a sequential number from 1 to N where N is the total of breakout rooms.
				$this->l->t('Room %n', $i),
```

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -555,11 +560,16 @@ public function userSeesAttendeesInRoom(string $user, string $identifier, int $s
 					$attendee['actorId'] .= '@' . rtrim($this->baseRemoteUrl, '/');
 				}
 
+				// Breakout room regex
+				if (isset($attendee['actorId']) && strpos($attendee['actorId'], '/') === 0 && preg_match($attendee['actorId'], $actual['actorId'])) {
+					$attendee['actorId'] = $actual['actorId'];
```
vitormattos, line 565: Could be good to break a bit this row to make more easy the use of xdebug and prevent wrap the row in editor.
```suggestion
				if (isset($attendee['actorId'])) {
					if (strpos($attendee['actorId'], '/') === 0) {
						if (preg_match($attendee['actorId'], $actual['actorId'])) {
					$attendee['actorId'] = $actual['actorId'];
```

lib/Migration/Version16000Date20221110151714.php
```diff
@@ -0,0 +1,63 @@
...
+
+		$table = $schema->getTable('talk_rooms');
+
+		if (!$table->hasColumn('breakout_room_mode')) {
+			$table->addColumn('breakout_room_mode', Types::INTEGER, [
+				'default' => 0,
+			]);
+			$table->addColumn('breakout_room_status', Types::INTEGER, [
+				'default' => 0,
+			]);
```
vitormattos, line 57: I didn't identified the place that we will use the status value in this PR.

nickvergessen: It's in the follow up #8370 but I want to use only 1 migration to reduce database time on update

docs/breakout-rooms.md
```diff
@@ -0,0 +1,48 @@
+# Breakout rooms management
+
+Group and public conversations can be used to host breakout rooms.
+
+* Only moderators can configure and remove breakout out rooms
+* Only moderators can start and stop breakout out rooms
```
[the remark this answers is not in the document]
nickvergessen: totally right :D

lib/Service/BreakoutRoomService.php
nickvergessen: We can not use `%n` as it is not a plural. We could use an ordinary `%s` placeholder but I find it better with "speaking" placeholders like in javascript.
But I will add a translator hint additionally

tests/integration/features/bootstrap/FeatureContext.php
nickvergessen: With xdebug just inspect the array?
If I split it like this PHPStorm complains that they should be combined into one:
image

lib/Service/BreakoutRoomService.php
nickvergessen: I think it should be obvious enough as there is no other item that could be a parent. Also the type hint indicates it as well.
Also I want to avoid breakings now with the follow up pull requests.

tests/integration/features/bootstrap/FeatureContext.php
vitormattos: PHPStorm isn't a good boy 😂
The problem with xdebug is because is necessary to check condition by condition to identify if the sentence will be true. But, the most important is to work.

lib/Service/BreakoutRoomService.php
vitormattos: This is a good point to follow up to server and change the `t` function to accept the same syntax as JS.

lib/Service/BreakoutRoomService.php
vitormattos: Is a good point to move to part 2.

lib/Service/BreakoutRoomService.php
vitormattos: Only to prevent to send in this array an user that isn't member of room.

vitormattos (approved):

PR #8391, opened 2022-11-23
Move all API version requirements to the same format

### LINKED ISSUE #8354: Fix apiVersion regexs
### How to use GitHub

---

The `apiVersion` requirements are sometimes plain strings, but should be regexes like this: `v4` -> `v(4)`.  
Additionally all these regex should have `^...$` in order to make them match to the whole parameter and not just parts of it.

nickvergessen:
This allows better parsing for OpenAPI specs later

Fix #8354 

### 🏁 Checklist

### DIFF
--- a/appinfo/routes/routesCallController.php
+++ b/appinfo/routes/routesCallController.php
@@ -24,8 +24,8 @@ declare(strict_types=1);
  */
 
 $requirements = [
-	'apiVersion' => 'v(4)',
-	'token' => '^[a-z0-9]{4,30}$',
+	'apiVersion' => 'v4',
+	'token' => '[a-z0-9]{4,30}',
 ];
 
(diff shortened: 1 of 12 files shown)


provokateurin (approved):

nickvergessen:
After some digging it seems that `^` and `$` are stripped away anyway.
I can see use doing both things now. Either the regex contain it (but that is not the case for most requirements across Nextcloud universe) or your script adds it?

nickvergessen:
https://github.com/nextcloud/3rdparty/blob/e0b9473487114accdba2e5a829ba763589ce5860/symfony/routing/Route.php#L520-L539

provokateurin:
i would omit them to make it easier to adapt other apps

PR #8402, opened 2022-11-28
Remove unused prop

marcoambrosini:


ShGKme (approved):
Looks ok, as this prop is unused, and this component is not used with this prop.

PR #8409, opened 2022-11-30
Fix some eslint warnings

marcoambrosini:


src/store/callViewStore.js
```diff
@@ -91,7 +91,7 @@ const mutations = {
 			throw new Error('Missing or empty sessionId argument in call to setParticipantHandRaised')
 		}
 		if (raisedHand && raisedHand.state) {
-			Vue.set(state.participantRaisedHands, sessionId, raisedHand)
+			set(state.participantRaisedHands, sessionId, raisedHand)
```
nickvergessen, line 94: Forgot to import?

> /home/runner/work/spreed/spreed/src/store/callViewStore.js
> Error:    94:4  error  'set' is not defined  no-undef
> Error:   103:3  error  'set' is not defined  no-undef

src/services/breakoutRoomsService.js
```diff
@@ -25,12 +25,12 @@ import { generateOcsUrl } from '@nextcloud/router'
...
- * @param mode Either manual, auto, or free, see constants file
- * @param amount The amount of breakout rooms to be created
- * @param attendeeMap A json encoded Map of attendeeId => room number (0 based)
+ * @param {string} token The conversation token
+ * @param {string} mode Either manual, auto, or free, see constants file
+ * @param {number} amount The amount of breakout rooms to be created
+ * @param {string} attendeeMap A json encoded Map of attendeeId => room number (0 based)
  * (Only considered when the mode is "manual")
- * @return {Promise<AxiosResponse<any>>}
+ * @return {Promise<any>} (AxiosResponse<any>)
```
ShGKme, line 33: To use types from packages in JSDocs we may use import.

```suggestion
 * @return {Promise<import('axios').AxiosResponse<any>>}
```

ShGKme (changes requested):
With import in JSDocs, the type will work as typing, not only text description.

src/services/breakoutRoomsService.js
Antreesy: We don't have `axios` as a dependency in spreed repository, only inside `@nextcloud/axios` bundle. But i'm not sure we could import a type from there, only if it was re-exported

ShGKme: > We don't have `axios` as a dependency in spreed repository, only inside `@nextcloud/axios` bundle. But i'm not sure we could import a type from there, only if it was re-exported

`axios` is a sub-dependency of `@nextcloud/axios`, it is not bundled to `@nextcloud/axios` build. So it is installed as well as ESLint or Webpack, for example.

This typing should work.

ShGKme (approved):
No more tons of warnings 🎉

PR #8445, opened 2022-12-08
⏺ Call recording ⚒ Server API - Add capability to toggle recording - Part 1

vitormattos:
Ref #8326 

### 🚧 TODO

- [ ] Document the system messages
  - [ ] `recording_started`
  - [ ] `recording_stopped`
  - [ ] `audio_recording_started` (Unsure if we implement this)
  - [ ] `audio_recording_stopped` (Unsure if we implement this)
- [  Document room API attribute for "call started at time" (`activeSince` of the room object) - suggestion: `callStartTime` https://github.com/nextcloud/spreed/pull/8369 
  - Ref https://github.com/nextcloud/spreed/pull/8369
  - [ ] Document it with the capability
- [ ] Document room API attribute for "the call is being recorded"
  - [ ] `callRecording` - `int`
    - [ ] 0 none
    - [ ] 1 video
    - [ ] 2 audio (Unsure if we implement this)
  - [ ] Document above constants in docs/constants.md
- [x] Document config capability whether recordings are on/off
  - [x] `config => call => recording` (bool)
- [x] Document capability for recording version
  - [x] `recording-v1`
- [ ] Document API endpoint to start a recording
  - [ ] `POST /ocs/v2.php/apps/spreed/api/v4/call/{token}/recording` Only available with `recording-v1` capability
      #8446
- [ ] Document API endpoint to stop a recording
  - [ ] `DELETE /ocs/v2.php/apps/spreed/api/v4/call/{token}/recording` Only available with `recording-v1` capability
      #8446

### 🏁 Checklist


lib/Config.php
```diff
@@ -132,6 +132,10 @@ public function canUserEnableSIP(IUser $user): bool {
 		return $this->canEnableSIP[$user->getUID()];
 	}
 
+	public function isBreakoutRoomsEnabled(): bool {
+		return $this->config->getAppValue('spreed', 'call_recording', 'yes') === 'yes';
```
nickvergessen, line 136: 1. Wrong function name
2. Should check for external signaling

```suggestion
	public function isRecordingEnabled(): bool {
		return $this->getSignalingMode() !== self::SIGNALING_INTERNAL && $this->config->getAppValue('spreed', 'call_recording', 'yes') === 'yes';
```

lib/Capabilities.php
```diff
@@ -145,6 +145,10 @@ public function getCapabilities(): array {
 			$capabilities['features'][] = 'reactions';
 		}
 
+		if ($this->talkConfig->isBreakoutRoomsEnabled()) {
+			$capabilities['features'][] = 'recording-v1';
```
nickvergessen, line 149: 1. Wrong config function
2. I would split this into 2 things (like with breakout rooms):
    1. A `recording-v1` which signals the API level
    https://github.com/nextcloud/spreed/blob/f121cccdab849dc5a5d2a960e8665f8adbb6d388/lib/Capabilities.php#L115
    3. A config => call => recording which signals the enabled/disabled:
    https://github.com/nextcloud/spreed/blob/f121cccdab849dc5a5d2a960e8665f8adbb6d388/lib/Capabilities.php#L124

nickvergessen (changes requested):

lib/Config.php
vitormattos: copy past.

vitormattos: I fixed in second PR and forgot to fix here.

lib/Capabilities.php
```diff
@@ -145,6 +146,10 @@ public function getCapabilities(): array {
 			$capabilities['features'][] = 'reactions';
 		}
 
+		if ($this->talkConfig->isRecordingEnabled()) {
+			$capabilities['features'][] = 'recording';
```
nickvergessen, line 150: This is supposed to be a config, not a feature. The recording-v1 is the feature. Just search for `breakout` in this file after rebasing on latest master for how it should be

lib/Config.php
```diff
@@ -132,6 +132,14 @@ public function canUserEnableSIP(IUser $user): bool {
 		return $this->canEnableSIP[$user->getUID()];
 	}
 
+	public function isRecordingEnabled(): bool {
+		$isSignalingInternal = $this->getSignalingMode() === self::SIGNALING_INTERNAL;
+		$callRecordingCapability = $this->config->getAppValue('spreed', 'call_recording', 'no');
```
nickvergessen, line 137: Please document in docs/settings.md but also the default should be yes, we don't want people to first have to enable this

nickvergessen (approved):

PR #8467, opened 2022-12-14
Wrap rooms creation in transaction

tcitworld:
### 🏁 Checklist

### DIFF
--- a/lib/Manager.php
+++ b/lib/Manager.php
@@ -34,6 +34,7 @@ use OCA\Talk\Model\SessionMapper;
 use OCA\Talk\Service\ParticipantService;
 use OCA\Talk\Service\RoomService;
 use OCP\App\IAppManager;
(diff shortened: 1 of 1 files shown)


nickvergessen:
Why is that? There is no benefit from a transaction over a single query.

tcitworld:
Yup, it's a read-after-write issue. I thought the read was there for a good reason, but I can change that to try and reconstruct the object.

nickvergessen:
We are currently heavily changing the affected code. Maybe you do a PR with a transaction for stable25 if it is having a real world negative impact at the moment on live servers, and we afterwards fix it properly for 26+ with direct construction of the object.
Can you raise a ticket for it, so we don't forget to track it?

nickvergessen:
Master part is now in #8520

nickvergessen:
25 rebased in https://github.com/nextcloud/spreed/pull/8531

PR #8471, opened 2022-12-16
Update documentation and expose room property

vitormattos:
Ref: #8326

### 🚧 TODO

- [x] `callStartTime`
  - [x] is not documented in https://github.com/nextcloud/spreed/blob/master/docs/conversation.md
  - [x] was missed to be added to the single room request (which is used by the HPB) https://github.com/nextcloud/spreed/blob/527ad87a78c3e18e9797681c076379e0c3bbe79e/lib/Controller/RoomController.php#L452-L472
- [x] `callRecording`
  - [x] is not exposed in the RoomController
    - [x] Also then needs documenting in https://github.com/nextcloud/spreed/blob/master/docs/conversation.md

### 🏁 Checklist


docs/conversation.md
```diff
@@ -91,6 +91,8 @@
...
+| `callStartTime`         | int     | v4    |         | Call start time                                                                                                                                                                                                                                                                                                                                                                                   |
+| `callRecording`         | int     | v4    |         | Call recording status                                                                                                                                                                                                                                                                                                                                                                             |
```
nickvergessen, line 95:
```suggestion
| `callStartTime`         | int     | v4    |         | Timestamp when the call was started (only available with `recording-v1` capability)                                                                                                                                                                                                                                                                                                               |
| `callRecording`         | int     | v4    |         | Type of call recording (see [Constants - Call recording status](constants.md#call-recording-status)) (only available with `recording-v1` capability)                                                                                                                                                                                                                                                                                                                                                                        |
```

lib/Controller/RoomController.php
```diff
@@ -399,6 +399,7 @@ protected function formatRoomV4(Room $room, ?Participant $currentParticipant, ?a
 			'hasPassword' => $room->hasPassword(),
 			'hasCall' => false,
 			'callStartTime' => 0,
+			'callRecording' => 0,
```
nickvergessen, line 402:
```suggestion
			'callRecording' => Room::RECORDING_NONE,
```

nickvergessen (changes requested):

marcoambrosini (approved):

nickvergessen (approved):

PR #8515, opened 2023-01-02
Hide expired messages when do not run expire job

vitormattos:
### 🚧 TODO

- [x] prevent display expired messages when don't run the job to do expire messages.

### 🏁 Checklist

### DIFF
--- a/lib/Controller/ChatController.php
+++ b/lib/Controller/ChatController.php
@@ -466,6 +466,7 @@ class ChatController extends AEnvironmentAwareController {
 		$this->preloadShares($comments);
 
 		$i = 0;
+		$now = $this->timeFactory->getDateTime('now', new \DateTimeZone('UTC'));
 		$messages = $commentIdToIndex = $parentIds = [];
 		foreach ($comments as $comment) {
 			$id = (int) $comment->getId();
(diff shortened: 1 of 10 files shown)


lib/Controller/ChatController.php
```diff
@@ -472,6 +472,11 @@ public function receiveMessages(int $lookIntoFuture,
 			$message = $this->messageParser->createMessage($this->room, $this->participant, $comment, $this->l);
 			$this->messageParser->parseMessage($message);
 
+			$expireDate = $message->getComment()->getExpireDate();
+			if ($expireDate instanceof \DateTime && $expireDate < $this->timeFactory->getDateTime()) {
```
nickvergessen, line 476: Calling `$this->timeFactory->getDateTime()` 200 times here. Want to move it outside of the loop?

Also should add it around line 520 where parents are rendered and look at other endpoints (search comes to mind)

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -2965,20 +2965,6 @@ public function waitForXSeconds($seconds): void {
 		sleep($seconds);
 	}
 
-	/**
-	 * @When apply message expiration job manually
-	 */
-	public function applyMessageExpirationJob(): void {
-		$currentUser = $this->currentUser;
-		$this->setCurrentUser('admin');
-		$this->sendRequest('GET', '/apps/spreedcheats/get_message_expiration_job');
```
nickvergessen, line 2974: then you can also remove that from the helper app.

nickvergessen (approved):

PR #8546, opened 2023-01-10
Allow to prepare virtual participants without immediately joining the call

SystemKeeper:
Currently Talkbuchet offers the method `addVirtualParticipant()` and `addVirtualParticipants(NUMBER)`. Those methods directly start a virtual participant, therefore joining the call sequentially. This is fine if you want to use Talkbuchet for testing with multiple participants in a call, but is not sufficient if you want to let an arbitrary number of participants join/leave at the same time (e.g. to test performance of layout calculations being done when multiple participants join/leave at once).

This PR extends `Talkbuchet.py` with the following methods:
`prepareVirtualParticipant()`
`prepareVirtualParticipants(NUMBER)`
`startVirtualParticipants()`
`startVirtualParticipantsParallel()`
`stopVirtualParticipants()`
`stopVirtualParticipantsParallel()`

while still keeping the previous functionality of `addVirtualParticipant()` / `addVirtualParticipants(NUMBER)`.

Basically the `prepare*` methods spin up the virtual participant, but does not join any call. You can prepare as many participants as you want (or your machine is able to handle) and then let them join "at once". Using `startVirtualParticipants()` will start each virtual participant in sequence, waiting for the previous one to finish its start. This is still faster than using `addVirtualParticipant*`, because every virtual participant is already up, but still not enough to simulate a parallel join/leave. For this you can use `startVirtualParticipantsParallel()`, which calls the start of each virtual participant in its own thread.

Example:
```
# Run Talkbuchet (using ARM image here)
./talkbuchet-run.sh --selenium-image "seleniarm/standalone-firefox:latest"

# Run in Talkbuchet
switchToVirtualParticipantMode()
setCredentials("test", "token")
setTarget("https://localhost")
setToken("roomToken")
setMedia(True, True)
prepareVirtualParticipants(10)
startVirtualParticipantsParallel()
```

### DIFF
--- a/docs/Talkbuchet-cli.py
+++ b/docs/Talkbuchet-cli.py
@@ -1296,6 +1296,12 @@ def switchToSiegeMode():
         if removeVirtualParticipants:
             removeVirtualParticipants()
 
+        del globals()['prepareVirtualParticipant']
+        del globals()['prepareVirtualParticipants']
+        del globals()['startVirtualParticipants']
+        del globals()['startVirtualParticipantsParallel']
+        del globals()['stopVirtualParticipants']
+        del globals()['stopVirtualParticipantsParallel']
         del globals()['addVirtualParticipant']
         del globals()['addVirtualParticipants']
         del globals()['removeVirtualParticipant']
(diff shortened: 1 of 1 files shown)


danxuliu (approved):
Tested and works :+1: 

Sorry for the delay and thanks! :-)

I have slightly adjusted the documentation. If you agree with the changes feel free to squash and merge :-)

PR #8550, opened 2023-01-12
Notify recording stored

vitormattos:
### 🚧 TODO

- [x] https://github.com/nextcloud/spreed/issues/8326 pending issues

### 🏁 Checklist

### DIFF
--- a/lib/Chat/SystemMessage/Listener.php
+++ b/lib/Chat/SystemMessage/Listener.php
@@ -382,6 +382,10 @@ class Listener implements IEventListener {
 		$listener = Server::get(self::class);
 		$manager = Server::get(Manager::class);
 
+		$request = Server::get(IRequest::class);
+		if ($request->getParam('_route') === 'ocs.spreed.Recording.shareToChat') {
+			return;
+		}
 		$room = $manager->getRoomByToken($share->getSharedWith());
 		$metaData = Server::get(IRequest::class)->getParam('talkMetaData') ?? '';
 		$metaData = json_decode($metaData, true);
--- a/lib/Controller/RecordingController.php
+++ b/lib/Controller/RecordingController.php
@@ -105,12 +105,30 @@ class RecordingController extends AEnvironmentAwareController {
 	 * @PublicPage
 	 * @RequireModeratorParticipant
 	 */
-	public function notificationDismiss(string $token, string $dateTime): DataResponse {
+	public function notificationDismiss(int $timestamp): DataResponse {
 		try {
 			$this->recordingService->notificationDismiss(
-				$token,
-				$this->participant->getAttendee()->getActorId(),
-				$dateTime
+				$this->getRoom(),
+				$this->participant,
+				$timestamp
+			);
+		} catch (InvalidArgumentException $e) {
+			return new DataResponse(['error' => $e->getMessage()], Http::STATUS_BAD_REQUEST);
+		}
+		return new DataResponse();
+	}
+
+	/**
+	 * @PublicPage
+	 * @RequireModeratorParticipant
+	 */
+	public function shareToChat(int $fileId, int $timestamp): DataResponse {
+		try {
+			$this->recordingService->shareToChat(
+				$this->getRoom(),
+				$this->participant,
+				$fileId,
+				$timestamp
 			);
 		} catch (InvalidArgumentException $e) {
 			return new DataResponse(['error' => $e->getMessage()], Http::STATUS_BAD_REQUEST);
--- a/lib/Controller/RecordingController.php
+++ b/lib/Controller/RecordingController.php
@@ -100,4 +100,21 @@ class RecordingController extends AEnvironmentAwareController {
 		}
 		return new DataResponse();
 	}
+
+	/**
+	 * @PublicPage
+	 * @RequireModeratorParticipant
+	 */
+	public function notificationDismiss(string $token, string $dateTime): DataResponse {
+		try {
+			$this->recordingService->notificationDismiss(
+				$token,
+				$this->participant->getAttendee()->getActorId(),
+				$dateTime
+			);
+		} catch (InvalidArgumentException $e) {
+			return new DataResponse(['error' => $e->getMessage()], Http::STATUS_BAD_REQUEST);
+		}
+		return new DataResponse();
+	}
 }
--- a/lib/Controller/RecordingController.php
+++ b/lib/Controller/RecordingController.php
@@ -102,7 +102,7 @@ class RecordingController extends AEnvironmentAwareController {
 	}
 
 	/**
-	 * @PublicPage
+	 * @NoAdminRequired
 	 * @RequireModeratorParticipant
 	 */
 	public function notificationDismiss(int $timestamp): DataResponse {
@@ -119,7 +119,7 @@ class RecordingController extends AEnvironmentAwareController {
 	}
 
 	/**
-	 * @PublicPage
+	 * @NoAdminRequired
 	 * @RequireModeratorParticipant
 	 */
 	public function shareToChat(int $fileId, int $timestamp): DataResponse {
--- a/lib/Notification/Notifier.php
+++ b/lib/Notification/Notifier.php
@@ -315,7 +315,7 @@ class Notifier implements INotifier {
 					'call' => [
 						'type' => 'call',
 						'id' => $room->getId(),
-						'name' => $room->getDisplayName((string) $participant->getAttendee()->getId()),
+						'name' => $room->getDisplayName((string) $participant->getAttendee()->getActorId()),
 						'call-type' => $this->getRoomType($room),
 					],
 				])
--- a/lib/Notification/Notifier.php
+++ b/lib/Notification/Notifier.php
@@ -320,7 +320,7 @@ class Notifier implements INotifier {
 				IAction::TYPE_POST
 			);
 		$dismissAction = $notification->createAction()
-			->setParsedLabel($l->t('Dismiss'))
+			->setParsedLabel($l->t('Dismiss notification'))
 			->setLink(
 				$this->urlGenerator->linkToRouteAbsolute(
 					'ocs.spreed.Recording.notificationDismiss',
--- a/lib/Notification/Notifier.php
+++ b/lib/Notification/Notifier.php
@@ -41,9 +41,7 @@ use OCA\Talk\Webinary;
 use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\Comments\ICommentsManager;
 use OCP\Comments\NotFoundException;
-use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\HintException;
-use OCP\IDBConnection;
 use OCP\IL10N;
 use OCP\IURLGenerator;
 use OCP\IUser;
@@ -61,7 +59,6 @@ use OCP\Share\IShare;
 
 class Notifier implements INotifier {
 	protected IFactory $lFactory;
-	private IDBConnection $db;
 	protected IURLGenerator $url;
 	protected Config $config;
 	protected IUserManager $userManager;
@@ -83,7 +80,6 @@ class Notifier implements INotifier {
 	protected array $participants = [];
 
 	public function __construct(IFactory $lFactory,
-								IDBConnection $db,
 								IURLGenerator $url,
 								Config $config,
 								IUserManager $userManager,
@@ -99,7 +95,6 @@ class Notifier implements INotifier {
 								Definitions $definitions,
 								AddressHandler $addressHandler) {
 		$this->lFactory = $lFactory;
-		$this->db = $db;
 		$this->url = $url;
 		$this->config = $config;
 		$this->userManager = $userManager;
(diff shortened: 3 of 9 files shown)


lib/Service/RecordingService.php
```diff
@@ -142,4 +147,44 @@ private function getRecordingFolder(string $owner, string $token): Folder {
...
+		$notificationLevel = $attendee->getNotificationLevel();
+		if ($notificationLevel === Participant::NOTIFY_NEVER) {
+			return;
+		}
+
+		$notification = $this->notificationManager->createNotification();
+
+		$notification
+			->setApp('spreed')
+			->setDateTime(new \DateTime())
```
nickvergessen, line 162: Use `ITimeFactory` for better unit testing

lib/Notification/Notifier.php
```diff
@@ -287,6 +293,36 @@ protected function shortenJsonEncodedMultibyteSave(string $subject, int $dataLen
 		return $temp;
 	}
 
+	private function parseStoredRecording(INotification $notification, Room $room, Participant $participant, IL10N $l): INOtification {
```
nickvergessen, line 296:
```suggestion
	private function parseStoredRecording(INotification $notification, Room $room, Participant $participant, IL10N $l): INotification {
```

lib/Notification/Notifier.php
```diff
@@ -287,6 +293,36 @@ protected function shortenJsonEncodedMultibyteSave(string $subject, int $dataLen
...
+						'apiVersion' => 'v1',
+						'token' => $room->getToken()
+					]
+				),
+				IAction::TYPE_POST
+			);
+
+		$notification
+			->setRichSubject(
+				$l->t('Record file of {call}'),
```
nickvergessen, line 313:
```suggestion
				$l->t('Recording for the call in {call} was uploaded.'),
```

lib/Notification/Notifier.php
```diff
@@ -287,6 +293,36 @@ protected function shortenJsonEncodedMultibyteSave(string $subject, int $dataLen
...
+				[
+					'call' => [
+						'type' => 'call',
+						'id' => $room->getId(),
+						'name' => $room->getDisplayName((string) $participant->getAttendee()->getId()),
+						'call-type' => $this->getRoomType($room),
+					],
+				])
+			->addParsedAction($shareAction);
```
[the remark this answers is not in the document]
vitormattos: As we talked in private chat, this will be solved in a follow up PR.

vitormattos:
@nickvergessen 

Couldn't the share to chat message cause confusion? For example: which chat?

image

lib/Notification/Notifier.php
```diff
@@ -287,6 +298,75 @@ protected function shortenJsonEncodedMultibyteSave(string $subject, int $dataLen
...
+					'ocs.spreed.Chat.shareObjectToChat',
+					[
+						'apiVersion' => 'v1',
+						'token' => $room->getToken()
+					]
+				),
+				IAction::TYPE_POST
+			);
+		$dismissAction = $notification->createAction()
+			->setParsedLabel($l->t('Dismiss'))
```
nickvergessen, line 321: Not sure about the wording. Dismiss might be confused that it dismisses the recording file which is not the case. Maybe 
```suggestion
			->setParsedLabel($l->t('Dismiss notification'))
```

lib/Notification/Notifier.php
```diff
@@ -287,6 +298,75 @@ protected function shortenJsonEncodedMultibyteSave(string $subject, int $dataLen
...
+			->andWhere($sql->expr()->eq('subject',
+				$sql->createNamedParameter($notification->getSubject())))
+			->andWhere($sql->expr()->eq('timestamp',
+				$sql->createNamedParameter($notification->getDateTime()->format('U')),
+				IQueryBuilder::PARAM_INT))
+			->setMaxResults(1);
+
+		$statement = $sql->executeQuery();
+		return (int) $statement->fetchOne();
+	}
```
nickvergessen, line 368: This is not okay, We can not make a query against the notifications tables.
But we can make an endpoint on the recording controller with a method to dismiss the notification and use normal notifications API there by specifying all the fields you did here + the userId and then call `OCP\Notifications\IManager::dismissNotification()`

nickvergessen (changes requested):

nickvergessen:
> Couldn't the share to chat message cause confusion? For example: which chat?

I think it is good enough and clear that it shares to the chat of that call.

lib/Notification/Notifier.php
vitormattos: Didn't exist a method to get the notification by other properties and also to dismiss notification.Didn't exist a method to get the notification by other properties and also to dismiss notification.

What do you suggest to get the notification_id or delete a notification using other properties?

lib/Notification/Notifier.php
```diff
@@ -287,6 +293,36 @@ protected function shortenJsonEncodedMultibyteSave(string $subject, int $dataLen
...
+		$shareAction = $notification->createAction()
+			->setParsedLabel($l->t('Share to chat'))
+			->setPrimary(true)
+			->setLink(
+				$this->urlGenerator->linkToRouteAbsolute(
+					'ocs.spreed.Chat.shareObjectToChat',
+					[
+						'apiVersion' => 'v1',
+						'token' => $room->getToken()
+					]
```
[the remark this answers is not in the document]
vitormattos: working now.

lib/Notification/Notifier.php
vitormattos: Is the text a bit big now:

image

lib/Notification/Notifier.php
```diff
@@ -69,6 +69,7 @@ class Notifier implements INotifier {
 	protected INotificationManager $notificationManager;
 	protected ICommentsManager $commentManager;
 	protected MessageParser $messageParser;
+	private IURLGenerator $urlGenerator;
```
nickvergessen, line 72:
```suggestion
	protected IURLGenerator $urlGenerator;
```

lib/Notification/Notifier.php
```diff
@@ -287,6 +293,62 @@ protected function shortenJsonEncodedMultibyteSave(string $subject, int $dataLen
...
+		Participant $participant,
+		IL10N $l
+	): INotification {
+		$parameters = $notification->getSubjectParameters();
+		$shareAction = $notification->createAction()
+			->setParsedLabel($l->t('Share to chat'))
+			->setPrimary(true)
+			->setLink(
+				$this->urlGenerator->linkToRouteAbsolute(
+					'ocs.spreed.Chat.shareObjectToChat',
```
nickvergessen, line 308:
```suggestion
				$this->urlGenerator->linkToOCSRouteAbsolute(
					'spreed.Chat.shareObjectToChat',
```

lib/Notification/Notifier.php
```diff
@@ -287,6 +293,62 @@ protected function shortenJsonEncodedMultibyteSave(string $subject, int $dataLen
...
+						'token' => $room->getToken()
+					]
+				),
+				IAction::TYPE_POST
+			);
+		$dismissAction = $notification->createAction()
+			->setParsedLabel($l->t('Dismiss notification'))
+			->setLink(
+				$this->urlGenerator->linkToRouteAbsolute(
+					'ocs.spreed.Recording.notificationDismiss',
```
nickvergessen, line 326:
```suggestion
				$this->urlGenerator->linkToOCSRouteAbsolute(
					'spreed.Recording.notificationDismiss',
```

lib/Notification/Notifier.php
nickvergessen: We could also have `/api/{apiVersion}/recording/{token}/notification` that does this. But both work I guess.

lib/Notification/Notifier.php
```diff
@@ -287,6 +293,62 @@ protected function shortenJsonEncodedMultibyteSave(string $subject, int $dataLen
...
+					[
+						'apiVersion' => 'v1',
+						'token' => $room->getToken(),
+						'dateTime' => $notification->getDateTime()->format('U'),
```
nickvergessen, line 330: Time sounds a bit error prone to me. Why not the file id?

lib/Service/RecordingService.php
```diff
@@ -142,4 +151,38 @@ private function getRecordingFolder(string $owner, string $token): Folder {
...
+			->setApp('spreed')
+			->setDateTime($this->timeFactory->getDateTime())
+			->setObject('chat', $room->getToken())
+			->setUser($attendee->getActorId())
+			->setSubject('record_file_stored', [
+				'objectType' => 'file',
+				'objectId' => $file->getId(),
+				'name' => $file->getName(),
+				'actorDisplayName' => $attendee->getDisplayName(),
+			]);
```
nickvergessen, line 174: It's quite clear that the object is a file on this subject?
With objectType and Id I meant the `$notification->setObject()` but maybe that's not a good idea.
So I would say let's remove `objectType`, `name` and `actorDisplayName` from the parameter list here as they are not needed.

```suggestion
			->setSubject('record_file_stored', [
				'objectId' => $file->getId(),
			]);
```

lib/Notification/Notifier.php
```diff
@@ -287,6 +293,62 @@ protected function shortenJsonEncodedMultibyteSave(string $subject, int $dataLen
...
+			->setLink(
+				$this->urlGenerator->linkToRouteAbsolute(
+					'ocs.spreed.Chat.shareObjectToChat',
+					[
+						'apiVersion' => 'v1',
+						'objectType' => 'file',
+						'objectId' => $notification->getObjectId(),
+						'metaData' => json_encode([
+							'name' => $parameters['name'],
+							'path' => $parameters['name'],
```
nickvergessen, line 315: Path and name are not the same. Please get the file from the IUserFolder and give the current path and name

lib/Notification/Notifier.php
```diff
@@ -287,6 +293,62 @@ protected function shortenJsonEncodedMultibyteSave(string $subject, int $dataLen
...
+		$shareAction = $notification->createAction()
+			->setParsedLabel($l->t('Share to chat'))
+			->setPrimary(true)
+			->setLink(
+				$this->urlGenerator->linkToRouteAbsolute(
+					'ocs.spreed.Chat.shareObjectToChat',
+					[
+						'apiVersion' => 'v1',
+						'objectType' => 'file',
+						'objectId' => $notification->getObjectId(),
```
nickvergessen, line 312:
```suggestion
						'objectId' => $parameters['objectId'],
```

Otherwise you add the room token instead of the file Id as a share

lib/Notification/Notifier.php
vitormattos: The fileId is stored inside a json.
We can remove the dateTime but, is a more specific attribute to prevent dismiss a wrong notification.

vitormattos:
The shared recording file:

image

lib/Chat/SystemMessage/Listener.php
```diff
@@ -382,6 +382,10 @@ public static function fixMimeTypeOfVoiceMessage($event): void {
 		$listener = Server::get(self::class);
 		$manager = Server::get(Manager::class);
 
+		$request = Server::get(IRequest::class);
+		if ($request->getParam('_route') === 'ocs.spreed.Recording.shareToChat') {
+			return;
+		}
```
vitormattos, line 388: I need do this to prevent run this method and duplicate the system message

nickvergessen: This sounds illegal 🙈

Please raise a follow up tech-debt issue to replace this with something better.
E.g. we could have a "state" variable in the RecordingService about this, that skips it and afterwards resets.

nickvergessen (changes requested):

lib/Controller/RecordingController.php
```diff
@@ -100,4 +100,39 @@ public function store(string $owner): DataResponse {
...
+	 */
+	public function shareToChat(int $fileId, int $timestamp): DataResponse {
+		try {
+			$this->recordingService->shareToChat(
+				$this->getRoom(),
+				$this->participant,
+				$fileId,
+				$timestamp
+			);
+		} catch (InvalidArgumentException $e) {
```
nickvergessen, line 133: Should log the exception for better debugging?

vitormattos: I think not because we already returned the error message on body of request. Do you prefer with log?

nickvergessen: I commented in the wrong place.

lib/Service/RecordingService.php
```diff
@@ -142,4 +156,88 @@ private function getRecordingFolder(string $owner, string $token): Folder {
...
+		try {
+			$this->chatManager->addSystemMessage(
+				$room,
+				$participant->getAttendee()->getActorType(),
+				$participant->getAttendee()->getActorId(),
+				$message,
+				$creationDateTime,
+				true
+			);
+		} catch (\Exception $e) {
```
nickvergessen, line 238: This is where we should log the exception.

nickvergessen (approved):

PR #8568, opened 2023-01-18
Combine mute button with volume level indicator

### LINKED ISSUE #2225: Have volume icon filled with current speaking level
From https://github.com/nextcloud/spreed/issues/2138#issuecomment-526524520

vcfdsfd

Antreesy:
Fix #2225

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Device check before call | -
Screenshot from 2023-01-23 17-29-09 | Screenshot from 2023-01-23 17-52-22 Screenshot from 2023-01-23 17-52-53
🌃 Dark mode | - 
Screenshot from 2023-01-23 17-29-49 | Screenshot from 2023-01-23 17-54-10 Screenshot from 2023-01-23 17-54-01
During a call | - 
Screenshot from 2023-01-23 17-30-13 |  Screenshot from 2023-01-23 17-55-03 Screenshot from 2023-01-23 17-55-12

### 🚧 TODO

- [ ] Cross-browser testing (working on Google)  
- [x] Design review (color of primary icon, overlay icon)
- [ ] Code review (probably need to move component from root folder src/components)

### 🏁 Checklist


nickvergessen:
For better accessibility we could make the overlap green like the bar used to be?

nimishavijay:
Looks very nice! :) I would be concerned about accessibility as well, we could use the variable `--color-success` like @nickvergessen mentioned :)

nickvergessen:
So it's quite good when enabled, but the "grey" vs. "black" when the mic is disabled I find to hidden.
Could we maybe color that differently?

src/components/VolumeIndicator/VolumeIndicator.vue
```diff
@@ -20,27 +20,53 @@
...
-		<span ref="volumeIndicator"
-			class="volume-indicator"
-			:class="{'volume-indicator--disabled': disabled}"
-			:style="{ 'height': currentVolumeIndicatorHeight + 'px' }" />
-	</div>
+	<span class="volume-indicator"
+		:style="{ height: size + 'px', width: size + 'px' }">
+		<span class="volume-indicator-primary"
+			:style="{
+				height: size - 1 - currentVolumeIndicatorHeight + 'px',
```
nickvergessen, line 27: I would go with 2 px border
```suggestion
				height: size - 2 - currentVolumeIndicatorHeight + 'px',
```

src/components/MediaDevicesPreview.vue
```diff
@@ -114,32 +111,6 @@ export default {
...
-			const maximumVolumeIndicatorHeight = this.$refs.volumeIndicator.parentElement.clientHeight - (parseInt(volumeIndicatorStyle.bottom, 10) * 2)
-
-			return maximumVolumeIndicatorHeight * currentVolumeProportion
-		},
-
-	},
-
 	mounted() {
 		this.mounted = true
 	},
```
ShGKme, line 116: `this.mounted` property of the component is unused now. Maybe remove it from `data()` as well as this lifecycle hook? 

(The same was done in `LocalMediaControls.vue`)

ShGKme: There is also no need to have `this.currentVolume` and `this.volumeThreshold` in component's `data` as they are already defined in the `devices` mixin. Maybe remove them too?

ShGKme (commented):
The update good and nice looks for me 🎙️

One small thing - one property became unused now in `MediaDevicesPreview` and two other properties duplicated mixin's properties.

ShGKme:
Not really related to this update, but on testing, I found that the volume indicator usually has a low load. That is because 100% volume means "loudest sound" and comes only on overload. So the indicator is usually on low height (at it is supposed to be).

It is the same as how the previous indicator worked. But now, with the Microphone Icon, it is more notable for me because the icon is thin on the bottom and wide on the top.

Example of a typical call:

mic

What about manually increasing the volume on the indicator? For example, I tried to add 25%:

min-2

It will lie on loud sounds and overloads. In such case, we may add red color to indicate an overload:

mic-loud

And personally for me, it always looked too abrupt. I tried to add a `height 0.4s leaner` transition, and it looks more smooth now.

mic-smooth

What do you think?

src/components/DeviceChecker/DeviceChecker.vue
```diff
@@ -71,7 +71,7 @@
 							:audio-enabled="audioOn"
 							:current-volume="currentVolume"
 							:volume-threshold="volumeThreshold"
-							:overlay-color="'#999999'" />
+							overlay-muted-color="#888888" />
```
Antreesy, line 74: @nickvergessen please check it:
this color overwrites default color in both light and dark themes, but looks clearly visible for me

src/components/VolumeIndicator/VolumeIndicator.vue
```diff
@@ -102,45 +105,56 @@ export default {
...
-		overlayMutedColor() {
-			return !this.audioEnabled ? this.overlayColor : undefined
+	methods: {
+		/**
+		 * Compute volume proportion based on currentVolume [-100, 0]
+		 * and volumeThreshold. Returns number between [0, 1]
+		 *
+		 * @param {number} overload to visually increase volume level
+		 */
+		computeVolumeLevel(overload = 0) {
```
Antreesy, line 119: @ShGKme please check it:
added method to compute proportion, so we can adjust range with `overload` param. Do you see any opportunities to optimize this code?

src/components/VolumeIndicator/VolumeIndicator.vue
```diff
@@ -102,45 +105,56 @@ export default {
...
-	top: 0;
 	left: 50%;
 	transform: translateX(-50%);
 
+	width: 100%;
+	height: 100%;
+
 	overflow: hidden;
+
+	transition: height 0.2s linear;
```
Antreesy, line 147: added `transition` to smoothly change the indicator

ShGKme: Currently, the indicator is made by two clipped icons on the top and the button with heights:
- `currentVolumeIndicatorHeight`
- `size - 2 - currentVolumeIndicatorHeight`

This `-2` emulates a 2px gap like a border. But with transition, sometimes it has an incorrect size. Sometimes the gap is large than 2px, and sometimes, there is no gap.

Instead of making this gap by the size difference, maybe add an actual border?

```css
.volume-indicator-overlay {
  border-top: 2px solid var(--color-main-background);
}
```

src/components/VolumeIndicator/VolumeIndicator.vue
```diff
@@ -102,45 +105,56 @@ export default {
...
+		 */
+		computeVolumeLevel(overload = 0) {
+			const computedLevel
+				= (this.volumeThreshold - this.currentVolume)
+				/ (this.volumeThreshold + overload)
+
+			if (computedLevel < 0) return 0
+			else if (computedLevel > 1) return 1
+			else return computedLevel
 		},
```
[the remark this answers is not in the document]
ShGKme: And I'd make the default value -25 (tested on some calls with receiver volume as input).

Or it looks filled enough on your tests with 20?

src/components/VolumeIndicator/VolumeIndicator.vue
ShGKme: The formula seems to be correct 👍

I added some comments in review.

ShGKme:
The indicator now is always green. What about adding a red on overload values (if `currentVolume > overload`)?

In the previous volume meter, overload was indicated by the gradient on the background. It was redder on the top. But adding a gradient to SVG will take a lot of work.

src/components/VolumeIndicator/VolumeIndicator.vue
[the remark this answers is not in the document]
Antreesy: `-20` is the max value i could physically achieve with loud music directly in front of the microphone, so i go with it.
Also changing the color from green to red hits too hard on the eyes, IMHO, so I want to reduce the chance of this event occurring.

Also agree with the negative value and overloadLimit prop, seems more suitable for reusability

src/components/VolumeIndicator/VolumeIndicator.vue
Antreesy: The component implements in three places:

- MediaDevicesPreview - bg color is `var(--color-main-background)`
- DeviceChecker - bg color is `var(--color-main-background)`
- LocalMediaControls - bg color is image and `rgba(34, 34, 34, 0.8)` over it. So, considering that this is the mostly visible place of this component, actual border won't fit here

ShGKme: Ho, yes, you are right. A CSS border doesn't work in `LocalMediaControls`...

But the current solution sometimes looks buggy...

| Normal | No gap | Large gap |
|---|---|---|
| image | image | image |

GIF:

mic-bug

ShGKme: *Thinks... 🤔 *

ShGKme:
I finally understood why the Volume Indicator feels less filled than it should be to me.
The problem is that in `vue-material-icons` the icon is not located from top to bottom.

image

Full-size height is not equal to and greater than the actual icon height. And on low volume, the indicator height is lower than the icon's bottom.

In SVG position of the icon is not very "pixel perfect," with very float coordinates in the path. But it's close to these values:

- Top: ~1/12 size
- Bottom: ~1/8 size

So I tried these values in computing, and it looks more natural to me:

- **volume-indicator-primary**: `size - size / 8 - 2 - currentVolumeIndicatorHeight`
- **volume-indicator-overlay**: `size / 8 + currentVolumeIndicatorHeight`
- **currentVolumeIndicatorHeight**: 
  ```js
  const offsetBottom = this.size / 8
  const offsetTop = this.size / 12
  const iconHeight = this.size - offsetBottom - offsetTop;
  return iconHeight * this.computeVolumeLevel()
  ```

src/components/VolumeIndicator/VolumeIndicator.vue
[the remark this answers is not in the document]
ShGKme: If -20 is the max value you could achieve, doesn't it mean that the sound is usually less on the calls? 

I tested on calls using other participants' output as a microphone input. And with -20 overload even with the height fix from the last comment, it is still not high to me...

But that might be my testing issue. How does the indicator look on your tests?

ShGKme (approved):
Looks great on my tests 🔥

nickvergessen (approved):
For me it's mostly still not visible that I'm talking, but let's see in real world whether people can actually hear me when it does not show.

nimishavijay:
> What about making it red when the currentVolume is larger than overload?

That sounds good to me! In the screenshot it looks like the red is flashing very momentarily, which can be a bit distracting, looks a bit like a glitch. Maybe it could turn red if the volume is too high for more than 1 second continuously? :)

PR #8578, opened 2023-01-20
perf(autoloader): Force own autoloader

ChristophWurst:
And don't register any generic PSR-4 nor PSR-0 loader in server.

https://github.com/nextcloud/server/blob/473c546b5c562e19ba8f9d3aef9cd64d2458cda7/lib/private/legacy/OC_App.php#L284-L290

### Blackfire comparison graph on IO wait
Bildschirmfoto vom 2023-01-20 13-51-25

### Blackfire recommendations before
Bildschirmfoto vom 2023-01-20 13-53-14

### Blackfire recommendations after
Bildschirmfoto vom 2023-01-20 13-50-49

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
:sloth:  | :snail: 

### 🏁 Checklist

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -21,9 +21,14 @@ local Pipeline(test_set, database, services) = {
 			},
 			commands: [
 				"bash tests/drone-run-integration-tests.sh || exit 0",
+				"composer --version",
+				"composer self-update --2",
 				"wget https://raw.githubusercontent.com/nextcloud/travis_ci/master/before_install.sh",
 				"bash ./before_install.sh $APP_NAME $CORE_BRANCH $DATABASEHOST",
 				"cd ../server",
+				"cd apps/$APP_NAME",
+				"composer install --no-dev",
+				"cd ../..",
 				"./occ app:enable $APP_NAME",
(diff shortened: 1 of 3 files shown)


ChristophWurst:
> metrics.composer.autoload.find_file.count 117 <= 50

Server and this app have authorative autoloaders. But I think the generic autoloader in server tries to load this app's files before the shipped autoloader gets a change.

nickvergessen:
Thanks, I was always tempted to add one, but I disliked the "ah forgot to run `bash build/autoload...`" from server. I didn't know it's possible to do it for 3rdparty but not the own classes :)

nickvergessen:
Seems CI doesn't like it
grafik

ChristophWurst:
that means `composer i` didn't happen

nickvergessen:
> that means composer i didn't happen

Right, it was run only after the app was enabled.
Fixed now

nickvergessen (approved):

ChristophWurst:
Tip: https://github.com/nextcloud/mail/pull/8020 to relax the autoloader on dev setups and make it find new files too

PR #8588, opened 2023-01-23
Don't break the maps app with talks files sidebar

nickvergessen:
Fix https://github.com/nextcloud/maps/issues/947

Unsure why this happens at all, but for now it might be easier to just disable it.
This however means Talk chats of files are not visible in the files sidebar of Photos you watch on the Maps app, but that is better than Maps not rendering at all.

We could create a follow up to have a look where it breaks, unless @danxuliu already has a clue.

### 🏁 Checklist

### DIFF
--- a/lib/Files/TemplateLoader.php
+++ b/lib/Files/TemplateLoader.php
@@ -97,7 +97,9 @@ class TemplateLoader implements IEventListener {
 
 		Util::addStyle(Application::APP_ID, 'At');
 		Util::addStyle(Application::APP_ID, 'icons');
-		Util::addScript(Application::APP_ID, 'talk-files-sidebar');
+		if (strpos(\OC::$server->getRequest()->getPathInfo(), '/apps/maps') !== 0) {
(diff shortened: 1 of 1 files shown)


tacruc (approved):
Tested Seems to work, Should we worry about the failing tests?

nickvergessen:
Yes we should. seems that while copying the patch over to master I cut off the opening `{`

nickvergessen:
Raised https://github.com/nextcloud/spreed/issues/8589 for proper fix

meichthys:
I'm not sure if these changes have been published yet, but I'm still seeing this issue with the following:
Talk v 15.0.3
Maps v0.2.5
Nextcloud v25.0.3 RC2

nickvergessen:
#8590 says "Next Patch" as milestone, so not released yet. Takes roughly 2 more weeks.

nickvergessen:
But until then you can patch it locally

meichthys:
Thanks!

PR #8605, opened 2023-01-26
Fix tooltip appearance for long names in ConversationList

### LINKED ISSUE #7254: When adding a user with a very long name to a conversation group, all information is not displayed
### How to use GitHub

---

When trying to add a user to a new conversation group, if the user's name is too long, it is not displayed completely

Insert a tooltip in the component to display the full information

**Additional context**
image

Antreesy:

Fix #7254 

Due to using the NcListItem component with title locked for modifications, we don't have opportunities to add `v-tooltip` here.
So, HTML-attribute `title` was applied for long names to show tooltip via native browser tools, when name is ellipsed.

*!IMPORTANT*
Current realisation works only if name is ellipsed initially (action button appears on hover and crops name even more) 

### 🖼️ Screenshots
image

### 🚧 TODO

- [ ] Cross-browser testing (working on Google)  
- [ ] Design review
- [ ] Code review

### 🏁 Checklist

### DIFF
--- a/src/components/LeftSidebar/ConversationsList/Conversation.vue
+++ b/src/components/LeftSidebar/ConversationsList/Conversation.vue
@@ -20,7 +20,8 @@
 -->
 
 <template>
-	<NcListItem :title="item.displayName"
+	<NcListItem ref="listItem"
+		:title="item.displayName"
 		:class="{'unread-mention-conversation': item.unreadMention}"
 		:anchor-id="`conversation_${item.token}`"
 		:actions-aria-label="t('spreed', 'Conversation actions')"
@@ -303,11 +304,20 @@ export default {
 		},
 	},
 
-	mounted() {
-		const titleSpan = document.getElementById(`conversation_${this.item.token}`)?.querySelector('.line-one__title')
-		if (!titleSpan) return
+	// TODO: move the implementation to @nextcloud-vue/NcListItem
+	watch: {
+		'item.displayName': {
+			immediate: true,
+			handler(value) {
+				this.$nextTick().then(() => {
+					const titleSpan = this.$refs.listItem?.$el?.querySelector('.line-one__title')
 
-		if (titleSpan.offsetWidth < titleSpan.scrollWidth) titleSpan.setAttribute('title', this.item.displayName)
+					if (titleSpan && titleSpan.offsetWidth < titleSpan.scrollWidth) {
+						titleSpan.setAttribute('title', value)
+					}
+				})
+			},
+		},
 	},
 
 	methods: {
--- a/src/components/LeftSidebar/ConversationsList/Conversation.vue
+++ b/src/components/LeftSidebar/ConversationsList/Conversation.vue
@@ -302,6 +302,13 @@ export default {
 			}
 		},
 	},
+
+	mounted() {
+		const titleSpan = document.getElementById(`conversation_${this.item.token}`).querySelector('.line-one__title')
+
+		if (titleSpan.offsetWidth < titleSpan.scrollWidth) titleSpan.setAttribute('title', this.item.displayName)
+	},
+
 	methods: {
 		async copyLinkToConversation() {
 			try {
--- a/src/components/LeftSidebar/ConversationsList/Conversation.vue
+++ b/src/components/LeftSidebar/ConversationsList/Conversation.vue
@@ -304,7 +304,8 @@ export default {
 	},
 
 	mounted() {
-		const titleSpan = document.getElementById(`conversation_${this.item.token}`).querySelector('.line-one__title')
+		const titleSpan = document.getElementById(`conversation_${this.item.token}`)?.querySelector('.line-one__title')
+		if (!titleSpan) return
 
 		if (titleSpan.offsetWidth < titleSpan.scrollWidth) titleSpan.setAttribute('title', this.item.displayName)
 	},
--- a/package-lock.json
+++ b/package-lock.json
@@ -72,6 +72,7 @@
 				"@vue/vue2-jest": "^29.2.1",
 				"babel-loader-exclude-node-modules-except": "^1.2.1",
 				"babel-plugin-add-module-exports": "^1.0.4",
+				"flush-promises": "^1.0.2",
 				"jest": "^29.2.2",
 				"jest-environment-jsdom": "^29.3.1",
 				"jest-localstorage-mock": "^2.4.26",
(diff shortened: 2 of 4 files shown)


src/components/LeftSidebar/ConversationsList/Conversation.vue
```diff
@@ -302,6 +302,14 @@ export default {
 			}
 		},
 	},
+
+	mounted() {
+		const titleSpan = document.getElementById(`conversation_${this.item.token}`)?.querySelector('.line-one__title')
+		if (!titleSpan) return
```
nickvergessen, line 308:
```suggestion
		if (!titleSpan) {
			return
		}
```

src/components/LeftSidebar/ConversationsList/Conversation.vue
```diff
@@ -302,6 +302,14 @@ export default {
...
+
+		if (titleSpan.offsetWidth < titleSpan.scrollWidth) titleSpan.setAttribute('title', this.item.displayName)
```
nickvergessen, line 310:
```suggestion
		if (titleSpan.offsetWidth < titleSpan.scrollWidth) {
			titleSpan.setAttribute('title', this.item.displayName)
		}
```

nickvergessen (commented):
We don't use inline if's in Nextcloud :)

src/components/LeftSidebar/ConversationsList/Conversation.vue
Antreesy: Got it, will fix tommorow. But if it's accepted by team, maybe it should be regulated by linter? 
https://eslint.org/docs/latest/rules/curly

src/components/LeftSidebar/ConversationsList/Conversation.vue
ShGKme, line 310: In Vue we should not directly mutate DOM elements that are fully controlled by Vue's Virtual DOM rendering. DOM should be defined with the component's template. Direct mutating is a very not Vue way. Vue can recreate this element on rendering, then the title will be lost.

But `NcListItem` has no features for setting a native title attribute...

I'd add a `// TODO: ` comment here to fix it in the future when/if NcListItem will support setting the native title attribute or displaying it for long titles.

Related issue: https://github.com/nextcloud/nextcloud-vue/issues/3687

src/components/LeftSidebar/ConversationsList/Conversation.vue
```diff
@@ -302,6 +303,23 @@ export default {
 			}
 		},
 	},
+
+	// TODO: move the implementation to @nextcloud-vue/NcListItem
```
Antreesy, line 307: Waiting for discussion of nextcloud/nextcloud-vue#3687, if the same approach could be implemented there

src/components/LeftSidebar/ConversationsList/Conversation.vue
```diff
@@ -302,6 +303,23 @@ export default {
...
+	watch: {
+		'item.displayName': {
+			immediate: true,
+			handler(value) {
+				this.$nextTick().then(() => {
+					const titleSpan = this.$refs.listItem?.$el?.querySelector('.line-one__title')
+
```
ShGKme, line 314: If `$refs.listItem` is always defined, then all optional chainings `?.` are redundant.

But if `$refs.listItem` maybe `undefined`, then it will fail anyway with a call of undefined. In such a case, it should have optional chaining on the function call `querySelector?.('.line-one__title')`.

`<NcListItem>` is always rendered: there is no `v-if` on it or anything. IMO, we may drop checking with optional chaining here.

ShGKme (approved):
Everything is good for me now. Only one small comment about optional chaining.

marcoambrosini (approved):

Antreesy:
`this.$nextTick()` prevent `'errors with notification when a new moderator is required before leaving'` test from src/components/LeftSidebar/ConversationsList/Conversation.spec.js to pass.

ShGKme:
> `this.$nextTick()` prevent `'errors with notification when a new moderator is required before leaving'` test from src/components/LeftSidebar/ConversationsList/Conversation.spec.js to pass.
> 
> Added `flush-promises` for Jest to work with multiple promises. Dependency could be removed after migration to Vue 3 and @vue/test-utils v2.0.0 (also added TODO for this).
> 
> Thanks to @ShGKme for the suggestion

For the record.

**What happens in the test:**
1. Click on the "Leave conversation" button, which has some async handler
2. `await trigger('click')` (= `await nextTick()`)
3. Check that the result of the handler is correct

**What was not correct:**

A promise of the async handler is not related to the nextTick's promise. Waiting for async in Vue reactivity doesn't guarantee some async method (handler) to be resolved.

**What is correct:**

Wait for the handler's promise to be resolved. It's not easy to do. But waiting for all promises to be resolved works fine, and that is a common practice. `flushPromises` does it. 

**Why it worked previously and failed after the update in this PR:**

The async handler was kind of the only async process in the component. Waiting for the `nextTick`'s promise was enough because its handler stayed later in the microtasks queue. 

1. Mount component
2. Click on the button - the async handler is executed
3. Mocked executor in async hander finished immediately - add the promise's resolve handler to the microtasks queue
4. `await nextTick` - add Vue scheduler's promise's resolve handler to the microtasks queue
5. Go to the microtasks queue. 3 is done
6. Go to the microtasks queue. 4 is done
7. ✅ Check chat 4 is done 

Calling `nextTick` early on the component's creation breaks it.
1. Mount component
2. `await nextTick` - add Vue scheduler's promise's resolve handler to the microtasks queue
3. Click on the button - the handler is executed
4. Mocked executor in async hander finished immediately - add the promise's resolve handler to the microtasks queue
5. Go to the microtasks queue. 2 - done
6. ❌ Check chat 4 is done 
7. Go to the microtasks queue. 5 - done. Too late...

ShGKme (approved):

src/components/LeftSidebar/ConversationsList/Conversation.vue
```diff
@@ -302,6 +303,23 @@ export default {
...
+					if (titleSpan && titleSpan.offsetWidth < titleSpan.scrollWidth) {
+						titleSpan.setAttribute('title', value)
```
raimund-schluessler, line 316: All this just to set the title attribute? It would be a minor change in the nextcloud-vue lib to add this there without the need for `watch` and `querySelector`.

ShGKme: As I know, it is not possible to detect if a text fit container or not without putting it in the container and looking at the result in DOM. watch + DOM access could be missed only if a title is always shown independently from title length.

But a change in nextcloud-vue may help to avoid manual mutating of Vue-rendered DOM.

PR #8656, opened 2023-01-31
fix(conversations): Notify participants about conversation changes so…

nickvergessen:
### 🚧 TODO

- [x] Trigger refresh of conversation list when recording is started/stopped
- [x] Trigger refresh when breakout rooms are setup/removed
- [x] Trigger refresh when breakout rooms are started/stopped

### 🏁 Checklist


danxuliu (commented):
This would make the external signaling server to send an event with type `update` and target `roomlist`. That should be enough to make the clients fetch again the data of the room :+1: 

However, the changed properties could also be included in the event by adding them to [`getPropertiesForSignaling`](https://github.com/nextcloud/spreed/blob/cf72ae74df6ef1ae5e424a404546dea783e49e53/lib/Room.php#L538), which could be used by the clients [to immediately update the data](https://github.com/nextcloud/spreed/pull/8477/commits/afde0e0d3a0b21eff34a884901c86047828f943f). Nevertheless, fine by me to merge already and do that in a follow up with more testing.

nickvergessen:
For recording that might work, but for breakout rooms it affects multiple rooms and can even be an invite/disinvite at the same time, so I thought we go with the better save then sorry approach.

PR #8672, opened 2023-02-02
Add PHP 8.2 support

nickvergessen:
### 🏁 Checklist

### DIFF
--- a/tests/php/Service/BreakoutRoomServiceTest.php
+++ b/tests/php/Service/BreakoutRoomServiceTest.php
@@ -51,6 +51,8 @@ class BreakoutRoomServiceTest extends TestCase {
 	private $participantService;
 	/** @var ChatManager|MockObject */
 	private $chatManager;
+	/** @var INotificationManager|MockObject */
+	private $notificationManager;
 	/** @var IEventDispatcher|MockObject */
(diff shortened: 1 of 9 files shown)


tests/php/Service/BreakoutRoomServiceTest.php
```diff
@@ -51,6 +51,8 @@ class BreakoutRoomServiceTest extends TestCase {
 	private $participantService;
 	/** @var ChatManager|MockObject */
 	private $chatManager;
+	/** @var INotificationManager|MockObject */
+	private $notificationManager;
```
vitormattos, line 55: **Optional**:
Union types works native since PHP 8.0 (https://3v4l.org/ArLAg):
```suggestion
	private INotificationManager|MockObject $notificationManager;
```
Maybe we can follow changing the code that we touch removing the union type from annotation and use the native way.

vitormattos (approved):

tests/php/Service/BreakoutRoomServiceTest.php
nickvergessen: Moved to https://github.com/nextcloud/spreed/issues/8677

PR #8676, opened 2023-02-03
Fix rounded corners on speaker's video

### LINKED ISSUE #8156: Video of other person in 1on1 call is not rounded
### How to use GitHub

---

## Steps to reproduce
1. Start a call
2. Look at the video of the other person

### Expected behaviour
Nice border-radius, just like for your own video.

### Actual behaviour
Sharp edges, ow!
image

## Talk app

**Talk app version:** Latest on cloud.nc.com

Antreesy:

Fix #8156 

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
One-to-One call | 
Screenshot from 2023-02-03 10-52-49 | Screenshot from 2023-02-03 10-48-16
Group call | 
Screenshot from 2023-02-03 10-51-57 |  Screenshot from 2023-02-03 10-49-37

- [X] Fulscreen view is rounded as well
- [X] Screencast isn't affected
- [X] Video in portrait mode is not cutted

### 🚧 TODO

- [ ] Code-review
- [ ] Visual check

### 🏁 Checklist

### DIFF
--- a/src/components/CallView/shared/VideoVue.vue
+++ b/src/components/CallView/shared/VideoVue.vue
@@ -21,6 +21,7 @@
 <template>
 	<div v-show="!placeholderForPromoted || sharedData.promoted"
 		:id="(placeholderForPromoted ? 'placeholder-' : '') + 'container_' + peerId + '_video_incoming'"
+		ref="videoContainer"
 		class="videoContainer"
 		:class="containerClass"
 		@mouseover="showShadow"
@@ -29,11 +30,13 @@
 		<transition name="fade">
 			<div v-show="showVideo"
 				:class="videoWrapperClass"
-				class="videoWrapper">
+				class="videoWrapper"
+				:style="videoWrapperStyle">
 				<video ref="video"
 					:disablePictureInPicture="!isBig"
 					:class="videoClass"
-					class="video" />
+					class="video"
+					@playing="updateVideoAspectRatio" />
 			</div>
 		</transition>
 		<transition name="fade">
@@ -164,7 +167,25 @@ export default {
 		},
 	},
 
+	data() {
+		return {
+			videoAspectRatio: null,
+			containerAspectRatio: null,
+			resizeObserver: null,
+		}
+	},
+
 	computed: {
+
+		videoWrapperStyle() {
+			if (!this.containerAspectRatio || !this.videoAspectRatio) {
+				return
+			}
+
+			return (this.containerAspectRatio > this.videoAspectRatio)
+				? `width: ${this.$refs.videoContainer.clientHeight * this.videoAspectRatio}px`
+				: `height: ${this.$refs.videoContainer.clientWidth / this.videoAspectRatio}px`
+		},
 
 		isSelectable() {
 			if (this.isStripe) {
@@ -475,6 +496,17 @@ export default {
 
 		// Set initial state
 		this._setStream(this.model.attributes.stream)
+
+		if (this.isBig) {
+			this.resizeObserver = new ResizeObserver(this.updateContainerAspectRatio)
+			this.resizeObserver.observe(this.$refs.videoContainer)
+		}
+	},
+
+	beforeDestroy() {
+		if (this.resizeObserver) {
+			this.resizeObserver.disconnect()
+		}
 	},
 
 	destroyed() {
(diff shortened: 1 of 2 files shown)


ShGKme (commented):
It also affects the layout of a portrait. I'd prefer not to have the video cut.

🏚️ Before | 🏡 After
---|---
image | image

nickvergessen:
> I'd prefer not to have the video cut.

Yeah that is actually a requirement from product management

Antreesy:
Ok, looking for a better solution

ShGKme (commented):
Adding ResizeObserver to round a border seems too overhead to me...

I haven't tried it on the flexbox, don't know if this is possible. But it is definitely possible with the gridbox layout.

Antreesy:
> But it is definitely possible with the gridbox layout.

I'm afraid that this is going to be the same trick for the `grid` as for the `flex` (see below), and these are issues:
- both require additional wrapper for `<video>` element;
- working only for portait mode, where container's `width` param is usually less then `height`;
- padding is calculated from parent width, so the total size is probably going to be larger then we need.

```
.videoAdditionalWrapper {
  width: 100%;
  height: 0;
  padding-top: 56.25%; // reverted aspect-ratio, can be obtained from eventListener
  position: relative;
}
.video {
  width: 100%;
  height: 100%;
  position: absolute;
  top: 0;
  left: 0;
}
```
So, if there is a more simple solution, i'll be glad to hear it.

For now, we can restrict ResizeObserver and EventListener to be called only for speaker's frame (which is only one at a time) to reduce memory consumption. Unfortunately, i couldn't find a better way to get `aspect-ratio` of video stream and orientation of container.

ShGKme:
We could get `aspect-ratio` from 

```js
model.attributes.stream.getVideoTracks()[0].getSettings().aspectRatio
```

But yep. It doesn't work with gridbox either... 

I cannot make it at the same time stretched + centralized + kept aspect ratio... (´。＿。｀)

Not very beautiful solution: use the fact that header's and paddings' heights are known and fixed.

ShGKme:
There is a working solution using [aspect-ratio](https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio). It has been supported in modern browsers since ~Jun..Sep 2021

src/components/CallView/shared/VideoVue.vue
```diff
@@ -475,6 +490,14 @@ export default {
...
 		this._setStream(this.model.attributes.stream)
+
+		if (this.isBig) {
+			this.$refs.video.addEventListener('playing', this.getVideoAspectRatio)
+		}
+	},
+
+	beforeDestroy() {
+		this.$refs.video.removeEventListener('playing', this.getVideoAspectRatio)
 	},
```
ShGKme, line 501: In Vue component's elements events are usually set with `v-on` directive in the template. Then all component's DOM events would be defined in the template and we wouldn't need to remove listeners manually.

```html
<video @playing="updateVideoAspectRatio"></video>
```

By the way, do we need to handle this event to update the aspect ratio? I'm not sure, but it seems to me that anytime the video is changed the stream would be also changed. Then the `model.attributes.stream` watcher with `_setStream` handler will be triggered. So we may calculate the aspect ratio immediately on the stream receiving from `model.attributes.stream.getVideoTracks()[0].getSettings().aspectRatio` while attaching the media stream before mounting to the DOM.

Antreesy: Unfortunately, this method doesn't return aspectRatio neither after setting stream, nor after `nextTick` call (see below):

image


these params appear in the settings only when video is on, which is - the best i could find - right after `playing` event
```
{
    "aspectRatio": 1.3333333333333333,
    "frameRate": 30,
    "height": 360,
    "width": 480
}
```

ShGKme (approved):
Looks good to me, tested in Chromium.
But I'm not sure about Safari, as it is kind of a new feature in CSS...

https://caniuse.com/mdn-css_properties_aspect-ratio

I want to test it on Mac later today.

ShGKme (changes requested):
It doesn't work on Safari, the video is always stretch like `width: 100%`

(┬┬﹏┬┬)

image

image

Ivansss:
Tested and works in Safari 👍

ShGKme (approved):
It's sad to realize that in 2023 there is still no solution for this problem in CSS without calculating the size manually in JS ಥ_ಥ

Thanks for all this work and attempts.

PR #8721, opened 2023-02-08
Get breakout rooms participants

marcoambrosini:
Screenshot 2023-02-08 at 16 45 29

### DIFF
--- a/src/components/RightSidebar/BreakoutRooms/BreakoutRoomsTab.vue
+++ b/src/components/RightSidebar/BreakoutRooms/BreakoutRoomsTab.vue
@@ -145,6 +145,11 @@ export default {
 			type: Object,
 			required: true,
 		},
+
+		isActive: {
+			type: Boolean,
+			required: true,
+		},
 	},
 
 	data() {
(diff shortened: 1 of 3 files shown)


src/components/RightSidebar/BreakoutRooms/BreakoutRoomsTab.vue
```diff
@@ -186,13 +192,35 @@ export default {
 		},
 	},
 
-	mounted() {
+	watch: {
+		isActive(newValue) {
```
nickvergessen, line 196: So everytime you go to this tab it immediately starts a new sync of the participants and starts the 30s interval.
Should be fine enough to do it on mount only?

src/services/breakoutRoomsService.js
```diff
@@ -95,11 +94,25 @@ const broadcastMessageToBreakoutRooms = async function(message, token) {
...
+ *
+ * @param token the conversation token
+ * @param message The message to be posted
+ * @return {Promise<AxiosResponse<any>>} The array of conversations
+ */
+const getBreakoutRoomsParticipants = async function(token) {
+	return await axios.get(generateOcsUrl('/apps/spreed/api/v4/room/{token}/breakout-rooms/participants', {
+		token,
+	})
+	)
```
nickvergessen, line 107:
```suggestion
	}))
```

src/store/breakoutRoomsStore.js
```diff
@@ -146,6 +150,22 @@ const actions = {
...
+			response.data.ocs.data.forEach(participant => {
+				context.dispatch('addParticipant', {
+					token: participant.roomToken,
+					participant,
+				})
+			})
+
+		} catch (error) {
+			console.error(error)
+			showError(t('spreed', 'An error occurred while sending a message to the breakout rooms'))
```
nickvergessen, line 166: wrong error message

src/components/RightSidebar/BreakoutRooms/BreakoutRoomsTab.vue
marcoambrosini: It's needed, but we're missing the part where we stop the interval if you navigate away from the tab, fixing this now

marcoambrosini: Actually, it's in mounted that it's not needed

src/store/breakoutRoomsStore.js
marcoambrosini: Removed as it's useless for the users

PR #8723, opened 2023-02-08
feat(performance): Update conversation data for chat messages and cal…

nickvergessen:
…ls from the notification

Ref #8710 

- [x] Require https://github.com/nextcloud/notifications/pull/1450
- [x] Update left sidebar when call notifications happen
- [x] Update left sidebar when chat notifications happen

### 🏁 Checklist

### DIFF
--- a/src/store/conversationsStore.js
+++ b/src/store/conversationsStore.js
@@ -97,7 +97,7 @@ const getters = {
 		return conversation.objectType !== 'room'
 	}),
 	/**
-	 * Get a conversation providing it's token
+	 * Get a conversation providing its token
 	 *
 	 * @param {object} state state object
 	 * @return {Function} The callback function returning the conversation object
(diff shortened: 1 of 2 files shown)


src/store/conversationsStore.js
```diff
@@ -477,6 +480,108 @@ const actions = {
 		}
 	},
 
+	async updateConversationLastMessageFromNotification({ getters, commit }, { notification }) {
+		let [token, messageId] = notification.objectId.split('/')
+		const conversation = { ...getters.conversation(token) }
+		messageId = parseInt(messageId, 10)
```
Antreesy, line 486: Guess we don't need that mutation

src/store/conversationsStore.js
```diff
@@ -477,6 +480,108 @@ const actions = {
...
+
+		const actor = notification.subjectRichParameters.user || notification.subjectRichParameters.guest || {
+			type: 'guest',
+			id: 'unknown',
+			name: t('spreed', 'Guest'),
+		}
+
+		const lastMessage = {
+			token,
+			id: parseInt(messageId, 10),
```
Antreesy, line 501: Because we're doing it here

PR #8726, opened 2023-02-09
"Modified since" support on roomlist

### LINKED ISSUE #7299: Skip old rooms from default room list and live updating?
### How to use GitHub

---

### Numbers
After getting some numbers I asked an individual on our instance with a lot of rooms (450+) if they are okay with checking their room last activity date. Since they agreed, here is the query + result:

#### By year+month only
```sql
SELECT YEAR(last_activity) AS year, MONTH(last_activity) AS month, COUNT(*)
FROM oc_talk_rooms
WHERE id IN (
	SELECT room_id
	FROM oc_talk_attendees
	WHERE actor_type = 'users'
	AND actor_id = '…')
GROUP BY year, month
ORDER BY year DESC, month DESC;
+------+-------+----------+
| year | month | COUNT(*) |
+------+-------+----------+
| 2022 |     5 |       38 |
| 2022 |     4 |       55 |
| 2022 |     3 |       28 |
| 2022 |     2 |       20 |
| 2022 |     1 |       12 |
| 2021 |    12 |       10 |
| 2021 |    11 |       14 |
| 2021 |    10 |       29 |
| 2021 |     9 |       14 |
| 2021 |     8 |        6 |
| 2021 |     7 |       16 |
| 2021 |     6 |       13 |
| 2021 |     5 |       16 |
| 2021 |     4 |       13 |
| 2021 |     3 |       13 |
| 2021 |     2 |        7 |
| 2021 |     1 |       11 |
| 2020 |    12 |        8 |
| 2020 |    11 |       10 |
| 2020 |    10 |        1 |
| 2020 |     9 |        9 |
| 2020 |     8 |        3 |
| 2020 |     7 |        4 |
| 2020 |     6 |        6 |
| 2020 |     5 |        8 |
| 2020 |     4 |        7 |
| 2020 |     3 |        8 |
| 2020 |     2 |        6 |
| 2020 |     1 |        3 |
| 2019 |    12 |        1 |
| 2019 |    11 |        9 |
| 2019 |    10 |        5 |
| 2019 |     9 |        3 |
| 2019 |     8 |        2 |
| 2019 |     7 |        5 |
| 2019 |     6 |        8 |
| 2019 |     5 |        3 |
| 2019 |     4 |        4 |
| 2019 |     3 |        1 |
| 2019 |     2 |        1 |
| 2019 |     1 |        2 |
| 2018 |    11 |        2 |
| 2018 |    10 |        6 |
| 2018 |     8 |        2 |
| 2018 |     7 |        1 |
| 2018 |     6 |        2 |
| 2018 |     5 |        1 |
| 2018 |     3 |        2 |
| 2018 |     2 |        1 |
| 2018 |     1 |        1 |
| 2017 |    12 |        1 |
| NULL |  NULL |        6 |
+------+-------+----------+
```

#### By year only
```sql
SELECT YEAR(last_activity) AS year, COUNT(*)
FROM oc_talk_rooms
WHERE id IN (
	SELECT room_id
	FROM oc_talk_attendees
	WHERE actor_type = 'users'
	AND actor_id = '…')
GROUP BY year
ORDER BY year DESC;
+------+----------+
| year | COUNT(*) |
+------+----------+
| 2022 |      153 |
| 2021 |      162 |
| 2020 |       73 |
| 2019 |       44 |
| 2018 |       18 |
| 2017 |        1 |
| NULL |        6 |
+------+----------+
```

### Problem / Impact
Currently we are doing multiple things every time we get the room list:
* Get user status for 1-1
* Get last comment and parse it
* Calculate unread messages

As per the above example it seems we could skip ~300 rooms for the user if we would only pull their rooms that had something happening in the last 4 months (as it is also very unlikely that they are needed ad hoc, and if they are and someone else posts in them, the next polling will bring them in). Another person on our instance with 550+ chats would also be reduced to ~160 chats:
```
+------+----------+
| year | COUNT(*) |
+------+----------+
| 2022 |      163 |
| 2021 |      199 |
| 2020 |       79 |
| 2019 |       51 |
| 2018 |       22 |
| 2017 |        1 |
| NULL |       36 |
+------+----------+
```

### New desired behaviour?
Since we have the search available, I wonder if we should just do similar to this:
* Only load conversations active in the last X months
* As a last entry add a dummy in the conversation list telling to use the search
* When the search is used we query the server already for open conversations and could just also return joined conversations matching the search term

### Potential issues
* Clients which keep a local copy of the conversation list need to adapt to the behaviour (unless we exclude them via user agent or something)
* When the active list is too short/empty, we might need a check on the size and could then make a second query (and we can cache the if) for those users to return more.

nickvergessen:
Fix #7299

When `modifiedSince` is provided only conversations with a newer `lastActivity` are returned.
Due to the nature of the data structure we can not return information that a conversation was deleted
or the user removed from a conversation. Therefore it is recommended to do a full refresh:

* Every 5 minutes
* When a signaling message of type `delete` or `disinvite` was received
* Internal signaling mode is used

### 🏁 Checklist

### DIFF
--- a/docs/conversation.md
+++ b/docs/conversation.md
@@ -13,14 +13,25 @@
     It is therefor recommended to use `format=json` or send the `Accept: application/json` header,
     to receive a JSON response.
 
+!!! note
+
+    When `modifiedSince` is provided only conversations with a newer `lastActivity` are returned.
+    Due to the nature of the data structure we can not return information that a conversation was deleted
+    or the user removed from a conversation. Therefore it is recommended to do a full refresh:
+    * Every 5 minutes
+    * When a signaling message of type `delete` or `disinvite` was received
+    * Internal signaling mode is used
+
+
 * Method: `GET`
 * Endpoint: `/room`
 * Data:
 
-| field            | type | Description                                                                                                      |
-|------------------|------|------------------------------------------------------------------------------------------------------------------|
-| `noStatusUpdate` | int  | Whether the "online" user status of the current user should be "kept-alive" (`1`) or not (`0`) (defaults to `0`) |
-| `includeStatus`  | bool | Whether the user status information of all one-to-one conversations should be loaded (default false)             |
+| field            | type | Description                                                                                                                                                                     |
+|------------------|------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `noStatusUpdate` | int  | Whether the "online" user status of the current user should be "kept-alive" (`1`) or not (`0`) (defaults to `0`)                                                                |
+| `includeStatus`  | bool | Whether the user status information of all one-to-one conversations should be loaded (default false)                                                                            |
+| `modifiedSince`  | int  | **Use with care as per note above.** When provided only conversations with a newer `lastActivity` (and one-to-one conversations when `includeStatus` is provided) are returned. |
 
 * Response:
     - Status code:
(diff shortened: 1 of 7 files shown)


SystemKeeper:
Is this basically https://github.com/nextcloud/spreed/issues/7299 ?

nickvergessen:
Yes

come-nc (approved):
:elephant: Only reviewed the PHP file

PR #8727, opened 2023-02-09
Feature/8339/breakout rooms request assistance

marcoambrosini:


Antreesy (approved):
Seems to work, but requires to test with receiving as moderator

PR #8748, opened 2023-02-14
Make participants counter always-on during the call

### LINKED ISSUE #8407: Participants count in call is not visible when sidebar is open
### How to use GitHub

---

## Steps to reproduce
1. Start a call and try to keep an overview about the participant count
2. It's visible in the topbar
Bildschirmfoto vom 2022-11-29 15-56-59

3. but as soon as you open the right sidebar for chatting or handling permissions etc, you can not see the count anymore. Keeping an eye on the participant cound and reading the chat in parallel is the most common usecase and task for webinar moderators.
Bildschirmfoto vom 2022-11-29 15-57-08

### Expected behaviour
The count should be visible.

### Actual behaviour
Only total participant count is visible.

### Solutions
1. Either we should something like (1/8) on the participants tab (but the string is dangerously long already and might trim in several langauges or screen resolutions, or 
2. We don't hide the participant icon when the sidebar is open
3. We move the information to another place where it is visible all the time (*cough* initial suggested and discussed subline *cough*)

Antreesy:
Fix #8407 

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
See mentioned issue | image

### 🚧 TODO

- [X] Show button when right sidebar is open
- [X] Correct hover effect (as per #8671)
- [ ] Fix: button counter react to attendees number change slowly when sidebar is closed
- [ ] Code review
- [ ] Visual check

### 🏁 Checklist

### DIFF
--- a/src/components/TopBar/TopBar.vue
+++ b/src/components/TopBar/TopBar.vue
@@ -309,15 +309,11 @@ export default {
 				this.notifyUnreadMessages(null)
 			}
 		},
+	},
 
-		// Starts and stops the getParticipantsMixin logic
-		isOneToOneConversation(newValue) {
-			if (newValue) {
-				this.initialiseGetParticipantsMixin()
-			} else {
-				this.stopGetParticipantsMixin()
-			}
-		},
+	beforeMount() {
+		// Initialises the get participants mixin for participants counter
+		this.initialiseGetParticipantsMixin()
 	},
 
 	mounted() {
@@ -335,6 +331,8 @@ export default {
 		document.removeEventListener('MSFullscreenChange', this.fullScreenChanged, false)
 		document.removeEventListener('webkitfullscreenchange', this.fullScreenChanged, false)
 		document.body.classList.remove('has-topbar')
+
+		this.stopGetParticipantsMixin()
 	},
 
 	methods: {
--- a/src/components/TopBar/TopBar.vue
+++ b/src/components/TopBar/TopBar.vue
@@ -107,20 +107,19 @@
 					{{ unreadMessagesCounter }}
 				</NcCounterBubble>
 			</div>
-
-			<!-- participants button -->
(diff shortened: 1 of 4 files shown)


marcoambrosini (commented):
Looks good, should we a tooltip that specifies that those are the participants in the call?

Ivansss:
Just as a suggestion, would make sense an icon like this?
https://fonts.google.com/icons?selected=Material%20Icons%3Acontact_phone%3A
I would also add a tooltip, but maybe using this icon above is more self explanatory :)

marcoambrosini:
hmm, it seems very different in style from our other icons, It's basically two icons in one and it's also inverted, as in the background is dark and the icons are bright.. Not sure about this one @Ivansss

Ivansss:
Yeah, after discussing it here at the office with @Antreesy we agreed to keep using the one currently used ;)

Antreesy:
So tooltip it is

src/components/TopBar/TopBar.vue
```diff
@@ -58,6 +58,19 @@
 		<CallTime v-if="isInCall"
 			:start="conversation.callStartTime" />
 
+		<!-- Participants counter -->
+		<NcButton v-if="isInCall && !isOneToOneConversation"
+			v-tooltip="t('spreed', 'Participants in call')"
+			class="top-bar__button"
+			type="tertiary"
+			@click="openSidebar('participants')">
```
nickvergessen, line 66: Tab is not available for guests.
Is this gracefully handled?

Antreesy: I'll check it. Should the counter be visible for guests in that case?

nickvergessen: I leave this up to you

marcoambrosini (approved):
Tested and works and looks good, good job :)

PR #8758, opened 2023-02-16
Apply aspect-ratio calculations from VideoVue to LocalVideo

Antreesy:
- [X] Copy settings and methods from #8676 for LocalVideo component.
This is required for subsequent improvements.

### 🏁 Checklist

### DIFF
--- a/src/components/CallView/shared/LocalVideo.vue
+++ b/src/components/CallView/shared/LocalVideo.vue
@@ -20,6 +20,7 @@
 
 <template>
 	<div id="localVideoContainer"
+		ref="videoContainer"
 		class="videoContainer videoView"
 		:class="videoContainerClass"
 		@mouseover="showShadow"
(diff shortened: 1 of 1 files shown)


nickvergessen:
Local video should always be fully visible in whatever aspect ratio it comes. The main reason is that we want people to know what others could see from their video/background area

Antreesy:
> Local video should always be fully visible in whatever aspect ratio it comes.

The full visibility of local video remains as is. It's just an enhancement to fix wrapper's height / width, so we could continue to apply same features or fixes to both components.

nickvergessen:
> The full visibility of local video remains as is. 

Hmm then there is a bug since before, e.g. the blue blanket in my video is not visible in the "own video", but shown for the other participant:

Bildschirmfoto vom 2023-02-17 15-43-27

nickvergessen (approved):
Seems to still work

ShGKme (commented):
This PR seems to be an almost duplication of https://github.com/nextcloud/spreed/pull/8676

Could we make it reusable with a new component or composable/mixin?

Antreesy:
> Could we make it reusable with a new component or composable/mixin?

Possible, but could be tricky, as we need to apply in for speaker's video only when big, and for local video in both cases

PR #8761, opened 2023-02-17
Always scroll chat when in a call

### LINKED ISSUE #7690: While in a call the chat doesn't keep scrolling to the end
### How to use GitHub

---

## Steps to reproduce
1. Join a call
2. Have the chat open
3. Move the window to a second screen to take notes in the other part
4. Colleagues post messages

### Expected behaviour
Chat should stay scrolled to the bottom and always show the last messages

### Actual behaviour
First message is like 80% visible and you can scroll to the bottom.
If you don't new messages are not shown

## Talk app

**Talk app version:** 14+

SystemKeeper:
Fixes #7690 

Before:
When in a call, but the window itself was not focused, the chat wouldn't scroll.

After:
When in a call, the chat will be scrolled when the window is visible (independent of being focused).

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -75,6 +75,7 @@ import MessagesGroup from './MessagesGroup/MessagesGroup.vue'
 import Axios from '@nextcloud/axios'
 import { subscribe, unsubscribe } from '@nextcloud/event-bus'
 import isInLobby from '../../mixins/isInLobby.js'
+import isInCall from '../../mixins/isInCall.js'
 import {
 	ATTENDEE,
 	CHAT,
@@ -102,6 +103,7 @@ export default {
 
 	mixins: [
 		isInLobby,
+		isInCall,
 	],
 
 	props: {
@@ -909,7 +911,7 @@ export default {
 		 */
 		smoothScrollToBottom() {
 			this.$nextTick(function() {
-				if (this.isWindowVisible && document.hasFocus()) {
+				if (this.isWindowVisible && (document.hasFocus() || this.isInCall)) {
 					// scrollTo is used when the user is watching
 					this.scroller.scrollTo({
 						top: this.scroller.scrollHeight,


SystemKeeper:
Honestly I'm not even sure why the `document.hasFocus()` is there in the first place. It was added here https://github.com/nextcloud/spreed/commit/c4e739c9a12b5ebf26f6f12bf9bb19527334a84b#diff-8161b5c80aa601ab393b624ced26b96b0fe998eb7bafa2e88545abd9b25ed410R784

But I guess that's more of a design decision. With this PR:
* When the window is not visible (determined by the browser) the chat will never be scrolled
* When in a call the chat will be scrolled when the window is visible
* When _not_ in a call, the chat will only be scrolled if the window is visible and focused.

So if we want to make the chat scroll even when not focused (like having a conversation open on the second screen), we could remove the in call check and the hasFocus check.

nickvergessen:
> Honestly I'm not even sure why the document.hasFocus() is there in the first place.

When you have another windows on top of the browser we didn't want to scroll?


> With this PR:
* When the window is not visible (determined by the browser) the chat will never be scrolled
* When in a call the chat will be scrolled when the window is visible
* When _not_ in a call, the chat will only be scrolled if the window is visible and focused.

I think that's a good step for now, we can still follow up later. What do you think @Antreesy @marcoambrosini

SystemKeeper:
> When you have another windows on top of the browser we didn't want to scroll?

Yes, that would be up to the browser to decide then. 

> I think that's a good step for now, we can still follow up later.

Agreed. Would be good test how the visibilty api works on different browsers and operating systems.

Antreesy (approved):
Tested, works good and exactly as described above.

:warning: When join a call, chat immediately scrolls down to the last message. I'm not sure, if we want that behavior, but looks fine

SystemKeeper:
> ⚠️ When join a call, chat immediately scrolls down to the last message. I'm not sure, if we want that behavior, but looks fine

That should be the case without this PR as well, right?

marcoambrosini (approved):

marcoambrosini:
Yeah that's the case currently, but @Antreesy raises a legitimate question, maybe we should open an issue for it?

PR #8772, opened 2023-02-17
Set local video in Stripe to 4:3 aspect-ratio

Antreesy:
### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image| image

As mentioned in corellated issue #8815 :

>... Even if the camera supports a higher resolution [in calls the resolution is requested to be 720x540](https://github.com/nextcloud/spreed/blob/de32c67ee0b1abf7f0f819ca03e577f951f6f9e7/src/utils/webrtc/VideoConstrainer.js#L187-L194). This is a default value chosen for performance reasons, both for CPU load when encoding the video and for network bandwidth when sending it.

Maximum capable video resolution is fixed at 4:3. This PR adjusts width and height of local video frame in stripe to fit in desired resolution and avoid its cropping

### 🚧 TODO

- [X] Remove video width lock at 300px
- [X] Set aspect-ratio to mentioned resolution at 4:3
- [ ] Code review 

### 🏁 Checklist

### DIFF
--- a/src/components/CallView/Grid/Grid.vue
+++ b/src/components/CallView/Grid/Grid.vue
@@ -882,12 +882,8 @@ export default {
 	padding: 16px;
 }
 
-.local-video-stripe {
-	width: 300px;
-}
-
 .stripe-wrapper {
-	width: calc(100% - 300px);
+	width: 100%;
 	position: relative;
 }
 
--- a/src/components/CallView/shared/LocalVideo.vue
+++ b/src/components/CallView/shared/LocalVideo.vue
@@ -72,11 +72,7 @@ import attachMediaStream from 'attachmediastream'
 import Hex from 'crypto-js/enc-hex.js'
 import SHA1 from 'crypto-js/sha1.js'
 
-import {
-	showError,
-	showInfo,
-	TOAST_PERMANENT_TIMEOUT,
-} from '@nextcloud/dialogs'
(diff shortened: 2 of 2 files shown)


ShGKme (commented):
I have this view on my local test. Is it was was expected (it is not cropped)? Shouldn't it be stretch so the video would have rounded corners?

image

ShGKme:
Hmm... Now it looks ok. 

image

ShGKme:
Step to reproduce:
1. Make a call from account 1 in a group chat
2. Join a call with account 2 with an enabled camera
   
3. By default, account 1 will be shown. The Video is cropped.
   image
4. Leave and re-join. The video is not cropped, but also not rounded:
   image
5. Switch to the grid view and back to the speaker view. Everything is ok.
   image

ShGKme:
I may admit that this may be a peculiarity of my environment... Does it work fine on yours?

ShGKme (commented):
The code looks fine, no cropping now. The aspect ratio is 1.33333 fallback until reconnect and reopen the local video on my test.

Is it what is expected in the PR? 

I'm not sure if we want to merge it without solving https://github.com/nextcloud/spreed/issues/8815.

change

Antreesy:
> Is it what is expected in the PR?

Yes, because videostream comes with `aspect-ratio: 1.33`, and previous styles cutted it even more, up to 1.24.
I want to fix at least this for now, and use as a basement in following up fixes on mentioned issue

ShGKme (approved):

PR #8798, opened 2023-02-20
Make recording files visible in the specific section Media tab of the…

### LINKED ISSUE #8663: Make recording files visible in the specific section Media tab of the sidebar
Original issue: https://github.com/nextcloud/spreed/issues/8326

Original requirement:

> The recording shall be shared in the chat as a "normal file" but have a special section in the Media tab of the sidebar.

vitormattos:
### 🚧 TODO

- close #8663

### 🏁 Checklist

### DIFF
--- a/lib/Model/Attachment.php
+++ b/lib/Model/Attachment.php
@@ -47,6 +47,8 @@ class Attachment extends Entity {
 	public const TYPE_MEDIA = 'media';
 	public const TYPE_OTHER = 'other';
 	public const TYPE_POLL = 'poll';
+	public const TYPE_RECORD_AUDIO = 'record-audio';
+	public const TYPE_RECORD_VIDEO = 'record-video';
 	public const TYPE_VOICE = 'voice';
 
 	/** @var int */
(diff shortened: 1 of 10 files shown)


lib/Model/Attachment.php
```diff
@@ -47,6 +47,8 @@ class Attachment extends Entity {
 	public const TYPE_MEDIA = 'media';
 	public const TYPE_OTHER = 'other';
 	public const TYPE_POLL = 'poll';
+	public const TYPE_RECORD_AUDIO = 'record-audio';
```
nickvergessen, line 50: Needs to be documented in https://github.com/nextcloud/spreed/blob/062ed28540c21d09a19cf7077033e1b0165658c4/docs/constants.md#shared-item-types

nickvergessen (changes requested):

nickvergessen:
I'm actually unsure if it should be 2 sections or one, but then again with the different types it might be better in split sections?
cc @marcoambrosini

nickvergessen:
PS screenshot in https://github.com/nextcloud/spreed/pull/8800

vitormattos:
Updated the web client implementation here: #8800 
I unified the sections only on the client side.

nickvergessen:
> I unified the sections only on the client side.

That's not good enough, as it needs more coordination with mobile clients, breaks the pagination, etc.

nickvergessen (approved):

PR #8808, opened 2023-02-21
Fix `NcRichContenteditable` styles overriding in `NewMessageForm`

ShGKme:
### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 TODO

Style overriding in https://github.com/nextcloud/spreed/pull/4333 worked on local tests but actually depends on building order and other apps. For example, `NcRichContenteditable` may be loaded on the page several times. Another instance may be loaded after the Talk.

~~Adding `!import` makes styles independent of order and duplications.~~
Increase specificity by combinator.

- [x] Fix `NcRichContenteditable` styles overriding in `NewMessageForm`

### 🏁 Checklist

### DIFF
--- a/src/components/NewMessageForm/NewMessageForm.vue
+++ b/src/components/NewMessageForm/NewMessageForm.vue
@@ -921,9 +921,10 @@ export default {
 			position: relative;
 		}
 
-		&__richContenteditable {
+		// Override NcRichContenteditable styles
+		& &__richContenteditable {
 			border: 1px solid var(--color-border-dark);
-			border-radius: calc($clickable-area / 2);
+			border-radius: calc(var(--default-clickable-area) / 2);
 			padding: 8px 16px 8px 44px;
 			max-height: 180px;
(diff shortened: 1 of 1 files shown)


src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -921,15 +921,16 @@ export default {
 			position: relative;
 		}
 
+		// Override NcRichContenteditable styles
 		&__richContenteditable {
```
Antreesy, line 925:
```suggestion
 	 & 	&__richContenteditable {
```
I'm not a fan of using `!important`, while it could cause some troubles in future with style overriding.
Probably mentioned above nesting should be enough

ShGKme: Agree. There is a better solution.

nickvergessen (approved):

PR #8826, opened 2023-02-22
Allow users to enter and exit breakout rooms

marcoambrosini:


src/components/TopBar/CallButton.vue
```diff
@@ -38,7 +38,7 @@
 			</template>
 			{{ startCallLabel }}
 		</NcButton>
-		<NcButton v-else-if="showLeaveCallButton && !canEndForAll"
+		<NcButton v-else-if="(showLeaveCallButton && !canEndForAll) && !isBreakoutRoom"
```
nickvergessen, line 41: nop

src/components/TopBar/CallButton.vue
```diff
@@ -48,20 +48,28 @@
 			</template>
 			{{ leaveCallLabel }}
 		</NcButton>
-		<NcActions v-else-if="showLeaveCallButton && canEndForAll"
+		<NcActions v-else-if="showLeaveCallButton && canEndForAll || isBreakoutRoom"
```
nickvergessen, line 51: combining && and || on the same level does mostly not what you think.

nickvergessen (approved):
🐘

PR #8837, opened 2023-02-24
 feat(recording): Add a shorter subject for the call recording notification

### LINKED ISSUE #8835: Adjust subject/message for recording notifications
Currently we only send a subject in the notification, which then looks like this on iOS:

I think it would be better to move the current subject to message and have more general subject:

For reference, this is in web right now:
Bildschirm­foto 2023-02-23 um 23 02 09

If we don't want (or can) change the way the notification is sent, I'd suggest we adjust the notification in iOS (we need to do it anyway because of the actions, so that would be easy). In this case we could use the subject as the notification body, which will show the app name as the title:

(Or add our own title)

If string freeze is a problem, we could move subject to message (so same string) and just use "Recording" as a subject for now as that's already available:
https://github.com/nextcloud/spreed/blob/79aeeace984bbca2774c8b0aaf53c581d3af519b/l10n/en_GB.js#L1250
Or leave the subject empty, which will result in displaying "Nextcloud Talk" as seen above.

CC: @mahibi

nickvergessen:
Fix #8835

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Bildschirmfoto vom 2023-02-24 07-12-36 | Bildschirmfoto vom 2023-02-24 07-12-19

### 🚧 TODO

- [x] Add a shorter subject to the notification which survives the pushing
- [x] Make the toast about the post-processing show infinite

### 🏁 Checklist

### DIFF
--- a/lib/Notification/Notifier.php
+++ b/lib/Notification/Notifier.php
@@ -345,8 +345,9 @@ class Notifier implements INotifier {
 			);
 
 		$notification
-			->setRichSubject(
-				$l->t('Recording for the call in {call} was uploaded to {file}.'),
+			->setRichSubject($l->t('Call recording now available'))
+			->setRichMessage(
+				$l->t('The recording for the call in {call} was uploaded to {file}.'),
 				[
 					'call' => [
 						'type' => 'call',
--- a/src/store/conversationsStore.js
+++ b/src/store/conversationsStore.js
@@ -22,7 +22,7 @@
 import Vue from 'vue'
 
 import { getCurrentUser } from '@nextcloud/auth'
-import { showInfo, showSuccess } from '@nextcloud/dialogs'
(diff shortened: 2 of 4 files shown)


SystemKeeper (approved):
Fixes #8835 

Can not test it right now, but from the code and screenshot, that’s exactly what I had in mind with #8835.
As the toast is now permanent, is there a way to close it once we receive the available notification?

nickvergessen:
> As the toast is now permanent, is there a way to close it once we receive the available notification?

Yeah, it has a closing X all the time (even when having the default 7 sec timeout)
grafik

SystemKeeper:
Sorry, I meant code-wise, when we receive then notification. But I guess there’s no way to identify the relation between notification and toast.

nickvergessen:
> Sorry, I meant code-wise, when we receive then notification. But I guess there’s no way to identify the relation between notification and toast.

yeah unluckily not.

PR #8846, opened 2023-02-24
Send notification when get upload problem

### LINKED ISSUE #8824: Add a warning in the admin recording settings about post_max_size and upload_max_filesize
After testing recordings for some minutes I ran into the following warning being logged:
```
Uploaded file might be larger than allowed by PHP settings 'upload_max_filesize' or 'post_max_size' 
```
https://github.com/nextcloud/spreed/blob/6533d77149c1d622e2f9ab710f2c0cd112d5e57d/lib/Service/RecordingService.php#L135-L140

That should:
1. Be logged as an error
2. Trigger a notification for the moderator so they know the recording will never appear and they need to reach out to the admin

vitormattos:
> PS: I couldn't reproduce the reported problem but I did changes to give a best error report and send the notification.

### Changes

* Notify as `error` when is an upload error
* Fix the error message to make more specific as the occurrence

### 🖼️ Screenshots

Screenshot_20230224_165212

### 🚧 TODO

- [ ] Close #8824

### 🏁 Checklist

### DIFF
--- a/lib/Chat/Notifier.php
+++ b/lib/Chat/Notifier.php
@@ -314,8 +314,16 @@ class Notifier {
 		// @todo this should be in the Notifications\Hooks
 		$notification->setApp('spreed');
 
-		$notification->setObject('chat', $chat->getToken());
-		$this->notificationManager->markProcessed($notification);
+		$objectTypes = [
+			'chat',
+			'recording',
+			'recording_information',
+			'remote_talk_share',
+		];
+		foreach ($objectTypes as $type) {
+			$notification->setObject($type, $chat->getToken());
+			$this->notificationManager->markProcessed($notification);
+		}
 
 		if (!$chatOnly) {
 			$notification->setObject('room', $chat->getToken());
--- a/lib/Chat/Notifier.php
+++ b/lib/Chat/Notifier.php
@@ -316,20 +316,19 @@ class Notifier {
 
 		$objectTypes = [
 			'chat',
-			'recording',
-			'recording_information',
-			'remote_talk_share',
 		];
+		if (!$chatOnly) {
+			$objectTypes = [
+				'call',
+				'chat',
+				'room',
+				'recording',
+				'recording_information',
+				'remote_talk_share',
+			];
+		}
 		foreach ($objectTypes as $type) {
 			$notification->setObject($type, $chat->getToken());
-			$this->notificationManager->markProcessed($notification);
-		}
-
-		if (!$chatOnly) {
-			$notification->setObject('room', $chat->getToken());
-			$this->notificationManager->markProcessed($notification);
-
-			$notification->setObject('call', $chat->getToken());
 			$this->notificationManager->markProcessed($notification);
 		}
 
--- a/lib/Notification/Notifier.php
+++ b/lib/Notification/Notifier.php
@@ -258,6 +258,9 @@ class Notifier implements INotifier {
 		if ($subject === 'record_file_stored') {
 			return $this->parseStoredRecording($notification, $room, $participant, $l);
 		}
+		if ($subject === 'record_file_store_fail') {
+			return $this->parseStoredRecordingFail($notification, $room, $participant, $l);
+		}
 		if ($subject === 'invitation') {
 			return $this->parseInvitation($notification, $room, $l);
 		}
@@ -295,6 +298,27 @@ class Notifier implements INotifier {
 			$temp = mb_substr($temp, 0, -5);
 		}
 		return $temp;
+	}
+
+	protected function parseStoredRecordingFail(
+		INotification $notification,
+		Room $room,
+		Participant $participant,
+		IL10N $l
+	): INotification {
+		$notification
+			->setRichSubject(
+				$l->t('Fail to store recording of call {call}, reach out to the admin.'),
+				[
+					'call' => [
+						'type' => 'call',
+						'id' => $room->getId(),
+						'name' => $room->getDisplayName($participant->getAttendee()->getActorId()),
+						'call-type' => $this->getRoomType($room),
+					],
+				]
+			);
+		return $notification;
 	}
 
 	protected function parseStoredRecording(
--- a/lib/Notification/Notifier.php
+++ b/lib/Notification/Notifier.php
@@ -308,7 +308,10 @@ class Notifier implements INotifier {
 	): INotification {
 		$notification
 			->setRichSubject(
-				$l->t('Failed to store recording of call {call}, reach out to the admin.'),
+				$l->t('Failed to upload call recording'),
+			)
+			->setRichMessage(
+				$l->t('The recording server failed to upload recording of call {call}. Please reach out to the adminstrators.'),
 				[
 					'call' => [
 						'type' => 'call',
--- a/lib/Notification/Notifier.php
+++ b/lib/Notification/Notifier.php
@@ -308,7 +308,7 @@ class Notifier implements INotifier {
 	): INotification {
 		$notification
 			->setRichSubject(
-				$l->t('Fail to store recording of call {call}, reach out to the admin.'),
+				$l->t('Failed to store recording of call {call}, reach out to the admin.'),
 				[
 					'call' => [
 						'type' => 'call',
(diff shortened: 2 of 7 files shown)


SystemKeeper:
Judging from the code this looks problematic for iOS (and I guess for android as well), because of the recording type. Will try it and report back.

SystemKeeper:
I see 2 problems here.

Visually: This notification will consist of a subject only, which might look weird on mobile devices (see #8835). As it's a bit shorter than the other notification, it might still be okay to go with a subject only. Personally I think a short subject and a more detailed message would look better.

Object type: Since this notification has type `recording` set, this would lead to showing recording actions on the native iOS notification (see again #8835 for an example). We have the following options:
* Check if actions are actually available in the notification. If not, don't register any actions with the notification. 
* We introduce a dedicated type like "information" which the mobile apps just display without further modifications (on iOS that means the notification subject is the title and the notification message is the body). Only downside here is that we can't navigate to the conversation if the notification is tapped on, because we could only assume that the object id is a room token.
* Like above but use a type like `room_information` or `room_message` so it's clear that it is an arbitrary message for a specific room and we can be sure that the object id is in fact a room token.
* For the record: We could use the type `chat` to send the failed notification. I would not go with this one as a chat notifications implies to have a sender. In this case we would show a notification from "Guest". Also you would be able to directly reply to that message, which doesn't make sense.

lib/Notification/Notifier.php
```diff
@@ -297,6 +300,27 @@ protected function shortenJsonEncodedMultibyteSave(string $subject, int $dataLen
...
 
+	protected function parseStoredRecordingFail(
+		INotification $notification,
+		Room $room,
+		Participant $participant,
+		IL10N $l
+	): INotification {
+		$notification
+			->setRichSubject(
+				$l->t('Fail to store recording of call {call}, reach out to the admin.'),
```
SystemKeeper, line 311:
```suggestion
				$l->t('Failed to store recording of call {call}, reach out to the admin.'),
```

Not sure if it's correct, but I would go with "Failed" here?!

mahibi:
> Like above but use a type like room_information or room_message (or just room) so it's clear that it is an arbitrary message for a specific room and we can be sure that the object id is in fact a room token.

i think this is the best solution

vitormattos:
Now we have the current types of objects:
```
call
chat
hosted
recording
remote_talk_share
room
```
In the following places:
```
lib/Federation/CloudFederationProviderTalk.php:312:			->setObject('remote_talk_share', $shareId)
lib/Chat/Notifier.php:317:		$notification->setObject('chat', $chat->getToken());
lib/Chat/Notifier.php:321:			$notification->setObject('room', $chat->getToken());
lib/Chat/Notifier.php:324:			$notification->setObject('call', $chat->getToken());
lib/Chat/Notifier.php:349:			->setObject('chat', $chat->getToken())
lib/Chat/Notifier.php:412:			->setObject('chat', $chat->getToken())
lib/Service/RecordingService.php:151:				->setObject('recording $room->getToken())
lib/Service/RecordingService.php:209:			->setObject('recording', $room->getToken())
lib/Service/RecordingService.php:220:			->setObject('recording', $room->getToken())
lib/Activity/Listener.php:181:				->setObject('room', $room->getId())
lib/Activity/Listener.php:226:				->setObject('room', $room->getId())
lib/Activity/Listener.php:259:					->setObject('room', $room->getId(), $roomName)
lib/Controller/PageController.php:204:						->setObject('room', $room->getToken());
lib/Controller/PageController.php:206:					$notification->setObject('call', $room->getToken());
lib/Notification/Notifier.php:523:			$notification->setObject('chat', $notification->getObjectId() . '/' . $comment->getId());
lib/Notification/Listener.php:134:				->setObject('room', $room->getToken())
lib/Notification/Listener.php:185:				->setObject('room', $room->getToken())
lib/Notification/Listener.php:235:				->setObject('room', $room->getToken());
lib/Notification/Listener.php:238:			$notification->setObject('call', $room->getToken());
lib/Notification/Listener.php:294:				->setObject('call', $room->getToken())
lib/Notification/Listener.php:314:				->setObject('call', $room->getToken())
lib/BackgroundJob/CheckHostedSignalingServer.php:153:				->setObject('hosted-signaling-server', $notificationSubject)
```

What's the best to use in this case?
`room_information` was never used but sounds interesting.
`room_message` could generate Misunderstanding that it's a message
`room` maybe also is good but too much generic

lib/Notification/Notifier.php
vitormattos: Thanks! I will push when solve the question about type.

vitormattos:
Looking about the type I identified that when we delete the room, have a method to delete notifications and this method now only consider the types `room`, `chat` and `call`.

`recording` is a new type and we need think that we will maintain this specific type or rename to `chat` that is the most generic type.

nickvergessen:
`room` is for invites and has actions as well so can not be reused
Let's use `recording_information` ?🤔

nickvergessen:
> `recording` is a new type and we need think that we will maintain this specific type or rename to `chat` that is the most generic type.

Can not be the same as the actions are completely different.

vitormattos:
> Can not be the same as the actions are completely different.

Then, we need to remove the recording notifications when the room is deleted.

SystemKeeper:
`recording_information` would be totally fine for me. 

> or rename to `chat` that is the most generic type.

From the name, yes, but not from the processing done in the background. When used like this, you would receive a notification from "Guest", which is not what we want here.

lib/Chat/Notifier.php
```diff
@@ -314,8 +314,16 @@ public function removePendingNotificationsForRoom(Room $chat, bool $chatOnly = f
 		// @todo this should be in the Notifications\Hooks
 		$notification->setApp('spreed');
 
-		$notification->setObject('chat', $chat->getToken());
-		$this->notificationManager->markProcessed($notification);
+		$objectTypes = [
```
nickvergessen, line 317: This function is called when the chat history is cleared, not sure we want to remove recording notifications and information at that case?

lib/Notification/Notifier.php
```diff
@@ -297,6 +300,27 @@ protected function shortenJsonEncodedMultibyteSave(string $subject, int $dataLen
...
 
+	protected function parseStoredRecordingFail(
+		INotification $notification,
+		Room $room,
+		Participant $participant,
+		IL10N $l
+	): INotification {
+		$notification
+			->setRichSubject(
+				$l->t('Failed to store recording of call {call}, reach out to the admin.'),
```
nickvergessen, line 311:
```suggestion
			->setRichSubject(
				$l->t('Failed to upload call recording'),
			)
			->setRichMessage(
				$l->t('The recording server failed to upload recording of call {call}. Please reach out to the adminstrators.'),
```

lib/Chat/Notifier.php
nickvergessen, line 317: This function is called when the chat history is cleared, not sure we want to remove recording notifications and information at that case?

nickvergessen (changes requested):

lib/Chat/Notifier.php
vitormattos: If we consider that the notification is part of chat history, yes.
This method also is called when the room is deleted and in this case, is necessary to delete the notification because all data related of the room is deleted.

nickvergessen: Well there is a `$chatOnly` parameter and that is the difference, but yeah maybe it's fine.

vitormattos: Is this new notification at context of `$chatOnly`, right?

nickvergessen (changes requested):

lib/Chat/Notifier.php
```diff
@@ -314,8 +314,16 @@ public function removePendingNotificationsForRoom(Room $chat, bool $chatOnly = f
...
+			'chat',
+			'recording',
+			'recording_information',
+			'remote_talk_share',
```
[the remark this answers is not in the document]
nickvergessen: should be fine, I wouldn't use array_merge however and just specify the `'chat'` entry again:

```php
		if (!$chatOnly) {
			$objectTypes = [
				'chat',
				'call',
				'room',
				'recording',
				'recording_information',
				'remote_talk_share',
			]);
		}
```

nickvergessen (approved):

PR #8861, opened 2023-02-27
Fix message type of files shared from mobile clients

### LINKED ISSUE #8858: Shared files are returned as system messages
### How to use GitHub

---

## Steps to reproduce
1. Open a conversation with shared files

### Expected behaviour
Preview of the file / image should be shown. The message is reactable and can be replied to.

### Actual behaviour
The message is received as a system message which neither is reactable, nor can be replied to.

### Additional infos

The problem comes from #8798, in particular this part:
https://github.com/nextcloud/spreed/blob/9184830ed995a258417f1d4e7d43752c3a9dcc1c/lib/Chat/Parser/SystemMessage.php#L414-L424

For files shared `$metaData['messageType']` is set, but looks like it's an empty string?! Before the change the check for `voice-message` would fail and therefore fallback to setting the message type to `ChatMananger::VERB_MESSAGE`, this is not the case anymore.

Comparison NC 25.0.4 and NC 26 Beta 5:
image

nickvergessen:
Test fails before the patch:
> grafik

Fix #8858 

### 🏁 Checklist

### DIFF
--- a/lib/Chat/Parser/SystemMessage.php
+++ b/lib/Chat/Parser/SystemMessage.php
@@ -418,6 +418,8 @@ class SystemMessage {
 						$chatMessage->setMessageType('record-audio');
 					} elseif ($metaData['messageType'] === 'record-video') {
 						$chatMessage->setMessageType('record-video');
+					} else {
+						$chatMessage->setMessageType(ChatManager::VERB_MESSAGE);
 					}
 				} else {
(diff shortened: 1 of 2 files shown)


SystemKeeper (approved):

nickvergessen:
Failing test seems unrelated

PR #8870, opened 2023-02-27
Allow to abort emoji and user autocompletion

### LINKED ISSUE #8849: Need a way out of the emoji autocomplete.
### How to use GitHub

---

## Steps to reproduce
1. Type smilies like `:D` and `:)` at the end of your message and try to submit
2. It's okay that emoji completion is started, but "ESC" should exit it (like it does on github), but that completely unfocuses the chat input.

### Expected behaviour
```
Hello :)
What? :D
```

### Actual behaviour
grafik

nickvergessen:
Fix #8849 

### 🏁 Checklist

### DIFF
--- a/src/components/NewMessageForm/NewMessageForm.vue
+++ b/src/components/NewMessageForm/NewMessageForm.vue
@@ -123,7 +123,7 @@
 						:placeholder="placeholderText"
 						:aria-label="placeholderText"
 						@shortkey="focusInput"
-						@keydown.esc.prevent="blurInput"
+						@keydown.esc="blurInput"
 						@paste="handlePastedFiles"
 						@submit="handleSubmit({ silent: false })" />
 				</div>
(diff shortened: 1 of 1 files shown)


src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -883,8 +883,19 @@ export default {
...
+		/**
+		 * Escape was used inside the chat input
+		 *
+		 * Unfocus it when there is no message, otherwise this is used to abort
+		 * autocompletion of emojis and users.
+		 *
+		 * @param {Event} event Key down event
+		 */
+		blurInput(event) {
+			if (this.text.trim() === '') {
```
Antreesy, line 895: Maybe we should check only what we have either `:`, `:<string>`, `/`, `/<string>`, `@` or `@<string>` as last symbol group in the input value to disable blur?
From current solution, i guess there is no way to blur component by `Esc` if there is something written

nickvergessen: Was fixed now properly by @ShGKme 

I was trying it the simple way as it's still a better thing then the current behaviour. But now it's properly.

marcoambrosini (approved):

lukyrys:
Is there any way to disable autocomplete? or eg. autoclose on basic smiles..  when i type ":D" on end of message i will need to double enter for send message.. this is annoying

PR #8874, opened 2023-02-27
Don't claim recording is enabled when no secret or servers are config…

nickvergessen:
…ured

E.g. currently on c.nc.c

### 🏁 Checklist

### DIFF
--- a/lib/Config.php
+++ b/lib/Config.php
@@ -160,12 +160,24 @@ class Config {
 	}
 
 	public function isRecordingEnabled(): bool {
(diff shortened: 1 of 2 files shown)


nickvergessen:
Let's see if current integration tests need adjusting

vitormattos (approved):

nickvergessen:
Drone passed on the first commit and after that only unit tests where adjusted.
Those passed on the second commit, so should be fine => merging

PR #8895, opened 2023-03-01
Fix scroll behaviour and "Scroll to bottom" button appearance

### LINKED ISSUE #8856: Unable to scroll to the bottom of the message list
### How to use GitHub

---

## Steps to reproduce
1. Open a conversation
2. Notice that the "scroll to bottom" arrow is visible
3. Try to scroll to the bottom with the mouse

### Expected behaviour
The arrow and the border above the input should be hidden (in fact, I am already at the bottom of the list)

### Actual behaviour
The only way to "move" to the bottom is to click on the arrow (it doesn't change the scroll location, as I am already at the bottom, but the arrow and the border disappears)

Bildschirm­foto 2023-02-25 um 22 23 19

When reloading the page it sometimes works even with scrolling, after switching from conversation 1 to conversation 2 it's not working.

## Talk app

**Talk app version:** 16 Beta 1 on c.nc.c

## Browser

**Browser name:** Chrome & Safari

**Browser version:** 110 & 16.3

Antreesy:
Fix #8856 

- [X] Hide button by default
- [X] Update its appeance when scrolling (button wasn't shown when scrolling up)
- [X] Fixed some typos, remove redundant methods

### 🚧 TODO

- [ ] Code review

### 🏁 Checklist

### DIFF
--- a/src/components/ChatView.vue
+++ b/src/components/ChatView.vue
@@ -40,20 +40,15 @@
 		</transition>
 		<MessagesList role="region"
 			:aria-label="t('spreed', 'Conversation messages')"
-			:is-chat-scrolled-to-bottom="isChatScrolledToBottom"
 			:token="token"
-			:is-visible="isVisible"
-			@set-chat-scrolled-to-bottom="setScrollStatus" />
+			:is-visible="isVisible" />
 		<NewMessageForm role="region"
 			:token="token"
-			:is-chat-scrolled-to-bottom="isChatScrolledToBottom"
 			:aria-label="t('spreed', 'Post message')" />
 	</div>
 </template>
 
 <script>
-import { getCapabilities } from '@nextcloud/capabilities'
-
 import MessagesList from './MessagesList/MessagesList.vue'
 import NewMessageForm from './NewMessageForm/NewMessageForm.vue'
 
@@ -78,7 +73,6 @@ export default {
 	data() {
 		return {
 			isDraggingOver: false,
-			isChatScrolledToBottom: getCapabilities()?.spreed?.config?.chat?.legacy || false,
 		}
 	},
 
@@ -133,10 +127,6 @@ export default {
 			const uploadId = new Date().getTime()
 			// Uploads and shares the files
 			this.$store.dispatch('initialiseUpload', { files, token: this.token, uploadId })
-		},
-
-		setScrollStatus(payload) {
-			this.isChatScrolledToBottom = payload
 		},
 	},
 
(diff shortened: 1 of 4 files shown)


src/components/ChatView.vue
```diff
@@ -78,7 +78,7 @@ export default {
 	data() {
 		return {
 			isDraggingOver: false,
-			isChatScrolledToBottom: getCapabilities()?.spreed?.config?.chat?.legacy || false,
+			isChatScrolledToBottom: getCapabilities()?.spreed?.config?.chat?.legacy || true,
```
nickvergessen, line 81: This basically rolls all the context changes back...

Antreesy: We still have a property `this.loadChatInLegacyMode` inside MessageList.vue, which basically does the same check.

I could change it back to `false`, but button is flickering in that case

ShGKme (commented):
image


Step to reproduce:

1. Create an empty chat
2. Open a chat with a history
3. Make some scroll
4. Switch to the empty chat
5. "Scroll to button" is expected to be invisible but it is

src/components/MessagesList/MessagesList.vue
```diff
@@ -1073,6 +1056,13 @@ export default {
 	flex: 1 0;
 	overflow-y: auto;
 	overflow-x: hidden;
+	border-bottom: 1px solid var(--color-border);
+	transition: border-bottom-color 150ms ease-in-out;
```
nickvergessen, line 1060: Any reason not to use one of these?
https://github.com/nextcloud/server/blob/24ffe328452e0e247017894ea9d49944470fd2bc/apps/theming/lib/Themes/DefaultTheme.php#L166-L167

nickvergessen (approved):
Seems to work, but I'm really scared

src/components/MessagesList/MessagesList.vue
Antreesy: To be sync with transition:fade effect. Could be replaced by variable though:
https://github.com/nextcloud/spreed/blob/915ab9d92b96d7e6b4dc8a6b769a743c8862d28f/src/assets/variables.scss#L100

src/components/MessagesList/MessagesList.vue
```diff
@@ -1057,7 +1057,7 @@ export default {
 	overflow-y: auto;
 	overflow-x: hidden;
 	border-bottom: 1px solid var(--color-border);
-	transition: border-bottom-color 150ms ease-in-out;
+	transition: $fade-transition;
```
nickvergessen, line 1060: This is not `all` but `border-bottom-color`, is it still fine?

PR #8058, opened 2022-09-30
fix and improve mobile UX

szaimen:
Fix https://github.com/nextcloud/spreed/issues/1937

Signed-off-by: szaimen <szaimen@e.mail.de>

<details>
<summary>For my own testing</summary>

```
docker run -it ^
-e TRUSTED_DOMAIN=192.168.178.155 ^
-e TALK_BRANCH=enh/1937/mobile ^
--name nextcloud-easy-test ^
-p 8443:443 ^
--volume="nextcloud_easy_test_npm_cache_volume:/var/www/.npm" ^
ghcr.io/szaimen/nextcloud-easy-test:latest
```

</details>

PR #8898, opened 2023-03-01
do not focus input on mobile

szaimen:
Follow-up to https://github.com/nextcloud/spreed/pull/8058

Regression of https://github.com/nextcloud/spreed/pull/4333

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

<details>
<summary>For my own testing</summary>

```
docker run -it \
--name nextcloud-easy-test \
-p 8443:443 \
-e TRUSTED_DOMAIN=192.168.24.128 \
--volume="nextcloud_easy_test_npm_cache_volume:/var/www/.npm" \
-e TALK_BRANCH=fix/noid/mobile-focus \
ghcr.io/szaimen/nextcloud-easy-test:latest
```

</details>

### DIFF
--- a/src/components/NewMessageForm/NewMessageForm.vue
+++ b/src/components/NewMessageForm/NewMessageForm.vue
@@ -880,6 +880,9 @@ export default {
 		},
 
 		focusInput() {
+			if (this.isMobile()) {
+				return
+			}
 			this.$nextTick().then(() => {
 				this.$refs.richContenteditable.$refs.contenteditable.focus()
 			})
@@ -895,6 +898,10 @@ export default {
 			if (!this.isTributePickerActive) {
 				this.blurInput()
 			}
+		},
+
+		isMobile() {
+			return /Android|iPhone|iPad|iPod/i.test(navigator.userAgent)
 		},
 	},
 }


src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -895,6 +898,10 @@ export default {
 			if (!this.isTributePickerActive) {
 				this.blurInput()
 			}
+		},
+
+		isMobile() {
+			return /Android|iPhone|iPad|iPod/i.test(navigator.userAgent)
```
nickvergessen, line 904: Should we use the `isMobile` mixing from the vue lib instead?
https://github.com/nextcloud/nextcloud-vue/blob/master/src/utils/IsMobileState.js
But seems to go by resolution rather then user agent.

szaimen: Yes, in this case here I would by purpose use the user agent instead of the resolution since the size of the window could also be small on desktop and you would still want to focus the input for them. This is also how it was done in https://github.com/nextcloud/spreed/pull/8058/files

szaimen:
Tested and works. Ready for review 👍

szaimen:
⚠️Rebased⚠️

nickvergessen (approved):

PR #8904, opened 2023-03-01
Use NcRichText component from `@nextcloud/vue`

julien-nc:
:tada: Circular dependency is definitely going down with the next `@nextcloud/vue` release which will include `NcRichText` and everything needed for the link picker and the reference widgets.

refs https://github.com/nextcloud/nextcloud-vue/pull/3841 https://github.com/nextcloud/nextcloud-vue/issues/3812

I'll update `@nextcloud/vue` and set this to "ready for review" once the new release is out.


julien-nc:
Ready IMO.

marcoambrosini (approved):

nickvergessen:
Jest is fully red?

julien-nc:
Here is the problem.
```
Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.
```
@juliushaertl @raimund-schluessler Any hint on why? Would it make sense to ignore some stuff with `transformIgnorePatterns` like suggested by jest?

raimund-schluessler:
> @juliushaertl @raimund-schluessler Any hint on why? Would it make sense to ignore some stuff with `transformIgnorePatterns` like suggested by jest?

Well, by default jest ignores all packages in `node_modules` and does not transform them. So the `transformIgnorePatterns` works in the opposite way than one would expect. Setting
```json
		"transformIgnorePatterns": [
			"/node_modules/(?!(.*)/)"
		],
```
as we do for `nextcloud/vue` at the moment will lead to everything in `node_modules` being transformed, which will likely make your tests run and pass, but will increase the runtime by a lot, so not sure you want this.

To be honest, I wonder why this didn't happen with the standalone package `nextcloud/vue-richtext` as it brought the same dependencies 🤔

nickvergessen:
> But I wonder how that worked anyhow, since the dependencies are required e.g. here

the package was excluded:
https://github.com/nextcloud/spreed/blob/be3a4f2f0fd51fc62454165ba6ce4a9d33ce1e0a/webpack.config.js#L27

And mainly for that very same reason.

raimund-schluessler:
So I adjusted the list of packages to transform for jest in 55932025ffbe5814fc2fef6fbcd29f6268ada470. It is quite extensive, and might change with a minor update of nextcloud/vue, but it works for now and doesn't require to transform all packages.

Antreesy (approved):
+ Jest is green;
+ `Message` and `Quote` rendered properly, import/order could be linted later;
+ We can close #8942 as duplicate

julien-nc:
@raimund-schluessler @Antreesy Thanks! :blue_heart:

PR #8905, opened 2023-03-01
Update breakout rooms actions 

marcoambrosini:
Screenshot 2023-03-01 at 17 29 55


nickvergessen:
After switching to a room and then returning to the room all content of the breakout room tab is gone?

nickvergessen (approved):
Fixes the moderator side of things

PR #8907, opened 2023-03-01
Document occ commands

### LINKED ISSUE #8406: Document occ commands
Especially the monitoring ones as there is no UI replacement.

vitormattos:
### 🚧 TODO

- [ ] close #8406

### 🏁 Checklist


docs/occ.md
```diff
@@ -0,0 +1,938 @@
+# Talk occ commands
+
+ * [talk:command:add](#talkcommandadd)
+ * [talk:command:add-samples](#talkcommandadd-samples)
+ * [talk:command:delete](#talkcommanddelete)
+ * [talk:command:list](#talkcommandlist)
+ * [talk:command:update](#talkcommandupdate)
+ * [talk:developer:update-docs](#talkdeveloperupdate-docs)
```
nickvergessen, line 8: Please skip this from the docs

tests/php/Command/DocumentationTest.php
```diff
@@ -0,0 +1,56 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+namespace OCA\Talk\Tests\php\Command;
+
+use OCP\App\IAppManager;
+use Test\TestCase;
+
+final class DocumentationTest extends TestCase {
+	public function testCommandIsDocumented(): void {
```
nickvergessen, line 29: Let's just use a github action which runs the command you added and checks that the diff is empty?

docs/occ.md
```diff
@@ -0,0 +1,938 @@
...
+* Is array: no
+* Default: `NULL`
+
+#### `enabled`
+
+Who can use this command: 0 - Disabled, 1 - Moderators, 2 - Users, 3 - Guests
+
+* Is required: yes
+* Is array: no
+* Default: `NULL`
```
nickvergessen, line 78: I think the layout should be a table and not a h2-h4 list.
Let's have a table for the arguments and one for the options similar to the output of `--help`
Otherwise every command is more than a 4k screen in size and that's not helpful anymore

docs/index.md
```diff
@@ -32,3 +32,4 @@
 ## Other
 
 * [Chat commands](commands.md)
+* [Occ commands](occ.md)
```
nickvergessen, line 35: You need to add it here as well:
https://github.com/nextcloud/spreed/blob/master/mkdocs.yml#L37

```
INFO     -  The following pages exist in the docs directory, but are not included in the "nav" configuration:
              - occ.md
```

nickvergessen (changes requested):
As mentioned we don't need the unit test, Just add a github action which executes the command and compares the result similar to the .github/workflow/node.yml

docs/occ.md
vitormattos: >  Otherwise every command is more than a 4k screen in size and that's not helpful anymore

I will check what I can do to change the layout

vitormattos: @nickvergessen look this:

Screenshot_20230302_124852

vitormattos: I just didn't format the markdown with the spaces because this will be a complex logic or, maybe, will require a new composer dependency to do this.

nickvergessen: Looks much better!

tests/php/Command/DocumentationTest.php
vitormattos: By this way we only will identify that need to update the documentation when push and run the CI.

I updated the unit test to verify if the documentation is up to date and not only to check if the .md file have the command name.

docs/occ.md
vitormattos: Summary removed

docs/occ.md
```diff
@@ -0,0 +1,388 @@
...
+| Arguments | Description | Is required | Is array | Default |
+|---|---|---|---|---|
+| `command-id` |  | yes | no | `NULL` |
+| `cmd` | The command as used in the chat "/help" => "help" | yes | no | `NULL` |
+| `name` | Name of the user posting the response | yes | no | `NULL` |
+| `script` | Script to execute (Must be using absolute paths only) | yes | no | `NULL` |
+| `response` | Who should see the response: 0 - No one, 1 - User, 2 - All | yes | no | `NULL` |
+| `enabled` | Who can use this command: 0 - Disabled, 1 - Moderators, 2 - Users, 3 - Guests | yes | no | `NULL` |
+
+## talk:developer:update-docs
```
nickvergessen, line 72: This should be skipped from the docs

Also it's not clean for me:
```
diff --git a/docs/occ.md b/docs/occ.md
index dee27d247..b1450ef4b 100644
--- a/docs/occ.md
+++ b/docs/occ.md
@@ -77,6 +77,21 @@ Update documentation of commands
 
 * `talk:developer:update-docs`
 
+| Arguments | Description | Is required | Is array | Default |
+|---|---|---|---|---|
+| `command` | The command to execute | yes | no | `NULL` |
+
+| Options | Accept value | Is value required | Is multiple | Default |
+|---|---|---|---|---|
+| `--help|-h` | Display this help message | no | no | no | false` |
+| `--quiet|-q` | Do not output any message | no | no | no | false` |
+| `--verbose|-v|-vv|-vvv` | Increase the verbosity of messages: 1 for normal output, 2 for more verbose output and 3 for debug | no | no | no | false` |
+| `--version|-V` | Display this application version | no | no | no | false` |
+| `--ansi` | Force ANSI output | no | no | no | false` |
+| `--no-ansi` | Disable ANSI output | no | no | no | false` |
+| `--no-interaction|-n` | Do not ask any interactive question | no | no | no | false` |
+| `--no-warnings` | Skip global warnings, show command output only | no | no | no | false` |
+
 ## talk:monitor:calls
 
 Prints a list with conversations that have an active call as well as their participant count
```

nickvergessen (approved):

PR #8949, opened 2023-03-03
feat(NewMessageUploadEditor): multiple files sharing support 🖼️+🖼️

Antreesy:
### ☑️ Resolves

* Support sending and receiving a group of shared files in a single message
* Fix #…

| 🏚️ Before | 🏡 After |
|------------|----------|
| B          | A        |

### 🚧 Tasks

- [ ] Mobile clients support:
  - As per backend setup, if message with file share contains a caption, it will replace the default message text (`{file}`). So we need to manually add references to the message text (see Message.vue)
  - When multiple files are loaded, they'll get indexed references - `file-1`, `file-2` (and `file-missing-3` for files deleted afterwards)
  - When sending multiple files, we still need to upload it via `apps/files_sharing/api/v1/shares` one by one, but metadata object should contain 'noMessage: true' to nont generate a separate message for each file. Instead, we're sending a new request to `apps/spreed/api/v1/chat/{token}/share-files` with caption and share ids of all the uploaded files
- [x] Autotests
- [ ] Capabilities
- [ ] Silent sending - check if possible
- [ ] User bubbles - don't render for prefilled messages
- [ ] Manual testing
- [ ] Visual check
- [ ] Code review

### 🏁 Checklist

### DIFF
--- a/src/services/filesSharingServices.js
+++ b/src/services/filesSharingServices.js
@@ -24,7 +24,7 @@ import { showError } from '@nextcloud/dialogs'
 import { generateOcsUrl } from '@nextcloud/router'
 
 /**
- * Appends a file as a message to the messagelist.
+ * Appends a file as a message to the message list.
  *
  * @param {string} path The file path from the user's root directory
  * @param {string} token The conversation's token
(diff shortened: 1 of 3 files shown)


src/components/UploadEditor.vue
```diff
@@ -64,6 +64,8 @@
 				<NcButton type="tertiary" @click="handleDismiss">
 					{{ t('spreed', 'Dismiss') }}
 				</NcButton>
+				<!-- FIXME replace with NcRichContenteditable -->
+				<input v-model="mediaCaption">
```
nickvergessen, line 68: Should use NcContentEditable aka the NewMessageForm similar to 
https://github.com/nextcloud/spreed/blob/master/src/components/BreakoutRoomsEditor/SendMessageDialog.vue#L28

nickvergessen (changes requested):

jancborchardt (commented):
Looks super nice generally! Some feedback:
- [ ] In the modal, the "Send" button is now not primary anymore, would be cool if the "Paper plane" icon button would be a primary button icon+text
- [x] Currently when composing, the images are above the text, but when sending the text comes first. In other messengers, when you send a message with a picture, usually the picture is above the text too.
- [ ] Sort of related to the previous point, maybe it would be good to put the photos next to each other or tile them somehow else when there are multiple ones. Otherwise the text and photos are very far apart and the message also takes a lot of space if you send a message with 10 pictures

Antreesy:
Rebased onto master, dropped all irrelevant/outdated/conflicted code

PR #8978, opened 2023-03-07
fix(a11y): Add aria-label to button with only numeric label

nickvergessen:
Before there is a plain text button with only a numeric value on it without any further clue for screenreader users.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Bildschirmfoto vom 2023-03-07 06-45-55 | grafik

### 🏁 Checklist


Antreesy (approved):
Works properly

PR #9051, opened 2023-03-13
feat: add utils for generating absolute links and copy a chat link

ShGKme:
There are many places with copying a link to a conversation. Currently, it is mostly duplicating of this:
```js
computed: {
	linkToConversation() {
		return window.location.protocol + '//' + window.location.host + generateUrl('/call/' + this.token)
	},
},

methods: {
	async copyLinkToConversation() {
		try {
			await navigator.clipboard.writeText(this.linkToConversation)
			showSuccess(t('spreed', 'Conversation link copied to clipboard'))
		} catch (error) {
			showError(t('spreed', 'The link could not be copied'))
		}
	},
},
```

Also, the Confirmation step on conversation creation shows the result of copying in plain text for some seconds instead of a toast message dialog, as it is in all other places.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
 | 

### 🚧 TODO

- [x] Add `generateAbsoluteUrl` util wrapper under `geenrateUrl` from `@nextcloud/router`
  - [ ] Add `generateAbsoluteUrl` directly to `@nextcloud/router`
- [x] Add `generateConversationLink` util
- [x] Add `copyConversationLink` method
- [x] Removed unused methods
- [x] Update `Confirmation.vue` to show the result of copying in a dialog

### DRAFT

This PR is a graft waiting for https://github.com/nextcloud/spreed/pull/9050: `generateAbsoluteUrl` is different on the desktop

### 🏁 Checklist


src/components/LeftSidebar/NewGroupConversation/NewGroupConversation.vue
```diff
@@ -148,6 +146,7 @@ import {
 } from '../../../services/conversationsService.js'
 import { EventBus } from '../../../services/EventBus.js'
 import { addParticipant } from '../../../services/participantsService.js'
+import { generateFullConversationLink } from '../../../services/urlService.js'
```
nickvergessen, line 149: 'generateFullConversationLink' is defined but never used

src/services/urlService.js
```diff
@@ -0,0 +1,76 @@
...
+ * @param {object} [params] parameters to be replaced into the address
+ * @param {import('@nextcloud/router').UrlOptions} [options] options for the parameter replacement
+ * @param {boolean} options.noRewrite True if you want to force index.php being added
+ * @param {boolean} options.escape Set to false if parameters should not be URL encoded (default true)
+ * @return {string} Full absolute URL
+ */
+export function generateAbsoluteUrl(url, params, options) {
+	// TODO: add this function to @nextcloud/router?
+	const fullPath = generateUrl(url, params, options)
+	if (!IS_DESKTOP) {
```
nickvergessen, line 39: 'IS_DESKTOP' is not defined

src/services/urlService.js
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/**
+ * Try to copy conversation link to a clipboard and display the result with dialogs
+ *
+ * @param {string} token - conversation token
+ * @param {string} [messageId] - messageId for message in conversation link
+ * @return {Promise<void>}
+ */
+export async function copyLinkToConversation(token, messageId) {
```
nickvergessen, line 69: I first was confused why you copy a link "into" a conversation, so maybe a better name would be:

```suggestion
export async function copyConversationLinkToClipboard(token, messageId) {
```

nickvergessen (changes requested):

src/services/urlService.js
ShGKme: Sorry, I miss-clicked, this PR should be a DRAFT before https://github.com/nextcloud/spreed/pull/9050

src/services/urlService.js
ShGKme: I kept the original method name.

Let's rename

src/components/LeftSidebar/NewGroupConversation/NewGroupConversation.vue
ShGKme: 🙈

ShGKme:
- Rebased after merge https://github.com/nextcloud/spreed/pull/9050
- Lint
- Rename `copyLinkToConversation` -> `copyConversationLinkToClipboard` according to the review

src/components/LeftSidebar/NewGroupConversation/Confirmation/Confirmation.vue
```diff
@@ -83,42 +85,11 @@ export default {
...
-				this.showConfirmationText = false
-			}, 800)
-		},
-		onError() {
-			this.confirmationText = t('spreed', 'Error')
-			this.showConfirmationText = true
-			setTimeout(function() {
-				this.showConfirmationText = false
-			}, 800)
+		onClickCopyLink() {
```
Antreesy, line 91: Shouldn't this method be asymc as well as the others?

ShGKme: `async/await` is not required here because we never need to wait copy.

For consistency I removed `await` everywhere for this method. I preferred removing await everywhere than adding here because:
- As we transpile to old ES5, async/await adds really much code after the transpiling
- We always can do copying "in parallel" while await attracts sequential execution

Antreesy (approved):
Nice and clean done, from my POV :tada:

nickvergessen (approved):

PR #9059, opened 2023-03-14
Feature/8334/conversation avatar frontend

### LINKED ISSUE #8334: 🖼️ Conversation avatars frontend
See meta ticket at https://github.com/nextcloud/spreed/issues/927

- [ ] Cropper like in user avatar
- [ ] Option to delete the avatar
- [ ] Use the new avatar URL in all places

marcoambrosini:
fix #8334 

https://user-images.githubusercontent.com/26852655/225372550-e83c6d05-9537-4008-8371-35054d3e36da.mov

- [x] Adapt server component to our use-case
- [x] Add services 
- [x] Add store method
- [x] Update conversationIcon component so that the picture is displayed everywhere


GretaD (approved):
Screenshot from 2023-03-15 21-07-31

src/components/ConversationIcon.vue
```diff
@@ -18,15 +18,20 @@
...
-	<div class="conversation-icon"
+	<div ref="conversation-icon"
+		class="conversation-icon"
+		:style="iconStyle"
 		:class="{'offline': offline}">
 		<div v-if="iconClass"
 			class="avatar icon"
+			:style="iconStyle"
 			:class="iconClass" />
+		<NcAvatar v-else-if="hasAvatar"
```
Antreesy, line 30: Maybe `hasAvatar` should have a priority here, and if it is falsy, check for `iconClass`?
Or it matters when we're waiting for picture to load?

src/components/ConversationSettings/ConversationAvatarEditor.vue
```diff
@@ -0,0 +1,310 @@
...
+
+<script>
+import VueCropper from 'vue-cropperjs'
+
+import axios from '@nextcloud/axios'
+import { getFilePickerBuilder, showError } from '@nextcloud/dialogs'
+import { generateUrl } from '@nextcloud/router'
+
+// import NcAvatar from '@nextcloud/vue/dist/Components/NcAvatar.js'
+import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
```
Antreesy, line 89: Please lint imports

src/components/ConversationSettings/ConversationAvatarEditor.vue
```diff
@@ -0,0 +1,310 @@
...
+		height: 300px;
+		overflow: hidden;
+
+		&-buttons {
+			width: 100%;
+			display: flex;
+			justify-content: space-between;
+		}
+
+		&::v-deep .cropper-view-box {
```
Antreesy, line 300:
```suggestion
		& :deep(.cropper-view-box) {
```

`v-deep` is deprecated

src/components/ConversationSettings/ConversationAvatarEditor.vue
```diff
@@ -0,0 +1,310 @@
...
+
+		async openFilePicker() {
+			const path = await picker.pick()
+			this.loading = true
+			try {
+				const { data } = await axios.post(generateUrl('/avatar'), { path })
+				if (data.status === 'success') {
+					this.loading = false
+				} else if (data.data === 'notsquare') {
+					const tempAvatar = generateUrl('/avatar/tmp') + '?requesttoken=' + encodeURIComponent(OC.requestToken) + '#' + Math.floor(Math.random() * 1000)
```
ShGKme, line 195: I may be wrong, but isn't `OC.requestToken` deprecated in favor of `getRequestToken` from [@nextcloud/auth](https://github.com/nextcloud/nextcloud-auth/)?

src/services/conversationsService.js
```diff
@@ -409,6 +409,14 @@ const validatePassword = async (password) => {
 	})
 }
 
+const setConversationPicture = async function(token, file) {
+	return axios.post(generateOcsUrl('apps/spreed/api/v1/room/{token}/avatar', { token }), file,)
+}
+
+const deleteConversationPicture = async function(token) {
+	return axios.delete(generateOcsUrl('apps/spreed/api/v1/room/{token}/avatar', { token }))
+}
```
ShGKme, line 418: In most of places it is called `avatar` in the code. Should we rename to `avatar` here too?

src/components/ConversationSettings/ConversationAvatarEditor.vue
```diff
@@ -0,0 +1,310 @@
...
+		conversation: {
+			type: Object,
+			required: true,
+		},
+	},
+
+	data() {
+		return {
+			// Todo: get real value here
+			showCropper: false,
```
ShGKme, line 133: What do you mean by real value here?

src/components/ConversationSettings/ConversationAvatarEditor.vue
```diff
@@ -0,0 +1,310 @@
...
+			return `account-property-${this.conversation.displayName}`
+		},
+
+		hasAvatar() {
+			return !!this.conversation.avatarVersion
+		},
+
+		fileTypeWarning() {
+			return t('spreed', 'The file must be a PNG or JPG')
+		},
```
ShGKme, line 160: I'd put this text also directly into the template. All other texts are in the template, and this text doesn't require any reactive computing to be `computed`

src/components/ConversationSettings/ConversationAvatarEditor.vue
```diff
@@ -0,0 +1,310 @@
...
+				<NcButton v-if="hasAvatar"
+					:aria-label="t('settings', 'Remove profile picture')"
+					@click="removeAvatar">
+					<template #icon>
+						<Delete :size="20" />
+					</template>
+				</NcButton>
+			</div>
+			<span>{{ fileTypeWarning }}</span>
+			<input :id="inputId"
```
ShGKme, line 56: Should have a class for ...

src/components/ConversationSettings/ConversationAvatarEditor.vue
```diff
@@ -0,0 +1,310 @@
...
+		display: flex;
+		flex-direction: column;
+		justify-content: center;
+		align-items: center;
+		gap: 16px 0;
+		width: 300px;
+
+		span {
+			color: var(--color-text-lighter);
+		}
```
ShGKme, line 273: ... these styles?

src/components/ConversationSettings/ConversationAvatarEditor.vue
```diff
@@ -0,0 +1,310 @@
...
+				viewMode: 1,
+				guides: false,
+				center: false,
+				highlight: false,
+				autoCropArea: 1,
+				minContainerWidth: 300,
+				minContainerHeight: 300,
+			},
+		}
+	},
```
ShGKme, line 147: `validMimeTypes` and `cropperOptions` are not a part of the component instance's reactive state (data). They never change and are actually constants. 

I know that it is a very common practice to add constants to `data` to add them into component instance properties, but in my opinion, it is not good practice because:
- It makes them feel like a changeable state, not a constant
- Vue (especially Vue 2) traverses the object, redefines all the properties, and creates an observer for watching, `Dep`s instances for each property, making them reactive. It's a bit worse in performance.

```html
<template>
  <!-- In template we can take options from $options -->
  <VueCropper v-bind="$options.CROPPER_OPTIONS" />
</template>

<script>
const VALID_MIME_TYPES = ['image/png', 'image/jpeg']
const CROPPER_OPTIONS = { /* ... */ }

export default {
	name: 'ConversationAvatarEditor',
	
	VALID_MIME_TYPES,
	CROPPER_OPTIONS,
	
	methods: {
	  foo(type) {
	    // In script we can use constants as it is
	    VALID_MIME_TYPES.includes(type)
	  }
	}
}
</script>
```

But as it is a common practice in Options API, I'm fine with any option.

ShGKme (commented):
Do we want to use the avatar also in mentions?

image

vitormattos (approved):
About API side, OK.
I ran the code at my environment to test the UI and worked fine for me.

ShGKme:
Is it expected?

image

nickvergessen:
That's an issue with the backend I guess. It returns all icons wrongly.

src/components/ConversationSettings/ConversationAvatarEditor.vue
marcoambrosini: What do you mean here @ShGKme ?

Antreesy:
Please take a look at my suggestions, there is also an opportunity for improvements:
https://github.com/nextcloud/spreed/commit/d38c8a91fb0093dce6f44849aa67c25f1f49d8c7

And a bugfix for 1-1 conversations and non-moderators in conversations:
https://github.com/nextcloud/spreed/commit/667b0cb8f17c225436bcf588dfac228f501d7bbd

src/components/ConversationIcon.vue
```diff
@@ -73,14 +82,17 @@ export default {
...
 			default: true,
 		},
+
 		hideCall: {
 			type: Boolean,
 			default: true,
 		},
-		disableMenu: {
+
+		 disableMenu: {
```
nickvergessen, line 91:
```suggestion
		disableMenu: {
```

src/store/conversationsStore.js
```diff
@@ -710,6 +712,31 @@ const actions = {
...
+		}
+
+	},
+
+	async deleteConversationAvatarAction(context, { token, file }) {
+		try {
+			const response = await deleteConversationAvatar(token, file)
+			const conversation = response.data.ocs.data
+			context.commit('addConversation', conversation)
+			showSuccess(t('spreed', 'You\'ve deleted the conversation picture.'))
```
nickvergessen, line 734:
```suggestion
			showSuccess(t('spreed', 'Conversation picture deleted'))
```

src/store/conversationsStore.js
```diff
@@ -710,6 +712,31 @@ const actions = {
 		}
 		context.commit('setCallRecording', { token, callRecording: CALL.RECORDING.OFF })
 	},
+
+	async setConversationAvatarAction(context, { token, file }) {
+		try {
+			const response = await setConversationAvatar(token, file)
+			const conversation = response.data.ocs.data
+			context.commit('addConversation', conversation)
+			showSuccess(t('spreed', 'You\'ve set the conversation picture.'))
```
nickvergessen, line 721:
```suggestion
			showSuccess(t('spreed', 'Conversation picture set'))
```

nickvergessen (commented):
Should improve translations by removing `'` and is more inline with other things.

src/components/ConversationSettings/ConversationAvatarEditor.vue
ShGKme: Second part of a question is in the next comment 😀

I meant, shouldn't we add a class for styling this span?

src/components/ConversationIcon.vue
```diff
@@ -139,21 +171,42 @@ export default {
...
+			return generateOcsUrl('apps/spreed/api/v1/room/{token}/avatar?v={avatarVersion}', {
+				token: this.item.token,
+				avatarVersion: this.item.avatarVersion,
+			})
 		},
 	},
+
+	mounted() {
+		// Get the size of the parent once the component is mounted
+		this.parentElement = this.$refs?.['conversation-icon']?.parentElement
```
Antreesy, line 203:
```suggestion
		this.parentElement = this.$refs.['conversation-icon'].parentElement
```

Antreesy (approved):
We missed this change somewhere, but should works properly in both cases

src/components/ConversationIcon.vue
```diff
@@ -65,6 +65,19 @@ import NcAvatar from '@nextcloud/vue/dist/Components/NcAvatar.js'
...
+	'share:password': 'icon-password',
+	emails: 'icon-mail',
+}
+
+const ICON_CLASS_BY_TYPE = {
+	[CONVERSATION.TYPE.CHANGELOG]: 'icon-changelog',
+	[CONVERSATION.TYPE.ONE_TO_ONE_FORMER]: 'icon-user',
+	[CONVERSATION.TYPE.GROUP]: 'icon-contacts',
+	[CONVERSATION.TYPE.PUBLIC]: 'icon-public',
+}
```
ShGKme, line 79: 😊

ShGKme (approved):

src/components/ConversationIcon.vue
nickvergessen: As mentioned in several of the backend PRs I don't think we want this.
All rooms should just return their avatar properly on the avatar endpoint, no matter if it is a custom uploaded one or a predefined fallback icon. This way the life of clients and frontend should be much easier as they simply use the avatar endpoint all the time, no more long list of if-switches to find the right class.

So drop this fallback behaviour from here, and then we fix the avatar endpoint to return proper icons in all the cases?

nickvergessen:
Needs a rebase after https://github.com/nextcloud/spreed/pull/9214

nickvergessen:
Will rebase later and fix the icon positions and colors

szaimen:
🎉🎉🎉

PR #9063, opened 2023-03-15
fix(avatar): Return the conversation data and fix the documentation

nickvergessen:
- [x] Return the full conversation object instead of (undocumented) custom data
- [x] Fix documentation missing headers

### 🏁 Checklist

### DIFF
--- a/docs/conversation.md
+++ b/docs/conversation.md
@@ -424,7 +424,7 @@ Get all (for moderators and in case of "free selection") or the assigned breakou
         + `403 Forbidden` When the current user is not a moderator/owner or the conversation is not a public conversation
         + `404 Not Found` When the conversation could not be found for the participant
 
-## Handle the avatar of conversation
+## Set conversations avatar
 
 * ⚠️ Preview - Might be modified before the capability is added
 * Required capability: `avatar`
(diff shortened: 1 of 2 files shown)


marcoambrosini (approved):
🙏

docs/conversation.md
```diff
@@ -453,6 +458,11 @@ Get all (for moderators and in case of "free selection") or the assigned breakou
...
         + `404 Not Found` When the conversation could not be found for the participant
 
+	- Data: See array definition in `Get user´s conversations`
+
+## Get conversations avatar (binary)
+
+* ⚠️ Preview - Might be modified before the capability is added
 * Required capability: `avatar`
 * Method: `GET`
 * Endpoint: `/room/{token}/avatar`
```
ShGKme, line 468: Shouldn't we mention `?v=<avatarVersion>` query param here for HTTP Cache?

nickvergessen: Query parameters are not part of the API, in fact the API does not eve handle the parameter, so you can not get an older version by specifying an older parameter or something a like, so I'd like not to add it to the docs

vitormattos (approved):

docs/conversation.md
ShGKme: It is not a part of API, but developers should use it. And I'm sure somebody will miss it.

Maybe we can add a note?

In the previous version, an example of the get image UTL (in POST avatar) contained `?v=<avatarVersion>`.

PR #9066, opened 2023-03-15
Allow to receive messages without marking notifications as unread

SystemKeeper:
Currently when retrieving new messages, notifications of the spreed app are marked as read by default:
https://github.com/nextcloud/spreed/blob/049c56f4d85cd35bbba5079061cce0f0f9fda513/lib/Chat/ChatManager.php#L606-L608
This is usually fine, but when we want to retrieve messages in the background on the mobile clients (to have the messages ready when we open the app again).
This PR adds a new parameter `markNotificationsAsRead` which allows to explicitly state that we do not want to mark notifications as read when retrieving messages.

### 🏁 Checklist


docs/chat.md
```diff
@@ -14,15 +14,17 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`: since Nextcloud 13
...
+| `setReadMarker`           | int  | `1` to automatically set the read timer after fetching the messages, use `0` when your client calls `Mark chat as read` manually. (Default: `1`)                                                                                        |
+| `includeLastKnown`        | int  | `1` to include the last known message as well (Default: `0`)                                                                                                                                                                            |
+| `noStatusUpdate`          | int  | Whether the "online" user status of the current user should be "kept-alive" (`1`) or not (`0`) (defaults to `0`)                                                                                                                        |
+| `markNotificationsAsRead` | int  | `0` to not mark notifications as read (Default: `1`, only available with `chat-read-status` capability                                                                                                                                                         |
```
nickvergessen, line 27:
```suggestion
| `markNotificationsAsRead` | int  | `0` to not mark notifications as read (Default: `1`, only available with `chat-read-status` capability)                                                                                                                                                        |
```

nickvergessen (changes requested):

lib/Capabilities.php
```diff
@@ -116,6 +116,7 @@ public function getCapabilities(): array {
 				'recording-v1',
 				'chat-get-context',
 				'single-conversation-status',
+				'chat-keep-notifications'
```
vitormattos, line 119: Using a comma at the last element of array every when is indented, when add another row don't will touch the previous row.

By example, you didn't touch any row, only added a new row. This is a good point when we need to use `git blame`.
```suggestion
				'chat-keep-notifications',
```

vitormattos (changes requested):

SystemKeeper:
Both comments addressed, rebased and squashed

vitormattos (approved):

docs/chat.md
```diff
@@ -14,15 +14,17 @@ Base endpoint is: `/ocs/v2.php/apps/spreed/api/v1`: since Nextcloud 13
...
+| `setReadMarker`           | int  | `1` to automatically set the read timer after fetching the messages, use `0` when your client calls `Mark chat as read` manually. (Default: `1`)                                                                                        |
+| `includeLastKnown`        | int  | `1` to include the last known message as well (Default: `0`)                                                                                                                                                                            |
+| `noStatusUpdate`          | int  | Whether the "online" user status of the current user should be "kept-alive" (`1`) or not (`0`) (defaults to `0`)                                                                                                                        |
+| `markNotificationsAsRead` | int  | `0` to not mark notifications as read (Default: `1`, only available with `chat-read-status` capability)                                                                                                                                 |
```
nickvergessen, line 27: Should be `chat-keep-notifications` ?

PR #9072, opened 2023-03-16
Populate icon-url on call rich objects

### LINKED ISSUE #8389: Document icon-url on call rich objects
https://github.com/nextcloud/spreed/pull/8333 introduced an optional `icon-url` for the call objects in rich messages.
This attribute is however not documented in https://github.com/nextcloud/server/blob/d18a293ee50f1a17942391b2031f39fed6ff2eae/lib/public/RichObjectStrings/Definitions.php#L166-L196

vitormattos:
https://github.com/nextcloud/spreed/pull/8333 introduced an optional `icon-url` for the call objects in rich messages.

Fix #8389

### 🏁 Checklist

### DIFF
--- a/lib/Chat/Parser/UserMention.php
+++ b/lib/Chat/Parser/UserMention.php
@@ -29,6 +29,7 @@ use OCA\Talk\GuestManager;
 use OCA\Talk\Model\Attendee;
 use OCA\Talk\Model\Message;
 use OCA\Talk\Room;
+use OCA\Talk\Service\AvatarService;
 use OCA\Talk\Service\ParticipantService;
 use OCP\Comments\ICommentsManager;
 use OCP\IGroup;
@@ -50,12 +51,14 @@ class UserMention {
 	protected IGroupManager $groupManager;
 	protected GuestManager $guestManager;
 	protected ParticipantService $participantService;
+	protected AvatarService $avatarService;
 	protected IL10N $l;
 
 	public function __construct(ICommentsManager $commentsManager,
 								IUserManager $userManager,
 								IGroupManager $groupManager,
 								GuestManager $guestManager,
+								AvatarService $avatarService,
 								ParticipantService $participantService,
 								IL10N $l) {
 		$this->commentsManager = $commentsManager;
@@ -63,6 +66,7 @@ class UserMention {
 		$this->groupManager = $groupManager;
 		$this->guestManager = $guestManager;
 		$this->participantService = $participantService;
+		$this->avatarService = $avatarService;
 		$this->l = $l;
 	}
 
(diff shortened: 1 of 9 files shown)


lib/Chat/Parser/UserMention.php
```diff
@@ -138,6 +138,7 @@ public function parseMessage(Message $chatMessage): void {
 					'id' => $chatMessage->getRoom()->getToken(),
 					'name' => $chatMessage->getRoom()->getDisplayName($userId),
 					'call-type' => $this->getRoomType($chatMessage->getRoom()),
+					'icon-url' => $chatMessage->getRoom()->getAvatar(),
```
nickvergessen, line 141: you need to call `AvatarService::getAvatarUrl()` ? Otherwise you just expose the "version" not a ready-to-use URL

vitormattos: Good observation. Fixing and checking if is possible cover this with tests

tests/php/Notification/NotifierTest.php
```diff
@@ -208,7 +213,8 @@ public function testPrepareOne2One(string $uid, string $displayName, string $par
 					'type' => 'call',
 					'id' => 1234,
 					'name' => $displayName,
-					'call-type' => 'one2one'
+					'call-type' => 'one2one',
+					'icon-url' => '',
```
nickvergessen, line 217: Please instead fix the test so it expects that the result of the avatar service is returned here

nickvergessen (changes requested):
See both comments

tests/integration/features/chat-2/rich-object-share.feature
```diff
@@ -67,7 +67,7 @@ Feature: chat-2/rich-object-share
     When user "participant1" shares rich-object "call" "R4nd0mT0k3n" '{"name":"Another room","call-type":"group"}' to room "public room" with 201 (v1)
     Then user "participant1" sees the following shared other in room "public room" with 200
       | room        | actorType | actorId      | actorDisplayName         | message  | messageParameters |
-      | public room | users     | participant1 | participant1-displayname | {object} | {"actor":{"type":"user","id":"participant1","name":"participant1-displayname"},"object":{"name":"Another room","call-type":"group","type":"call","id":"R4nd0mT0k3n"}} |
+      | public room | users     | participant1 | participant1-displayname | {object} | {"actor":{"type":"user","id":"participant1","name":"participant1-displayname"},"object":{"name":"Another room","call-type":"group","type":"call","id":"R4nd0mT0k3n","icon-url":""}} |
```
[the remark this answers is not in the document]
vitormattos: Did improvements at this point.

tests/php/Notification/NotifierTest.php
vitormattos: I  looked the unit test and I think that the integration tests at this case is best because to reproduce this at unit tests we will need to mock a lot of methods, including the string of avatar URL.

tests/integration/features/chat-2/rich-object-share.feature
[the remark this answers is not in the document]
nickvergessen: > We can have, but only is possible after avatar change

Why? the URL should be set and return an image at all time.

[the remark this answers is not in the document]
nickvergessen: Reverted it and the test still works

nickvergessen (approved):

PR #9075, opened 2023-03-16
Fix order of message of signaling server to be returned to client

vitormattos:
### 🚧 TODO

- [ ] Close https://github.com/nextcloud/spreed/issues/7863

### 🏁 Checklist

### DIFF
--- a/lib/Signaling/Messages.php
+++ b/lib/Signaling/Messages.php
@@ -129,7 +129,8 @@ class Messages {
 		$query->select('*')
 			->from('talk_internalsignaling')
 			->where($query->expr()->eq('recipient', $query->createNamedParameter($sessionId)))
-			->andWhere($query->expr()->lte('timestamp', $query->createNamedParameter($time)));
+			->andWhere($query->expr()->lte('timestamp', $query->createNamedParameter($time)))
+			->orderBy('id', 'ASC');
 
 		$delete = $this->db->getQueryBuilder();
 		$delete->delete('talk_internalsignaling')


lib/Signaling/Messages.php
```diff
@@ -129,7 +129,8 @@ public function getAndDeleteMessages(string $sessionId): array {
 		$query->select('*')
 			->from('talk_internalsignaling')
 			->where($query->expr()->eq('recipient', $query->createNamedParameter($sessionId)))
-			->andWhere($query->expr()->lte('timestamp', $query->createNamedParameter($time)));
+			->andWhere($query->expr()->lte('timestamp', $query->createNamedParameter($time)))
+			->orderBy('id');
```
nickvergessen, line 133:
```suggestion
			->orderBy('id', 'ASC');
```

vitormattos: ASC is the default to Doctrine DBAL:
https://github.com/nextcloud/3rdparty/blob/master/doctrine/dbal/src/Query/QueryBuilder.php#L1243
Follow as default at the latest version of DBAL:
https://github.com/doctrine/dbal/blob/3.6.x/src/Query/QueryBuilder.php#L1277

And also to
MySQL:
https://dev.mysql.com/doc/refman/8.0/en/sorting-rows.html
PostgreSQL:
https://www.postgresql.org/docs/current/queries-order.html
SQLite:
https://www.sqlitetutorial.net/sqlite-order-by/
And Oracle:
https://www.oracletutorial.com/oracle-basics/oracle-order-by/

nickvergessen: Explicit is better :)

vitormattos: https://github.com/nextcloud/spreed/issues/9077

nickvergessen (approved):

PR #9110, opened 2023-03-20
Refactor `VideoBottomBar.vue` component

Antreesy:
### 🖼️ Screenshots

Grid view | Speaker view
---|---
image | image

| Tests covering #9202  |
| --- |
| image |

- [x] Move checks to computed properties
- [x] Rename classes
- [x] Remove redundant wrappers,  classes, properties
- [x] Improve component `template` readability
- [x] Component appearance is untouched
- [x] Cover with unit tests

### 🚧 TODO

- [ ] Code review :warning: It will be easier to analyze commits one by one
- [ ] Visual check

### 🏁 Checklist

### DIFF
--- a/src/components/CallView/shared/VideoBottomBar.spec.js
+++ b/src/components/CallView/shared/VideoBottomBar.spec.js
@@ -163,7 +163,7 @@ describe('VideoBottomBar.vue', () => {
 					propsData: componentProps,
 				})
 
-				const participantName = wrapper.findComponent('.participant-name')
+				const participantName = wrapper.find('.participant-name')
 				expect(participantName.exists()).toBeFalsy()
 
 				const indicators = wrapper.findAllComponents(NcButton)
@@ -179,7 +179,7 @@ describe('VideoBottomBar.vue', () => {
 					propsData: componentProps,
 				})
 
-				const participantName = wrapper.findComponent('.participant-name')
+				const participantName = wrapper.find('.participant-name')
 				expect(participantName.isVisible()).toBeTruthy()
 				expect(participantName.text()).toBe(PARTICIPANT_NAME)
 			})
@@ -192,7 +192,7 @@ describe('VideoBottomBar.vue', () => {
 					propsData: componentProps,
 				})
 
-				const participantName = wrapper.findComponent('.participant-name')
+				const participantName = wrapper.find('.participant-name')
 				expect(participantName.isVisible()).toBeFalsy()
 			})
 		})
--- a/src/components/CallView/shared/VideoBottomBar.spec.js
+++ b/src/components/CallView/shared/VideoBottomBar.spec.js
@@ -2,6 +2,8 @@ import { createLocalVue, shallowMount } from '@vue/test-utils'
 import { cloneDeep } from 'lodash'
 import Vuex, { Store } from 'vuex'
 
+import AlertCircle from 'vue-material-design-icons/AlertCircle.vue'
+import HandBackLeft from 'vue-material-design-icons/HandBackLeft.vue'
 import VideoIcon from 'vue-material-design-icons/Video.vue'
 import VideoOff from 'vue-material-design-icons/VideoOff.vue'
 
@@ -13,6 +15,7 @@ import VideoBottomBar from './VideoBottomBar.vue'
 
 import { CONVERSATION, PARTICIPANT } from '../../../constants.js'
 import storeConfig from '../../../store/storeConfig.js'
+import { findNcButton } from '../../../test-helpers.js'
 import { ConnectionState } from '../../../utils/webrtc/models/CallParticipantModel.js'
 
 jest.mock('@nextcloud/event-bus', () => ({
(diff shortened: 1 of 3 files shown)


ShGKme (changes requested):
The component still seems to me to be a "super-component" with many flags, but it's much better with tests and a bit less complex now 🎉

One blocking thing for me: this component now has new props. Are they used and set somewhere?

src/components/CallView/shared/VideoBottomBar.vue
```diff
@@ -179,86 +169,109 @@ export default {
...
+		// The current promoted participant
+		isPromoted: {
+			type: Boolean,
+			default: false,
+		},
+		// Is the current selected participant
+		isSelected: {
+			type: Boolean,
+			default: false,
+		},
```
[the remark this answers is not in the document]
Antreesy: They came before with v-bind, just haven't been defined in the component

src/components/CallView/shared/VideoBottomBar.vue
```diff
@@ -20,87 +21,76 @@
...
 			<transition name="fade">
-				<div v-show="showNameIndicator"
-					class="bottom-bar__nameIndicator"
-					:class="{'bottom-bar__nameIndicator--promoted': boldenNameIndicator}">
+				<div v-show="showParticipantName"
+					class="participant-name"
+					:class="participantNameClass">
 					{{ participantName }}
 				</div>
 			</transition>
```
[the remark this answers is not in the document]
Antreesy: Not sure about this one. Name is separated from other media indicators, and visible on each `mouseover` event inside callView because of `showVideoOverlay`
So i would go with v-show here and v-if in other places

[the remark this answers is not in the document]
ShGKme: If it is mouseover effect, maybe show it with CSS?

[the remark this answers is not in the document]
Antreesy: Not exactly. The timer is defined by Grid.vue, and `showVideoOverlay` changed by that timer function passed as prop 
Grid.vue => Video.vue => VideoBottomBar.vue.
- If we want to test it properly, it requires some integration tests - do we need it here?
- If we want to change that behavior (with CSS), it should be done in Grid.vue - could it break current behavior? Now everything _(visually)_ works as before
- If it triggered quite often during the call - is it better to leave `v-show`? as i know, it just add `display: none` to the DOM element, without mounting/removing it, so it shoud be better for performance?

src/components/CallView/shared/VideoBottomBar.spec.js
```diff
@@ -0,0 +1,561 @@
...
+			? t('spreed', 'Mute')
+			: t('spreed', 'Muted')
+	}
+	const videoIndicatorAriaLabel = function() {
+		return this.sharedData.remoteVideoBlocker?.isVideoEnabled()
+			? t('spreed', 'Disable video')
+			: t('spreed', 'Enable video')
+	}
+	const screenSharingIndicatorAriaLabel = () => t('spreed', 'Show screen')
+	const followingButtonAriaLabel = () => t('spreed', 'Stop following')
```
ShGKme, line 49: With these methods, cannot we just find with the `findNcButton`?

src/test-helpers.js
```diff
@@ -46,7 +46,13 @@ const findNcButton = function(wrapper, text) {
 	return items.at(0)
 }
 
+const findNcButtonByMethod = function(wrapper, method) {
+	const text = method.call(wrapper.props())
+	return findNcButton(wrapper, text)
+}
+
```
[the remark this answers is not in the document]
Antreesy: replaced with updated test-helpers

src/test-helpers.js
```diff
@@ -20,26 +20,40 @@
  *
  */
 
+import Wrapper from '@vue/test-utils'
+
```
ShGKme, line 24: Auto-import from WebStorm?

src/test-helpers.js
```diff
@@ -20,26 +20,40 @@
...
 import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
 
 // helpers
-const findNcActionButton = function(wrapper, text) {
+/**
+ *
+ * @param {Wrapper} wrapper root wrapper to look for NcActionButton
+ * @param {string | Array<string>} text or array of possible texts to look for NcButtons
+ * @return {Wrapper}
+ */
```
ShGKme, line 34:
```suggestion
/**
 *
 * @param {import('@vue/test-utils').Wrapper} wrapper root wrapper to look for NcActionButton
 * @param {string | Array<string>} text or array of possible texts to look for NcButtons
 * @return {import('@vue/test-utils').Wrapper}
 */
```

src/test-helpers.js
```diff
@@ -20,26 +20,40 @@
...
 	return items.at(0)
 }
 
-const findNcButton = function(wrapper, text) {
+/**
+ *
+ * @param {Wrapper} wrapper root wrapper to look for NcButton
+ * @param {string | Array<string>} text or array of possible texts to look for NcButtons
+ * @return {Wrapper}
+ */
```
ShGKme, line 51:
```suggestion
/**
 *
 * @param {import('@vue/test-utils').Wrapper} wrapper root wrapper to look for NcActionButton
 * @param {string | Array<string>} text or array of possible texts to look for NcButtons
 * @return {import('@vue/test-utils').Wrapper}
 */
```

ShGKme, line 51:
```suggestion
/**
 *
 * @param {import('@vue/test-utils').Wrapper} wrapper root wrapper to look for NcButton
 * @param {string | Array<string>} text or array of possible texts to look for NcButtons
 * @return {import('@vue/test-utils').Wrapper}
 */
```

src/test-helpers.js
Antreesy: Thanks

ShGKme (approved):

PR #9175, opened 2023-03-27
User agent for recording backend requests

mishamosher:
When Cloudflare or similar is used for the main/frontend site, it stops the Recording Server requests from ever reaching the server, due to the `User-Agent` reported by the recording server, `Python-urllib/X.Y` ([urllib.request.Request](https://docs.python.org/3/library/urllib.request.html#urllib.request.Request)).

This PR adds a user-configurable setting for such `User-Agent`.

### DIFF
--- a/recording/src/nextcloud/talk/recording/BackendNotifier.py
+++ b/recording/src/nextcloud/talk/recording/BackendNotifier.py
@@ -28,6 +28,7 @@ import json
 import logging
 import os
 import ssl
+from nextcloud.talk import recording
 from secrets import token_urlsafe
 from urllib.request import Request, urlopen
 from urllib3 import encode_multipart_formdata
@@ -98,6 +99,7 @@ def backendRequest(backend, data):
         'OCS-ApiRequest': 'true',
         'Talk-Recording-Random': random,
         'Talk-Recording-Checksum': checksum,
+        'User-Agent': 'Mozilla/5.0 (Recording) Nextcloud-Talk v' + recording.__version__,
     }
 
     backendRequest = Request(url, data, headers)


nickvergessen:
Optimal solution would be to hardcode it to:
```
Mozilla/5.0 (Recording) Nextcloud-Talk v{{version}}
```

mishamosher:
As in, remove the possibility to define it in the `server.conf` file, and have such user agent hardcoded directly into `BackendNotifier.py`?

Only issue I have is that I can't seem to find a Nextcloud version const/var in the Python code. Only bit I have is this: [recording/__init__.py, line 20](https://github.com/nextcloud/spreed/blob/master/recording/src/nextcloud/talk/recording/__init__.py#L20)

Would like to confirm ff it's acceptable to use that constant (`0.1` for now).

mishamosher:
> yes, modifying shouldn't be needed then anymore of the HPB also has the same problem.

I see. Will remove all the user-configurable changes then.

> Seems to be it:
> 
> https://github.com/nextcloud/spreed/blob/b0b5430b8ef59891297dac0e7c8f1652622901e7/recording/src/nextcloud/talk/recording/Server.py#L43-L45

This ends up referring the very same [recording/__init__.py, line 20](https://github.com/nextcloud/spreed/blob/c4bb8002b2124d7af8ec26ee44396df8d8df3015/recording/src/nextcloud/talk/recording/__init__.py#L20), as recording comes from the import `from nextcloud.talk import recording`, retrieving the value `0.1` in the end.

That action (`/api/v1/welcome`) is used by the Nextcloud backend (_not_ the recording server), to query the recording server and show the version of the recording server when configuring it in the Talk settings. It does _not_ retrieve the Nextcloud backend version (v26, v27, etc.), but returns the recording server version (0.1) instead.

nickvergessen:
Yeah that's totally fine and was just confirming that this is the version we are looking for 

So it should results in 
```
Mozilla/5.0 (Recording) Nextcloud-Talk v0.1
```
at the moment, exactly what we want 👍

nickvergessen:
I rebased and dropped the commits that are not necessary (all your commits and then the revert)

PR #9176, opened 2023-03-27
Ensure no double slashes on requests URLs from the recording server to the backend server

mishamosher:
A very small improvement to avoid double slashes on requests URLs from the recording server to the backend server.


mishamosher:
@nickvergessen I've noticed the related PRs (https://github.com/nextcloud/spreed/pull/9189, https://github.com/nextcloud/spreed/pull/9175) got a backport to stable26.

Should this PR also be backported?

PR #9207, opened 2023-03-29
Fix test scripts

marcoambrosini:
- Fixes broken `npm run test:watch` command
- Shortens non-watch into simply `npm run test`


Antreesy (approved):

package.json
```diff
@@ -8,9 +8,9 @@
 		"build": "webpack --node-env production --progress",
 		"dev": "webpack --node-env development --progress",
 		"watch": "webpack --node-env development --progress --watch",
-		"test:unit": "TZ=UTC jest",
+		"test": "TZ=UTC jest",
+		"test:watch": "TZ=UTC jest --watchAll",
```
Antreesy, line 12:
```suggestion
		"test:watch": "TZ=UTC jest --watch",
```
Maybe simple --watch is enough? If you change a common file, like `constants.js`, it triggers 19 / 33 tests to run, only for files that only use constants

package.json
```diff
@@ -8,9 +8,9 @@
 		"build": "webpack --node-env production --progress",
 		"dev": "webpack --node-env development --progress",
 		"watch": "webpack --node-env development --progress --watch",
-		"test:unit": "TZ=UTC jest",
+		"test": "TZ=UTC jest",
```
nickvergessen, line 11: Can we keep the name as `test:unit` to be in line with the Nextcloud eco system?

marcoambrosini: as far as I can see this is not something that's standardised and it's just redundant text since we don't have other commands. Just about as annoying as prefixing every css variable with "default".

E.G. both **photos** and **text** just use `npm run test`
**mail** uses `test:unit` and `test:unit:watch`

package.json
marcoambrosini: sounds good!

package.json
ShGKme: `test:unit` is most likely legacy from using https://cli.vuejs.org

I'd prefer `test` name as it is more common in JS ecosystem and it has npm short run:

```bash
npm run test
# alias
npm test
```

We may old version as an alias aka:

```
"test": "TZ=UTC jest",
"test:unit": "npm test",
```

ShGKme (changes requested):
Should also be updated in makefile

https://github.com/nextcloud/spreed/blob/8d817cd16f75775d5de221920886d453cc07f940/Makefile#L41-L42

ShGKme (approved):
Fine to me.

But we may keep `test:unit`as an alias so old command would also work. For consistency with the other apps or if we missed it in some other CIs.

PR #9246, opened 2023-04-04
fix(avatar): Fix missing avatars for user and group conversation sugg…

nickvergessen:
…estions

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Bildschirmfoto vom 2023-04-04 13-07-18 | Bildschirmfoto vom 2023-04-04 13-02-08

### 🏁 Checklist

### DIFF
--- a/src/components/ConversationIcon.vue
+++ b/src/components/ConversationIcon.vue
@@ -24,21 +24,22 @@
 		class="conversation-icon"
 		:style="{'--icon-size': `${size}px`}"
 		:class="{'offline': offline}">
-		<template v-if="item.token">
-			<NcAvatar v-if="!isOneToOne"
-				:url="avatarUrl"
-				:size="size" />
-			<NcAvatar v-else
-				:size="size"
-				:user="item.name"
-				:disable-menu="disableMenu"
-				:display-name="item.displayName"
-				:preloaded-user-status="preloadedUserStatus"
-				:show-user-status-compact="disableMenu"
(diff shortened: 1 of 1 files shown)


src/components/ConversationIcon.vue
```diff
@@ -155,6 +156,21 @@ export default {
 			return this.$store?.getters.getMainContainerSelector()
 		},
 
+		iconClass() {
+			if (this.item.token) {
+				// Existing conversations use the /avatar endpoint… Always!
```
Antreesy, line 161: Screenshot from 2023-04-04 14-44-06

Somehow not

nickvergessen: Hmm interesting, seems like the token is not populated for open conversations, will have a look

Antreesy: Works as a patch for the moment:
image

constants are here: https://github.com/nextcloud/spreed/commit/eae9143fd24151a93e33213950f2e43eb025cc6b

nickvergessen: We don't want those constants. The frontend should not generate avatars directly anymore when it's conversations.

Antreesy: In that case everything should work if we just remove `<template>` tags, because group icon is show up as well for me
Other additions are unneccessary

Antreesy (approved):

PR #9253, opened 2023-04-04
Avoid out-of-memory errors when uploading a recording

mishamosher:
Currently, the recording server (and the backend server) loads the entirety of the recording file into RAM, which is far from optimal.

This PR changes such behavior and uses a streaming POST request instead.

# For reviewers:

I've noted that `newFile($fileName, $content)` ([source code](https://github.com/nextcloud/server/blob/215aef3cbdc1963be1bb6bca5218ee0a4b7f1665/lib/public/Files/Folder.php#L113)) accepts `string|resource|null`, and `resource` is used in this case.

In my tests it performs a *copy* (and not a simple file rename), which temporarily wastes storage space in my current implementation.

I'm unsure how to force it to perform a file rename instead, or even if the current implementation (copy) is the desired solution so that compatibility with streaming-based file storage is not affected.

### DIFF
--- a/lib/Service/RecordingService.php
+++ b/lib/Service/RecordingService.php
@@ -2,7 +2,8 @@
 
 declare(strict_types=1);
 /**
- * @copyright Copyright (c) 2022 Vitor Mattos <vitor@php.rio>
+ * @copyright Copyright (c) 2022, Vitor Mattos <vitor@php.rio>
+ * @copyright Copyright (c) 2023, Elmer Miroslav Mosher Golovin (miroslav@mishamosher.com)
  *
  * @author Vitor Mattos <vitor@php.rio>
  *
@@ -127,14 +128,16 @@ class RecordingService {
 			throw new InvalidArgumentException('owner_participant');
 		}
 
-		$content = $this->getContentFromFileArray($file, $room, $participant);
+		$resource = $this->getResourceFromFileArray($file, $room, $participant);
 
 		$fileName = basename($file['name']);
-		$this->validateFileFormat($fileName, $content);
+		$fileRealPath = realpath($file['tmp_name']);
+
+		$this->validateFileFormat($fileName, $fileRealPath);
 
 		try {
 			$recordingFolder = $this->getRecordingFolder($owner, $room->getToken());
-			$fileNode = $recordingFolder->newFile($fileName, $content);
+			$fileNode = $recordingFolder->newFile($fileName, $resource);
 			$this->notifyStoredRecording($room, $participant, $fileNode);
 		} catch (NoUserException $e) {
 			throw new InvalidArgumentException('owner_invalid');
@@ -203,7 +206,15 @@ class RecordingService {
 		$this->notificationManager->notify($notification);
 	}
 
-	public function getContentFromFileArray(array $file, Room $room, Participant $participant): string {
+	/**
+	 * Gets a resource that represents the file contents of the file array.
+	 *
+	 * @param array $file File array from which a resource will be returned
+	 * @param Room $room The Talk room that requests the resource
+	 * @param Participant $participant The Talk participant that requests the resource
+	 * @return resource Resource representing the file contents of the file array
+	 */
+	public function getResourceFromFileArray(array $file, Room $room, Participant $participant) {
 		if ($file['error'] !== 0) {
 			$error = self::UPLOAD_ERRORS[$file['error']];
 			$this->logger->error($error);
@@ -220,17 +231,30 @@ class RecordingService {
 			throw new InvalidArgumentException('invalid_file');
 		}
 
-		$content = file_get_contents($file['tmp_name']);
-		unlink($file['tmp_name']);
+		$resource = fopen($file['tmp_name'], 'r');
+		if ($resource === false) {
+			throw new InvalidArgumentException('fopen_failed');
+		}
 
-		if (!$content) {
+		$resourceStat = fstat($resource);
+		if ($resourceStat === false) {
+			throw new InvalidArgumentException('fstat_failed');
+		}
+
+		if ($resourceStat['size'] === 0) {
 			throw new InvalidArgumentException('empty_file');
 		}
-		return $content;
+
+		return $resource;
 	}
 
-	public function validateFileFormat(string $fileName, $content): void {
-		$mimeType = $this->mimeTypeDetector->detectString($content);
+	public function validateFileFormat(string $fileName, string $fileRealPath): void {
+		if (!is_file($fileRealPath)) {
+			$this->logger->warning("An invalid file path ($fileRealPath) was provided");
+			throw new InvalidArgumentException('file_invalid_path');
+		}
+
+		$mimeType = $this->mimeTypeDetector->detectContent($fileRealPath);
 		$allowed = self::DEFAULT_ALLOWED_RECORDING_FORMATS;
 		if (!array_key_exists($mimeType, $allowed)) {
 			$this->logger->warning("Uploaded file detected mime type ($mimeType) is not allowed");
--- a/recording/src/nextcloud/talk/recording/BackendNotifier.py
+++ b/recording/src/nextcloud/talk/recording/BackendNotifier.py
@@ -27,11 +27,12 @@ import hmac
 import json
 import logging
 import os
+import requests
 import ssl
 from nextcloud.talk import recording
+from requests import Request, Session
+from requests_toolbelt import MultipartEncoder
 from secrets import token_urlsafe
-from urllib.request import Request, urlopen
-from urllib3 import encode_multipart_formdata
 
 from .Config import config
 
@@ -62,13 +63,12 @@ def doRequest(backend, request, retries=3):
     """
     context = None
 
-    if config.getBackendSkipVerify(backend):
-        context = ssl.create_default_context()
-        context.check_hostname = False
-        context.verify_mode = ssl.CERT_NONE
+    backendSkipVerify = config.getBackendSkipVerify(backend)
 
     try:
-        urlopen(request, context=context)
+        session = Session()
+        preparedRequest = session.prepare_request(request)
+        session.send(preparedRequest, verify=not backendSkipVerify)
     except Exception as exception:
         if retries > 1:
             logger.exception(f"Failed to send message to backend, {retries} retries left!")
@@ -102,7 +102,7 @@ def backendRequest(backend, data):
         'User-Agent': recording.USER_AGENT,
     }
 
-    backendRequest = Request(url, data, headers)
+    backendRequest = Request('POST', url, headers, data=data)
 
     doRequest(backend, backendRequest)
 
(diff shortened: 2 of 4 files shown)


lib/Service/RecordingService.php
```diff
@@ -156,17 +167,23 @@ public function getContentFromFileArray(array $file, Room $room, Participant $pa
 			throw new InvalidArgumentException('invalid_file');
 		}
 
-		$content = file_get_contents($file['tmp_name']);
-		unlink($file['tmp_name']);
```
danxuliu, line 160: Just to be fully sure, `$file['tmp_name']` is automatically removed by PHP and therefore there is no need to unlink it like it was done before, right?

lib/Service/RecordingService.php
```diff
@@ -156,17 +167,23 @@ public function getContentFromFileArray(array $file, Room $room, Participant $pa
...
+		$resource = fopen($file['tmp_name'], 'r');
+		$resourceStat = fstat($resource);
```
danxuliu, line 171: `fopen` and `fstat` can return `false` in case of failure. Although it should not happen I think it would be good to check if they failed and throw an exception in that case.

recording/src/nextcloud/talk/recording/BackendNotifier.py
```diff
@@ -59,16 +61,26 @@ def doRequest(backend, request, retries=3):
...
-
-    if config.getBackendSkipVerify(backend):
-        context = ssl.create_default_context()
-        context.check_hostname = False
-        context.verify_mode = ssl.CERT_NONE
+    backendSkipVerify = config.getBackendSkipVerify(backend)
 
     try:
-        urlopen(request, context=context)
+        if isinstance(request, UrllibRequest):
```
danxuliu, line 69: Rather than using both `Request` from `urllib` and `requests`, would it be possible to convert the `urllib` requests to use `requests` instead? I guess the `urllib` requests were kept just because it was the existing code (and in general I would prefer that :-) ), but in this case I think a unified approach would be better. Or is there any benefit in using `Request` from `urllib` in addition to the one from `requests`?

recording/src/nextcloud/talk/recording/BackendNotifier.py
```diff
@@ -59,16 +61,26 @@ def doRequest(backend, request, retries=3):
     :param backend: the backend to send the request to.
     :param request: the request to send.
     :param retries: the number of times to retry in case of failure.
+    :param library: the library to use for the request. Possible values are: 'urllib', 'requests'.
```
danxuliu, line 64: `library` seems to be a leftover from an older implementation that can be removed, as it is not explicitly given and the request type is detected via `isinstance`.

recording/src/nextcloud/talk/recording/BackendNotifier.py
```diff
@@ -59,16 +61,26 @@ def doRequest(backend, request, retries=3):
...
+
+            if backendSkipVerify:
+                context = ssl.create_default_context()
+                context.check_hostname = False
+                context.verify_mode = ssl.CERT_NONE
+
+            urlopen(request, context=context)
+        elif isinstance(request, RequestsRequest):
+            session = Session()
+            prepped = session.prepare_request(request)
```
danxuliu, line 80: Although `prepped` is fine, I would prefer to name it `preparedRequest` :-)

danxuliu (changes requested):
Thanks a lot for your contributions and sorry for the delay.

Regarding the copy instead of the rename of the uploaded file I am afraid that I do not know, although from a quick look to [the DAV code](https://github.com/nextcloud/server/blob/98ed72b3ed7e81a75d9a323c70a5e7f5af265a23/apps/dav/lib/Connector/Sabre/File.php#L138) it seems that a copy is done also in that case (except in the case of chunked files, in which the assembled file is moved, but if I recall correctly that assembled file is nevertheless copied from the different chunks). Thus I would assume that, even if it would be better to do a move rather than a copy when possible, a copy is needed in all cases (but please note that it is just a guess).

Besides the comments, could you rebase into latest master? In Nextcloud Talk we prefer rebasing the branch and force pushing it (`--force-with-lease` :-) ) rather than merging the master branch into the pull request branch.

Could you also squash the commit that updates the copyright into the commit that modified the file in first place, the commits to adjust the tests into the commit that modified the code under test, and the commit that fixes the formatting into the commits that added the wrong formatted code?

Thanks!

lib/Service/RecordingService.php
mishamosher: The [official behavior](https://www.php.net/manual/en/features.file-upload.post-method.php) is:

> The file will be deleted from the temporary directory at the end of the request if it has not been moved away or renamed.

In my implementation, neither of those actions (move/rename) is performed on the original temporary file, so it is guaranteed to be removed after the PHP script ends.

An additional `unlink` could be added after [this line](https://github.com/nextcloud/spreed/pull/9253/files#diff-b8455de898ee16a5504ba2177a4674041bf91734999d5405b8ca9cba40cf7eaeR136), but I'm not sure if all the implementations of `file->newFile(...)` guarantee that the file is fully copied after the function call ends.

lib/Service/RecordingService.php
mishamosher: I'm not that familiar with the totality of the spreed/Talk source code. Is there a recommended exception that should be used in those cases, or the built-in ones should suffice (UnexpectedValueException)?

recording/src/nextcloud/talk/recording/BackendNotifier.py
mishamosher: From the docs, I get the impression that `requests` is more oriented towards `functional programming` and `urllib` towards `object-oriented programming`, although `object-oriented programming` is not impossible to achieve with `requests`. Kinda like PHP 4/5 vs PHP 7/8 differences.

Lately `requests` has been moving towards an `object-oriented` approach, so this difference is slowly becoming a non-issue.

If removing `urllib` entirely is not an issue, would it be acceptable to add a separate PR for such change, and keep the current implementation for this PR?

lib/Service/RecordingService.php
danxuliu: > In my implementation, neither of those actions (move/rename) is performed on the original temporary file, so it is guaranteed to be removed after the PHP script ends.

Sounds good :-)
 
> An additional `unlink` could be added after [this line](https://github.com/nextcloud/spreed/pull/9253/files#diff-b8455de898ee16a5504ba2177a4674041bf91734999d5405b8ca9cba40cf7eaeR136), but I'm not sure if all the implementations of `file->newFile(...)` guarantee that the file is fully copied after the function call ends.

I am not aware of any implementation that does an asynchronous copy. But if there were, I guess it would fail anyway due to the temporal file being deleted at the end of the request. So in practice I think that the implicit unlink is fine (and if it turns out that left over files are eventually found we can always fix it later).

lib/Service/RecordingService.php
danxuliu: The rest of the method just uses generic `InvalidArgumentException` with a descriptive parameter, so that could be (ab)used again in this case :-) (even if it does not exactly match [the official description of the exception](https://www.php.net/manual/en/class.invalidargumentexception.php))

recording/src/nextcloud/talk/recording/BackendNotifier.py
danxuliu: > From the docs, I get the impression that `requests` is more oriented towards `functional programming` and `urllib` towards `object-oriented programming`, although `object-oriented programming` is not impossible to achieve with `requests`. Kinda like PHP 4/5 vs PHP 7/8 differences.
> 
> Lately `requests` has been moving towards an `object-oriented` approach, so this difference is slowly becoming a non-issue.

I am not familiar enough neither with `requests` nor `urllib`, so I can not provide any insight on that, sorry :-)

> If removing `urllib` entirely is not an issue, would it be acceptable to add a separate PR for such change, and keep the current implementation for this PR?

Yes, it would be fine. But I thought that it would just require changing `backendRequest = UrllibRequest(url, data, headers)` to `backendRequest = Request('POST', url, headers, data)` (based on the change done in `uploadRecording`) :thinking: Is something else required, or why would you like to do it in a different pull request? For the record, my idea was to first move the existing `urllib` code to `requests` and then, in a separate commit, adjust it to stream the upload.

lib/Service/RecordingService.php
mishamosher: Will keep the current implementation then (:

lib/Service/RecordingService.php
mishamosher: Got it. Thanks.

recording/src/nextcloud/talk/recording/BackendNotifier.py
mishamosher: After reading all of the `urllib` calls, all of them are `POST` requests. Completely replacing `urllib` by `requests` should require minimal changes as you suggest.

Asked about a separate PR because of the description, but that can be adjusted anyways.

tl;dr if completely replacing `urllib` by `requests` is desired, so be it in this very PR (:

mishamosher:
@danxuliu done with most (if not all) of the requested changes.

recording/src/nextcloud/talk/recording/BackendNotifier.py
danxuliu: > After reading all of the `urllib` calls, all of them are `POST` requests. Completely replacing `urllib` by `requests` should require minimal changes as you suggest.

He, now you made me wonder about what would happen if non-POST requests are needed at some point. But... let's worry about that in the future if needed ;-)

> Asked about a separate PR because of the description, but that can be adjusted anyways.

In general I prefer separate pull requests that are rebased one on top of the other (to nickvergessen's dismay :-P ), but in this case I think it is fine to do the groundwork of moving from `urllib` to `requests` as a separate commit in the same pull request.

> tl;dr if completely replacing `urllib` by `requests` is desired, so be it in this very PR (:

Perfect, thanks :-)

danxuliu (commented):
Thanks!

Would it be possible to split the first commit in one commit that changes from `urllib` to `requests`, and then in another commit to use streaming POST requests? If moving to `requests` requires using `MultipartEncoder` and, with that, it implicitly provides streaming POST requests, or if `encode_multipart_formdata` can not be used with `requests` and something complex is needed that will be just replaced by `MultipartEncoder` in the next commit then it is fine in a single commit, of course. But if possible I would prefer it to be split to have more "self-contained" commits.

Regarding the PHP changes I would prefer if the tests are kept in sync with the code under test even at the commit level, so I would squash the third commit into the second one.

All this is nitpicking (sorry!), so I can do the changes myself if you want, just let me know :-)

mishamosher:
@danxuliu changes applied.

danxuliu (approved):
Tested and works :+1:

I took the liberty to adjust the second commit, as it was using the already removed `encode_multipart_formdata`. I hope that is fine :-)

Independently of that, excellent work! Thanks!

mishamosher:
The requests library uses urllib as a dependency, so it was not really
removed.

To be honest, I'm just hoping to have the recording functionality working
out-of-the-box in my setup (:

Thanks to you for taking the time to review the PR.

danxuliu:
> The requests library uses urllib as a dependency, so it was not really removed.

Argh, you are right... I forgot that it was a dependency. I meant removed from the imports, but it could have been added back to the imports instead of using already the `MultipartEncoder`. But I enabled auto-merge and now it is too late :facepalm: 

> To be honest, I'm just hoping to have the recording functionality working out-of-the-box in my setup (:

We are almost there :-) I hope... :-P

> Thanks to you for taking the time to review the PR.

Thanks to you for applying the changes :-)

PR #9256, opened 2023-04-04
Show empty content only when there is nothing to show

Pytal:
### ☑️ Resolves

- For https://github.com/nextcloud/nextcloud-vue/issues/3838

Adapt to component changes requested by designers in https://github.com/nextcloud/nextcloud-vue/pull/3852

Now the panel is visible when there is an error in addition to when there are resources

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/RightSidebar/SharedItems/SharedItemsTab.vue
+++ b/src/components/RightSidebar/SharedItems/SharedItemsTab.vue
@@ -42,7 +42,8 @@
 		<NcRelatedResourcesPanel class="related-resources"
 			provider-id="talk"
 			:item-id="conversation.token"
-			@has-resources="value => hasRelatedResources = value" />
+			@has-resources="value => isRelatedResourcesVisible = value"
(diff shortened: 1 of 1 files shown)


nickvergessen:
This looks like a wrong approach to me. The problem is that the error occurs, not that it is shown.

Also as you can see it removes the "No shared items" which is not what we want.

nickvergessen:
The actual problem comes from:
> Bildschirmfoto vom 2023-04-05 10-10-02

There is an error, but nothing in the logs. So also telling the user to contact the administration is not the right thing to do.

nickvergessen:
Let's start with finding out what is going on:
https://github.com/nextcloud/related_resources/pull/206

Unluckily I didn't check what the issue was when taking above screenshot and it seems to not always happen

nickvergessen:
The actual error to fix is at https://github.com/nextcloud/related_resources/issues/207

PR #9259, opened 2023-04-05
Add note about using modifiedSince with includeStatus set to true

SystemKeeper:
### ☑️ Resolves

One to one conversations will always be returned when we use `modifiedSince` and also set `includeStatus` to `true` as implemented here:

https://github.com/nextcloud/spreed/blob/f112fb92cc641af27d0d0644940a658a974ba319/lib/Controller/RoomController.php#L202-L207

This should be added to the docs.

### 🏁 Checklist

### DIFF
--- a/docs/conversation.md
+++ b/docs/conversation.md
@@ -15,7 +15,7 @@
 
 !!! note
 
-    When `modifiedSince` is provided only conversations with a newer `lastActivity` are returned. If `includeStatus` is set to `true` this will also return all one-to-one conversations.
+    When `modifiedSince` is provided only conversations with a newer `lastActivity` are returned. If `includeStatus` is set to `true` all one-to-one conversations will be returned to make sure the latest user status is returned.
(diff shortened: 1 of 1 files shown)


Ivansss (approved):

docs/conversation.md
```diff
@@ -15,7 +15,7 @@
 
 !!! note
 
-    When `modifiedSince` is provided only conversations with a newer `lastActivity` are returned.
+    When `modifiedSince` is provided only conversations with a newer `lastActivity` are returned. If `includeStatus` is set to `true` this will also return all one-to-one conversations.
```
nickvergessen, line 18:
```suggestion
    When `modifiedSince` is provided only conversations with a newer `lastActivity` are returned. If `includeStatus` is set to `true` all one-to-one conversations will be returned to make sure the latest user status is returned.
```

nickvergessen (approved):

PR #9275, opened 2023-04-06
Fix squeezed user mentions in NcRichcontenteditable

Antreesy:
### ☑️ Resolves

* Fix container reference, passing to subcomponents in `NewMessageForm` (without changes, document.body by default)
* Prevents tribute to appear in `#content-vue` (which resolves this flickering)

image

* As a result, tribute doesn't inherit `box-sizing: border-box` from `NcContent` and looks normal

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot from 2023-04-06 14-30-17 | Screenshot from 2023-04-06 14-30-31

### 🚧 Tasks

- [ ] Code review

### 🏁 Checklist

### DIFF
--- a/src/components/NewMessageForm/NewMessageForm.vue
+++ b/src/components/NewMessageForm/NewMessageForm.vue
@@ -308,12 +308,12 @@ export default {
 		},
 
 		/**
-		 * When this component is used to send message to a breakout room we
-		 * pass an id of modal containing component to render properly.
+		 * Selector for popovers and pickers to be rendered inside container properly.
+		 * Container must be mounted before passing its ID as a prop
 		 */
-		containerId: {
+		container: {
 			type: String,
-			default: null,
+			required: true,
 		},
 
 		/**
@@ -409,14 +409,8 @@ export default {
 			return this.text !== ''
 		},
 
-		container() {
-			return this.containerId ?? this.$store.getters.getMainContainerSelector()
-		},
-
 		containerElement() {
-			// TODO can't find DOM element by #content-vue. undefined is passed
-			//  for NcRichContenteditable to use 'document.body' by default
-			return document.querySelector(this.container) ?? undefined
+			return document.querySelector(this.container)
 		},
 
 		isOneToOne() {
@@ -928,6 +922,20 @@ export default {
 	},
 }
 </script>
+
+<style lang="scss">
+// Enforce NcAutoCompleteResult to have proper box-sizing
+.tribute-container {
+	position: absolute;
+	box-sizing: content-box !important;
+
+	& *,
+	& *::before,
+	& *::after {
+		box-sizing: inherit !important;
+	}
+}
+</style>
 
 <style lang="scss" scoped>
 @import '../../assets/variables';
--- a/src/components/BreakoutRoomsEditor/SendMessageDialog.vue
+++ b/src/components/BreakoutRoomsEditor/SendMessageDialog.vue
@@ -30,7 +30,7 @@
 				role="region"
 				:token="token"
 				:breakout-room="true"
-				:container-id="modalContainerId"
+				:container="modalContainerId"
 				:aria-label="t('spreed', 'Post message')"
 				:broadcast="broadcast"
 				@sent="handleMessageSent"
--- a/src/components/ChatView.vue
+++ b/src/components/ChatView.vue
@@ -42,8 +42,10 @@
 			:aria-label="t('spreed', 'Conversation messages')"
 			:token="token"
 			:is-visible="isVisible" />
-		<NewMessageForm role="region"
+		<NewMessageForm v-if="containerId"
+			role="region"
 			:token="token"
+			:container="containerId"
 			:aria-label="t('spreed', 'Post message')" />
 	</div>
 </template>
@@ -73,6 +75,7 @@ export default {
 	data() {
 		return {
 			isDraggingOver: false,
+			containerId: undefined,
 		}
 	},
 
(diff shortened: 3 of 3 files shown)


nickvergessen:
- [x] Works while being in a call
- [ ] But outside it looks too huge for me (Firefox)

marcoambrosini (approved):

marcoambrosini (changes requested):
same here, chrome

Screenshot 2023-04-06 at 14 20 50

src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -929,6 +929,13 @@ export default {
 }
 </script>
 
+<style lang="scss">
+// Tribute container inherits box-sizing: border-box from NcContent, so height should be overrided
+.tribute-container .autocomplete-result {
+	height: 64px !important;
+}
+</style>
```
ShGKme, line 937: If the problem is different `box-sizing`, cannot we set only correct `box-sizing` value?

Antreesy: Thought about it. Will try that later

src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -311,9 +311,8 @@ export default {
 		 * When this component is used to send message to a breakout room we
 		 * pass an id of modal containing component to render properly.
 		 */
-		containerId: {
+		container: {
 			type: String,
-			default: null,
```
nickvergessen, line 316: GitHub Actions / eslint

Prop 'container' requires default value to be set

Antreesy: By default we need it as 'undefined', and it works like we need that, now

@shgkme, is it necessary for Vue to assign a default value in that case?

nickvergessen: It's necessary for eslint, just specify default: undefined?

Antreesy: If other changes will be approved, I'll add it

nickvergessen (dismissed):
tribute no longer shows when running talk in fullscreen (press `F` when chat input is not focused or use the option in the … menu):
> image

image

src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -308,12 +308,12 @@ export default {
...
-		 * pass an id of modal containing component to render properly.
+		 * Selector for popovers and pickers to be rendered inside container properly.
+		 * If undefined, subcomponents treat it as document.body.
 		 */
-		containerId: {
+		container: {
 			type: String,
-			default: null,
+			default: undefined,
 		},
```
ShGKme, line 317: I'd remove comment about "`If undefined, subcomponents treat it as document.body.`".

This fact is out of scope of this component. It is some fact about implementation of a component from another library. We don't know, what value it uses by default (it may change any time), and actually it is not important to the component's user: we don't need container to be exactly `document.body` (otherwise we should set `'body'` as the default value).

At the same time more important thing is not mentioned: the container have to be `box-sizing: content-box`.

src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -409,13 +409,8 @@ export default {
...
 
-		container() {
-			return this.containerId ?? this.$store.getters.getMainContainerSelector()
-		},
-
 		containerElement() {
-			// TODO can't find DOM element by #content-vue. undefined is passed
-			//  for NcRichContenteditable to use 'document.body' by default
+			// returns DOM Element. If undefined, subcomponents treat it as body.
 			return document.querySelector(this.container) ?? undefined
```
ShGKme, line 414: Same here

ShGKme (approved):
Works in all scenarios.

`NcAutoCompleteResult` is required to be `box-sizing: content-box`.

It is rendered in the `NcRichContenteditable`'s `menu-container`. By default the `menu-container` is `body` with `box-sizing: content-box` in the Nextcloud layout.

Previously `NewMessageForm` passed `#content-vue` as the `menu-container`. It was fine for positioning, but it had different `box-sizing`.

New solution leaves it up to the `NcRichContenteditable`.

I'd also think about fixing the problem of containers in the `@nextcloud/vue` library.

Antreesy:
> tribute no longer shows when running talk in fullscreen (press `F` when chat input is not focused or use the option in the … menu):

Same for "Leave call" button, "Broadcast to breakout rooms" modal and other popovers and modals in fullscreen. Problem persists on master branch, it's not related to this PR and should be fixed, but as a follow up @nickvergessen

marcoambrosini (approved):

Antreesy:
Follow up for fullsreeens is in issue https://github.com/nextcloud/spreed/issues/9299

src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -929,6 +923,20 @@ export default {
...
+	position: absolute;
+	box-sizing: content-box !important;
+
+	& *,
+	& *::before,
+	& *::after {
+		box-sizing: inherit !important;
+	}
+}
+</style>
```
ShGKme, line 938: Do we need to set `position: absolute;` here?
It seems that it is set by tribute as an inline style anyway.

ShGKme (approved):
Fine to me, tested in:
- General chat view
- Chat view in fullscreen
- Message all rooms model

src/components/NewMessageForm/NewMessageForm.vue
Antreesy: It set as a style rule after mounting, before that - it is static and cause flickering on the right

PR #9288, opened 2023-04-08
Delete dot

Valdnet:
Unification of error messages.

### DIFF
--- a/src/utils/signaling.js
+++ b/src/utils/signaling.js
@@ -430,7 +430,7 @@ Signaling.Internal.prototype._sendMessageWithCallback = function(ev) {
 		}.bind(this))
 		.catch(function(err) {
 			console.error(err)
-			showError(t('spreed', 'Sending signaling message has failed.'))
+			showError(t('spreed', 'Sending signaling message has failed'))
 		})
 }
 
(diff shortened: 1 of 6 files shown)


src/utils/signaling.js
```diff
@@ -1041,12 +1041,12 @@ Signaling.Standalone.prototype.helloResponseReceived = function(data) {
 
 	if (!this.settings.helloAuthParams.internal && (!this.hasFeature('audio-video-permissions') || !this.hasFeature('incall-all') || !this.hasFeature('switchto'))) {
 		showError(
-			t('spreed', 'The configured signaling server needs to be updated to be compatible with this version of Talk. Please contact your administrator.'),
```
nickvergessen, line 1044: Why is that? Is there a rule for it? I think it's quite unlucky to remove a dot as it breaks all languages. Also it's a full sentence, so it should have a dot, no?

Valdnet: With error messages consisting of two or more sentences, there is always the problem of whether there should be a dot at the end of the sentence.
In my opinion, multi-sentence error messages should be avoided. The message should be worded in one sentence.
Some applications use a dot at the end of the next sentence, while some do not. It is important that the spelling rules are the same for all messages.
The change has been undone.

nickvergessen: I guess https://docs.nextcloud.com/server/latest/developer_manual/basics/front-end/l10n.html

Then again, why would a full sentence not end with a dot? Only when it's in a toast?
Is it different when it's shown in a modal, or any other UI element? That's such a weird rule and it will be broken all the time (especially by new apps and when community sends PRs, etc) and until it's then noticed and fixed (like here) community translated it already, but then it's being broken by removing the trailing dot and shows up untranslated again unless translators translate it a second time...

nickvergessen: My question remains, when should the dot at the end of a full sentence be skipped and why?

marcoambrosini (approved):

nickvergessen:
There is still unresolved questions. Please don't merge such PRs @marcoambrosini

PR #9297, opened 2023-04-12
Fix broken styles for navigation buttons in call

Antreesy:
### ☑️ Resolves

* Fix overriding of `position: absolute` for navigation buttons with base styles

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot from 2023-04-11 16-27-09 | Screenshot from 2023-04-11 16-27-16
Screenshot from 2023-04-11 16-26-25 | Screenshot from 2023-04-11 16-26-30

### 🚧 Tasks

- [ ] Code review
- [ ] Visual check

### 🏁 Checklist

### DIFF
--- a/src/components/CallView/Grid/Grid.vue
+++ b/src/components/CallView/Grid/Grid.vue
@@ -37,7 +37,7 @@
 		</NcButton>
 		<transition :name="isStripe ? 'slide-down' : ''">
 			<div v-if="!isStripe || stripeOpen" class="wrapper" :style="wrapperStyle">
-				<div :class="{'stripe-wrapper': isStripe, 'wrapper': !isStripe}">
+				<div :class="[isStripe ? 'stripe-wrapper' : 'grid-wrapper']">
 					<NcButton v-if="hasPreviousPage && gridWidth > 0"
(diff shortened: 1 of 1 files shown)


src/components/CallView/Grid/Grid.vue
```diff
@@ -1011,7 +1017,7 @@ export default {
 }
 
 .stripe--collapse {
-	position: absolute;
+	position: absolute !important;
 	top: calc(-1 * var(--default-clickable-area));
 	right: 0;
 	z-index: 2;
```
ShGKme, line 1023: Could we wrap a button in a new absolute block instead of overriding its styles with `!important`?

ShGKme (approved):
Works fine on all layouts

src/components/CallView/Grid/Grid.vue
Antreesy: I don't want to add a new wrapper.
But adding some weight to this rule should be enough for solving, so i would go with a solution like this:


```css
.grid-main-wrapper {
    position: relative;

    & .stripe--collapse {
        position: absolute;
    }
}
```

PR #9304, opened 2023-04-13
Reactions in video calls

### LINKED ISSUE #9266: Frontend "Reactions in video calls"
* Overview in https://github.com/nextcloud/spreed/issues/9249

---

- [ ] Add an emoji icon in the call view
- [ ] Tabbing it allows to react with a predefined list of emojis
- [ ] Emoji is sent as signaling message
- [ ] When a signaling message is received, the emoji should be shown

Antreesy:
### ☑️ Resolves

* Overview in #9249
* Close #9266

### 🖼️ Screenshots

**Reacting** - user's click is throttled to 1 reaction / 2 seconds. Horizontal and vertical offset are randomized to make it look more "natural"
One reaction | Spamming with reactions
-- | --
image | image

---
**Overlapping** - reactions source is located around 10-30% of call container width from the left, and each reaction flows up and dissapears at around 30-35% of screen height.
Speaker view | Screensharing
--- | ---
image | image

**Screen capture**

[reactions.webm](https://user-images.githubusercontent.com/93392545/233122383-ca5c4f14-8a7c-4166-96d7-95b9f685d62b.webm)

### 🚧 Tasks
Ongoing tasks:
- [ ] The available reactions are: ❤️ 🎉 👏🏽 👍🏽 👎🏽 😂 🤩 🤔 😮 😢 (to be discussed)
- [ ] Code review
- [ ] Animation behavior confirmation

API tasks:
- [x] Emoji is sent as signaling message
- [x] When a signaling message is received, the emoji is shown
- [x] Get supported reactions from capabilities

Visual tasks:
- [x] The emojis reactions are accessible from a menu in the TopBar. Menu doesn't close on click.
- [x] Received (and self) reaction appears at the bottom of the screen and floats to the top with a sender name next to it and fades away.
- [x] The color of the name tag could be the avatar color
- [x] For small screens, size of emoji is 24px. For bigger screens: 44px (starting from 1920px)
- [x] The total time taken for the emoji to float to the top and disappear is 2000ms 
- [x] The total height travelled by the emoji depends on the screen. It can vary by 10% to look more natural. Generally it should disappear before it reaches the mid point of the screen to not be too distracting. (30-35% of screen height)
- [x] The emoji should start fading while it reaches the top. However, it doesn't look right if it starts fading right from the start. So from 0% to 50% of the animation, only the position would change and opacity remains constant. For 50% to 100% of the animation, the position would change as well as the opacity would change from 1 to 0. Transition is linear.
- [x]  One person can react maximum once per 2 seconds.
- [x] If many people are reacting simultaneously, the number of reactions shown on screen must be limited to avoid overcrowding the screen.

### 🏁 Checklist

### DIFF
--- a/src/components/CallView/shared/ReactionToaster.vue
+++ b/src/components/CallView/shared/ReactionToaster.vue
@@ -54,6 +54,16 @@ export default {
 			type: String,
 			required: true,
 		},
+
+		/**
+		 * Supported reactions
+		 */
+		supportedReactions: {
+			type: Array,
+			validator: (prop) => prop.every(e => typeof e === 'string'),
+			required: true,
+		},
+
 		callParticipantModels: {
 			type: Array,
 			required: true,
@@ -110,6 +120,11 @@ export default {
 		},
 
 		handleReaction(model, reaction) {
+			// prevent receiving anything rather than defined reactions in capabilities
+			if (!this.supportedReactions.includes(reaction)) {
+				return
+			}
+
 			this.reactionsQueue.push({
 				reaction,
 				name: this.getParticipantName(model),
--- a/src/components/CallView/shared/ReactionToaster.vue
+++ b/src/components/CallView/shared/ReactionToaster.vue
@@ -1,0 +1,228 @@
+<!--
+  - @copyright Copyright (c) 2023 Maksim Sukharev <antreesy.web@gmail.com>
+  -
+  - @author Maksim Sukharev <antreesy.web@gmail.com>
+  -
+  - @license GNU AGPL version 3 or any later version
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+-->
+
+<template>
+	<transition-group name="toast" class="toaster" tag="ul">
+		<li v-for="toast in toasts"
+			:key="toast.seed"
+			class="toast"
+			:style="styled(toast.name, toast.seed)">
+			<span class="toast__reaction">
+				{{ toast.reaction }}
+			</span>
+			<span class="toast__name">
+				{{ toast.name }}
+			</span>
+		</li>
+	</transition-group>
+</template>
+
+<script>
+import Hex from 'crypto-js/enc-hex.js'
+import SHA1 from 'crypto-js/sha1.js'
+
+import { subscribe, unsubscribe } from '@nextcloud/event-bus'
+
+import usernameToColor from '@nextcloud/vue/dist/Functions/usernameToColor.js'
+
+export default {
+	name: 'ReactionToaster',
+
+	props: {
+		/**
+		 * The conversation token
+		 */
+		token: {
+			type: String,
+			required: true,
+		},
+		callParticipantModels: {
+			type: Array,
+			required: true,
+		},
+	},
+
+	data() {
+		return {
+			registeredModels: {},
+			reactionsQueue: [],
+			intervalId: null,
+			animationLength: 2000,
+			toasts: [],
+		}
+	},
+
+	computed: {
+		participants() {
+			return this.$store.getters.participantsList(this.token)
+		},
+	},
+
+	watch: {
+		callParticipantModels(models) {
+			// subscribe connected models for reaction signals
+			const addedModels = models.filter(model => !this.registeredModels[model.attributes.peerId])
+			addedModels.forEach(addedModel => {
+				this.registeredModels[addedModel.attributes.peerId] = addedModel
+				this.registeredModels[addedModel.attributes.peerId].on('reaction', this.handleReaction)
+			})
+
+			// unsubscribe disconnected models
+			const removedModels = Object.keys(this.registeredModels).filter(registeredModelId => !models.find(model => model.attributes.peerId === registeredModelId))
+			removedModels.forEach(removedModel => {
+				this.registeredModels[removedModel].off('reaction', this.handleReaction)
+				delete this.registeredModels[removedModel]
+			})
+		},
+	},
+
+	mounted() {
+		this.intervalId = setInterval(this.processReactionsQueue, this.animationLength / 4)
+		subscribe('send-reaction', this.handleOwnReaction)
+	},
+
+	beforeDestroy() {
+		clearInterval(this.intervalId)
+		unsubscribe('send-reaction', this.handleOwnReaction)
+	},
+
+	methods: {
+		handleOwnReaction({ model, reaction }) {
+			this.handleReaction(model, reaction)
+		},
+
+		handleReaction(model, reaction) {
+			this.reactionsQueue.push({
+				reaction,
+				name: this.getParticipantName(model),
+				seed: Math.random(),
+			})
+		},
+
+		processReactionsQueue() {
+			if (this.reactionsQueue.length > 0) {
+				// Move reactions from queue to visible array
+				this.toasts.push(this.reactionsQueue.shift())
+
+				// Delete reactions from array after animation ends
+				setTimeout(() => {
+					this.toasts.shift()
+				}, this.animationLength)
+			}
+		},
+
+		getParticipantName(model) {
+			const { name, peerId } = model.attributes
+			if (name) {
+				return name
+			}
+
+			const participant = this.participants.find(participant => participant.sessionIds.includes(peerId))
+			if (participant?.displayName) {
+				return participant.displayName
+			}
+
+			return this.$store.getters.getGuestName(this.token, Hex.stringify(SHA1(peerId)))
+		},
+
+		styled(name, seed) {
+			const color = usernameToColor(name)
+
+			return {
+				'--background-color': `rgb(${color.r}, ${color.g}, ${color.b})`,
+				'--animation-length': `${this.animationLength + 300}ms`,
+				'--horizontal-offset': `${10 + 20 * seed}%`,
+				'--vertical-offset': 30 + 5 * seed,
+			}
+		},
+	},
+}
+</script>
+
+<style lang="scss" scoped>
+.toaster {
+	position: absolute;
+	bottom: 20px;
+	left: 0;
+	display: flex;
+	flex-direction: column;
+	gap: 20px;
+	width: 100%;
+	z-index: 1;
+}
+
+.toast {
+	position: absolute;
+	bottom: 0;
+	left: var(--horizontal-offset, 0);
+	display: flex;
+	align-items: center;
+	gap: 8px;
+	animation: toast-floating var(--animation-length) linear;
+
+	&__reaction {
+		font-size: 250%;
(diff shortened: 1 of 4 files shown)


src/components/CallView/shared/ReactionToaster.vue
```diff
@@ -0,0 +1,165 @@
...
+			}, this.animationLength)
+		},
+
+		styled(user, seed) {
+			const styles = {}
+
+			const color = usernameToColor(user)
+			styles['--background-color'] = `rgb(${color.r}, ${color.g}, ${color.b})`
+			styles['--animation-length'] = `${this.animationLength + 300}ms`
+			styles['--horizontal-offset'] = `${5 + 10 * seed}%`
```
nickvergessen, line 89: not sure which one is which, but I think the left-right should vary more. 5 pixels or 10% is too little. I would at least allow for a left-right offset of 30-40%
```suggestion
			styles['--horizontal-offset'] = `${5 + 30 * seed}%`
```

src/components/CallView/shared/ReactionToaster.vue
```diff
@@ -0,0 +1,165 @@
...
+				{{ toast.user }}
+			</span>
+		</li>
+	</transition-group>
+</template>
+
+<script>
+import usernameToColor from '@nextcloud/vue/dist/Functions/usernameToColor.js'
+
+import { EventBus } from '../../../services/EventBus.js'
```
marcoambrosini, line 41: This is deprecated and we should use @nextcloud/event-bus instead from now on.

src/components/TopBar/ReactionMenu.vue
```diff
@@ -0,0 +1,170 @@
...
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+-->
+
+<template>
+	<NcActions v-tooltip="t('spreed', 'React')"
```
marcoambrosini, line 23:
```suggestion
	<NcActions v-tooltip="t('spreed', 'Send a reaction')"
```

src/components/TopBar/ReactionMenu.vue
```diff
@@ -0,0 +1,170 @@
...
+		:aria-label="t('spreed', 'React')"
```
marcoambrosini, line 24:
```suggestion
		:aria-label="t('spreed', 'Send a reaction')"
```

src/components/TopBar/TopBar.vue
```diff
@@ -74,6 +74,13 @@
 			{{ participantsInCall }}
 		</NcButton>
 
+		<!-- TopBar menu -->
```
marcoambrosini, line 77:
```suggestion
		<!-- Reactions menu -->
```

src/constants.js
```diff
@@ -38,6 +38,8 @@ export const CALL = {
 		AUDIO_STARTING: 4,
 		FAILED: 5,
 	},
+
+	REACTIONS: ['❤️', '🎉', '👏', '👍', '👎', '😂', '🤩', '🤔', '😮', '😥'],
```
marcoambrosini, line 42:
```suggestion
	REACTIONS: ['❤️', '🎉', '👏', '👍', '😂', '🤩', '🤔', '😮', '😥'],
```
Not nice

nickvergessen: As per design call we want a "not nice" option.

nimishavijay: How about 😐 instead of 👎

marcoambrosini: I'd avoid that at all and require people who want to express dissent to do it:
* not while someone is presenting speaking
* in a more nuanced way than just an emoji reaction

nickvergessen: Manager says: "User x quit"

You want to immediately react to that visually.

I agree that we shouldn't have the 💩 emoji, but a thumbs down also for getting feedback like "should we do or not" is not to intrusive/negative as we agreed in the design planning call.

Antreesy: "Guns don't kill people, people do"

I agree that "thumb down" is the most neutral emoji we could offer here for negative reactions. It's up to users to decide, whether they're going to use it or not.

If we'll dig deeper, other emojis could be also considered as inappropriate:
https://mashable.com/article/emoji-culture-world

src/components/CallView/shared/VideoVue.vue
```diff
@@ -482,6 +482,19 @@ export default {
...
+			handler(model, oldModel) {
+				if (oldModel) {
+					oldModel.off('reaction', this._handleReaction)
+				}
+
+				if (model) {
+					model.on('reaction', this._handleReaction)
+				}
+			},
+		},
```
danxuliu, line 497: If I am not mistaken adding a handle for the reaction event emitted by the model should not be done in the VideoView, unless it is desired that only reactions for visible participants are shown.

Moreover, I do not think that the props and hence the model are nullified when the component is destroyed (but I might be wrong, I am far from a Vue expert), so the event listener is not properly removed when the component is destroyed. This could explain why reactions multiply after switching from speaker mode to grid view and back again: new VideoView components are created, but although the older ones are destroyed the listeners are still running (this might happen already in the example code included in #9311, but I have not checked).

I would suggest to do something like passing [`callParticipantModels`](https://github.com/nextcloud/spreed/blob/c74f038d317ade3e20dd4493b7513733a46aa75a/src/components/CallView/CallView.vue#L207) to the `ReactionToaster` and watching it for changes, adding or removing a listener for each model as needed. Something similar to (**untested**, based on the (horrible :-P ) code of [`updateDataFromCallParticipantModels`](https://github.com/nextcloud/spreed/blob/c74f038d317ade3e20dd4493b7513733a46aa75a/src/components/CallView/CallView.vue#L438), so if you have any cleaner idea please do that instead ;-) ):
```
data: {
    registeredModels: {},
},
...
watch: {
    callParticipantModels(models) {
        const addedModels = models.filter(model => !this.registeredModels[model.attributes.peerId])
        const removedModelIds = Object.keys(this.registeredModels).filter(registeredModelId => models.find(model => model.attributes.peerId === registeredModelId) === undefined)

        removedModelIds.forEach(removedModelId => {
            this.registeredModels[addedModel.attributes.peerId].off('reaction', this._handleReaction)
            delete this.registeredModels[addedModel.attributes.peerId]
        })

        addedModels.forEach(addedModel => {
            this.registeredModels[addedModel.attributes.peerId] = addedModel
            this.registeredModels[addedModel.attributes.peerId].on('reaction', this._handleReaction)
        })
    },
},
methods: {
    _handleReaction(model, reaction) {
        ...
    },
},
```

src/constants.js
jancborchardt: So neither Teams nor Jitsi have a thumbs down reaction like this. I agree with @marcoambrosini also that a simple thumbs down reaction is not very productive without elaboration. On GitHub e.g. it’s often very frustrating to see thumbs down without further explanation.

We have the 🤔 emoji which is enough, and if in doubt I would rather go with what other alternatives use.

nickvergessen: image

nickvergessen: The code point is moved to the backend now (so it can be consumed by clients in similar fashion and be toggled based on the server support) at https://github.com/nextcloud/spreed/pull/9337/files#diff-d8666c6d9eaabdff763631d59b9df820f9625645a705199e82ca1844afe0eeb4R130

However, I'm strongly against censoring into a positive only reaction set not allowing to express disconsent or gather a quick yay/nay overview.

src/components/CallView/shared/VideoVue.vue
Antreesy: Thanks for the tip!

src/components/CallView/shared/ReactionToaster.vue
```diff
@@ -0,0 +1,228 @@
...
+			})
+		},
+	},
+
+	mounted() {
+		this.intervalId = setInterval(this.processReactionsQueue, this.animationLength / 4)
+		subscribe('send-reaction', this.handleOwnReaction)
+	},
+
+	beforeDestroy() {
```
danxuliu, line 102: The registered models should be iterated over and deregistered (`.off('reaction', this.handleReaction)`) when the component is destroyed (or before being destroyed, that should not matter).

The models will be (probably) deregistered anyway when the models are removed from `callParticipantModels` even if the component was already destroyed (although I am not fully sure about that, I do not know if watchers are still run after the component is destroyed), but it would be cleaner to do it when the component is destroyed and thus no longer needed.

src/components/TopBar/ReactionMenu.vue
```diff
@@ -0,0 +1,176 @@
...
+				const reactionRand = this.reactions[Math.floor(Math.random() * 10)]
+				this.localCallParticipantModel.sendReaction(reactionRand)
+
+				// show reaction to yourself
+				emit('send-reaction', {
+					model: this.localCallParticipantModel,
+					reaction: reactionRand,
+				})
+			}, 1000)
+		},
```
ShGKme, line 157: **thought:** we may keep this method and only remove the button from UI

ShGKme (approved):
Looks great by the code to me, works fine with a spam button, not brakes without reactions support.
👏 👍 🤩 😮 ❤️ 🎉

src/components/TopBar/ReactionMenu.vue
Antreesy: Then could it be used by someone through the devtools? Or it will not be accessible in production?

Antreesy: I'd like to remove it at all, as it's not hard to reproduce again. Or at least move this implementation in unit test, there it could be used for good

nickvergessen: https://github.com/nextcloud/spreed/blob/788ec820eb612b706d5b70b8c70b2494730dcfb5/src/components/NewMessageForm/NewMessageForm.vue#L533

Maybe make sure to wrap it with OC.debug :P

src/components/CallView/shared/ReactionToaster.vue
```diff
@@ -110,6 +120,11 @@ export default {
 		},
 
 		handleReaction(model, reaction) {
+			// prevent receiving anything rather than defined reactions in capabilities
+			if (!this.supportedReactions.includes(reaction)) {
+				return
+			}
+
```
ShGKme, line 136: Do we need this check on frontend? Backend should handle it anyway.

src/components/TopBar/ReactionMenu.vue
ShGKme: > Maybe make sure to wrap it with OC.debug :P

In context of "how easy is it to run a testing method", it is still very easy to set `OC.debug` to `true`

src/components/CallView/shared/ReactionToaster.vue
Antreesy: We don't have validation on backend, according to this: https://github.com/nextcloud/spreed/pull/9311#issue-1668282354
Still, any text could be sended with command: `document.querySelector('#localVideoContainer').__vue__.localCallParticipantModel.sendReaction('👎')`
So it seems fine to me to make an additional check here.

marcoambrosini (approved):

src/components/CallView/shared/ReactionToaster.vue
```diff
@@ -0,0 +1,253 @@
...
+	},
+
+	methods: {
+		handleOwnReaction({ model, reaction }) {
+			this.handleReaction(model, reaction)
+		},
+
+		handleReaction(model, reaction) {
+			// prevent spamming to queue from a single account
+			if (this.reactionsQueue.some(item => item.id === model.attributes.peerId)) {
```
nickvergessen, line 128: Unless debug?

Issue #9273, opened 2023-04-06
Translate chat messages


nickvergessen:
- Since Nextcloud 26 a core API to translate strings is available in the server: https://github.com/nextcloud/server/blob/a0ecc37d036ace7bd32675f89b046bf0bff1fe89/core/Controller/TranslationApiController.php#L37
  - API docs at https://github.com/nextcloud/documentation/pull/10047
  - Capability at https://github.com/nextcloud/spreed/pull/9321
- There should be a … Message actions to translate messages.
- A modal opens translating the message to the currently selected user language and allowing in a dropdown to translate to other languages

---

- [ ] Android https://github.com/nextcloud/talk-android/issues/2931
- [x] iOS https://github.com/nextcloud/talk-ios/issues/1185
- [x] Web https://github.com/nextcloud/spreed/issues/9528

nickvergessen:
- API docs at https://github.com/nextcloud/documentation/pull/10047
  - Capability at https://github.com/nextcloud/spreed/pull/9321

jancborchardt:
> A modal opens translating the message to the currently selected user language and allowing in a dropdown to translate to other languages

This sounds a bit complicated – I would rather go with how Instagram and Twitter do it and have a simple "Translate this message" action which only translates it to your language and places it below the original.
(Also talked about with @AndyScherzinger 

We could of course have a modal to translate to other languages too, but that should be a separate second step. Otherwise it’s basically having full DeepL/Gtranslate app in there where you still need to manually do everything → instead of just simply clicking a button.

AndyScherzinger:
I think putting it in the 3dot menu on web or the long-press on mobile would be fine since showing it directly would mean we need to process all messages upfront to detect their original language to then decide when to show the translations action (as in show it when the language is not the one of the device). So putting it in the menu would be a good start.

jancborchardt:
@nickvergessen thanks for the explanation – ok then I’m good with a modal. Could be the same dialog as in Text.

PR #9321, opened 2023-04-17
feat(chat): Add capabilities for translation options

nickvergessen:
### ☑️ Resolves

* Capability and it's docs for #9273 
* Actual API is missing documentation in server, will try to see if they are scheduled or in progress already

### 🖼️ Screenshots

image

### 🏁 Checklist

### DIFF
--- a/lib/Capabilities.php
+++ b/lib/Capabilities.php
@@ -32,26 +32,18 @@ use OCP\Comments\ICommentsManager;
 use OCP\IConfig;
 use OCP\IUser;
 use OCP\IUserSession;
+use OCP\Translation\ITranslationManager;
 
 class Capabilities implements IPublicCapability {
-	protected IConfig $serverConfig;
-	protected Config $talkConfig;
-	protected ICommentsManager $commentsManager;
-	protected IUserSession $userSession;
-	private IAppManager $appManager;
 
 	public function __construct(
-		IConfig $serverConfig,
-		Config $talkConfig,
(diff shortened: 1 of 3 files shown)


lib/Capabilities.php
```diff
@@ -32,26 +32,18 @@
...
-		Config $talkConfig,
-		ICommentsManager $commentsManager,
-		IUserSession $userSession,
-		IAppManager $appManager,
+		protected IConfig $serverConfig,
+		protected Config $talkConfig,
+		protected ICommentsManager $commentsManager,
+		protected IUserSession $userSession,
+		protected IAppManager $appManager,
+		protected ITranslationManager $translationManager,
```
skjnldsv, line 45: image

skjnldsv (approved):

lib/Capabilities.php
```diff
@@ -32,26 +32,18 @@
...
-		IConfig $serverConfig,
-		Config $talkConfig,
-		ICommentsManager $commentsManager,
-		IUserSession $userSession,
-		IAppManager $appManager,
+		protected IConfig $serverConfig,
+		protected Config $talkConfig,
+		protected ICommentsManager $commentsManager,
+		protected IUserSession $userSession,
+		protected IAppManager $appManager,
```
max-nextcloud, line 44: This was private `before`.
Is that change on purpose?

nickvergessen: Yeah, I have protected as it prevents extending

PR #9325, opened 2023-04-17
Fix toasts messages on Fullscreen mode

Antreesy:
### ☑️ Resolves

* Follow up from https://github.com/nextcloud/spreed/pull/9309#issuecomment-1510925486
* Requires https://github.com/nextcloud/nextcloud-dialogs/pull/803
* Set the global selector for toast messages, to render toast messages inside `#content-vue` (which fixes Fullscreen Mode)
* Tested on Webclient

### 🚧 Tasks

- [ ] Code review
- [ ] Visual check

### 🏁 Checklist

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -42,6 +42,7 @@ import debounce from 'debounce'
 import PreventUnload from 'vue-prevent-unload'
 
 import { getCurrentUser } from '@nextcloud/auth'
+import { setGlobalToastOptions } from '@nextcloud/dialogs'
 import { emit, subscribe, unsubscribe } from '@nextcloud/event-bus'
 import { generateUrl } from '@nextcloud/router'
 
@@ -254,6 +255,8 @@ export default {
 		document.addEventListener('visibilitychange', this.changeWindowVisibility)
 
 		this.onResize()
+
+		setGlobalToastOptions({ selector: this.$store.getters.getMainContainerSelector().slice(1) })
 
 		window.addEventListener('unload', () => {
 			console.info('Navigating away, leaving conversation')


src/utils/showToasts.js
```diff
@@ -0,0 +1,69 @@
...
+		}
+	}
+}
+
+export const showError = createToastFunction(baseShowError)
+export const showInfo = createToastFunction(baseShowInfo)
+export const showMessage = createToastFunction(baseShowMessage)
+export const showSuccess = createToastFunction(baseShowSuccess)
+export const showUndo = createToastFunction(baseShowUndo)
+export const showWarning = createToastFunction(baseShowWarning)
```
ShGKme, line 60: **issue (blocking):** `showUndo` is broken

It has 3 arguments: https://github.com/nextcloud/nextcloud-dialogs/blob/master/lib/toast.ts#L203

Antreesy:
* Works now as a partial patch for solving https://github.com/nextcloud/nextcloud-dialogs/issues/380

Will move it to a draft, and see later, if this could be done in the upstream repository

src/utils/showToasts.js
```diff
@@ -0,0 +1,69 @@
...
+} from '@nextcloud/dialogs'
+
+import store from '../store/index.js'
+
+/**
+ * Add a reference to Talk app container ID if no other selectors are provided
+ * This is necessary to render toast messages in Fullscreen mode properly
+ *
+ * @param {Function} baseFunction callback of 'nextcloud/dialogs'
+ */
```
ShGKme, line 44: **issue (blocking):** JSDoc is not correct:
* `baseFunction` is not a callback
* No `@return`
* We may set less base type to note lose auto-complete in IDE


**suggestion:** add return and update `baseFunction description

```js
/**
 * Add a reference to Talk app container ID if no other selectors are provided
 * This is necessary to render toast messages in Fullscreen mode properly
 *
 * @param {function} baseFunction showX function from '@nextcloud/dialogs'
 * @return {function} wrapped baseFunction with a new default selector
 */
```

**suggestion:** using `@template` (not a part of the standard but works). It adds auto-complete for new functions.

```js
/**
 * Add a reference to Talk app container ID if no other selectors are provided
 * This is necessary to render toast messages in Fullscreen mode properly
 *
 * @template {Function} T
 * @param {T} baseFunction showX function from '@nextcloud/dialogs'
 * @return {T} wrapped function with a new default selector
 */
```

More strict typing require `@nextcloud/dialogs` improvements, because is doesn't export types.

src/utils/showToasts.js
ShGKme, line 1: **thought: the module name** 

I'd to name this module with something that more obviously says it exports decorated functions from `@nextcloud/dialog`. `showToast` sounds to me like a function while the module is a set of functions and constants. 

- `dialogs.js`
- `nextcloudDialogs.js`
- `wrappedDialogs.js`

What do you think?

ShGKme (changes requested):
Mostly works, but one function is broken and JSDoc is not complete.

Alternative solution: to fix https://github.com/nextcloud/nextcloud-dialogs/issues/380. Then we would only need to change the default container on fullscreen toggle.

src/App.vue
```diff
@@ -255,6 +256,8 @@ export default {
 
 		this.onResize()
 
+		setGlobalToastOptions({ selector: this.$store.getters.getMainContainerSelector().slice(1) })
+
```
ShGKme, line 260: I haven't tested yet, but it seems, it should be watch. Otherwise, it is set once and doesn't update global selector with the getMainContainerSelector update.

Issue #9248, opened 2023-04-04
Typing indicators


nickvergessen:
- [ ] Original issue: https://github.com/nextcloud/spreed/issues/2094
- [x] Server/API https://github.com/nextcloud/spreed/issues/9272
- [x] Frontend https://github.com/nextcloud/spreed/issues/9271
- [ ] Android https://github.com/nextcloud/talk-android/issues/2930
- [x] iOS https://github.com/nextcloud/talk-ios/issues/1181

nickvergessen:
* Account-wide setting:
  * Similar to the read-status setting a account wide setting is added that enables the sending **and** receiving of those messages. So a user can only see the typing indicator, when they are also sharing their own typing status.
  * Setting is stored in oc_preferences like the read-status but not copied to conversation level as it's not needed for the technical side
  * The user's setting is populated as capability setting
* UI:
  * There should be a single line on top of the input field `X is typing …`
  * Maybe with avatars (depends on the looks of it)
  * Showing up to 3? user names
  * If more users are typing we show `X users are typing`
* Technology:
  * To not overload the PHP server with requests we will communicate this information as signaling messages
  * Internal signaling server support is optional
  * The clients/frontend send a signaling message every X seconds when their chat input is non-empty
  * Receiving clients expire the information after X seconds or when receiving a signaling message that the other side stopped

nickvergessen:
Regarding the questions:

* Jan: If it's not a big performance concern I would say yes to both. Cause then guests have a similarly nice experience as people with accounts
* Me: It's more like they don't have a settings page (yet), so either the option is very visible or it's quite some extra efforts (but might be worth it anyway so they can see the keyboard shortcuts, etc)

jancborchardt:
First mockups:
- The line is shown inbetween the last message and the input field, not shifting any messages above (in case people stop typing)
- Includes avatar(s) and a line in `--color-text-maxcontrast`, with the name in bold
- Single person indicator centers the small avatar horizontally with the bigger avatars in the chat (maybe right-align is also fine). Text is "**Name** is typing"
- Multiple people are simply shown comma-separated and with the overlapped avatars (needs a component at some point), up until 3 people like: "**Marco** and **Nimisha** are typing"
- For more than 3 people, only 3 keep showing as avatars and full name (could be either the ones who have most recently typed or have started typing first, whatever is more performant), and ends with "and X others are typing" like "**Nimisha, Marco, Joas** and 3 others are typing"

[...]

PR #9344, opened 2023-04-20
Typing indicators in chat [web draft]

### LINKED ISSUE #9271: Frontend "Typing indicators"
- Overview issue at https://github.com/nextcloud/spreed/issues/9248

---

- [ ] When the chat input is non empty a signaling message is sent in periodical interval
- [ ] When the signaling messages is received a small max contrast line is shown above the chat input

Antreesy:
### ☑️ Resolves

* Overview at #9248
* Close #9271 

### 🖼️ Screenshots

| 🏡 After |
| --- |
| image |
| image |
| image |

### 🚧 Tasks
API part
- [ ] When the chat input is non empty a signaling message is sent in periodical interval
- [ ] When the signaling messages is received a small max contrast line is shown above the chat input
- [ ] Ability to disable sending / receiving of signals (like reading status)

Visual part
- [x] The line is shown inbetween the last message and the input field, not shifting any messages above (in case people stop typing)
- [x] It includes avatar(s) and a text in --color-text-maxcontrast, with the name in bold
- [x] Single person indicator centers the small avatar horizontally with the bigger avatars in the chat. Text is "Name is typing"
- [ ] Multiple people are simply shown comma-separated and with the overlapped avatars, up until 3 people like: "Marco and Nimisha are typing"
- [x] For more than 3 people, only 3 keep showing as avatars and full name (could be either the ones who have most recently typed or have started typing first, whatever is more performant), and ends with "and X others are typing" like "Nimisha, Marco, Joas and 3 others are typing"

### 🏁 Checklist


src/components/TypingIndicator/TypingIndicator.vue
```diff
@@ -0,0 +1,142 @@
...
+					condensed
+					:condensed-overlap="8"
+					disable-menu
+					disable-tooltip />
+			</div>
+
+			<p class="indicator__main">
+				<b>{{ visibleParticipantsNamesList }}</b>
+				<span v-if="hiddenParticipantsCount" class="indicator__corresponding">
+					{{ n('spreed', 'and %n other are typing...', 'and %n others are typing...', hiddenParticipantsCount) }}
```
nickvergessen, line 41: Doesn't work for translations

Antreesy: Fixed with commit below.

:warning: Please notice that `translate` and `translatePlural` functions escape HTML and sanitize text to prevent XSS attacks by default, so it should be safe to use `v-html` directive. But it's better to be checked: cc @ShGKme 

Link to the library: https://github.com/nextcloud/nextcloud-l10n/blob/1e43c9bd8ba3e814f0a897e86430a77d6cdaacbd/lib/translation.ts#L60-L63

src/components/TypingIndicator/TypingIndicator.vue
```diff
@@ -0,0 +1,153 @@
...
+				return t('spreed', '<strong>{user1}</strong> and <strong>{user2}</strong> are typing...',
+					{ user1, user2 })
+			}
+
+			if (this.typingParticipants.length === 3) {
+				return t('spreed', '<strong>{user1}</strong>, <strong>{user2}</strong> and <strong>{user3}</strong> are typing...',
+					{ user1, user2, user3 })
+			}
+
+			return n('spreed', '<strong>{user1}</strong>, <strong>{user2}</strong>, <strong>{user3}</strong> and %n other are typing...',
```
nickvergessen, line 101: > so it should be safe to use v-html directive

We do that already in some places, yeah.

the problem is this way enables translators to add or remove more HTML.
You can also escape the displaynames yourself and stop the escaping of parameters:
https://github.com/nextcloud/spreed/blob/d41fb8ef032f8fa4a8e2ea48c7eae5f6d60708a9/src/components/MessagesList/MessagesList.vue#L431-L438

src/components/TypingIndicator/TypingIndicator.vue
```diff
@@ -0,0 +1,159 @@
...
+				'{user1}, {user2}, {user3} and %n others are typing...',
+				this.hiddenParticipantsCount, { user1, user2, user3 }, { escape: false })
+		},
+	},
+
+	methods: {
+		prepareNamesList() {
+			return this.visibleParticipants.reverse()
+				.map(participant => this.getParticipantName(participant))
+				.map(name => name ? `<strong>${name}</strong>` : undefined)
```
nickvergessen, line 111: you need to escape the names here

Antreesy: Good point 
image

nickvergessen: > Good point. But shouldn't it be forbidden to set a name with these special characters?

Too late and therefore no.

src/components/TypingIndicator/TypingIndicator.vue
```diff
@@ -0,0 +1,161 @@
...
+
+		indicatorMessage() {
+			if (!this.typingParticipants) {
+				return ''
+			}
+
+			const [user1, user2, user3] = this.prepareNamesList()
+
+			if (this.typingParticipants.length === 1) {
+				return t('spreed', '{user1} is typing...',
```
jancborchardt, line 89: Just a detail here @Antreesy – instead of 3 dots we should use the actual ellipsis character, and with a space between the text and it, like so: ` …`

> Maksim is typing …

Antreesy: Thanks, did it in other PR, but forgot to change it here as well.
Will be shipped with the next commit, when API is merged!

Antreesy:
* Rebased onto master, connected to capabilities from #9455
* Commits from API #9431 are here to not break additional code
* Component was renamed for easier integration with #9430

src/components/SettingsDialog/SettingsDialog.vue
```diff
@@ -60,6 +60,14 @@
 				@update:checked="toggleReadStatusPrivacy">
 				{{ t('spreed', 'Share my read-status and show the read-status of others') }}
 			</NcCheckboxRadioSwitch>
+			<NcCheckboxRadioSwitch id="typing_status_privacy"
```
nickvergessen, line 63: Only if capability is returned

src/components/AvatarWrapper/AvatarWrapper.vue
```diff
@@ -134,7 +139,7 @@ export default {
 			if (this.isDeletedUser) {
 				return 'X'
 			}
-			const customName = this.name !== t('spreed', 'Guest') ? this.name : '?'
+			const customName = this.name && this.name !== t('spreed', 'Guest') ? this.name : '?'
```
nickvergessen, line 142: Is this an effective change? What is the difference and is it needed to do this on stable26 and others?

Antreesy: `this.name &&` check that name is non-empty, like ""

nickvergessen: ah, then I guess it should be backported

nickvergessen (approved):
Seems to work fine

PR #9346, opened 2023-04-21
Allow to mark conversation as unread

### LINKED ISSUE #6808: Allow to mark conversation as unread
### How to use GitHub

---

## Steps to reproduce
1. Have a chat with messages
2. Open it and scroll to the end

### Expected behaviour
* Show an option `Mark as unread`
* Calls the new API: https://github.com/nextcloud/spreed/pull/6807
* Switch away from the conversation if open, to not mark it as read again with the next message polling

### Actual behaviour
Still see "Mark as read" option in the left sidebar which is not functional

## Talk app

**Talk app version:** master

Antreesy:
### ☑️ Resolves

* Fix #6808 

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
`Mark as read` is shown for all conversation | Different buttons for read and unread conversations
image | image

### 🚧 Tasks

- [ ] Code review

### 🏁 Checklist

- [x]  🔖 Capability is added or not needed

### DIFF
--- a/src/services/conversationsService.js
+++ b/src/services/conversationsService.js
@@ -207,6 +207,20 @@ const clearConversationHistory = async function(token) {
 }
 
 /**
+ * Set conversation as unread
+ *
+ * @param {string} token The token of the conversation to be set as unread
+ */
+const setConversationUnread = async function(token) {
+	try {
+		const response = axios.delete(generateOcsUrl('apps/spreed/api/v1/chat/{token}/read', { token }))
+		return response
+	} catch (error) {
+		console.debug('Error while setting the conversation as unread: ', error)
+	}
+}
+
+/**
  * Add a conversation to the favorites
  *
  * @param {string} token The token of the conversation to be favorites
@@ -441,6 +455,7 @@ export {
 	setConversationName,
 	setConversationDescription,
 	clearConversationHistory,
+	setConversationUnread,
 	setConversationPermissions,
 	setCallPermissions,
 	setMessageExpiration,
--- a/src/store/conversationsStore.js
+++ b/src/store/conversationsStore.js
@@ -46,6 +46,7 @@ import {
 	setConversationDescription,
 	deleteConversation,
 	clearConversationHistory,
+	setConversationUnread,
 	setNotificationLevel,
 	setNotificationCalls,
 	setConversationPermissions,
@@ -129,7 +130,7 @@ const mutations = {
 		Vue.delete(state.conversations, token)
 	},
 	/**
-	 * Resets the store to it's original state
+	 * Resets the store to its original state
 	 *
 	 * @param {object} state current store state;
 	 */
@@ -428,15 +429,25 @@ const actions = {
 	},
 
 	async markConversationRead({ commit, getters }, token) {
-		const conversation = Object.assign({}, getters.conversations[token])
-		if (!conversation) {
+		if (!getters.conversations[token]) {
 			return
 		}
 
-		conversation.unreadMessages = 0
-		conversation.unreadMention = false
+		commit('updateUnreadMessages', { token, unreadMessages: 0, unreadMention: false })
+	},
 
-		commit('addConversation', conversation)
+	async markConversationUnread({ commit, dispatch, getters }, { token }) {
+		if (!getters.conversations[token]) {
+			return
+		}
+
+		try {
+			await setConversationUnread(token)
+			commit('updateUnreadMessages', { token, unreadMessages: 1 })
+			await dispatch('fetchConversation', { token })
+		} catch (error) {
+			console.debug('Error while marking conversation as unread: ', error)
+		}
 	},
 
 	async updateLastCommonReadMessage({ commit, getters }, { token, lastCommonReadMessage }) {
--- a/src/store/conversationsStore.js
+++ b/src/store/conversationsStore.js
@@ -124,7 +124,7 @@ const mutations = {
 	 * Deletes a conversation from the store.
 	 *
 	 * @param {object} state current store state;
-	 * @param {object} token the token of the conversation to delete;
+	 * @param {string} token the token of the conversation to delete;
 	 */
 	deleteConversation(state, token) {
 		Vue.delete(state.conversations, token)
@@ -201,7 +201,7 @@ const mutations = {
 
 const actions = {
 	/**
-	 * Add a conversation to the store and index the displayname.
+	 * Add a conversation to the store and index the displayName.
 	 *
 	 * @param {object} context default store context;
 	 * @param {object} conversation the conversation;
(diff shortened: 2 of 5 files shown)


src/store/conversationsStore.js
```diff
@@ -439,6 +440,20 @@ const actions = {
...
 
+	async markConversationUnread({ dispatch, getters }, { token }) {
+		const conversation = Object.assign({}, getters.conversations[token])
+		if (!conversation) {
+			return
+		}
+
+		try {
+			await setConversationUnread(token)
+			dispatch('fetchConversation', { token })
```
nickvergessen, line 451: Instead of dispatching a fetch, can we directly manipulate the data?
Unread counter is set to 1?

… Ah, unreadMention could be a problem?

nickvergessen: would still prefer to see it bold + counter immediate, can turn blue after another server round trip (or we fix the api)

Antreesy:
```suggestion
      conversation.unreadMessages = 1
			dispatch('fetchConversation', { token })
```

Could be patched like this, so if `axios.delete` will return successful response, we could see that immediately, and fetch and update conversation afterwards (because on server it also updates `lastReadMessage` and should be received)

src/store/conversationsStore.js
```diff
@@ -439,6 +440,20 @@ const actions = {
 		commit('addConversation', conversation)
 	},
 
+	async markConversationUnread({ dispatch, getters }, { token }) {
+		const conversation = Object.assign({}, getters.conversations[token])
+		if (!conversation) {
+			return
+		}
```
ShGKme, line 447: **issue (blocking):** `!conversation` is always falsy, because `conversation` is always an object (`thruly`)

This `if` never works and `conversation` is not used outside this `if`.

I assume, the goal was to check if `getters.conversations[token]` exists? Then we should not clone the object. But in this I'm not sure, that we need to return. 

Is it possible that this action will be called to not existed conversation?

src/store/conversationsStore.js
```diff
@@ -439,6 +440,20 @@ const actions = {
...
+
+		try {
+			await setConversationUnread(token)
+			dispatch('fetchConversation', { token })
+		} catch (error) {
+			console.debug('Error while marking conversation as unread: ', error)
+		}
```
ShGKme, line 454: **question:** no `await dispatch('fetchConversation')` 

Is it intended that dispatch is called without await here?

src/store/conversationsStore.js
Antreesy: I copied this from "read" method, so now it should be checked at both places

src/store/conversationsStore.js
Antreesy: Nope, it's my mistake

src/store/conversationsStore.js
Antreesy: So, getter always returns an object or undefined, and I don't see any reason to use Object.assign here,
 and in another method (but we must look at Git blame, to understand when and why it was initially added)

src/store/conversationsStore.js
ShGKme: > I would say it's okay, as refreshing the conversation after doing the actual action should not be blocking?

`await` here has no affect on UI. The marker on UI updates when store update as the result of `fetchConversation` action. 

But without `await` this action resolves before the state is asctually updated and an exception from `fetchConversation` is not cough by this `catch`, making `unhandledrejection`.

marcoambrosini (approved):
tested and works

src/services/conversationsService.js
```diff
@@ -206,6 +206,20 @@ const clearConversationHistory = async function(token) {
...
+/**
+ * Set conversation as unread
+ *
+ * @param {string} token The token of the conversation to be set as unread
+ */
+const setConversationUnread = async function(token) {
+	try {
+		const response = axios.delete(generateOcsUrl('apps/spreed/api/v1/chat/{token}/read', { token }))
+		return response
+	} catch (error) {
```
nickvergessen, line 218: For me catching in the service is a real anti-pattern. at least when the endpoint can return 400 Bad Request etc.

src/store/conversationsStore.js
```diff
@@ -428,15 +429,25 @@ const actions = {
...
+	async markConversationUnread({ commit, dispatch, getters }, { token }) {
+		if (!getters.conversations[token]) {
+			return
+		}
+
+		try {
+			await setConversationUnread(token)
+			commit('updateUnreadMessages', { token, unreadMessages: 1 })
+			await dispatch('fetchConversation', { token })
+		} catch (error) {
```
nickvergessen, line 448: catching again? you are already catching in the `setConversationUnread`?
Whatever happens to be catched here is not expected and should be console.error at least.

Maybe you at least want to move the first 2 calls out of the try?

src/services/conversationsService.js
Antreesy: But we do it almost for every service in this file, what's the difference?

src/store/conversationsStore.js
Antreesy: Here I'm agree, that we could omit try-catch blocks:
`setConversationUnread` - catches errore in servise.js
`commit('updateUnreadMessages')`  - mutation, for which we have checked before that conversation exists in `if` block
`dispatch('fetchConversation')` - also catches errors by its own

src/services/conversationsService.js
marcoambrosini: We've been doing try and catch mostly in store and components, that's true

src/store/conversationsStore.js
```diff
@@ -19,6 +19,7 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  *
  */
+import { isEmpty } from 'lodash'
```
ShGKme, line 22: **issue (blocking):** Lodash does not support tree shaking.

```js
import { isEmpty } from 'lodash'
``` 
adds about **71.7kB** to bundle.

```js
import isEmpty from 'lodash/isEmpty.js'
```
would add **6.7kB** and actually adds **0.14kB** because we already have it with `isEqual`.

```suggestion
import isEmpty from 'lodash/isEmpty.js'
```

ShGKme (changes requested):

src/store/conversationsStore.js
ShGKme: > `dispatch('fetchConversation')` - also will catch errors by its own

It re-throw it

https://github.com/nextcloud/spreed/blob/077a584c86a2c4e1942e91835647b47382738a41/src/store/conversationsStore.js#L633-L638

But it's ok to not catch exception here to me, because we don't show anything with `showError`, don't check that `setConversationUnread` was actually successful, and don't check is all action was successful on the call.

src/services/conversationsService.js
Antreesy: Then i'd prefer to keep it consistent for now, and move discussion to the new issue

for conversationService: 14 / 26 requests use try/catch
for conversationStore: 7 / ..40 actions use try/catch

PR #9377, opened 2023-04-25
feat(controllers): Migrate server annotations to attributes

nickvergessen:
### ☑️ Resolves

* Server ref https://github.com/nextcloud/server/pull/37905

### 🏁 Checklist

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -41,6 +41,8 @@ use OCP\App\IAppManager;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\Attribute\BruteForceProtection;
+use OCP\AppFramework\Http\Attribute\NoCSRFRequired;
+use OCP\AppFramework\Http\Attribute\PublicPage;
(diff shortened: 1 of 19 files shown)


lib/Controller/PageController.php
```diff
@@ -141,51 +142,41 @@ public function showCall(string $token): Response {
...
-	 * @NoCSRFRequired
-	 *
 	 * @param string $token
 	 * @param string $callUser
 	 * @param string $password
 	 * @return TemplateResponse|RedirectResponse
 	 * @throws HintException
 	 */
+	#[NoCSRFRequired]
+	#[PublicPage]
```
vitormattos, line 179: Maybe would be good put all ordered.

vitormattos (approved):

lib/Controller/PageController.php
vitormattos: We also can verify if possible create a rule at friendsofphp/php-cs-fixer about this.

nickvergessen: it's alphabetically ordered?

> We also can verify if possible create a rule at nextcloud/coding-standard about this.

I guess that would be cool

Issue #7708, opened 2022-08-11
Replace input fields with TextField, PasswordField and NumberField component


nickvergessen:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

- [ ] apps/spreed/src/components  (2 usages found)
    - [ ] SetGuestUsername.vue  (1 usage found)
        - [ ]  36 `<input ref="usernameInput"`
    - [ ] UploadEditor.vue  (1 usage found)
        - [ ]  31 `<input id="file-upload"`
- [ ] apps/spreed/src/components/AdminSettings  (12 usages found)
    - [ ] HostedSignalingServer.vue  (3 usages found)
        - [ ]  37 `<input v-model="hostedHPBNextcloudUrl"`
        - [ ]  45 `<input v-model="hostedHPBFullName"`
        - [ ]  53 `<input v-model="hostedHPBEmail"`
    - [ ] SignalingServer.vue  (2 usages found)
        - [ ]  25 `<input ref="signaling_server"`
    - [ ] SignalingServers.vue  (2 usages found)
        - [ ]  74 `<input v-model="secret"`
    - [ ] SIPBridge.vue  (1 usage found)
        - [ ]  55 `<input v-model="sharedSecret"`
    - [ ] StunServer.vue  (1 usage found)
        - [ ]  27 `<input ref="stun_server"`
    - [ ] TurnServer.vue  (2 usages found)
        - [ ]  41 `<input ref="turn_server"`
        - [ ]  51 `<input ref="turn_secret"`
- [ ] apps/spreed/src/components/ConversationSettings  (9 usages found)
    - [ ] LinkShareSettings.vue  (3 usages found)
        - [ ]  63 `<input id="link_share_settings_link_password"`
- [ ] apps/spreed/src/components/LeftSidebar/NewGroupConversation/PasswordProtect  (1 usage found)
    - [ ] PasswordProtect.vue  (1 usage found)
        - [ ]  23 `<input ref="password"`
- [ ] apps/spreed/src/components/LeftSidebar/NewGroupConversation/SetContacts  (1 usage found)
    - [ ] SetContacts.vue  (1 usage found)
        - [ ]  26 `<input ref="setContacts"`
[...]

PR #9384, opened 2023-04-26
Refactor Administrator settings for Talk

Antreesy:
### ☑️ Resolves

* Part of #7708
* Replace native inputs with NcTextField
* Replace outdated NcMultiSelects and native selects with NcSelect 
* Unify styles for each section (minor visual changes)

### 🖼️ Screenshots

| 🏚️ Before / 🏡 After |
|---|
| image |
| image |
| image |
| image |
| image |
| image |
| image |

### 🚧 Tasks

- [ ] Code review
- [ ] Approve of visual changes

### 🏁 Checklist


src/components/AdminSettings/SIPBridge.vue
```diff
@@ -162,18 +163,31 @@ export default {
...
-	&__shared-secret,
-	&__dialin-info {
-		width: 480px;
+.form {
+	&__textfield,
+	&__select {
+		width: 300px;
+		:deep(.vs__search) {
+			width: 100%;
+		}
```
ShGKme, line 186: What is `vs__search` element? I haven't found it...

ShGKme (commented):
Looks very good to me, nice work! Only some minor comments

src/components/AdminSettings/SIPBridge.vue
Antreesy: There is a hidden input inside `NcSelect` to search through the options

src/components/AdminSettings/SIPBridge.vue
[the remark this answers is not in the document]
Antreesy: We have kind of 'subsection' in several places. How about highlight them, and replace others for standard labels?
image

src/components/AdminSettings/SIPBridge.vue
ShGKme: Oh, this is a class from `vue-select` library. But if it is hidden, why should it have overridden width?

Antreesy: Because it adds horizontal scrolled and change height

Antreesy:
Guess, some movements of AdminSettings could be safely implemented in this PR as a part of https://github.com/nextcloud/spreed/issues/9402

ShGKme:
> We have kind of 'subsection' in general settings, maybe that's what bother you.

What bothers me is:
- Is "Dial-in information" a sub-section header or a textarea's label?
- Is "Restrict SIP configuration" a subsection header or a selects's?

> Replaced with labels, screenshots are updated

Now the additional part of a label is moved from the label to the input message.

Could we keep the original look and change only the structure?

image

ShGKme:
> Guess, some movements of AdminSettings could be safely implemented in this PR as a part of https://github.com/nextcloud/spreed/issues/9402

But these components are not shared across the applications, they are used only in one very separated feature - AdminSettings.

Antreesy:
> Now the additional part of a label is moved from the label to the input message.

We have styling like that already (small hint description under the select in "Limit to groups" section), so it's fine by me to use same approach here. 
image

src/views/AdminSettings.vue
```diff
@@ -69,33 +69,4 @@ export default {
...
-		vertical-align: middle;
-	}
-
-	.icon-checkmark-color.hidden {
-		display: none;
-	}
-
-	.error {
-		border-color: var(--color-error);
-	}
```
ShGKme, line 99: Were those files also redundant? 😀

Antreesy: icons from material-design-icons library either wrapped in the NcButton and already have these styles or have different classes, like `check-icon` instead of `icon-checkmark`
In both cases, they're not usable anymore

Antreesy: same for error handling, we need to use `&.error :deep(.input-field__input)` to adjust input appearance

ShGKme (approved):
🕸️

PR #9407, opened 2023-05-01
Add documentation of signaling messages

SystemKeeper:
This is mostly a shameless copy of @danxuliu's work over at [the android client](https://github.com/nextcloud/talk-android/blob/master/app/src/main/java/com/nextcloud/talk/signaling/SignalingMessageReceiver.java), but I think we should have this documented here as well.
I only documented the field explicitly that had a comment on the android side before. There's room for improvements here for sure.

### 🏁 Checklist


nickvergessen:
I guess that's replacing the todo here?
https://github.com/nextcloud/spreed/blob/master/docs/internal-signaling.md#internal-signaling-api

SystemKeeper:
> I guess that's replacing the todo here?
> https://github.com/nextcloud/spreed/blob/master/docs/internal-signaling.md#internal-signaling-api

Actually I think that this is something different. Because the internal signaling API would be the endpoint to send / receive messages through the API, isn't it? This PR includes the format of the messages, both for internal and external signaling.
But I think it would be good to link this document there as well?!

PR #9410, opened 2023-05-01
Use deep selector

marcoambrosini:
// FIXME: add a description for this pr

https://vuejs.org/api/sfc-css-features.html#scoped-css
https://stackoverflow.com/questions/48032006/how-do-i-use-deep-or-or-v-deep-in-vue-js

### DIFF
--- a/src/components/NewMessageForm/SimplePollsEditor/SimplePollsEditor.vue
+++ b/src/components/NewMessageForm/SimplePollsEditor/SimplePollsEditor.vue
@@ -193,17 +193,15 @@ export default {
 }
 
 // Upstream
-::v-deep .checkbox-radio-switch {
-	&__label {
-		align-items: unset;
-		height: unset;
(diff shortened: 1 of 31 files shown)


src/components/NewMessageForm/SimplePollsEditor/SimplePollsEditor.vue
```diff
@@ -193,17 +193,15 @@ export default {
...
+	align-items: unset;
+	height: unset;
+	margin: 4px 0;
+	padding: 8px;
+	width: 100%;
+	border-radius: var(--border-radius-large);
+	span {
+		align-self: flex-start;
 	}
 }
```
ShGKme, line 206: Here is the same as in `src/components/MessagesList/MessagesGroup/Message/MessagePart/Poll.vue`

Antreesy:
Rebased and signed off a commit

PR #9411, opened 2023-05-01
Fix diverging user status

### LINKED ISSUE #8744: User status in topbar and conversation list is not synchronized
### How to use GitHub

---

## Steps to reproduce
1. Join a conversation and leave the conversation open

### Expected behaviour
When the conversation list is updated and we're in a 1:1 conversation, the status should be reflected in the top bar as well. Interestingly enough the message is updated, but the icon is not in this case?

## Talk app

**Talk app version:** 15.0.3 on c.nc.c

nickvergessen:
### ☑️ Resolves

* Fix #8744

### 🚧 Tasks

- [x] clearAt was invalid type (and only handed in to the event so far)
- [x] clearAt was missing from docs
- [x] emit event to update NcAvatar's of the user when we have a new status from the room list

### 🏁 Checklist


Antreesy (approved):
Tested on web client

lib/Controller/RoomController.php
```diff
@@ -818,10 +818,11 @@ protected function formatParticipantList(array $participants, bool $includeStatu
 				}
 
 				if (isset($statuses[$userId])) {
+					$clearAt = $statuses[$userId]->getClearAt();
 					$result['status'] = $statuses[$userId]->getStatus();
 					$result['statusIcon'] = $statuses[$userId]->getIcon();
 					$result['statusMessage'] = $statuses[$userId]->getMessage();
-					$result['statusClearAt'] = $statuses[$userId]->getClearAt();
+					$result['statusClearAt'] = $clearAt ? $clearAt->getTimestamp() : null;
```
vitormattos, line 825: Only suggestion:

If you like, you can change at other places of this PR but isn't requirement to merge this PR.

Since PHP 8.0 we can use nullsafe operator. This would be good to be used, is the same behavior when we to this at JS:

```javascript
err?.response?.data?.ocs?.data?.message
```

Ref: https://wiki.php.net/rfc/nullsafe_operator

```suggestion
					$result['statusClearAt'] = $statuses[$userId]->getClearAt()?->getTimestamp();
```

docs/conversation.md
```diff
@@ -100,6 +100,7 @@
...
 | `statusMessage`         | string  | v4    |         | Optional: Only available for one-to-one conversations and when  `includeStatus=true` is set                                                                                                                                                                                                                                                                                                       |
+| `statusClearAt`         | int     | v4    |         | Optional: Only available for one-to-one conversations and when  `includeStatus=true` is set                                                                                                                                                                                                                                                                                                       |
```
vitormattos, line 103: I think that would be good talk that this value can be null. Isn't only integer every time.

lib/Controller/RoomController.php
nickvergessen: yeah, but we backport this, so can't use it for now. But we can change in master afterwards

PR #9413, opened 2023-05-02
fix(permissions): Reset custom permissions on promotion

### LINKED ISSUE #9298: Custom permission flag shown for moderators
### How to use GitHub

---

## Steps to reproduce
1. Invite a user to a room
2. Grant "all permissions"
3. Promote to moderator

### Expected behaviour
Looks like a normal moderator

### Actual behaviour
Is a moderator but also shows custom permissions icon, but there is no way to reset:
image

nickvergessen:
### ☑️ Resolves

* Fix #9298 

### 🚧 Tasks

- [x] Reset permissions on promotion
- [x] Add a repair step to execute on update

### 🏁 Checklist

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -16,7 +16,7 @@ And in the works for the [coming versions](https://github.com/nextcloud/spreed/m
 
 	]]></description>
 
-	<version>17.0.0-dev</version>
+	<version>17.0.0-dev.1</version>
 	<licence>agpl</licence>
 
 	<author>Daniel Calviño Sánchez</author>
--- a/lib/Migration/Version16000Date20230502145340.php
+++ b/lib/Migration/Version16000Date20230502145340.php
@@ -1,0 +1,59 @@
+<?php
+
+declare(strict_types=1);
+
+/**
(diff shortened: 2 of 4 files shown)


appinfo/info.xml
```diff
@@ -73,6 +73,7 @@ And in the works for the [coming versions](https://github.com/nextcloud/spreed/m
 			<step>OCA\Talk\Migration\CreateHelpCommand</step>
 			<step>OCA\Talk\Migration\ClearResourceAccessCache</step>
 			<step>OCA\Talk\Migration\CacheUserDisplayNames</step>
+			<step>OCA\Talk\Migration\ResetCustomPermissionsForModerators</step>
```
vitormattos, line 76: Need to be repair step?
Isn't this a pontual change?

nickvergessen: yeah, shouldn't happen anymore after the code change, unless we miss something.

vitormattos: Was the other steps created with the same intention to run only a time?

nickvergessen: No, they are "safety nets" to repair undesired admin actions, etc.

PR #9422, opened 2023-05-02
Support call sounds on Safari

SystemKeeper:
### ☑️ Resolves

Until now we don't support call sounds on Safari. This has 2 reasons:
* Safari does not support .ogg
* Safari does only allow to play sounds if this is due to a user interaction

To solve the first issue we also provide FLAC files for the sounds which are working fine on Safari. To solve the second issue we create the audio object when we press the start call button. As this is done inside of a click event handler, the audio objects are free to play at any time. 
While this approach in theory also works iOS, it does not work anymore as soon as you grant camera and microphone permissions there - so we keep the note about it not being supported on those devices.

Follow-up to https://github.com/nextcloud/spreed/pull/5410
Tested on
* Safari 16.3
* Chrome 112
* Firefox 112

### 🏁 Checklist


img/COPYING
```diff
@@ -5,3 +5,11 @@
 ## leave_call.ogg
 @author Marco Ambrosini (marcoambrosini@icloud.com) 
 @license CC0-1.0
+
+## join_call.flac
```
nickvergessen, line 9: you add 3 files but only 2 licenses?

src/store/soundsStore.js
```diff
@@ -55,6 +60,76 @@ const mutations = {
 	setUserId(state, userId) {
 		state.userId = userId
 	},
+
+	/**
+	 * Sets the join audio object
+	 *
```
nickvergessen, line 66:
```suggestion
	 *
```

> ### GitHub Actions / eslint
>
> Trailing spaces not allowed                                              


Same for the other docs.

src/store/soundsStore.js
```diff
@@ -55,6 +60,76 @@ const mutations = {
...
+		state.audioObjectsCreated = true
+	},
+}
+
+/**
+ * Helper method to get the full file path of an audio file in spreed/img.
+ * Example: getFullAudioFilepath('myAudio') => spreed/img/myAudio.ogg
+ * 
+ * @param {string} fileName The name of the file without extension
+ * @returns The full path to the file in the spreed/img directory and adds .ogg/.flac depending on supported codecs
```
nickvergessen, line 109:
```suggestion
 * @return The full path to the file in the spreed/img directory and adds .ogg/.flac depending on supported codecs
```

src/store/soundsStore.js
```diff
@@ -55,6 +60,76 @@ const mutations = {
...
+ */
+const getFullAudioFilepath = function(fileName) {
+	const tempAudio = new Audio()
+
+	// Prefer the .ogg version of sounds, but fall back to .flac in case .ogg is not supported (Safari)
+	if (tempAudio.canPlayType("audio/ogg")) {
```
nickvergessen, line 115:
```suggestion
	if (tempAudio.canPlayType('audio/ogg')) {
```

src/store/soundsStore.js
```diff
@@ -55,6 +60,76 @@ const mutations = {
...
+	} else {
+		return generateFilePath('spreed', 'img', fileName + '.flac')
+	}
+}
+
+/**
+ * Creates a HTMLAudioElement with the specified filePath and loads it
+ * 
+ * @param {string} filePath Path to the file
+ * @returns The created and loaded HTMLAudioElement
```
nickvergessen, line 126:
```suggestion
 * @return {HTMLAudioElement} The created and loaded HTMLAudioElement
```

img/COPYING
SystemKeeper: We did not specify a license for `LibremPhoneCall.ogg` before, so I did not add it for the FLAC version.

marcoambrosini (approved):
tested and works

marcoambrosini (approved):
tested and works

marcoambrosini (approved):
tested and works

img/COPYING
SystemKeeper: @marcoambrosini any idea about the third license ?

marcoambrosini: No, that file was already there when I added the join and leave call sounds.

SystemKeeper:
@nickvergessen Should this be backported?

nickvergessen:
I think it would be quite a surprise for most users in a patch release. We can of course bump the minor version to indicate something a like, but otherwise I'm also fine with not backporting

PR #9429, opened 2023-05-02
chore(front): make unit-testing cross-platform

ShGKme:
### ☑️ Resolves

Bash ENV doesn't work on some other shells, for example, CMD or PowerShell (Windows).
Also, there is a problem with different slashes on Unix/Windows.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks

- [x] Use global setup to setup ENV instead of bash
- [x] Add stubs for `.tflite` files instead of blob to `src/utils/media/effects/virtual-background/vendor/*`
- [x] Refactoring: use the stub transformer for styles

### 🏁 Checklist


nickvergessen:
> Bash ENV doesn't work on some other shells, for example, CMD or PowerShell (Windows).
Also, there is a problem with different slashes on Unix/Windows.

Windows is not supported as server (dev) environment *cough*
https://docs.nextcloud.com/server/latest/admin_manual/installation/system_requirements.html#server

That being said, I think the timezone UTC was needed for some tests

ShGKme:
> Windows is not supported as server (dev) environment cough

Yes. But this is not a part of the server, this is frontend, that is also used in the desktop client.

> I think the timezone UTC was needed for some tests

I did not remove it. I only moved it from bash to the Jest global setup

Antreesy (approved):
:seal: 
Tests which require UTC timezone, are still working, style stubs didn't break anything also.

PR #9433, opened 2023-05-03
Updates lib/Config.php PHPDoc comments.

fsamapoor:
Updates PHPDoc comments in order to make the parameters match their declared types.


nickvergessen (approved):
Could also just remove those doc blocks, but either way is fine for me

Issue #8444, opened 2022-12-07
Ability to set an emoji as a conversation avatar


e-caste:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

**Is your feature request related to a problem? Please describe.**
Whatsapp allows to set an emoji as a group chat icon. I think this could be a very nice UX addition for 2 reasons:
1. facilitate quick group chat creation (without having to choose a picture which takes longer)
2. make easily recognisable group chats (without having to sacrifice the precious characters in the title to make it start with an emoji)

**Describe the solution you'd like**
While creating a group chat, the user can spawn an emoji picker and choose from those.  
At that point either:
1. the selected emoji gets rasterized and saved as image on the server or
2. the corresponding image for the selected emoji is chosen from a high resolution emoji pack such as https://twemoji.twitter.com/ which also contains SVG versions  

the image is then used as the conversation avatar.

Whatsapp UI:
1. tap the icon button
![IMG_8441](https://user-images.githubusercontent.com/48513706/206276065-9048f29b-118d-43b4-9e0d-15d28f4c6534.jpg)
2. choose "Emoji and Sticker"
![IMG_8442](https://user-images.githubusercontent.com/48513706/206276061-0915af9a-1c61-4983-bb1c-8792031b94c3.jpg)
3. choose emoji and background color
![IMG_8443](https://user-images.githubusercontent.com/48513706/206276070-6c7fcd9b-238e-49bc-83f2-78ea14a53ea3.PNG)

**Describe alternatives you've considered**
Emoji in the chat title, manually picked picture as a conversation avatar.

**Additional context**
Issue opened as requested by @nickvergessen in #927.

nickvergessen:
So as a quick plan the following idea came up:
* If the room name contains an emoji as first character, we will use that emoji and return it as an SVG on the conversation avatar endpoint
* The background of the emoji is automatically generated based on the full displayname of the room, so multiple rooms with the same emoji could still have a different icon for easier distinction.

PR #9443, opened 2023-05-04
feat(avatar): Allow to select an emoji + color as avatar

nickvergessen:
Ref #8444 

### 🚧 Tasks

- [x] Allow picking emoji and color
- [x] Fallback to dark/bright mode if no color is given
- [x] Document API
- [x] Write integration test

### 🏁 Checklist

### DIFF
--- a/lib/Controller/AvatarController.php
+++ b/lib/Controller/AvatarController.php
@@ -27,6 +27,7 @@ declare(strict_types=1);
 namespace OCA\Talk\Controller;
 
 use InvalidArgumentException;
+use OCA\Mail\Service\Avatar\Avatar;
 use OCA\Talk\Middleware\Attribute\RequireModeratorParticipant;
 use OCA\Talk\Middleware\Attribute\RequireParticipant;
 use OCA\Talk\Service\AvatarService;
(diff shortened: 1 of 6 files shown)


lib/Controller/AvatarController.php
```diff
@@ -78,12 +79,39 @@ public function uploadAvatar(): DataResponse {
...
 	#[NoCSRFRequired]
 	#[RequireParticipant]
 	public function getAvatar(bool $darkTheme = false): Response {
 		$file = $this->avatarService->getAvatar($this->getRoom(), $this->userSession->getUser(), $darkTheme);
 
+		if ($file->getMimeType() === 'image/svg+xml' && str_contains($file->getContent(), AvatarService::THEMING_PLACEHOLDER)) {
+			$color = $darkTheme ? AvatarService::THEMING_DARK_BACKGROUND : AvatarService::THEMING_BRIGHT_BACKGROUND;
+			$file->putContent(str_replace(AvatarService::THEMING_PLACEHOLDER, $color, $file->getContent()));
+		}
+
```
vitormattos, line 114: Will be best put this inside the previous method: `avatarService->getAvatar`

lib/Service/AvatarService.php
```diff
@@ -81,7 +85,40 @@ public function setAvatarFromRequest(Room $room, ?array $file): void {
...
+		foreach ($avatarFolder->getDirectoryListing() as $file) {
+			$file->delete();
+		}
+
+		$avatarName = $this->random->generate(16, ISecureRandom::CHAR_HUMAN_READABLE) . '.svg';
+		$avatarFolder->newFile($avatarName, $svgContent);
+		$this->roomService->setAvatar($room, $avatarName);
+	}
+
+	public function setAvatar(Room $room, \OC_Image $image, bool $allowSvg = false): void {
```
vitormattos, line 121: Maybe other name then my suggestion would be best but the idea is to rename this method because isn't more the unique method to set avatar.
```suggestion
	public function setAvatarFromBinary(Room $room, \OC_Image $image, bool $allowSvg = false): void {
```

vitormattos (approved):

lib/Service/AvatarService.php
nickvergessen: The thing is we can't have a \OC_Image as it does not support SVG. And then the differences are so much, that I'm not sure it's worth to enforce them into 1 method

vitormattos: My suggestion isn't to use only a method, is only to rename the actual to `setAvatarFromBinary` and maintain the new as `setSVGAvatar`. Only to don't have a generic method called `setAvatar`

vitormattos (changes requested):

vitormattos (approved):

PR #9462, opened 2023-05-05
Add messageId to richObject in ReferenceProvider

SystemKeeper:
This adds the messageId to the richObject provided by the reference provider in case a link to a message is provided.

Server PR for definition: https://github.com/nextcloud/server/pull/38102

### 🏁 Checklist

### DIFF
--- a/lib/Collaboration/Reference/TalkReferenceProvider.php
+++ b/lib/Collaboration/Reference/TalkReferenceProvider.php
@@ -177,6 +177,7 @@ class TalkReferenceProvider extends ADiscoverableReferenceProvider implements IS
 		$roomName = $room->getDisplayName($this->userId);
 		$title = $roomName;
 		$description = '';
+		$messageId = null;
 
 		if ($participant instanceof Participant
 			|| $this->roomManager->isRoomListableByUser($room, $this->userId)) {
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):

lib/Collaboration/Reference/TalkReferenceProvider.php
```diff
@@ -238,6 +240,7 @@ protected function fetchReference(Reference $reference): void {
 			'name' => $roomName,
 			'link' => $reference->getUrl(),
 			'call-type' => $this->getRoomType($room),
+			'message-id' => $messageId,
```
nickvergessen, line 243: instead of adding with null, we could also skip it in that case:

```php
		$referenceData = [
			'id' => $room->getToken(),
			'name' => $roomName,
			'link' => $reference->getUrl(),
			'call-type' => $this->getRoomType($room),
		];

		if ($messageId) {
			$referenceData['message-id'] = $messageId;
		}

		$reference->setRichObject('call', $referenceData);
```

SystemKeeper: I like that, will change accordingly!

nickvergessen (approved):

PR #9484, opened 2023-05-06
Fullscreen - use Material Design icons

starypatyk:
### ☑️ Resolves

* Align icons with **Viewer** (https://github.com/nextcloud/viewer/pull/1588#issuecomment-1510802067)

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
obraz | obraz
obraz | obraz

### 🚧 Tasks

- [x] Use Material Design icons
- [ ] Adjust spelling

### 🏁 Checklist

### DIFF
--- a/src/components/TopBar/TopBarMenu.vue
+++ b/src/components/TopBar/TopBarMenu.vue
@@ -79,10 +79,13 @@
 		</NcActionButton>
 
 		<!-- Fullscreen -->
-		<NcActionButton :icon="iconFullscreen"
-			:aria-label="t('spreed', 'Toggle fullscreen')"
+		<NcActionButton :aria-label="t('spreed', 'Toggle fullscreen')"
 			:close-after-click="true"
 			@click="toggleFullscreen">
+			<template #icon>
+				<Fullscreen v-if="!isFullscreen" :size="20" />
+				<FullscreenExit v-else :size="20" />
+			</template>
 			{{ labelFullscreen }}
 		</NcActionButton>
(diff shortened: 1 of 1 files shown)


src/components/TopBar/TopBarMenu.vue
```diff
@@ -79,10 +79,13 @@
 		</NcActionButton>
 
 		<!-- Fullscreen -->
-		<NcActionButton :icon="iconFullscreen"
```
nickvergessen, line 82: You can delete the iconFullscreen computed too?

starypatyk: Indeed. My omission - a bit embarrassing. Thanks for pointing this out. :+1: Done in https://github.com/nextcloud/spreed/pull/9484/commits/c9adfe233edd86b4892580b981ecf0fc1eb9b15b

I am a bit concerned about the `isInCall` check that was in the removed code. It seems that the icon is always in a menu with a light background - even during a call. So, I hope this will be fine.
obraz

SystemKeeper: > It seems that the icon is always in a menu with a light background - even during a call

When using dark mode, the menu is dark as well:
image

starypatyk: I have checked in dark mode. It is OK.

In conversation:
obraz

During a call:
obraz

nickvergessen (approved):

PR #9509, opened 2023-05-10
fix(chat): Prevent empty chat messages

### LINKED ISSUE #9507: Empty messages should not be possible to send
### How to use GitHub

---

## Steps to reproduce
1. Open a conversation
2. Press spacebar 
3. Press return

### Expected behaviour
Nothing should happen, because the message is empty (after trim).

### Actual behaviour
The message is actually send as an empty message. This happens since switching to `NCRichContentEditable` (I think). Corresponding issue of the vue-lib https://github.com/nextcloud/nextcloud-vue/issues/4068.
Not sure if we should prevent this on frontend or backend.

## Talk app

Master, stable26

nickvergessen:
### ☑️ Resolves

* Fix #9507

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
With only new lines the submit button is active | With only new lines the submit button is inactive
Bildschirmfoto vom 2023-05-10 07-17-54 | Bildschirmfoto vom 2023-05-10 07-17-35

### 🏁 Checklist

### DIFF
--- a/src/components/NewMessageForm/NewMessageForm.vue
+++ b/src/components/NewMessageForm/NewMessageForm.vue
@@ -439,7 +439,7 @@ export default {
 		},
 
 		hasText() {
-			return this.text !== ''
+			return this.text.trim() !== ''
 		},
 
 		containerElement() {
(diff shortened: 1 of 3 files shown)


vitormattos (approved):

src/components/NewMessageForm/NewMessageForm.vue
```diff
@@ -592,7 +592,7 @@ export default {
 				}
 			}
 
-			if (this.text !== '') {
+			if (this.hasText) {
 				// FIXME: remove after issue is resolved: https://github.com/nextcloud/nextcloud-vue/issues/3264
 				const temp = document.createElement('textarea')
 				temp.innerHTML = this.text
```
Antreesy, line 598:
```suggestion
				temp.innerHTML = this.text.trim()
```
We trim messages at server, probably, so temporary message could be aligned to it as well

nickvergessen: Fixed

Antreesy (approved):

PR #9512, opened 2023-05-10
Translate chat messages

### LINKED ISSUE #9528: Translate chat messages [web frontend]
- Overview issue https://github.com/nextcloud/spreed/issues/9273

---

- Since Nextcloud 26 a core API to translate strings is available in the server: https://github.com/nextcloud/server/blob/a0ecc37d036ace7bd32675f89b046bf0bff1fe89/core/Controller/TranslationApiController.php#L37
  - API docs at https://github.com/nextcloud/documentation/pull/10047
  - Capability at https://github.com/nextcloud/spreed/pull/9321
- There should be a … Message actions to translate messages.
- A modal opens translating the message to the currently selected user language and allowing in a dropdown to translate to other languages

marcoambrosini:
### ☑️ Resolves

* Fix #9528

### Requisites

* Download `nextcloud/translate` and install according to Readme: https://github.com/nextcloud/translate
* If models are not downloaded with occ command, it could be done manually: https://github.com/nextcloud-releases/translate/releases/tag/v1.1.3 (and put in translate/models/ afterwards)

### 🖼️ Screenshots

image
image

### 🚧 Tasks

- [ ] Code review
- [ ] Visual check
- [ ] Handle rich content properly

### 🏁 Checklist

### DIFF
--- a/src/components/ChatView.vue
+++ b/src/components/ChatView.vue
@@ -108,8 +108,6 @@ export default {
 	mounted() {
 		// Postpone render of NewMessageForm until application is mounted
 		this.containerId = this.$store.getters.getMainContainerSelector()
-
-		this.$store.dispatch('getTranslationConfiguration')
 	},
 
 	methods: {
--- a/src/components/ChatView.vue
+++ b/src/components/ChatView.vue
@@ -108,6 +108,8 @@ export default {
 	mounted() {
 		// Postpone render of NewMessageForm until application is mounted
 		this.containerId = this.$store.getters.getMainContainerSelector()
+
+		this.$store.dispatch('getTranslationConfiguration')
 	},
 
 	methods: {
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -179,6 +179,7 @@ the main body of the message as well as a quote.
 		<div class="message-body__scroll">
 			<MessageButtonsBar v-if="showMessageButtonsBar"
 				ref="messageButtonsBar"
+				:is-translation-available="isTranslationAvailable"
 				:is-action-menu-open.sync="isActionMenuOpen"
 				:is-emoji-picker-open.sync="isEmojiPickerOpen"
 				:is-reactions-menu-open.sync="isReactionsMenuOpen"
@@ -193,8 +194,14 @@ the main body of the message as well as a quote.
 				:common-read-icon-tooltip="commonReadIconTooltip"
 				:show-sent-icon="showSentIcon"
 				:sent-icon-tooltip="sentIconTooltip"
+				@show-translate-dialog="isTranslateDialogOpen = true"
 				@delete="handleDelete" />
 		</div>
+
+		<MessageTranslateDialog v-if="isTranslateDialogOpen"
+			:text="message"
+			:rich-parameters="richParameters"
+			@close="isTranslateDialogOpen = false" />
 
 		<div v-if="isLastReadMessage"
 			v-observe-visibility="lastReadMessageVisibilityChanged">
@@ -225,6 +232,7 @@ import NcRichText from '@nextcloud/vue/dist/Components/NcRichText.js'
 import Quote from '../../../Quote.vue'
 import CallButton from '../../../TopBar/CallButton.vue'
 import MessageButtonsBar from './MessageButtonsBar/MessageButtonsBar.vue'
+import MessageTranslateDialog from './MessageButtonsBar/MessageTranslateDialog.vue'
 import Contact from './MessagePart/Contact.vue'
 import DeckCard from './MessagePart/DeckCard.vue'
 import DefaultParameter from './MessagePart/DefaultParameter.vue'
@@ -245,6 +253,7 @@ export default {
 	name: 'Message',
 
 	components: {
+		MessageTranslateDialog,
 		CallButton,
 		MessageButtonsBar,
 		NcButton,
@@ -422,6 +431,7 @@ export default {
 			isReactionsMenuOpen: false,
 			isForwarderOpen: false,
 			detailedReactionsLoading: false,
+			isTranslateDialogOpen: false,
 		}
 	},
 
@@ -580,7 +590,9 @@ export default {
 		},
 
 		showMessageButtonsBar() {
-			return !this.isSystemMessage && !this.isDeletedMessage && !this.isTemporary && (this.isHovered || this.isActionMenuOpen || this.isEmojiPickerOpen || this.isFollowUpEmojiPickerOpen || this.isReactionsMenuOpen || this.isForwarderOpen)
+			return !this.isSystemMessage && !this.isDeletedMessage && !this.isTemporary
+				&& (this.isHovered || this.isActionMenuOpen || this.isEmojiPickerOpen || this.isFollowUpEmojiPickerOpen
+					|| this.isReactionsMenuOpen || this.isForwarderOpen || this.isTranslateDialogOpen)
 		},
 
 		isTemporaryUpload() {
@@ -649,6 +661,10 @@ export default {
 
 		detailedReactionsLoaded() {
 			return this.$store.getters.reactionsLoaded(this.token, this.id)
+		},
+
+		isTranslationAvailable() {
+			return this.$store.getters.getTranslationAvailableLanguages()?.length !== 0
 		},
 	},
 
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -221,6 +221,7 @@ import CheckAll from 'vue-material-design-icons/CheckAll.vue'
 import EmoticonOutline from 'vue-material-design-icons/EmoticonOutline.vue'
 import Reload from 'vue-material-design-icons/Reload.vue'
 
+import { getCapabilities } from '@nextcloud/capabilities'
 import { showError, showSuccess, showWarning, TOAST_DEFAULT_TIMEOUT } from '@nextcloud/dialogs'
 import moment from '@nextcloud/moment'
 
@@ -245,6 +246,8 @@ import { useIsInCall } from '../../../../composables/useIsInCall.js'
 import { ATTENDEE, CONVERSATION, PARTICIPANT } from '../../../../constants.js'
 import participant from '../../../../mixins/participant.js'
 import { EventBus } from '../../../../services/EventBus.js'
+
+const isTranslationAvailable = getCapabilities()?.spreed?.config?.chat?.translations?.length > 0
 
 /**
  * @property {object} scrollerBoundingClientRect provided by MessageList.vue
@@ -413,7 +416,7 @@ export default {
 
 	setup() {
 		const isInCall = useIsInCall()
-		return { isInCall }
+		return { isInCall, isTranslationAvailable }
 	},
 
 	data() {
@@ -661,10 +664,6 @@ export default {
 
 		detailedReactionsLoaded() {
 			return this.$store.getters.reactionsLoaded(this.token, this.id)
-		},
-
-		isTranslationAvailable() {
-			return this.$store.getters.getTranslationAvailableLanguages()?.length !== 0
 		},
 	},
 
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -199,7 +199,7 @@ the main body of the message as well as a quote.
 		</div>
 
 		<MessageTranslateDialog v-if="isTranslateDialogOpen"
-			:text="message"
+			:message="message"
 			:rich-parameters="richParameters"
 			@close="isTranslateDialogOpen = false" />
 
--- a/src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
@@ -106,6 +106,15 @@
 					@click="action.callback(messageApiData)">
 					{{ action.label }}
 				</NcActionButton>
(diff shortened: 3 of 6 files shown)


src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -580,7 +596,8 @@ export default {
 		},
 
 		showMessageButtonsBar() {
-			return !this.isSystemMessage && !this.isDeletedMessage && !this.isTemporary && (this.isHovered || this.isActionMenuOpen || this.isEmojiPickerOpen || this.isFollowUpEmojiPickerOpen || this.isReactionsMenuOpen || this.isForwarderOpen)
+			return true
+			// return !this.isSystemMessage && !this.isDeletedMessage && !this.isTemporary && (this.isHovered || this.isActionMenuOpen || this.isEmojiPickerOpen || this.isFollowUpEmojiPickerOpen || this.isReactionsMenuOpen || this.isForwarderOpen)
```
nickvergessen, line 600:
```suggestion
			return !this.isSystemMessage && !this.isDeletedMessage && !this.isTemporary && (this.isHovered || this.isActionMenuOpen || this.isEmojiPickerOpen || this.isFollowUpEmojiPickerOpen || this.isReactionsMenuOpen || this.isForwarderOpen)
```

but I guess you want to add your dialog as well? is it `showTranslateDialog` ? Could it also be named `isTranslateOpen` or something alike?

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
```diff
@@ -559,6 +581,10 @@ export default {
 				return EmojiIndex.emoji(emojiStrings).native
 			})
 		},
+
+		handleTranslate() {
+			return 1
```
nickvergessen, line 586: ?

src/components/TranslateDialog.vue
```diff
@@ -0,0 +1,194 @@
...
+			<label for="from">
+				{{ t('spreed', 'From') }}
+			</label>
+			<NcSelect v-model="selectedFrom"
+				input-id="from"
+				:aria-label="t('spreed', 'From:')"
+				:placeholder="t('spreed', 'From:')"
+				:options="availableFroms"
+				:clearable="false"
+				track-by="from"
```
nickvergessen, line 35: track-by does not exist in NcSelect? At least it did not for me yesterday

src/components/TranslateDialog.vue
```diff
@@ -0,0 +1,194 @@
...
+			<label for="to">
+				{{ t('spreed', 'To') }}
+			</label>
+			<NcSelect v-model="selectedTo"
+				input-id="to"
+				:aria-label="t('spreed', 'To:')"
+				:placeholder="t('spreed', 'To:')"
+				:options="availableTos"
+				:clearable="false"
+				track-by="To"
```
nickvergessen, line 47: If track-by still exists thi sneeds to be lower case?

src/components/TranslateDialog.vue
```diff
@@ -0,0 +1,194 @@
...
+				:clearable="false"
+				track-by="from"
+				no-wrap
+				@input="handleSelectFrom" />
+			<label for="to">
+				{{ t('spreed', 'To') }}
+			</label>
+			<NcSelect v-model="selectedTo"
+				input-id="to"
+				:aria-label="t('spreed', 'To:')"
```
nickvergessen, line 43: Not a valid aria label.

```suggestion
				:aria-label="t('spreed', 'Target language to translate into')"
```

similar thing for the from block

src/components/TranslateDialog.vue
```diff
@@ -0,0 +1,194 @@
...
+			type: String,
+			required: true,
+		},
+	},
+
+	data() {
+		return {
+			translatedMessage: '',
+			availableTos: [],
+			selectedFrom: '',
```
nickvergessen, line 105: selectedFrom should be null and send to the server, similar to the iOS app to try to detect the origin language?

nickvergessen (changes requested):

src/components/TranslateDialog.vue
Antreesy: It worked for NcMultiSelect before, my bad

nickvergessen: Yeah, but NcMultiselect is deprecated because it's not accessible. So going with NcSelect is correct, you just need to map the "track-by" atttribute in your data to the id field

Antreesy:
So, i get a successful responce on axios.get(generateOcsUrl('/translation/languages')):
`{"ocs":{"meta":{"status":"ok","statuscode":200,"message":"OK"},"data":{"languages":[{"from":"de","fromLabel":"German","to":"en","toLabel":"English"},{"from":"en","fromLabel":"English","to":"de","toLabel":"German"}],"languageDetection":true}}}`

But when trying to translate, get an error:
`{"ocs":{"meta":{"status":"failure","statuscode":400,"message":""},"data":{"message":"Unable to translate","from":"en"}}}`

src/components/TranslateDialog.vue
```diff
@@ -0,0 +1,194 @@
...
+			selectedTo: '',
+			loading: false,
+		}
+	},
+
+	computed: {
+		availableFroms() {
+			return Array.from(new Set(this.availableTranslations.map(element => element.from)))
+		},
+	},
```
Antreesy, line 115:
```js
		optionsFrom() {
			return this.availableFroms.map(element => {
				const model = this.availableTranslations.find(language => language.from === element)
				return {
					id: model.from,
					label: model.fromLabel,
				}
			})
		},

		optionsTo() {
			return this.availableTranslations.filter(language => language.from === this.selectedFrom?.id)
				.map(model => {
					return {
						id: model.to,
						label: model.toLabel,
					}
				})
		},
```

Something like this could be used for Select options

src/components/TranslateDialog.vue
```diff
@@ -0,0 +1,194 @@
...
+
+<style lang="scss" scoped>
+.translate-dialog {
+	display: flex;
+	flex-direction: column;
+	min-height: 400px;
+	padding: calc(var(--default-grid-baseline) * 4);
+	&__message {
+		margin: calc(var(--default-grid-baseline) * 4);
+	}
```
Antreesy, line 183:
```suggestion
  &-translation {
    border: 1px solid var(--color-primary-element);
    border-radius: var(--border-radius-large);
    }
}
```

Maybe also wrap it with some border?

src/components/TranslateDialog.vue
Antreesy: Suggested changes for Select options, see below

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageTranslateDialog.vue
```diff
@@ -0,0 +1,295 @@
...
+				<NcRichText class="translate-dialog__message translate-dialog__message-translation"
+					:text="translatedMessage"
+					:arguments="richParameters"
+					:reference-limit="0" />
+				<NcButton class="translate-dialog__copy-button"
+					@click="handleCopyTranslation">
+					<template #icon>
+						<ContentCopy />
+					</template>
+					{{ t('spreed', 'Copy translation') }}
```
Antreesy, line 72:
```suggestion
					{{ t('spreed', 'Copy translated text) }}
```
cc @nimishavijay

src/store/messagesStore.js
```diff
@@ -1313,6 +1334,18 @@ const actions = {
 	async easeMessageList(context, { token }) {
 		context.commit('easeMessageList', { token })
 	},
+
+	async getTranslationConfiguration(context) {
+		try {
+			const response = await getTranslationLanguages()
```
nickvergessen, line 1340: We populate this via initial state, so you could save the request (and we have a feature toggle already ready?)

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageTranslateDialog.vue
```diff
@@ -0,0 +1,295 @@
...
+		},
+
+		async translateMessage(sourceLanguage = null) {
+			try {
+				this.isLoading = true
+				const response = await translateText(this.text, sourceLanguage, this.selectedTo?.id)
+				this.translatedMessage = response.data.ocs.data.text
+			} catch (error) {
+				console.error(error)
+				showError(t('spreed', 'The message could not be translated'))
```
nickvergessen, line 231: The returned error object actually has an error message in the user's language that would be cool to show.

```suggestion
				showError(error.response?.data?.ocs?.data?.message ?: t('spreed', 'The message could not be translated'))
```

src/components/ChatView.vue
```diff
@@ -108,6 +108,8 @@ export default {
 	mounted() {
 		// Postpone render of NewMessageForm until application is mounted
 		this.containerId = this.$store.getters.getMainContainerSelector()
+
+		this.$store.dispatch('getTranslationConfiguration')
```
nickvergessen, line 112: This means we load the configuration when switching chat to call and back?
Anyway as mentioned in the other comment we have a capability with the value that should be used

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageTranslateDialog.vue
```diff
@@ -0,0 +1,295 @@
...
+			} finally {
+				this.isLoading = false
+			}
+		},
+
+		async handleCopyTranslation() {
```
nickvergessen, line 237: neat feature ✨

nimishavijay:
I messed around a bit more with how this could look and I'd like to propose one more design change: use a border instead of a background color. 

- [ ] Give --border-radius-large border radius to both the texts
- [ ] Use a --color-border border for the source text
- [ ] Use a --color-primary-element border for the target text
- [ ] use --color-main-background background for both the texts



what do you think? :)

src/store/messagesStore.js
nickvergessen: hmm okay wouldn't work for desktop unless that loads it itself?

Antreesy: Not aware about toggle, but yes, restore it as initial, getting from capabilities

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageTranslateDialog.vue
```diff
@@ -0,0 +1,295 @@
...
+				this.isLoading = true
+				const response = await translateText(this.message, sourceLanguage, this.selectedTo?.id)
+				this.translatedMessage = response.data.ocs.data.text
+			} catch (error) {
+				console.error(error)
+				showError(error.response?.data?.ocs?.data?.message ?? t('spreed', 'The message could not be translated'))
+			} finally {
+				this.isLoading = false
+			}
+		},
```
ShGKme, line 234: On the first modal open it always fails to me from the request in `mounted`:

Request:

```js
fetch("https://nextcloud.dev.shgk.me/ocs/v2.php/translation/translate", {
  "headers": {
    "accept": "application/json, text/plain, */*",
    "accept-language": "en-US,en;q=0.9,ru;q=0.8",
    "content-type": "application/json",
    "requesttoken": "<DELETED>",
    "sec-ch-ua": "\"Chromium\";v=\"112\", \"Not:A-Brand\";v=\"99\"",
    "sec-ch-ua-mobile": "?0",
    "sec-ch-ua-platform": "\"Windows\"",
    "sec-fetch-dest": "empty",
    "sec-fetch-mode": "cors",
    "sec-fetch-site": "same-origin"
  },
  "referrerPolicy": "no-referrer",
  "body": "{\"text\":\"Hallo!\",\"fromLanguage\":null,\"toLanguage\":\"en\"}",
  "method": "POST",
  "mode": "cors",
  "credentials": "include"
});
```

Response:

```json
{"ocs":{"meta":{"status":"failure","statuscode":400,"message":""},"data":{"message":"Could not detect language"}}}
```

Antreesy: It's on API side then, we always send text and 'to' language

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
```diff
@@ -354,9 +365,14 @@ export default {
 			type: String,
 			required: true,
 		},
+
+		isTranslationAvailable: {
+			type: Boolean,
+			required: true,
+		},
```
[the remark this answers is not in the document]
Antreesy: Could be done later, when we'll finally reach Message refactoring =)

ShGKme (approved):

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageTranslateDialog.vue
ShGKme: Yep, it seems, API is used correctly

nickvergessen: That's what I said in the office. You send a single word and fromLanguage is null.
But on a single word language models don't work and respond with a language, so we don't have a from when talking to translations provider and it returns with could not translate

ShGKme: Maybe we should not show the error message in this case?
Otherwise, all small messages show an error toast on opening the translation modal.
Also, if only one language is available, we can pre-select it.

nickvergessen: Well we could try to hide the error from empty runs yeah, but let's do that in a follow up

nickvergessen (approved):

PR #9513, opened 2023-05-10
fix(call): Fix incorrectly selected media when actually joining a call

### LINKED ISSUE #9501: Blur state is inverted in media settings on page load
### How to use GitHub

---

Bildschirmfoto vom 2023-05-09 11-03-59

## Steps to reproduce
1. Start a call for the first time in a room
2. See blur on the preview and continue

### Expected behaviour
Blur is on

### Actual behaviour
Blur is off

## Talk app

**Talk app version:** Master and Talk desktop

Assuming regression from https://github.com/nextcloud/spreed/issues/9270

nickvergessen:
### ☑️ Resolves

* Fix #9501 
* Ref is https://github.com/nextcloud/spreed/commit/27105c9259e862d2b154600d8b51026b368b6367
* But the change was not fully done, so the webrtc code still checked without `videoDisabled_2790450198` but the media setting set `nextcloud_per_dGFsaw==_videoDisabled_2790450198` now both paths are consistent.

### ❓ Questions
* We could also fix it the other way around and basically revert https://github.com/nextcloud/spreed/commit/27105c9259e862d2b154600d8b51026b368b6367 so it also writes without `nextcloud_per_dGFsaw==_` again. The advantage would be that existing settings would be reused, which I would actually prefer.
* Or in case the value is null, we check the old browser storage name to fallback. But that might be very cumbersome.

### 🏁 Checklist

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -275,12 +275,12 @@ export default {
 			if (this.isInCall) {
 				this.$store.dispatch('setForceCallView', true)
 
-				const enableAudio = !localStorage.getItem('audioDisabled_' + this.token)
-				const enableVideo = !localStorage.getItem('videoDisabled_' + this.token)
-				const enableVirtualBackground = !!localStorage.getItem('virtualBackgroundEnabled_' + this.token)
-				const virtualBackgroundType = localStorage.getItem('virtualBackgroundType_' + this.token)
-				const virtualBackgroundBlurStrength = localStorage.getItem('virtualBackgroundBlurStrength_' + this.token)
-				const virtualBackgroundUrl = localStorage.getItem('virtualBackgroundUrl_' + this.token)
+				const enableAudio = !BrowserStorage.getItem('audioDisabled_' + this.token)
+				const enableVideo = !BrowserStorage.getItem('videoDisabled_' + this.token)
+				const enableVirtualBackground = !!BrowserStorage.getItem('virtualBackgroundEnabled_' + this.token)
+				const virtualBackgroundType = BrowserStorage.getItem('virtualBackgroundType_' + this.token)
+				const virtualBackgroundBlurStrength = BrowserStorage.getItem('virtualBackgroundBlurStrength_' + this.token)
+				const virtualBackgroundUrl = BrowserStorage.getItem('virtualBackgroundUrl_' + this.token)
 
 				EventBus.$once('joined-conversation', async ({ token }) => {
 					if (params.token !== token) {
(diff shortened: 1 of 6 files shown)


danxuliu (commented):
[`localStorage` is still used for media state in _App.vue_](https://github.com/nextcloud/spreed/blob/425fbcbe574dbb94c8cb37b8be7f0ed42d78bc3a/src/App.vue#L278-L319) (to restore the state when switching to another room while in a call).

It is also used in a [store test](https://github.com/nextcloud/spreed/blob/04c973ed6aee1d8113ccf44d91b5431b954d6c70/src/store/callViewStore.spec.js), but I have not checked for what (nor why the test do not fail), and to [save the nick of guests](https://github.com/nextcloud/spreed/blob/15b0436b20bc1b5240069575d4f6b0e0cdc40b25/src/components/SetGuestUsername.vue#L104-L106).

Regarding whether to fully embrace `BrowserStorage` or go back to `localStorage` so existing settings are reused I have no preference; I guess that the move to `BrowserStorage` was done for some reason, so fine to move it, but reuse the existing settings is good, so fine to go back as well :-P

nickvergessen:
> and to save the nick of guests.

Yeah that has to stay, so text, collabora and talk use the same name.

Just trying to fix the regression between media settings and the actual call for now.

nickvergessen:
For the background i guess the idea was only consistency
https://github.com/nextcloud/spreed/pull/9347#discussion_r1176979062
So might also be okay to revert for the time being and then fully migrate with a transition in the future?

Cc @ShGKme

ShGKme:
> For the background I guess the idea was only consistency

At least consistency, as we have an abstraction for Nextcloud apps. The same as with Axios where we always use `@nextcloud/axios` instead of any other HTTP client or custom axios extension.

In the general case, it also helps to keep values in a specific app scope, less global between apps.

> So might also be okay to revert for the time being and then fully migrate with a transition in the future?

Is the only problem - to keep the current local (in browser) settings for local media settings?

nickvergessen:
> Is the only problem - to keep the current local (in browser) settings for local media settings?

Well that would be why "reverting" would be better then "going forward with this PR".

> At least consistency, as we have an abstraction for Nextcloud apps.

I get that, but it's also kind of annoying to break the setting.

But either thing needs to be done, so webrtc + media settings save/read from the right place. Currently whatever you do in the media settings is ignored in the call view and vice-versa

nickvergessen:
But okay, then let's go forward with this PR :)

nickvergessen:
> It is also used in a [store test](https://github.com/nextcloud/spreed/blob/04c973ed6aee1d8113ccf44d91b5431b954d6c70/src/store/callViewStore.spec.js), but I have not checked for what (nor why the test do not fail),

So that works because BrowserStorage still uses localStorage and all the keys just have a prefix but the regex of the keys only checks for the end.
https://github.com/nextcloud/spreed/blob/04c973ed6aee1d8113ccf44d91b5431b954d6c70/src/store/callViewStore.spec.js#L101
If that would be changed to `/^callprefs-XXTOKENXX-isgrid$/` it would fail with:
```
    Expected: StringMatching /^callprefs-XXTOKENXX-isgrid$/
    Received: "nextcloud_per_dGFsaw==_callprefs-XXTOKENXX-isgrid"
```
But the missing leading `^` means it just checks for the end, so all good.

nickvergessen:
> It is also used in a [store test](https://github.com/nextcloud/spreed/blob/04c973ed6aee1d8113ccf44d91b5431b954d6c70/src/store/callViewStore.spec.js), but I have not checked for what (nor why the test do not fail),

@danxuliu So that works because BrowserStorage still uses localStorage and all the keys just have a prefix but the regex of the keys only checks for the end.
https://github.com/nextcloud/spreed/blob/04c973ed6aee1d8113ccf44d91b5431b954d6c70/src/store/callViewStore.spec.js#L101
If that would be changed to `/^callprefs-XXTOKENXX-isgrid$/` it would fail with:
```
    Expected: StringMatching /^callprefs-XXTOKENXX-isgrid$/
    Received: "nextcloud_per_dGFsaw==_callprefs-XXTOKENXX-isgrid"
```
But the missing leading `^` means it just checks for the end, so all good.

src/App.vue
```diff
@@ -240,6 +240,8 @@ export default {
 	},
 
 	beforeMount() {
+		this.migrateDirectLocalStorageToNextcloudBrowserStorage()
+
```
ShGKme, line 244: `migrateDirectLocalStorageToNextcloudBrowserStorage` is not a vue component method. It doesn't use any component's properties, it is not used as a component's property. It should be a function outside of the component.

A small drawback of using it as a component method is that the very first time it is called is only after the root application component is already initialized and created. So we may lose the value if it is got from the storage before (vuex store init, global application init, most of other imported ES modules).

Suggestion: https://github.com/nextcloud/spreed/blob/master/src/init.js

nickvergessen: Like this?

src/init.js
```diff
@@ -84,3 +85,44 @@ EventBus.$on('signaling-recording-status-changed', (token, status) => {
...
+ * BrowserStorage
+ */
+const migrateDirectLocalStorageToNextcloudBrowserStorage = () => {
+	const storageKeys = Array.from(Array(localStorage.length), (i) => localStorage.key(i)).filter(key => key.startsWith('audioDisabled_')
+		|| key.startsWith('videoDisabled_')
+		|| key.startsWith('virtualBackgroundEnabled_')
+		|| key.startsWith('virtualBackgroundType_')
+		|| key.startsWith('virtualBackgroundBlurStrength_')
+		|| key.startsWith('virtualBackgroundUrl_')
+	)
```
ShGKme, line 103: A very minor performance note: Web Storage API is a synchronous Browser API. Reading 300-400 entries takes about 0.2ms of blocking time. Even without `filter` (no matter if `Object.keys` or `localStorage.key` is used).

An alternative could be to store only one flag with `isMigrated` or migration version.

src/init.js
```diff
@@ -84,3 +85,44 @@ EventBus.$on('signaling-recording-status-changed', (token, status) => {
...
+
+/**
+ * Migrate localStorage to @nextcloud/browser-storage
+ *
+ * In order to preserve the user settings while migrating to the abstraction,
+ * we loop over the localStorage entries and add the matching ones to the
+ * BrowserStorage
+ */
+const migrateDirectLocalStorageToNextcloudBrowserStorage = () => {
+	const storageKeys = Array.from(Array(localStorage.length), (i) => localStorage.key(i)).filter(key => key.startsWith('audioDisabled_')
```
ShGKme, line 97: **issue (blocking):** Sorry, I made a typo on the last suggestion, `i` must be index, not item.

```suggestion
	const storageKeys = Array.from(Array(localStorage.length), (_, i) => localStorage.key(i)).filter(key => key.startsWith('audioDisabled_')
```

Or, if `Array.from` looks complex:

```suggestion
	const storageKeys = Array(localStorage.length).fill().map((_, i) => localStorage.key(i)).filter(key => key.startsWith('audioDisabled_')
```

ShGKme (changes requested):

ShGKme:
Except for one issue, all works fine and looks good.

Also checked that no other storage item is missed.

ShGKme (approved):

PR #9530, opened 2023-05-12
(follow-up) Minor javascript refactoring

Antreesy:
### ☑️ Resolves

* From https://github.com/nextcloud/spreed/pull/9430#discussion_r1191165370 and https://github.com/nextcloud/spreed/pull/9430#discussion_r1192518267
* Replace 'kinda' radio-buttons at NotificationSettings with `NcCheckboxRadioSwitch`

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks

- [ ] Code review
- [ ] Visual check

### 🏁 Checklist

### DIFF
--- a/src/components/CallView/Grid/Grid.vue
+++ b/src/components/CallView/Grid/Grid.vue
@@ -438,7 +438,7 @@ export default {
 		// Hides or displays the `grid-navigation next` button
 		hasNextPage() {
 			if (this.displayedVideos.length !== 0 && this.hasPagination) {
-				return this.displayedVideos[this.displayedVideos.length - 1] !== this.videos[this.videos.length - 1]
+				return this.displayedVideos.at(-1) !== this.videos.at(-1)
 			} else {
 				return false
 			}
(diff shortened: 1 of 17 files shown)


nickvergessen (approved):

ShGKme:
I re-checked, and I was a bit wrong in [that comment](https://github.com/nextcloud/spreed/pull/9430#discussion_r1192518267).

It is not directly mentioned in the Vue documentation, but if a custom event **has a native event object as a payload**, its native event modifiers still **do work** on custom component events (most of the modifiers, incl. `prevent`, `exact`, `stop`). 

[Prooflink](https://v2.template-explorer.vuejs.org/#%3CComp%20%40click.stop.prevent%20%2F%3E)

Although, it changes almost nothing. I checked all the cases. In all the cases `.prevent` was useless because its target had no default event handler anyway (`button[type="button"]`, `div`, `esc` for `input`). The only difference is that `exact`. Now these buttons in the danger zone work also with `CTRL`/`SHIFT`/`ALT` key modifiers. 

image

cc @Antreesy

PR #9596, opened 2023-05-23
Fix Fullscreen mode support in ViewerOverlay

### LINKED ISSUE #9444: Viewer overlay doesn't support full screen mode
### How to use GitHub

---

## Steps to reproduce
1. Start a call
2. Turn on full-screen mode
3. Open a document

### Expected behaviour

You can see a video in the viewer overlay

### Actual behaviour

The viewer overlay is hidden above the viewer

ShGKme:
### ☑️ Resolves

* Fix #9444

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks

In the Fullscreen mode, Talk Vue app (`#content-vue`) is a [top layer](https://developer.mozilla.org/en-US/docs/Glossary/Top_layer) element. It is not possible to display any other elements outside the Talk app on top of the top layer using CSS.

Both Viewer and ViewerOverlayCallView are rendered outside the Talk app and had this problem. With Viewer it is solved now by a `fixme` hack with manual moving a DOM element with Viewer to the Talk app.

This PR fixes this problem in a similar way, in a fullscreen mode `Teleport` is used to move the ViewerOverlay not to the body, but to the Talk app (`#content-vue`).

A do not like this solution. Formally this is not correct to use Portal/Teleport to move something **inside** the Vue app. It should only be used to teleport something **outside** the app and Vue app content should never be mutated directly without Vue and its Virtual DOM.

Alternative solutions could be:
1. Use the [full version of the Portal library](https://v2.portal-vue.linusb.org)
   - Requires adding a heavy library
   - Doesn't solve the same problem with Viewer anyway
2. Create and use a new child of `#content-vue` as Talk Vue app, so its parent could be in the Fullscreen allowing to render new elements there
   - Changes the Talk app layout by making `#content-vue` not a vue app or changes the Talk app server-side template. I'm afraid, it could break many things. I'd consider doing something like these in techdept later.

So I haven't found any better solution :( 

### 🏁 Checklist

### DIFF
--- a/src/components/CallView/shared/ViewerOverlayCallView.vue
+++ b/src/components/CallView/shared/ViewerOverlayCallView.vue
@@ -19,7 +19,16 @@
 
 <template>
 	<div ref="ghost" class="viewer-overlay-ghost">
-		<Portal>
+		<!--
+			Viewer Overlay should be teleported to be to the top of DOM to be on top of the Viewer,
+			because by default Viewer is on top of an entire Talk (#content-vue).
+			In the fullscreen mode Viewer is manually moved to #content-vue which is top layer by Fullscreen API.
+			FIXME: this is not correct to use Portal/Teleport to move something inside the Vue app.
+			Alternative solutions could be:
+			- Use full version of the Portal library (doesn't solve the same problem with Viewer)
+			- Use a new child of #content-vue as Talk Vue app
+		-->
+		<Portal :selector="isFullscreen ? '#content-vue' : 'body'">
 			<!-- Add .app-talk to use Talk icon classes outside of #content-vue -->
 			<div class="viewer-overlay app-talk"
 				:style="{
@@ -181,6 +190,10 @@ export default {
 	computed: {
 		conversation() {
 			return this.$store.getters.conversation(this.token)
+		},
+
+		isFullscreen() {
+			return this.$store.getters.isFullscreen()
 		},
 	},
 


ShGKme:
Tested in a call with and without participants in normal and Fullscreen modes

nickvergessen:
Work when:
1. Fullscreen
2. Join call
3. Open viewer

Does not work when:
1. Join call
2. Open viewer
3. Fullscreen

ShGKme:
> Does not work when:
> 
> 1. Join call
> 2. Open viewer
> 3. Fullscreen

Yes. But this is not related to the viewer overlay. It didn't work even without overlay, because the viewer is moved only once on opening.

Do you prefer to fix this problem in this PR together with the overlay bug?

ShGKme:
Hmm, Portal also doesn't react to the target selector update. So a little bit related.

ShGKme:
A new commit adds support to the fullscreen mode toggle after overlay opening.

But it still requires a fix for the same issue with the viewer.

src/components/CallView/shared/ViewerOverlayCallView.vue
```diff
@@ -19,7 +19,17 @@
...
+			Viewer Overlay should be teleported to be to the top of DOM to be on top of the Viewer,
+			because by default Viewer is on top of an entire Talk (#content-vue).
+			In the fullscreen mode Viewer is manually moved to #content-vue which is top layer by Fullscreen API.
+			FIXME: this is not correct to use Portal/Teleport to move something inside the Vue app.
+			Alternative solutions could be:
+			- Use full version of the Portal library (doesn't solve the same problem with Viewer)
+			- Use a new child of #content-vue as Talk Vue app
+		-->
+		<!-- Also Portal's selector is not reactive. We need to re-mount the node on selector change using key -->
+		<Portal :selector="portalSelector" :key="portalSelector">
```
nickvergessen, line 32: GitHub Actions / eslint

Attribute ":key" should go before ":selector"

PR #9607, opened 2023-05-24
Fix Viewer with Fullscreen mode toggle

ShGKme:
### ☑️ Resolves

* Fix https://github.com/nextcloud/spreed/pull/9596#issuecomment-1560640396

#### Fix 

To work correctly in the Fullscreen mode, the Viewer should be mounted in Talk Vue app (`#content-vue`). Currently, it is moved from the body to `#content-vue` only once on Viewer open. It doesn't help when Fullscreen mode is enabled or disabled after the Viewer was opened.

See also: https://github.com/nextcloud/spreed/pull/6879

This PR moves the Viewer to `watch`.

#### Remove a 1-second timeout

Currently, there is a 1000ms timeout to wait until Viewer was mounted. But 1 second too much.

This PR uses `nextTick` to wait for one Vue scheduler iteration is enough.
Now opening a file is almost 1000ms faster in Talk 😀

### 🏁 Checklist

### DIFF
--- a/src/composables/useViewer.js
+++ b/src/composables/useViewer.js
@@ -17,6 +17,8 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
+import { nextTick, watch } from 'vue'
+
 import { useIsInCall } from './useIsInCall.js'
 import { useStore } from './useStore.js'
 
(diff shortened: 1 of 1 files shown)


ShGKme:
This PR is on top of https://github.com/nextcloud/spreed/pull/9596 to test with both issues resolved.

Rebase and merge after that PR merge.

src/composables/useViewer.js
```diff
@@ -71,11 +73,26 @@ export function useViewer() {
 		// opened, so for the time being it is reparented if needed shortly
 		// after calling it.
 		// @see https://github.com/nextcloud/viewer/issues/995
-		setTimeout(() => {
-			if (store.getters.isFullscreen()) {
-				document.getElementById('content-vue').appendChild(document.getElementById('viewer'))
-			}
-		}, 1000)
+		watch(
+			() => store.getters.isFullscreen(),
```
nickvergessen, line 77: Is watching better, or would it make sense to emit an event in https://github.com/nextcloud/spreed/blob/47e98d86d617925253f69f4a5a7ad67d63fd4d79/src/store/windowVisibilityStore.js#L76-L78

ShGKme: IMO, `watch` is better:
- This is a Vue way to react to a reactive state's change
- No need to manually remove a listener on clean-up, watch is managed by Vue + Handler is managed by Vue's scheduler
- Using an event emitter (a bus) for communication between Vue components is kind of an official anti-pattern in Vue (not for communication between different applications on the page). Here we have a direct and implicit dependency on isFullscreen state. With a global event, we have a star (as topologically)

nickvergessen: Roger

nickvergessen:
Time to rebase?

nickvergessen (approved):

PR #9608, opened 2023-05-24
fix(mediasettings): Hide virtual background options when not supported

### LINKED ISSUE #9597: Don't show background options on Safari as it's unsupported
### How to use GitHub

---

image

Since Safari does not support WASM, the blur option is disabled. But also the background options should be disabled (or the tab should be hidden), because it won't work as well).

## Talk app

**Talk app version:** 17 RC1 and master

nickvergessen:
### ☑️ Resolves

* Fix #9597 

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
grafik | Bildschirmfoto vom 2023-05-24 16-49-48

### 🏁 Checklist

### DIFF
--- a/src/components/MediaSettings/MediaSettings.vue
+++ b/src/components/MediaSettings/MediaSettings.vue
@@ -97,7 +97,8 @@
 					</template>
 					{{ t('spreed', 'Devices') }}
 				</NcButton>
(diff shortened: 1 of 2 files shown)


SystemKeeper (approved):
Tested and works

PR #9625, opened 2023-05-25
fix(TypingIndicator): Actors are only unique by type+id

nickvergessen:
### 🏁 Checklist

### DIFF
--- a/src/components/NewMessage/NewMessageTypingIndicator.vue
+++ b/src/components/NewMessage/NewMessageTypingIndicator.vue
@@ -70,7 +70,8 @@ export default {
 
 		typingParticipants() {
 			return this.$store.getters.participantsListTyping(this.token).filter(participant => {
-				return participant.actorId !== this.$store.getters.getActorId()
+				return participant.actorType !== this.$store.getters.getActorType()
+					|| participant.actorId !== this.$store.getters.getActorId()
 			})
 		},
 


src/components/NewMessage/NewMessageTypingIndicator.vue
```diff
@@ -70,7 +70,8 @@ export default {
 
 		typingParticipants() {
 			return this.$store.getters.participantsListTyping(this.token).filter(participant => {
-				return participant.actorId !== this.$store.getters.getActorId()
+				return participant.actorType !== this.$store.getters.getActorType()
+					|| participant.actorId !== this.$store.getters.getActorId()
```
ShGKme, line 74: I'm not sure, I understand what is our goal...

Am I right, that:
1. We have types: `users`, `guests`, `bots`.
2. In a group all of them have unique actorId, but bots may have the same actorId as users
3. So we need to use type+id as an identifier here to separate some bot and the current user.

If I understand it correctly, then this code seems weird to me. It doesn't make `type+id`. It seems to me that here `actorType !== getActorType()` is used as `actorType !== 'users'` to check if `actorType === 'bots'.

ShGKme: I meant, `participant.actorType !== this.$store.getters.getActorType()` doesn't look like `unique by type+id` for me.

Did I understand something wrong?

SystemKeeper: I think the idea is to filter out myself here. So it’s not myself if it’s another actorId or another actorType?!

nickvergessen: Exactly.

In theory there could be a type=group id=admin and type=user id=admin
We don't want to filter out that group admin would be typing, just because there id is the same as my user id "admin".

Now with groups that might not work, but a guest has id=md5/sha1(ofSomeData) and your ldap user backend could be configured to be using random or md5 or such data as well and it could match.

ShGKme (approved):

PR #9657, opened 2023-05-29
Add background job to check TURN certificate

### LINKED ISSUE #9606: Expired SSL cert of turnserver still shows green checkmark in admin settings
### How to use GitHub

---

## Steps to reproduce
1. Have a turnserver
2. Wait 3 months until cert expired

SystemKeeper:
Fixes #9606 

This PR adds a background job which checks the certificate of the TURN server (if `turns:` is used) and sends a notification to admins when the certificate is about to expire (less than 10 days).
Adding a test for this is a bit above my head - not sure how this can be done.

### 🚧 Tasks

- [ ] Add test?
- [ ] Allow to specify a different group to be notified? (Like it is done at [updatenotification](https://github.com/nextcloud/server/blob/fa23698c23a7ef9380f13c4fe439ff3aa6a9b9b0/apps/updatenotification/lib/Notification/BackgroundJob.php#L228-L237))

### 🏁 Checklist

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -63,6 +63,7 @@ And in the works for the [coming versions](https://github.com/nextcloud/spreed/m
 		<job>OCA\Talk\BackgroundJob\CheckHostedSignalingServer</job>
 		<job>OCA\Talk\BackgroundJob\CheckMatterbridges</job>
 		<job>OCA\Talk\BackgroundJob\ExpireChatMessages</job>
+		<job>OCA\Talk\BackgroundJob\CheckTurnCertificate</job>
 	</background-jobs>
 
 	<repair-steps>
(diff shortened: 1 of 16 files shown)


nickvergessen:
Should do the same for all others as well? recording server, HPB, stun?

SystemKeeper:
> Should do the same for all others as well? recording server, HPB, stun?

STUN doesn't use TLS, so not certificate involved here.
HPB and Recording server could be added, yes. TURN was the most important to me, because it might be hard to detect and might only fail in some cases. If HPB or Recording server have a problem it is more obvious and would also be reflected in the admin UI test. But yea, might make sense to have a more non-turn-specific approach.

nickvergessen:
Let's see tomorrow

nickvergessen:
As discussed we would like to:

- Move the check into a service
- Add a frontend endpoint
    - That is then executed on the admin page and shows a ✅  / ❌  on the server list

SystemKeeper:
Done as discussed @nickvergessen

nickvergessen (approved):
Okay backend-wise.

Please review frontend @Antreesy

Antreesy (commented):
Frontend code (Vue + JS) seems clean and proper, and local signaling server still works for me, although I don't have TURN configured

Antreesy:
Seems fine to backport this as well, or it's not necessary?

nickvergessen:
Looking at the diff size, let's not do it. :)

Issue #6431, opened 2021-10-29
List of open conversations


jeekajoo:
**Is your feature request related to a problem? Please describe.**
Unless we begin to search the name of an existing conversation, it is not listed and so we cannot find/join it.
It means we have to know it exists. And we have to know the name of the conversation.

**Describe the solution you'd like**
List all existing conversations that we can join

**Describe alternatives you've considered**
Add button to show all existing conversations that we can join

Thanks !

---

- [x] Frontend https://github.com/nextcloud/spreed/issues/9675
- [x] iOS https://github.com/nextcloud/talk-ios/issues/1262
- [x] Android https://github.com/nextcloud/talk-android/issues/3070

nickvergessen:
# 🖍️ Design

The rough design idea for web is to make the + menu a action menu with 2 entries:
![grafik](https://github.com/nextcloud/spreed/assets/213943/94af4e90-5bc1-4d1d-bbd0-19a61b00949a)

Afterwards we show the "RoomSelector" modal that is also used when you "Forward message" from the … menu of a chat message
![grafik](https://github.com/nextcloud/spreed/assets/213943/13a14057-6521-4985-8d56-efc884a16b80)

The list of conversations can be accessed by simply searching for an empty string `''` on the existing endpoint:
https://nextcloud-talk.readthedocs.io/en/latest/conversation/#get-open-conversations

mahibi:
how should this be done for mobiles?
On android, we have the plus button at the bottom:
![grafik](https://github.com/nextcloud/spreed/assets/2932790/2908c784-77a3-493c-afdf-660e47c5f587)

a) should this show 2 options, eg. "create new conversation" and "list open conversations" with an expanded menu list this one:
https://stackoverflow.com/questions/30699302/android-design-support-library-expandable-floating-action-buttonfab-menu

b) or should the "list open conversations" be placed just as an option inside the screen that is opened when clicking the plus button?

![grafik](https://github.com/nextcloud/spreed/assets/2932790/7f23bd8e-2e04-42a5-846f-68d656d99a8a)

cc @jancborchardt @nimishavijay @AndyScherzinger

SystemKeeper:
Please see also https://github.com/nextcloud/talk-ios/issues/1262

jancborchardt:
Expanded FAB doesn't seem to be a thing anymore according to the spec, or am I missing something?
https://m3.material.io/components/floating-action-button/guidelines

[...]

PR #9688, opened 2023-06-02
Filter conversation list for unread / mentioned conversations only

### LINKED ISSUE #9677: Frontend - Filter conversation list for unread / mentioned conversations only
* Overview issue in https://github.com/nextcloud/spreed/issues/9668

DorraJaouad:
### ☑️ Resolves

* Fixes #9677 
* The search box gets expanded when it is focused on to cover the filter button. 
* During the textual search operation, no filter can be applied

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

  Default View 
  image

Focused on input or during the textual search 
image

### 🚧 Tasks

- [ ] Visual review
- [ ] Code review

### 🏁 Checklist

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.spec.js
+++ b/src/components/LeftSidebar/LeftSidebar.spec.js
@@ -626,8 +626,8 @@ describe('LeftSidebar.vue', () => {
 				const ncModalComponent = wrapper.findComponent(NcModal)
 				expect(ncModalComponent.exists()).toBeTruthy()
 
-				const input = ncModalComponent.find('input[placeholder="Enter a name for this conversation"]')
-				expect(input.element.value).toBe(groupsResults[1].label)
+				const input = ncModalComponent.findComponent({name: 'NcTextField', ref: 'conversationName'})
+				expect(input.props('value')).toBe(groupsResults[1].label)
 
 				// nothing created yet
 				expect(createOneToOneConversationAction).not.toHaveBeenCalled()
@@ -648,9 +648,9 @@ describe('LeftSidebar.vue', () => {
 				await wrapper.vm.$nextTick()
 				const ncModalComponent = wrapper.findComponent(NcModal)
 				expect(ncModalComponent.exists()).toBeTruthy()
-
-				const input = ncModalComponent.find('input[placeholder="Enter a name for this conversation"]')
-				expect(input.element.value).toBe(circlesResults[1].label)
+				console.log(wrapper.html())
+				const input = ncModalComponent.findComponent({name: 'NcTextField', ref: 'conversationName'})
+				expect(input.props('value')).toBe(circlesResults[1].label)
 
 				// nothing created yet
 				expect(createOneToOneConversationAction).not.toHaveBeenCalled()
--- a/src/components/LeftSidebar/LeftSidebar.spec.js
+++ b/src/components/LeftSidebar/LeftSidebar.spec.js
@@ -7,6 +7,8 @@ import Vuex from 'vuex'
 import { subscribe, unsubscribe } from '@nextcloud/event-bus'
 import { loadState } from '@nextcloud/initial-state'
 
+import NcModal from '@nextcloud/vue/dist/Components/NcModal.js'
+
 import LeftSidebar from './LeftSidebar.vue'
 
 import router from '../../__mocks__/router.js'
@@ -47,8 +49,8 @@ describe('LeftSidebar.vue', () => {
 				// to prevent user status fetching
 				NcAvatar: true,
 				// to prevent complex dialog logic
-				NewGroupConversation: true,
 				NcActions: true,
+				NcModal: true,
 			},
 		})
 	}
@@ -612,9 +614,6 @@ describe('LeftSidebar.vue', () => {
 			})
 
 			test('shows group conversation dialog when clicking search result', async () => {
-				const eventHandler = jest.fn()
-				EventBus.$once('new-group-conversation-dialog', eventHandler)
-
 				const wrapper = await testSearch(SEARCH_TERM, [...groupsResults], [])
 
 				const resultsListItems = findNcListItems(wrapper, groupsResults.map(item => item.label))
@@ -622,7 +621,13 @@ describe('LeftSidebar.vue', () => {
 				expect(resultsListItems).toHaveLength(groupsResults.length)
 
 				await resultsListItems.at(1).findAll('a').trigger('click')
-				expect(eventHandler).toHaveBeenCalledWith(groupsResults[1])
+				// Wait for the component to render
+				await wrapper.vm.$nextTick()
+				const ncModalComponent = wrapper.findComponent(NcModal)
+				expect(ncModalComponent.exists()).toBeTruthy()
+
+				const input = ncModalComponent.find('input[placeholder="Enter a name for this conversation"]')
+				expect(input.element.value).toBe(groupsResults[1].label)
 
 				// nothing created yet
 				expect(createOneToOneConversationAction).not.toHaveBeenCalled()
@@ -630,8 +635,6 @@ describe('LeftSidebar.vue', () => {
 			})
 
 			test('shows circles conversation dialog when clicking search result', async () => {
-				const eventHandler = jest.fn()
-				EventBus.$once('new-group-conversation-dialog', eventHandler)
 
 				const wrapper = await testSearch(SEARCH_TERM, [...circlesResults], [])
 
@@ -640,7 +643,14 @@ describe('LeftSidebar.vue', () => {
 				expect(resultsListItems).toHaveLength(circlesResults.length)
 
 				await resultsListItems.at(1).findAll('a').trigger('click')
-				expect(eventHandler).toHaveBeenCalledWith(circlesResults[1])
+
+				// Wait for the component to render
+				await wrapper.vm.$nextTick()
+				const ncModalComponent = wrapper.findComponent(NcModal)
+				expect(ncModalComponent.exists()).toBeTruthy()
+
+				const input = ncModalComponent.find('input[placeholder="Enter a name for this conversation"]')
+				expect(input.element.value).toBe(circlesResults[1].label)
 
 				// nothing created yet
 				expect(createOneToOneConversationAction).not.toHaveBeenCalled()
@@ -704,6 +714,7 @@ describe('LeftSidebar.vue', () => {
 			const wrapper = mountComponent()
 			const buttonEl = findNcActionButton(wrapper, 'Create a new conversation')
 			expect(buttonEl.exists()).toBeTruthy()
+
 		})
 		test('does not show new conversation button if user cannot start conversations', () => {
 			loadStateSettings.start_conversations = false
--- a/src/components/LeftSidebar/LeftSidebar.spec.js
+++ b/src/components/LeftSidebar/LeftSidebar.spec.js
@@ -13,7 +13,7 @@ import router from '../../__mocks__/router.js'
 import { searchPossibleConversations, searchListedConversations } from '../../services/conversationsService.js'
 import { EventBus } from '../../services/EventBus.js'
 import storeConfig from '../../store/storeConfig.js'
-import { findNcListItems } from '../../test-helpers.js'
+import { findNcListItems, findNcActionButton } from '../../test-helpers.js'
 
 jest.mock('@nextcloud/initial-state', () => ({
 	loadState: jest.fn(),
@@ -48,6 +48,7 @@ describe('LeftSidebar.vue', () => {
 				NcAvatar: true,
 				// to prevent complex dialog logic
 				NewGroupConversation: true,
+				NcActions: true,
 			},
 		})
 	}
@@ -701,14 +702,14 @@ describe('LeftSidebar.vue', () => {
 			loadStateSettings.start_conversations = true
 
 			const wrapper = mountComponent()
-			const buttonEl = wrapper.findComponent({ name: 'NewGroupConversation' })
+			const buttonEl = findNcActionButton(wrapper, 'Create a new conversation')
 			expect(buttonEl.exists()).toBeTruthy()
 		})
 		test('does not show new conversation button if user cannot start conversations', () => {
 			loadStateSettings.start_conversations = false
 
 			const wrapper = mountComponent()
-			const buttonEl = wrapper.findComponent({ name: 'NewGroupConversation' })
+			const buttonEl = findNcActionButton(wrapper, 'Create a new conversation')
 			expect(buttonEl.exists()).toBeFalsy()
 		})
 	})
--- a/src/components/LeftSidebar/LeftSidebar.spec.js
+++ b/src/components/LeftSidebar/LeftSidebar.spec.js
@@ -7,15 +7,13 @@ import Vuex from 'vuex'
 import { subscribe, unsubscribe } from '@nextcloud/event-bus'
 import { loadState } from '@nextcloud/initial-state'
 
-import NcModal from '@nextcloud/vue/dist/Components/NcModal.js'
-
 import LeftSidebar from './LeftSidebar.vue'
 
 import router from '../../__mocks__/router.js'
 import { searchPossibleConversations, searchListedConversations } from '../../services/conversationsService.js'
 import { EventBus } from '../../services/EventBus.js'
 import storeConfig from '../../store/storeConfig.js'
-import { findNcListItems, findNcActionButton } from '../../test-helpers.js'
+import { findNcListItems } from '../../test-helpers.js'
 
 jest.mock('@nextcloud/initial-state', () => ({
 	loadState: jest.fn(),
@@ -623,7 +621,7 @@ describe('LeftSidebar.vue', () => {
 				await resultsListItems.at(1).findAll('a').trigger('click')
 				// Wait for the component to render
 				await wrapper.vm.$nextTick()
-				const ncModalComponent = wrapper.findComponent(NcModal)
+				const ncModalComponent = wrapper.findComponent({ name: 'NcModal' })
 				expect(ncModalComponent.exists()).toBeTruthy()
 
 				const input = ncModalComponent.findComponent({ name: 'NcTextField', ref: 'conversationName' })
@@ -646,7 +644,7 @@ describe('LeftSidebar.vue', () => {
 
 				// Wait for the component to render
 				await wrapper.vm.$nextTick()
-				const ncModalComponent = wrapper.findComponent(NcModal)
+				const ncModalComponent = wrapper.findComponent({ name: 'NcModal' })
 				expect(ncModalComponent.exists()).toBeTruthy()
 				const input = ncModalComponent.findComponent({ name: 'NcTextField', ref: 'conversationName' })
 				expect(input.props('value')).toBe(circlesResults[1].label)
@@ -711,7 +709,7 @@ describe('LeftSidebar.vue', () => {
 			loadStateSettings.start_conversations = true
 
 			const wrapper = mountComponent()
-			const buttonEl = findNcActionButton(wrapper, 'Create a new conversation')
+			const buttonEl = wrapper.findComponent({ name: 'NewGroupConversation' })
 			expect(buttonEl.exists()).toBeTruthy()
 
 		})
@@ -719,7 +717,7 @@ describe('LeftSidebar.vue', () => {
 			loadStateSettings.start_conversations = false
 
 			const wrapper = mountComponent()
-			const buttonEl = findNcActionButton(wrapper, 'Create a new conversation')
+			const buttonEl = wrapper.findComponent({ name: 'NewGroupConversation' })
 			expect(buttonEl.exists()).toBeFalsy()
 		})
 	})
--- a/src/components/LeftSidebar/LeftSidebar.spec.js
+++ b/src/components/LeftSidebar/LeftSidebar.spec.js
@@ -626,7 +626,7 @@ describe('LeftSidebar.vue', () => {
 				const ncModalComponent = wrapper.findComponent(NcModal)
 				expect(ncModalComponent.exists()).toBeTruthy()
 
-				const input = ncModalComponent.findComponent({name: 'NcTextField', ref: 'conversationName'})
+				const input = ncModalComponent.findComponent({ name: 'NcTextField', ref: 'conversationName' })
 				expect(input.props('value')).toBe(groupsResults[1].label)
 
 				// nothing created yet
@@ -648,8 +648,7 @@ describe('LeftSidebar.vue', () => {
 				await wrapper.vm.$nextTick()
 				const ncModalComponent = wrapper.findComponent(NcModal)
 				expect(ncModalComponent.exists()).toBeTruthy()
-				console.log(wrapper.html())
-				const input = ncModalComponent.findComponent({name: 'NcTextField', ref: 'conversationName'})
+				const input = ncModalComponent.findComponent({ name: 'NcTextField', ref: 'conversationName' })
 				expect(input.props('value')).toBe(circlesResults[1].label)
 
 				// nothing created yet
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -24,13 +24,13 @@
 		<div class="new-conversation"
 			:class="{ 'new-conversation--scrolled-down': !isScrolledToTop }">
 			<SearchBox ref="searchbox"
-				v-model="searchText"
+				:value.sync="searchText"
 				class="conversations-search"
 				:class="{'conversations-search--expanded': isFocused}"
-				:disabled="isFiltered"
+				:disabled="isFiltered !== null"
 				:is-searching="isSearching"
-				@blur="setIsFocused(false)"
-				@focus="setIsFocused(true)"
+				@focus="setIsFocused"
+				@blur="setIsFocused"
 				@input="debounceFetchSearchResults"
 				@submit="onInputEnter"
 				@keydown.enter.native="handleEnter"
@@ -46,7 +46,7 @@
 					</template>
 					<NcActionButton close-after-click
 						class="filter-actions__button"
-						:primary="isFiltered === 'mentions'"
+						:class="{'filter-actions__button--active': isFiltered === 'mentions'}"
 						@click="handleFilter('mentions')">
 						<template #icon>
 							<AtIcon :size="20" />
@@ -264,7 +264,6 @@ export default {
 			preventFindingUnread: false,
 			roomListModifiedBefore: 0,
 			forceFullRoomListRefreshAfterXLoops: 0,
-			isNewGroupConversationOpen: false,
 			isFocused: false,
 			isFiltered: null,
 		}
@@ -389,12 +388,18 @@ export default {
 		getFocusableList() {
 			return this.$el.querySelectorAll('li.acli_wrapper .acli')
 		},
-		setIsFocused(value) {
-			this.isFocused = value
+		setIsFocused(event) {
+			if (event.relatedTarget?.className.includes('input-field__clear-button')) {
+				return
+			}
+			this.isFocused = event.type === 'focus'
+
 		},
 
 		handleFilter(filter) {
 			this.isFiltered = filter
+			// Clear the search input once a filter is active
+			this.searchText = ''
 		},
 
 		focusCancel() {
@@ -417,10 +422,6 @@ export default {
 				this.fetchSearchResults()
 			}
 		}, 250),
-
-		toggleNewGroupConversation(value) {
-			this.isNewGroupConversationOpen = value
-		},
 
 		async fetchPossibleConversations() {
 			this.contactsLoading = true
@@ -510,6 +511,7 @@ export default {
 		// Reset the search text, therefore end the search operation.
 		abortSearch() {
 			this.searchText = ''
+			this.isFocused = false
 			if (this.cancelSearchPossibleConversations) {
 				this.cancelSearchPossibleConversations()
 			}
@@ -755,16 +757,16 @@ export default {
 
 }
 
-//FIXME : this should be changed once the disabled style for input is added
+//FIXME : upstream: this should be changed once the disabled style for NcInputField is added
 :deep(.input-field__input[disabled="disabled"]){
 	background-color: var(--color-background-dark);
 }
 
 .options{
 	position: relative;
-	transition: all 0.3s ease; //gets hidden once the search box is expanded
 	left : calc(65% + 4px);
 	display: flex;
+	height: var(--default-clickable-area);
 }
 
 .filter-actions__button--active{
@@ -774,10 +776,6 @@ export default {
 		font-weight: bold;
 	}
 
-}
-
-.hidden-visually{
-	height:100%;
 }
 
 .settings-button {
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -37,7 +37,8 @@
 				@abort-search="abortSearch" />
 
 			<!-- Options -->
-			<div class="options">
+			<div class="options"
+				:class="{'options--hidden': isFocused}">
 				<NcActions ref="filterMainBtn"
 					class="filters-button">
 					<template #icon>
@@ -69,7 +70,7 @@
 						<template #icon>
 							<FilterRemoveIcon :size="20" />
 						</template>
-						{{ t('spreed','Clear filter :') }} {{ filterText }}
+						{{ t('spreed', 'Clear filters: {filterText}', { filterText: filterText }) }}
 					</NcActionButton>
 				</NcActions>
 
@@ -87,7 +88,6 @@
 					</NcActionButton>
 				</NcActions>
 			</div>
-
 			<!-- New Conversation -->
 			<NewGroupConversation ref="newGroupConversation"
 				:show-modal="isNewGroupConversationOpen"
@@ -297,7 +297,8 @@ export default {
 				conversations = conversations.filter(conversation => conversation.unreadMessages > 0)
 				break
 			case ('is:mentioned'):
-				conversations = conversations.filter(conversation => conversation.unreadMention)
+				conversations = conversations.filter(conversation => conversation.unreadMention || (conversation.unreadMessages > 0
+				 && (conversation.type === CONVERSATION.TYPE.ONE_TO_ONE || conversation.type === CONVERSATION.TYPE.ONE_TO_ONE_FORMER)))
 				break
 			default:
 				if (this.searchText !== '') {
@@ -764,7 +765,6 @@ export default {
 }
 
 .conversations-search {
-	flex-grow: 1;
 	transition: all 0.3s ease;
 	width: 65%;
 	z-index: 2;
@@ -774,7 +774,7 @@ export default {
 		border-radius: var(--border-radius-pill);
 	}
 	&--expanded {
-		width: 95%;
+		width: 90%;
 	}
 
 	&--filter{
@@ -791,6 +791,10 @@ export default {
 	left : calc(65% + 10px);
 	display: flex;
 	gap: 0.25px;
+
+	&--hidden{
+		visibility: hidden;
+	}
 }
 
 .filterButton--Clear{
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -476,7 +476,6 @@ export default {
 			} else {
 				// For other types, show the modal directly
 				this.$refs.newGroupConversation.showModalForItem(item)
-				this.toggleNewGroupConversation(true)
 			}
 		},
 
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -24,6 +24,18 @@
 		<div class="new-conversation"
 			:class="{ 'new-conversation--scrolled-down': !isScrolledToTop }">
 
+			<SearchBox v-model="searchText"
+				class="conversations-search"
+				:class="{ 'conversations-search--expanded': isFocused }"
+				:is-searching="isSearching"
+				@blur="isFocused(false)"
+				@focus="isFocused(true)"
+				@focusCancel ="isFocused = false"
+				@input="debounceFetchSearchResults"
+				@submit="onInputEnter"
+				@keydown.enter.native="handleEnter"
+				@abort-search="abortSearch" />
+
 			<!-- Options -->
 			<div class="options">
 				<NcActions class ='filters-button'>
@@ -67,23 +79,11 @@
 					</NcActionButton>
 				</NcActions>
 			</div>
-
-			<SearchBox v-model="searchText"
-				class="conversations-search"
-				:class="{ 'conversations-search--expanded': isFocused }"
-				:is-searching="isSearching"
-				@focus="isFocused = true"
-				@focusCancel ="isFocused = false"
-				@input="debounceFetchSearchResults"
-				@submit="onInputEnter"
-				@keydown.enter.native="handleEnter"
-				@abort-search="abortSearch" />
 			
 			<!-- New Conversation -->
 			<NewGroupConversation ref="newGroupConversation"
 				:show-modal="isNewGroupConversationOpen"
-				@open-modal="toggleNewGroupConversation(true)"
-				@close-modal="toggleNewGroupConversation(false)" />
+				@update-modal="toggleNewGroupConversation" />
 		</div>
 
 		<template #list>
@@ -392,6 +392,9 @@ export default {
 	methods: {
 		getFocusableList() {
 			return this.$el.querySelectorAll('li.acli_wrapper .acli')
+		},
+		isFocused(value){
+			this.isFocused = value
 		},
 		focusCancel() {
 			return this.abortSearch()
@@ -742,22 +745,24 @@ export default {
 
 .conversations-search {
 	flex-grow: 1;
-	transition: all 0.7s ease;
+	transition: all 0.3s ease;
 	width: 65%;
+	z-index: 1;
 	position : absolute;
-	left: 8px; //padding
+	padding: 0 8px ;
 
 	:deep(.input-field__input) {
 		border-radius: var(--border-radius-pill);
 	}
 	&--expanded {
-		width: 95%;
+		width: 100%;
 	}
 }
 
 
 .options{
 	position: relative;
+	z-index: 2;
 	left : calc(65% + 7px);
 	display: flex;
 	gap: 0.5px;
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -23,42 +23,62 @@
 	<NcAppNavigation :aria-label="t('spreed', 'Conversation list')">
 		<div class="new-conversation"
 			:class="{ 'new-conversation--scrolled-down': !isScrolledToTop }">
+
+			<!-- Options -->
+			<div class="options">
+				<NcActions class ='filters-button'>
+					<template #icon>
+						<FilterIcon :size="15" />
+					</template>
+					<NcActionButton close-after-click
+						@click="insertValue(t('spreed','is:mentioned'))">
+						<template #icon>
+							<AtIcon :size="20" />
+						</template>
+						{{ t('spreed','Filter unread mentions') }}
+					</NcActionButton>
+
+					<NcActionButton close-after-click
+						@click="insertValue(t('spreed','is:unread'))">
+						<template #icon>
+							<MessageBadge :size="20" />
+						</template>
+						{{ t('spreed','Filter unread messages') }}
+					</NcActionButton>
+				</NcActions>
+
+				<NcActions >
+					<template #icon>
+						<PlusIcon :size="20" />
+					</template>
+					<NcActionButton v-if="canStartConversations"
+						close-after-click
+						@click="toggleNewGroupConversation(true)">
+						<template #icon>
+							<PlusIcon :size="20" />
+						</template>
+						{{ t('spreed','Create a new conversation') }}
+					</NcActionButton>
+					<NcActionButton close-after-click>
+						<template #icon>
+							<ListIcon :size="20" />
+						</template>
+						{{ t('spreed','List of open conversations') }}
+					</NcActionButton>
+				</NcActions>
+			</div>
+
 			<SearchBox v-model="searchText"
 				class="conversations-search"
+				:class="{ 'conversations-search--expanded': isFocused }"
 				:is-searching="isSearching"
+				@focus="isFocused = true"
+				@focusCancel ="isFocused = false"
 				@input="debounceFetchSearchResults"
 				@submit="onInputEnter"
 				@keydown.enter.native="handleEnter"
 				@abort-search="abortSearch" />
-
-			<!-- Options -->
-			<NcActions>
-				<NcActionButton v-if="canStartConversations"
-					close-after-click
-					@click="toggleNewGroupConversation(true)">
-					<template #icon>
-						<PlusIcon :size="20" />
-					</template>
-					{{ t('spreed','Create a new conversation') }}
-				</NcActionButton>
-
-				<NcActionButton close-after-click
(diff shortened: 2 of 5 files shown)


src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -21,96 +21,77 @@
...
 	<NcAppNavigation :aria-label="t('spreed', 'Conversation list')">
-		<div class="new-conversation"
-			:class="{ 'new-conversation--scrolled-down': !isScrolledToTop }">
-			<SearchBox v-model="searchText"
-				class="conversations-search"
-				:is-searching="isSearching"
-				@input="debounceFetchSearchResults"
-				@submit="onInputEnter"
-				@keydown.enter.native="handleEnter"
+		<div class="new-conversation" :class="{ 'new-conversation--scrolled-down': !isScrolledToTop }">
```
Antreesy, line 24: So, something in Eslint configuration on your local PC is wrong, and a lot of lines were changed, which shouldn't happen usually.
Please, check for Eslint plugin settings, and don't forget to run `lint:fix` script

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -21,96 +21,77 @@
...
+			<SearchBox v-model="searchText" class="conversations-search" :is-searching="isSearching"
+				@input="debounceFetchSearchResults" @submit="onInputEnter" @keydown.enter.native="handleEnter"
 				@abort-search="abortSearch" />
 			<NewGroupConversation v-if="canStartConversations" />
 		</div>
+		<div class="button-container">
+			<button class="button" @click="insertValue('is:unread')" :title="t('spreed', 'Unread messages')">{{
```
Antreesy, line 31: We need to use `NcButton` from Vue components library. It comes in spreed repo as node_modules dependencies, so you could take a look, how it is used in other components; there should be a plenty of examples

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -21,96 +21,77 @@
...
+				t('spreed', 'unread') }}</button>
```
Antreesy, line 32: Translations don't work for you, because it's a new line.
After PR will be merged, all new lines appear here: https://app.transifex.com/nextcloud/nextcloud/spreed/.
Then a community should provide a translation for it (as far as i know, it has to be done manually for each language), and 1-2 times a week, bot will merge new translations into project.

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -192,8 +173,8 @@ export default {
 			isCirclesEnabled: loadState('spreed', 'circles_enabled'),
 			canStartConversations: loadState('spreed', 'start_conversations'),
 			initialisedConversations: false,
-			cancelSearchPossibleConversations: () => {},
-			cancelSearchListedConversations: () => {},
+			cancelSearchPossibleConversations: () => { },
```
Antreesy, line 176: After setting up linter, please look through the file and revert all changes, which are unintended, and not related to this issue

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -654,6 +649,26 @@ export default {
 	justify-content: flex-start !important;
 }
 
+.button-container {
```
Antreesy, line 652: Also, styles shouldn't be neccessary, when you insert `NcButton` component

Antreesy (changes requested):
PR looks a bit messy now. Please, figure out the problem with Eslint and put the code in order.
And don't hesitate to ask questions!

src/components/LeftSidebar/LeftSidebar.vue
DorraJaouad: Yes, I removed the default configuration so it won't cause that change again

src/components/LeftSidebar/LeftSidebar.vue
DorraJaouad: I was using the NcButton BUT it wasn't flexible with the modification( changing the span font size ) and I switched to the normal Button tag with which I could modify the style adequately.

src/components/LeftSidebar/LeftSidebar.vue
DorraJaouad: Corrected

src/components/LeftSidebar/LeftSidebar.vue
DorraJaouad: that's the container of the two buttons. What do you think? should it be a button filter or the 3 dots on the right that drops a list of the two filters with the conversation addition option?

src/components/LeftSidebar/LeftSidebar.vue
Antreesy: What you're intending to do, should be possible by using `:deep()` selector from Vue.
Try to experiment with it to understand how it works, but as I said before, we should stick to the native appearance of NcButton, because it was made bold for a reason
And also I spend a week or two replacing buttons in project with NcButton, so don't ruin my job)

src/components/LeftSidebar/LeftSidebar.vue
Antreesy: Sorry, I meant not the container class, but all new styles for buttons (as NcButton has it by default)

I'm thinking, that we should take as little space as possible for this feature, so three-dots menu seems more preferable to me. But we could also wait for design team proposal, or suggest our option in specific issue (you could put your screenshots there)

src/components/LeftSidebar/NewGroupConversation/NewGroupConversation.vue
```diff
@@ -263,6 +259,9 @@ export default {
 				this.passwordProtect = false
 			}
 		},
+		showModal(newVal) {
```
Antreesy, line 262: We can't change `props` directly, like we doing with `data`
Instead, we need to emit a new value, 'false' in our case, and capture it in `LeftSidebar` with @close="toggle(false)", like it works now

src/components/LeftSidebar/NewGroupConversation/NewGroupConversation.vue
```diff
@@ -309,6 +304,7 @@ export default {
 			this.password = ''
 			this.listable = CONVERSATION.LISTABLE.NONE
 			this.$store.dispatch('purgeNewGroupConversationStore')
+			this.$emit('close');
```
Antreesy, line 307: all emits should be defined in the component:

You need to add a line on a root level (same as mixins, props, data e.t.c.)
```
emits: ['close'],
```

Maybe, lint:fix could do it automatically

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -21,32 +21,51 @@
...
+				</NcActionButton>
+				<NcActionButton  @click="insertValue(t('spreed','is:mentioned'))" close-after-click>
+					<template #icon>
+						<AtIcon :size="20" />
+					</template>
+					{{ t('spreed','Filter unread mentions') }}
+				</NcActionButton>
+				<NcActionButton @click="renderNewGroupConversation" close-after-click>
+					<template #icon>
+						<Plus :size="20" />
```
Antreesy, line 49:
```suggestion
						<PlusIcon :size="20" />
```
Not critical, but we could import it as `PlusIcon` for icons names consistency

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -62,14 +81,12 @@
 								:title="item.label"
 								@click="createAndJoinConversation(item)">
 								<template #icon>
-									<ConversationIcon :item="iconData(item)"
-										:disable-menu="true" />
+									<ConversationIcon :item="iconData(item)" :disable-menu="true" />
```
Antreesy, line 84: I'd revert this commit: https://github.com/nextcloud/spreed/pull/9688/commits/41dff5ce63f5dff91a8e19c80de62d4e00560c56
What worries me, as it touches a lot of lines of code, and it will be difficult to backport something later (because this PR doesn't go to NC27, as it's a new feature for NC28)
If you want, we could made a specific PR for it later

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -202,6 +226,8 @@ export default {
 			preventFindingUnread: false,
 			roomListModifiedBefore: 0,
 			forceFullRoomListRefreshAfterXLoops: 0,
+			showNewGroupConversation: false,
```
Antreesy, line 229:
```suggestion
			isNewGroupConversationShown: false,
```

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -211,10 +237,21 @@ export default {
...
 				const lowerSearchText = this.searchText.toLowerCase()
-				conversations = conversations.filter(conversation =>
-					conversation.displayName.toLowerCase().includes(lowerSearchText)
-					|| conversation.name.toLowerCase().includes(lowerSearchText)
-				)
+				switch (this.searchText) {
+				case t('spreed', 'is:unread'):
+					conversations = conversations.filter(conversation => conversation.unreadMessages > 0)
+					break
+				case t('spreed', 'is:mentioned'):
```
Antreesy, line 244:
```suggestion
				case 'is:mentioned':
```

Antreesy (changes requested):

Antreesy:
Also tests should be adjusted:
https://github.com/nextcloud/spreed/actions/runs/5176171012/jobs/9324649445?pr=9688#step:7:2218

look through the project to see, how `findNcActionButton` helper is used in tests, and fix spec.js file
You could run tests locally, see scripts in package.json

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -21,32 +21,51 @@
...
+
+			<!-- Options -->
+			<NcActions>
+				<NcActionButton  @click="insertValue(t('spreed','is:unread'))" close-after-click>
+					<template #icon>
+						<MessageBadge :size="20" />
+					</template>
+					{{ t('spreed','Filter unread messages') }}
+				</NcActionButton>
+				<NcActionButton  @click="insertValue(t('spreed','is:mentioned'))" close-after-click>
```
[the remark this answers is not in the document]
DorraJaouad: it will appear on the search box, is it okay to be always in English?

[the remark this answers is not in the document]
Antreesy: I think so, just like here:
image

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -618,6 +666,7 @@ export default {
 	display: flex;
 	padding: 8px 0 8px 12px;
 	align-items: center;
+	padding-bottom: 0px;
```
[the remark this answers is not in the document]
DorraJaouad: those are for .new-conversation

src/components/LeftSidebar/LeftSidebar.vue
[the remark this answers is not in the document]
DorraJaouad: GitHub is always in English, I guess. I saw in other apps that those filters are usually translated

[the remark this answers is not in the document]
Antreesy: We could ask @jancborchardt and @nimishavijay for opinion (and for design review as well)

src/components/LeftSidebar/LeftSidebar.vue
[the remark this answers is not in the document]
Antreesy: Ok, but:
* It was intended to do that way: https://github.com/nextcloud/spreed/pull/7977
* we could modify existing property: `padding: 8px 0 0 12px;`, if we need to.

src/components/LeftSidebar/LeftSidebar.vue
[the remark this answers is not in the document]
DorraJaouad: Sure, this is the other suggestion for the design 
buttons

Antreesy:
I reverted some codeline wrapping changes and fixed a couple of things: https://github.com/nextcloud/spreed/commit/2865709a81604048604b3c40ace044e94fe449a3
You could cherry-pick this commit and apply, to reduce amount of lines changed.

Also could be done:
* squash commits into one (or two commits: one for new functionality, one for refactoring and moving NewGroupConversation component
* we may get rid of `EventBus.$on('new-group-conversation-dialog')` and call `showModalForItem` directly from LeftSidebar (see how it's done in other component): https://github.com/nextcloud/spreed/blob/2858a005151bb6700de6c1be978659a9f0d10d76/src/components/BreakoutRoomsEditor/SendMessageDialog.vue#L110

src/components/LeftSidebar/NewGroupConversation/NewGroupConversation.vue
```diff
@@ -184,6 +185,10 @@ export default {
 
 	mixins: [participant],
 
+	props: {
+		showModal: Boolean,
```
Antreesy, line 189: sorry for misleading in the chat, but it's better to use full syntax:
```
showModal: {
			type: Boolean,
			required: true,
		},
```

jancborchardt:
Ah just saw in the issue it will be showing "is:unread" in the search field so that would answer it. It is quite technical though, but something for a follow-up.

DorraJaouad:
> Nice! A detail: The sorting of the action menu entries should be the other way around, cause currently the least important (unread messages) is up top.
> 
> Also, could you add a screenshot of how it looks like when you have it filtered by e.g. mentions and the action menu is closed? (Needs to be obvious it is in a non-default state and offer a one-click reset, e.g. by showing a chip "Unread mentions" with an x to reset.)

Thank you. The filter gets cleared by clicking on the `x` button from the search box. 
image

Antreesy:
> Ah just saw in the issue it will be showing "is:unread" in the search field so that would answer it.

We asked about it in the comment above, actually. Should the query requests like this be language-independent (like Github uses it), or translated for users?
* is:unread
* t('spreed', 'is:unread')

I'd prefer first option, as we are currently using the same string for filter checking, and that could raise some misalignments in future, but it's up to user experience.

src/components/LeftSidebar/NewGroupConversation/NewGroupConversation.vue
```diff
@@ -168,7 +153,6 @@ import {
 	createPrivateConversation,
 	setConversationPassword,
 } from '../../../services/conversationsService.js'
-import { EventBus } from '../../../services/EventBus.js'
```
Antreesy, line 171: If EventBus is not used in LeftSidebar anywhere, don't forget to remove it too

src/components/LeftSidebar/NewGroupConversation/NewGroupConversation.vue
```diff
@@ -265,39 +256,26 @@ export default {
...
-
 		setIsAvatarEdited(value) {
 			this.isAvatarEdited = value
 		},
 
 		showModalForItem(item) {
 			if (item) {
 				// Preload the conversation name from group selection
-				this.conversation.displayName = item.label
+				this.newConversation.displayName = item.label
```
Antreesy, line 267: When you rebase the branch later, you might face the conflict there, bc this line is on master already

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -470,8 +473,9 @@ export default {
 					params: { token: conversation.token },
 				}).catch(err => console.debug(`Error while pushing the new conversation's route: ${err}`))
 			} else {
-				// For other types we start the conversation creation dialog
-				EventBus.$emit('new-group-conversation-dialog', item)
+				// For other types, show the modal directly
+				this.$refs.newGroupConversation.showModalForItem(item)
+				this.toggleNewGroupConversation(true)
```
Antreesy, line 478:
```suggestion
        this.$refs.newGroupConversation.showModalForItem(item)
```

this method already opens modal by this line, so we don't need to do it twice:
```js
	this.$emit('open-modal')
```

src/components/LeftSidebar/LeftSidebar.spec.js
```diff
@@ -612,26 +614,27 @@ describe('LeftSidebar.vue', () => {
...
 				expect(resultsListItems).toHaveLength(groupsResults.length)
 
 				await resultsListItems.at(1).findAll('a').trigger('click')
-				expect(eventHandler).toHaveBeenCalledWith(groupsResults[1])
+				// Wait for the component to render
+				await wrapper.vm.$nextTick()
+				const ncModalComponent = wrapper.findComponent(NcModal)
+				expect(ncModalComponent.exists()).toBeTruthy()
+
+				const input = ncModalComponent.find('input[placeholder="Enter a name for this conversation"]')
```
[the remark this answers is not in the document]
DorraJaouad: It doesn't work because it is not rendered as a NcTextField element 
`<div data-v-f50b337c="" class="input-field__main-wrapper"><input data-v-f50b337c="" id="inputxhsjs" type="text" placeholder="Enter a name for this conversation" aria-describedby="" aria-live="polite" trailingbuttonicon="close" class="input-field__input">`

the only "identifier" I could find was obviously the placeholder.

marcoambrosini (changes requested):
I think it's not very good that we're mixing the creation of a new conversation with the filtering of the conversations. This functionality should be done in a separate filtering button. 

| `search field` | `filtering options action button` | `new conversation action button` (in the future could also contain `join open conversation action button`) |

The search field can expand to the whole width of the line when clicked and the other elements can hide during the search operation 

blocking because I think that further design conversation is needed @nextcloud/designers

DorraJaouad:
> 

The design team suggested adding a divider between filters and other options ( there is a 4th option coming up) [in this issue](https://github.com/nextcloud/spreed/issues/6431#issuecomment-1582603659) which seems good. 

At first thought, I had this [idea ](https://github.com/nextcloud/spreed/pull/9688#discussion_r1219528387)in mind for the filters. We can put the new conversation action button and the list of open conversations action button together in a drop-down list instead of the '`+`'

marcoambrosini:
cc @jancborchardt @nimishavijay @szaimen I think this deserves its own button, also to display active state when filtering is active

Antreesy:
> As soon as typing something into the input field, the plus button could be exchanged with a dropdown to show the filter options.

Doesn't that mean that we're blocking a new conversation dialog when search for something, and vice versa?
* Sometimes i'm trying to enter a name for a new conversation in the search box instead of clicking on "+"
* Also I'd like to see only unread conversations without typing anything, what also adds additional layer of filtering

If "Plus" button is staying, we need to find a compromiss, for example, to have both buttons accessible:

Screenshot from 2023-06-15 16-45-53

marcoambrosini:
@szaimen the filtering options should be visible when not searching. The + button and the filter button could go away during the search operation, and the search input field could expand to fill the whole width of the row. For the purpose of talk search, I don't think we need filtering of the search results. That would be also a bit misleading because the search results are not only joined conversations, but also not joined ones, for which there 's no read/unread state.

There's a good example of this behaviour in telegram desktop.

https://github.com/nextcloud/spreed/assets/26852655/63185000-6f70-43d6-b807-e4a31c500ae0

@Antreesy your screenshot looks good, what we could do to further improve this is
- Display the abort search button when the input field is selected (currently id displays only when something is typed)
- Change the placeholder text from "search in conversation" into "search"
- Reduce the size of the filter icon

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -644,10 +731,25 @@ export default {
 
 .conversations-search {
 	flex-grow: 1;
+	transition: all 0.7s ease;
+	width: 65%;
+	position : absolute;
+	left: 8px; //padding
```
Antreesy, line 737:
```suggestion
	padding: 0 8px;
```
and with that we may use `width: 100%` in expanded mode

src/components/LeftSidebar/NewGroupConversation/NewGroupConversation.vue
```diff
@@ -195,12 +179,20 @@ export default {
...
 	mixins: [participant],
 
+	props: {
+		showModal: {
+			type: Boolean,
+			required: true,
+		},
+	},
+
+	emits: ['open-modal', 'close-modal'],
```
Antreesy, line 194:
```suggestion
	emits: ['update-modal'],
```

src/components/LeftSidebar/NewGroupConversation/NewGroupConversation.vue
```diff
@@ -264,40 +255,28 @@ export default {
...
 		showModalForItem(item) {
 			if (item) {
 				// Preload the conversation name from group selection
-				this.conversation.displayName = item.label
+				this.newConversation.displayName = item.label
 				this.$store.dispatch('updateSelectedParticipants', item)
 			}
 
-			this.showModal()
+			this.$emit('open-modal')
```
Antreesy, line 272:
```suggestion
			this.$emit('update-modal', true)
```

src/components/LeftSidebar/NewGroupConversation/NewGroupConversation.vue
```diff
@@ -309,6 +288,7 @@ export default {
 			this.password = ''
 			this.listable = CONVERSATION.LISTABLE.NONE
 			this.$store.dispatch('purgeNewGroupConversationStore')
+			this.$emit('close-modal')
```
Antreesy, line 291:
```suggestion
			this.$emit('update-modal', false)
```

src/components/LeftSidebar/SearchBox/SearchBox.vue
```diff
@@ -26,6 +26,8 @@
 			:label="placeholderText"
 			:show-trailing-button="isSearching"
 			trailing-button-icon="close"
+			@blur="isFocused(false)"
+			@focus="isFocused(true)"
```
Antreesy, line 30: Seems more readable to me to pass methods and use `@focus` and `@blur` in LeftSidebar:
```suggestion
			v-on:"$listeners"
```

src/components/LeftSidebar/SearchBox/SearchBox.vue
```diff
@@ -103,6 +105,19 @@ export default {
 		EventBus.$off('route-change', this.focusInputIfRoot)
 	},
 	methods: {
+		isFocused(value){
+			if (value){
+				if (this.localValue != t('spreed','is:mentioned') & this.localValue != t('spreed','is:unread')){
```
Antreesy, line 110: ... and thus we don't need this logic and emits here

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -333,6 +413,14 @@ export default {
 			}
 		}, 250),
 
+		toggleNewGroupConversation(value) {
+			this.isNewGroupConversationOpen = value
+		},
+
+		insertValue(value) {
```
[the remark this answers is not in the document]
nickvergessen: The biggest advantage of the textfield would be to one day be able to bookmark filters with the browser, similar to github.
But yeah for now radio-ed buttons and a fitler-remove icon when active sounds good too.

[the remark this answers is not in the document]
nickvergessen: Seeing the implementations on the mobile clients maybe this is the best way to go at the moment, as they handle it exactly like this.

nickvergessen (commented):
1. Currently the filter is still "hacked" via the search text and we should avoid that so we can:
2. The search filter can not be applied together with a textual filter
3.  When a filter is applied and the list is empty, we should show a "empty content" message like https://github.com/nextcloud/spreed/blob/1ad25e8b717436a792991932c17e680658bb67f8/src/components/LeftSidebar/LeftSidebar.vue#L126
4. I personally wouldn't automatically clear the filter when reopening the filter options, but have a third option there to clear it, so you can see what filter you are on?

src/components/LeftSidebar/SearchBox/SearchBox.vue
```diff
@@ -71,7 +72,7 @@ export default {
 		},
 	},
 
-	emits: ['update:value', 'input', 'submit', 'abort-search'],
+	emits: ['update:value', 'input', 'submit', 'abort-search', 'focus', 'focusCancel'],
```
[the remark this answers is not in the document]
Antreesy: Should have been reverted, as we're using `v-on="$listeners"` and emitting native events automatically

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -203,19 +269,32 @@ export default {
...
-				conversations = conversations.filter(conversation =>
-					conversation.displayName.toLowerCase().includes(lowerSearchText)
-					|| conversation.name.toLowerCase().includes(lowerSearchText)
-				)
+			switch (this.filterText) {
+			case ('is:unread'):
+				conversations = conversations.filter(conversation => conversation.unreadMessages > 0)
+				break
+			case ('is:mentioned'):
+				conversations = conversations.filter(conversation => conversation.unreadMention)
```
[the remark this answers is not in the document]
DorraJaouad: I think it does list the unread 1-1 chats already.

DorraJaouad:
> 1. Currently the filter is still "hacked" via the search text and we should avoid that so we can:
> 2. The search filter can not be applied together with a textual filter
> 3. When a filter is applied and the list is empty, we should show a "empty content" message like https://github.com/nextcloud/spreed/blob/1ad25e8b717436a792991932c17e680658bb67f8/src/components/LeftSidebar/LeftSidebar.vue#L126
> 4. I personally wouldn't automatically clear the filter when reopening the filter options, but have a third option there to clear it, so you can see what filter you are on?

There is also another option to show the filter type icon instead of the standard filter icon whenever that filter type (e.g: is:mentioned) is applied. The user can see which filter is applied.

src/components/LeftSidebar/LeftSidebar.vue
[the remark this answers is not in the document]
nickvergessen: it should list them also when selecting "mentions" as any unread 1-1 message is shown/handled  as "mention" in all other places

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -23,15 +23,77 @@
...
+					</NcActionButton>
+
+					<NcActionButton v-if="isFiltered"
+						close-after-click
+						class="filterButton--Clear"
+						@click="clearFilter">
+						<template #icon>
+							<FilterRemoveIcon :size="20" />
+						</template>
+						{{ t('spreed','Clear filter :') }} {{ filterText }}
```
nickvergessen, line 72:
```suggestion
						{{ t('spreed','Clear filters') }}
```

Otherwise this does not work for translations

src/components/LeftSidebar/LeftSidebar.vue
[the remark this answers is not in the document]
nickvergessen:
```suggestion
				conversations = conversations.filter(conversation => conversation.unreadMention || (conversation.unreadMessages > 0 && (conversation.type === CONVERSATION.TYPE.ONE_TO_ONE || conversation.type === CONVERSATION.TYPE.ONE_TO_ONE_FORMER)))
```

src/components/LeftSidebar/LeftSidebar.spec.js
```diff
@@ -611,26 +614,27 @@ describe('LeftSidebar.vue', () => {
...
 
 				const resultsListItems = findNcListItems(wrapper, groupsResults.map(item => item.label))
 				expect(resultsListItems.exists()).toBeTruthy()
 				expect(resultsListItems).toHaveLength(groupsResults.length)
 
 				await resultsListItems.at(1).findAll('a').trigger('click')
-				expect(eventHandler).toHaveBeenCalledWith(groupsResults[1])
+				// Wait for the component to render
+				await wrapper.vm.$nextTick()
+				const ncModalComponent = wrapper.findComponent(NcModal)
```
Antreesy, line 626:
```suggestion
				const ncModalComponent = wrapper.findComponent({ name: 'NcModal' })
```
Should work also, and we don't need to import that component

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -23,15 +23,77 @@
...
-			<NewGroupConversation v-if="canStartConversations" />
+
+			<!-- Options -->
+			<div class="options"
+				:class="{'options--hidden': isFocused}">
+				<NcActions ref="filterMainBtn"
+					class="filters-button">
+					<template #icon>
+						<FilterIcon v-if="!isFiltered" :size="15" />
+						<FilterCheckIcon v-else-if="isFiltered" :size="15" />
```
Antreesy, line 46:
```suggestion
						<FilterCheckIcon v-else :size="15" />
```
That's also works, though I'd change them, to avoid '!' negation

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -23,15 +23,77 @@
...
+					</template>
+					<NcActionButton close-after-click
+						class="filterButton--Option"
```
Antreesy, line 49:
```suggestion
						class="filters-button__option"
```
We don't use capital letters in classes, and (not 100% sure about it) try to write it close to BEM conception (block__element, block--modifier, block__element--modifier)

And option is more to Select component, maybe call it 'filter-actions__button'?

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -23,15 +23,77 @@
...
+						@click="handleFilter('is:mentioned')">
+						<template #icon>
+							<AtIcon :size="20" />
+						</template>
+						{{ t('spreed','Filter unread mentions') }}
+					</NcActionButton>
+
+					<NcActionButton close-after-click
+						class="filterButton--Option"
```
Antreesy, line 58:
```suggestion
						class="filterButton--Option"
:class={'className--active': isFiltered === 'mentions'} 
```
Concept to show, how we could modify appearance for 'selected' filtration

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -23,15 +23,77 @@
...
+					</NcActionButton>
+
+					<NcActionButton v-if="isFiltered"
+						close-after-click
+						class="filterButton--Clear"
+						@click="clearFilter">
+						<template #icon>
+							<FilterRemoveIcon :size="20" />
+						</template>
+						{{ t('spreed', 'Clear filters: {filterText}', { filterText: filterText }) }}
```
Antreesy, line 73: I think, Joas meant just 'Clear (all) filters'. We can't translate this message as it's divided in two parts, before and after colon. It's different for some languages, and translators couldn't rewrite it to sound correctly

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -46,7 +108,7 @@
 					<template v-if="!initialisedConversations">
 						<LoadingPlaceholder type="conversations" />
 					</template>
-					<Hint v-else-if="searchText && !conversationsList.length"
+					<Hint v-else-if="(searchText && !conversationsList.length) || (!conversationsList.length && isFiltered)"
```
Antreesy, line 111:
```suggestion
					<Hint v-else-if="(searchText || isFiltered) && !conversationsList.length"
```
Easier to read and understand, right? And it's big enough, we could move it to computed property

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -311,12 +404,27 @@ export default {
 		getFocusableList() {
 			return this.$el.querySelectorAll('li.acli_wrapper .acli')
 		},
+		setIsFocused(value) {
+			this.isFocused = value
+		},
+
+		handleFilter(filter) {
+			this.isFiltered = true
+			this.filterText = filter
```
Antreesy, line 413:
```suggestion
			this.isFiltered = filter
```
Maybe just leave one property, with optional values `['mentions' , 'unread', null]`?

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -311,12 +404,27 @@ export default {
...
+
+		},
+
+		clearFilter() {
```
Antreesy, line 417: and replace with hadleFilter(null)

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -654,11 +765,41 @@ export default {
...
+	position : absolute;
 
 	:deep(.input-field__input) {
 		border-radius: var(--border-radius-pill);
 	}
+	&--expanded {
+		width: 90%;
+	}
+
+	&--filter{
```
Antreesy, line 780: So that means, we cannot use filters and search together? What if I want to find all unread conversation with 'Talk' in theirs name?
I think we need a two-layer filtration here, but maybe I missed something in the discussion

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -654,11 +765,41 @@ export default {
...
+}
+
+.options{
+	position: relative;
+	z-index: 1;
+	left : calc(65% + 10px);
+	display: flex;
+	gap: 0.25px;
+
+	&--hidden{
```
Antreesy, line 795: Try to use class 'visually-hidden', it should work better

src/components/LeftSidebar/SearchBox/SearchBox.vue
```diff
@@ -52,7 +53,7 @@ export default {
 		 */
 		placeholderText: {
 			type: String,
-			default: t('spreed', 'Search conversations or users'),
+			default: t('spreed', 'Search…'),
```
Antreesy, line 56:
```suggestion
			default: t('spreed', 'Search …'),
```

src/components/LeftSidebar/LeftSidebar.vue
Antreesy: so, `visibility: hidden;` throws out buttons from the tabbable stream, so instead of Search => Filter => Add, I jump from Search => First Chat

marcoambrosini (commented):
Can we make the filter button primary instead of changing the icon when the filters are active?

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -23,15 +23,79 @@
 	<NcAppNavigation :aria-label="t('spreed', 'Conversation list')">
 		<div class="new-conversation"
 			:class="{ 'new-conversation--scrolled-down': !isScrolledToTop }">
-			<SearchBox v-model="searchText"
+			<SearchBox ref="searchbox"
+				v-model="searchText"
 				class="conversations-search"
+				:class="{'conversations-search--expanded': isFocused,
+					'conversations-search--filter' : isFiltered }"
```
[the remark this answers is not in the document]
Antreesy: And modify styles `:deep(input[disabled="true"])` with FIXME comment, until component appearance in disabled state isn't changed in vue components library

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -23,15 +23,63 @@
...
+			<div class="options"
+				:class="{'hidden-visually': isFocused}">
+				<NcActions class="filter-actions"
+					:primary="isFiltered !== null">
+					<template #icon>
+						<FilterIcon :size="15" />
+					</template>
+					<NcActionButton close-after-click
+						class="filter-actions__button"
+						:primary="isFiltered === 'mentions'"
```
Antreesy, line 49:
```suggestion
						:class="{'filter-actions__button--active': isFiltered === 'mentions'}"
```
Action buttons don't have that prop, so we need our own class, like you did for the next button

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -203,19 +264,32 @@ export default {
 			preventFindingUnread: false,
 			roomListModifiedBefore: 0,
 			forceFullRoomListRefreshAfterXLoops: 0,
+			isNewGroupConversationOpen: false,
```
Antreesy, line 267:
```suggestion

```

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -334,6 +418,10 @@ export default {
 			}
 		}, 250),
 
+		toggleNewGroupConversation(value) {
+			this.isNewGroupConversationOpen = value
+		},
+
```
Antreesy, line 424:
```suggestion
```

We don't have any components for this, as you removed button. It's fine, let's clear it here and implement in another PR

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -654,11 +741,43 @@ export default {
...
 	:deep(.input-field__input) {
 		border-radius: var(--border-radius-pill);
 	}
+	&--expanded {
+		width: 90%;
+	}
+
+}
+
+//FIXME : this should be changed once the disabled style for input is added
```
Antreesy, line 758:
```suggestion
//FIXME upstream: this should be changed once the disabled style for NcInputField is added
```

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -654,11 +741,43 @@ export default {
...
+:deep(.input-field__input[disabled="disabled"]){
+	background-color: var(--color-background-dark);
+}
+
+.options{
+	position: relative;
+	transition: all 0.3s ease; //gets hidden once the search box is expanded
+	left : calc(65% + 4px);
+	display: flex;
+}
```
Antreesy, line 768:
```suggestion
  height: var(--default-clickable-area);
}
```
let's do it like this, because now i see in devtools both yours .hidden-visually and the one from server (overlapping styles)

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -654,11 +741,43 @@ export default {
...
+	background-color: var(--color-primary-element-light);
+	border-radius: 6px;
+	:deep(.action-button__longtext){
+		font-weight: bold;
+	}
+
+}
+
+.hidden-visually{
+	height:100%;
```
Antreesy, line 780:
```suggestion
```

src/components/LeftSidebar/SearchBox/SearchBox.vue
```diff
@@ -25,7 +25,9 @@
 			:value.sync="localValue"
 			:label="placeholderText"
 			:show-trailing-button="isSearching"
+			:disabled="disabled"
 			trailing-button-icon="close"
+			v-on="$listeners"
 			@trailing-button-click="abortSearch"
```
Antreesy, line 31: Need to check that line and method, because text doesn't cleared up then clicking on 'X'


https://github.com/nextcloud/spreed/assets/93392545/f8a7bba9-813f-4a4a-bd5e-a7b73cf3c17c

Antreesy:
```vue
				:value.sync="searchText"
```
Try this to synchronise both SearchBox and TextField values, and then check for abortSearch()

marcoambrosini (changes requested):
I don't think that the search box should be disabled when a filter is active, why are we doing that? The search box should stay clickable so that the search operation can be initiated immediately without worrying about filters.

DorraJaouad:
> I don't think that the search box should be disabled when a filter is active, why are we doing that? The search box should stay clickable so that the search operation can be initiated immediately without worrying about filters.

For example, the user can type a text in the search box and activates a filter right after. As the text remains there, It will be mistakenly thought that the search text is used to filter out the search output. Vice versa, if the search box remains clickable, the user can try to write a textual filter while the filter is active. 

>2. the search filter can not be applied together with a textual filter

nickvergessen:
> For example, the user can type a text in the search box and activates a filter right after. As the text remains there, It will be mistakenly thought that the search text is used to filter out the search output. Vice versa, if the search box remains clickable, the user can try to write a textual filter while the filter is active.

I commented this in the beginning as well. It would be cool to to be able to filter and additionally still search. Should be doable?

DorraJaouad:
Yes, we can still amend it but the whole feature was kinda built on the fact that both textual and new filters cannot be combined together. The search box expansion was added purposely to hide the filters during the search operation. Also, there was a reason behind this separation. 

>..For the purpose of talk search, I don't think we need filtering of the search results. That would be also a bit misleading because the search results are not only joined conversations, but also not joined ones, for which there 's no read/unread state.
https://github.com/nextcloud/spreed/pull/9688#issuecomment-1594055712

I would change it if you all think it is better to combine them. It was like that in the very beginning.

marcoambrosini:
> the user can type a text in the search box and activates a filter right after

@DorraJaouad this shouldn't be possible because as long as there's a search operation undergoing we are not displaying the filter button.

@nickvergessen  [As I raised previously](https://github.com/nextcloud/spreed/pull/9688#issuecomment-1594055712) I think that mixing filtering and searching functionality would complicate things a lot and I see very little benefit to it. 

This is what I would suggest as a first iteration of this: 
When a filter is active and the user initiates a search, we should just disregard the filter and return the search results normally. Once the user selects a search results we go back to whatever filtering option the user had previously selected. 

Also, we shouldn't forget placing an empty content view if the filtering doesn't return any conversation :)

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -23,15 +23,63 @@
...
+						<template #icon>
+							<FilterRemoveIcon :size="20" />
+						</template>
+						{{ t('spreed', 'Clear filters') }}
+					</NcActionButton>
+				</NcActions>
+				<!-- New Conversation -->
+				<NewGroupConversation v-if="canStartConversations"
+					ref="newGroupConversation" />
+			</div>
```
nickvergessen, line 80: Solution for:

> My final pain point would be that the + button should stay visible when searching. If you can not find the conversation you are looking for, you will  have to create it, but for that you (currently) first have to clear the search term, which is rather unexpected (in the past we even had the feature that the current search term was automatically prefilled as room name, but that seems to be broken even on stable27)

```suggestion
			</div>

			<!-- New Conversation -->
			<NewGroupConversation v-if="canStartConversations"
				ref="newGroupConversation" />
```

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -203,19 +264,31 @@ export default {
...
+			default:
+				if (this.searchText !== '') {
+					const lowerSearchText = this.searchText.toLowerCase()
+					conversations = conversations.filter(conversation =>
+						conversation.displayName.toLowerCase().includes(lowerSearchText)
+								|| conversation.name.toLowerCase().includes(lowerSearchText)
+					)
+					break
+				}
 			}
```
nickvergessen, line 292: Solution for Marco's comment:

> When a filter is active and the user initiates a search, we should just disregard the filter and return the search results normally. Once the user selects a search results we go back to whatever filtering option the user had previously selected.

```suggestion
			if (this.searchText !== '') {
				const lowerSearchText = this.searchText.toLowerCase()
				conversations = conversations.filter(conversation =>
					conversation.displayName.toLowerCase().includes(lowerSearchText)
							|| conversation.name.toLowerCase().includes(lowerSearchText)
				)
			} else if (this.isFiltered === 'unread') {
				conversations = conversations.filter(conversation => conversation.unreadMessages > 0)
			} else if (this.isFiltered === 'mentions') {
				conversations = conversations.filter(conversation => conversation.unreadMention || (conversation.unreadMessages > 0
					&& (conversation.type === CONVERSATION.TYPE.ONE_TO_ONE || conversation.type === CONVERSATION.TYPE.ONE_TO_ONE_FORMER))
			}
```

nickvergessen (changes requested):
> https://github.com/nextcloud/spreed/pull/9688#issuecomment-1594055712 I think that mixing filtering and searching functionality would complicate things a lot and I see very little benefit to it.

Alright, was not aware of this since there were too many comments when I returned and got asked. Also makes a bit of sense, so lets see how that feels naturally.

> When a filter is active and the user initiates a search, we should just disregard the filter and return the search results normally. Once the user selects a search results we go back to whatever filtering option the user had previously selected.

Should be easily doable by making the switch block an if-statement with the search as first branch.

---

My final pain point would be that the + button should stay visible when searching. If you can not find the conversation you are looking for, you will  have to create it, but for that you (currently) first have to clear the search term, which is rather unexpected (in the past we even had the feature that the current search term was automatically prefilled as room name, but that seems to be broken even on stable27)

---

After this we should get the PR in and then consider doing follow ups, to not drag this PR longer than needed.

nickvergessen:
> Also, we shouldn't forget placing an empty content view if the filtering doesn't return any conversation :)

This is already there

marcoambrosini:
@nickvergessen how about reintroducing that feature as the last search result + empty content of a search instead of leaving the + button visible?

nickvergessen:
Then let's move that to a follow up too.

So @DorraJaouad todo:
- [ ] Remove disable from the searchbox input field
- [ ] replace switch with the if from https://github.com/nextcloud/spreed/pull/9688#discussion_r1244656659

nickvergessen (approved):
I think we need some padding between the button and the input, but otherwise good to go from my side:
grafik

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -23,15 +23,64 @@
...
 				@input="debounceFetchSearchResults"
 				@submit="onInputEnter"
 				@keydown.enter.native="handleEnter"
 				@abort-search="abortSearch" />
-			<NewGroupConversation v-if="canStartConversations" />
+
+			<!-- Options -->
+			<div class="options"
+				:class="{'hidden-visually': isFocused}">
+				<NcActions class="filter-actions"
```
Antreesy, line 41:
```suggestion
		<NcActions class="filter-actions"
		  :class="{'hidden-visually': isFocused}"
```

It's probably possible to move these styles to the NcActions now, without additional div wrapper

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -311,12 +383,24 @@ export default {
 		getFocusableList() {
 			return this.$el.querySelectorAll('li.acli_wrapper .acli')
 		},
+		setIsFocused(event) {
+			if (event.relatedTarget?.className.includes('input-field__clear-button') || this.searchText !== '') {
```
Antreesy, line 387: Like that behaviour, when blurred input stays expanded with text. Make sense, as we don't have filters applied at the moment
image

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -654,11 +738,44 @@ export default {
 }
 
 .conversations-search {
-	flex-grow: 1;
-
+	transition: all 0.3s ease;
+	z-index: 1;
+	// new conversation button width : 52 px
+	// Filters button width : 44 px
```
Antreesy, line 744: Could be replaced with `var(--default-clickable-area)`

+ to Joas' commit: we could decrease width up to 3 more px to get a nice gap between input and button

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -654,11 +738,44 @@ export default {
...
+	// Total : 97 px
+	width : calc(100% - 97px);
+	display : flex;
 	:deep(.input-field__input) {
 		border-radius: var(--border-radius-pill);
 	}
+	&--expanded {
+
+		// Gets expanded : 100 % - (52px + 1px)
+		width : calc(100% - 53px );
```
Antreesy, line 755: Works fine in current state, but I'd like to refactor styles a bit in follow-up, because `pos:absolute` might misbehaves on different resolutions

src/components/LeftSidebar/SearchBox/SearchBox.vue
```diff
@@ -69,29 +71,20 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		disabled: {
```
Antreesy, line 74: We don't use `disabled` prop anymore, so could be removed completely from component (it makes a little sense to keep it just in case)

src/mixins/arrowNavigation.js
```diff
@@ -43,7 +43,7 @@ const arrowNavigation = {
 		mountArrowNavigation() {
 			document.addEventListener('keydown', (event) => {
 				// https://www.w3.org/WAI/GL/wiki/Using_ARIA_menus
-				if (this.isFocused()) {
+				if (this.isSearching) {
```
Antreesy, line 46: Also check as a follow-up: I'm not sure if this mixin is still working

Antreesy (approved):
Works and looks good, visually and by the code
Nothing blocking from my side, what we couldn't done later (but I left some commits to consider as a follow-up)

jancborchardt (approved):
Looks nice! :)

PR #9706, opened 2023-06-05
Don't scroll conversation list for visible conversations

### LINKED ISSUE #9582: Don't scroll to conversations that are already visible
### How to use GitHub

---

## Steps to reproduce
1. Select a conversation which is at the middle of the conversation list

### Expected behaviour
The conversation list should not scroll as the selected conversation is already visible on the screen

### Actual behaviour
The conversation list scrolls so that the selected conversation is at the top of the conversation list.

## Talk app

**Talk app version:** Seeing this on stable26 and stable27. On 26 it's only happening after a page reload (which is already annyoing), on 27 it's always happening.

Antreesy:
### ☑️ Resolves

* Fix #9582

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist


SystemKeeper (approved):
Did not look at the code, but working great on Safari and Chrome on MacOS!

ShGKme (changes requested):
Looks good in the code.

And on my test it works great now on selecting new chat (both manually and programmatically).

pr-1

But it breaks scrolling after opening a conversation from search. (That is why we needed scrolling here).

pr-2

My guess - that was the result of removing that `// FIXME` part.

Antreesy:
> My guess - that was the result of removing that `// FIXME` part.

Yes, it seems that we need to wait for $nextTick for the conversation list to re-render. Pushed changes, tested it on Chrome and Firefox.

ShGKme (approved):
All works great now with:
- select by click
- select programmatically
- select in view, above, below
- select from search

Just some comments from the code side

PR #9709, opened 2023-06-06
Fix error when creating a conversation from search results

Antreesy:
### ☑️ Resolves

* Regression from #9465 

To test:
* have a user group (`admin` is default one)
* type group name in search field
* click on a search result -> create conversation modal should be opened

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks

- [ ] Test on browser

### 🏁 Checklist


ShGKme (approved):
🦤

PR #9721, opened 2023-06-08
Fix dark/light theme in `LoadingPlaceholder` and refactor dark theme detection

### LINKED ISSUE #9720: The loading animation should not be as dark. Please use a brighter color for it.
As discussed with @jancborchardt and @nimishavijay, the loading animation should not be as dark. Please use a brighter color for it.

See the following screenshot:
image

ShGKme:
### ☑️ Resolves

* Fix #9720
* Fix https://github.com/nextcloud/talk-desktop/issues/45

### 🖼️ Screenshots

🏚️ Before LoadingPlaceholder | 🏡 After LoadingPlaceholder
---|---
Default on 🌑 Dark system |
image | image
🌑 Dark theme with 🌑 Dark system |
image | image
🌕 Light theme with 🌑 Dark system |
image | image 
Default on 🌕 Light system  |
image | image
🌕 Light theme with 🌕 Light system |
image | image
🌑 Dark theme with 🌕 Light system |
image | image

### Fix dark/light theme in `LoadingPlaceholder`

In Nextcloud we have CSS variables for themes:
1. `:root` - default values based on system `preferred-color-scheme`
2. On `<body>` - set by current theme if not "System default theme" is used in settings. These styles override defaults.

The `LoadingPlaceholder` used variables from `documentElement` (`:root`) instead of `body`. It caused a problem when the system-preferred theme was not equal to the selected in Nextcloud settings.

- [x] Use styles from `body` instead of `documentElement` in `LoadingPlaceholder`
- [x] Move styles getting from instance's `mounted` to component's ES module
  - Render the component right away with the styles instead of useless re-render after mount.
  - Do not get the value from DOM API for each instance, get only once.

### Refactor dark theme detection

- [x] Add `isDarkTheme` util

In some components, we checked if a dark theme is enabled by this line:
```js
const isDarkTheme = window.getComputedStyle(document.body)
  .getPropertyValue('--background-invert-if-dark') === 'invert(100%)'
```

Problems:
- Duplication
- It is evaluated for each instance again using DOM API to get the same value

New `isDarkTheme` util checks if the dark theme is enabled only once and then exports the result.

It assumes that the dark/light theme is not been changing without page reload. But this case didn't fully work early anyway...

No changes after the refactoring

image | image
---|---

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/Mention.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/Mention.vue
@@ -47,6 +47,8 @@ import { generateOcsUrl } from '@nextcloud/router'
 
 import NcUserBubble from '@nextcloud/vue/dist/Components/NcUserBubble.js'
 
+import { isDarkTheme } from '../../../../../utils/isDarkTheme.js'
+
 export default {
 	name: 'Mention',
 
@@ -97,7 +99,7 @@ export default {
 				&& loadState('spreed', 'user_group_ids', []).includes(this.id)
 		},
 		avatarUrl() {
-			return generateOcsUrl('apps/spreed/api/v1/room/{token}/avatar' + (this.darkTheme ? '/dark' : ''), {
+			return generateOcsUrl('apps/spreed/api/v1/room/{token}/avatar' + (isDarkTheme ? '/dark' : ''), {
 				token: this.id,
 			})
 		},
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/Mention.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/Mention.vue
@@ -96,10 +96,6 @@ export default {
 			return this.isGroupMention
 				&& loadState('spreed', 'user_group_ids', []).includes(this.id)
 		},
-		isDarkTheme() {
-			return window.getComputedStyle(document.body)
-				.getPropertyValue('--background-invert-if-dark') === 'invert(100%)'
-		},
 		avatarUrl() {
 			return generateOcsUrl('apps/spreed/api/v1/room/{token}/avatar' + (this.darkTheme ? '/dark' : ''), {
 				token: this.id,
--- a/src/utils/isDarkTheme.js
+++ b/src/utils/isDarkTheme.js
@@ -1,0 +1,41 @@
+/*
+ * @copyright Copyright (c) 2023 Grigorii Shartsev <me@shgk.me>
+ *
+ * @author Grigorii Shartsev <me@shgk.me>
+ *
+ * @license AGPL-3.0-or-later
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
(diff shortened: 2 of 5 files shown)


src/utils/isDarkTheme.js
```diff
@@ -0,0 +1,41 @@
...
+ * Check if dark theme is used
+ *
+ * @return {boolean}
+ */
+function checkIfDarkTheme() {
+	// Nextcloud uses --background-invert-if-dark for dark theme filters in CSS
+	// Values:
+	// - 'invert(100%)' for dark theme
+	// - 'no' for light theme
+	return window.getComputedStyle(document.body).getPropertyValue('--background-invert-if-dark') === 'invert(100%)'
```
szaimen, line 32: @skjnldsv isnt there a better way to get the setting in the frontend?

szaimen:
> @szaimen Do you think, we should backport it to 27?

If this is only happening since 27 I would say yes since it is a bug :)

ShGKme:
> If this is only happening since 27 I would say yes since it is a bug :)

No, it is an old bug

szaimen:
> > If this is only happening since 27 I would say yes since it is a bug :)
> 
> No, it is an old bug

Then I'd say definitely backport to 27 and maybe even further?

szaimen (commented):
✅ LGTM from the screenshots. Thanks for tackling this! :)

src/utils/isDarkTheme.js
ShGKme: In theory, we can use data theme(s) attributes, but I'm not sure, if it works with custom themes from appstore

skjnldsv: No idea tbh :/
We should most likely provide an official way for this

Ideally it's
1. has the user picked a custom theme
2. is the system asking for dark mode

src/components/MessagesList/MessagesGroup/Message/MessagePart/Mention.vue
```diff
@@ -96,12 +98,8 @@ export default {
 			return this.isGroupMention
 				&& loadState('spreed', 'user_group_ids', []).includes(this.id)
 		},
-		isDarkTheme() {
-			return window.getComputedStyle(document.body)
-				.getPropertyValue('--background-invert-if-dark') === 'invert(100%)'
-		},
 		avatarUrl() {
-			return generateOcsUrl('apps/spreed/api/v1/room/{token}/avatar' + (this.darkTheme ? '/dark' : ''), {
+			return generateOcsUrl('apps/spreed/api/v1/room/{token}/avatar' + (isDarkTheme ? '/dark' : ''), {
```
Antreesy, line 102: Fixed this file for chat mentions:
image

Because of the wrong naming, it wasn't working before, so not your fault to miss it =)

Antreesy (approved):
Everything else os working good, thanks for handling this!
Tested on Ubuntu + Chrome for all 6 variations

src/utils/isDarkTheme.js
szaimen: That sounds good!

Antreesy:
FYI: removed label as non-blocking for 27 release

PR #9727, opened 2023-06-09
Make pagination buttons in call (grid, stripe views) visible on call view hover

### LINKED ISSUE #9505: Call view pagination button is no longer visible
### How to use GitHub

---

## Steps to reproduce
1. Hover the call view and try to find the `>` button to go to the next page of videos

Without hovering the button | Hovering the button
---|---
Bildschirmfoto vom 2023-05-09 15-08-43 | Bildschirmfoto vom 2023-05-09 15-08-47

### Expected behaviour
The button should have some background when the mouse is hovering the call view.

### Actual behaviour
I guess a CSS selector needs changing due to Vue lib changes?

## Talk app

**Talk app version:** Desktop 0.4.0

Antreesy:
### ☑️ Resolves

* Fix #9505

### 🖼️ Screenshots

🏚️ Before: see issue

🏡 After:
Without hover | With hover
---|---
Screenshot from 2023-06-09 09-08-27 | Screenshot from 2023-06-09 09-08-34
Screenshot from 2023-06-09 09-07-31 | Screenshot from 2023-06-09 09-07-34

### 🚧 Tasks

- [ ] Visual check

### 🏁 Checklist

### DIFF
--- a/src/components/CallView/Grid/Grid.vue
+++ b/src/components/CallView/Grid/Grid.vue
@@ -963,14 +963,6 @@ export default {
 }
 
 .grid-navigation {
-	z-index: 2;
-	background-color: rgba(0, 0, 0, 0.5);
-
-	&:hover,
-	&:focus {
-		background-color: rgba(0, 0, 0, 0.8) !important;
-	}
-
 	.grid-wrapper & {
 		position: absolute;
 		top: calc(50% - var(--default-clickable-area) / 2);
@@ -1022,10 +1014,14 @@ export default {
 	position: absolute !important;
 	top: calc(-1 * var(--default-clickable-area));
 	right: 0;
+}
+
+.stripe--collapse,
+.grid-navigation {
 	z-index: 2;
 	opacity: .7;
 
-	.app-content:hover & {
+	#call-container:hover & {
 		background-color: rgba(0, 0, 0, 0.1) !important;
 
 		&:hover,


DorraJaouad (approved):
Visual check: 
tested and shown as described.

src/components/CallView/Grid/Grid.vue
```diff
@@ -872,6 +872,10 @@ export default {
 	position: relative;
 	bottom: 0;
 	left: 0;
+
+	&:hover .grid-navigation {
+		background-color: rgba(0, 0, 0, 0.8) !important;
+	}
```
ShGKme, line 878: Isn't it the same as `hover` and `focus` values, so there is no hover/focus effect now on those buttons?

Antreesy: I left previous styles, so `hover` and `focus` works for buttons as before. But if we're moving a mouse over `<div class="wrapper">` now, buttons should also be highlighted

Antreesy: There was a misunderstanding, as wrapper:hover buttons and buttons:hover have the same styles applied, we couldn't distinguish them. Should look better now with a commit below

src/components/CallView/Grid/Grid.vue
```diff
@@ -872,6 +872,10 @@ export default {
 	position: relative;
 	bottom: 0;
 	left: 0;
+
+	&:hover .grid-navigation {
+		background-color: rgba(0, 0, 0, 0.5) !important;
+	}
```
ShGKme, line 878: This is the same value that this button has initially without wrapper hovering:

https://github.com/nextcloud/spreed/blob/2aabfbb04a5deea7976b74ffe9b1a21076e5d4ec/src/components/CallView/Grid/Grid.vue#L969-L971

ShGKme (changes requested):
Now there is no `.wrapper:hover` effect on navigation buttons.

What about using the same styles as are used for stipe-collapse button in almost the same place? IMO, they should have consistent styles anyway.

https://github.com/nextcloud/spreed/blob/2aabfbb04a5deea7976b74ffe9b1a21076e5d4ec/src/components/CallView/Grid/Grid.vue#L1025-L1040

ShGKme:
Currently these buttons are not well-aligned relative to grid/video/page border. It became visible with a new wrapper:hover effect.

image

We are choosing between 8px (same as other paddings on this page) and 0px (in-line with video).

What do you think?
cc @marcoambrosini 
cc @SystemKeeper
cc @mahibi 

8px | 0px 
---|---
image | image
image | image

SystemKeeper:
I would go with the 8px personally.

ShGKme (approved):

PR #9629, opened 2023-05-26
chore: update workflows from templates

skjnldsv:
This PR was automatically generated by the workflow dispatcher

PR #9733, opened 2023-06-09
Fix CI with backports from master

ShGKme:
Fixes failing CI

Manual backport of:
- https://github.com/nextcloud/spreed/pull/9629
  - Fixes failing CI because of wrong PHP extensions in `Set up php`
- https://github.com/nextcloud/spreed/pull/9666
  - Fixes failing CI because of out of space (there is an alternative solution - use Ubuntu 22.04)
- https://github.com/nextcloud/spreed/pull/9672
  - I'm not sure this is actually required...

### DIFF
--- a/.github/workflows/appstore-build-publish.yml
+++ b/.github/workflows/appstore-build-publish.yml
@@ -44,7 +44,7 @@ jobs:
           expression: "//info//dependencies//nextcloud/@min-version"
 
       - name: Read package.json node and npm engines version
-        uses: skjnldsv/read-package-engines-version-actions@1bdcee71fa343c46b18dc6aceffb4cd1e35209c6 # v1.2
+        uses: skjnldsv/read-package-engines-version-actions@0ce2ed60f6df073a62a77c0a4958dd0fc68e32e7 # v2.1
         id: versions
         # Continue if no package.json
         continue-on-error: true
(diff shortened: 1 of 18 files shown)


ShGKme:
@skjnldsv We had many problems on CI on stable27:
- With PHP extensions (no GC)
- With out of space

All it works fine (almost) on master. So I decided that it is better to have the same CI on stable27 than its own fixes like using Ubuntu 22.04 or manually change ONLY php extensions.

But I might be wrong if some CI updates in master are actually hard-connected with master and should not be used on 27...

Could you confirm, that this is ok?

.github/workflows/appstore-build-publish.yml
```diff
@@ -83,7 +83,7 @@ jobs:
         if: steps.check_composer.outputs.files_exists == 'true'
         run: |
           cd ${{ env.APP_NAME }}
-          composer install --no-dev --classmap-authoritative
+          composer install --no-dev
```
SystemKeeper, line 86: We should create a follow up issue to this and clarify (also for master).

SystemKeeper (approved):
Approving this now as release needs to be done. Can't find anything obviously wrong.

skjnldsv (approved):

PR #9777, opened 2023-06-14
feat(MessageGroup) - group system messages

### LINKED ISSUE #9678: Frontend - Group system messages from calls
* Overview issue at https://github.com/nextcloud/spreed/issues/1148
* Proposal at https://github.com/nextcloud/spreed/issues/1148#issuecomment-992477832

Antreesy:
### ☑️ Resolves

* Fix #9678
* New component and util for handling groups of system messages added
* Combine messages for following actions:
  * participants being added to the conversations
  * participants promoted to / demoted from moderators
  * participants joined / left the call
  * participant reconnected to the call
* Light sort of messages within one group:
  `user added/removed` => `moderator promoted/demoted` => `call started` => `recording started` => `call joined/left` => `call ended`

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot from 2023-08-03 12-52-48 | image 

### 🚧 Tasks

- [ ] Design: appearance of header and collapsed messages, toggle button, animation
- [ ] RFC: re-sorting system messages within one MessagesSystemGroup: define a list of system messages being sorted, and theirs order (sorting of related messages brings weird results , like `call_left` and `call_joined` in various combinations, so they're not sorted between each other)
- [ ] Follow-up - test coverage
- [ ] Follow-up - messages re-rendering

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesGroup/MessagesSystemGroup.vue
+++ b/src/components/MessagesList/MessagesGroup/MessagesSystemGroup.vue
@@ -164,6 +164,27 @@ export default {
 				return 'user_added'
 			}
 
+			// Group users reconnected in a minute
+			if (message1.systemMessage === 'call_joined'
+				&& message2.systemMessage === 'call_left'
+				&& message1.timestamp - message2.timestamp < 60 * 1000
+				&& message1.actorId === message2.actorId
+				&& message1.actorType === message2.actorType) {
+				return 'call_reconnected'
+			}
+
+			// Group users joined call one by one
+			if (message1.systemMessage === 'call_joined'
+				&& message1.systemMessage === message2.systemMessage) {
+				return 'call_joined'
+			}
+
+			// Group users left call one by one
+			if (message1.systemMessage === 'call_left'
+				&& message1.systemMessage === message2.systemMessage) {
+				return 'call_left'
+			}
+
 			return ''
 		},
 
@@ -177,12 +198,21 @@ export default {
 					groups.push({ messages: [message], type: '', collapsed: true })
 					forceNextGroup = false
 				} else {
+					if (groupingType === 'call_reconnected') {
+						groups.push({ messages: [groups.at(-1).messages.pop()], type: '', collapsed: true })
+						forceNextGroup = true
+					}
 					groups.at(-1).messages.push(message)
 					groups.at(-1).type = groupingType
 				}
 				lastMessage = message
 			}
 			return groups
+		},
+
+		checkIfSelfIsActor(message) {
+			return message.messageParameters.actor.id === this.$store.getters.getActorId()
+				&& message.messageParameters.actor.type + 's' === this.$store.getters.getActorType()
 		},
 
 		checkIfSelfIsUser(message) {
@@ -192,6 +222,17 @@ export default {
 
 		createCombinedSystemMessage({ messages, type }) {
 			const combinedMessage = cloneDeep(messages[0])
+
+			// Handle cases when users reconnected to the call
+			if (type === 'call_reconnected') {
+				if (this.checkIfSelfIsActor(combinedMessage)) {
+					combinedMessage.message = t('spreed', 'You reconnected to the call')
+				} else {
+					combinedMessage.message = t('spreed', '{actor} reconnected to the call')
+				}
+
+				return combinedMessage
+			}
 
 			// clear messageParameters to be filled later
 			const actor = messages[0].messageParameters.actor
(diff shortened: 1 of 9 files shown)


src/components/MessagesList/MessagesGroup/MessagesSystemGroup.vue
```diff
@@ -0,0 +1,456 @@
...
+						combinedMessage.message = t('spreed', 'You, {user0} and {hiddenUsers} more participants left the call', { hiddenUsers: usersCounter - 2 })
+						break
+					}
+				} else {
+					switch (usersCounter) {
+					case 2 :
+						combinedMessage.message = t('spreed', '{user0} and {user1} left the call')
+						break
+					case 3 :
+						combinedMessage.message = t('spreed', '{user0}, {user1} and 1 more participant left the call')
```
nickvergessen, line 359: that's not how plurals work :-X

Antreesy: :see_no_evil: Sure, will overwrite with `n('spreed')` later

marcoambrosini (changes requested):
Very nice, could we please always align the button with the date? Meaning to the first line of the system message

src/components/MessagesList/MessagesGroup/MessagesSystemGroup.vue
```diff
@@ -0,0 +1,614 @@
...
+	width: 100%;
+	min-width: 0;
+
+	&--header {
+		position: relative;
+		& &__toggle {
+			position: absolute;
+			top: 50%;
+			right: 0;
+			transform: translateY(-50%);
```
Antreesy, line 605:
```suggestion
			top: -6px;
			right: 0;
```

image
_Note to self_

Antreesy:
follow-up: messages are re-rendered on each fetch, so hidden messages collapse each time and don't persist a state

jancborchardt (changes requested):
Very nice @Antreesy! One detail – as per [design guidelines](https://docs.nextcloud.com/server/latest/developer_manual/design/foundations.html#typography) we never use italic font as it can be difficult to read for people. Since they are in a container already they do not need to be italic as well. :)

Antreesy:
Marked as a draft, as it will conflict with currently pending refactor+bugfix PR and needs to be rebased afterwards:
- [ ] Related PR: #9897

ShGKme:
Is it still work in progress, or ready for code review?

Antreesy:
> Is it still work in progress, or ready for code review?

There still is a room for improvement, but mostly about error cleanup and refactoring.
Main Feature is working, so the code could be checked already

marcoambrosini (commented):
Hi, I think we can hide this action until te group is hovered, what do you think @nextcloud/designers

Antreesy:
> we can hide this action until te group is hovered

Sounds nice, and UI will be less disturbed, but don't we still need to mark grouped messages somehow (so the user knows, whether it's a single system message or a grouped one, without hovering)?

marcoambrosini:
Maybe underlying "and 7 more participants" and making it into an `<a>` that expands the content on click?

nimishavijay:
Looks really really nice! :)

> we can hide this action until the group is hovered

If there are no a11y issues with this, then this sounds great! This would be the ideal solution. Maybe @Pytal or @JuliaKirschenheuter could help with figuring out if showing the button only on hover is okay? :) 

> Maybe underlying "and 7 more participants" and making it into an <a> that expands the content on click?

I'm not too sure about this as it's not really a link at all, and we rarely (if at all) use link styling to indicate other actions.

Pytal:
> Looks really really nice! :)
> 
> > we can hide this action until the group is hovered
> 
> If there are no a11y issues with this, then this sounds great! This would be the ideal solution. Maybe @Pytal or @JuliaKirschenheuter could help with figuring out if showing the button only on hover is okay? :)

Hiding the action to reduce visual clutter sounds good but make sure that it is tabbable and visible when focused/hovered

Antreesy:
@nimishavijay @marcoambrosini stylized toggle button as`MessageButtonBar`, which we have for normal message (with the same logic). Button is visible on hover, and in open state:

image

nimishavijay (approved):
Looks really nice! :)

src/components/MessagesList/MessagesList.vue
```diff
@@ -1113,6 +1116,10 @@ export default {
 		onWindowFocus() {
 			this.refreshReadMarkerPosition()
 		},
+
+		isSystemMessagesGroup(group) {
+			return group.isSystemMessagesGroup ? MessagesSystemGroup : MessagesGroup
+		},
```
ShGKme, line 1122:
```suggestion
		messagesGroupComponent(group) {
			return group.isSystemMessagesGroup ? MessagesSystemGroup : MessagesGroup
		},
```

`is` prefix is usually used for boolean flags

ShGKme: `MessagesGroup` can now be removed in the `components` options, because it is not used in the template by registration name

src/utils/createCombinedSystemMessage.js
ShGKme, line 1: I'd not make this method a utility. It is not a universal utility, used only in one component and very specific messages-related logic. Also, with this util component now uses store in 2 different ways at the same time:
1. Injected store from `this.$store`
2. Directly imported `store from '../store/index.js'`

If it uses a store, it should be composable as it is part of a component (it uses the store from a component context).

Or it can be a service/"module in the component's directory" with the actor data passed as arguments.

ShGKme (approved):
Seems to works fine. Nothing blocking, but some naming/architecture comments

nickvergessen:
> Conflicting files
src/components/MessagesList/MessagesGroup/Message/Message.vue

Antreesy:
Rebased onto master
@ShGKme please take a look again - i've turned the utiil into composable

src/components/MessagesList/MessagesGroup/MessagesSystemGroup.vue
```diff
@@ -0,0 +1,337 @@
...
+				<ul v-show="messagesCollapsed.messages?.length === 1 || !messagesCollapsed.collapsed"
+					class="messages"
+					:class="{'messages--collapsed': messagesCollapsed.messages?.length > 1}">
+					<Message v-for="message of messagesCollapsed.messages"
+						:key="message.id"
+						v-bind="message"
+						:next-message-id="getNextMessageId(message)"
+						:previous-message-id="getPrevMessageId(message)" />
+				</ul>
+			</div>
```
ShGKme, line 52: This component has many methods calls in template (that a called during the re-rendering) and those methods are not small utils or formatting methods but data processing methods.

It won't be a problem if it doesn't have unnecessary re-renderings, but currently `MessagesList` re-renders each new message.

For example, a user has a chat with mostly only calls, but large calls, for example, 100 people. Then in the history there will be many groups with 100 system messages about leaving the call. And every message, reaction, poll update will re-render the list with recalculating all these methods looping over 100 messages for each group.

IMO, we can keep it in this way but then do one of:
- Find a reason of re-renderings
- Refactor all these components, remove unnecessary props
- Remove these calculating inside rendering
- Wait for Vue 3 migration =D

ShGKme: Also, I'd prefer to use `in` in `v-for` directives to use only one style. (`v-for of` is not very often used in our components).

ShGKme (approved):
Tested manually in basic cases ✅

These components and grouping logic has many edge cases. IMO, we should cover it with tests.

src/components/MessagesList/MessagesGroup/MessagesSystemGroup.vue
Antreesy: Moved to follow-up

Issue #9626, opened 2023-05-25
File upload dialogue inaccessible when image preview is open during call


ZetaTom:
### How to use GitHub
* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Steps to reproduce
1. participate in active call
2. open preview of any previously shared image in chat
3. drag file into chat to upload

### Expected behaviour
Dialogue appears and one is able to confirm the upload.

### Actual behaviour
Confirmation dialogue for upload is hidden behind open preview window and thus not accessible.

## Browser
**Operating system:** Fedora Linux 38

**Browser name:** Firefox

**Browser version:** 113.0

PR #9792, opened 2023-06-16
Fix media settings in files sidebar

### LINKED ISSUE #8709: Device preview prevents call from starting while editing document
Hello devs.

Here is my problem. When opening a never before opened document and enter the chat for that very document via sidebar I cannot start a call because obviousely talk tries to open the device preview window. It works if I go back to talk and disable the "always show preview screen before joining a call in this conversation". Then starting a call works while having the document open.

I was thinking, maybe pop-up windows are disabled in the browser but it shouldn't work then in the talk app, right?

Question: Is it possible to disable the device preview globally?

## Steps to reproduce
1. open a document
2. start a call
3. nothing happens

### Expected behaviour
video call should start

### Actual behaviour
video call does not start

## Talk app

**Talk app version:** 15.0.3

## Browser

**Microphone available:** yes

**Camera available:** yes

**Operating system:** Ubuntu 22.0.4

**Browser name:** Firefox 109.1 / Chromium

## Server configuration

**Operating system**: Ubuntu

**Web server:** Apache

**Database:** MySQL

**PHP version:** 8.1

**Nextcloud Version:** 25.0.3

marcoambrosini:
When starting a call from the files sidebar and while editing a file, the media settings was not display because it was covered by the file dialog. Increasing the z-index of the media settings dialog didn't fix the issue (it's unclear to me why not). Removing the container of the media settings dialog and thereby assigning the default body as container fixes the issue.

### ☑️ Resolves

* Fix #8709 

### 🖼️ Screenshots

Screenshot 2023-06-16 at 13 45 12

### DIFF
--- a/src/components/MediaSettings/MediaSettings.vue
+++ b/src/components/MediaSettings/MediaSettings.vue
@@ -22,7 +22,6 @@
 <template>
 	<NcModal v-if="modal"
 		class="talk-modal"
-		:container="container"
 		@close="closeModal">
 		<div class="media-settings">
 			<h2 class="media-settings__title">
@@ -616,4 +615,5 @@ export default {
 	display: flex !important;
 	max-width: 500px !important;
 }
+
 </style>


Antreesy (changes requested):
As expected, it breaks opening media settings on fullscreen

marcoambrosini:
Hmmm nice catch. Why is it that in Files the media settings are visible even in fullscreen mode while in talk they're not?

ShGKme:
> Hmmm nice catch. Why is it that in Files the media settings are visible even in fullscreen mode while in talk they're not?

As far as I know, in Files there is no real "Fullscreen mode". Only open Viewer and Fullscreen "styles" for the Sidebar.

In Talk the Fullscreen mode is a "real" fullscreen with native [Fullscreen API](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API) for the Talk app (`div#content-vue`) element. Only Talk is "fullscreen-ed", you see no header and ect. In this mode, Talk app is a [top layer](https://developer.mozilla.org/en-US/docs/Glossary/Top_layer) element. No element outside the Talk (`div#content-vue`) can be above, no matter what z-index it has.

That was the reason, why Daniel created the `uiMode` store here: https://github.com/nextcloud/spreed/pull/6119

We need some huge rework here in general. I'm working on it as a part of https://github.com/nextcloud/spreed/issues/9626

rriemann:
I have already notified this to our Nextcloud sales engineer. It is so important for our use to have this fixed. Please make it happen for the next update. :pray:

Antreesy:
Just a thought: maybe we could try to replace mainContainer for the sidebars? For the main view of Talk app it's obvious to have block like #content-talk, but for Files all and others we should also cover the whole screen:

https://github.com/nextcloud/spreed/blob/master/src/mainFilesSidebar.js#L72

PR #9812, opened 2023-06-19
fix(NewMessageNewFileDialog) - fix creation of file from Blank template

### LINKED ISSUE #9779: File doesn't added to conversation, when create from attachments with 'Blank' or no template
### How to use GitHub

---

## Steps to reproduce
1. Open a conversation, select `Attachments` button => `New text file`
2. Select 'Blank' template, if templates are available
3. Create file => see error
image
5. Check that file is created at Files/Talk
6. Try to create file with the same name => see error:
image

### Expected behaviour
File is added, as when it's created from template

### Actual behaviour
File is created, but not visible in conversation

## Talk app

**Talk app version:** 27

**Custom Signaling server configured:** no

Antreesy:
### ☑️ Resolves

* Fix #9779
* Regression from https://github.com/nextcloud/spreed/commit/5ae4c8207f03887ecfd6fec5f6510a1d66689888 - joining of empty template with others into one array
* `this.selectedTemplate?.filename` checks for the path to file, for example, `Templates/Filename.md`. With `/Blank` file, we receive a search error, but file will still be created in `Talk` folder

To reproduce:
1) in Files, click "+" => "Set Templates folder" (could be wrong with button title)
2) Add some templates [from there](https://github.com/nextcloud/example-files/tree/master)
3) in Talk, click :paperclip: => "New text file"
4) Try to create a file with Blank or any of the existing templates
image
5) Everything should work
5.1) Or use only steps 3-4, template is Blank by default

### 🚧 Tasks

- [ ] Code review

### 🏁 Checklist

### DIFF
--- a/src/components/NewMessage/NewMessageNewFileDialog.vue
+++ b/src/components/NewMessage/NewMessageNewFileDialog.vue
@@ -210,17 +210,13 @@ export default {
 
 			let fileData
 			try {
-				let response
-				if (this.selectedTemplate.fileid === -1) {
-					// Don't send template information for Blank
-					response = await createNewFile(filePath)
-				} else {
-					response = await createNewFile(
+				const response = this.selectedTemplate.fileid === -1
+					? await createNewFile(filePath)
+					: await createNewFile(
 						filePath,
 						this.selectedTemplate?.filename,
 						this.selectedTemplate?.templateType,
 					)
-				}
 				fileData = response.data.ocs.data
 			} catch (error) {
 				console.error('Error while creating file', error)
--- a/src/components/NewMessage/NewMessageNewFileDialog.vue
+++ b/src/components/NewMessage/NewMessageNewFileDialog.vue
@@ -210,11 +210,17 @@ export default {
 
 			let fileData
 			try {
-				const response = await createNewFile(
-					filePath,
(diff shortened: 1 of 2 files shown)


src/components/NewMessage/NewMessageNewFileDialog.vue
```diff
@@ -210,11 +210,17 @@ export default {
...
+				if (this.selectedTemplate.fileid === -1) {
+					// Don't send template information for Blank
+					response = await createNewFile(filePath)
+				} else {
+					response = await createNewFile(
+						filePath,
+						this.selectedTemplate?.filename,
+						this.selectedTemplate?.templateType,
+					)
+				}
```
ShGKme, line 223: What about a ternary operator here?

```suggestion
				const response = this.selectedTemplate.fileid === -1 
					? await createNewFile(filePath)
				  : await createNewFile(
						filePath,
						this.selectedTemplate?.filename,
						this.selectedTemplate?.templateType,
					)
```

ShGKme (approved):
Works. 

Though, I didn't understand what was actually changed in the mentioned commit, what caused regression...

marcoambrosini (approved):

Antreesy:
Changed code based on suggestion, tested, still works

PR #9815, opened 2023-06-19
fix(Poll) - hide voters number from open poll if user has not voted

Antreesy:
### ☑️ Resolves

* Currently, when user hasn't voted yet in the poll, server returns voters number as 0 always. See reference: https://github.com/nextcloud/spreed/blob/9b525c2229d73e6989a7182949fa45fdbc84f457/lib/Controller/PollController.php#L245-L256
* PR aims to align frontend behavior with backend - hide poll voters number for people who are not moderators/owners/voted

### 🖼️ Screenshots

🏚️ Before (admin) | 🏚️ Before (user) | 🏡 After (admin) | 🏡 After (user)
---|---|---|---
image | image | image | image
image | image | image | image
image | image | image | image
image | image | image | image
image | image | image | image
image | image | image | image
image | image | image | image

### 🚧 Tasks

- [ ] Visual check
- [ ] Code review
- [ ] Change `Poll` to `Open poll` ???

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/Poll.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/Poll.vue
@@ -63,9 +63,7 @@
 						</h2>
 					</div>
 					<div class="poll__summary">
-						<template v-if="currentUserIsPollCreator || currentUserIsModerator || pollIsPublic">
-							{{ n('spreed', 'Poll results • %n vote', 'Poll results • %n votes', votersNumber) }}
-						</template>
+						{{ pollSummaryText }}
 					</div>
 
 					<!-- options -->
@@ -108,7 +106,7 @@
 				</template>
 
 				<!-- Results page -->
-				<template v-if="modalPage === 'results'">
+				<template v-else-if="modalPage === 'results'">
 					<!-- Title -->
 					<div class="poll__header">
 						<PollIcon :size="20" />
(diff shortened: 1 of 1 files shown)


src/components/MessagesList/MessagesGroup/Message/MessagePart/Poll.vue
```diff
@@ -315,61 +308,46 @@ export default {
...
+			return this.$store.getters.conversation(this.token).participantType
 		},
 
-		/**
-		 * Current actor type
-		 */
-		actorType() {
-			return this.$store.getters.getActorType()
+		selfIsOwnerOrModerator() {
+			return (this.poll.actorType === this.$store.getters.getActorType() && this.poll.actorId === this.$store.getters.getActorId())
```
ShGKme, line 316: Now I feel, we need an util function for this :D

ShGKme (commented):
Looks good by screenshots and the code, but haven't tested manually yet

ShGKme (approved):
Tested, works good, same as on screenshots

ShGKme:
> Change Poll to Open poll ???

Where?

src/components/MessagesList/MessagesGroup/Message/MessagePart/Poll.vue
Antreesy: Maybe a getter? I tried to implement it, as `selfIsActor`, and it shows itself pretty good. Though we have a different checks in different places for this, so should be unified first

Antreesy:
> Where?

For polls which are not closed yet, like this:
`Open poll ° 3 votes`

marcoambrosini (approved):
Not tested but screenshots look good. Agree with shortening the text.

ShGKme:
> For polls which are not closed yet, like this:
> `Open poll ° 3 votes`

Yes, that makes sense to me

Issue #9719, opened 2023-06-07
Typing indicator is too close to the message input


ShGKme:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

The typing indicator is too close to the message input.

![image](https://github.com/nextcloud/spreed/assets/25978914/97d2e760-dea9-4cce-81e5-99a6e0df1138)

![image](https://github.com/nextcloud/spreed/assets/25978914/ed5f2d20-b44c-4e55-832e-1760b2a38c52)

What about positioning it in the middle of the `.wrapper`'s padding?

![image](https://github.com/nextcloud/spreed/assets/25978914/7ce8127d-a075-4c88-a5be-d1f29cd3c49a)

AniketSaraswat112:
HI @ShGKme I would like to work on this issue please assign this to me.

Antreesy:
Case "More than 3 persons are typing in the call" should be considered when working on that issue, because in the sidebar chat indicator wraps into two lines of text
Original issue: https://github.com/nextcloud/spreed/issues/9589
Fix to that issue: 18324284787803a431b9a65432b96289143f20ea

PR #9826, opened 2023-06-21
fix(styles): prevent typing indicator overlay

dijeth:
### ☑️ Resolves

* Fix [typing indicator is too close to the message input](https://github.com/nextcloud/spreed/issues/9719)
* Fix typing indicator overlays the message above

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks

- [ ] Behaviour check

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -1075,7 +1075,7 @@ export default {
 
 .scroll-to-bottom {
 	position: absolute !important;
-	bottom: 76px;
+	bottom: 140px;
 	right: 24px;
 	z-index: 2;
 }
--- a/src/components/NewMessage/NewMessage.vue
+++ b/src/components/NewMessage/NewMessage.vue
@@ -22,7 +22,7 @@
 -->
 
 <template>
-	<div class="wrapper" :class="{'wrapper--has-typing-indicator': showTypingStatus}">
+	<div class="wrapper">
 		<NewMessageTypingIndicator v-if="showTypingStatus"
 			:token="token" />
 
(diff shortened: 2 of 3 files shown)


src/components/NewMessage/NewMessage.vue
```diff
@@ -804,14 +804,8 @@ export default {
 
 .wrapper {
 	position: relative;
-	display: flex;
-	justify-content: center;
 	padding: 12px 0 12px;
```
Antreesy, line 807:
```suggestion
	padding: 12px 0;
```
This as well could be shortened

src/components/NewMessage/NewMessageTypingIndicator.vue
```diff
@@ -139,11 +139,10 @@ export default {
 
 <style lang="scss" scoped>
 .indicator {
-	position: absolute;
-	bottom: calc(100% - 2em);
-	left: 0;
 	width: 100%;
 	padding-right: 12px;
+	padding-left: 12px;
+	margin-bottom: 12px;
```
Antreesy, line 145:
```suggestion
	margin-bottom: 4px;
```

Smaller offset seems also good for, and as benefit, we save some pixels for the chat itself

Antreesy (commented):
Hello and thanks for contributing to us!
It's a nice work, but i have found a couple of places, which could be improved. Please see the comments below:

src/components/NewMessage/NewMessageTypingIndicator.vue
```diff
@@ -139,11 +139,10 @@ export default {
 
 <style lang="scss" scoped>
 .indicator {
-	position: absolute;
-	bottom: calc(100% - 2em);
-	left: 0;
 	width: 100%;
 	padding-right: 12px;
+	padding-left: 12px;
```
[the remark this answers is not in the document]
dijeth: Oh, yes... You are right!

dijeth:
Thank you for the comments!
I've done all of them and added a sticky position to `scroll-to-bottom` button and now its position depends on the height of "new message" block.

image

src/components/NewMessage/NewMessage.vue
```diff
@@ -803,71 +801,57 @@ export default {
...
+		position: relative;
+	}
 
-		&__quote {
-			margin: 0 16px 12px 24px;
-			background-color: var(--color-background-hover);
-			padding: 8px;
-			border-radius: var(--border-radius-large);
+	// Override NcRichContenteditable styles
+	&__richContenteditable {
```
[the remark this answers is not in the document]
dijeth: oh... fixed

dijeth:
If we hardcode the position of the button, it may happen that button will overlap the text.
Also, the position of the button seems strange when no one is typing. Perhaps we could set the `bottom` value depending on the typing status.

image

Wouldn't it be better if we hid the button using opacity? This way the scrolling will not jump and the button position will be adaptive. But we will need to remove the extra space under the text...

scroll-button

src/components/MessagesList/MessagesList.vue
```diff
@@ -1075,7 +1075,7 @@ export default {
 
 .scroll-to-bottom {
 	position: absolute !important;
-	bottom: 76px;
+	bottom: 140px;
```
Antreesy, line 1078: To make this property reliable on typing indicator, we need to pass it somehow through the parent component (which I think is an overkill)

Antreesy:
> But we will need to remove the extra space under the text...

That's the side effect of `position: sticky` I want to avoid

src/components/NewMessage/NewMessage.vue
```diff
@@ -803,71 +801,57 @@ export default {
...
-			padding: 8px 16px 8px 44px;
-			max-height: 180px;
-
-			&:hover,
-			&:focus,
-			&:active {
-				border: 1px solid var(--color-primary-element);
-			}
-		}
+	&__emoji-picker {
```
Antreesy, line 829: Sorry, this unchanged styles, moved to another position, distract me. Can we try to minimize amount of changes in this PR. It could affect us later when we backport something to older versions

Antreesy (commented):
Almost there! I have some thoughts on how to make scroll-to-bottom button work, but it requires a lot of code movements. How about we squash your commits into one and merge, and work on button in separate PR?

Antreesy (approved):
Position of "Scroll to bottom" goes to follow-up, as it needs more than a style refactoring.
The indicator itself is looking great now, thanks again!

PR #9830, opened 2023-06-21
feat(chat): Expose the "top most parent" ID as thread ID

### LINKED ISSUE #9679: API - Message threads
* Overview at https://github.com/nextcloud/spreed/issues/1469
* Rough idea is that when a reply message or a message that has a reply is clicked, we filter the chat stream with that "thread" ID (root parent message).
* All newly written messages should be addresses to that root parent as well (needs a new API parameter, so that you can still reply within a thread for quoting and reactions)
* The UI of course needs to see a button to "go back" to the unfiltered list.

nickvergessen:
### ☑️ Resolves

* Fix #9679 

### 🚧 Tasks

- [x] ~~Add threading system~~
    - [x] Expose top most parent id as thread id
- [x] Define capability
- [x] Adjust documentation

### 🏁 Checklist

### DIFF
--- a/docs/capabilities.md
+++ b/docs/capabilities.md
@@ -121,3 +121,6 @@
 * `config => call => supported-reactions` - A list of emojis supported as call reactions. If the list is absent or empty, clients should not show the emoji reaction option in calls.
 * `config => chat => typing-privacy` - User defined numeric value to enable 1 or disable 0 the typing indicator to other users
 * `typing-privacy` - Support toggle typing privacy
+
+## 18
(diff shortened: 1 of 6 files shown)


Antreesy (approved):
Able to receive from web client, message by default gets `threadId=0`, 1-st level replies get the same `threadId` as `parent` param, deep replies inherit `threadId` of the original first message

Note: at the moment thread messages would look messy in the original message list. Also in order to show threads, we need to receive all messages by threadId, including the original one (by id), and also all the original thread messages (which have thread started - maybe, additional param?)

PR #9811, opened 2023-06-18
Experiments to increase left sidebar performance

ShGKme:
See: https://github.com/nextcloud/spreed/issues/9630#issuecomment-1596290654

How to test:
1. `npm ci`
2. Add Many chats. 200 chats for slow machines, 500 for fast one
3. Open Firefox
4. In the console use:
- `PERF_TEST.TRY_USE_SMART_UPDATE()` - update only changed properties in the state
- `PERF_TEST.TRY_USE_VIRTUAL_SCROLLING()` - use virtual scrolling
- `PERF_TEST.TRY_USE_DUMMY_CONVERSATION()` - render conversations as `<div>`
- `PERF_TEST.TRY_USE_SIMPLE_CONVERSATION()` - render conversations without `<NcListItem>` as a `<RouterLink>` with `icon` and some data
- `PERF_TEST.TRY_HIDE_CONVERSATION_ICON()` - render conversations without `<ConversationIcon>`
- `PERF_TEST.TRY_HIDE_NC_ACTION_BUTTON()` - render conversations without `<NcActionButton>`
- `PERF_TEST.TRY_RESET()` - reset
5. `PERF_TEST.START()` - start the test (run at least 2-3 times)

PR #9832, opened 2023-06-21
Optimize conversations update on Frontend

### LINKED ISSUE #9630: 🔥 Sever performance/CPU issue with at least firefox due to Talk page rendering
### How to use GitHub

---

## Steps to reproduce
1. Have 200+ conversations
2. Have the tab open and others help sending messages so chats sort up and down regularly

### Expected behaviour
🚤 Everything works fast and properly

### Actual behaviour
🔥  PC heats up, CPU is 100% and Firefox eventually becomes unresponsive

## Talk app

**Talk app version:** 17rc2 at least

ShGKme:
### ☑️ Resolves

* Maybe fix #9630
* Decrease re-rendering time for `LeftSidebar` on re-fetching conversations every 30 seconds when:
  - External signaling is using
  - This is not full re-fetching every 5 minutes

### 🖼️ Screenshots

No visual changes

### 🚧 Tasks

- [x] Add new action `addConversationIfChange` to add new conversation only if it was moidified according to `lastActivity` and `modifiedSince`
- [x] Add new action `postAddConversation` for common post-action of adding a conversation
- [x] Use new `addConversationIfChange` in `fetchConversations` action
- [x] Fix old tests
- [x] Add new test for this case

### 🏁 Checklist

### Alternative and additional solutions
- In comparison, compare `lastActivity` and old `lastActivity`, not `modifiedSince`
- In addition, cache conversations list with KeepAlive to make returning from search fast
  - See: https://github.com/nextcloud/spreed/pull/9833
- Instead, decrease number of any state changes by full property comparison
  - Works with any signaling
  - Also works on 5-minute full re-fresh
  - Not domain-specific, a bit more code, but universal solution
  - See: https://github.com/nextcloud/spreed/pull/9811
- Instead/In addition use virtual scrolling
  - Increase `LeftSidebar` performance in general in any case, including memory usage, interactive reaction and etc.
  - Tricky in scrolling part, e2e testing and *perhaps* more risky
  - See: https://github.com/nextcloud/spreed/pull/9811

### DIFF
--- a/src/store/conversationsStore.js
+++ b/src/store/conversationsStore.js
@@ -203,7 +203,7 @@ const mutations = {
 
 const actions = {
 	/**
-	 * Add a conversation to the store and index the displayName.
+	 * Add a conversation to the store and index the displayName
 	 *
 	 * @param {object} context default store context;
 	 * @param {object} conversation the conversation;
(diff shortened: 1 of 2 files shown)


nickvergessen (approved):
🤞🏼

src/store/conversationsStore.js
```diff
@@ -203,14 +203,42 @@ const mutations = {
...
+	 *
+	 * @param {object} context dispatch context
+	 * @param {object} payload mutation payload
+	 * @param {object} payload.conversation the conversation
+	 * @param {number|0} payload.modifiedSince timestamp of last state or 0 if unknown
+	 */
+	addConversationIfChanged(context, { conversation, modifiedSince }) {
+		if (conversation.lastActivity >= modifiedSince) {
+			context.commit('addConversation', conversation)
+		}
```
Antreesy, line 228:
```suggestion
		} else {
		  // Smart update
		}
```

Don't we want to apply smartUpdate here to cover rest cases (full re-fetch at least)?

Antreesy (approved):
Looks good from the code perspective (but still lagging for me because of internal signalling)

src/store/conversationsStore.js
ShGKme: If we use the smartUpdate approach in else, then we do not need `if` branch, as `smartUpdate` covers all possible updates anyway

ShGKme:
> but still lagging for me because of internal signalling

To test on internal you can add:
```js
this.roomListModifiedBefore = ~~(Date.now() / 1000)
```

after this `if`

https://github.com/nextcloud/spreed/blob/d1adff82227f0cb3fc7099fb3196d2d10941fcb5/src/components/LeftSidebar/LeftSidebar.vue#L493-L499

PR #9847, opened 2023-06-22
Add local feature flags for optimization testing on production

ShGKme:
### ☑️ Resolves

* Helps with #9630

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 Tasks

- [x] Add `localFeatureFlagService`
  - [x] Adds `OCA.Talk.FEATURE_FLAGS` interface with methods to set and reset flags
  - [x] Exposes flags as Vue plugin in Vue prototype
- [x] Add soft conversations update feature flag
  - Updates conversations without recreating objects when possible
- [x] Add hidden avatars feature flag
  - Displays gray circle instead of the conversation avatar
- [x] Add revert user status sync feature flag
  - [x] Reverts https://github.com/nextcloud/spreed/pull/9411
- [x] Add Virtual Scrolling flag
  - [x] Refactor LeftSidebar (required here)
    - [x] Instead of mutating getter result as computed side effect, move sorting conversations into `conversationsList` getter
    - [x] Divide storing all conversations and filtered (search) conversations
    - [x] Organize rendering in `LeftSIdebar`, remove duplications in conditions mix
  - [x] Add the virtual scrolling
  - [x] Fix scrolling
- [x] Fix tests

### 🏁 Checklist


ShGKme:
Should be properly tested. Or drop some changes.

See by commits.

nickvergessen:
```
[Vue warn]: Duplicate keys detected: '9d1d6399-0e8e-4206-b89e-81eae3689a0f'. This may cause an update error.

found in

---> <LeftSidebar> at src/components/LeftSidebar/LeftSidebar.vue
       <NcContent>
         <App> at src/App.vue
           <Root> [vue.runtime.esm.js:4605](webpack:///talk/node_modules/vue/dist/vue.runtime.esm.js)
    VueJS 27
    _callee$ LeftSidebar.vue:195
    tryCatch LeftSidebar.vue:2
    makeInvokeMethod LeftSidebar.vue:2
    defineIteratorMethods LeftSidebar.vue:2
    asyncGeneratorStep LeftSidebar.vue:3
    _next LeftSidebar.vue:4
    (Async: promise callback)
    asyncGeneratorStep LeftSidebar.vue:3
    _next LeftSidebar.vue:4
    _asyncToGenerator LeftSidebar.vue:4
    _asyncToGenerator LeftSidebar.vue:4
    fetchPossibleConversations LeftSidebar.vue:223
    _callee3$ LeftSidebar.vue:274
    tryCatch LeftSidebar.vue:2
    makeInvokeMethod LeftSidebar.vue:2
    defineIteratorMethods LeftSidebar.vue:2
    asyncGeneratorStep LeftSidebar.vue:3
    _next LeftSidebar.vue:4
    _asyncToGenerator LeftSidebar.vue:4
    _asyncToGenerator LeftSidebar.vue:4
    fetchSearchResults LeftSidebar.vue:280
    debounceFetchSearchResults LeftSidebar.vue:170
    later index.js:27
    (Async: setTimeout handler)
    debounced index.js:38
    VueJS 4
    localValue SearchBox.vue:48
    VueJS 8
```

I guess this is from spliting the search and the normal conversation list?

src/store/conversationsStore.js
```diff
@@ -122,6 +123,71 @@ const mutations = {
...
+		// Update conversations
+		for (const [token, oldConversation] of Object.entries(state.conversations)) {
+			const newConversation = newConversations[token]
+			// If a conversation has activity - update conversation properties as well
+			if (oldConversation.lastActivity !== newConversation.lastActivity) {
+				applyObjectChanges(oldConversation, newConversation, true)
+				// If the conversation's last message was changed - update the last message
+				if (oldConversation.lastMessage && newConversation.lastMessage && oldConversation.lastMessage.id !== newConversation.lastMessage.id) {
+					applyObjectChanges(oldConversation.lastMessage, newConversation.lastMessage, true)
+				}
```
Antreesy, line 200: It's the only `Object type` property for now, but to be more secure, could we make a recursive update here?
Or inside `applyObjectChanges()`?

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -39,76 +39,87 @@
...
+							:ref="`conversation-${item.token}`"
+							:item="item" />
 					</template>
-					<Hint v-else-if="searchText && !conversationsList.length"
-						:hint="t('spreed', 'No matches')" />
-					<template v-if="isSearching">
-						<template v-if="!listedConversationsLoading && searchResultsListedConversations.length > 0">
+
+					<!-- Search results -->
+					<template v-else-if="isSearching">
```
Antreesy, line 71:
```suggestion
					<template v-else>
```
Will it make sense here?

Antreesy (commented):
Not sure for now, what could be the reason, but when switching conversations, all `<Conversation />` components are also updated (can tell it from triggering `updated` hook)

src/components/LeftSidebar/LeftSidebar.vue
ShGKme: With large `<template>` blocks I wanted it to be more explicitly visible that this block is searching (not only by manual comments but also by the code).

But yep, it can be safely replaced with `v-else`

ShGKme:
> [Vue warn]: Duplicate keys detected: '9d1d6399-0e8e-4206-b89e-81eae3689a0f'. This may cause an update error.

@nickvergessen I cannot reproduce. Could you give addition info:
- Is it from the last commit?
- What flags have you enabled?
- Is it the list or the searching state? If searching, what it looked like?
- What chats do you have?

nickvergessen:
I enabled all flags, searched, removed switched conversations, created 1-1, etc. No detailed things I followed, just saw it afterwards in the console

ShGKme:
> Not sure for now, what could be the reason, but when switching conversations, all <Conversation /> components are also updated (can tell it from triggering updated hook) - not from this PR, same on master

Nice catch.

Yes, it updates. It has a `computed` that depends on current token. Switching conversation triggers this computed.

https://github.com/nextcloud/spreed/blob/7f5e7d5cc8209dfb571ed83cd593576c48efe754/src/components/LeftSidebar/ConversationsList/Conversation.vue#L330-L337

To improve it, we can make `isActive` a prop, then the component should not update if `isActive` value is not actually toggled.

@nickvergessen do we want to do it in the same pr? As a flag?
It is a small change.

nickvergessen:
> do we want to do it in the same pr? As a flag?

You are the expert, does it make sense?

Antreesy:
> do we want to do it in the same pr? As a flag?

I'd say, we should include it as part of `LeftSidebar` refactoring, independently from flags, as it's crucial in our case (same as 5-min problem)

ShGKme:
> To improve it, we can make isActive a prop, then the component should not update if isActive value is not actually toggled.

UPDATE: this computed is totally useless now, the current `NcListItem` implementation doesn't use it at all. We can just remove it.

ShGKme:
> You are the expert, does it make sense?

Well, the question was more about adding this change as a part of this PR with a flag or separately.

Anyway, it can be removed at all (done).

> I enabled all flags, searched, removed switched conversations, created 1-1, etc. No detailed things I followed, just saw it afterwards in the console

I still cannot reproduce :( 
Do you have user groups?

@Antreesy Have you seen such an error?

My only clue is that some conversations may have the same token, because I use `token` instead of an `id` for `key` in virtual scrolling.

src/store/conversationsStore.js
ShGKme: > but to be more secure, could we make a recursive update here?

Speaking about security, after your message I realized that there is a bug. I will fix it soon (now). Objects and arrays inside `lastMessage` were not updated.

---

> but to be more secure, could we make a recursive update here?

Well, on one hand, yes, on another no.

With the recursive update, it's a bit more complex to debug/control what is updating and how except setting the depth of blockLists. To see that "a conversation is updated" or "a lastMessage is updated".

But yep, with just a full recursive update we can make it universal and safe.

Both implementations are ready. What would you prefer?

Antreesy: > we can make it universal and safe.

I'd vote for this one.

Just a thought, `lastMessage` property 'message' is one and only what we use in the component's render. Apart from that prop others are not used, and doubtly will change (only a reactions to last message, but that's not the reason to re-render a component)
So we could maybe rewrite a `lastMessage` object (and other 2-nd level object like it) completely to make it simple?

ShGKme:
@Antreesy 
Soft update is updated to use a simple recursion without any domain-specific logic

src/store/conversationsStore.js
```diff
@@ -122,6 +137,55 @@ const mutations = {
...
+				if (!Object.hasOwn(newObject, key)) {
+					del(target, key)
+				}
+			}
+
+			// Add new properties and update old ones
+			for (const key of Object.keys(newObject)) {
+				if (!Object.hasOwn(target, key)) {
+					// Add new property
+					set(target, key, newObject[key])
```
Antreesy, line 168:
```suggestion
					Vue.set(target, key, newObject[key])
```
maybe a matter of preference, but in order to keep consistent I'd write with like all the other mutations (or we could rewrite all the others later)

Antreesy (approved):
Changes seems fine. Observed no errors while testing, just a bit lagging (0.5s) when receiving response from server (slow machine + internal server + 200 chats)

src/store/conversationsStore.js
ShGKme: Vue.set is deprecated

Antreesy: Then let's raise a follow up later

ShGKme: Hm, Compat built has only the old `Vue.set` and not the new `set` from `2.7`.
Then it is the opposite for easy migration with the Compat
https://v3-migration.vuejs.org/migration-build.html#feature-reference

ShGKme:
`set` and `del` replaced with `Vue.set` and `Vue.delete`.

Squashed fixup commits.

Tested each separated flag in the production build, including the search, renames, new chats, removed chats, and new messages.

I still couldn't reproduce `Duplicate keys detected`.

ShGKme:
As 17.0.1 has already been released and deployed, I will check also `Unread mentions` button later.

nickvergessen:
Can you rebase this on top of the tag `v17.0.1` (commit https://github.com/nextcloud/spreed/commit/d9464862f5e47b28856fac029a12615fde6e9db5) so we can build production JS assets and swap them on our servers for testing?

nickvergessen:
Or you make a new branch based on top of the commit.

Antreesy (approved):
| Test                                          | Initial render            | Re-fetch and re-render | Search input type + clear | Notes                           |
|-----------------------------------------------|---------------------------|------------------------|---------------------------|---------------------------------|
| DEFAULT                                       | Slow, up to 5 sec         | Slow, up to 5 sec      | Slow, up to 5 sec         |                                 |
| CONVERSATIONS_LIST__SOFT_CONVERSATIONS_UPDATE | --                        | No freezes             | --                        | IMO good for backport by itself |
| CONVERSATIONS_LIST__VIRTUAL_SCROLLING         | Bit faster, up to 2-3 sec | Fast, up to 0.5-1 sec  | Fast, up to 0.5-1 sec     | Scroll behaviour is broken      |

Other flags doesn't show significant results for me (internal signaling, firefox, 300 group chats + 10 one-to-one
Also would like to propose to cherry-pick some commits (41f7a0cd14a25028902ca878aefd650aa8dd70c0, 6a03653151cc5e58910f034c92974c80c58c0939, 24f464bb15ba20b4d581298f0b6be5d2be9ae6f4) and merge it to master + backport already, as they shouldn't break any behaviour for now, but only improve it

ShGKme:
> I will check also Unread mentions button later.

A bug is not related to this PR

ShGKme:
> Or you make a new branch based on top of the commit.

https://github.com/nextcloud/spreed/pull/9865

ShGKme:
@Antreesy The results look a bit weird. Have you tested on production or development build? In case of the development build, have you disabled strict mode in Vuex?

Antreesy:
> Have you tested on production or development build?

My laptop can't handle production build anymore =(
But for `strict: true` and `strict: false` I got more or less the same freeze durations

ShGKme:
> But for strict: true and strict: false I got more or less the same freeze durations

It has a huge effect on handling updates. With strict mode, it will be slower even with virtual scrolling.

`CONVERSATIONS_LIST__VIRTUAL_SCROLLING` shouldn't be so slow on "Re-fetch and re-render"... I'd expect the result to be caused by the `strict`.

nickvergessen:
Using the first flag spams errors on converstion list update:
Bildschirmfoto vom 2023-06-28 06-03-40

nickvergessen:
- [ ] With the virutal scrolling flag, the user status seems incorrectly being recycled sometimes.

PR #9861, opened 2023-06-26
follow-up(ChatView) - align scroll button position with mobile clients

Antreesy:
### ☑️ Resolves

* Follow-up from #9826 (cc @dijeth )
* Scroll button was moved to the parent component level
* Its position updated based on NewMessage component's height (which is updated if someone is typing)

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot from 2023-06-26 11-50-23 | image
Screenshot from 2023-06-26 11-50-29 | image

### 🚧 Tasks

- [ ] Visual check / Testing
- [ ] Code review

### 🏁 Checklist

### DIFF
--- a/src/components/ChatView.vue
+++ b/src/components/ChatView.vue
@@ -58,7 +58,7 @@
 				:style="`bottom: ${scrollButtonOffset}px`"
 				@click="smoothScrollToBottom">
 				<template #icon>
-					<ChevronDown :size="20" />
+					<ChevronDoubleDown :size="20" />
 				</template>
 			</NcButton>
 		</transition>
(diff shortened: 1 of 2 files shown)


marcoambrosini (commented):
I think the button should be within the boundaries of the message list, with 8px bottom margin to the new message form component

Antreesy:
> I think the button should be within the boundaries of the message list, with 8px bottom margin to the new message form component 

You mean to lift it up on 8px, not lower? For now I restored it's initial position, so it will be a design change.
And as you came up with this suggestion, maybe we also should replace icon with double chevron, like on Android? https://pictogrammers.com/library/mdi/icon/chevron-double-down/

marcoambrosini:
Icon change sounds good. I mean exactly like the screenshot you posted, ~ 8px above the `newmessageform` border.

marcoambrosini (approved):

Issue #9448, opened 2023-05-04
Vue 3 Migration


ShGKme:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

# ⚰️ Vue 2 End-Of-Life - 2024 

**Vue 2 will reach End of Life (EOL) on December 31st, 2023:** https://v2.vuejs.org/lts/

See also: https://v3-migration.vuejs.org/

## Dependency compatibilities

Nextcloud Dependencies:

- [x] `@nextcloud/vue`: https://github.com/nextcloud/nextcloud-vue/issues/2154
- [x] `nextcloud-vue-collections`: https://github.com/nextcloud-libraries/nextcloud-vue-collections/tree/vue3
- [x] `@nextcloud/dialogs`: https://github.com/nextcloud-libraries/nextcloud-dialogs/pull/884
- [x] `@nextcloud/upload`: https://github.com/nextcloud-libraries/nextcloud-upload/pull/1197

Other dependencies:

- [x] vue@2 -> vue@3
- [x] vue-router@3 -> vue-router@4
- [x] vuex@3 -> vuex@4
- [x] vuex@4
- [x] vue-cropperjs@4 -> vue-cropperjs@5
- [x] vue-draggable-resizable@2 -> vue-draggable-resizable@3
- [x] vue-observe-visibility@1 -> vue-observe-visibility@2
- [x] `vue-outside-events` - migrate to `vueuse` #12346
- [x] `vue-prevent-unload` - get rid of or migrate to `vueuse` if any #12349
- [x] `vue-shortkey` - check `vue3-shortkey` 
- [x] `vue-virtual-scroller@1` -> `vue-virtual-scroller@2`
- [x] `vue2-leaflet` -> `@vue-leaflet/vue-leaflet`

### 📄 Preparations

Details are to be added to the separated issues.

- [x] Migrate all deprecated in 2.7 things
  - [x] https://github.com/nextcloud/spreed/issues/9449
  - [x] https://github.com/nextcloud/spreed/pull/9410
- [x] Prepare for the migration (with ESLint warnings whenever possible **// todo: create an issue**)
[...]

PR #9892, opened 2023-07-03
feat(store) replace Vuex `settingsStore` with equivalent Pinia store

Antreesy:
### ☑️ Resolves

* Part of #9448 
* This PR as a beginner should be a proper example of how to migrate other stores - feel free to suggest changes!

### 🖼️ Screenshots

* No visual changes
* Only NewMessage Typing indicator and Talk settings private statuses switches are affected

### 🚧 Tasks

- [ ] Manual testing
- [ ] Code review

### 🏁 Checklist


src/stores/settingsStore.js
```diff
@@ -0,0 +1,59 @@
...
+import { defineStore } from 'pinia'
+
+import { getCapabilities } from '@nextcloud/capabilities'
+import { loadState } from '@nextcloud/initial-state'
+
+import { PRIVACY } from '../constants.js'
+import { setReadStatusPrivacy, setTypingStatusPrivacy } from '../services/settingsService.js'
+
+export const useSettingsStore = defineStore('settingsStore', {
+	state: () => ({
```
ShGKme, line 32: **nitpick:** store key format

I'm not sure we need should add `*Store` prefix to all stores' keys. This key is used only internally in Pinia and in the store devtools. So with this prefix, we will have `*Store` for every item in the list of stores in the store dev tools.

This is also not how stores are named in the [docs](https://pinia.vuejs.org/core-concepts/#defining-a-store), and not how pinia stores are called in other places there Pinia is used in nextcloud, for example, in files:
https://github.com/nextcloud/server/blob/686ad69133bb9359f9386080aface284c8639398/apps/files/src/store/viewConfig.ts#L33-L37

ShGKme: PS I mean only the key passed to the `defineStore`, not the composable variable name.

ShGKme (changes requested):
It seems that we also need to init Pinia at least in files sidebar app. And then we will need to init pinia in other apps used store.

All the other things looks great to me.

---

While I believe, moving to Pinia has only benefits and no drawbacks, I'm not sure we should do it now (as was discussed with @nickvergessen on the Talk Team week).

This is not a part of Vue 3 migration as both Pinia and Vuex are not linked to a specific Vue version. We can do it before the migration as well as after the migration.

Nevertheless, I'm personally fine to move to Pinia any time.

src/stores/settings.js
```diff
@@ -0,0 +1,59 @@
...
+
+import { PRIVACY } from '../constants.js'
+import { setReadStatusPrivacy, setTypingStatusPrivacy } from '../services/settingsService.js'
+
+export const useSettingsStore = defineStore('settings', {
+	state: () => ({
+		readStatusPrivacy: loadState('spreed', 'read_status_privacy', PRIVACY.PRIVATE),
+		typingStatusPrivacy: loadState('spreed', 'typing_privacy', PRIVACY.PRIVATE),
+		supportTypingStatus: getCapabilities()?.spreed?.config?.chat?.['typing-privacy'] !== undefined,
+	}),
```
ShGKme, line 36: I didn't notice it at first.

`supportTypingStatus` is not an application state, it never changes, and we do not support capabilities updates without reloading.

I prefer not to store constants in a global reactive store.

Antreesy: I wanted to ease access to constants from capabilities.
Can we do it another way, maybe? get them from server once, store as simple JS constant in, say, `src/capabilities.js`, and import from there?

ShGKme: > Can we do it another way, maybe? get them from server once, store as simple JS constant in, say, `src/capabilities.js`, and import from there?

We can. But it seems, it won't change much.

If this module looks like

```js
export const talkCapabilities = {
  config: {
    chat: {
      typingPrivacy: getCapabilities()?.spreed?.config?.chat?.['typing-privacy'] !== undefined
    }
  }
}
```

then it will help to simplify one complex line

```js
getCapabilities()?.spreed?.config?.chat?.['typing-privacy'] !== undefined
```

to two simple

```js
import { talkCapabilities } from '...'

talkCapabilities.config.chat.typingPrivacy
```

ShGKme: I personally would prefer to see "more original" capability path instead of new names.

```js
// Most explicit usage of capabilities
// But many optional chaining, long due to accessing spreed from all capabilities and casting to boolean by expression
getCapabilities()?.spreed?.config?.chat?.['typing-privacy'] !== undefined

// Close to capabilities
talkCapabilities.config.chat.typingPrivacy

// New name, required to jump into declaration
supportTypingPrivacy
```

We can discuss adding capabilities module. But speaking about this PR, I'm against the solution of mixing individual constant capabilities with global reactive state storages.

Antreesy: Then I'll revert these changes for current PR for now

nickvergessen: Also it's not "support" typing privacy, but whether it's on/off for the current user?

nickvergessen: right, we have `spreed.feature.typing-privacy` as well as `spreed.config.chat.typing-privacy`. but yeah the config one is the user setting.

I'd say simplicity is key and we can keep it directly in the code, until we faced too many typos of one and then we do it with a method :D

Antreesy:
Rebased onto master because of conflicting package-lock.json

ShGKme (approved):
🍍 Tested, works fine 🍍

Note, OptionsApi-like define store was used here. It is easier for migration purpose and looks more structured. But CompositionApi-like defineStore have more possibilities, such as:
- More complex init
- Internal (aka private) methods and variables
- Possibility to use `watch` and etc.

PR #9909, opened 2023-07-06
Use PHP8's match expression instead of switch

shdehnavi:
### ☑️ Resolves

* Using PHP8's `match` expression instead of `switch`

### 🏁 Checklist

### DIFF
--- a/lib/Activity/Provider/Base.php
+++ b/lib/Activity/Provider/Base.php
@@ -106,19 +106,11 @@ abstract class Base implements IProvider {
 	}
 
 	protected function getRoom(Room $room, string $userId): array {
-		switch ($room->getType()) {
-			case Room::TYPE_ONE_TO_ONE:
-			case Room::TYPE_ONE_TO_ONE_FORMER:
-				$stringType = 'one2one';
-				break;
-			case Room::TYPE_GROUP:
-				$stringType = 'group';
-				break;
-			case Room::TYPE_PUBLIC:
-			default:
-				$stringType = 'public';
-				break;
-		}
+		$stringType = match ($room->getType()) {
+			Room::TYPE_ONE_TO_ONE, Room::TYPE_ONE_TO_ONE_FORMER => 'one2one',
+			Room::TYPE_GROUP => 'group',
+			default => 'public',
+		};
 
 		return [
 			'type' => 'call',
--- a/lib/Chat/Parser/UserMention.php
+++ b/lib/Chat/Parser/UserMention.php
@@ -202,16 +202,11 @@ class UserMention {
 	 * @throws \InvalidArgumentException
 	 */
 	protected function getRoomType(Room $room): string {
-		switch ($room->getType()) {
-			case Room::TYPE_ONE_TO_ONE:
-			case Room::TYPE_ONE_TO_ONE_FORMER:
-				return 'one2one';
-			case Room::TYPE_GROUP:
-				return 'group';
-			case Room::TYPE_PUBLIC:
-				return 'public';
-			default:
-				throw new \InvalidArgumentException('Unknown room type');
-		}
+		return match ($room->getType()) {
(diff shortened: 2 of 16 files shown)


nickvergessen:
For now I'll pass on this one.
I don't think it helps with readability of the code at the moment.

nickvergessen:
Thanks anyway for all the efforts

shdehnavi:
@nickvergessen Thanks for the feedback, I still suggest using `match` expression instead of the traditional `switch` statement in some cases. The `match` expression provides some advantages over the traditional `switch` statement. Here are a few reasons why we might consider using `match` instead:

**Simplicity and readability:** The `match` expression offers a more concise and expressive syntax compared to the `switch` statement. It allows you to write code in a more straightforward and readable manner, especially when you're dealing with simple value matching.

**Strict type checking:** The `match` expression performs strict type checking by default, which means it takes into account the type of the value being matched against. This helps to prevent unexpected behavior and reduces the chances of introducing bugs related to type coercion.

**Exhaustiveness checking:** When you use the `match` expression, PHP can automatically check if you have handled all possible cases. If you miss any cases, PHP will issue a warning or error, depending on your error reporting settings. This feature helps catch potential logic errors and makes your code more robust.

**Return values:** The `match` expression allows you to use the result of the matching directly as a value. It can be assigned to a variable or used in an expression, eliminating the need for explicit break statements like in the `switch` statement.

**Performance improvements:** In some cases, the `match` expression can provide better performance than the `switch` statement. This is because the `match` expression is implemented using a more efficient internal mechanism, leveraging the advantages of PHP 8.0's JIT compilation.

nickvergessen:
Might be me, but I'm still not convinced. Let's take this as example:

grafik

> Simplicity and readability

- For me the readability is reduced, because I have to scan several starting points `TYPE_ONE_TO_ONE_FORMER` is to be found way later in the line compared to `TYPE_ONE_TO_ONE` and `TYPE_GROUP`
- TYPE_PUBLIC is removed and therefore does not give a search hit anymore when introducing a similar type in the future

> Strict type checking

- Since we work with models basically all the time and they have typed values, there is implicit strict types already

> Exhaustiveness checking

- If that is the case, how come that `TYPE_PUBLIC` can be removed without an issue? :)

> Return values

- I give you that one 👍🏼 

> Performance improvements

The given switch statements are all so "simple" that the performance gain should be rather small. There are other things which would make more sense to look into.

---

I think before switching to `match` I would prefer to move to if (when returning directly) or if/elseif (when onlly setting a var)

shdehnavi:
> Might be me, but I'm still not convinced. Let's take this as example:
> 
> grafik
> 
> > Simplicity and readability
> 
> * For me the readability is reduced, because I have to scan several starting points `TYPE_ONE_TO_ONE_FORMER` is to be found way later in the line compared to `TYPE_ONE_TO_ONE` and `TYPE_GROUP`
> * TYPE_PUBLIC is removed and therefore does not give a search hit anymore when introducing a similar type in the future
> 
> > Strict type checking
> 
> * Since we work with models basically all the time and they have typed values, there is implicit strict types already
> 
> > Exhaustiveness checking
> 
> * If that is the case, how come that `TYPE_PUBLIC` can be removed without an issue? :)
> 
> > Return values
> 
> * I give you that one 👍🏼
> 
> > Performance improvements
> 
> The given switch statements are all so "simple" that the performance gain should be rather small. There are other things which would make more sense to look into.
> 
> I think before switching to `match` I would prefer to move to if (when returning directly) or if/elseif (when onlly setting a var)

you are right in the above code, I merged match arms and it reduces the readability, So we can use separated arms to make it more readable. However, i think you prefer to not use match expressions, is that true?

nickvergessen:
> However, i think you prefer to not use match expressions based on the rest reasons you mentioned, is it true?

For now, yes

PR #9910, opened 2023-07-07
fix(NewGroupConversation) - refactoring & focus on Confirmation page 

### LINKED ISSUE #7584: Final step of conversation creation doesn't focus
### How to use GitHub

---

## Steps to reproduce
1. `+`
2. Enter a name
3. Tick "Allow guests"
4. Use keyboard tab to navigate to the next page
5. Tab through the user list and navigate to the next page

### Expected behaviour
The dialog is focused at the top

### Actual behaviour
Finish button is still focused (as it's the same button we just change the label) and pressing tab jumps to the message input in the background

DorraJaouad:
### ☑️ Resolves

* Fix #7584

The focus for every page is the following 
- First Page: conversation Name text field
- Second Page: search input
- Third Page: Copy Link button

### 🚧 Tasks

- [ ] Code review

### 🏁 Checklist


Antreesy (changes requested):
Actually, `<Confirmation />` component is a pile of props and boilerplate, without complex logic.
If we get rid of it and move the code to the `<NewGroupConversation />`, it will be just 30 lines of template + 3 lines of script + 15 lines of styles.

And that could allow us to access buttons directly, and watch after `success` and `error` - if they are truthy, we'll focus an appropriate button

src/components/LeftSidebar/NewGroupConversation/Confirmation/Confirmation.vue
```diff
@@ -87,6 +88,12 @@ export default {
 		},
 	},
 
+	mounted() {
+		const interval = setTimeout(() => {
+			this.$refs.copyLink.$el.focus()
+		}, 500);
+	},
```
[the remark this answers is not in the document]
Antreesy: And don't forget to lint files, you have a semicolon

src/components/LeftSidebar/NewGroupConversation/NewGroupConversation.vue
```diff
@@ -94,14 +94,35 @@
...
 				<!-- Third page -->
-				<Confirmation v-else-if="page === 2"
-					class="new-group-conversation__content"
-					:token="newConversation.token"
-					:conversation-name="conversationNameTrimmed"
-					:error="error"
-					:is-loading="isLoading"
-					:success="success"
-					:is-public="isPublic" />
+				<div v-else-if="page === 2" class="confirmation new-group-conversation__content">
```
Antreesy, line 97:
```suggestion
				<div v-else-if="page === 2" class="new-group-conversation__content">
```

We could omit that class, as it doesn't have any styles applied

src/components/LeftSidebar/NewGroupConversation/NewGroupConversation.vue
```diff
@@ -265,6 +285,15 @@ export default {
 				this.passwordProtect = false
 			}
 		},
+
+		success(value) {
```
Antreesy, line 289: Let's also add the same, but watch for `error` and focus `Close` button instead, as it'll be only available

Antreesy (approved):
Tested on web, and works so far!
Let's add one more button in focus, and squash commits into one (or two)

nickvergessen:
27 Backport is in #9954 
26 is canceled, so removing the label

PR #9958, opened 2023-07-12
chore(deps): Bump @nextcloud/vue to v7.12.1

DorraJaouad:
Tested for: 
 * Input Fields and RichContenteditable: add focus and select methods.
 
It works.


Antreesy (approved):
Tested on web client (Chrome + Linux)

PR #9968, opened 2023-07-13
Run psalm with all threads

provokateurin:
### ☑️ Resolves

Speeds up running psalm locally

### 🏁 Checklist

### DIFF
--- a/.github/workflows/psalm.yml
+++ b/.github/workflows/psalm.yml
@@ -39,4 +39,4 @@ jobs:
         run: composer i
 
       - name: Run coding standards check
-        run: composer run psalm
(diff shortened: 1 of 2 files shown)


.github/workflows/psalm.yml
```diff
@@ -39,4 +39,4 @@ jobs:
         run: composer i
 
       - name: Run coding standards check
-        run: composer run psalm
+        run: composer run psalm:ci
```
nickvergessen, line 42: Will be overwritten by all workflow template updates.

nickvergessen (changes requested):

PR #9977, opened 2023-07-14
fix(status): Don't emit event when the user status did not update

nickvergessen:
### ☑️ Resolves

* Attempt to contribute to #9630 
* If you think it makes sense, we can also apply it here: https://github.com/nextcloud/spreed/blob/04c973ed6aee1d8113ccf44d91b5431b954d6c70/src/mixins/getParticipants.js#L145-L153

### 🏁 Checklist

### DIFF
--- a/src/store/conversationsStore.js
+++ b/src/store/conversationsStore.js
@@ -240,13 +240,21 @@ const actions = {
 	 */
 	postAddConversation(context, conversation) {
 		if (conversation.type === CONVERSATION.TYPE.ONE_TO_ONE && conversation.status) {
-			emit('user_status:status.updated', {
-				status: conversation.status,
-				message: conversation.statusMessage,
-				icon: conversation.statusIcon,
-				clearAt: conversation.statusClearAt,
-				userId: conversation.name,
-			})
+			const prevStatus = context.state.conversations[conversation.token] || {}
+
+			// Only emit the event when something actually changed
+			if (prevStatus?.status !== conversation.status
+				|| prevStatus?.statusMessage !== conversation.statusMessage
+				|| prevStatus?.statusIcon !== conversation.statusIcon
+				|| prevStatus?.statusClearAt !== conversation.statusClearAt) {
+				emit('user_status:status.updated', {
+					status: conversation.status,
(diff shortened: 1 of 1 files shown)


ShGKme (approved):
Although reverting emitting `user_status:status.updated` didn't help in the experiment, it make sense.

ShGKme:
> If you think it makes sense, we can also apply it here

Yep, let's add it here too

src/store/conversationsStore.js
```diff
@@ -240,13 +240,21 @@ const actions = {
...
-				userId: conversation.name,
-			})
+			const prevStatus = context.state.conversations[conversation.token] || {}
+
+			// Only emit the event when something actually changed
+			if (prevStatus?.status !== conversation.status
+				|| prevStatus?.statusMessage !== conversation.statusMessage
+				|| prevStatus?.statusIcon !== conversation.statusIcon
+				|| prevStatus?.statusClearAt !== conversation.statusClearAt) {
+				emit('user_status:status.updated', {
```
Antreesy, line 250: Note:
* LeftSidebar gets status data from `this.$store.getters.conversationsList`
* `user_status:status.updated` change participant's status only in RightSidebar for current room, if the participant is this room's attendee

Do you want to soft update participants list also, or only statuses? because in the mixin we purge participants list completely, and add it back from response, then update all participants with the same statuses, which we've just added.

nickvergessen: > `user_status:status.updated` change participant's status only in RightSidebar for current room, if the participant is this room's attendee

That's not entirely true. It's a global event from the vue library and will affect very place that renders a user status and listens to it, including all avatars (also your own in the top right).

Antreesy: Indeed, code in `apps/user_status` is subscribed to it, my bad.

Then a check for changes should be valid and helpful in both places, but there's still a pending question with syncing statuses. We could try to solve it as a follow-up

nickvergessen:
Participants list is purged atm, so can't really check the previous status without much more efforts.
So let's go with this change here for now.

PR #9979, opened 2023-07-14
[stable27] chore(deps) - remove unused dependencies

Antreesy:
Manual backport of #9975 
Tested on web client, main + files sidebar (Chrome, Linux)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,7 +24,6 @@
 				"@nextcloud/router": "^2.1.2",
 				"@nextcloud/vue": "^7.11.6",
 				"attachmediastream": "^2.1.0",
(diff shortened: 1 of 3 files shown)


ShGKme:
We usually do not backport dependency updates when not necessary (cc @nickvergessen)

Antreesy:
> We usually do not backport dependency updates when not necessary

I think it's fine to make a cleanup, as we're going to support v.17 till the next release

nickvergessen:
Fine by me if you can garantee that it does not break anything under any circumstances for any user on any browser and OS.

Otherwise I'm all for being defensive and not changing deps unless there was a bug or sec fix we need

PR #9980, opened 2023-07-14
fix(LeftSidebar) redo arrow navigation

### LINKED ISSUE #7530: When creating a conversation it's not possible to select participants with the keyboard

### LINKED ISSUE #9944: Pressing enter after search no longer joins the first conversation in the search results
## Steps to reproduce
1. search in conversation 
2. press enter

### Expected behaviour
The first conversation in the search results should be joined 

### Actual behaviour
The first conversation in the unfiltered list (before searching) is joined

### LINKED ISSUE #9952: Searching a user name in the conversation creation dialog does not allow to use "Enter" to take the first hit
### How to use GitHub

---

Following the pattern of #9944  and basically like the assignee/review list on github works

Antreesy:
### ☑️ Resolves

* Fix #9952
* Fix #9944
* Fix #7530
* Behaviour was updated, should be checked and discussed before merge

#### :derelict_house:  Before
* Due to vue-lib update a long time ago, an old mixin based on lib classes didn't work at least since `stable25`. That makes it hard to reproduce, so behaviour was restored according to mixin inner functions description

#### 🏡 After
* New composable copies an old approach to implement arrow navigation, although it should be now reusable for other components
  * `Enter` - :warning: Instead of heading directly to the new conversation after `Enter` keydown, we are focusing the first element in the list, explicitly saying, that it's selected. To open conversation, `Enter` should be pressed once again
    * _updated_: it's possible to select item with one keydown (without confirmation) - default composable behaviour 
  * `ArrowUp` / `ArrowDown` - If we haven't a focused element yet, focusing a first one, to show user that we're in the navigation mode
  * `Escape` - return the focus back to the default element (if someting was focused). To perform default action, `Escape` should be pressed once again

https://github.com/nextcloud/spreed/assets/93392545/2c60ec1b-6825-4871-afd1-8f12817183bd

### 🚧 Tasks

- [ ] Check and approve new behaviour
- [ ] Follow-up: it's possible to join the first element simultaneously (but there are misalignments with #9952):
    * it'll be always a first element, whether it's an open conv, button or an already joined conversation
    * or joined already conversations will be omitted from this navigation, which IMO breaks UX 
    * We can't highlight first element natively, as it blurs the input and prevent typing
    * Doing it only visually requires JS manipulations with DOM, which we want to avoid
- [ ] Code review
- [ ] Manual testing

### 🏁 Checklist


src/composables/useArrowNavigation.js
```diff
@@ -0,0 +1,153 @@
...
+
+	/**
+	 * Focus the first element if not focused yet, otherwise proceed
+	 *
+	 * @param {Event} [event] Keydown event to prevent, if defined and not focused yet
+	 *
+	 * @return {boolean} If first element was focused or not
+	 */
+	function focusFirstElementIfNotFocused(event) {
+		if (focusedIndex.value === null) {
```
DorraJaouad, line 104: `if (focusedIndex.value !== 0)`

Because in case the user is focusing on `y`th list element , clicks back to the searchBox and hits enter again, the `focusedIndex.value === y `and nothing will happen

src/components/LeftSidebar/NewGroupConversation/SetContacts/SetContacts.vue
```diff
@@ -127,11 +137,14 @@ export default {
...
-	async mounted() {
-		// Focus the input field of the current component.
-		this.focusInput()
-		// Perform a search with an empty string
-		await this.fetchSearchResults()
+	mounted() {
+		this.$nextTick(() => {
+			this.mountArrowNavigation(this.$refs.wrapper, this.$refs.setContacts)
+			// Focus the input field of the current component.
+			this.focusInput()
```
DorraJaouad, line 144: Pressing Esc in the modal will close it.

I suggest to also add the Escape behavior from useArrowNavigation when the focus is on the search field or on an item from the participants list. Otherwise, the closing modal applies.

DorraJaouad (commented):
Nice :) .
Just have tiny remarks.

src/composables/useArrowNavigation.js
Antreesy: Fixed a bit differently: `focusedIndex` should be reset after each `blur` event, which is not an iteration over list

ShGKme (approved):

PR #9997, opened 2023-07-17
fix(searchBox) - Conversation search UX improvements 

### LINKED ISSUE #9974: Conversation search UX improvements
A couple of things that weren't addressed in conversation filtering.

- [ ] Upon clicking into the search conversation field, even if the user hasn't typed anything yet, the "abort search" button should be immediately visible. 
- [ ] If the conversations are mention-filtered before clicking the search field, the list should become unfiltered immediately.

DorraJaouad:
### ☑️ Resolves

* Fix #9974

* The new conversation '+' button is now hidden when the search field is expanded
* The abort search button is visible whenever the search box is focused.
* The list gets immediately cleared out from filters once the search field is clicked on. 
* If the search results are empty, a new button shows up to quickly create a conversation with the search text as a name
* The animation is more expedited ( set to 0.15s instead of 0.3s) as it requires to be quick.

https://github.com/nextcloud/spreed/assets/84044328/0f19a2ef-9d79-4573-8f0c-19f4e3ed8310

### Other
And if the search results are empty, you can create a new conversation with the search text, it will be set to private by default.
image

# screenshots

Before
image

After 
image

### 🚧 Tasks

- [ ] Code review

### 🏁 Checklist

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -446,6 +446,7 @@ export default {
 		},
 
 		setIsFocused(event) {
+			console.log(event.type)
 			if (this.searchText !== '') {
 				return
 			}
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -111,6 +111,17 @@
 				<ul ref="scroller"
 					class="scroller"
 					@scroll="debounceHandleScroll">
+					<NcListItem v-if="noMatchFound && searchText"
+						:title="t('spreed', 'Create a new conversation')"
+						@click="createConversation(searchText)">
+						<template #icon>
+							<ChatPlus :size="30" />
+						</template>
+						<template #subtitle>
+							{{ searchText }}
+						</template>
+					</NcListItem>
+
 					<NcAppNavigationCaption :class="{'hidden-visually': !isSearching}"
 						:title="t('spreed', 'Conversations')" />
 					<Conversation v-for="item of conversationsList"
@@ -146,7 +157,7 @@
 							<NcAppNavigationCaption v-if="searchResultsUsers.length === 0"
 								:title="t('spreed', 'Users')" />
 							<Hint v-if="contactsLoading" :hint="t('spreed', 'Loading')" />
-							<Hint v-else :hint="t('spreed', 'No search results')" />
+							<Hint v-else :hint="t('spreed', 'No matches found')" />
 						</template>
 					</template>
 					<template v-if="showStartConversationsOptions">
@@ -238,6 +249,7 @@ import SearchBox from './SearchBox/SearchBox.vue'
 import { useArrowNavigation } from '../../composables/useArrowNavigation.js'
 import { CONVERSATION } from '../../constants.js'
 import {
+	createPrivateConversation,
 	searchPossibleConversations,
 	searchListedConversations,
 } from '../../services/conversationsService.js'
@@ -319,7 +331,7 @@ export default {
 	computed: {
 		conversationsList() {
 			let conversations = this.$store.getters.conversationsList
-			if (this.searchText !== '') {
+			if (this.searchText !== '' || this.isFocused) {
 				const lowerSearchText = this.searchText.toLowerCase()
 				conversations = conversations.filter(conversation =>
 					conversation.displayName.toLowerCase().includes(lowerSearchText)
@@ -434,11 +446,10 @@ export default {
 		},
 
 		setIsFocused(event) {
-			if (event.relatedTarget?.className.includes('input-field__clear-button') || this.searchText !== '') {
+			if (this.searchText !== '') {
 				return
 			}
 			this.isFocused = event.type === 'focus'
-
 		},
 
 		handleFilter(filter) {
@@ -549,6 +560,17 @@ export default {
 				// For other types, show the modal directly
 				this.$refs.newGroupConversation.showModalForItem(item)
 			}
+		},
+
+		async createConversation(name) {
+			const response = await createPrivateConversation(name)
+			const conversation = response.data.ocs.data
+			this.$store.dispatch('addConversation', conversation)
+			this.abortSearch()
+			this.$router.push({
+				name: 'conversation',
+				params: { token: conversation.token },
+			}).catch(err => console.debug(`Error while pushing the new conversation's route: ${err}`))
 		},
 
 		hasOneToOneConversationWith(userId) {
@@ -796,8 +818,7 @@ export default {
 	}
 	&--expanded {
 
-		// Gets expanded : 100 % - (52px + 1px)
-		width : calc(100% - 53px );
+		width : calc(100% - 8px);
 	}
 
 }
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -27,9 +27,10 @@
 				:class="{'conversations-search--expanded': isFocused}">
 				<SearchBox ref="searchBox"
 					:value.sync="searchText"
-					:is-searching="isSearching"
+					:is-focused="isFocused"
 					@focus="setIsFocused"
 					@blur="setIsFocused"
+					@trailing-blur="setIsFocused"
 					@input="debounceFetchSearchResults"
 					@abort-search="abortSearch" />
 			</div>
@@ -75,7 +76,8 @@
 			</div>
 
 			<!-- Actions -->
-			<div class="actions">
+			<div class="actions"
+				:class="{'hidden-visually': isFocused}">
 				<NcActions class="conversations-actions">
 					<template #icon>
 						<DotsVertical :size="20" />
@@ -219,6 +221,7 @@ import debounce from 'debounce'
 import { ref } from 'vue'
 
 import AtIcon from 'vue-material-design-icons/At.vue'
+import ChatPlus from 'vue-material-design-icons/ChatPlus.vue'
 import DotsVertical from 'vue-material-design-icons/DotsVertical.vue'
 import FilterIcon from 'vue-material-design-icons/Filter.vue'
 import FilterRemoveIcon from 'vue-material-design-icons/FilterRemove.vue'
@@ -279,6 +282,7 @@ export default {
 		FilterIcon,
 		FilterRemoveIcon,
 		Plus,
+		ChatPlus,
 		List,
 		DotsVertical,
 	},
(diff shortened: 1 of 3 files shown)


marcoambrosini (approved):
Very nice, thanks for the fix :)

marcoambrosini (changes requested):
Actually, there is still a bug that happens if we try and create a new conversation during a search operation. 

https://github.com/nextcloud/spreed/assets/26852655/a89cf063-4b31-4f9d-8e91-6417907da583

I think that the best solution here is to remove the + button too during a search operation. There's really no need to keep it there and it's easy enough to interrupt the search given the presence of the `x` button even if there's no text in the search-box

nickvergessen:
> I think that the best solution here is to remove the + button too during a search operation. There's really no need to keep it there and it's easy enough to interrupt the search given the presence of the x button even if there's no text in the search-box

I always search for things and when I can not find the room have to create it. Would be annoying if I have to clear the search first, before being able to create it (in the past we added code to use the search term as default value for the name input on the dialog, in case you remember)

marcoambrosini:
> But maybe adding a placeholder/option in the search list could be a replacement?

Definitely, let's create a `NcListItem` if there are no matches that creates a new conversation without asking anything about participants and settings. Just the name and no participants like in old talk?

DorraJaouad:
> > But maybe adding a placeholder/option in the search list could be a replacement?
> 
> Definitely, let's create a `NcListItem` if there are no matches that creates a new conversation without asking anything about participants and settings. Just the name and no participants like in old talk?

something like this ? 
image

Or we can make a placeholder right below the search box that contains a link and says something like : 
"No conversations found. Consider initiating **a new conversation.**"

But I think we should just open the dialog and the search text is pasted in the Name field instead of directly creating one.

nickvergessen:
> Just the name and no participants like in old talk?

Or we bring up the dialog on page one with the name prefilled. I'd prefer that one.

marcoambrosini:
> Or we bring up the dialog on page one with the name prefilled. I'd prefer that one.

I think we should skip the dialog for this one and allow a very quick way to get a conversation going. 
If a conversation is new and it des not have participants other than you we can display a couple of buttons to **make public and copy link** in one action and **add participants** in another one.

That'd be for the future, but for now I think it would be really valuable to have a very quick way to create a conv

@DorraJaouad let's use the ListItem component with the `chatPlus` icon. 

Screenshot 2023-07-20 at 14 18 07

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -78,14 +79,26 @@
 			<!-- New Conversation -->
 			<NewGroupConversation v-if="canStartConversations"
 				ref="newGroupConversation"
-				class="new-conversation__button" />
+				class="new-conversation__button"
+				:class="{'hidden-visually': isFocused}" />
```
Antreesy, line 83: It will conflict with open conversations, so as I said earlier, let's wait for it to be merged

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -499,6 +515,10 @@ export default {
 			}
 		},
 
+		createConversationIfNotFound(name) {
+			this.$refs.newGroupConversation.createConversationIfNotFound(name)
```
Antreesy, line 519:
```suggestion
			this.$refs.newGroupConversation.createConversation(name)
```
Child component doesn't know, what the conversation is not found, so we shouldn't name the method like that

src/components/LeftSidebar/SearchBox/SearchBox.vue
```diff
@@ -66,6 +66,11 @@ export default {
 		/**
 		 * If true, this component displays an 'x' button to abort the search
 		 */
+		isFocused: {
+			type: Boolean,
+			default: false,
+		},
+
 		isSearching: {
```
Antreesy, line 74: We probably don't need this prop anymore, especially if it's calculated in parent based on this child's value

Antreesy (commented):
Haven't tested it yet, sorry

src/components/LeftSidebar/LeftSidebar.vue
DorraJaouad: yes, I will merge it after the PR of open conversations.

src/components/LeftSidebar/LeftSidebar.vue
DorraJaouad: There is already a function with that name for creating the conversation using the dialog, perhaps I should just put the function directly in leftSidebar without using the newGroupConversation component?

Antreesy: You are right, that will be better in our case

marcoambrosini (changes requested):
Nice :) 

One last thing:
Upon creating a new conversation via the list-item, the search operation should end  and the route should change to that conversation

---

**edit. found another one:**
 
Screenshot 2023-07-21 at 12 59 46

If you're not searching and filtering and the filtering has no results, the create conversation shortcut shouldn't be there. Also the empty-content shouldn't say "no search result", but something more like "No unread messages" or "No unread mentions"

Antreesy:
> Also the empty-content shouldn't say "no search result"

We could go with common wording for all cases "No matches found", like in this PR:
https://github.com/nextcloud/spreed/pull/9955/files#diff-7a1abf98ca549db81a418814a97629c796bc85dc32b40f9cffb65ce3ad40d555R125

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -384,7 +399,8 @@ export default {
 			return this.$el.querySelectorAll('li.acli_wrapper .acli')
 		},
 		setIsFocused(event) {
-			if (event.relatedTarget?.className.includes('input-field__clear-button') || this.searchText !== '') {
+			if ((event.type === 'click' && event.relatedTarget?.className.includes('input-field__clear-button'))
```
Antreesy, line 402: Same here. How could I trigger a "click" event from here? We're calling this method only from `@focus` and `@blur` of SearchBox

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -273,7 +288,7 @@ export default {
 	computed: {
 		conversationsList() {
 			let conversations = this.$store.getters.conversationsList
-			if (this.searchText !== '') {
+			if (this.searchText !== '' || this.isFocused) {
```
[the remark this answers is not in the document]
DorraJaouad: This was added to reset the search result to the full list if the user focuses on the search field when a filter is activated ( the filter should be immediately cleared when focusing on input and not when typing a search text)

src/components/LeftSidebar/LeftSidebar.vue
DorraJaouad: Changed, it was a hopeless attempt to fix the incorrect tabbing flow at root . Should be a follow-up I suppose.

src/components/LeftSidebar/SearchBox/SearchBox.vue
```diff
@@ -66,7 +66,7 @@ export default {
 		/**
 		 * If true, this component displays an 'x' button to abort the search
 		 */
-		isSearching: {
+		isFocused: {
```
Antreesy, line 69: * Will left some notes here about focusing

Antreesy (changes requested):
Note: `showStartConversationsOptions()` method returns `this.isSearching && this.canStartConversations` and is used inside `<template v-if="isSearching">` in one place and could be moved to this template in another place

Please check if this method could be removed, and replaced with `v-if="canStartConversations" + v-else`

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -112,6 +112,17 @@
 				<ul ref="scroller"
 					class="scroller"
 					@scroll="debounceHandleScroll">
+					<NcListItem v-if="noMatchFound && searchText"
+						:title="t('spreed', 'Create a new conversation')"
+						@click="createConversation(searchText)">
+						<template #icon>
+							<ChatPlus :size="30" />
```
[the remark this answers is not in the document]
DorraJaouad: don't worry XD , I know that,  I am still working on the PR and haven't pushed all changes yet

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -27,78 +27,82 @@
...
-						{{ t('spreed','Filter unread messages') }}
-					</NcActionButton>
-
-					<NcActionButton v-if="isFiltered"
-						close-after-click
-						class="filter-actions__clearbutton"
-						@click="handleFilter(null)">
+			<transition-group name="radial-reveal">
+				<!-- Filters -->
+				<div v-show="!isFocused" key="filters" class="filters">
```
Antreesy, line 41:
```suggestion
				<div v-show="!isSearching" key="filters" class="filters">
```

Non-blocking: I'd replace this parameter for better handling of keyboard navigation:

`isFocused`:

https://github.com/nextcloud/spreed/assets/93392545/35cbe42a-d215-4132-9a37-e26d6dd7272d

`isSearching`:

https://github.com/nextcloud/spreed/assets/93392545/1a8c929f-9586-402f-9ced-703375f6455c

:warning: Fancy animation won't work with this configuration, so would be wise to discuss with design team

Also outline radius for trailing button should be round, like here:
```
:deep(.input-field__input) {
}
```

Antreesy (approved):
Works on web client, looks also cool!

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -35,71 +35,73 @@
...
+							</template>
+							{{ t('spreed', 'Clear filters') }}
+						</NcActionButton>
+					</NcActions>
+				</div>
 
-					<NcActionButton close-after-click
-						@click="showModalListConversations">
+				<!-- Actions -->
+				<div v-show="!isFocused"
```
Antreesy, line 79: Maybe put them into one `<div>` wrapper with pos: absolute and display: flex, as they are following the same logic?
then `<transition-group>` should be replaces with `<transition>`, and i think, it should work without `key` param

DorraJaouad: that leads to a different animation as it will be counted on the wrapper and not on each button but let's wait for @marcoambrosini to check whether to keep it or not.

https://github.com/nextcloud/spreed/assets/84044328/574243bc-ee0f-4749-bd49-e8ee2d0a4a78

marcoambrosini: Looks good, as discussed we've increased the animation speed

marcoambrosini (approved):

PR #10018, opened 2023-07-19
chore(deps): add @types/jest

ShGKme:
### ☑️ Resolves

We use global Jest function without typings. IDEs doesn't understand them, show warnings and no auto-complete.

`@types/jest` adds global jest typings.

**Alternative solution**: use `@jest/globals` with explicit import everywhere:

```js
import { describe, it, test, expect, jest } from '@jest/globals'

describe('...', () => {
  it('..', () => {})
})
```

**Docs**: https://jestjs.io/docs/getting-started#type-definitions

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
JetBrains IDEs | -
image | image
Microsoft VSCode | -
image | image
JetBrains Fleet | -
image | image

### 🏁 Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -64,6 +64,7 @@
         "@nextcloud/eslint-config": "^8.2.1",
         "@nextcloud/stylelint-config": "^2.3.1",
         "@nextcloud/webpack-vue-config": "^5.5.1",
+        "@types/jest": "^29.5.3",
         "@vue/test-utils": "^1.3.6",
         "@vue/vue2-jest": "^29.2.4",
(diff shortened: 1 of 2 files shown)


Antreesy (approved):
Added it recently to the local IDE [configuration](https://intellij-support.jetbrains.com/hc/en-us/community/posts/115000357324-Get-rid-of-Unresolved-function-method-variable-warning-in-Jest-test-files)

But as long as it's not going to production build - fine to have it in dev-deps

nickvergessen (approved):

PR #10021, opened 2023-07-19
Revert mapping of cjs module

Antreesy:
### ☑️ Resolves

* Regression from #9919 
* After bumping #9915, all the packages are using a new major version of `axios`, which makes it unnecessary to map it with CJS module 
* Affects only Jest work, so as long as tests have passed, rest should be fine

### 🏁 Checklist


ShGKme (approved):
🧹

PR #10048, opened 2023-07-24
fix(MessagesList) - check for temporary messages while soft updating

Antreesy:
### ☑️ Resolves

* Regression from #9897 
* When post a new message (not the first in the message group), response from server replaces temporary message, but group length remains the same, and component with temp message won't re-render => infinite loading

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks

- [ ] Code review
- [ ] Manual testing

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -339,13 +339,25 @@ export default {
 
 			// Check if we have this group in the old list already and it is unchanged
 			return newGroups.map(newGroup => oldGroupsMap.has(newGroup.id)
-				&& newGroup.messages.length === oldGroupsMap.get(newGroup.id).messages.length
-				&& newGroup.dateSeparator === oldGroupsMap.get(newGroup.id).dateSeparator
-				&& newGroup.previousMessageId === oldGroupsMap.get(newGroup.id).previousMessageId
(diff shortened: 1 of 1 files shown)


src/components/MessagesList/MessagesList.vue
```diff
@@ -339,15 +339,27 @@ export default {
...
+				|| group1.previousMessageId !== group2.previousMessageId
+				|| group1.nextMessageId !== group2.nextMessageId) {
+				return false
+			}
+
+			// Check for temporary messages, replaced with messages from server
+			const array1 = group1.messages.map(message => message.id)
+			const array2 = group2.messages.map(message => message.id)
+
+			return array1.every(item => array2.includes(item))
```
ShGKme, line 360: It's unlikely we will have hundreds messages here, but I do not like `O(n^2)` here for this check.

Temp messages have a special `id` starting with `temp-` and it is changed after posting. Instead of a quadratic search, can we check only `id`s here between correspond objects?

Antreesy: LIke this?
```
return array1.every((message, index) => array2[index].id === message.id)
```

PR #10051, opened 2023-07-24
refactor(conversations): separate search results and conversations

ShGKme:
### ☑️ Resolves

LeftSidebar refactoring

### 🖼️ Screenshots

No visual changes

### 🚧 Tasks

- Fix "// TODO": instead of mutating the getter result as computed side effect, move sorting conversations into `conversationsList` getter
- Divide storing all conversations in the LeftSidebar into: "All conversations list" + "Filtered conversations list" + "Search result conversations list"
  - It allows to easily add `v-show` (`KeepAlive`) or `VirtualScrolling` in future
- Organize rendering in `LeftSIdebar`
  - Make conditions less labyrinthine =D
  - Remove duplication in conditions (the "no search result" part)

### 🏁 Checklist

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -790,7 +790,7 @@ export default {
 }
 
 .new-conversation {
-  position: relative;
+	position: relative;
 	display: flex;
 	padding: 8px 4px 8px 12px;
 	align-items: center;
@@ -799,17 +799,17 @@ export default {
 		border-bottom: 1px solid var(--color-placeholder-dark);
 	}
 
-  .filters {
-    position: absolute;
-    top : 8px;
-    right: 56px;
-  }
+	.filters {
+		position: absolute;
+		top : 8px;
+		right: 56px;
+	}
 
-  .actions {
-    position: absolute;
-    top: 8px;
-    right: 8px;
-  }
+	.actions {
+		position: absolute;
+		top: 8px;
+		right: 8px;
+	}
 }
 
 // Override vue overflow rules for <ul> elements within app-navigation
@@ -831,7 +831,7 @@ export default {
 }
 
 .conversations-search {
-  padding: 4px 0;
+	padding: 4px 0;
 	transition: all 0.15s ease;
 	z-index: 1;
 	// New conversation button width : 52 px
@@ -865,6 +865,7 @@ export default {
 :deep(.input-field__clear-button) {
 	border-radius: var(--border-radius-pill) !important;
 }
+
 :deep(.app-navigation ul) {
 	padding: 0 !important;
 }
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -113,89 +113,98 @@
 		<template #list>
 			<li ref="container" class="left-sidebar__list" @scroll="debounceHandleScroll">
 				<ul class="scroller">
-					<NcListItem v-if="noMatchFound && searchText && canStartConversations"
-						:title="t('spreed', 'Create a new conversation')"
-						@click="createConversation(searchText)">
-						<template #icon>
-							<ChatPlus :size="30" />
-						</template>
-						<template #subtitle>
-							{{ searchText }}
-						</template>
-					</NcListItem>
-
-					<NcAppNavigationCaption :class="{'hidden-visually': !isSearching}"
-						:title="t('spreed', 'Conversations')" />
-					<Conversation v-for="item of conversationsList"
-						:key="item.id"
-						:ref="`conversation-${item.token}`"
-						:item="item" />
-					<template v-if="!initialisedConversations">
-						<LoadingPlaceholder type="conversations" />
+					<!-- Conversations List -->
+					<template v-if="!isSearching">
+						<NcAppNavigationCaption :title="t('spreed', 'Conversations')" class="hidden-visually" />
+						<Conversation v-for="item of filteredConversationsList"
+							:key="`conversation_${item.id}`"
+							:ref="`conversation-${item.token}`"
+							:item="item" />
+						<LoadingPlaceholder v-if="!initialisedConversations" type="conversations" />
+						<Hint v-else-if="filteredConversationsList.length === 0" :hint="t('spreed', 'No matches found')" />
 					</template>
-					<Hint v-else-if="noMatchFound"
-						:hint="t('spreed', 'No matches found')" />
-					<template v-if="isSearching">
-						<template v-if="!listedConversationsLoading && searchResultsListedConversations.length > 0">
+
+					<!-- Search results -->
+					<template v-else-if="isSearching">
+						<!-- Create a new conversation -->
+						<NcListItem v-if="searchResultsConversationList.length === 0 && canStartConversations"
+							:title="t('spreed', 'Create a new conversation')"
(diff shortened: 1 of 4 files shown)


src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -114,76 +114,87 @@
...
-								:key="item.id"
-								:title="item.label"
-								@click="createAndJoinConversation(item)">
-								<template #icon>
-									<ConversationIcon :item="iconData(item)"
-										:disable-menu="true" />
-								</template>
-							</NcListItem>
+						<!-- Search results: new conversations -->
+						<template v-if="showStartConversationsOptions">
```
Antreesy, line 162:
```suggestion
						<template v-if="canStartConversations">
```

Method is redundant, it uses `isSearching && canStartConversations` inside of the same template

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -114,76 +114,87 @@
...
+									</template>
+								</NcListItem>
+							</template>
 						</template>
 
-						<NcAppNavigationCaption v-if="sourcesWithoutResults"
-							:title="sourcesWithoutResultsList" />
+						<!-- Search results: no results (yet) -->
+						<NcAppNavigationCaption v-if="sourcesWithoutResults" :title="sourcesWithoutResultsList" />
 						<Hint v-if="contactsLoading" :hint="t('spreed', 'Loading')" />
```
Antreesy, line 192: Like this, maybe?
```suggestion
						<Hint v-if="contactsLoading" :hint="t('spreed', 'Loading …')" />
```

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -114,76 +114,87 @@
...
 							<Conversation v-for="item of searchResultsListedConversations"
 								:key="item.id"
 								:item="item"
 								is-search-result />
 						</template>
+
+						<!-- Search results: users -->
 						<template v-if="searchResultsUsers.length !== 0">
 							<NcAppNavigationCaption :title="t('spreed', 'Users')" />
 							<NcListItem v-for="item of searchResultsUsers"
```
Antreesy, line 151: Results are cached between different search attempts, until new response is received, maybe reset it?

```js
async fetchPossibleConversations() {
			this.contactsLoading = true
			this.searchResults = []
			this.searchResultsUsers = []
			this.searchResultsGroups = []
			this.searchResultsCircles = []
}
```
Or even move specific searchResults to computed properties?

Antreesy (approved):
Nothing blocking, just a couple of remarks to improve

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -114,76 +114,87 @@
...
-										:disable-menu="true" />
-								</template>
-							</NcListItem>
+						<!-- Search results: new conversations -->
+						<template v-if="canStartConversations">
+							<!-- New conversations: Groups -->
+							<template v-if="searchResultsGroups.length !== 0">
+								<NcAppNavigationCaption :title="t('spreed', 'Groups')" />
+								<NcListItem v-for="item of searchResultsGroups"
+									:key="item.id"
```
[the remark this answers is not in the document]
ShGKme: While we have `NcAppNavigationCaption` between `v-for`-ed elements it is not needed (they are not mixed into one list in the end). But we can add it for future just in case.

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -114,76 +114,87 @@
...
+					<!-- Conversations List -->
+					<template v-else-if="!isSearching">
+						<Conversation v-for="item of filteredConversationsList"
+							:key="item.id"
+							:ref="`conversation-${item.token}`"
+							:item="item" />
+						<Hint v-if="filteredConversationsList.length === 0" :hint="t('spreed', 'No matches')" />
 					</template>
-					<Hint v-else-if="noMatchFound"
-						:hint="t('spreed', 'No matches found')" />
```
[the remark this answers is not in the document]
ShGKme: Yep, nice catch. Changed in [Add open conv list dialog](https://github.com/nextcloud/spreed/commit/c08bc6445c68e9940d3b8b0fc39a622aeb0a031e) and lost during rebase.

src/components/LeftSidebar/LeftSidebar.vue
ShGKme: I'd make it in a separated from refactoring PR or discuss it in an issue as it changes behavior a bit. It makes the search "slower" from UX.

src/components/LeftSidebar/LeftSidebar.vue
ShGKme: I'm not aware, but it was "Loading" before refactoring and it will be a new translation token.

src/components/LeftSidebar/LeftSidebar.vue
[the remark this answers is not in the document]
Antreesy: Let's add it then, as a safety measure (and to not disturb my console during the search).
From the DOM perspective, it's still a single `<ul>`, where `<li>` have identical keys

ShGKme:
- Updated according to the review
- Rebased onto master with many `LeftSidebar` updates
- Stylelint styles, there were wrong tabulation

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -113,89 +113,103 @@
...
-							<ChatPlus :size="30" />
-						</template>
-						<template #subtitle>
-							{{ searchText }}
-						</template>
-					</NcListItem>
-
-					<NcAppNavigationCaption :class="{'hidden-visually': !isSearching}"
+					<NcAppNavigationCaption v-if="!isSearching"
+						class="hidden-visually"
```
Antreesy, line 117: It's always hidden from the DOM now. Maybe remove NcAppNavigationCaption from here at all?

Antreesy (approved):
Still works :seal:

nickvergessen:
> Conflicting files
> src/components/LeftSidebar/LeftSidebar.vue

ShGKme:
Rebased onto master

ShGKme:
There is a regression. 

On the first load from a conversation page, before there was the current conversation in the list together with loading. Now there is no conversation before loading.

Actually, I don't know if it was intended before :D

@Antreesy @DorraJaouad @marcoambrosini @nickvergessen What do you think?

Before | After
---|---
image | image

Antreesy:
> Actually, I don't know if it was intended before :D

Probably not, as when you join a room, you fetch it separately from others. That's why it ended up alone in the store and in the list
Don't think, that it will be an issue with cached conversations, as we have a starting point already

marcoambrosini:
> before there was the current conversation in the list together with loading

I think it's a nice to have if in the content there's that conversation opened

nickvergessen:
Both behaviours are fine to me, so whatever is better to do/maintain

ShGKme:
> I think it's a nice to have if in the content there's that conversation opened

I checked that it is possible to make it work with both this loading and virtual scrolling.

Though, it is noticeable only after re-login when the new cache feature doesn't work, I decided to fix it.

So no regression now

Antreesy (approved):

ShGKme:
Rebased onto master and fixup.

PR #10053, opened 2023-07-24
feat(attachments): Add option for captions

### LINKED ISSUE #5354: attachment caption
Wouldn't it be nice if we had the possibility to add a caption to the file upload?
This time the user posts the file share in the first message and sometimes a second message with a description.

nickvergessen:
### ☑️ Resolves

* Fix #5354 

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🏁 Checklist

### DIFF
--- a/lib/Chat/Parser/Listener.php
+++ b/lib/Chat/Parser/Listener.php
@@ -86,7 +86,7 @@ class Listener {
 
 		try {
 			$parser->parseMessage($message);
-			$event->stopPropagation();
+			// Disabled so we can parse mentions in captions: $event->stopPropagation();
 		} catch (\OutOfBoundsException $e) {
 			// Unknown message, ignore
 		}
--- a/lib/Chat/Parser/SystemMessage.php
+++ b/lib/Chat/Parser/SystemMessage.php
@@ -406,6 +406,10 @@ class SystemMessage {
 				} else {
 					$chatMessage->setMessageType(ChatManager::VERB_MESSAGE);
 				}
+
+				if (isset($metaData['caption']) && $metaData['caption'] !== '') {
+					$parsedMessage = $metaData['caption'] . '{file}';
+				}
 			} catch (\Exception $e) {
 				$parsedMessage = $this->l->t('{actor} shared a file which is no longer available');
 				if ($currentUserIsActor) {
--- a/lib/Chat/SystemMessage/Listener.php
+++ b/lib/Chat/SystemMessage/Listener.php
@@ -381,6 +381,10 @@ class Listener implements IEventListener {
 		$metaData = json_decode($metaData, true);
 		$metaData = is_array($metaData) ? $metaData : [];
 
+		if (isset($metaData['noMessage'])) {
+			return;
+		}
+
 		if (isset($metaData['messageType']) && $metaData['messageType'] === 'voice-message') {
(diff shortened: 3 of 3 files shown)


Antreesy (approved):

SystemKeeper:
Do we have a limit on how many files can be in one message?!

docs/chat.md
```diff
@@ -190,12 +190,44 @@ See [OCP\RichObjectStrings\Definitions](https://github.com/nextcloud/server/blob
...
 
-| field         | type   | Description                                                                                            |
-|---------------|--------|--------------------------------------------------------------------------------------------------------|
-| `messageType` | string | A message type to show the message in different styles. Currently known: `voice-message` and `comment` |
+| field         | type   | Description                                                                                                                                                          |
+|---------------|--------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `messageType` | string | A message type to show the message in different styles. Currently known: `voice-message` and `comment`                                                               |
+| `caption`     | string | A caption message that should be shown together with the shared file (only available with `media-caption` capability)                                                |
```
SystemKeeper, line 196: Will the caption be markdown rendered?

docs/chat.md
```diff
@@ -190,12 +190,44 @@ See [OCP\RichObjectStrings\Definitions](https://github.com/nextcloud/server/blob
...
+
+* Required capability: `media-caption`
+* Method: `POST`
+* Endpoint: `/chat/{token}/share-files`
+* Data:
+
+| field              | type   | Description                                                                                                                       |
+|--------------------|--------|-----------------------------------------------------------------------------------------------------------------------------------|
+| `shareIds`         | int[]  | Array with the integer ID of the shares                                                                                           |
+| `caption`          | string | A caption message that should be shown together with the shared files                                                             |
```
SystemKeeper, line 211: Same as above, markdown?

SystemKeeper:
> Do we have a limit on how many files can be in one message?!

It's 64 currently: 

https://github.com/nextcloud/spreed/pull/10053/files#diff-1c8f2a0b4aae87ba3143f0028e4964ba5906d5a2c0a966f67a3f82d47ed11e5fR298

docs/chat.md
Antreesy: Last time I tested it with web frontend, markdown worked fine together, so don't see a reason to prevent it. Could it be tricky to realize on mobiles?

SystemKeeper: No, would be fine, but if it should be just plain, I wouldn’t add markdown in the first place :)

Antreesy:
These methods should return messages with multiple objects:
https://github.com/nextcloud/spreed/blob/d0d52ce2901c56ef9b03012c029a09a641c21bac/lib/Controller/ChatController.php#L924

https://github.com/nextcloud/spreed/blob/d0d52ce2901c56ef9b03012c029a09a641c21bac/lib/Controller/ChatController.php#L977

nickvergessen:
> These methods should return messages with multiple objects:

I tried looking into that, but it's problematic. The database has no record of the "shared object" in that table. It basically assumes there is a one-to-one mapping of shared item and message and therefore storing/parsing the message ID is good enough.

But in order to achieve this we will need to add a new column to the database to store the object/parameter that should be rendered.

Antreesy (approved):
Tested against updated frontend (linked above)

PR #10058, opened 2023-07-26
Fix duplicating messages and improve performance

### LINKED ISSUE #10027: [Bug] Dissapearing messages / duplicated messages in chat
### How to use GitHub

---

Seems like this mutation causes incorrect object reference to be patched with wrong data:
https://github.com/nextcloud/spreed/blob/319742603e7d571000ce73da4472c48b2616b2a1/src/store/messagesStore.js#L284-L295

Resulting in replacing the message in store and in DOM tree (message `id=2017713` dissapears, and message `id=2017786` is duplicated):

```js
2017713: {
    "id": 2017786,
    "token": "TOKEN",
    "actorType": "users",
    "actorId": "actor",
    "actorDisplayName": "...",
    "timestamp": 1689252346,
    "message": "> temporary messages has relatively longer id with prefix 'temp-'...",
    ...
},
...
2017786: {
    "id": 2017786,
    "token": "TOKEN",
    "actorType": "users",
    "actorId": "actor",
    "actorDisplayName": "...",
    "timestamp": 1689252346,
    "message": "> temporary messages has relatively longer id with prefix 'temp-'...",
    ...
}
```
image
---

### Steps to reproduce
Algorithm for reproducing is not defined, as it happens sometime after fetching a new messages from server.
Upd: happens quite often for system messages:
image

### Expected behaviour
Messages should not be duplicated

### Actual behaviour
See screenshot above

## Talk app

**Talk app version:** at least 17

ShGKme:
### ☑️ Resolves

* Fix #10027 

#### Steps to reproduce:

1. Open **ChatA**
2. Switch to **ChatB**
3. From another tab or user send a message to **ChatA**
4. Wait until `LeftSidebar` will be updated
5. Return back to **ChatA**
6. You can see duplicate messages 😣

#### Why it happens:

1. In the `messagesStore` we update `conversation.lastMessage` in the conversation store using a new received message in the chat. Here a reference is passed to the `conversationsStore`'s action
https://github.com/nextcloud/spreed/blob/d381c9c031822aaafcde2337ea49f78ce597a8ce/src/store/messagesStore.js#L1065-L1069

2. In the `conversationsStore` then this reference is saved in the store:

https://github.com/nextcloud/spreed/blob/d381c9c031822aaafcde2337ea49f78ce597a8ce/src/store/conversationsStore.js#L151-L153

3. At this step `conversationsState.conversation[token].lastMessage` stores a reference to a message object from `messagesStore`

4. Then we patch `conversationsState.conversation[token].lastMessage` after receiveing new conversation on re-fetching. This mutates the object in the messages store.

https://github.com/nextcloud/spreed/blob/d381c9c031822aaafcde2337ea49f78ce597a8ce/src/store/conversationsStore.js#L129-L131

#### Another problem: 

There is also `conversation.lastMessage.messageParameters` array. `patchObject` is not patching sub-arrays, so it caused almost unnecessary re-renders anyway.

### 🖼️ Screenshots

Green highlight = re-rendering. After PR there are more re-renderings in changed conversation items, but no re-renderings in not changed items.

🏚️ Before | 🏡 After
---|---
image | image
before-rerender | after-rerender

### 🚧 Tasks

- [x] Instead of patching conversation - re-assign it when the new conversation has larger `lastActivity`
  - Drawbacks:
    - It assumes that `lastActivity` is enough to know if there were any changes in the conversation
    - It could be a little bit less optimal than patching

### 👀 Alternatives

1. Clone `lastMessage` on update from `messagesStore`
  - But there is a patching error anyway. So
    1. Adjust `patchObject` to handle arrays (check length and elements)
      - But it seemed to be too overkill to me
2. Adjust `patchObject` to handle "special" cases and list custom setters for specific keys (`lastMessage`)
   - But it seemed to be too overkill to me
3. Store `lastMessageId` instead of `lastMessage` in the conversation, same as `lastCommonReadMessage`
   - I'd like this solution, but it would require many changes. The best alternative solution for me

### 🏁 Checklist


src/store/conversationsStore.js
```diff
@@ -118,16 +117,15 @@ const mutations = {
...
-	 * - delete properties
+	 * Update stored conversation if it has changes according to lastActivity
 	 *
 	 * @param {object} state the state
 	 * @param {object} conversation the new conversation object
 	 */
-	patchConversation(state, conversation) {
-		patchObject(state.conversations[conversation.token], conversation)
+	updateConversationIfHasActivity(state, conversation) {
+		if (state.conversations[conversation.token].lastActivity < conversation.lastActivity) {
```
nickvergessen, line 126: I think we also need to do it when the unread counter changes, because reading a chat does not update the lastactivity

ShGKme:
@nickvergessen do you think the change in the last commit is too much and should be replaced with just direct `'unreadMessages'`, `'unreadMention'`, `'unreadMentionDirect'` checks?

ShGKme:
@nickvergessen @Antreesy If this is fine for you, I'll add some tests and ready to merge

src/store/conversationsStore.js
```diff
@@ -209,7 +251,9 @@ const actions = {
 	 * @param {object} conversation the conversation;
 	 */
 	addConversation(context, conversation) {
-		context.dispatch('emitUserStatusUpdatedWhenChanged', conversation)
+		if (conversation.type === CONVERSATION.TYPE.ONE_TO_ONE) {
```
Antreesy, line 254:
```suggestion
		if (conversation.status) {
```

Group conversations don't have a `status` field, so I'm not sure if we needed that additional check before and need that now.

src/store/conversationsStore.js
```diff
@@ -85,6 +84,23 @@ const DUMMY_CONVERSATION = {
 	isDummyConversation: true,
 }
 
+/**
+ * Emit global event for user status update with the status from a 1-1 conversation
+ *
+ * @param {object} conversation - a 1-1 conversation
+ */
+function emitUserStatusUpdated(conversation) {
+	if (conversation.status) {
```
Antreesy, line 93:
```suggestion
```

And thus we don't that check here, as we do it outside of each call

src/store/conversationsStore.js
```diff
@@ -118,16 +134,42 @@ const mutations = {
...
 	 * @param {object} conversation the new conversation object
 	 */
-	patchConversation(state, conversation) {
-		patchObject(state.conversations[conversation.token], conversation)
+	updateConversationIfHasChanged(state, conversation) {
+		const oldConversation = state.conversations[conversation.token]
+
+		// Update 1-1 conversation, if its status was changed
+		if (conversation.type === CONVERSATION.TYPE.ONE_TO_ONE && conversation.status) {
+			if (oldConversation.status !== conversation.status
```
Antreesy, line 147:
```suggestion
		if (conversation.status && (oldConversation.status !== conversation.status
```

Antreesy (approved):
Nothing blocking from me.
Tested with replies, works properly now

src/store/conversationsStore.js
ShGKme: @nickvergessen Do you remember if there was a reason to add this check here: https://github.com/nextcloud/spreed/pull/9411

Except readability

nickvergessen: It could be that the user has no status now, but had one before. Then the event should trigger and delete the status.
Can e.g. happen when they go invisible, or when the inactive status gets purged.

That being said, it doesn't make sense with the rest of the code, so I guess no.

ShGKme: As we never have this request without `includeStatus`, I added emit for null status as emitting about status removal.

ShGKme:
Updated user status update emitting.

Added test for all these cases around conversations updating and user status update emitting.

Antreesy (approved):
:seal:

ShGKme:
Rebased onto master and squashed

PR #10084, opened 2023-07-30
feat(MessagesList) - history mode for old messages

### LINKED ISSUE #6046: "MessagesList.vue:646 Message to focus not found in DOM 988" when I clicked a search result in the Nextcloud unified search prompt

Antreesy:
### ☑️ Resolves

* Attempt to close #6046
* Introduce a history mode for old messages:
  * opened via direct link
  * opened via unified search
* after context is loaded, additional parameter is checking, whether context message is old enough to not load new messages after it, and disable polling for new messages
* Old messages are not longer freezing a browser tab

### 🖼️ Screenshots

https://github.com/nextcloud/spreed/assets/93392545/889c7230-b283-41a1-a861-9632bec8fb82

### 🚧 Tasks

- [ ] Known issues:
  - [ ] Normal mode loads context around lastReadMessage - if it exists out of `HISTORY_LIMIT` scheduled time, it is also counted as history mode
  - [ ] Switching between modes purges the messages list, whatever is fetched there
- [ ] Suggestion (current or follow-up):
  - [ ] Don't purge current messages list, instead introduce a new state param `historyMessages` to store them separately
  - [ ] Also don't disable polling for chat, if was previously loaded in normal mode => instead store them for a moment 
- [ ] Manual testing
- [ ] Design review
- [ ] Code review

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -42,10 +42,14 @@ get the messagesList array and loop through the list to generate the messages.
 			:date-separator="item.dateSeparator"
 			:previous-message-id="item.previousMessageId"
 			:next-message-id="item.nextMessageId" />
+
+		<!-- Loading messages placeholder -->
 		<template v-if="showLoadingAnimation">
 			<LoadingPlaceholder type="messages"
 				:count="15" />
 		</template>
+
+		<!-- No messages placeholder -->
 		<NcEmptyContent v-else-if="showEmptyContent"
 			:title="t('spreed', 'No messages')"
 			:description="t('spreed', 'All messages have expired or have been deleted.')">
@@ -53,6 +57,17 @@ get the messagesList array and loop through the list to generate the messages.
 				<Message :size="64" />
 			</template>
 		</NcEmptyContent>
+
+		<!-- History mode information -->
+		<!-- TODO proper wording -->
+		<div v-if="isConversationInHistoryMode" class="scroller__wrapper">
+			<div class="scroller__wrapper-content">
+				<span>{{ t('spreed', 'This conversation is in history mode') }}</span>
+				<NcButton @click="clearRouterHash">
+					{{ t('spreed', 'Load current history') }}
+				</NcButton>
+			</div>
+		</div>
 	</div>
 </template>
 
@@ -68,6 +83,7 @@ import { getCapabilities } from '@nextcloud/capabilities'
 import { subscribe, unsubscribe } from '@nextcloud/event-bus'
 import moment from '@nextcloud/moment'
 
+import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
 import NcEmptyContent from '@nextcloud/vue/dist/Components/NcEmptyContent.js'
 
 import LoadingPlaceholder from '../LoadingPlaceholder.vue'
@@ -84,6 +100,7 @@ export default {
 		LoadingPlaceholder,
 		MessagesGroup,
 		Message,
+		NcButton,
 		NcEmptyContent,
 	},
 
@@ -204,6 +221,10 @@ export default {
 
 		hasMoreMessagesToLoad() {
 			return this.$store.getters.hasMoreMessagesToLoad(this.token)
+		},
+
+		isConversationInHistoryMode() {
+			return this.$store.getters.isConversationInHistoryMode(this.token)
 		},
 
 		/**
@@ -602,8 +623,11 @@ export default {
 
 				this.isInitialisingMessages = false
 
-				// get new messages
-				await this.lookForNewMessages()
+				if (!this.isConversationInHistoryMode) {
+					await this.lookForNewMessages()
+				} else {
+					this.$store.dispatch('cancelLookForNewMessages', { requestId: this.chatIdentifier })
+				}
 
 				if (this.loadChatInLegacyMode || focusMessageId === null) {
 					// don't scroll if lookForNewMessages was polling as we don't want
@@ -766,7 +790,8 @@ export default {
 			const tolerance = 10
 
 			// For chats, scrolled to bottom or / and fitted in one screen
-			if (scrollOffset < clientHeight + tolerance && scrollOffset > clientHeight - tolerance && !this.hasMoreMessagesToLoad) {
+			if (scrollOffset < clientHeight + tolerance && scrollOffset > clientHeight - tolerance
+				&& (!this.hasMoreMessagesToLoad || this.isConversationInHistoryMode)) {
 				this.setChatScrolledToBottom(true)
 				this.displayMessagesLoader = false
 				this.previousScrollTopValue = scrollTop
@@ -1082,29 +1107,32 @@ export default {
 				&& from.token === to.token
 				&& from.hash !== to.hash) {
 
+				// the hash is cleared, need to purge messages list and load new messages
+				if (this.isConversationInHistoryMode && !to.hash) {
+					await this.$store.dispatch('deleteMessages', this.token)
+					this.$nextTick(async () => {
+						await this.handleStartGettingMessagesPreconditions()
+					})
+				}
+
 				// the hash changed, need to focus/highlight another message
 				if (to.hash && to.hash.startsWith('#message_')) {
 					const focusedId = this.getMessageIdFromHash(to.hash)
 					if (this.messagesList.find(m => m.id === focusedId)) {
 						// need some delay (next tick is too short) to be able to run
-						// after the browser's native "scroll to anchor" from
-						// the hash
+						// after the browser's native "scroll to anchor" from the hash
 						window.setTimeout(() => {
 							// scroll to message in URL anchor
 							this.focusMessage(focusedId, true)
 						}, 2)
 					} else {
-						// Update environment around context to fill the gaps
-						this.$store.dispatch('setFirstKnownMessageId', {
-							token: this.token,
-							id: focusedId,
+						// the message is far from current list, need to purge messages list and load old messages
+						// TODO delete only for 'history' mode - need to know conditions (context timestamp)
+						this.$store.dispatch('deleteMessages', this.token)
+						this.$nextTick(async () => {
+							await this.handleStartGettingMessagesPreconditions()
+							this.focusMessage(focusedId, true)
 						})
-						this.$store.dispatch('setLastKnownMessageId', {
-							token: this.token,
-							id: focusedId,
-						})
-						await this.getMessageContext(focusedId)
-						this.focusMessage(focusedId, true)
 					}
 				}
 			}
@@ -1121,6 +1149,11 @@ export default {
 
 		onWindowFocus() {
 			this.refreshReadMarkerPosition()
+		},
+
+		clearRouterHash() {
+			this.$router.push({ name: 'conversation', params: { token: this.token } })
+				.catch(err => console.debug(`Error while pushing the new conversation's route: ${err}`))
 		},
 	},
 }
@@ -1149,6 +1182,21 @@ export default {
 		height: 40px;
 		width: 40px;
 		transform: translatex(-64px);
+	}
+
+	&__wrapper {
+		max-width: 800px;
+		padding: 8px 0 8px 52px;
+		margin: 0 auto;
+
+		&-content {
+			display: flex;
+			justify-content: center;
+			align-items: center;
+			gap: 8px;
+			padding-right: 140px;
+			color: var(--color-text-maxcontrast);
+		}
 	}
 }
 
--- a/src/constants.js
+++ b/src/constants.js
@@ -27,6 +27,7 @@ export const SIGNALING = {
 export const CHAT = {
 	FETCH_LIMIT: 100,
 	MINIMUM_VISIBLE: 5,
+	HISTORY_LIMIT: 24 * 60 * 60, // 24 hours in seconds
 }
 
 export const CALL = {
--- a/src/store/messagesStore.js
+++ b/src/store/messagesStore.js
@@ -167,7 +167,7 @@ const getters = {
 		return []
 	},
 	message: (state) => (token, id) => {
-		if (state.messages[token][id]) {
+		if (state.messages[token]?.[id]) {
 			return state.messages[token][id]
 		}
 		return {}
@@ -410,8 +410,8 @@ const mutations = {
 		}
 	},
 
-	loadedMessagesOfConversation(state, { token }) {
-		Vue.set(state.loadedMessages, token, true)
+	setLoadedMessagesOfConversation(state, { token, value }) {
+		Vue.set(state.loadedMessages, token, value)
 	},
 
 	// Decreases reaction count for a particular reaction on a message
@@ -700,6 +700,7 @@ const actions = {
 	 */
 	deleteMessages(context, token) {
 		context.commit('deleteMessages', token)
+		context.commit('setLoadedMessagesOfConversation', { token, value: false })
 	},
 
 	/**
@@ -828,7 +829,7 @@ const actions = {
 			})
 		}
 
-		context.commit('loadedMessagesOfConversation', { token })
+		context.commit('setLoadedMessagesOfConversation', { token, value: true })
 
 		if (minimumVisible > 0) {
 			// There are not yet enough visible messages loaded, so fetch another chunk.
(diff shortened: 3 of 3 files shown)


lib/Controller/ChatController.php
```diff
@@ -818,6 +818,9 @@ public function getMessageContext(
 			$commentsHistory = $this->chatManager->getHistory($this->room, $messageId, $limit, true, $threadId);
 			$commentsHistory = array_reverse($commentsHistory);
 			$commentsFuture = $this->chatManager->waitForNewMessages($this->room, $messageId, $limit, 0, $currentUser, false);
+			$commentsFuture = array_filter($commentsFuture, function (IComment $comment) use ($threadId) {
```
nickvergessen: If we do this (and I'm not sure about it), it should be in the DB query

Antreesy: Sure. That's a temporary workaround

src/stores/chat.ts
```diff
@@ -66,11 +74,10 @@ export const useChatStore = defineStore('chat', () => {
...
 
-		// FIXME temporary show all messages for given token from all chat blocks - no behaviour change
-		const contextBlock = chatBlocks[token].reduce<Set<number>>((acc, set) => {
-			set.forEach((id) => acc.add(id))
-			return acc
-		}, new Set())
+		// Look for a set containing given context id (return first block as fallback for not found / constants)
+		const contextBlock = (messageId <= 0)
+			? chatBlocks[token][0]
+			: chatBlocks[token].find((set) => set.has(messageId)) ?? chatBlocks[token][0]
```
DorraJaouad, line 82: I think we should insure that the id is central meaning that if for example, on one side, it has less than 10 messages then we need to check neighboring blocks or fetch.

src/composables/useGetMessages.ts
```diff
@@ -167,33 +196,37 @@ export function useGetMessagesProvider() {
...
-				EventBus.emit('focus-message', focusMessageId)
-			}, 2)
-			return
+			contextMessageId.value = focusMessageId
+		} else if (threadId && from.query.threadId !== to.query.threadId) {
+			// FIXME temporary get thread messages from the start
+			contextMessageId.value = threadId
+		} else {
+			// last known message in the most recent block store
+			contextMessageId.value = conversationLastMessageId.value
```
DorraJaouad, line 233: Shouldn't it be lastRead ? It can be far away from last known

src/composables/useGetMessages.ts
```diff
@@ -167,33 +196,37 @@ export function useGetMessagesProvider() {
...
 		}
+		stopFetchingOldMessages.value = false
 
-		if (to.query.threadId && from.query.threadId !== to.query.threadId) {
-			// FIXME temporary get thread messages from the start
-			const hasMessageInStore = ('id' in (store.getters.message(to.params.token, to.query.threadId) as ChatMessage | Record<string, never>))
-			if (!hasMessageInStore) {
-				await getMessageContext(to.params.token, +to.query.threadId, +to.query.threadId)
-			}
+		const hasMessageInStore = chatStore.hasMessage(to.params.token, { messageId: contextMessageId.value, threadId })
```
DorraJaouad, line 236: `hasMessage` method doesn't need `threadId` as a second attr IMO. All uses are just looking for existence of message and if it does, it definitely has `threadId`

src/composables/useGetMessages.ts
DorraJaouad: This block logic is redundant in `handleStartGettingMessagesPreconditions`. Either onRouteChange or that method is enough

src/composables/useGetMessages.ts
```diff
@@ -203,31 +236,27 @@ export function useGetMessagesProvider() {
...
 			try {
-				// Start from message hash or unread marker
-				let startingMessageId = focusMessageId !== null ? focusMessageId : conversation.value!.lastReadMessage
-				// Check if thread is initially opened
-				if (threadId.value) {
-					// FIXME temporary get thread messages from the start
-					startingMessageId = threadId.value
+				if (!contextMessageId.value) {
+					throw new Error(`[DEBUG] spreed: context message ID is ${contextMessageId.value}`)
 				}
```
DorraJaouad, line 268: why would you log the id when `!contextMessageId.value` ?

src/composables/useGetMessages.ts
```diff
@@ -78,13 +84,36 @@ export function useGetMessagesProvider() {
...
+		}
+		const firstKnownMessageId = chatStore.getFirstKnownId(currentToken.value, { messageId: contextMessageId.value, threadId: contextThreadId.value })
+		const firstKnownMessage = store.getters.message(currentToken.value, firstKnownMessageId) as ChatMessage | undefined
+		if (!firstKnownMessage) {
+			// Do not block attempts to fetch history inside each block
+			return false
+		}
+		return firstKnownMessage.messageType === MESSAGE.TYPE.SYSTEM
+			&& ['conversation_created', 'history_cleared'].includes(firstKnownMessage.systemMessage)
+	})
```
DorraJaouad, line 106: If it is thread view, we need to check if first known has threadId as an id ;)

src/composables/useGetMessages.ts
```diff
@@ -167,33 +196,37 @@ export function useGetMessagesProvider() {
...
-			return
+			contextMessageId.value = focusMessageId
+		} else if (threadId && from.query.threadId !== to.query.threadId) {
+			// FIXME temporary get thread messages from the start
+			contextMessageId.value = threadId
+		} else {
+			// last known message in the most recent block store
+			contextMessageId.value = conversationLastMessageId.value
 		}
+		stopFetchingOldMessages.value = false
```
DorraJaouad: This is not necessary, this part is reached when there is a change in the same conversation. Once we have set that flag, we are certain there are no more old messages to fetch

src/composables/useGetMessages.ts
```diff
@@ -78,13 +84,36 @@ export function useGetMessagesProvider() {
...
+		return chatStore.getLastKnownId(currentToken.value, { threadId: contextThreadId.value })
+	})
+
+	const isChatEndReached = computed(() => {
+		const conversation = store.getters.conversation(currentToken.value) as Conversation | undefined
+		if (!conversation || !conversation.lastMessage) {
+			// Do not block attempts to fetch new messages inside each block
+			return false
+		}
+		return chatStore.getLastKnownId(currentToken.value, { messageId: contextMessageId.value, threadId: contextThreadId.value }) >= conversationLastMessageId.value
```
DorraJaouad: Implementation is not accurate. 
It should be truthy on two cases: 
- If it is thread view,  thread.last.id is in the most recent block
- Otherwise, *absolute* last known (last id in the most recent block) is equal to conversation last message id

src/components/MessagesList/MessagesList.vue
```diff
@@ -714,6 +719,22 @@ export default {
...
+				}
+				this.displayMessagesLoader = true
+				await this.getNewMessages(this.token, false)
+				this.displayMessagesLoader = false
+				if (this.$refs.scroller.scrollHeight !== scrollHeight) {
+					// scroll to previous position + added height
+					this.$refs.scroller.scrollTo({
+						top: scrollTop,
+					})
+				}
```
DorraJaouad, line 731: Image

That is not needed

src/components/MessagesList/MessagesList.vue
DorraJaouad, line 1: Don't forget to add `NcLoadingIcon` to the bottom when it is loading newer messages, or dynamically change the position of the already existing one (change its absolute position based on scrolling direction)

src/components/MessagesList/MessagesList.vue
```diff
@@ -138,12 +138,16 @@ export default {
 
 	setup(props) {
 		const {
+			contextMessageId,
 			loadingOldMessages,
+			loadingNewMessages,
```
DorraJaouad, line 155: You can susbtitute `displayMessagesLoader` with this. I haven't checked it throughly but from first sight, it seems doable

DorraJaouad (commented):
I haven't tested it yet

src/stores/chat.ts
Antreesy: This would be covered by 800px threshold (to fetch old/new messages). Should be good enough?

src/composables/useGetMessages.ts
Antreesy: Was thinking about it too. Telegram has two threshold when scrolling to bottom, which I like:
- last read message
- last message in the chat

src/composables/useGetMessages.ts
Antreesy: When we separate chatBlocks and threadBlocks, it should make sense.
You might have a thread message fetched already, but without thread context around you can't open a thread with this message. Then `hasMessage` should return false, and we'd need to fetch from `context/message_id?thread_id`

src/composables/useGetMessages.ts
Antreesy: handleStartGettingMessagesPreconditions is fired when switching between conversations
onRouteChange  - when navigating within the same conversation

so they're not overlapping

src/composables/useGetMessages.ts
Antreesy: we can get the old context [1-100] and reach the beginning there with 304 response, but when we return to current context [400-500], this flag would block us from getting old messages between 100-400

Although we can completely drop `stopFetchingOldMessages` flag

src/composables/useGetMessages.ts
Antreesy: Agreed with 1st
2nd is the same as `stopFetchingOldMessages`, we need to know it for currently opened context to fetch back and forth

src/components/MessagesList/MessagesList.vue
Antreesy: image

src/components/MessagesList/MessagesList.vue
Antreesy: Actually needed, as scroller content is snapped to one of borders, if height is changed:
- if scrollTop is 0, it will remain scrollTop,
- if scrolled to bottom, it will remain at bottom,

that's why we do it when scrolling up

src/stores/chat.ts
DorraJaouad: That means we are relying on `handleScroll` being called initially. Maybe it is fine

src/composables/useGetMessages.ts
```diff
@@ -303,11 +295,10 @@ export function useGetMessagesProvider() {
 				// Empty chat, no messages to load
 				debugTimer.end(`${token} | get context`, 'status 304')
 				store.dispatch('loadedMessagesOfConversation', { token })
-
-				stopFetchingOldMessages.value = true
```
DorraJaouad: How are we going to stop useless fetching requests when we reach the top in expiring conversation? We don't have the typical system messages? We need sthg to mark this result

Antreesy: bringing back. once per context is the best we could achieve without storing and invalidating it per block

src/composables/useGetMessages.ts
```diff
@@ -165,35 +212,41 @@ export function useGetMessagesProvider() {
...
+		// Reset blocker for fetching old messages
+		stopFetchingOldMessages.value = false
 		if (from.name !== 'conversation' || to.name !== 'conversation'
 			|| from.params.token !== to.params.token || typeof to.params.token !== 'string') {
 			// Only handle route changes within the same conversation
 			return
 		}
 
 		const focusMessageId = getMessageIdFromHash(to.hash)
+		const threadId = to.query.threadId ? +to.query.threadId : 0
```
```suggestion
		const threadId = +(to.query.threadId ?? 0)
```

src/composables/useGetMessages.ts
```diff
@@ -165,35 +212,41 @@ export function useGetMessagesProvider() {
...
-			// need some delay (next tick is too short) to be able to run
-			// after the browser's native "scroll to anchor" from the hash
-			window.setTimeout(() => {
-				EventBus.emit('focus-message', focusMessageId)
-			}, 2)
-			return
+			contextMessageId.value = focusMessageId
+		} else if (threadId && from.query.threadId !== to.query.threadId) {
+			// FIXME temporary get thread messages from the start
+			contextMessageId.value = threadId
```
DorraJaouad: You should focus on thread.last instead

src/composables/useGetMessages.ts
```diff
@@ -165,35 +212,41 @@ export function useGetMessagesProvider() {
...
+		} else if (conversation.value?.lastReadMessage && conversation.value.lastReadMessage > contextMessageId.value) {
+			// focus last read message first
+			contextMessageId.value = conversation.value.lastReadMessage
```
DorraJaouad, line 230: I would place this above the thread check condition block

src/composables/useGetMessages.ts
```diff
@@ -165,35 +212,41 @@ export function useGetMessagesProvider() {
...
-			// after the browser's native "scroll to anchor" from the hash
-			window.setTimeout(() => {
-				EventBus.emit('focus-message', focusMessageId)
-			}, 2)
-			return
+			contextMessageId.value = focusMessageId
+		} else if (threadId && from.query.threadId !== to.query.threadId) {
+			// FIXME temporary get thread messages from the start
+			contextMessageId.value = threadId
+		} else if (conversation.value?.lastReadMessage && conversation.value.lastReadMessage > contextMessageId.value) {
```
DorraJaouad, line 228: At this point of condition tree, what could be `contextMessageId` if it is not hash id nor in thread?

src/composables/useGetMessages.ts
DorraJaouad: Ehm

src/composables/useGetMessages.ts
DorraJaouad: > onRouteChange - when navigating within the same conversation

OnRouteChange is a listener to `route-change` event which is triggered on every route change except when conversation is not successfully fetched in `fetchSingleConversation`

DorraJaouad (commented):
Tested, it works correctly (test was not thread focused). Just need to address the above comments and it is good to go

src/composables/useGetMessages.ts
Antreesy: Yes, but we're currently blocking it with `from.name !== 'conversation' || to.name !== 'conversation' || from.params.token !== to.params.token`

We can take a look again into unifying onRouteChange and handleStartGettingMessagesPreconditions or extracting common logic later.
At the moment:
- `handleStartGettingMessagesPreconditions` checks if user 'canGetMessages';
- `onRouteChange` triggers only afterwards under separate conditions, assuming the check above is true

src/composables/useGetMessages.ts
Antreesy: It would be a previous value, so either:
- from `handleStartGettingMessagesPreconditions` - hash id, thread id, or last read
- from previous `onRouteChange` - "from" hash id, "from" thread id, (other options would be last read or last fetched, but they shouldn't trigger router to change)
so we're switching to most recent context in any case

src/composables/useGetMessages.ts
Antreesy: Dropping it, and will focus only hash id, last read or last known

If threadId is given, it will find a message or fetch context around it

src/composables/useGetMessages.ts
Antreesy: `contextMessageId` can still be 0 as per previous unchanged logic. Keeping it as-is, adding a comment

DorraJaouad (approved):

Issue #6212, opened 2021-09-08
Suggest short break after 60 minutes during call


tobiasKaminsky:
To help users maintain a healthy digital well-being/work environment, I think it would be great if after 60minutes an automated message pops up in chat, suggesting to do a short break.
Chat is less intrusive, maybe with a nice coffee mug and a "funny" text? @jancborchardt 

---

- [x] https://github.com/nextcloud/spreed/issues/9947
- [x] https://github.com/nextcloud/talk-android/issues/3175
- [x] https://github.com/nextcloud/talk-ios/issues/1298

Spartachetto:
The idea is really interesting. 
I am sorry to propose options, but you could consider to have a "funny" text good for more informal situations and some citation to health guidelines for more formal ones.

Consider also that **unluckily**  attended several meetings in this period in which such message would've been seen as annoying...

nickvergessen:
Same here. Maybe we can have a look when the different call types are a bit better seperated

Minnietheandroid:
Never been crazy about suggestions from software myself, funny or otherwise:)
If you go ahead with this, please make it 'opt in' or something we can disable easily.

nickvergessen:
Of course we are not going to force this up on everyone.

nickvergessen:
Related #1218

jancborchardt:
Maybe just a small popover on the call time display:

> Note the call is running since 1 hour already.

Nothing else suggesting a break or wrap up, just pointing out the time.

nickvergessen:
@jancborchardt just for the record, tjhe popup does not appear in call recordings, right?

jancborchardt:
@nickvergessen yeah it's fine if it doesn't display in the recording. E.g. a connection warning would also not display, right?

nickvergessen:
All done 🎉

PR #10086, opened 2023-08-01
feat(CallTime)- add one hour hint during the call

### LINKED ISSUE #9947: Show hint when call is running since 1 hour
Overview at https://github.com/nextcloud/spreed/issues/6212

* Design: https://github.com/nextcloud/spreed/issues/6212#issuecomment-1623340533
  Maybe just a small popover on the call time display:
  > Note the call is running since 1 hour already.

  Nothing else suggesting a break or wrap up, just pointing out the time.

DorraJaouad:
### ☑️ Resolves

* Fix #9947 

### Description

The Hint text 
[ [design](https://github.com/nextcloud/spreed/issues/6212#issuecomment-1623340533)]
> Note the call is running since 1 hour already. 

* The Hint gets shown at 1:00:00, and the popover stays open for 10s if the user didn't trigger outside focus.
* After that, The Hint remains in the popover and it is accessed by clicking on the call time.

* If the User is a moderator, the hint is added to the recording buttons popover.

### 🖼️ Screenshots

This is a recording for a hint faked at 2:20 :

https://github.com/nextcloud/spreed/assets/84044328/8ee08983-3f84-4d30-96b5-9035907d8c0f

The moderator's popover ( e.g in case the recording is starting)

image

### 🚧 Tasks

- [ ] To test : 
If you don't want to wait for one hour, change this line and you can only wait for 20 sec and it can be repeated with 1 min interval.
```
      this.untilOneHourHint = (1000 * 20) - this.timer + 1
```
- [ ] code review
- [ ] Visual review

### 🏁 Checklist

### DIFF
--- a/src/components/TopBar/CallTime.vue
+++ b/src/components/TopBar/CallTime.vue
@@ -20,17 +20,17 @@
 
 <template>
 	<NcPopover class="call-time"
-		close-after-click="true"
 		:menu-title="callTime"
 		:shown.sync="showPopover"
+		:focus-trap="isShowRecordingStatus"
 		:triggers="[]"
 		:container="container">
 		<template #trigger>
-			<NcButton :disabled="!isShowRecordingStatus || !isModerator"
+			<NcButton :disabled="isButtonDisabled"
 				:wide="true"
 				:title="recordingButtonTitle"
 				type="tertiary"
-				@click="showPopover = true">
+				@click="showPopover = !showPopover">
 				<template v-if="isShowRecordingStatus" #icon>
 					<RecordCircle v-if="isRecording"
 						:size="20"
@@ -42,24 +42,34 @@
 				{{ formattedTime }}
 			</NcButton>
 		</template>
-		<NcButton v-if="isStartingRecording"
-			type="tertiary-no-background"
-			:wide="true"
-			@click="stopRecording">
-			<template #icon>
-				<NcLoadingIcon :size="20" />
-			</template>
-			{{ t('spreed', 'Cancel recording start') }}
-		</NcButton>
-		<NcButton v-else
-			type="tertiary-no-background"
-			:wide="true"
-			@click="stopRecording">
-			<template #icon>
-				<StopIcon :size="20" />
-			</template>
-			{{ t('spreed', 'Stop recording') }}
-		</NcButton>
+
+		<!--one hour hint-->
+		<span v-if="isCallDurationHintShown" class="call-duration-hint">
+			{{ t('spreed', 'The call has been running for one hour.') }}
+		</span>
+
+		<!--Moderator's buttons-->
+		<template v-if="isShowRecordingStatus">
+			<hr v-if="isCallDurationHintShown" class="solid">
+			<NcButton v-if="isStartingRecording"
+				type="tertiary-no-background"
+				:wide="true"
+				@click="stopRecording">
+				<template #icon>
+					<NcLoadingIcon :size="20" />
+				</template>
+				{{ t('spreed', 'Cancel recording start') }}
+			</NcButton>
+			<NcButton v-else
+				type="tertiary-no-background"
+				:wide="true"
+				@click="stopRecording">
+				<template #icon>
+					<StopIcon :size="20" />
+				</template>
+				{{ t('spreed', 'Stop recording') }}
+			</NcButton>
+		</template>
 	</NcPopover>
 </template>
 
@@ -101,7 +111,9 @@ export default {
 		return {
 			callTime: undefined,
 			showPopover: false,
+			isCallDurationHintShown: false,
 			timer: null,
+			untilCallDurationHintShown: null,
 		}
 	},
 
@@ -165,6 +177,10 @@ export default {
 			return this.isStartingRecording || this.isRecording
 		},
 
+		isButtonDisabled() {
+			return !this.isShowRecordingStatus && !this.isCallDurationHintShown
+		},
+
 		recordingButtonTitle() {
 			if (this.isStartingRecording) {
 				return t('spreed', 'Starting the recording')
@@ -173,6 +189,17 @@ export default {
 			}
 
 			return ''
+		},
+	},
+
+	watch: {
+		callTime(value) {
+			if (value && !this.untilCallDurationHintShown) {
+				this.untilCallDurationHintShown = (1000 * 60 * 60) - value + 1000
+				setTimeout(() => {
+					this.showCallDurationHint()
+				}, this.untilCallDurationHintShown)
+			}
 		},
 	},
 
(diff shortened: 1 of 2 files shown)


DorraJaouad:
@jancborchardt , how about 

> The call has been running for one hour.

src/components/TopBar/CallTime.vue
```diff
@@ -143,6 +150,10 @@ export default {
 			return hours + ' : ' + minutes + ' : ' + seconds
 		},
 
+		showOneHourHint() {
+			return Math.floor((this.callTime / (1000 * 60 * 60)) % 24) >= 1
```
Antreesy, line 154: Don't know how good it is performance-wise
This value should be re-calculated every time `this.callTime` is updated, so every second?

I'd experiment here a bit, if it's posiible in `mounted` hook get a diff between 1 hour and currently elapsed time, ans use `setTimeout` with this amount of time

src/components/TopBar/CallTime.vue
```diff
@@ -176,6 +187,27 @@ export default {
 		},
 	},
 
+	watch: {
+		showOneHourHint(value) {
+			if (value) {
```
Antreesy, line 192: This also will be change only once, so a method fits here better, I suppose

jancborchardt:
@DorraJaouad your text auggestion sounds better indeed!

And it looks good, except it could use some more padding to the top and sides so it's consistent on all sides.

src/components/TopBar/CallTime.vue
```diff
@@ -204,7 +236,10 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-
+.one-hour-hint{
+	display : flex ;
+	padding : 6px;
```
jancborchardt, line 241: Code style here:
- No space before ":"
- No space before ";"

jancborchardt: And padding should be set based on multiples of --default-grid-baseline.

Also keep in mind sides might need more than top and bottom to look visually consistent. :)

jancborchardt (changes requested):
See code comments

src/components/TopBar/CallTime.vue
```diff
@@ -179,6 +187,10 @@ export default {
 	mounted() {
 		// Start the timer when mounted
 		this.timer = setInterval(this.computeElapsedTime, 1000)
+		this.untilOneHourHint = (1000 * 60 * 60) - this.timer + 1
```
Antreesy, line 190: If I remember correctly, `this.timer` stores internal id of the interval, to be able to reset it (not the result)
Can we maybe use `this.callTime` here?

DorraJaouad: `this.callTime` is still undefined, it gets its value after `this.computeElapsedTime` which is called in a delay of 1s.

Antreesy: this.timer is also can't be used here:
image
image

Antreesy: So, we know the time of call start from `conversation.callStartTime`, which is fetched only every 30 seconds
BUT: this value is equal to `message.timestamp` of system message with systemMessage: `call_started`

So either we could patch a conversation in conversationStore with this value, or return back to watcher, and check if `callStart !== 0 && !timeout`, to start it only once

src/components/TopBar/CallTime.vue
```diff
@@ -20,17 +20,17 @@
 
 <template>
 	<NcPopover class="call-time"
-		close-after-click="true"
 		:menu-title="callTime"
 		:shown.sync="showPopover"
+		:focus-trap=" isShowRecordingStatus "
```
Antreesy, line 25:
```suggestion
		:focus-trap="isShowRecordingStatus"
```

src/components/TopBar/CallTime.vue
```diff
@@ -20,17 +20,17 @@
...
 		:triggers="[]"
 		:container="container">
 		<template #trigger>
-			<NcButton :disabled="!isShowRecordingStatus || !isModerator"
+			<NcButton :disabled="(!isShowRecordingStatus || !isModerator ) && !isCallDurationHintShown"
```
Antreesy, line 29:
```suggestion
			<NcButton :disabled="isButtonDisabled"
```
let's move it to computed as well

src/components/TopBar/CallTime.vue
```diff
@@ -42,24 +42,32 @@
...
-			:wide="true"
-			@click="stopRecording">
-			<template #icon>
-				<StopIcon :size="20" />
-			</template>
-			{{ t('spreed', 'Stop recording') }}
-		</NcButton>
+
+		<!--one hour hint-->
+		<span v-if="isCallDurationHintShown" class="one-hour-hint"> {{ t('spreed', 'The call has been running for one hour.') }} </span>
```
Antreesy, line 47:
```suggestion
		<span v-if="isCallDurationHintShown" class="one-hour-hint">
		  {{ t('spreed', 'The call has been running for one hour.') }}
		</span>
```

src/components/TopBar/CallTime.vue
```diff
@@ -42,24 +42,32 @@
...
+			<NcButton v-if="isStartingRecording"
+				type="tertiary-no-background"
+				:wide="true"
+				@click="stopRecording">
+				<template #icon>
+					<NcLoadingIcon :size="20" />
+				</template>
+				{{ t('spreed', 'Cancel recording start') }}
+			</NcButton>
+			<NcButton v-else-if="isRecording"
```
Antreesy, line 61:
```suggestion
		<NcButton v-else
```
Rollback

src/components/TopBar/CallTime.vue
```diff
@@ -185,6 +195,18 @@ export default {
 		clearInterval(this.timer)
 	},
 
+	watch: {
```
Antreesy, line 198: eslint warning here

src/components/TopBar/CallTime.vue
```diff
@@ -185,6 +195,18 @@ export default {
...
+
+		callTime(value) {
+			if (value && !this.untilCallDurationHintShown) {
+				this.untilCallDurationHintShown = (1000 * 35) - value + 1
```
Antreesy, line 202:
```suggestion
				const untilCallDurationHintShown = (1000 * 60 * 60) - value + 1
```
`const` is not working here?

src/components/TopBar/CallTime.vue
```diff
@@ -199,12 +221,37 @@ export default {
 			}
 			this.callTime = new Date() - this.callStart
 		},
+
+		showOneHourHint() {
```
Antreesy, line 225:
```suggestion
		showCallDurationHint() {
```

src/store/messagesStore.js
```diff
@@ -501,6 +501,20 @@ const actions = {
 			})
 		}
 
+		if (message.systemMessage === 'call_started') {
```
Antreesy, line 504:
```suggestion
		if (message.systemMessage === 'call_started' && context.getters.conversation(token).lastActivity < message.timestamp) {
```

We need to check in addition if we're processing a new message now, or an old one, to not trigger a call UI by coincidence
Not sure about the correct line patch, but should be understndable

src/store/messagesStore.js
```diff
@@ -501,6 +501,20 @@ const actions = {
...
+			context.dispatch('setConversationProperties', {
+				token: message.token,
+				properties: { callStartTime: message.timestamp },
+			})
+		}
+
+		if (message.systemMessage === 'call_ended_everyone') {
```
Antreesy, line 511:
```suggestion
		if ((message.systemMessage === 'call_ended' || message.systemMessage === 'call_ended_everyone')
		  && context.getters.conversation(token).lastActivity < message.timestamp) {
```
Should work for both cases

Antreesy (approved):
Tested on web client, works good!

Also call timer is much more responsive now

PR #10088, opened 2023-08-02
fix(recording): Fix transcription opt-in

nickvergessen:
### ☑️ Resolves
Before:
```
var_dump(!'no' === 'yes');
bool(false)

var_dump(!'yes' === 'yes');
bool(false)
```

### 🏁 Checklist

### DIFF
--- a/lib/Service/RecordingService.php
+++ b/lib/Service/RecordingService.php
@@ -147,7 +147,7 @@ class RecordingService {
 			throw new InvalidArgumentException('owner_permission');
 		}
 
-		if (!$this->serverConfig->getAppValue('spreed', 'call_recording_transcription', 'no') === 'yes') {
+		if ($this->serverConfig->getAppValue('spreed', 'call_recording_transcription', 'no') === 'no') {
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


SystemKeeper (approved):

lib/Service/RecordingService.php
```diff
@@ -147,7 +147,7 @@ public function store(Room $room, string $owner, array $file): void {
 			throw new InvalidArgumentException('owner_permission');
 		}
 
-		if (!$this->serverConfig->getAppValue('spreed', 'call_recording_transcription', 'no') === 'yes') {
+		if ($this->serverConfig->getAppValue('spreed', 'call_recording_transcription', 'no') === 'yes') {
```
Altahrim, line 150:
```suggestion
		if ($this->serverConfig->getAppValue('spreed', 'call_recording_transcription', 'no') !== 'yes') {
```
It fixs the bug of priority, but changes the meaning.
We want to return only if config is different from `yes`, no?

nickvergessen: thanks, fixed by comparing to `no`

Altahrim (approved):

PR #10105, opened 2023-08-04
feat(Message) - 🔔 add frontend support for message reminders

### LINKED ISSUE #10076: Notify me later about this message - Frontend
- Overview ticket at https://github.com/nextcloud/spreed/issues/10075

Antreesy:
### ☑️ Resolves

* Fix #10076
* Add an item into MessageButtonsBar (with submenu)

### 🖼️ Screenshots
Main menu | Sub menu
--- | ---
image | image

https://github.com/nextcloud/spreed/assets/93392545/b5caa45d-d950-41ad-be25-91e61abec5f7

### 🚧 Tasks

- [ ] Manual testing
- [ ] Design review
- [ ] Code review

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
@@ -144,6 +144,15 @@
 						{{ t('spreed', 'Back') }}
 					</NcActionButton>
 
+					<NcActionButton v-if="currentReminder"
+						close-after-click
+						@click.stop="removeReminder">
+						<template #icon>
+							<CloseCircleOutline :size="20" />
+						</template>
+						{{ clearReminderLabel }}
+					</NcActionButton>
+
 					<NcActionSeparator />
 
 					<NcActionButton v-for="option in reminderOptions"
@@ -227,6 +236,7 @@ import CalendarClock from 'vue-material-design-icons/CalendarClock.vue'
 import Check from 'vue-material-design-icons/Check.vue'
 import CheckAll from 'vue-material-design-icons/CheckAll.vue'
 import ClockOutline from 'vue-material-design-icons/ClockOutline.vue'
+import CloseCircleOutline from 'vue-material-design-icons/CloseCircleOutline.vue'
 import EmoticonOutline from 'vue-material-design-icons/EmoticonOutline.vue'
 import EyeOffOutline from 'vue-material-design-icons/EyeOffOutline.vue'
 import File from 'vue-material-design-icons/File.vue'
@@ -251,7 +261,7 @@ import Forwarder from './Forwarder.vue'
 
 import { PARTICIPANT, CONVERSATION, ATTENDEE } from '../../../../../constants.js'
 import { EventBus } from '../../../../../services/EventBus.js'
-import { setMessageReminder } from '../../../../../services/remindersService.js'
+import { getMessageReminder, removeMessageReminder, setMessageReminder } from '../../../../../services/remindersService.js'
 import { copyConversationLinkToClipboard } from '../../../../../services/urlService.js'
 
 // Keep version in sync with @nextcloud/vue in case of issues
@@ -275,6 +285,7 @@ export default {
 		AlarmIcon,
 		ArrowLeft,
 		CalendarClock,
+		CloseCircleOutline,
 		Check,
 		CheckAll,
 		ClockOutline,
@@ -449,6 +460,7 @@ export default {
 		return {
 			frequentlyUsedEmojis: [],
 			submenu: null,
+			currentReminder: null,
 			customReminderDateTime: new Date(moment().add(2, 'hours').minute(0).second(0).valueOf()),
 		}
 	},
@@ -585,6 +597,18 @@ export default {
 				},
 			].filter(option => option.timestamp !== null)
 		},
+
+		clearReminderLabel() {
+			return t('spreed', 'Clear reminder – {timeLocale}', { timeLocale: moment(this.currentReminder.timestamp * 1000).format('ddd LT') })
+		},
+	},
+
+	watch: {
+		submenu(value) {
+			if (value === 'reminder') {
+				this.getReminder()
+			}
+		},
 	},
 
 	methods: {
@@ -702,6 +726,25 @@ export default {
 
 		getTimestamp(momentObject) {
 			return momentObject?.minute(0).second(0).millisecond(0).valueOf() || null
+		},
+
+		async getReminder() {
+			try {
+				const response = await getMessageReminder(this.token, this.id)
+				this.currentReminder = response.data.ocs.data
+			} catch (error) {
+				console.debug(error)
+			}
+		},
+
+		async removeReminder() {
+			try {
+				await removeMessageReminder(this.token, this.id)
+				showSuccess(t('spreed', 'A reminder was successfully removed'))
+			} catch (error) {
+				console.error(error)
+				showError(t('spreed', 'Error occurred when removing a reminder'))
+			}
 		},
 
 		async setReminder(timestamp) {
--- a/src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
@@ -77,16 +77,16 @@
 					<NcActionSeparator />
 					<NcActionButton v-if="isPrivateReplyable"
 						icon="icon-user"
-						:close-after-click="true"
+						close-after-click
 						@click.stop="handlePrivateReply">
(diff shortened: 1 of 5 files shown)


src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBarSetReminder.vue
```diff
@@ -0,0 +1,176 @@
...
+			}
+		},
+	},
+}
+</script>
+
+<style lang="scss" scoped>
+.popover_nested-list {
+	:deep(.action-button::after) {
+		content: "\FF1E";
```
nickvergessen, line 166: The arrow right looks kind of weird to me. Not sure we need it. Also the files app has nothing:
https://github.com/nextcloud/server/issues/39181#issuecomment-1664832347

That being said, it looks like files is replacing the popup or something, instead of adding another one to the right.
But not into the details there

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBarSetReminder.vue
```diff
@@ -0,0 +1,176 @@
...
+
+	methods: {
+		getTimeStamp(momentObject) {
+			return momentObject?.minute(0).second(0).millisecond(0).valueOf() || null
+		},
+
+		async setReminder(timestamp) {
+			try {
+				await setMessageReminder(this.token, this.id, timestamp / 1000)
+				showSuccess(t('spreed', 'A reminder was successfully created'))
```
nickvergessen, line 154: Should we add the date/time to the success message?

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBarSetReminder.vue
nickvergessen: Looks good enough like that, but opening to the right could be problematic?

Edit: ah seems to open to the left in the call case, but yeah the arrow would be wrong at that moment, so let's not have it for now:
grafik

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
```diff
@@ -64,6 +64,18 @@
 					</template>
 					{{ messageDateTime }}
 				</NcActionButton>
+
+				<MessageButtonsBarSetReminder :id="id" :token="token">
+					<template #trigger>
+						<NcActionButton type="tertiary">
+							<template #icon>
+								<ClockOutline :size="20" />
```
nickvergessen, line 72: Nimisha asked for AlarmClock to be used iirc.

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBarSetReminder.vue
ShGKme: What about icon from `var(--icon-triangle-e-light)`?

Antreesy: > What about icon from `var(--icon-triangle-e-light)`?

image

ShGKme: > Let's just not use any icon for now, because it could show into the wrong direction (as outlined above)

It seems to me to be a "default" behavior for multi-level context menus everywhere.

image

image

DorraJaouad:
Should the chosen reminder be highlighted and/or marked on the message (e.g.: by an alarm clock next to the time)?

nickvergessen:
We can not read existing reminders for now

Antreesy:
> Should the chosen reminder be highlighted and/or marked on the message (e.g.: by an alarm clock next to the time)?

it's not possible atm to get reminder / reminders for the specific messages, so we can't track it

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBarSetReminder.vue
```diff
@@ -0,0 +1,244 @@
...
+			let parent = this.$parent
+			while (parent) {
+				if (parent.$options.name === 'NcActions') {
+					parent.closeMenu(false)
+					return
+				}
+				parent = parent.$parent
+			}
+		},
+	},
```
ShGKme, line 208: I'd highly prefer to avoid this way to close the menu...
- `closeMenu` is not mentioned as a public method of `NcActions`, I'd consider it to be an internal method
- A proper way to control `NcActions` open state is the `open` prop
- Iterating parents is also not a vue way to communicate between component 
- And `$parent` is available only by internal undocumented functions and properties in Composition API, it is not a recommended way to work with components 

If we do want to close the menu here, we can provide a method from `NcActions`'s parent component.

ShGKme: And it doesn't work for me...

ShGKme: Do we need to close the menu after setting a custom reminder? It is closed on hover loose. And if we do, should be also close on other options?

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBarSetReminder.vue
```diff
@@ -0,0 +1,244 @@
...
+		},
+
+		/**
+		 * The message id.
+		 */
+		id: {
+			type: [String, Number],
+			required: true,
+		},
+	},
```
ShGKme, line 106: `id` could also be HTML attribute. What about renaming it to `messageId`?

marcoambrosini (commented):
Nice addition :) 

I think that opening the date picker from within a tooltip looks bad, is unreliable and too complex. I would just add an action button that says `custom` and open a small dialog with the date picker.

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBarSetReminder.vue
marcoambrosini: IMO:
Close on set: yes 
Closed on hover loose: no

ShGKme (approved):
🔔

DorraJaouad (approved):
Tested and all good 🦅! 

In my opinion, I would recommend closing on clicking and not on hover loose for the non-custom reminders unless we want to encourage the user to set multiple reminders.

Antreesy:
Appearance was aligned with https://github.com/nextcloud/server/issues/39181#issuecomment-1669394720

ShGKme (approved):
Tested after menu design update, everything seems to work fine, looks fine code-wise.

API request is made in a correct local timezone on both presets and custom.

nickvergessen (changes requested):

DorraJaouad (approved):

marcoambrosini:
To elaborate on [my point above](https://github.com/nextcloud/spreed/pull/10105#pullrequestreview-1564772478):

This is a pattern that we use in several places in Nextcloud that I think we should break: **creating baby forms within action menus**. We shouldn't include forms, especially if they involve multi-selects and pickers, into tiny action menus. 

Having a text input with an inline submit button is **probably** ok but far from ideal, but in this case we have a baby form with a submit button that looks exactly like an actionbutton, creating visual confusion and usability issues. For example here  there's no way to guarantee that the picker itself won't block the submit button below. (see screenshot)

This came up in a recent design call and there was some consensus on that we should probably even deprecate these complex action input elements. cc @nextcloud/designers 

Screenshot 2023-08-09 at 17 38 06

nickvergessen:
> This is a pattern that we use in several places in Nextcloud that I think we should break: creating baby forms within action menus. We shouldn't include forms, especially if they involve multi-selects and pickers, into tiny action menus.
>
> Having a text input with an inline submit button is probably ok but far from ideal, but in this case we have a baby form with a submit button that looks exactly like an actionbutton, creating visual confusion and usability issues. For example here there's no way to guarantee that the picker itself won't block the submit button below. (see screenshot)

Well that is the result of yesterdays call was. we have less than 24h left to merge and backport the feature in 2 apps and server, so I guess miniform it is for 27.1

nickvergessen:
I don't see a request to https://nextcloud-talk.readthedocs.io/en/latest/chat/#get-reminder-for-chat-message when opening the respective submenu and neither a "Clear reminder" option for the currently set reminder?

Ref from iOS https://github.com/nextcloud/talk-ios/pull/1320#issuecomment-1670757931

Antreesy:
> I don't see a request

Just rebased branch and added it, please take a look

Issue #10094, opened 2023-08-02
🐣  Bots part 2


nickvergessen:
- [x] Commands (currently marked as Beta) should be marked as Deprecated #10151
- [x] Allow bots to react #10138 
- [x] Installed bots should be listed in the admin settings similarly like commands are currently
  - [x] Additionally with a link to documentation how to write them
  - [x] And a link to the list of available bots (currently planned to be hosted in the docs of Talk for the time being)

PR #10138, opened 2023-08-07
feat(bot): Allow reactions by bots

nickvergessen:
### ☑️ Resolves

* Task from https://github.com/nextcloud/spreed/issues/10094#issue-1833746694

### 🏁 Checklist

### DIFF
--- a/lib/Controller/BotController.php
+++ b/lib/Controller/BotController.php
@@ -27,6 +27,10 @@ declare(strict_types=1);
 namespace OCA\Talk\Controller;
 
 use OCA\Talk\Chat\ChatManager;
+use OCA\Talk\Chat\ReactionManager;
+use OCA\Talk\Exceptions\ReactionAlreadyExistsException;
+use OCA\Talk\Exceptions\ReactionNotSupportedException;
+use OCA\Talk\Exceptions\ReactionOutOfContextException;
 use OCA\Talk\Exceptions\UnauthorizedException;
 use OCA\Talk\Manager;
 use OCA\Talk\Middleware\Attribute\RequireLoggedInModeratorParticipant;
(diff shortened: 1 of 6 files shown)


lib/Controller/BotController.php
```diff
@@ -107,16 +100,40 @@ public function sendMessage(string $token, string $message, string $referenceId
...
+	#[PublicPage]
+	public function sendMessage(string $token, string $message, string $referenceId = '', int $replyTo = 0, bool $silent = false): DataResponse {
+		try {
+			$bot = $this->getBotFromHeaders($token, $message);
+		} catch (\InvalidArgumentException $e) {
+			$response = new DataResponse([], $e->getCode());
+			if ($e->getCode() === Http::STATUS_UNAUTHORIZED) {
+				$response->throttle(['action' => 'bot']);
+			}
 			return $response;
```
Altahrim, line 137: I am looking for a way to avoid repetition here and keep the same logic in one place. Maybe a wrapper like this:
```php
private function handleInvalidArgumentException(\InvalidArgumentException $e) {
	$response = new DataResponse([], $e->getCode());
	if ($e->getCode() === Http::STATUS_UNAUTHORIZED) {
		$response->throttle(['action' => 'bot']);
	}
	return $response;
}
```

Block from l. 130 to l. 138 is used several times. I wonder we could extract it in a method, but then we will have something like: 
```php
$bot = $this->loadBot($token, $message);
if ($bot instanceof DataResponse) {
	return $bot;
}
```
And I don't like it ^^

lib/Controller/BotController.php
```diff
@@ -149,6 +166,78 @@ public function sendMessage(string $token, string $message, string $referenceId
...
+			$this->reactionManager->deleteReactionMessage(
+				$room,
+				$actorType,
+				$actorId,
+				$messageId,
+				$reaction
+			);
+		} catch (ReactionNotSupportedException | ReactionOutOfContextException | NotFoundException) {
+			return new DataResponse([], Http::STATUS_NOT_FOUND);
+		} catch (\Exception) {
```
Altahrim, line 234: Maybe we should catch `\Throwable` here?

lib/Controller/BotController.php
```diff
@@ -107,16 +100,40 @@ public function sendMessage(string $token, string $message, string $referenceId
...
+				return $botAttempt;
 			} catch (UnauthorizedException) {
 			}
 		}
 
-		if (!$bot instanceof Bot) {
-			$this->logger->debug('No valid Bot entry found');
-			$response = new DataResponse([], Http::STATUS_UNAUTHORIZED);
-			$response->throttle(['action' => 'bot']);
+		$this->logger->debug('No valid Bot entry found');
```
Altahrim, line 108: Still useful?

lib/Controller/BotController.php
nickvergessen: If it's not expected, don't catch it? :P

lib/Controller/BotController.php
nickvergessen: For debugging yes

lib/Controller/BotController.php
Altahrim: Agree, but…
Already a lot of things can be caught with `Exception`. Are all of them expected?

lib/Controller/BotController.php
nickvergessen: yeah, I don't like mixed return types, that's why I went for the throwing instead.

Altahrim (approved):

PR #10146, opened 2023-08-08
feat(chat): Option to get and clear reminders

### LINKED ISSUE #10077: Notify me later about this message - API
- Overview ticket at https://github.com/nextcloud/spreed/issues/10075

nickvergessen:
### ☑️ Resolves

* Fix #10077

### 🚧 Tasks

- [x] Allow updating an existing reminder (before it was triggered)
- [x] API to get existing reminder for a message
- [x] API to delete reminder before the notification triggered

### 🏁 Checklist


SystemKeeper (approved):
Did not look too detailed at the code, just tested it with the iOS PR and worked (Set Reminder, Clear Reminder and the actual push notification)

lib/BackgroundJob/Reminder.php
```diff
@@ -0,0 +1,49 @@
...
+use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\BackgroundJob\TimedJob;
+
+class Reminder extends TimedJob {
+	public function __construct(
+		ITimeFactory $time,
+		protected ReminderService $reminderService,
+	) {
+		parent::__construct($time);
+		// Every time (5 minutes minus 1 second)
```
Altahrim, line 39: Why "minus 1 second"?

I also wonder if we can set the interval to match the next reminder. It would also mean to change the interval if a new reminder is set before… 
It's not in the scope of this pull request of course :) Just a thought

lib/Chat/ChatManager.php
```diff
@@ -730,26 +725,14 @@ public function getMessagesById(Room $chat, array $commentIds): array {
...
-				'commentId' => $comment->getId(),
-			]);
-		$this->notificationManager->markProcessed($notification);
+	/**
+	 * Get messages by ID
+	 *
+	 * @param int[] $commentIds
+	 * @return IComment[] Key is the message id
+	 */
+	public function getMessagesById(array $commentIds): array {
```
Altahrim, line 734: Could also be like this to enforce type: 
```suggestion
	public function getMessagesById(int ...$commentIds): array {
```
If accepted, maybe in a future pull request to uniformize other methods like `getCommentsById` or `getMessagesForRoomId`

lib/Chat/ChatManager.php
```diff
@@ -730,26 +725,14 @@ public function getMessagesById(Room $chat, array $commentIds): array {
...
+		return $this->commentsManager->getCommentsById(array_map('strval', $commentIds));
```
Altahrim, line 735: I wonder if conversion to `string` shouldn't be done at lower level.
It seems we always use `int` everywhere else.

Probably also out of scope of this pull request

lib/Controller/ChatController.php
```diff
@@ -689,37 +692,58 @@ public function deleteMessage(int $messageId): DataResponse {
 	#[RequireModeratorOrNoLobby]
 	#[RequireLoggedInParticipant]
 	#[UserRateLimit(limit: 60, period: 3600)]
-	public function remindLater(int $messageId, int $timestamp): DataResponse {
+	public function setReminder(int $messageId, int $timestamp): DataResponse {
 		try {
-			$message = $this->chatManager->getComment($this->room, (string) $messageId);
+			$this->chatManager->getComment($this->room, (string) $messageId);
```
Altahrim, line 697: No `exists` method?
Depending of the implementation, it could be faster than a `get`

lib/Controller/ChatController.php
```diff
@@ -689,37 +692,58 @@ public function deleteMessage(int $messageId): DataResponse {
...
 	#[RequireModeratorOrNoLobby]
 	#[RequireLoggedInParticipant]
-	public function dismissReminder(int $messageId): DataResponse {
+	public function getReminder(int $messageId): DataResponse {
 		try {
-			$message = $this->chatManager->getComment($this->room, (string) $messageId);
+			$this->chatManager->getComment($this->room, (string) $messageId);
 		} catch (NotFoundException) {
 			return new DataResponse([], Http::STATUS_NOT_FOUND);
 		}
```
Altahrim, line 720: Do we need this part?
If the reminder exists, message should exists too, right?

lib/Controller/ChatController.php
```diff
@@ -689,37 +692,58 @@ public function deleteMessage(int $messageId): DataResponse {
...
+
+	#[NoAdminRequired]
+	#[RequireModeratorOrNoLobby]
+	#[RequireLoggedInParticipant]
+	public function deleteReminder(int $messageId): DataResponse {
+		try {
+			$this->chatManager->getComment($this->room, (string) $messageId);
+		} catch (NotFoundException) {
+			return new DataResponse([], Http::STATUS_NOT_FOUND);
+		}
```
Altahrim, line 741: Same question here

lib/Service/ReminderService.php
```diff
@@ -0,0 +1,132 @@
...
+			return;
+		}
+
+		$shouldFlush = $this->notificationManager->defer();
+
+		$messageIds = [];
+		foreach ($reminders as $reminder) {
+			$messageIds[] = $reminder->getMessageId();
+		}
+		$messageIds = array_unique($messageIds);
```
Altahrim, line 99: We can use `array_map` here:
```php
$messageIds = array_map(
    static fn(Reminder $reminder) => $reminder->getMessageId(), 
    $reminders
);

File: lib/Service/ReminderService.php [hunk 8]
```diff
@@ -0,0 +1,132 @@
...
+				]);
+			$this->notificationManager->notify($notification);
+		}
+
+		if ($shouldFlush) {
+			$this->notificationManager->flush();
+		}
+
+		$this->reminderMapper->deleteExecutedReminders($executeBefore);
+	}
```
Altahrim, line 131: I  think this function should be in a database transaction

Altahrim (commented):
Whoops, didn't saw this one was merged during review…

lib/BackgroundJob/Reminder.php
nickvergessen: > Why "minus 1 second"?

I was thinking about setting it to 1 sec in the beginning. It's weird edgecasing in my head I guess.

If the cron job runs at 8:55AM and the next time 5min later at 9:00AM it would be skipped there, because it's not 300 seconds since the last run. But on our instance that would not be enough as an "empty" run without heavy jobs takes like 5 seconds. So maybe we just set it to a minute.

lib/Chat/ChatManager.php
nickvergessen: commentsManager is from server and it operates based on strings (for legacy reasons, I can tell you the story another day). So when interacting with that we cast to string and towards talks chatManager we always use ints because that is how the DB works.

lib/Service/ReminderService.php
nickvergessen: Why is that?

nickvergessen: Would still happen for the reminders that were sent before the crash, and I would like to avoid a delete per reminder.
So I guess it's a price we have to pay (and since it's only in programmatic cases I'm fine with it).

PR #10167, opened 2023-08-10
Changelog conversation update for 17.1

nickvergessen:
### 🖼️ Screenshots

grafik

### 🏁 Checklist


nickvergessen:
Will summon the backport bot, but we should not merge it yet, so that it is translated for people when we update the beta.

lib/Chat/Changelog/Manager.php
```diff
@@ -69,64 +69,68 @@ public function updateChangelog(string $userId): void {
 	public function getChangelogs(): array {
 		return [
 			$this->l->t(
-				"Welcome to Nextcloud Talk!\n"
+				"# Welcome to Nextcloud Talk!\n"
```
Antreesy, line 72: Not sure we should use `<h1>`, as per CEO / readers / someone else's standards it's required to use no more than 1 header on a web-page

Maybe stick with `##` for big headers, and `###` for subheaders?

nickvergessen:
```suggestion
				"## Welcome to Nextcloud Talk!\n"
```

Antreesy (approved):

PR #10180, opened 2023-08-11
fix(desktop): fix disabling avatar menu for desktop

ShGKme:
### ☑️ Resolves

The author's avatar in the messages list has an avatar menu on the Desktop. It causes infinity requests to loop on click and leads to rate limit.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks

- [x] to `disableMenu` we need to pass `true`, not `false` as disabling value 🙈

### 🏁 Checklist


nickvergessen (approved):
Would be more of a fan fixing it, as I hit it daily.
But yeah at least no more confusion.

nickvergessen:
Backport ?

PR #10186, opened 2023-08-11
Add OpenAPI

provokateurin:
### ☑️ Resolves

Adds all the utilities and annotations required for OpenAPI.

### 🏁 Checklist

### DIFF
--- a/.github/workflows/openapi.yml
+++ b/.github/workflows/openapi.yml
@@ -1,0 +1,52 @@
+name: OpenAPI
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+      - stable*
+
+permissions:
+  contents: read
+
+concurrency:
+  group: openapi-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  openapi:
+    runs-on: ubuntu-latest
+
+    if: ${{ github.repository_owner != 'nextcloud-gmbh' }}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+
+      - name: Set up php
+        uses: shivammathur/setup-php@c5fc0d8281aba02c7fda07d3a70cc5371548067d # v2
+        with:
+          php-version: '8.2'
+          extensions: xml
+          coverage: none
+          ini-file: development
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set up dependencies
+        run: composer i
+
+      - name: Regenerate OpenAPI
+        run: composer run openapi
+
+      - name: Check openapi.json changes
+        run: |
+          bash -c "[[ ! \"`git status --porcelain `\" ]] || (echo 'Please run \"composer run openapi\" and commit the openapi.json, see the section \"Show changes on failure\" for details' && exit 1)"
+
+      - name: Show changes on failure
+        if: failure()
+        run: |
+          git status
+          git --no-pager diff
+          exit 1 # make it red to grab attention
--- a/lib/Controller/RoomController.php
+++ b/lib/Controller/RoomController.php
@@ -64,6 +64,7 @@ use OCA\Talk\Webinary;
 use OCP\App\IAppManager;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\Attribute\BruteForceProtection;
+use OCP\AppFramework\Http\Attribute\IgnoreOpenAPI;
 use OCP\AppFramework\Http\Attribute\NoAdminRequired;
 use OCP\AppFramework\Http\Attribute\PublicPage;
 use OCP\AppFramework\Http\DataResponse;
@@ -84,8 +85,8 @@ use OCP\UserStatus\IUserStatus;
 use Psr\Log\LoggerInterface;
 
 /**
- * @psalm-import-type SpreedRoom from ResponseDefinitions
- * @psalm-import-type SpreedRoomParticipant from ResponseDefinitions
+ * @psalm-import-type TalkParticipant from ResponseDefinitions
+ * @psalm-import-type TalkRoom from ResponseDefinitions
  */
 class RoomController extends AEnvironmentAwareController {
 	public const EVENT_BEFORE_ROOMS_GET = self::class . '::preGetRooms';
@@ -160,7 +161,7 @@ class RoomController extends AEnvironmentAwareController {
 	 * @param int $noStatusUpdate When the user status should not be automatically set to online set to 1 (default 0)
 	 * @param bool $includeStatus Include the user status
 	 * @param int $modifiedSince Filter rooms modified after a timestamp
-	 * @return DataResponse<Http::STATUS_OK, SpreedRoom[], array{X-Nextcloud-Talk-Hash: string, X-Nextcloud-Talk-Modified-Before: numeric-string}>
+	 * @return DataResponse<Http::STATUS_OK, TalkRoom[], array{X-Nextcloud-Talk-Hash: string, X-Nextcloud-Talk-Modified-Before: numeric-string}>
 	 *
 	 * 200: Return list of rooms
 	 */
@@ -243,7 +244,7 @@ class RoomController extends AEnvironmentAwareController {
 	 * Get listed rooms with optional search term
 	 *
 	 * @param string $searchTerm search term
-	 * @return DataResponse<Http::STATUS_OK, SpreedRoom[], array{}>
+	 * @return DataResponse<Http::STATUS_OK, TalkRoom[], array{}>
 	 *
 	 * 200: Return list of matching rooms
 	 */
@@ -260,9 +261,9 @@ class RoomController extends AEnvironmentAwareController {
 	}
 
 	/**
-	 * Get all (for moderators and in case of "free selection) or the assigned breakout room
+	 * Get all (for moderators and in case of "free selection") or the assigned breakout room
 	 *
-	 * @return DataResponse<Http::STATUS_OK, SpreedRoom[], array{}>|DataResponse<Http::STATUS_BAD_REQUEST, array{error: string}, array{}>
+	 * @return DataResponse<Http::STATUS_OK, TalkRoom[], array{}>|DataResponse<Http::STATUS_BAD_REQUEST, array{error: string}, array{}>
 	 *
 	 * 200: Breakout rooms returned
 	 * 400: Getting breakout rooms is not possible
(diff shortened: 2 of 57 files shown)


openapi.json
nickvergessen, line 1: File should be excluded from packaging via:
https://github.com/nextcloud/spreed/blob/1d9cf11441c6d25b2a2962938cdccd0cc2e11139/Makefile#L90

provokateurin:
I remember seeing a psalm failure with not enough memory at some point, I believe it's a memory leak that nukes psalm. Just giving it more ram didn't fix it, although I don't have this problem locally :thinking:

openapi.json
provokateurin: The file needs to be included for the https://apps.nextcloud.com/apps/ocs_api_viewer to work.

Makefile
```diff
@@ -89,6 +89,7 @@ appstore:
 	--exclude=mkdocs.yml \
 	--exclude=Makefile \
 	--exclude=node_modules \
+	--exclude=openapi.json \
```
provokateurin, line 92: Revert as this file should be packaged as it is used by https://apps.nextcloud.com/apps/ocs_api_viewer

provokateurin:
Ok here are my investigation results:
1. Reproducing sucessfully on clean ubuntu:jammy and debian:bookworm docker images -> No ubuntu specific problem
2. Installed PHP 8.2 on ubuntu:jammy -> No PHP 8.1 specific problem
3. Reproducing successfully on a clean archlinux:latest docker image -> No non-ArchLinux specific problem

I think this could be a problem with certain PHP extensions, but I will need some further investigations to confirm this.

nickvergessen:
I found out, that even with old dependencies the problem existed.

So I rebased this PR and turned around the stuff:

1. set up CI, dependency, etc.
2. then the doc modifications

Turns out, after 1 psalm still worked, but not after 2

So I started a git bisect:
```
$ git bisect log
git bisect start
# status: waiting for both good and bad commits
# good: [152a2cd9e04015eb0273ba77e128ea3ca69df1a3] Add OpenAPI CI
git bisect good 152a2cd9e04015eb0273ba77e128ea3ca69df1a3
# good: [152a2cd9e04015eb0273ba77e128ea3ca69df1a3] Add OpenAPI CI
git bisect good 152a2cd9e04015eb0273ba77e128ea3ca69df1a3
# bad: [8bb03633903b7b9d0b0550545858d2e06f38e6ae] Add annotations for PollController
git bisect bad 8bb03633903b7b9d0b0550545858d2e06f38e6ae
# good: [a7527f2ca7a812bece1944d4b573ebcafd56de29] Add annotations for ChatController
git bisect good a7527f2ca7a812bece1944d4b573ebcafd56de29
# good: [4cfc169c02f47418d79763e2da2c48dbed9da628] Add annotations for GuestController
git bisect good 4cfc169c02f47418d79763e2da2c48dbed9da628
# bad: [81588e0c022112ab054bb55cd420a47b199c0132] Add annotations for MatterbridgeController
git bisect bad 81588e0c022112ab054bb55cd420a47b199c0132
# good: [5c8a9f3b46980ee02ca3b13b0bbff11b01bc0b56] Add annotations for HostedSignalingServerController
git bisect good 5c8a9f3b46980ee02ca3b13b0bbff11b01bc0b56
# first bad commit: [81588e0c022112ab054bb55cd420a47b199c0132] Add annotations for MatterbridgeController
```

So I had a look at that commit and found the problem:
https://github.com/nextcloud/spreed/pull/10186/files#diff-4489f22fe309430bc8a3e0cc90a0410de61dd9b7b231d5b6f8680310f3b75ad5R59
```
	 * @return DataResponse<Http::STATUS_OK, SpreedMatterbridge&SpreedMatterbridgeProcessState, array{}>
```

from `getBridgeOfRoom` fixes the memory issue.

After some testing the following patch also fixes it:
```diff
diff --git a/lib/Controller/MatterbridgeController.php b/lib/Controller/MatterbridgeController.php
index 65e15eb1f..aac46265a 100644
--- a/lib/Controller/MatterbridgeController.php
+++ b/lib/Controller/MatterbridgeController.php
@@ -40,6 +40,7 @@ use OCP\IRequest;
  * @psalm-import-type SpreedMatterbridge from ResponseDefinitions
  * @psalm-import-type SpreedMatterbridgeParts from ResponseDefinitions
  * @psalm-import-type SpreedMatterbridgeProcessState from ResponseDefinitions
+ * @psalm-import-type SpreedMatterbridgeWithProcessState from ResponseDefinitions
  */
 class MatterbridgeController extends AEnvironmentAwareController {
 
@@ -56,7 +57,7 @@ class MatterbridgeController extends AEnvironmentAwareController {
        /**
         * Get bridge information of one room
         *
-        * @return DataResponse<Http::STATUS_OK, SpreedMatterbridge&SpreedMatterbridgeProcessState, array{}>
+        * @return DataResponse<Http::STATUS_OK, SpreedMatterbridgeWithProcessState, array{}>
         */
        #[NoAdminRequired]
        #[RequireLoggedInModeratorParticipant]
diff --git a/lib/ResponseDefinitions.php b/lib/ResponseDefinitions.php
index 49b82c41a..664aa7284 100644
--- a/lib/ResponseDefinitions.php
+++ b/lib/ResponseDefinitions.php
@@ -209,6 +209,8 @@ namespace OCA\Talk;
  *     running: bool,
  * }
  *
+ * @psalm-type SpreedMatterbridgeWithProcessState = SpreedMatterbridge&SpreedMatterbridgeProcessState
+ *
  * @psalm-type SpreedSignalingSettings = array{
  *     helloAuthParams: array{
  *         "1.0": array{
```

nickvergessen:
So now that we are able to solve the problem, I will take over and migrate the responses to the expected namespace and see where we can simply the things.

provokateurin:
Thanks for the debugging. The openapi-extractor still doesn't support reading the namespace as per https://github.com/nextcloud/openapi-extractor/issues/11

provokateurin:
Since this PR there have been a few fixes in openapi-extractor, so please upgrade it to the latest version before continuing.
It would be nice if I could reviewyour changes as well, since this whole thing is quite complex.

nickvergessen:
> The openapi-extractor still doesn't support reading the namespace as per https://github.com/nextcloud/openapi-extractor/issues/11

Neither the version it seems (locking 0.0.1 while talk is at 18.0…). Will try to have a look

nickvergessen:
> The version field does not indicate the version of the software but the version of the document instead.

Okay, so it updates automatically everytime something changes? Or is hardcoded? Because it should reflect to breaking changes (api removed), etc I guess?

nickvergessen:
> It would be nice if I could reviewyour changes as well, since this whole thing is quite complex.

Pushed them as fixups for now.

But yeah the annoying part (namespace) and then also unify the naming pattern of the responses seems to require work in the openapi-extractor first. So will try to get that running from source tomorrow

provokateurin:
> Okay, so it updates automatically everytime something changes? Or is hardcoded? Because it should reflect to breaking changes (api removed), etc I guess?

No, right now it is completely hardcoded to 0.0.1.
I think the best would be to automatically bump the version if an existing spec is detected and the new version doesn't match the old version.
Let's discuss this in the openapi-extractor repo, can you open an issue there?

.github/workflows/openapi.yml
```diff
@@ -0,0 +1,52 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set up dependencies
+        run: composer i
+
+      - name: Regenerate OpenAPI
+        run: composer run openapi
+
+      - name: Check webpack build changes
```
provokateurin, line 43: Wrong name

vendor-bin/openapi-extractor/composer.json
```diff
@@ -0,0 +1,16 @@
...
+			"url": "https://github.com/nextcloud/openapi-extractor"
+		}
+	],
+	"config": {
+		"platform": {
+			"php": "8.1"
+		}
+	},
+	"require-dev": {
+		"nextcloud/openapi-extractor": "dev-main"
```
provokateurin, line 14:
```suggestion
		"nextcloud/openapi-extractor": "dev-main#aef9a921bebdcdc42b110a51bfb1e2ca281c307a"
```
Should be pinned

lib/Controller/RoomController.php
```diff
@@ -276,13 +293,27 @@ public function getBreakoutRooms(): DataResponse {
...
+	 */
 	#[PublicPage]
 	#[BruteForceProtection(action: 'talkRoomToken')]
 	#[BruteForceProtection(action: 'talkSipBridgeSecret')]
 	public function getSingleRoom(string $token): DataResponse {
 		try {
 			$isSIPBridgeRequest = $this->validateSIPBridgeRequest($token);
 		} catch (UnauthorizedException $e) {
+			/**
+			 * A hack to fix type collision
```
nickvergessen, line 314: 🙈

lib/Controller/RoomController.php
```diff
@@ -834,14 +965,26 @@ protected function formatParticipantList(array $participants, bool $includeStatu
 		return new DataResponse(array_values($results), Http::STATUS_OK, $headers);
 	}
 
+	/**
+	 * Add a participant to a room
+	 *
+	 * @param string $newParticipant New participant
+	 * @param string $source Source of the participant
+	 * @return DataResponse<Http::STATUS_OK, array{type: ?int}, array{}>|DataResponse<Http::STATUS_NOT_FOUND|Http::STATUS_NOT_IMPLEMENTED, array<empty>, array{}>|DataResponse<Http::STATUS_BAD_REQUEST, array{error: ?string}, array{}>
+	 * return DataResponse<Http::STATUS_OK|Http::STATUS_BAD_REQUEST|Http::STATUS_NOT_FOUND|Http::STATUS_NOT_IMPLEMENTED, ?array{type?: int, error?: string}, array{}>
```
nickvergessen, line 974: is this an earlier attempt from the previous line?

lib/Controller/RoomController.php
```diff
@@ -834,14 +965,26 @@ protected function formatParticipantList(array $participants, bool $includeStatu
...
+	 */
 	#[NoAdminRequired]
 	#[RequireLoggedInModeratorParticipant]
 	public function addParticipantToRoom(string $newParticipant, string $source = 'users'): DataResponse {
 		if ($this->room->getType() === Room::TYPE_ONE_TO_ONE
 			|| $this->room->getType() === Room::TYPE_ONE_TO_ONE_FORMER
 			|| $this->room->getType() === Room::TYPE_NOTE_TO_SELF
 			|| $this->room->getObjectType() === 'share:password') {
-			return new DataResponse([], Http::STATUS_BAD_REQUEST);
+			return new DataResponse(['error' => null], Http::STATUS_BAD_REQUEST);
```
nickvergessen, line 987: Can we roll this back? empty array should be supported totally fine and since it changes the returned type of "error" it would potentially break GSON and other libs when the client defined it as "its there and string, or absent"

vendor-bin/openapi-extractor/composer.json
nickvergessen: The commited composer.lock file is the pin

lib/Controller/RoomController.php
provokateurin: Oh, probably yes

lib/Controller/RoomController.php
provokateurin: Yes, but then you need to add `array<empty>` to the union type

nickvergessen: Done in https://github.com/nextcloud/spreed/pull/10186/commits/59d25a7abcb951db211b6d70369904403a2d18a8

lib/Controller/RoomController.php
nickvergessen: Deduplicated in https://github.com/nextcloud/spreed/pull/10186/commits/59d25a7abcb951db211b6d70369904403a2d18a8

nickvergessen:
After aligning with Kate in the chat, I squashed the fixups, kept my changes in individual commits for easier retrospective and rebased it.
So we can merge it and rebase all open PHP PRs

nickvergessen (approved):

PR #10188, opened 2023-08-11
feat(MediaSettings) - Adjust the preview background update to only change in video when confirmed

### LINKED ISSUE #9379: Preview video backgrounds when in a call
Currently when clicking a video background in the media settings during a call, the video stream is immediately changed in the call video.

It would be ideal to preview the background and other effects in the media settings and only change the call background once the user confirms it by clicking a button.

DorraJaouad:
### ☑️ Resolves

* Fix #9379

### 🚧 Tasks

- [ ] Code review
- [ ] Visual review

### 🏁 Checklist


src/components/MediaSettings/MediaSettings.vue
```diff
@@ -170,7 +170,7 @@
 					class="call-button"
 					:force-join-call="true"
 					:silent-call="silentCall" />
-				<NcButton v-else @click="closeModal">
+				<NcButton v-else @click="closeModalAndConfirmBackground">
 					{{ t('spreed', 'Done') }}
```
Antreesy, line 174:
```suggestion
				<NcButton v-else-if="updatedBackground" @click="closeModalAndConfirmBackground">
					{{ t('spreed', 'Apply changes') }}
```

Don't know why we had that button before, as modal could be closed with outside click, or with a cross 
cc @marcoambrosini

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -421,16 +422,31 @@ export default {
...
 
-		handleUpdateBackground(background) {
+		closeModalAndConfirmBackground() {
+			if (this.updatedBackground) {
+				this.handleBackgroundUpdate()
+			}
+			this.closeModal()
+		},
+
+		handleBackgroundUpdate() {
```
Antreesy, line 432:
```suggestion
		handleUpdateBackground(background) {
```

I'd pass a value, to make this method more independent

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -448,7 +464,7 @@ export default {
 			if (this.isInCall) {
 				localMediaModel.disableVirtualBackground()
 			} else {
-				BrowserStorage.setItem('virtualBackgroundEnabled_' + this.token, 'false')
+				BrowserStorage.removeItem('virtualBackgroundEnabled_' + this.token)
```
Antreesy, line 467: Aligned with App.vue :+1:

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -421,16 +422,31 @@ export default {
 			}
 		},
 
-		handleUpdateBackground(background) {
+		closeModalAndConfirmBackground() {
```
Antreesy, line 425: `closeModalAndApplyBackground()` is also fine for me

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -421,16 +422,31 @@ export default {
...
+			if (this.updatedBackground === 'none') {
+				this.clearBackground()
+			} else if (this.updatedBackground === 'blur') {
+				this.blurBackground()
+			} else {
+				this.setBackgroundImage(this.updatedBackground)
+			}
+		},
+
+		handleUpdateBackgroundPreview(background) {
```
Antreesy, line 442:
```suggestion
		handleUpdateVirtualBackground(background) {
```

Antreesy (commented):
Seems to work, nothing blocking - just remarks about wording / naming

src/components/MediaSettings/MediaSettings.vue
DorraJaouad: The `mediaSettings `also has Devices settings. It will be weird to show `'Apply changes'` button only for the background change and not for changing the media devices. I'd prefer to keep it as a permanent `'Done'` button for any changes in general OR  show it for both changes. 

P.S: If we go with the conditioned button, I think it should be removed if the chosen option is the same as the current one in order to reflect 'real changes' only.

marcoambrosini (commented):
Hi @DorraJaouad  & @Antreesy :)
I think "Apply settings" works well for every setting there, and we should use it it in the devices page too. By that I mean that both camera and microphone device settings shouldn't immediately affect the streams in the call but only the preview videos and mic level. 
Changes should occur only when "apply settings" button is clicked.  
What do you think?

Antreesy:
> Changes should occur only when "apply settings" button is clicked. What do you think?

Should be possible to split this logic to preview and apllied device, same as for background, so I'd like to see it

src/components/MediaSettings/MediaSettings.vue
Antreesy, line 467: Please also check BrowserStorage (devtools - application - local storage):
- url, type and enabled flag are stored by tokens for each conversation
- they appear on screen in Media Settings
- but option in the component points to "None", not the defined before image

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -399,6 +403,8 @@ export default {
 
 		closeModal() {
 			this.modal = false
+			this.updatedBackground = undefined
+			this.deviceIdChanged = false
 		},
 
 		toggleAudio() {
```
Antreesy, line 410: Small wish: Audio/Video toggle buttons here in TopBar are not syncing (although they could with the BrowserStorage).
How about:
1) on `toggleAudio` and `toggleVideo` also add a flag `this.deviceStateChanged = true` and follow it for 'Apply Changes` button
2) on LocalAudioControlButton (and LocalVideoControlButton) add `subscribe('talk:media-settings:hide', this.updateDeviceState)`
```
updateDeviceState() {
if (BrowserStorage.getItem('audioDisabled_' + this.token)) {
  this.model.disableAudio()
} else {
  this.model.enableAudio()
}
}
```
3) also for LocalAudioControlButton expose method `toggleAudio` and replace in TopBarMediaControls duplicated method with `this.$refs.audioControl.toggelAudio()`

Antreesy (approved):

Antreesy (approved):
:tada:

PR #10203, opened 2023-08-14
feat(conversationsStore) - cache conversations to BrowserStorage

Antreesy:
### ☑️ Resolves

* Allow to store fetched conversations in BrowserStorage, and recover it instantly from cache, while waiting for a server response
* Conversations updated with `patchConversations()` action, which is called approx. every 30 seconds
* If conversation was deleted, it will be removed from the view with the next full fetch, with redirect to `/apps/spreed/not-found`

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks

- [ ] Manual testing
- [ ] Code review

### 🏁 Checklist

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.spec.js
+++ b/src/components/LeftSidebar/LeftSidebar.spec.js
@@ -14,6 +14,7 @@ import { searchPossibleConversations, searchListedConversations } from '../../se
 import { EventBus } from '../../services/EventBus.js'
 import storeConfig from '../../store/storeConfig.js'
 import { findNcListItems, findNcActionButton } from '../../test-helpers.js'
+import { requestTabLeadership } from '../../utils/requestTabLeadership.js'
 
 jest.mock('@nextcloud/initial-state', () => ({
 	loadState: jest.fn(),
@@ -131,6 +132,8 @@ describe('LeftSidebar.vue', () => {
 
 			const wrapper = mountComponent()
 
+			await requestTabLeadership()
+
 			expect(fetchConversationsAction).toHaveBeenCalledWith(expect.anything(), expect.anything())
 			expect(conversationsListMock).toHaveBeenCalled()
 
@@ -157,6 +160,9 @@ describe('LeftSidebar.vue', () => {
 
 		test('re-fetches conversations every 30 seconds', async () => {
 			const wrapper = mountComponent()
+
+			await requestTabLeadership()
+
 			expect(wrapper.exists()).toBeTruthy()
 			expect(fetchConversationsAction).toHaveBeenCalled()
 
@@ -175,6 +181,9 @@ describe('LeftSidebar.vue', () => {
 
 		test('re-fetches conversations when receiving bus event', async () => {
 			const wrapper = mountComponent()
+
+			await requestTabLeadership()
+
 			expect(wrapper.exists()).toBeTruthy()
 			expect(fetchConversationsAction).toHaveBeenCalled()
 
@@ -302,6 +311,8 @@ describe('LeftSidebar.vue', () => {
 			}
 
 			const wrapper = mountComponent()
+
+			await requestTabLeadership()
 
 			expect(fetchConversationsAction).toHaveBeenCalledWith(expect.anything(), expect.anything())
 			expect(conversationsListMock).toHaveBeenCalled()
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -261,7 +261,9 @@ import {
 	searchListedConversations,
 } from '../../services/conversationsService.js'
 import { EventBus } from '../../services/EventBus.js'
+import { talkBroadcastChannel } from '../../services/talkBroadcastChannel.js'
 import CancelableRequest from '../../utils/cancelableRequest.js'
+import { requestTabLeadership } from '../../utils/requestTabLeadership.js'
 
 export default {
 
@@ -331,6 +333,8 @@ export default {
 			preventFindingUnread: false,
 			roomListModifiedBefore: 0,
 			forceFullRoomListRefreshAfterXLoops: 0,
+			isFetchingConversations: false,
+			isCurrentTabLeader: false,
 			isFocused: false,
 			isFiltered: null,
 		}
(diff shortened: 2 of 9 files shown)


src/store/conversationsStore.js
```diff
@@ -339,6 +340,47 @@ const actions = {
...
+	 * @param {object} context default store context
+	 */
+	putConversationsToStorage(context) {
+		const conversations = context.getters.conversationsList
+		if (!conversations.length) {
+			return
+		}
+
+		const serializedConversations = JSON.stringify(conversations)
+		BrowserStorage.setItem('cachedConversations_', serializedConversations)
```
ShGKme, line 382: **nitpick (non-blocking):** maybe remove trailing `_`?

```suggestion
		BrowserStorage.setItem('cachedConversations', serializedConversations)
```

ShGKme (approved):
Tested, works fine. Only some naming preferences.

Checked that LS is pruned on logout, not security issues ✅

---

From the UX side: after reload, a user sees an old conversations list with old last messages. That could be confusing. Should we add any small load indicator?

---

`BrowserStorage` is reused by tabs. If we store conversations in the browser storage, we should do it only once.

Currently it is used only for initialization. But if we open 5 tabs with Talk, then we have 5 apps with 5 conversation re-fetchings and 5 parallel updating `localStorage`. The most optimal solution here could be to run re-fetching only on one tab/worker and listen to the `storage` windows event or `BroadcastChannel`.

ShGKme:
Upd: instead of shared worker, it could be easier and better for browser support to use [Web Locks API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API)

ShGKme:
Notes about using Web Locks API here.

We don't have actually a "resource" that we want to block. But we can use a Leader pattern.
With Locks API we can lock a tab to be a "Leader" who is responsible for everything that should be done only in one tab. Then let's call "being a leader" a resource that is lock for infinity time (until the tab is not closed).

Then we can create a helper like

```js
async function onTalkLeaderTab(() => {
  // Promise of waiting to be a leader
  return new Promise((resolve) =>
    // Request "being a leader"
    navigator.locks.request('talk:leader', () => {
      // Since we got this "resource" this tab is a leader. Resolve promise
      resolve()
      
      // Infinity promise, resource is blocked forever, while tab is not closed
      return new Promise(() => {})
  }
}

onTalkLeaderTab()
  .then(() => {
    // Now we are the leader
    // Start the setInterval for re-fetching or anything
  })
```

One important thing: it only works with HTTPS ⚠️

Then we can use the `BroadcastChannel` to inform other tabs about new conversations (cc https://github.com/nextcloud-libraries/nextcloud-event-bus/pull/600)

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -668,16 +680,41 @@ export default {
...
+		async onFetchingConversationsLeader() {
+			return new Promise((resolve) => {
+				window.navigator.locks.request('talk:conversations-fetching-leader', () => {
+					resolve()
+
+					// return an infinity promise, resource is blocked until 'leader' tab is closed
+					return new Promise(() => {})
+				})
+			})
+		},
```
ShGKme, line 716: I'd make this method universal `util` like `onTalkLeader` from the last comment.

This method is not a part of the `LeftSidebar` Vue component instance: it is not related to its state or its properties.

And at the same time, if we need a talk leader anywhere else, we can reuse this lock as a general "`talk:leader`"
 resource.

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -410,13 +412,23 @@ export default {
...
+
+		this.onFetchingConversationsLeader().then(() => {
+			this.isFetchingConversationsLeader = true
+			this.fetchConversations()
+		})
 	},
 
 	mounted() {
-		// Refreshes the conversations every 30 seconds
+		// TODO update process when '@nextcloud/event-bus' broadcasting between tabs is available
```
ShGKme, line 426: **thought:** We can use the `BroadcastChannel` even without `@nextcloud/event-bus` support. It's quite an easy-to-use API.

I'm not sure if it will be better in performance. On one hand, it clones data and interacts with all the tabs. On the other hand, WebStorage is a sync (blocking) PI + spends some time on serialization/deserialization.

But it would remove set interval and updating only once per 30 sec in subtabs. 🤔

src/components/LeftSidebar/LeftSidebar.spec.js
```diff
@@ -26,6 +26,13 @@ jest.mock('../../services/conversationsService', () => ({
 // short-circuit debounce
 jest.mock('debounce', () => jest.fn().mockImplementation(fn => fn))
 
+window.navigator.locks = {
+	request: jest.fn().mockImplementation((name, callback) => {
+		callback()
+		return new Promise(() => {})
+	}),
```
ShGKme, line 33: `new Promise(() => {})` is an infinity promise... It should be `return Promise.resolve()`. And I have no idea why it works 😅

src/components/LeftSidebar/LeftSidebar.vue
ShGKme, line 716: **issue (blocking):** We should also check for lock availability here and consider any tab to be a leader otherwise. It is not available without HTTPS.

ShGKme (changes requested):

src/store/conversationsStore.js
```diff
@@ -339,6 +340,62 @@ const actions = {
 				context.dispatch('updateConversationIfHasChanged', newConversation)
 			}
 		}
+
+		context.dispatch('cacheConversations')
+	},
```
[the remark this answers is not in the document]
Antreesy: Added an option, if we will need it one day, but right now updating is working with the next fetch (except from deleted conversations)

src/utils/talkLeader.js
```diff
@@ -0,0 +1,51 @@
...
+ * Assign the first window, which requested that, a 'fetching leader'.
+ * It will fetch conversations with defined interval and cache to BrowserStorage
+ * Other will update list with defined interval from the BrowserStorage
+ * Once 'leader' is closed, next requested tab will be assigned, and so on
+ */
+const onTalkLeaderTab = async () => {
+	return new Promise((resolve) => {
+		// Locks are supported only with HTTPS protocol,
+		// so we don't lock anything for another cases
+		if (window.location.protocol !== 'https:') {
```
ShGKme, line 34: Could you change the check of a protocol to API availability?

It has no effect on web-client (as Web Locks API is available on any supported by Talk web browser), but it would make it work in the Talk Desktop client with multiple windows.

```suggestion
		if (navigator.locks === undefined) {
```

ShGKme (approved):
Tested, everything seems to work fine. Tested with 2-3 tabs, with changing leader, creating new conversations, deleting conversations.

ShGKme:
Follow-up: because x-talk-hash header is checked not in any HTTP response, but only in re-fetching conversations for the `LeftSidebar`, it is not checked not in sub-tabs.

Antreesy:
> Follow-up: because x-talk-hash header is checked not in any HTTP response, but only in re-fetching conversations for the `LeftSidebar`, it is not checked not in sub-tabs.

Rebased, follow-up fixed with additional broadcasted message specifically for talk hash.
Also a logic has been moved to the store and App.vue, to be able to use a 'leader' locked status within different processes in one tab.

image

Antreesy:
> Follow-up: because x-talk-hash header is checked not in any HTTP response, but only in re-fetching conversations for the `LeftSidebar`, it is not checked not in sub-tabs.

Rebased, follow-up fixed with additional broadcasted message specifically for talk hash.
Also a logic has been moved to the store and App.vue, to be able to use a 'leader' locked status within different processes in one tab.

image

src/store/talkHashStore.js
```diff
@@ -22,6 +22,8 @@
 
 import { showError, TOAST_PERMANENT_TIMEOUT } from '@nextcloud/dialogs'
 
+const talkBroadcastChannel = new BroadcastChannel('nextcloud:talk')
+
```
ShGKme, line 26: It works fine and is totally my personal preference, but I'd also move this to a separated module, so we won't need to create it again each time and bother about typos in the channel name.

src/utils/talkLeader.js
```diff
@@ -0,0 +1,51 @@
...
+
+			// return an infinity promise, resource is blocked until 'leader' tab is closed
+			return new Promise(() => {})
+		})
+	})
+}
+
+export {
+	onTalkLeaderTab,
+}
```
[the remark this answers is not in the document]
Antreesy: Relatively simple solution, code-wise

ShGKme:
Follow-ups:
- Don't cache conversations when nothing actually changed (we know it from the patching action)
- Add methods to post specific messages into the broadcast channel into the talkBroadcastChannel service as well as methods to add message handlers
- Think about centralized way to control what and when is synced, when to post messages

ShGKme (approved):

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -618,7 +649,13 @@ export default {
 		 */
 		async handleShouldRefreshConversations(options) {
 			if (options?.all === true) {
-				this.roomListModifiedBefore = 0
+				if (this.isCurrentTabLeader) {
+					this.roomListModifiedBefore = 0
```
Antreesy, line 653:
```suggestion
					this.roomListModifiedBefore = 0
					this.forceFullRoomListRefreshAfterXLoops = 10
```

follow-up: restart cycle

PR #10262, opened 2023-08-18
Virtual scrolling for conversations list

ShGKme:
PR is made on top of https://github.com/nextcloud/spreed/pull/10051

### ☑️ Resolves

Performance issues:
1. Too many conversations avatars request: 
   Opening a page with 500 conversations immediately makes 500 requests for conversation avatars
2. Initial loading is slow for long chat lists
3. Scrolling a long conversations list and interacting with the menu could be slow on a long conversations list

### 🖼️ Screenshots

Almost no visual changes, except scrolling behavior.

By default scroll to position in virtual scrolling just "teleport" to a new position instantly:

To the end | To unread
---|---
instant | instant-2

It doesn't look good and it is different from the current Talk behavior. And it makes it less understandable from what direction (up/down) the user have moved.

But we cannot just use smooth scrolling... Smooth scrolling from the beginning to the end means - loading **ALL** avatars in between and not very performant in general.

The idea is "imitating" the smooth scrolling:
1. If the new position is far from the current - first instantly move to a distance of one visual list height from the target
2. Then scroll with smooth just for a small number of elements

Scroll close | Scroll a bit far | Scroll from the end of 500
---|---|---
scroll-close | scroll-far-1 | scroll-far-2

Throw 500 chats scrolling is not super smooth but smooth a bit :D

### 🚧 Tasks

- [x] Choose and add a virtual scroller
  - Currently, one popular from a Vue core team member is used: https://github.com/Akryum/vue-virtual-scroller
  - But I'd switch to our own scroller in future
- [x] Use a virtual scroller for the conversations list
  - [x] Change scrolling and unread mentions handling with the virtual list
- [x] Use `` with implicit `key` instead of `<NcAvatar>` for conversations
  - Without a `key` on `` there is a glitch effect on scrolling because Vue updates `src` of the images. Even though images are cached after the first load, the glitch effect is still visible.
  - It is still a problem for 1-1

With glitching | Without glitching
---|---
with-glitch | no-glitch

#### No virtual scrolling in the search results

A virtual scroller is used ONLY for the main conversations list (incl. filtered conversations) and **NOT** for the search results. The search results are more complicated because it is not a homogeneous list but several lists with different elements in between.

#### Glitching effect with 1-1 conversations avatars

To avoid it we need to stop using `<NcAvatar>` for 1-1 (what is problematic with the user status) or add `key` to `<NcAvatar>` and make it more performant on props updating.

### 👀 Alternative solutions 

If we only want to fix the first problem with avatar fetching, we can just implement lazy loading for conversation avatars.

### 🏁 Checklist

### DIFF
--- a/src/components/LeftSidebar/ConversationsListVirtual.vue
+++ b/src/components/LeftSidebar/ConversationsListVirtual.vue
@@ -1,0 +1,156 @@
+<!--
+  - @copyright Copyright (c) 2023 Grigorii Shartsev <me@shgk.me>
+  -
+  - @author Grigorii Shartsev <me@shgk.me>
+  -
+  - @license AGPL-3.0-or-later
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -->
+
+<template>
+	<RecycleScroller ref="scroller"
+		item-tag="ul"
+		:items="conversations"
+		:item-size="CONVERSATION_ITEM_SIZE"
+		key-field="token">
+		<template #default="{ item }">
+			<Conversation :item="item" />
+		</template>
+		<template #after>
+			<LoadingPlaceholder v-if="loading" type="conversations" />
+		</template>
+	</RecycleScroller>
+</template>
+
+<script>
+import { RecycleScroller } from 'vue-virtual-scroller'
+
+import LoadingPlaceholder from '../LoadingPlaceholder.vue'
+import Conversation from './ConversationsList/Conversation.vue'
+
+import 'vue-virtual-scroller/dist/vue-virtual-scroller.css'
+
+const CONVERSATION_ITEM_SIZE = 64
+
+export default {
+	name: 'ConversationsListVirtual',
+
+	components: {
+		LoadingPlaceholder,
+		Conversation,
+		RecycleScroller,
+	},
+
+	props: {
+		conversations: {
+			type: Array,
+			required: true,
+		},
+
+		loading: {
+			type: Boolean,
+			default: false,
+		},
+	},
+
+	setup() {
+		return {
+			CONVERSATION_ITEM_SIZE,
+		}
+	},
+
+	methods: {
+		/**
+		 * Get an index of the first fully visible conversation in viewport
+		 *
+		 * @public
+		 * @return {number}
+		 */
+		getFirstItemInViewportIndex() {
+			// (ceil to include partially) of (absolute number of items above viewport) + 1 (next item is in viewport) - 1 (index starts from 0)
+			return Math.ceil(this.$refs.scroller.$el.scrollTop / CONVERSATION_ITEM_SIZE)
+		},
+
+		/**
+		 * Get an index of the last fully visible conversation in viewport
+		 *
+		 * @public
+		 * @return {number}
+		 */
+		getLastItemInViewportIndex() {
+			// (floor to include only fully visible) of (absolute number of items below and in viewport) - 1 (index starts from 0)
+			return Math.floor((this.$refs.scroller.$el.scrollTop + this.$refs.scroller.$el.clientHeight) / CONVERSATION_ITEM_SIZE) - 1
+		},
+
+		/**
+		 * Scroll to conversation by index
+		 *
+		 * @public
+		 * @param {number} index - index of conversation to scroll to
+		 * @return {Promise<void>}
+		 */
+		async scrollToItem(index) {
+			const firstItemIndex = this.getFirstItemInViewportIndex()
+			const lastItemIndex = this.getLastItemInViewportIndex()
+
+			const viewportHeight = this.$refs.scroller.$el.clientHeight
+
+			/**
+			 * Scroll to a position with smooth scroll imitation
+			 *
+			 * @param {number} to - target position
+			 * @return {void}
+			 */
+			const doScroll = (to) => {
+				const ITEMS_TO_BORDER_AFTER_SCROLL = 1
+				const padding = ITEMS_TO_BORDER_AFTER_SCROLL * CONVERSATION_ITEM_SIZE
+				const from = this.$refs.scroller.$el.scrollTop
+				const direction = from < to ? 1 : -1
+
+				// If we are far from the target - instantly scroll to a close position
+				if (Math.abs(from - to) > viewportHeight) {
+					this.$refs.scroller.scrollToPosition(to - direction * viewportHeight)
+				}
+
+				// Scroll to the target with smooth scroll
+				this.$refs.scroller.$el.scrollTo({
+					top: to + padding * direction,
+					behavior: 'smooth',
+				})
+			}
+
+			if (index < firstItemIndex) { // Item is above
+				await doScroll(index * CONVERSATION_ITEM_SIZE)
+			} else if (index > lastItemIndex) { // Item is below
+				// Position of item + item's height and move to bottom
+				await doScroll((index + 1) * CONVERSATION_ITEM_SIZE - viewportHeight)
+			}
+		},
+
+		/**
+		 * Scroll to conversation by token
+		 *
+		 * @param {string} token - token of conversation to scroll to
+		 * @return {void}
+		 */
+		scrollToConversation(token) {
+			const index = this.conversations.findIndex(conversation => conversation.token === token)
+			if (index !== -1) {
+				this.scrollToItem(index)
+			}
+		},
+	},
+}
+</script>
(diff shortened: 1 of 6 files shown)


src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -680,55 +698,39 @@ export default {
...
+		 */
 		handleUnreadMention() {
-			this.unreadNum = 0
-			const unreadMentions = document.querySelectorAll('.unread-mention-conversation')
-			unreadMentions.forEach(x => {
-				if (this.elementIsBelowViewpoint(this.$refs.container, x)) {
-					if (this.unreadNum === 0) {
-						this.firstUnreadPos = x.offsetTop
+			this.$nextTick(() => {
+				this.unreadNum = 0
```
Antreesy, line 709: As we don't use an exact amount of unread mentions, maybe, transform it to the boolean value? Or discard it and keep only `firstUnreadIndex` for both tracking and navigating purposes?

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -680,55 +698,39 @@ export default {
...
+				const lastConversationInViewport = this.$refs.scroller.getLastItemInViewportIndex()
+				for (let i = this.filteredConversationsList.length - 1; i > lastConversationInViewport; i--) {
```
Antreesy, line 711:
```suggestion
				for (let i = lastConversationInViewport; i < this.filteredConversationsList.length - 1; i++) {
```
I'd go with another order, as unread mentions have newer lastActivity, and stay higher in the list

src/components/LeftSidebar/ConversationsListVirtual.vue
```diff
@@ -0,0 +1,123 @@
...
+		</template>
+	</RecycleScroller>
+</template>
+
+<script>
+import { RecycleScroller } from 'vue-virtual-scroller'
+
+import 'vue-virtual-scroller/dist/vue-virtual-scroller.css'
+
+import Conversation from './ConversationsList/Conversation.vue'
```
Antreesy, line 39: lint

ShGKme: But Linter on the CI says everything is fine...

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -680,55 +698,39 @@ export default {
...
-						top: conversation.offsetTop - conversation.offsetHeight,
-						behavior: 'smooth',
-					})
-				}
-			})
+			if (!this.$refs.scroller) {
+				return
+			}
+
+			await this.$refs.scroller.scrollToConversation(token)
```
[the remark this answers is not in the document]
ShGKme: Hmm... I need to check it, thanks

src/components/LeftSidebar/LeftSidebar.vue
ShGKme: I kept the current algorithm, we scroll to the latest (most down) unread mention.

Actually, first I suddenly implemented it as you proposed :D 

And then there was a problem. If you have a long list of unread mentions, then, even if you scroll to the first one, you still have some unread mentions below the viewport.

So 
1. You click on "Unread mentions"
2. It scrolls to the beginning of the long list of unread mentions
3. But you still see the button
4. You click again and move just one move conversation down
5. And then again, until you move to the last one

Antreesy: Is that a feature, like an inbox, to process all one by one?)

src/components/LeftSidebar/ConversationsListVirtual.vue
Antreesy: It's a warning, not an error

src/components/LeftSidebar/LeftSidebar.vue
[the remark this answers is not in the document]
ShGKme: Well, the scrolling doesn't work on the `master` as well, because conversation can be joined before the list is loaded. But it helped me to find a regression (https://github.com/nextcloud/spreed/pull/10051#issuecomment-1683867869)

src/components/LeftSidebar/LeftSidebar.vue
ShGKme: Well, we have a filter already to have all unread now :D

So I think the primary feature of this button is to indicate that you have something unread below.

src/components/LeftSidebar/ConversationsListVirtual.vue
ShGKme: 🙈

ShGKme:
- Rebased onto refactoring PR
- Fixed according to the review - linting, `unreadNum` variable
- Fixed regression with not showing joined conversation during loading

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -111,17 +111,18 @@
 		</div>
 
 		<template #list>
-			<li ref="container" class="left-sidebar__list" @scroll="debounceHandleScroll">
-				<ul class="scroller">
+			<li ref="container" class="left-sidebar__list">
+				<ul class="scroller h-100">
 					<!-- Conversations List -->
 					<template v-if="!isSearching">
```
Antreesy, line 117: UPD: using `v-show` and keeping scroller in DOM helps with keeping the scrol position

src/components/LeftSidebar/ConversationsListVirtual.vue
```diff
@@ -34,10 +34,10 @@
 <script>
 import { RecycleScroller } from 'vue-virtual-scroller'
 
-import 'vue-virtual-scroller/dist/vue-virtual-scroller.css'
-
 import Conversation from './ConversationsList/Conversation.vue'
 
+import 'vue-virtual-scroller/dist/vue-virtual-scroller.css'
```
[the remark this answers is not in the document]
ShGKme: > The node_module folder is not shipped, so they would need copying to another location?

No, they wouldn't. `<style src="...">` is a `.vue` file feature, not native (Native `<style>` has no `src` attr). It is supposed to work exactly the same as `import 'path to .css`.

> I know what we're doing it everywhere in the project, bu isn't it possible (and cleaner) to import external styles through [style tag](https://vuejs.org/api/sfc-spec.html#src-imports)?

It is possible. And sounds logical. But I personally would use `import` here:
- We work with styles as with modules and dependencies. For me, it's more visible to import dependencies with import. Lib's styles and lib's are imported close to each other
- It is generally more common practice import css directly in SFC. I've never seen using `<style src` for dependencies. Only for component's css.
- Not applicable in this case, but sometimes IDEs don't properly process using `src` in `SFC`.

src/components/LeftSidebar/ConversationsListVirtual.vue
```diff
@@ -0,0 +1,131 @@
...
+		},
+
+		async scrollToItem(index) {
+			const firstItemIndex = this.getFirstItemInViewportIndex()
+			const itemsInViewport = this.getItemsCountInViewport()
+			const lastItemIndex = this.getLastItemInViewportIndex()
+
+			// Make a trick with "smooth" scroll
+
+			if (index < firstItemIndex) {
```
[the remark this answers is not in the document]
ShGKme: Thanks.

Fixed, but by correcting the fist and last item calculation.

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -686,55 +695,38 @@ export default {
...
-				// See https://github.com/vuejs/vue/issues/4952#issuecomment-280661367
-				// Fixed in Vue 3
-				// Temp solution - use unique ref name for each v-for element. The value is still array but with one element
-				// TODO: Vue3: remove [0] here or use object for template refs
-				const conversation = this.$refs[`conversation-${token}`]?.[0].$el
-				if (!conversation) {
-					return
-				}
+		async scrollToConversation(token) {
+			await this.$nextTick()
```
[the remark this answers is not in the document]
ShGKme: > When picking a conversation with search, and it's on the bottom, scroller previous position is already dropped, and we scroll through the whole list to the picked one

I'm not sure I understand what you suggest here. Is it about adding `v-show` mentioned in the next comment?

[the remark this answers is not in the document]
Antreesy: Yes, then destroy and mount again component, previously scroll position is not keeped anywhere.
Patch is required to trigger `abortSearch()` earlier and from proper place, when `scrollToconversation()` will be triggered when join a conversation, and smoothscroll should jump to nearest place and smooth scroll further (your hack)

ShGKme:
Rebased onto master with new `cachedConversations` feature.

See commits from `fixup! remove unused code`:
- Fix linting errors and remove unused code
- Fix scrolling behavior
- Fix some issues with abort search and scrolling to newly selected conversations

@Antreesy please, have a look.

src/components/LeftSidebar/LeftSidebar.vue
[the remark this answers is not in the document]
ShGKme: I removed `resetSearchFilter` at all, and used a direct click event from the child component.

ShGKme:
Join a conversation from search update:

Before | After
---|---
before | after

src/components/LeftSidebar/LeftSidebar.vue
Antreesy: Still thinking, that it's an issue: try scroll to the bottom and use search from there

Antreesy (approved):
Tested, better now
Let's rebase onto master and squash, and think again, if `v-show` is applicable

Antreesy:
> I'd not mix this feature with virtual scrolling.

Then let's consider it for the follow-up

nickvergessen:
So time for the fixup rebase?

ShGKme:
Forced pushed with squashed commits and very minor changes aka typos in comments

Issue #10283, opened 2023-08-20
follow-up(LeftSidebar) - caching, broadcast optimizations


Antreesy:
Follow-ups:
- Don't cache conversations when nothing actually changed (we know it from the patching action)
- Add methods to post specific messages into the broadcast channel into the talkBroadcastChannel service as well as methods to add message handlers
- Think about centralized way to control what and when is synced, when to post messages

_Originally posted by @ShGKme in https://github.com/nextcloud/spreed/issues/10203#issuecomment-1684264487_

PR #10284, opened 2023-08-20
follow-up(LeftSidebar) - caching optimization

Antreesy:
### ☑️ Resolves

* Part of #10283
* If full fetch was forced from signaling (leader or sub-tab), counter will be reset to 10 again (like we do after usual full fetch)
* cashing was moved to `patchConversations` action called by `fetchConversations` action
* a flag was added to `patchConversations` to watch, if there was changes after fetch to cache them
* `updateConversationIfHasChanged` logic was moved from mutation to the action

---
### 🖼️ Screenshots
No visual changes

### 🚧 Tasks
No due tasks

### 🏁 Checklist


ShGKme (approved):
All looks good

PR #10301, opened 2023-08-22
fix(BrowserStorage) - clear BrowserStorage when different user logged in

Antreesy:
### ☑️ Resolves

* If another user is logged in, we should check, whether it's `uid` matches with last cached `uid`.
  * if it matches, we could safely restore and use data from storage;
  * if not, we wipe everything related to the last user and containing possibly sensitive data;

### 🖼️ Screenshots
No visual changes

### 🚧 Tasks

- [ ] Add encryption?
- [ ] Another possible checks if `uid` is not enough?

### 🏁 Checklist

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -227,6 +227,22 @@ export default {
 		},
 	},
 
+	beforeCreate() {
+		const authorizedUser = getCurrentUser()?.uid || null
+		const lastLoggedInUser = BrowserStorage.getItem('last_logged_in_user')
+
+		if (authorizedUser !== lastLoggedInUser) {
+			// TODO introduce helper/util to list and clear all sensitive data
+			// or create BrowserSensitiveStorage for this purposes,
+			// if we have more than one source
+			BrowserStorage.removeItem('cachedConversations')
+		}
+
+		if (authorizedUser) {
+			BrowserStorage.setItem('last_logged_in_user', authorizedUser)
+		}
+	},
+
 	beforeDestroy() {
 		if (!getCurrentUser()) {
 			EventBus.$off('should-refresh-conversations', this.debounceRefreshCurrentConversation)


src/App.vue
```diff
@@ -227,6 +227,16 @@ export default {
...
 
+	beforeCreate() {
+		const authorizedUser = getCurrentUser().uid
+		const lastLoggedInUser = BrowserStorage.getItem('last_logged_in_user')
+
+		if (authorizedUser !== lastLoggedInUser) {
+			BrowserStorage.clear()
+		}
+		BrowserStorage.setItem('last_logged_in_user', authorizedUser)
+	},
```
ShGKme, line 238: For guest `getCurrentUser().uid === null`.

On the first open `authorizedUser === lastLoggedInUser === null` and it works fine. But it saves a string `'null'` as `last_logged_in_user` value.

After a reload `authorizedUser !== lastLoggedInUser` because `null !== 'null'` and everything is cleared. While it's good to clear the cache for guests, maybe we shouldn't clear all the settings. Or at least do it implicitly.

ShGKme (changes requested):

src/App.vue
Antreesy: > For guest `getCurrentUser().uid === null`.

Actually, `getCurrentUser() === null` and `getCurrentUser().id` throws an error :see_no_evil: 
Thanks for noticing

> On the first open `authorizedUser === lastLoggedInUser === null` and it works fine. But it saves a string `'null'` as `last_logged_in_user` value.

Solvable by pushing a storage value through JSON.parse()

Antreesy: > Since we remember camera setting etc as well, could we just wipe the conversation list thing for now?

If we could treat a guest (non-logged) as user with uid === null, we could keep the cache. Guests don't fetch conversations, so it'll never appear for them. And if someone joins as a guest -> `'someone_uid' !== null` -> wipe

nickvergessen: Just thinking about the session issues we have atm on our instance.
If I happen to rejoin to a public conversation after hybernate, it would wipe all my camera/mic preferences

Antreesy:
In fix-up:

* don't save uid for non-autorized users: `null === null` on checking for guests, all preferences are keeped
* when there's a mismatch, clear only conversations (as it's the only part of sensitive data)
* TODO: 
  * unify sensitive data somehow

nickvergessen (approved):

src/App.vue
```diff
@@ -227,6 +227,22 @@ export default {
 		},
 	},
 
+	beforeCreate() {
+		const authorizedUser = getCurrentUser()?.uid || null
+		const lastLoggedInUser = BrowserStorage.getItem('last_logged_in_user')
+
+		if (authorizedUser !== lastLoggedInUser) {
```
ShGKme, line 234: Isn't it still comparing `null !== 'null'`?

Antreesy: We don't save uid for guests, so as long as there is nothing saved, `BrowserStorage.getItem()` returns normal `null`

PR #10317, opened 2023-08-23
Fix capabilities endpoint with openai translation integration

### LINKED ISSUE #10074: Reduce size of capabilities response with translations enabled
### How to use GitHub

---

If translations are enabled, the capabilities of Talk contain a huge amount of entries in the `translations` key. Here it was about 500 Kbytes of JSON. Most of the data in there is redundant (`fromLabel` / `toLabel`), so a smaller representation would be rather easy. Also if multiple apps will need the same data in the future, the capabilities will grow a lot again.

Compressing the response of course is easy (in my example the 500 Kbytes go down to ~24 KBytes), but services caching the capabilities of multiple servers (like the HPB or the SIP bridge) will require a lot of memory to do so.

It would be great if the size of the capabilities could be reduced for the case where translations are enabled. Maybe the list of possible translation sources / targets could be moved to a separate endpoint, so it's only transmitted if necessary.

nickvergessen:
### ☑️ Resolves

* Fix #10074 

### 🚧 Tasks

- [x] Add new boolean capability
- [x] Use boolean in Message.vue
- [x] Fix getting translation sets in MessageTranslateDialog.vue @Antreesy 
- [x] Raise issues for mobile apps @nickvergessen 
    - https://github.com/nextcloud/talk-ios/issues/1376
    - https://github.com/nextcloud/talk-android/issues/3347

### 🏁 Checklist

### DIFF
--- a/docs/capabilities.md
+++ b/docs/capabilities.md
@@ -115,7 +115,7 @@
 
 ## 17
 * `avatar` - Avatar of conversation
-* `config => chat => translations` - List of translations tuples, JSON encoded sample `{"from":"de","fromLabel":"German","to":"en","toLabel":"English"}`. Those tuples should be provided as options when translating chat messages.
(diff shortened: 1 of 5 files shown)


Antreesy (approved):
API is compatible!
I've changed a way to receive language pairs: from capabilities to server request.
That means fetching them each time, but I doubt that it will have a big impact, then creating a dedicated store with reactivity for them

PR #10326, opened 2023-08-23
fix(arrowNavigation) - update way to listen for focus/blur events

Antreesy:
### ☑️ Resolves

* Regression from #9980
* Before composable add new eventListener each time (on navigation initialization => after each fetch). Thus some items could receive several event listeners from it. Now it keep listeners in memory to remove them later (on reset method => before fetch).
* Before if search text was changed, and first item was removed from the list, but navigation is not initialized yet, on `Enter` key code tries to click on removed element, which causes page reload - fixed with additional check 

### 🖼️ Screenshots

No visual changes

### 🚧 Tasks

- [ ] Manual testing
- [ ] Code review

### 🏁 Checklist

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -3,7 +3,7 @@
   -
   - @author Marco Ambrosini <marcoambrosini@icloud.com>
   -
-  - @license GNU AGPL version 3 or any later version
+  - @license AGPL-3.0-or-later
   -
   - This program is free software: you can redistribute it and/or modify
   - it under the terms of the GNU Affero General Public License as
@@ -167,6 +167,7 @@
 							<NcAppNavigationCaption :title="t('spreed', 'Users')" />
 							<NcListItem v-for="item of searchResultsUsers"
 								:key="`user_${item.id}`"
+								:anchor-id="`user_${item.id}`"
 								:title="item.label"
 								@click="createAndJoinConversation(item)">
 								<template #icon>
@@ -182,6 +183,7 @@
 								<NcAppNavigationCaption :title="t('spreed', 'Groups')" />
 								<NcListItem v-for="item of searchResultsGroups"
 									:key="`group_${item.id}`"
+									:anchor-id="`group_${item.id}`"
 									:title="item.label"
 									@click="createAndJoinConversation(item)">
 									<template #icon>
@@ -195,6 +197,7 @@
 								<NcAppNavigationCaption :title="t('spreed', 'Circles')" />
 								<NcListItem v-for="item of searchResultsCircles"
 									:key="`circle_${item.id}`"
+									:anchor-id="`circle_${item.id}`"
 									:title="item.label"
 									@click="createAndJoinConversation(item)">
 									<template #icon>
@@ -309,7 +312,7 @@ export default {
 		const leftSidebar = ref(null)
 		const searchBox = ref(null)
 
-		const { initializeNavigation } = useArrowNavigation(leftSidebar, searchBox)
+		const { initializeNavigation } = useArrowNavigation(leftSidebar, searchBox, '.list-item')
 
 		return {
 			initializeNavigation,
@@ -539,9 +542,6 @@ export default {
 				this.searchResultsGroups = this.searchResults.filter((match) => match.source === 'groups')
 				this.searchResultsCircles = this.searchResults.filter((match) => match.source === 'circles')
 				this.contactsLoading = false
-				this.$nextTick(() => {
-					this.initializeNavigation('.list-item')
-				})
 			} catch (exception) {
 				if (CancelableRequest.isCancel(exception)) {
 					return
@@ -563,9 +563,6 @@ export default {
 				const response = await request({ searchText: this.searchText })
 				this.searchResultsListedConversations = response.data.ocs.data
 				this.listedConversationsLoading = false
-				this.$nextTick(() => {
-					this.initializeNavigation('.list-item')
-				})
 			} catch (exception) {
 				if (CancelableRequest.isCancel(exception)) {
 					return
@@ -577,6 +574,7 @@ export default {
 
 		async fetchSearchResults() {
 			await Promise.all([this.fetchPossibleConversations(), this.fetchListedConversations()])
+			this.initializeNavigation()
 		},
 
 		/**
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -312,10 +312,11 @@ export default {
 		const leftSidebar = ref(null)
 		const searchBox = ref(null)
 
-		const { initializeNavigation } = useArrowNavigation(leftSidebar, searchBox, '.list-item')
+		const { initializeNavigation, resetNavigation } = useArrowNavigation(leftSidebar, searchBox, '.list-item')
 
 		return {
 			initializeNavigation,
+			resetNavigation,
 			leftSidebar,
 			searchBox,
 		}
@@ -513,6 +514,7 @@ export default {
 			}, 500)
 		},
 		debounceFetchSearchResults: debounce(function() {
+			this.resetNavigation()
 			if (this.isSearching) {
 				this.fetchSearchResults()
 			}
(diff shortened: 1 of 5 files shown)


nickvergessen:
Doesn't seem to be caused by this, but I can not escape the conversation list with a search being done. The tabbing flow jumps back to the "clear search" instead of continuing with the list of users or groups:
[Bildschirmaufzeichnung vom 2023-08-24, 10-11-45.webm](https://github.com/nextcloud/spreed/assets/213943/fcafd263-9e18-4aac-a7fc-4d548076506f)

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -519,9 +520,13 @@ export default {
...
 		},
 		debounceFetchSearchResults: debounce(function() {
+			this.resetNavigation()
 			if (this.isSearching) {
 				this.fetchSearchResults()
 			}
+			this.$nextTick(() => {
+				this.initializeNavigation('.list-item')
+			})
 		}, 250),
```
DorraJaouad, line 530: so whenever a character is entered as input, all event listeners are removed and then added again to almost the same items ? is it better maybe to compare the previous `itemElements` and skip the common ones ?

DorraJaouad (commented):
Nothing is blocking, except for a  suggestion for a possible improvement.

src/components/LeftSidebar/LeftSidebar.vue
Antreesy: Order of items could change, which will break a flow (item with index 2 could become 6th, and focus will jump weirdly)
So it's safe to remove old listeners and add new ones, it shouldn't load browser too much for 10-20 items

DorraJaouad: The indexes will follow the new `itemElements` hence the order is correct. I am thinking about cases where a user can have +50 items for example.

ShGKme: > Order of items could change, which will break a flow (item with index 2 could become 6th, and focus will jump weirdly) So it's safe to remove old listeners and add new ones, it shouldn't load browser too much for 10-20 items

Can we listen the keydown event not on items itself but on a whole container?

ShGKme: Or create vue component with `v-on` directive instead of manually query elements and manage listeners. Then Vue will handle optimizations around managing listeners itself.

*Universal component, with a scoped-slot

Antreesy: I think that it's better to keep navigation logic in a separate composable, and not duplicate it for each component we need. Currently we use it only for two item lists, but it could be every list in the application.
In addition, that container should also be able to perform other functions (be a virtual scroller, for example). In that case, we need also to reinvent a way to work for each case.

src/composables/useArrowNavigation.js
```diff
@@ -31,16 +32,37 @@ import { onMounted, ref, unref } from 'vue'
...
 	const isConfirmationEnabled = ref(null)
 
+	// Set focused index according to selected element
+	const handleFocusEvent = (event) => {
+		const newIndex = itemElementsIdMap.value.findIndex(id => id === event.target?.id)
+
+		if (newIndex !== -1) {
+			focusedIndex.value = newIndex
+		}
+	}
```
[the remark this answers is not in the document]
DorraJaouad: https://github.com/nextcloud/spreed/assets/84044328/2992f8e1-abc1-4881-ac45-a78e04651067

I am just clicking on the arrow down key and the navigation gets trapped between two items. I don't know exactly the reason but removing  `handleFocusEvent` solved it 🤔.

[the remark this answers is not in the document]
Antreesy: Thanks for checking! I saw and fixed this bug for the joined conversations before, will check now for the rest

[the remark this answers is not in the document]
DorraJaouad: > nativelyFocusElement is handled by arrow keys, focuses the element, and (surprisingly) triggers handleFocusEvent

because there is a focus event emitted to the `listElementRef` upon `.focus()`, I think.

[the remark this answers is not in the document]
Antreesy: Duplicated mutation is fixed, as well as items trap, with adding ids to the missing ones

src/composables/useArrowNavigation.js
```diff
@@ -31,16 +32,37 @@ import { onMounted, ref, unref } from 'vue'
...
+
+	// Reset focused index if focus moved out of navigation area
+	const handleBlurEvent = (event) => {
+		if (!listRef.value.contains(event.relatedTarget)) {
+			focusedIndex.value = null
+		}
+	}
```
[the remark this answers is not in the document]
Antreesy: Fixed with checking for serachBox on Blur event

src/composables/useArrowNavigation.js
```diff
@@ -49,16 +49,20 @@ export function useArrowNavigation(listElementRef, defaultElementRef, selector,
...
 		}
 	}
 
-	// Reset focused index if focus moved out of navigation area
+	// Reset focused index if focus moved out of navigation area or moved to the defaultRef
 	const handleBlurEvent = (event) => {
-		if (!listRef.value.contains(event.relatedTarget)) {
+		if (!listRef.value.contains(event.relatedTarget)
+			|| defaultRef.value?.$el.contains(event.relatedTarget)
+			|| defaultRef.value.contains?.(event.relatedTarget)) {
```
DorraJaouad, line 65: I am not sure if I understand correctly your assumptions here 
because `defaultRef` is always assumed to be a DOM element 
```
		// depending on ref, listElementRef could be either a component or a DOM element
		listRef.value = unref(listElementRef)?.$el ?? unref(listElementRef)
		defaultRef.value = unref(defaultElementRef)
```

DorraJaouad (approved):
Thanks, it seems ready now :

- Navigation including tabbing and keys works correctly.
- Mouse focuses are processed in the same way whether they are outside or on SearchBox.
- No more extra listeners reassignments to conversations.

Tested on both lists and it functions as expected 🦅

src/composables/useArrowNavigation.js
Antreesy: That was important for `leftSidebar`, as we're using it only as DOM element in the code

`defaultRef`, on the other hand, could be either a DOM element or a Vue component, so we should check here both options:
- DOM doesn't have `$el` prop
- Vue component doesn't have `contain()` method

src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
```diff
@@ -20,7 +20,8 @@
 -->
 
 <template>
-	<li class="participant-row"
+	<li :id="`${participant.source}_${participant.id}`"
```
ShGKme, line 23: Is it really required to be `id` here and `anchor` in the `LeftSidebar`?

ID must be unique on a whole page, and I'm really not sure that IDs like `user_${item.id}` are guaranteed to be unique across all applications on the page. 

If it is actually only used as additional Node data to get item id from `event.target?.id` can we use some custom attr here? For example, `data-arrow-navigation-id` or at least `data-id`?

P.S. duplicated IDs is one of the most large a11y issue we currently have in Nextcloud...

ShGKme (changes requested):

src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
Antreesy: For the LeftSidebar they are used by RouterLink in `<Conversation/>` components, but not for the `<NcListItem/>` and `<Participant/>`
So I think, we could find a way to avoid it

ShGKme: > For the LeftSidebar they are used by RouterLink in `<Conversation/>`

But they were not used in Talk, right?

Antreesy: Router works without them, and I couldn't find anythig related or based on DOM anchor element id

They have an empty id attribute, but that's an upstream regression of NcListItem
image

ShGKme (approved):
Fine by me with change `id` to `data-nav-id`.

But I'm still not a fun of mixing vanilla DOM manipulation with Vue.

Not tested.

PR #10356, opened 2023-08-24
fix(LeftSidebar): wrong user status after scrolling

ShGKme:
### ☑️ Resolves

On scrolling user status in 1-1 conversations in the `LeftSidebar` is not correct.

- User's avatar and user status is displayed by `NcAvatar` component.
- With virtual scrolling components are reused. During scrolling new visible conversations are just given props of previous.
- It works only for components that react on props update
- `NcAvatar` supports `user` update but doesn't support `preloadedUserStatus` update

### 🚧 Tasks

- [x] Add `key` to `NcAvatar` and re-mount avatar for 1-1 conversation during scrolling.

### Drawbacks and alternative solution

It is less performant that adding support of `preloadUserStatus` change to `NcAvatar`.

Ideally it should be fixed on `@nextcloud/vue`. But I cannot say how much is the difference.

### 🏁 Checklist

### DIFF
--- a/src/components/ConversationIcon.vue
+++ b/src/components/ConversationIcon.vue
@@ -35,7 +35,11 @@
 			:height="size"
 			:alt="item.displayName"
 			class="avatar icon">
+		<!-- NcAvatar doesn't fully support props update and works only for 1 user -->
+		<!-- Using key on NcAvatar forces NcAvatar re-mount and solve the problem, could not really optimal -->
+		<!-- TODO: Check if props update support in NcAvatar is more performant -->
 		<NcAvatar v-else
+			:key="item.token"
 			:size="size"
 			:user="item.name"
 			:disable-menu="disableMenu"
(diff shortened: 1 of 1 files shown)


Antreesy (approved):
Tested with desktop client on c.nc.c, all statuses are at their own places.

follow-up: 
* does it make sense to keep img tag now?
image
* 
can we upstream handling of `preloadedUserStatus`? watch for prop changes to update should be enough, I suppose?

ShGKme:
> does it make sense to keep img tag now?

`key` on `` helps to keeps Image DOM element during scrolling and re-create it on long scrolling when `src` actually changes. It removes glitch when `src` is updated on existing image.

`key` on `<NcAvatar>` recreates a whole component instance. It also re-creates its ``. But I'd call it a temporal solution. We should not re-create `<NcAvatar>`.

So we should add status update support to `<NcAvatar>` and check if there is anything we can disable when it is not needed for performant update.

We can also consider adding `key` to `img` inside `NcAvatar`. But need to check if it won't have any drawbacks.

PR #10360, opened 2023-08-25
fix(chat): Fix primary color selection on quotes

nickvergessen:
### ☑️ Resolves

* Fix my primary color is pink, but the quotes were blue

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Bildschirmfoto vom 2023-08-25 10-13-02 | Bildschirmfoto vom 2023-08-25 10-13-10

### 🏁 Checklist

### DIFF
--- a/src/components/Quote.vue
+++ b/src/components/Quote.vue
@@ -283,7 +283,7 @@ export default {
 	}
 
 	&.quote-own-message::before {
-		background-color: var(--color-primary-default);
+		background-color: var(--color-primary-element);
 	}
 
 	&__main {


src/components/Quote.vue
```diff
@@ -283,7 +283,7 @@ export default {
 	}
 
 	&.quote-own-message::before {
-		background-color: var(--color-primary-default);
```
Antreesy, line 286: Shouldn't it be aligned with your theming as well?

https://github.com/nextcloud/server/blob/85ff1449d33df76db17d741e46485799e7121977/apps/theming/lib/Themes/CommonThemeTrait.php#L58

Antreesy (approved):
Fine as a quick fix, but not sure, whether it should be upstreamed or not

src/components/Quote.vue
nickvergessen: No, default is default blue
The other option would have been `--color-primary` but that could be white on white, etc. So we have to use element here

Issue #9202, opened 2023-03-29
Talk web test coverage


Antreesy:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---
I would like to define specific files (components, stores, service) to cover it prior to moving from Vue 2.7 to Vue 3

This is a current test coverage on spreed repositories:
Components                                                                                            | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
------------------------------------------------------------------------------------------------|---------|----------|---------|---------|---
  ChatView.vue                                                                                  |    4.16 |        0 |       0 |    4.16 | 74-135                                                
  ConversationIcon.vue                                                                          |      68 |       48 |   83.33 |      68 | 87,114,116,118,120,122,134,146                        
  ConversationsOptionsList.vue                                                                  |     100 |      100 |     100 |     100 |                                                       
  Hint.vue                                                                                      |     100 |      100 |     100 |     100 |                                                       
  LoadingPlaceholder.vue                                                                        |     100 |      100 |     100 |     100 |                                                       
  LobbyScreen.vue                                                                               |    7.69 |        0 |       0 |    7.69 | 71-108                                                
  MediaDevicesSelector.vue                                                                      |       0 |        0 |       0 |       0 | 62-193                                                
  Quote.vue                                                                                     |    3.22 |        0 |       0 |    3.22 | 111-242                                               
  SetGuestUsername.vue                                                                          |    2.63 |        0 |       0 |    2.94 | 56,73-155                                             
  UploadEditor.vue                                                                              |       0 |        0 |       0 |       0 | 98-188                                                
  AllowedGroups.vue                                                                             |       0 |        0 |       0 |       0 | 104-238                                               
  CommandVue.vue                                                                                |       0 |        0 |       0 |       0 | 81-99                                                 
  Commands.vue                                                                                  |       0 |      100 |       0 |       0 | 69-83                                                 
  GeneralSettings.vue                                                                           |       0 |        0 |       0 |       0 | 64-142                                                
  HostedSignalingServer.vue                                                                     |       0 |        0 |       0 |       0 | 135-253                                               
[...]

PR #10393, opened 2023-08-28
techdebt(Jest) - introduce response helper, clear output, raise coverage

Antreesy:
### ☑️ Resolves

* add test helpers `generateOCSResponse` and `generateOCSErrorResponse` to ease codebase, unify request mocks and enhance typings
* mock `IntersectionObserver` and `@nextcloud/initial-state` to reduce amount of senseless logs in jest output
* update tests for :pineapple: Pinia stores, increase coverage (part of #9202) 

### 🖼️ Screenshots
No visual changes

Jest output:
🏚️ Before | 🏡 After
---|---
**3 212 lines** (323 lines - test coverage) | **634 lines** :arrow_down_small: 90%  (323 lines - test coverage)

### 🚧 Tasks

- [ ] Code review

### 🏁 Checklist

### DIFF
--- a/src/stores/__tests__/bots.spec.js
+++ b/src/stores/__tests__/bots.spec.js
@@ -1,0 +1,92 @@
+import { setActivePinia, createPinia } from 'pinia'
+
+import {
+	getConversationBots,
+	disableBotForConversation,
+	enableBotForConversation,
+} from '../../services/botsService.js'
+import { generateOCSResponse } from '../../test-helpers.js'
+import { useBotsStore } from '../bots.js'
+
+jest.mock('../../services/botsService', () => ({
+	getConversationBots: jest.fn(),
+	disableBotForConversation: jest.fn(),
+	enableBotForConversation: jest.fn(),
+}))
+
+describe('botsStore', () => {
+	const token = 'TOKEN'
+	const bots = [
+		{ id: 1, name: 'Dummy bot 1', state: 0 },
+		{ id: 2, name: 'Dummy bot 2', state: 1 },
+	]
+	let botsStore
+
+	beforeEach(async () => {
+		setActivePinia(createPinia())
+		botsStore = useBotsStore()
+	})
+
+	afterEach(async () => {
+		jest.clearAllMocks()
+	})
+
+	it('returns an empty array when bots are not loaded yet for conversation', async () => {
+		// Assert: check initial state of the store
+		expect(botsStore.getConversationBots(token)).toEqual([])
+	})
+
+	it('processes a response from server and stores bots', async () => {
+		// Arrange
+		const response = generateOCSResponse({ payload: bots })
+		getConversationBots.mockResolvedValueOnce(response)
+
+		// Act: load bots from server
+		await botsStore.loadConversationBots(token)
+
+		// Assert
+		expect(getConversationBots).toHaveBeenCalledWith(token)
+		expect(botsStore.getConversationBots(token)).toEqual(bots)
+	})
+
+	it('updates bots in the store after new response from server', async () => {
+		// Arrange
+		const newBot = { id: 3, name: 'Dummy bot 3', state: 2 }
+		const responseFirst = generateOCSResponse({ payload: bots })
+		const responseSecond = generateOCSResponse({ payload: [...bots, newBot] })
+		getConversationBots
+			.mockResolvedValueOnce(responseFirst)
+			.mockResolvedValueOnce(responseSecond)
+
+		// Act: load bots from server twice
+		await botsStore.loadConversationBots(token)
+		await botsStore.loadConversationBots(token)
+
+		// Assert
+		expect(getConversationBots).toHaveBeenCalledTimes(2)
+		expect(botsStore.getConversationBots(token)).toEqual([...bots, newBot])
+	})
+
+	it('toggles bots state according to their current state', async () => {
+		// Arrange
+		const response = generateOCSResponse({ payload: bots })
+		getConversationBots.mockResolvedValueOnce(response)
+		await botsStore.loadConversationBots(token)
+
+		const enabledBot = { ...bots[0], state: 1 }
+		const responseEnabled = generateOCSResponse({ payload: enabledBot })
+		enableBotForConversation.mockResolvedValueOnce(responseEnabled)
+		const disabledBot = { ...bots[1], state: 0 }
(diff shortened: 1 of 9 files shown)


src/stores/__tests__/bots.spec.js
```diff
@@ -0,0 +1,83 @@
...
+
+describe('botsStore', () => {
+	beforeEach(async () => {
+		setActivePinia(createPinia())
+	})
+
+	it('returns an empty array if bots are not available', async () => {
+		const botsStore = useBotsStore()
+		expect(botsStore.getConversationBots('empty-token')).toEqual([])
+	})
```
ShGKme, line 25: It says that it tests stores when "bots are not available" but neither this test nor others actually touch bots "availability". The store doesn't check the capability `bots-v1`.

I think it should be renamed to something like `returns empty bots array when they not loaded`. Then it would also document not obvious behavior - that both conversation without bots and conversation with not loaded bots have `[]` as bots list, bot `[]` and `null`.

src/stores/__tests__/bots.spec.js
```diff
@@ -0,0 +1,83 @@
...
+
+		// Disable bot with id: 2
+		const disabledBot = { ...bots[1], state: 0 }
+		const responseDisabled = generateOCSResponse({ payload: disabledBot })
+		disableBotForConversation.mockResolvedValueOnce(responseDisabled)
+
+		await botsStore.toggleBotState(token, bots[1])
+		expect(disableBotForConversation).toHaveBeenCalledWith('TOKEN', bots[1].id)
+		expect(botsStore.getConversationBots('TOKEN')).toEqual([enabledBot, disabledBot])
+	})
```
ShGKme, line 82:
```suggestion
	it('toggles bots state', async () => {
		const token = 'TOKEN'
		const botsStore = useBotsStore()

		const bots = [
			{ id: 1, name: 'Dummy bot 1', state: 0 },
			{ id: 2, name: 'Dummy bot 2', state: 1 },
		]
		const response = generateOCSResponse({ payload: bots })
		getConversationBots.mockResolvedValueOnce(response)
		await botsStore.loadConversationBots(token)

		// Enable bot with id: 1
		const enabledBot = { ...bots[0], state: 1 }
		const responseEnabled = generateOCSResponse({ payload: enabledBot })
		enableBotForConversation.mockResolvedValueOnce(responseEnabled)
		await botsStore.toggleBotState(token, bots[0])

		// Disable bot with id: 2
		const disabledBot = { ...bots[1], state: 0 }
		const responseDisabled = generateOCSResponse({ payload: disabledBot })
		disableBotForConversation.mockResolvedValueOnce(responseDisabled)
		await botsStore.toggleBotState(token, bots[1])

		expect(enableBotForConversation).toHaveBeenCalledWith(token , bots[0].id)		
		expect(disableBotForConversation).toHaveBeenCalledWith(token , bots[1].id)
		expect(botsStore.getConversationBots(token)).toEqual([enabledBot, disabledBot])
	})
```

ShGKme: In unit tests it is a good practice to follow the AAA pattern:
1. Arrange - prepare the environment for the test, set up the initial state and constants:
   - Setup initial store state
2. Act - make the action being tested
3. Assert - do the actual check

Then it is clear what exactly in what case is tested. And it saves from creating tests that are actually a sequence of several tests.

It is different from the e2e test where we test not units but long user scenarios with many actions and complex initial setup.

src/stores/__tests__/bots.spec.js
```diff
@@ -0,0 +1,83 @@
...
+
+		// New response with updated bots list
+		const newBot = { id: 3, name: 'Dummy bot 3', state: 2 }
+		response = generateOCSResponse({ payload: [...oldBots, newBot] })
+		getConversationBots.mockResolvedValueOnce(response)
+		await botsStore.loadConversationBots(token)
+
+		expect(getConversationBots).toHaveBeenCalledWith(token)
+		expect(botsStore.getConversationBots(token)).toEqual([...oldBots, newBot])
+	})
```
ShGKme, line 51: For example, this test definitely is not one test. It testing failed and in the report we have
```
❌ receives a response from server, stores and updates bots
```

It is hard to understand, what was broken. Does it not load bots, or doesn't store them, or are there issues with updates?

It should be a load conversation bots test + update a bot test.

If many tests have exactly the same initial state and arrangement, it can be reused with `beforeEach`.

src/test-helpers.js
```diff
@@ -72,8 +72,53 @@ function findNcListItems(wrapper, text) {
...
+		response: {
+			status,
+			data: {
+				ocs: {
+					data: Array.isArray(payload) ? [...payload] : { ...payload },
+				},
+			},
+		},
+	}
+}
```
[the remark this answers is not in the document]
Antreesy: I faced a couple of times a situation, when server response was mutated, and test passed because of that, although it didn't.
It probably takes more resources to compare output with expected, than by reference, but i think, that it's better to have the mocked data immutable

Changed to `cloneDeep` until further discussion

src/stores/__tests__/bots.spec.js
```diff
@@ -0,0 +1,92 @@
...
+		const response = generateOCSResponse({ payload: bots })
+		getConversationBots.mockResolvedValueOnce(response)
+
+		// Act: load bots from server
+		await botsStore.loadConversationBots(token)
+
+		// Assert
+		expect(getConversationBots).toHaveBeenCalledWith(token)
+		expect(botsStore.getConversationBots(token)).toEqual(bots)
+	})
```
ShGKme, line 50: I didn't mean that we must literally add `// A*` comments =D

(But fine by me)

src/test-helpers.js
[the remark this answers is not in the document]
ShGKme: > I faced a couple of times a situation, when server response was mutated, and test passed because of that, although it didn't.

I'd say that this is an issue of the test. For example, using one object for response twice and probably making several actual tests inside one test.

ShGKme (approved):
Thanks for cleaning up the output log, that was annoying 🧹🧹🧹

src/stores/__tests__/bots.spec.js
Antreesy: I feel it to be more readable / understandable for the following development

Antreesy:
Rebased, squashed fixups only

PR #10444, opened 2023-08-31
Remove unneccessary output from php process

SystemKeeper:
### ☑️ Resolves

* Removes unneccessary output of "Accepted" and "Closed" messages from the PHP dev server.
* Fixes the trap. Since we can have only one trap per signal, this is now combined.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

### DIFF
--- a/tests/integration/run.sh
+++ b/tests/integration/run.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+PROCESS_ID=$$
+
 APP_NAME=spreed
 NOTIFICATIONS_BRANCH="master"
 GUESTS_BRANCH="master"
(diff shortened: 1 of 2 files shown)


tests/integration/run.sh
```diff
@@ -17,12 +19,14 @@ echo ''
...
-PHP_CLI_SERVER_WORKERS=3 php -S localhost:8080 -t ${ROOT_DIR} &
+touch phpserver.log
+PHP_CLI_SERVER_WORKERS=3 php -S localhost:8080 -t ${ROOT_DIR} &> phpserver.log &
 PHPPID1=$!
 echo -e "Running on process ID: \033[1;35m$PHPPID1\033[0m"
 
-# also kill php process in case of ctrl+c
-trap 'pkill -P $PHPPID1; kill -TERM $PHPPID1; wait $PHPPID1' TERM
+# Output filtered php server logs
+tail -f phpserver.log | grep --line-buffered -v " Accepted" | grep --line-buffered -v " Closing" &
```
nickvergessen, line 28:
```suggestion
tail -f phpserver.log | grep --line-buffered -v -E ":\d+ Accepted$" | grep --line-buffered -v -E ":\d+ Closing$" &
```

Not sure regex works in `grep -E` the same way, but filtering out based on the word only looks a bit risky, so let's try to make sure it is really only the word and before there is a port?

Edit: Or it's okay, because all other output is actually from behat and not the php build in server.

Can you do the same on the federation build in server below (line 31/35)

tests/integration/run.sh
```diff
@@ -17,12 +19,14 @@ echo ''
 echo -e "\033[0;36m#\033[0m"
 echo -e "\033[0;36m# Starting PHP webserver\033[0m"
 echo -e "\033[0;36m#\033[0m"
-PHP_CLI_SERVER_WORKERS=3 php -S localhost:8080 -t ${ROOT_DIR} &
+touch phpserver.log
```
nickvergessen, line 22: To clear on repeating attempts locally:
```suggestion
echo "" > phpserver.log
```
And add to `.gitignore`?

tests/integration/run.sh
nickvergessen: Can you do the same on the federation build in server below (line 31/35)
Maybe to a separate file for easier debugging

SystemKeeper: > Edit: Or it's okay, because all other output is actually from behat and not the php build in server.

Yes, it's only from the PHP server, but we can match with the suggested regex to be sure!

> Can you do the same on the federation build in server below (line 31/35)

Yep, good idea.

tests/integration/run.sh
SystemKeeper: Good idea - most of the time I only think about CI...
Should we put the file in `/tmp` to have it outside of the repo?

nickvergessen: No, inside is good so it's findable 🙈

tests/integration/run.sh
SystemKeeper: Not completely sure why `\d` isn't working on the runners, but `[0-9]` is working fine, so using this.

SystemKeeper:
Done as discussed. In the conversation test we are now down from approx. 18000 to 8000 lines.

nickvergessen (approved):

PR #10467, opened 2023-09-05
Add a welcome window to set username for guests

### LINKED ISSUE #855: 👤 Ask Guest to enter a name when connecting
If a guest joins a conversation with a public link, ask him to enter a name.

I know you can change it here: 
image

But i think the most won't do this... so why don't ask him when he connects.

DorraJaouad:
### ☑️ Resolves

* Fix #855
- The window is shown only when there is no username set previously. 

### 🖼️ Screenshots

🏚️ Chat View | 🏡 Lobby view
---|---
image | image

### 🚧 Tasks

- [ ] visual review
- [ ] code review

### 🏁 Checklist

### DIFF
--- a/src/components/ChatView.vue
+++ b/src/components/ChatView.vue
@@ -23,7 +23,7 @@
 		@dragover.prevent="handleDragOver"
 		@dragleave.prevent="handleDragLeave"
 		@drop.prevent="handleDropFiles">
-		<GuestWelcomeWindow v-if="isGuestAndhasNotUserName" />
+		<GuestWelcomeWindow v-if="isGuestAndhasNotUsername" :token="token" />
 		<TransitionWrapper name="slide-up" mode="out-in">
 			<div v-show="isDraggingOver"
 				class="dragover">
@@ -113,8 +113,8 @@ export default {
 			return this.$store.getters.getActorType() === 'guests'
 		},
 
-		isGuestAndhasNotUserName() {
-			const userName = localStorage.getItem('nick')
+		isGuestAndhasNotUsername() {
+			const userName = this.$store.getters.getDisplayName()
 			return !userName && this.isGuest
 		},
 
--- a/src/components/ChatView.vue
+++ b/src/components/ChatView.vue
@@ -23,6 +23,7 @@
 		@dragover.prevent="handleDragOver"
 		@dragleave.prevent="handleDragLeave"
 		@drop.prevent="handleDropFiles">
+		<GuestWelcomeWindow v-if="isGuestAndhasNotUserName" />
 		<TransitionWrapper name="slide-up" mode="out-in">
 			<div v-show="isDraggingOver"
 				class="dragover">
@@ -70,6 +71,7 @@ import ChevronDoubleDown from 'vue-material-design-icons/ChevronDoubleDown.vue'
 
 import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
 
+import GuestWelcomeWindow from './GuestWelcomeWindow.vue'
 import MessagesList from './MessagesList/MessagesList.vue'
 import NewMessage from './NewMessage/NewMessage.vue'
 import TransitionWrapper from './TransitionWrapper.vue'
@@ -87,6 +89,7 @@ export default {
 		MessagesList,
 		NewMessage,
 		TransitionWrapper,
+		GuestWelcomeWindow,
 	},
 
 	props: {
@@ -109,6 +112,12 @@ export default {
 		isGuest() {
 			return this.$store.getters.getActorType() === 'guests'
 		},
+
+		isGuestAndhasNotUserName() {
+			const userName = localStorage.getItem('nick')
+			return !userName && this.isGuest
+		},
+
 		dropHintText() {
 			if (this.isGuest) {
 				return t('spreed', 'You need to be logged in to upload files')
--- a/src/components/ChatView.vue
+++ b/src/components/ChatView.vue
@@ -23,7 +23,7 @@
 		@dragover.prevent="handleDragOver"
 		@dragleave.prevent="handleDragLeave"
 		@drop.prevent="handleDropFiles">
-		<GuestWelcomeWindow v-if="isGuestAndhasNotUsername" :token="token" />
+		<GuestWelcomeWindow v-if="isGuestWithoutDisplayName" :token="token" />
 		<TransitionWrapper name="slide-up" mode="out-in">
 			<div v-show="isDraggingOver"
 				class="dragover">
@@ -113,7 +113,7 @@ export default {
 			return this.$store.getters.getActorType() === 'guests'
 		},
 
-		isGuestAndhasNotUsername() {
+		isGuestWithoutDisplayName() {
 			const userName = this.$store.getters.getDisplayName()
 			return !userName && this.isGuest
 		},
--- a/src/components/GuestWelcomeWindow.vue
+++ b/src/components/GuestWelcomeWindow.vue
@@ -1,0 +1,149 @@
+<!--
+  - @copyright Copyright (c) 2022, Dorra Jaouad <dorra.jaoued1@gmail.com>
+  -
+  - @author Dorra Jaouad <dorra.jaoued1@gmail.com>
+  -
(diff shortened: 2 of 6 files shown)


szaimen (commented):
Hi @DorraJaouad looks pretty good already but I have one small suggestions

szaimen (approved):
Fine by me from design perspective :)

src/components/GuestWelcomeWindow.vue
```diff
@@ -0,0 +1,149 @@
...
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+
+<template>
+	<NcModal v-if="showModal"
```
Antreesy, line 24: if userName is reactive, we don't need `v-if` here

src/components/GuestWelcomeWindow.vue
```diff
@@ -0,0 +1,149 @@
...
+			showModal: true,
+		}
+	},
+
+	computed: {
+		container() {
+			return this.$store.getters.getMainContainerSelector()
+		},
+
+		token() {
```
Antreesy, line 91: we coul pass token as a prop, and don't call a store one more time

src/components/GuestWelcomeWindow.vue
```diff
@@ -0,0 +1,149 @@
...
+			return this.$store.getters.getToken()
+		},
+
+		conversation() {
+			return this.$store.getters.conversation(this.token)
+		},
+
+		conversationDisplayName() {
+			return this.conversation && this.conversation.displayName
```
Antreesy, line 100:
```suggestion
			return this.conversation?.displayName
```

src/components/GuestWelcomeWindow.vue
```diff
@@ -0,0 +1,149 @@
...
+		},
+
+		conversationDescription() {
+			return this.conversation && this.conversation.description
```
Antreesy, line 104:
```suggestion
			return this.conversation?.description
```

src/components/GuestWelcomeWindow.vue
```diff
@@ -0,0 +1,149 @@
...
+</script>
+
+<style lang="scss" scoped>
+.modal__content {
+	padding: calc(var(--default-grid-baseline) * 4);
+	background-color: var(--color-main-background);
+
+	:deep(.username-form__input) {
+		width: auto !important;
+	}
```
Antreesy, line 128: That's overriding of `<SetGuestUsername />` styles, why don't do it in the component itself? You have a prop `firstWelcome` for that already

src/components/GuestWelcomeWindow.vue
```diff
@@ -0,0 +1,149 @@
...
+.description {
+	margin: 0px 12px 12px 12px;
+}
+
+.input-label {
+	margin: 0px 12px;
+}
+
+.submit-button {
+	margin: 0 auto;
```
Antreesy, line 147: Maybe, center all the content, with modal_content as flex container? And thus we don't need `conversation-information` wrapper at all

src/components/GuestWelcomeWindow.vue
```diff
@@ -0,0 +1,149 @@
...
+		:close-on-click-outside="false"
+		size="small">
+		<div class="modal__content">
+			<div class="conversation-information">
+				<ConversationIcon :item="conversation"
+					:show-user-status="false"
+					:disable-menu="true" />
+				<h2>{{ conversationDisplayName }}</h2>
+			</div>
+			<h3 class="description">
```
Antreesy, line 36: I'd go for stylized `<p>` tag here instead of heading, because of semantic

src/components/GuestWelcomeWindow.vue
```diff
@@ -0,0 +1,149 @@
...
+				{{ conversationDescription }}
+			</h3>
+
+			<h3 class="input-label">
+				{{ t('spreed', 'Enter your name') }}
+			</h3>
+			<SetGuestUsername ref="setGuestUsername"
+				class="guest-user-form"
+				first-welcome
```
Antreesy, line 45: Maybe don't use this component here, NcTextField will be enough? That means copy-pasting the method, but if you don't use a half of component functionality (NcButton), why do we need it?

src/components/GuestWelcomeWindow.vue
```diff
@@ -0,0 +1,149 @@
...
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+
+<template>
+	<NcModal v-if="showModal"
+		:container="container"
+		:can-close="false"
+		:close-on-click-outside="false"
```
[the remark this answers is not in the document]
szaimen: I disagree. this was the whole idea around this feature, no?

src/components/GuestWelcomeWindow.vue
DorraJaouad: It means basically copying more than half of the `SetGuestUsername` component if I go with simply NcTextField. It seems more like a duplication .

src/components/GuestWelcomeWindow.vue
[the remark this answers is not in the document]
Antreesy: Maybe it was, no one explained me 😁. Looking at the issue, I assume that behavior was changed to the current one long time ago, for some reasons, which we should consider.

Nothing blocking from my side, just a note for the future, if that feature will reveal some other hidden problems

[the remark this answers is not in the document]
szaimen: I see. It was my fault then to explain that the modal should be blocking until the name was entered. Sorry for that!

DorraJaouad:
Rebased onto master

src/components/ChatView.vue
```diff
@@ -109,6 +112,12 @@ export default {
 		isGuest() {
 			return this.$store.getters.getActorType() === 'guests'
 		},
+
+		isGuestAndhasNotUsername() {
```
Antreesy, line 116:
```suggestion
		isGuestAndHasNotUsername() {
```

src/components/GuestWelcomeWindow.vue
```diff
@@ -0,0 +1,156 @@
...
+		conversationDisplayName() {
+			return this.conversation?.displayName
+		},
+
+		conversationDescription() {
+			return this.conversation?.description
+		},
+
+		submitMessage() {
+			return t('spreed', 'Submit name and join')
```
Antreesy, line 119: If it doesn't require a logic, we can simply put `t()` function straight into the template

src/components/GuestWelcomeWindow.vue
```diff
@@ -0,0 +1,156 @@
...
+			</p>
+
+			<p class="input-label">
+				{{ t('spreed', 'Enter your name') }}
+			</p>
+
+			<NcTextField :value.sync="guestUserName"
+				:placeholder="t('spreed', 'Guest')"
+				class="username-form__input"
+				:show-trailing-button="false"
```
Antreesy, line 46:
```suggestion
				:show-trailing-button="false"
				:label="t('spreed', 'Enter your name')"
```
To resolve https://github.com/nextcloud/spreed/pull/10467#discussion_r1315980609

src/components/GuestWelcomeWindow.vue
```diff
@@ -0,0 +1,156 @@
...
+					disable-menu />
+				<h2>{{ conversationDisplayName }}</h2>
+			</div>
+			<p class="description">
+				{{ conversationDescription }}
+			</p>
+
+			<p class="input-label">
+				{{ t('spreed', 'Enter your name') }}
+			</p>
```
Antreesy, line 41: We shoud either use prop `label` or at least native label tag

Antreesy (commented):
Tiny things before you'll push again with tests

szaimen:
🎉🎉🎉

PR #10470, opened 2023-09-05
fix(guestUsername)- Migrate to browserStorage

DorraJaouad:
The guest username is now stored in `guestName` in browserStorage.


Antreesy:
By migrating I also meant modifying  `src/init.js` function:
```const migrateDirectLocalStorageToNextcloudBrowserStorage = () => {
```

Antreesy (dismissed):
Code-wise looks good
Need to be tested before backporting

Antreesy:
Sorry, I didn't check that `localStorage.getItem('nick')` is being used across all apps. (FIXME comment):
* https://github.com/nextcloud/nextcloud-browser-storage/issues/3

To keep a guest name persistent for the user, we have to wait for issue to resolve.
Duplicate and store the same value in both localStorage and BS at the moment doesn't make much sense

PR #10475, opened 2023-09-06
(guestNameStore) - Migrate to Pinia in guestNameStore

DorraJaouad:
The guestNameStore is officially a Pinia store and the guest name is now stored as expected ( previously, only the token was stored) 

Tests are adjusted.


src/store/guestNameStore.js
```diff
@@ -20,86 +20,88 @@
...
 	 * @return {Array} the participants array (if there are participants in the store)
 	 */
-	getGuestName: (state) => (token, actorId) => {
-		if (state.guestNames[token] && state.guestNames[token][actorId]) {
-			return state.guestNames[token][actorId]
+    getGuestName(state) {
+		return (token, actorId) => {
+		  if (state.guestNames[token] && state.guestNames[token][actorId]) {
+			return state.guestNames[token][actorId] 
+		  }
```
Antreesy, line 42:
```suggestion
    getGuestName(token, actorId) {
			return this.guestNames[token]?.[actorId]  ?? t('spreed', 'Guest')
	}
```

Let's simplify that

`state` and `getters` should be available with `this.`

src/store/guestNameStore.js
```diff
@@ -20,86 +20,88 @@
...
-			Vue.set(state.guestNames[token], actorId, t('spreed', 'Guest'))
-		} else if (noUpdate) {
-			return
-		}
-		state.guestNames[token][actorId] = actorDisplayName
-	},
-}
-
-const actions = {
+		addGuestName({ noUpdate, token, actorId, actorDisplayName }) {
```
Antreesy, line 71: `addGuestName`, `setGuestNameIfEmpty`, `forceGuestName` - are basically one action, with one additional prop `noUpdate` - maybe keep only first one?

src/store/guestNameStore.js
```diff
@@ -20,86 +20,88 @@
  *
  */
 import Vue from 'vue'
+import { defineStore } from 'pinia'
+
+export const useGuestNameStore = defineStore('guestNameStore', {
```
Antreesy, line 25:
```suggestion
export const useGuestNameStore = defineStore('guestName', {
```

See `src/stores/bots.js` or `src/stores/settings.js` for example
* Pinia store should be moved to the different folder, as others, and renamed to `guestName` for simplicity
* Vuex store usages should be removed from the `src/store/storeConfig.js`
* Test coverage - 100%

src/store/guestNameStore.js
DorraJaouad: next time :')

src/components/MessagesList/MessagesGroup/MessagesGroup.spec.js
```diff
@@ -1,27 +1,34 @@
...
 		localVue.use(Vuex)
+		pinia = createPinia()
+		setActivePinia(pinia)
+
+		const guestNameStore = useGuestNameStore()
 
 		testStoreConfig = cloneDeep(storeConfig)
 		getGuestNameMock = jest.fn()
-		testStoreConfig.modules.guestNameStore.getters.getGuestName = () => getGuestNameMock
+		guestNameStore.getGuestName = getGuestNameMock
```
Antreesy, line 31: You added a store here, so you can directly use it for test `renders guest display name`:
```js
	// Arrange
	guestNameStore.addGuestName({
		token: TOKEN,
		actorId: 'actor-1',
		actorDisplayName: 'guest-one-display-name',
	})
```

And remove mock from the file completely - we've tested it in store test already:
```js
	// Assert
-	expect(guestNameStore.getGuestName).toHaveBeenCalledWith(TOKEN, 'actor-1')
```

Antreesy (approved):
Seems to work, store is filled from messages and participant list.

Not a part of PR, just a note: messages don't keep display names of guests at the moment of sending, and `actorId` is reset on each page reload - but should be fine within one session / webinar 
image

PR #10494, opened 2023-09-12
feat(ParticipantsTab) - virtual scrolling for participants list

### LINKED ISSUE #5907: Stop using image-avatars in huge rooms
### Steps
1. Create 1k users and add to a dummy group:
```sh
#!/bin/sh

php occ -vvv group:add "dummies" --display-name "1k dummies"

i=10
while [ $i -ne 1000 ]
do
        i=$(($i+1))
        echo "$i"
        
        
	USERID=$(uuidgen)
	OC_PASS="123456" php occ -vvv user:add --password-from-env --display-name "Dummy #$i" -g "dummies" $USERID
done
```

2. Create a new conversation and invite the group as attendee
3. See your browser die as it requests the 1k avatars for the participant list and the chat log.

### Thoughts

At some point certain things make no sense anymore, so there are some thoughts:

* Requesting 1k images as avatars in a sidebar that is not usable
  => I think we could fall back to the 2 letter js solution?
* Starting with 1k system messages that an individual user was added
  => Skip the list when you are adding a group/circle at the moment and assume it is okay to only list the group name
* Stop automatically refreshing the participant list
  => instead have a manual button?
* Listing all participants
  => There is a search already, so maybe we just show moderators and a label "X participants" until a search term is used

Antreesy:
### ☑️ Resolves

* Fix #5907 
* Based on #10262 and #10356
  * Same performance issues with avatar requests and long list re-rendering (and initial rendering)
* Fix `data-nav-id="undefined_undefined"` for `<Participant />` components without arrow navigation

### 🖼️ Screenshots

No visual changes

[virtual scroll.webm](https://github.com/nextcloud/spreed/assets/93392545/767b54a6-a2df-4aff-bf71-5fca0131fd71)

### 🚧 Tasks

- [ ] Code review
- [ ] Manual testing

### 🏁 Checklist

### DIFF
--- a/src/components/RightSidebar/Participants/ParticipantsList/ParticipantsListVirtual.vue
+++ b/src/components/RightSidebar/Participants/ParticipantsList/ParticipantsListVirtual.vue
@@ -1,0 +1,82 @@
+<!--
+  - @copyright Copyright (c) 2023 Maksim Sukharev <antreesy.web@gmail.com>
+  -
+  - @author Grigorii Shartsev <me@shgk.me>
+  -
+  - @license AGPL-3.0-or-later
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -->
+
+<template>
+	<RecycleScroller ref="scroller"
+		list-tag="ul"
+		item-tag="li"
+		:items="participants"
+		:item-size="PARTICIPANT_ITEM_SIZE"
+		key-field="attendeeId">
+		<template #default="{ item }">
+			<Participant :participant="item" tag="div" />
+		</template>
+		<template v-if="loading" #after>
+			<LoadingPlaceholder type="participants" :count="dummyParticipants" />
+		</template>
+	</RecycleScroller>
+</template>
+
+<script>
+import { RecycleScroller } from 'vue-virtual-scroller'
+
+import LoadingPlaceholder from '../../../LoadingPlaceholder.vue'
+import Participant from './Participant/Participant.vue'
+
+import 'vue-virtual-scroller/dist/vue-virtual-scroller.css'
+
+const PARTICIPANT_ITEM_SIZE = 64
+
+export default {
+	name: 'ParticipantsListVirtual',
+
+	components: {
+		LoadingPlaceholder,
+		Participant,
+		RecycleScroller,
+	},
+
+	props: {
+		participants: {
+			type: Array,
+			required: true,
+		},
+
+		loading: {
+			type: Boolean,
+			default: false,
+		},
+	},
+
+	setup() {
+		return {
+			PARTICIPANT_ITEM_SIZE,
+		}
+	},
+
+	computed: {
+		dummyParticipants() {
+			const dummies = 6 - this.participants.length
+			return dummies > 0 ? dummies : 0
+		},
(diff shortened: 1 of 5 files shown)


src/components/RightSidebar/Participants/ParticipantsTab.vue
```diff
@@ -260,6 +265,14 @@ export default {
...
 
 <style scoped>
+.h-100 {
+	height: 100%;
+}
+
+/** Subtract search field height **/
+.h-crop {
+	height: calc(100% - 36px);
+}
```
ShGKme, line 275: We can use flexbox here to not calculate height in px, but flex content.

src/components/RightSidebar/Participants/ParticipantsTab.vue
```diff
@@ -20,33 +20,36 @@
...
+				:loading="!participantsInitialised"
+				class="h-100" />
+
+			<template v-else>
+				<NcAppNavigationCaption v-if="canAdd" :title="t('spreed', 'Participants')" />
+
+				<ParticipantsList v-if="filteredParticipants.length"
+					:items="filteredParticipants"
+					:loading="!participantsInitialised" />
+				<Hint v-else :hint="t('spreed', 'No search results')" />
```
ShGKme, line 43: Cannot we use virtual scrolling also for search results here?

src/components/RightSidebar/Participants/ParticipantsList/ParticipantsListVirtual.vue
```diff
@@ -0,0 +1,75 @@
...
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -->
+
+<template>
+	<RecycleScroller ref="scroller"
+		item-tag="ul"
```
ShGKme, line 24: It makes structure:
```
ul
  li
    ul
      li
        participant1
  li
    ul
      li
        participant2
  li
    ul
      li
        participant3    
```

In `LeftSidebar` this structure was the only possible, because `NcListItem` always renders as `li` element and `NcAppSidebar` always renders as `<ul>`. But here we can use any layout, making an entire scroller a `ul` element and each item true `li` element.

src/components/RightSidebar/Participants/ParticipantsTab.vue
Antreesy: Won't it be the same problem, as in LeftSidebar? I meant that we use here different components with unknown amount of elements, like filtered conversations + open conversations + e.t.c

ShGKme: Yep, you are right. I missed `ParticipantsSearchResults` below

src/components/RightSidebar/Participants/ParticipantsList/ParticipantsListVirtual.vue
Antreesy: `<Participant />` is already a `li` element, which is used in flat `ul` lists (ParticipantsList, BreakoutRoomItem)

RecycleScroller creates some additional wrappers by default:
image

the best we could do is
```html
<div virtual-scroller>
  <li>
    <ul>
      <Participant>
```

ShGKme: > `<Participant />` is already a `li` element,

Yes, but it is our component, and we can easily make it configurable with dynamic root node
```vue
<component :is="tag">
</component>

<script>
export default {
  props: {
    tag: {
      type: String,
      default: 'li',
    }
  }
}
</script>
```

It could also be useful for `BreakoutRoomItem` where the structure is also `ul>li>ul>li>1-item`

Then in RecycleScroller we can use `list-tag="ul" item-tag="li"`

Antreesy: Updated, but `BreakoutRoomItem` was fine, it has structure `ul "BR tab" > li "BR item" > ul "participants list" > <Participant tag="li"/>`

ShGKme (approved):

DorraJaouad (approved):
Tested

ShGKme:
Backport?

PR #10523, opened 2023-09-17
fix(participantsStore) - handle participants fetch through the store

### LINKED ISSUE #10497: [Bug] Participants list is fetched twice from server
### How to use GitHub

---
## Overview
Mixin `getParticipants.js` doesn't share data between components (`TopBar` and `ParticipantsTab`), so they are not aware of CancelableRequest processing already => create a new one, resulting in second server request

Possible steps to solve:
- [ ] implement logic from messageStore - it keeps cancelable requests in store, making them available for all instance
- [ ] move logic from mixin to the participantsStore
- [ ] Replace mixin with composable

## Steps to reproduce
1. Open DevTools - Network
2. Join group conversation
3. See result

### Expected behaviour
image

### Actual behaviour
image

## Talk app

**Talk app version:** master and below (code is there since 2021: 449260f9435bece772b0c4d025152157db7c3e93)

Antreesy:
### ☑️ Resolves

- Fix #10497
  - Previously for users and guest moderators participants were fetched twice, from `RightSidebar` and `TopBar`
  - `CancelledRequest` logic was there, but because it was in mixin, it doesn't work for different components as expected
  - TopBar duplicates request for participants counter in call only - it's not needed for registered users at all
  - TopBar watches if guest was promoted to guest-moderator to fetch participants

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks

- [x] create `fetchParticipants` action from the current code and move to `participantsStore`
- [x] cover an action with tests
- [x] move logic for guests from TopBar to RightSidebar
- [x] handle case when guest moderator was demoted

### 🏁 Checklist

### DIFF
--- a/src/components/RightSidebar/Participants/ParticipantsTab.vue
+++ b/src/components/RightSidebar/Participants/ParticipantsTab.vue
@@ -70,7 +70,6 @@ import getParticipants from '../../../mixins/getParticipants.js'
 import { searchPossibleConversations } from '../../../services/conversationsService.js'
 import { EventBus } from '../../../services/EventBus.js'
 import { addParticipant } from '../../../services/participantsService.js'
-import { useGuestNameStore } from '../../../stores/guestName.js'
 import CancelableRequest from '../../../utils/cancelableRequest.js'
 
 export default {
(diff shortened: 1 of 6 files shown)


DorraJaouad (approved):
Tested

PR #10542, opened 2023-09-20
feat: Allow guests to contacts user via public profile "Talk to" link

### LINKED ISSUE #674: Website integration
### Expected behaviour
1. Any non-Nextcloud user clicks on link on website
2. Call is setup to a specific Nextcloud user
3. Smartphone rings
4. Audio/Video/Text Chat

### Actual behaviour
Non-existing

## Browser
All

jonas2515:
### ☑️ Resolves

* Fix #674

Depends on https://github.com/nextcloud/server/pull/40288

### 🚧 Tasks

- [x] Conversation name should be in recipient language
- [x] Admin config to disable globally/per-user? (will be done in another PR)
- [x] Check that API endpoint respects the users profile setting (right now only the button is hidden) https://github.com/nextcloud/server/pull/41055
- [x] Make sure the actual conversation is only created on a subsequent POST request. Otherwise like preview generation etc. will randomly create new conversations.

### 🏁 Checklist

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -48,6 +48,7 @@ use OCP\IUserManager;
 use OCP\IUserSession;
 use OCP\L10N\IFactory;
 use OCP\Notification\IManager as INotificationManager;
+use OCP\Profile\IProfileManager;
 use OCP\Security\Bruteforce\IThrottler;
 use OCP\Security\RateLimiting\ILimiter;
 use OCP\Security\RateLimiting\IRateLimitExceededException;
@@ -75,6 +76,7 @@ class PageController extends Controller {
 		protected Config $talkConfig,
 		protected IGroupManager $groupManager,
 		protected IUserManager $userManager,
+		protected IProfileManager $profileManager,
 		protected ILimiter $limiter,
 		protected IFactory $l10nFactory,
 	) {
@@ -148,25 +150,26 @@ class PageController extends Controller {
 		return $this->pageHandler($token, $callUser);
 	}
 
-	protected function createPrivateRoom(string $targetUserId): ?Room {
+	/**
+	 * @throws \InvalidArgumentException
+	 */
+	protected function createPrivateRoom(string $targetUserId): Room {
 		$user = $this->userManager->get($targetUserId);
 		if (!$user instanceof IUser) {
-			return null;
+			throw new \InvalidArgumentException('user');
+		}
+
+		if ($this->profileManager->isProfileFieldVisible('talk', $user, null)) {
+			throw new \InvalidArgumentException('profile');
 		}
 
 		$l = $this->l10nFactory->get('spreed', $this->l10nFactory->getUserLanguage($user));
 
-		try {
-			$room = $this->roomService->createConversation(
-				Room::TYPE_PUBLIC,
-				$l->t('Contact request'),
-				$user,
-			);
-		} catch (\InvalidArgumentException) {
-			return null;
-		}
-
-		return $room;
+		return $this->roomService->createConversation(
+			Room::TYPE_PUBLIC,
+			$l->t('Contact request'),
+			$user,
+		);
 	}
 
 	/**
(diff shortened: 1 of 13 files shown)


lib/Controller/PageController.php
```diff
@@ -142,6 +147,27 @@ public function duplicateSession(): Response {
...
+	protected function createPrivateRoom(string $forUser): ?Room {
+		$user = $this->userManager->get($forUser);
+		if (!$user instanceof IUser) {
+			return null;
+		}
+
+		try {
+			$objectType = '';
+			$objectId = '';
+			$room = $this->roomService->createConversation(Room::TYPE_PUBLIC, "Contact request", $user, $objectType, $objectId);
```
nickvergessen, line 163: you still have local changes with the l10nFactory, can you push them?

lib/Controller/PageController.php
```diff
@@ -142,6 +147,27 @@ public function duplicateSession(): Response {
...
+		} catch (InvalidArgumentException $e) {
```
nickvergessen, line 164:
```suggestion
		} catch (\InvalidArgumentException $e) {
```

nickvergessen (commented):
I can afterwards take care of the rebase and adding the admin config as discussed.

lib/Controller/PageController.php
jonas2515: Ah indeed, done!

lib/Controller/PageController.php
```diff
@@ -152,12 +183,37 @@ public function duplicateSession(): Response {
...
+						'create-anonymous-conversation',
+						5, // Five conversations
+						60 * 60, // Per hour
+						$this->request->getRemoteAddress(),
+					);
+				} catch(IRateLimitExceededException $exception) {
+					return new TooManyRequestsResponse();
+				}
+				
+				// FIXME: add rate limiting
```
jonas2515, line 212: Hmm, we added rate-limiting above, so I assume this FIXME can be removed?

Also, looking at this again, shouldn't the rate-limiting block be moved above the call to `createPrivateRoom()`? It looks wrong to error out after creating the new room when the user has been rate-limited...

nickvergessen: In Nextcloud we keep rate limiting/brute forcing in the controller, so that the services can be reused by other mechanisms.

But yeah, the registration should happen before creating the room 😅 (I can take care)

Regarding the comment, yes the fixme can be removed.

lib/Controller/PageController.php
```diff
@@ -152,11 +181,34 @@ public function duplicateSession(): Response {
...
+						'create-anonymous-conversation',
+						5, // Five conversations
+						60 * 60, // Per hour
+						$this->request->getRemoteAddress(),
+					);
+				} catch(IRateLimitExceededException) {
+					return new TooManyRequestsResponse();
+				}
+
+				$room = $this->createPrivateRoom($callUser);
```
nickvergessen, line 202: - [ ] Need to check if the user enabled it before this action @nickvergessen 
- [ ] Need to check if the admin allowed it before this action @nickvergessen

lib/Profile/TalkAction.php
```diff
@@ -80,9 +81,8 @@ public function getIcon(): string {
 	public function getTarget(): ?string {
 		$visitingUser = $this->userSession->getUser();
 		if (
-			!$visitingUser
-			|| $this->config->isDisabledForUser($this->targetUser)
-			|| $this->config->isDisabledForUser($visitingUser)
+			$this->config->isDisabledForUser($this->targetUser)
+			|| ($visitingUser && $this->config->isDisabledForUser($visitingUser))
```
nickvergessen, line 85: - [ ] Need to check if the admin allowed it before this action @nickvergessen

nickvergessen: Implemented now, but requires https://github.com/nextcloud/server/pull/41055

lib/Controller/PageController.php
nickvergessen: > Need to check if the user enabled it before this action

Implemented now, but requires https://github.com/nextcloud/server/pull/41055

nickvergessen (dismissed):

nickvergessen:
Ah, this is still open:

> - [ ] Make sure the actual conversation is only created on a subsequent POST request. Otherwise like preview generation etc. will randomly create new conversations.

nickvergessen:
Need to check if @DorraJaouad has time at the end of the cycle to work on this

nickvergessen (approved):
PHP code-wise it's good for now

jancborchardt (changes requested):
Very cool! :) Some feedback:
- [ ] The whole container should be a bit larger
- [ ] Textarea could use autosize instead of using manual drag to resize (which is also a bit buggy)

nickvergessen:
Also I think as general UX, we might want the lobby to be more obvious to the user recipient of the request, especially while posting/answering, or e.g. when the right sidebar is closed

PR #10543, opened 2023-09-20
feat(Viewer) - allow to list shared media

### LINKED ISSUE #3347: Opening viewer cycle through all share talk medias
cc @danxuliu 

https://github.com/nextcloud/spreed/blob/6ab75d7816a015a2d4d5fb001e45d914db522e2b/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue#L143-L146

- [ ] Create a new endpoint to get a list of filesInfos for all share files in a room
- [ ] Use this list
- [ ] Bonus: use pagination to not load a list of 1k shared files

Antreesy:
### ☑️ Resolves

* Closes #3347
    * as `list` is required, and `loadMore()` is optional
* Update useViewer() mixin according to docs: https://github.com/nextcloud/viewer#open-a-file
* Pass a list of adjacent media files to the Viewer to be able to list through them
  * Could be reused, when support of several files in one message will be added
  * Request is paginated by 20 elements and `loadMore()`

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks

- [ ] Code review
- [ ] Visual check

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
@@ -85,6 +85,7 @@ import Tooltip from '@nextcloud/vue/dist/Directives/Tooltip.js'
 import AudioPlayer from './AudioPlayer.vue'
 
 import { useViewer } from '../../../../../composables/useViewer.js'
+import { SHARED_ITEM } from '../../../../../constants.js'
 
 const PREVIEW_TYPE = {
 	TEMPORARY: 0,
@@ -232,6 +233,11 @@ export default {
 		isSharedItemsTab: {
 			type: Boolean,
 			default: false,
+		},
+
+		sharedItemsType: {
+			type: String,
+			default: '',
 		},
 	},
 
@@ -473,7 +479,18 @@ export default {
 			event.preventDefault()
 
 			const fileInfo = this.generateViewerObject(this)
-			this.openViewer(this.internalAbsolutePath, [fileInfo], fileInfo)
+
+			if (this.isSharedItemsTab && this.sharedItemsType === SHARED_ITEM.TYPES.MEDIA) {
+				// Get available media files from store and put them to the list to navigate through slides
+				const mediaFiles = this.$store.getters.sharedItems(this.$store.getters.getToken())?.media
+				const list = Object.values(mediaFiles).reverse()
+					.map(item => this.generateViewerObject(item.messageParameters.file))
+
+				this.openViewer(this.internalAbsolutePath, list, fileInfo)
+			} else {
+				this.openViewer(this.internalAbsolutePath, [fileInfo], fileInfo)
+
+			}
 		},
 	},
 }
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
@@ -481,12 +481,20 @@ export default {
 			const fileInfo = this.generateViewerObject(this)
 
 			if (this.isSharedItemsTab && this.sharedItemsType === SHARED_ITEM.TYPES.MEDIA) {
-				// Get available media files from store and put them to the list to navigate through slides
-				const mediaFiles = this.$store.getters.sharedItems(this.$store.getters.getToken())?.media
-				const list = Object.values(mediaFiles).reverse()
+				const token = this.$store.getters.getToken()
+				const getRevertedList = (items) => Object.values(items).reverse()
 					.map(item => this.generateViewerObject(item.messageParameters.file))
 
-				this.openViewer(this.internalAbsolutePath, list, fileInfo)
+				// Get available media files from store and put them to the list to navigate through slides
+				const mediaFiles = this.$store.getters.sharedItems(token).media
+				const list = getRevertedList(mediaFiles)
+				const loadMore = async () => {
+					const { messages } = await this.$store.dispatch('getSharedItems',
+						{ token, type: SHARED_ITEM.TYPES.MEDIA })
+					return getRevertedList(messages)
+				}
+
+				this.openViewer(this.internalAbsolutePath, list, fileInfo, loadMore)
 			} else {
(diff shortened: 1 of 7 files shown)


skjnldsv (approved):

src/composables/useViewer.js
```diff
@@ -80,10 +118,20 @@ export function useViewer() {
...
 
+	const generateViewerObject = (file) => ({
+		fileid: parseInt(file.id, 10),
+		filename: generateAbsolutePath(file.path),
+		basename: file.name,
+		mime: file.mimetype,
+		hasPreview: file.previewAvailable === 'yes' || file['preview-available'] === 'yes',
+		etag: file.etag,
+		permissions: generatePermissions(file.permissions),
+	})
```
DorraJaouad, line 129: The fields should match the definition of fileInfo in Viewer
https://github.com/nextcloud/viewer/blob/af7635343546808dab410c76d57104723a7ca6cf/src/services/Viewer.js#L40
`etag` and `permissions` are extra ?

src/components/RightSidebar/SharedItems/SharedItems.vue
```diff
@@ -61,6 +61,7 @@
 				:key="item.id"
 				:small-preview="isList"
 				:row-layout="isList"
+				:shared-items-type="type"
```
DorraJaouad, line 64: We can do a follow-up to remove `isVoiceMessage` prop :)

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
```diff
@@ -472,44 +478,33 @@ export default {
...
-				}
-				if (this.permissions & OC.PERMISSION_DELETE) {
-					permissions += 'D'
-				}
-				if (this.permissions & OC.PERMISSION_SHARE) {
-					permissions += 'R'
+			const fileInfo = this.generateViewerObject(this)
+
+			if (this.isSharedItemsTab && this.sharedItemsType === SHARED_ITEM.TYPES.MEDIA) {
+				const token = this.$store.getters.getToken()
```
DorraJaouad, line 484: We can add `token` prop to `sharedItems` and pass it to `FilePreview` as we are calling it from the store in both components. All other components that are using `FilePreview` have the token prop already.

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
```diff
@@ -472,44 +478,33 @@ export default {
...
-					permissions += 'W'
-				}
-				if (this.permissions & OC.PERMISSION_DELETE) {
-					permissions += 'D'
-				}
-				if (this.permissions & OC.PERMISSION_SHARE) {
-					permissions += 'R'
+			const fileInfo = this.generateViewerObject(this)
+
+			if (this.isSharedItemsTab && this.sharedItemsType === SHARED_ITEM.TYPES.MEDIA) {
```
DorraJaouad, line 483: Maybe it's also good to have it for the chat messages as well?

DorraJaouad: BTW, `isSharedItemsTab `should be renamed because it is also true in `SharedItemsBrowser` . Maybe isOverviewGrid or sthg similar

DorraJaouad (changes requested):
Tested 🦅and it works. There is room for improvement.

src/composables/useViewer.js
Antreesy: Maybe, but without `etag` file isn't opening in Viewer, and permissions could be required for CRUD operations on the file.
Besides, the more information we provide to the Viewer, the lesser is chance for it to not fetch an additional data, or provide an image from cache. So I'd tend to leave the prop fields as-is

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
Antreesy: It may be an overdo + tricky for some components, as they don't have a list, and props are provided not via template:
```js
richParameters[p] = {
	component: FilePreview,
	props: parameters,
}
``` 

But we could do it in a follow-up as well, as other components don't contain multiple media items, and may need this logic it only for NC28

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
Antreesy: > Maybe it's also good to have it for the chat messages as well?

I personally don't think, that it should be implemented across all messages, only within a single one. But that's for NC28 also, as we don't have multiple items in one message.

Antreesy: > `isSharedItemsTab `should be renamed

It's a part of different PR, i could fix it there, to not raise conflicts

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
DorraJaouad: Still simple, we can just add token to `parameters`

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
DorraJaouad: > > Maybe it's also good to have it for the chat messages as well?
> 
> I personally don't think, that it should be implemented across all messages, only within a single one. But that's for NC28 also, as we don't have multiple items in one message.

I actually think that it should be implemented across all messages ( for easier media navigation when checking the chat). It is common in other apps/websites.

DorraJaouad (approved):
Good to go for now.

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
Antreesy: > it should be implemented across all messages 

I'd discuss it in public or design channel maybe

PR #10544, opened 2023-09-20
fix(notifications): Allow sessions to mark themselves as inactive

### LINKED ISSUE #1575: Notifications for chats are not showing when browser tab/window is inactive
## Steps to reproduce
1. Join a chat and stay in room but change browser tab
2. Make someone in the room send you a message
3. No notification raised

### Expected behaviour
As the browser tab or window is not active, user should be marked as inactive and the receive notification.

### Actual behaviour
No notification is raised because of user not being marked "inactive" on window change.
So you are to miss all messages from other until you go back to Nextcloud talk tab/window

## I removed many of items asked in this bug in order to keep it clear. Those were not revelant.
If you miss any information, please feel free to ask.

## Browser ALL

**Operating system:** Ubuntu
**Browser name:** Firefox/Chrome

**Browser version:** 65 for Firefox

## Spreed app

**Spreed app version:** 5.0.2

**Custom TURN server configured:** yes

**Operating system**: Ubuntu

**Web server:** Apache

**Database:** Maria

**PHP version:**

**Nextcloud Version:** 15.0.4

nickvergessen:
### ☑️ Resolves

* Fix #1575
* Current session (opened room) will be marked as inactive if:
  * Tab was hidden or minimized
  * Tab was blurred, and there were not any movements on it for 3 minutes
* If inactive, it will receive notifications for this tab as well

### 🏁 Checklist

### DIFF
--- a/lib/Migration/Version18000Date20230920182747.php
+++ b/lib/Migration/Version18000Date20230920182747.php
@@ -1,0 +1,55 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2023, Joas Schilling <coding@schilljs.com>
+ *
+ * @author Joas Schilling <coding@schilljs.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Talk\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version18000Date20230920182747 extends SimpleMigrationStep {
+	/**
+	 * @param IOutput $output
+	 * @param Closure(): ISchemaWrapper $schemaClosure
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('talk_sessions');
+		if (!$table->hasColumn('state')) {
+			$table->addColumn('state', Types::SMALLINT, [
+				'default' => 1, // active
+				'unsigned' => true,
+			]);
+			return $schema;
+		}
+		return null;
+	}
+}
--- a/src/composables/useActiveSession.js
+++ b/src/composables/useActiveSession.js
@@ -1,0 +1,134 @@
+/*
+ * @copyright Copyright (c) 2023 Maksim Sukharev <antreesy.web@gmail.com>
+ *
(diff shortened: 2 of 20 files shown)


lib/Migration/Version18000Date20230920182747.php
```diff
@@ -0,0 +1,56 @@
...
+	 * @param IOutput $output
+	 * @param Closure(): ISchemaWrapper $schemaClosure
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$modified = false;
```
SystemKeeper, line 45: Not needed?

nickvergessen: fixed

src/composables/useActiveSession.js
```diff
@@ -0,0 +1,132 @@
...
+import { SESSION } from '../constants.js'
+import { setSessionState } from '../services/participantsService.js'
+import { useStore } from './useStore.js'
+
+const supportSessionState = getCapabilities()?.spreed?.features?.includes('session-state')
+const INACTIVE_TIME_MS = 3 * 60 * 1000
+/**
+ * Check whether the current session is active or not:
+ * - tab or browser window was moved to background or minimized
+ * - there was not movements within tab window for a long time
```
nickvergessen, line 34: and not in a call?

src/composables/useActiveSession.js
```diff
@@ -0,0 +1,132 @@
...
+		if (currentState.value === SESSION.STATE.INACTIVE) {
+			return
+		}
+		clearTimeout(inactiveTimer.value)
+		inactiveTimer.value = null
+		currentState.value = SESSION.STATE.INACTIVE
+
+		try {
+			await setSessionState(token.value, SESSION.STATE.INACTIVE)
+			showInfo(t('spreed', 'Session has been marked as inactive'))
```
nickvergessen, line 100: console.info

Not sure about `showInfo` might be helpful if you see it on another screen though.

src/composables/useActiveSession.js
```diff
@@ -0,0 +1,132 @@
...
+	const currentState = ref(SESSION.STATE.ACTIVE)
+
+	watch(token, () => {
+		// Joined conversation has active state by default
+		currentState.value = SESSION.STATE.ACTIVE
+	})
+
+	watch(windowIsVisible, (value) => {
+		// Change state if tab is hidden or minimized
+		// FIXME upstream: Notifications app doesn't work in background
```
nickvergessen, line 57: Not sure what you mean, the notifications app should work fine.
It even has code to track visibility because we change the page title to prefix it with `*` when you receive something, so pinned tabs are highlighted.

Antreesy: I don't see native browser notifications, when tab is minimized, only when tab is opened again, new notification is added to the list and trigger browser notification to show

src/composables/useActiveSession.js
Antreesy: Looks like this. It will provide at least some info, why you start to receive notifications from the chat, which is currently opened. But for sure, we could remove this and leave only `console.info`

image

nickvergessen: Let's do both

src/composables/useActiveSession.js
```diff
@@ -0,0 +1,135 @@
...
+	const currentState = ref(SESSION.STATE.ACTIVE)
+
+	watch(token, () => {
+		// Joined conversation has active state by default
+		currentState.value = SESSION.STATE.ACTIVE
+	})
+
+	watch(windowIsVisible, (value) => {
+		// Change state if tab is hidden or minimized
+		// FIXME: Notifications app doesn't show browser notifications if in background
```
nickvergessen, line 58:
```suggestion
```

I think this does not belong here. Also I just tested it again and it works as expected on my side.

Antreesy: Rebased, squashed fixup and removed this comment

nickvergessen (commented):
Frontend code seems to work 👍🏼

Antreesy (approved):
Backend also works fine

Issue #10559, opened 2023-09-21
Watchers are not optimally written for RightSidebar


Antreesy:
Watchers are not optimally written for RightSidebar

_Originally posted by @Antreesy in https://github.com/nextcloud/spreed/pull/10557#discussion_r1333324770_

PR #10588, opened 2023-09-25
fix(RightSidebar) - Optimize update active tab ( sharedItems) 

DorraJaouad:
Minimized fetch requests of shared Items
Regression from #10558
Part of #10559

### 🚧 Tasks

- [ ] code review

### DIFF
--- a/src/components/RightSidebar/SharedItems/SharedItemsTab.vue
+++ b/src/components/RightSidebar/SharedItems/SharedItemsTab.vue
@@ -20,7 +20,8 @@
 -->
 
 <template>
-	<div v-if="!loading && active">
+	<LoadingComponent v-if="loading" class="tab-loading" />
+	<div v-else>
 		<!-- Shared items grouped by type -->
 		<template v-for="type in sharedItemsOrder">
 			<div v-if="sharedItems[type]" :key="type">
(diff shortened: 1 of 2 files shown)


src/components/RightSidebar/SharedItems/SharedItemsTab.vue
```diff
@@ -164,13 +165,17 @@ export default {
 		hasSharedItems() {
 			return Object.keys(this.$store.getters.sharedItems(this.token)).length > 0
 		},
+
+		isOneToOne() {
```
Antreesy, line 169: Now, if we switch to 1-1 conversation, overview will be loaded regardless the tab is open or not.
Maybe watch not after `isOneToOne`, but after `this.$store.getters.getSidebarStatus` to check three items:
 - sidebar is open
 - tab is active
 - token has changed

when it could be the same logic as we have in chatIdentifier in MessageList:
```js
		chatIdentifier() {
			return this.token + ':' + this.active + ':' + this.isSidebarOpen
		},

		watch:
			chatIdentifier() {
				if (this.token && this.active && this.isSidebarOpen) {
					this.getSharedItemsOverview()
				}
			}
		},
```

src/components/RightSidebar/SharedItems/SharedItemsTab.vue
```diff
@@ -164,13 +165,17 @@ export default {
...
+			return this.conversation.type === CONVERSATION.TYPE.ONE_TO_ONE
+				|| this.conversation.type === CONVERSATION.TYPE.ONE_TO_ONE_FORMER
+		},
 	},
 
 	watch: {
 		active: {
```
Antreesy, line 176: Then we need only one watcher instead of two we have here atm

Antreesy (commented):
Nice addition!

Let's also add a spinner to this component:
```vue
<template>
	<LoadingComponent v-if="loading" />
	<div v-else>
	...
	</div>
</template>
```

Antreesy (approved):
:seal:

PR #10589, opened 2023-09-25
Add key prop to NclistItem to re-render conversation with the appropr…

DorraJaouad:
### ☑️ Resolves

Sometimes ( not sure what the trigger is yet), the active conversation class gets passed to all conversation items in the virtual scroller one by one successively.

Adding a key prop to NcListItem will force proper re-render.

### 🚧 Tasks

- [ ] code review

### DIFF
--- a/src/components/LeftSidebar/ConversationsList/Conversation.vue
+++ b/src/components/LeftSidebar/ConversationsList/Conversation.vue
@@ -21,6 +21,7 @@
 
 <template>
 	<NcListItem ref="listItem"
+		:key="item.token"
 		:name="item.displayName"
 		class="conversation-item"
 		:class="{'unread-mention-conversation': item.unreadMention}"


ShGKme:
> Was lucky to capture the screencast:

Was it uploaded correctly? I see only this static image in the video...
image

Anyway, is there a way to reproduce the bug? So we can debug it to find what causes the problem. Or at least a screenshot with a bug =D

> Not sure, that a force-rerendering with `:key` will help with performance. Furthermore, maybe it's hidden conversations that triggering it, cc @ShGKme for some thoughts

As `NcListItem` doesn't have any local state (except NcAvatar which already has a key, unfortunately), it won't break anything. But it makes scrolling a less performance. Component instances are not reused anymore but are fully re-created (including DOM elements) during the scrolling.

DorraJaouad:
https://github.com/nextcloud/spreed/assets/84044328/f9cb9721-348b-4664-b871-1e6830bb556b

ShGKme:
Yep, I can see the problem now. Though, still cannot reproduce it...

ShGKme:
> Furthermore, maybe it's hidden conversations that triggering it

Having duplicated IDs is also weird. but it should not trigger an active state change. Vue Router adds an active state to all active links even if there are many.

DorraJaouad:
Tested when switching to filters as the glitching is also visible there. 

| With key  |  Without key |
|------------|----------|
| With-Key | Without-key |

ShGKme (approved):
Still cannot reproduce to debug, let's go with the key for now

ShGKme:
> Also fixes #10698 ?
> 
> If so we should backport this to stable27?

Yes. 

The problem with titles caused by direct DOM mutation here:
https://github.com/nextcloud/spreed/blob/6de7100adcdc88d72d88d129a8eb3da41a4b0e19/src/components/LeftSidebar/ConversationsList/Conversation.vue#L339-L341

Vue is not aware of this change and doesn't update the title because it is not bind to the application state.

`key` fix this indirectly by destroying and creating a DOM node from scratch.

PR #10591, opened 2023-09-26
[stable26] Optimize update active tab ( sharedItems)

DorraJaouad:
Backport of #10588
and added the loading spinner component as in https://github.com/nextcloud/spreed/pull/10323/commits/a0a6dfb007cbdea45d8e1d1c596cc2f8b80ab340


Antreesy:
Wrap a watcher into the `$nextTick` to wait for participants tab to mount. Otherwise, when switch to a group conversation, active tab is not switched, but the `activeTab` parameter does. Because of that SharedItemsTab stucks with loading.

It was changes with `@nextcloud/vue` in version v7.11.0, but stable26 has previous one:
https://github.com/nextcloud/spreed/pull/9371/commits/d5fcc1324eb61292a22c250b9ce6b6031d9ca2e7

Antreesy (approved):
Tested

PR #10593, opened 2023-09-26
feat(user_status): Increase participant limit to 1k users

nickvergessen:
Currently limiting to 1k users because the user_status API would yield an error on Oracle otherwise. Clients should use a virtual scrolling mechanism so the data should not be a problem nowadays

### 🏁 Checklist


Antreesy (approved):

jancborchardt (approved):
Thanks a bunch! :)

Issue #10405, opened 2023-08-29
Frontend "Note to self"


nickvergessen:
- Overview at https://github.com/nextcloud/spreed/issues/2196
- First attempt at https://github.com/nextcloud/spreed/pull/3320

---

- [ ] Show conversation
- [ ] "Note to self" as quick action to "Forward a message to your own note-to-self conversation"?

nickvergessen:
The following restrictions apply to the newly added conversation type, so make sure that the UI options are hidden:

- [ ] Can not add participants to the room
- [ ] Can not allow guests
- [ ] Can not make read-only
- [ ] Can not make listable
- [ ] Can not change permissions
- [ ] Can not set lobby
- [ ] Can not enable SIP
- [ ] Can not configure breakout rooms
- [ ] Can not call

PR #10603, opened 2023-09-27
feat(conversation) - add Note to self conversation

DorraJaouad:
### ☑️ Resolves

* Feature #10405

### 🖼️ Screenshots

Forward a message to Note to self
https://github.com/nextcloud/spreed/assets/84044328/6335d4b9-7221-4842-8f86-0770ae325646

Restore
https://github.com/nextcloud/spreed/assets/84044328/1e3f593f-5be0-4604-9a6a-df601f829c1f

### 🚧 Tasks

- [x] Remove the non-applicable settings
- [x] Add the option to restore the note to self conversation if it was removed
- [x] Add the option to forward a message to note to self 
- [x] Add forward message action to the store and some follow-up refactoring where it is used
- [x] Write the relevant tests for the new action

### 🏁 Checklist

### DIFF
--- a/docs/constants.md
+++ b/docs/constants.md
@@ -8,6 +8,7 @@
 * `3` Public
 * `4` Changelog
 * `5` Former "One to one" (When a user is deleted from the server or removed from all their conversations, `1` "One to one" rooms are converted to this type)
+* `6` Note to self 
 
 ### Object types
 
--- a/src/components/ConversationSettings/ConversationSettingsDialog.vue
+++ b/src/components/ConversationSettings/ConversationSettingsDialog.vue
@@ -36,7 +36,8 @@
 
 		<template v-if="!isBreakoutRoom">
 			<!-- Notifications settings and devices preview screen -->
-			<NcAppSettingsSection id="notifications"
+			<NcAppSettingsSection v-if="!isNoteToSelf"
+				id="notifications"
 				:title="t('spreed', 'Personal')">
 				<NcCheckboxRadioSwitch :checked.sync="showMediaSettings"
 					type="switch">
@@ -49,8 +50,8 @@
 			<NcAppSettingsSection v-if="canFullModerate"
 				id="conversation-settings"
 				:title="t('spreed', 'Moderation')">
-				<ListableSettings :token="token" />
-				<LinkShareSettings ref="linkShareSettings" />
+				<ListableSettings v-if="!isNoteToSelf" :token="token" />
+				<LinkShareSettings v-if="!isNoteToSelf" ref="linkShareSettings" />
 				<ExpirationSettings :token="token" can-full-moderate />
 			</NcAppSettingsSection>
 			<NcAppSettingsSection v-else
@@ -60,7 +61,7 @@
 			</NcAppSettingsSection>
 
 			<!-- Meeting: lobby and sip -->
-			<NcAppSettingsSection v-if="canFullModerate"
+			<NcAppSettingsSection v-if="canFullModerate && !isNoteToSelf"
 				id="meeting"
 				:title="t('spreed', 'Meeting')">
 				<LobbySettings :token="token" />
@@ -68,7 +69,7 @@
 			</NcAppSettingsSection>
 
 			<!-- Conversation permissions -->
-			<NcAppSettingsSection v-if="canFullModerate"
+			<NcAppSettingsSection v-if="canFullModerate && !isNoteToSelf"
 				id="permissions"
 				:title="t('spreed', 'Permissions')">
 				<ConversationPermissionsSettings :token="token" />
@@ -99,7 +100,7 @@
 			<NcAppSettingsSection v-if="canLeaveConversation || canDeleteConversation"
 				id="dangerzone"
 				:title="t('spreed', 'Danger zone')">
-				<LockingSettings :token="token" />
+				<LockingSettings v-if="canFullModerate && !isNoteToSelf" :token="token" />
 				<DangerZone :conversation="conversation"
 					:can-leave-conversation="canLeaveConversation"
 					:can-delete-conversation="canDeleteConversation" />
@@ -171,6 +172,10 @@ export default {
 
 		canUserEnableSIP() {
 			return this.conversation.canEnableSIP
+		},
+
+		isNoteToSelf() {
+			return this.conversation.type === CONVERSATION.TYPE.NOTE_TO_SELF
 		},
 
 		token() {
--- a/src/components/LeftSidebar/ConversationsList/Conversation.vue
+++ b/src/components/LeftSidebar/ConversationsList/Conversation.vue
@@ -80,7 +80,7 @@
 				</template>
 				{{ t('spreed', 'Conversation settings') }}
 			</NcActionButton>
-			<NcActionButton v-if="canLeaveConversation"
+			<NcActionButton v-if="canLeaveConversation && !isNoteToSelf"
 				:close-after-click="true"
 				@click="leaveConversation">
 				<template #icon>
--- a/src/components/LeftSidebar/ConversationsList/Conversation.vue
+++ b/src/components/LeftSidebar/ConversationsList/Conversation.vue
@@ -80,7 +80,7 @@
 				</template>
 				{{ t('spreed', 'Conversation settings') }}
 			</NcActionButton>
-			<NcActionButton v-if="canLeaveConversation && !isNoteToSelf"
+			<NcActionButton v-if="canLeaveConversation"
 				:close-after-click="true"
 				@click="leaveConversation">
 				<template #icon>
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -96,7 +96,7 @@
 						<template #icon>
 							<Note :size="20" />
 						</template>
-						{{ t('spreed','New Personal Note') }}
+						{{ t('spreed','New personal note') }}
 					</NcActionButton>
 
 					<NcActionButton close-after-click
@@ -409,7 +409,7 @@ export default {
 		},
 
 		hasNoteToSelf() {
-			return this.conversationsList.filter(conversation => conversation.type === CONVERSATION.TYPE.NOTE_TO_SELF).length
+			return this.conversationsList.find(conversation => conversation.type === CONVERSATION.TYPE.NOTE_TO_SELF)
 		},
 
 		sourcesWithoutResults() {
@@ -632,7 +632,7 @@ export default {
 			}
 		},
 
-		addAndOpenConversation(conversation) {
+		switchToConversation(conversation) {
 			this.$store.dispatch('addConversation', conversation)
 			this.abortSearch()
 			this.$router.push({
@@ -644,13 +644,13 @@ export default {
 		async createConversation(name) {
 			const response = await createPrivateConversation(name)
 			const conversation = response.data.ocs.data
-			this.addAndOpenConversation(conversation)
+			this.switchToConversation(conversation)
 		},
 
 		async restoreNoteToSelfConversation() {
 			const response = await fetchNoteToSelfConversation()
 			const conversation = response.data.ocs.data
-			this.addAndOpenConversation(conversation)
+			this.switchToConversation(conversation)
 		},
 
 		hasOneToOneConversationWith(userId) {
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -90,6 +90,15 @@
 						{{ t('spreed','Create a new conversation') }}
 					</NcActionButton>
 
+					<NcActionButton v-if="!hasNoteToSelf"
+						close-after-click
+						@click="restoreNoteToSelfConversation">
+						<template #icon>
+							<Note :size="20" />
+						</template>
+						{{ t('spreed','New Personal Note') }}
+					</NcActionButton>
+
 					<NcActionButton close-after-click
 						@click="showModalListConversations">
 						<template #icon>
@@ -239,6 +248,7 @@ import FilterIcon from 'vue-material-design-icons/Filter.vue'
 import FilterRemoveIcon from 'vue-material-design-icons/FilterRemove.vue'
 import List from 'vue-material-design-icons/FormatListBulleted.vue'
 import MessageBadge from 'vue-material-design-icons/MessageBadge.vue'
+import Note from 'vue-material-design-icons/NoteEditOutline.vue'
 import Plus from 'vue-material-design-icons/Plus.vue'
 
 import { showError } from '@nextcloud/dialogs'
@@ -267,6 +277,7 @@ import { CONVERSATION } from '../../constants.js'
 import BrowserStorage from '../../services/BrowserStorage.js'
 import {
 	createPrivateConversation,
+	fetchNoteToSelfConversation,
 	searchPossibleConversations,
 	searchListedConversations,
 } from '../../services/conversationsService.js'
@@ -303,6 +314,7 @@ export default {
 		ChatPlus,
 		List,
 		DotsVertical,
+		Note,
 	},
 
 	mixins: [
@@ -394,6 +406,10 @@ export default {
 
 		isSearching() {
 			return this.searchText !== ''
+		},
+
+		hasNoteToSelf() {
+			return this.conversationsList.filter(conversation => conversation.type === CONVERSATION.TYPE.NOTE_TO_SELF).length
 		},
 
 		sourcesWithoutResults() {
@@ -616,15 +632,25 @@ export default {
 			}
 		},
 
-		async createConversation(name) {
-			const response = await createPrivateConversation(name)
-			const conversation = response.data.ocs.data
+		addAndOpenConversation(conversation) {
 			this.$store.dispatch('addConversation', conversation)
 			this.abortSearch()
 			this.$router.push({
 				name: 'conversation',
 				params: { token: conversation.token },
 			}).catch(err => console.debug(`Error while pushing the new conversation's route: ${err}`))
+		},
+
+		async createConversation(name) {
+			const response = await createPrivateConversation(name)
+			const conversation = response.data.ocs.data
+			this.addAndOpenConversation(conversation)
+		},
+
+		async restoreNoteToSelfConversation() {
+			const response = await fetchNoteToSelfConversation()
+			const conversation = response.data.ocs.data
+			this.addAndOpenConversation(conversation)
 		},
 
 		hasOneToOneConversationWith(userId) {
--- a/src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/Forwarder.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/Forwarder.vue
@@ -68,7 +68,6 @@
 </template>
 
 <script>
-import cloneDeep from 'lodash/cloneDeep.js'
 
 import Check from 'vue-material-design-icons/Check.vue'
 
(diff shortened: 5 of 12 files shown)


docs/constants.md
```diff
@@ -8,6 +8,7 @@
 * `3` Public
 * `4` Changelog
 * `5` Former "One to one" (When a user is deleted from the server or removed from all their conversations, `1` "One to one" rooms are converted to this type)
+* `6` Note to self
```
nickvergessen, line 11: Ops, thanks!

src/components/LeftSidebar/ConversationsList/Conversation.vue
```diff
@@ -83,7 +83,7 @@
 				</template>
 				{{ t('spreed', 'Conversation settings') }}
 			</NcActionButton>
-			<NcActionButton v-if="canLeaveConversation"
+			<NcActionButton v-if="canLeaveConversation && !isNoteToSelf"
```
nickvergessen, line 86: This should not be needed. I will adjust the backend

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -90,6 +90,15 @@
 						{{ t('spreed','Create a new conversation') }}
 					</NcActionButton>
 
+					<NcActionButton v-if="!hasNoteToSelf"
+						close-after-click
+						@click="restoreNoteToSelfConversation">
```
nickvergessen, line 95: We agreed to not do that, instead the "Note to self" Message action would recreate it.
Now that the code is already here, I wouldn't mind.

Cc @jancborchardt

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
```diff
@@ -111,6 +111,14 @@
 						</template>
 						{{ t('spreed', 'Go to file') }}
 					</NcActionLink>
+					<NcActionButton v-if="canForwardMessage && !isInNoteToSelf"
+						close-after-click
+						@click="forwardToNote">
+						<template #icon>
+							<Note :size="16" />
+						</template>
+						{{ t('spreed', 'Add to Note') }}
```
nickvergessen, line 120:
```suggestion
						{{ t('spreed', 'Note to self') }}
```
?

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
```diff
@@ -702,6 +717,54 @@ export default {
 			this.$emit('update:isReactionsMenuOpen', true)
 		},
 
+		async forwardToNote() {
+			const NoteToSelf = this.$store.getters.conversationsList.filter(conversation => conversation.type === CONVERSATION.TYPE.NOTE_TO_SELF)[0]
+			const messageToBeForwarded = cloneDeep(this.messageObject)
```
nickvergessen, line 722: Is this the same code as in ForwardMessage.vue?

src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
```diff
@@ -594,7 +594,8 @@ export default {
 
 		showModeratorLabel() {
 			return this.isModerator
-				&& ![CONVERSATION.TYPE.ONE_TO_ONE, CONVERSATION.TYPE.ONE_TO_ONE_FORMER, CONVERSATION.TYPE.CHANGELOG].includes(this.conversation.type)
+				&& ![CONVERSATION.TYPE.ONE_TO_ONE, CONVERSATION.TYPE.ONE_TO_ONE_FORMER,
+					CONVERSATION.TYPE.CHANGELOG, CONVERSATION.TYPE.NOTE_TO_SELF].includes(this.conversation.type)
```
nickvergessen, line 598: This whole participants tab should not be there

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
```diff
@@ -702,6 +717,54 @@ export default {
 			this.$emit('update:isReactionsMenuOpen', true)
 		},
 
+		async forwardToNote() {
+			const NoteToSelf = this.$store.getters.conversationsList.filter(conversation => conversation.type === CONVERSATION.TYPE.NOTE_TO_SELF)[0]
```
nickvergessen, line 721: Might by empty, so `[0]` will fail. In that case the function from the service file should be called

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
```diff
@@ -702,6 +717,54 @@ export default {
...
+							objectId: richObject.id,
+							objectType: richObject.type,
+							metaData: JSON.stringify(richObject),
+							referenceId: '',
+						},
+					})
+					showSuccess(t('spreed', 'Message added to Note to self'))
+				} catch (error) {
+					console.error('Error while adding message to Note', error)
+					showError(t('spreed', 'Error while adding message to Note'))
```
nickvergessen, line 745:
```suggestion
					showError(t('spreed', 'Error while forwarding message to "Note to self"'))
```

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
```diff
@@ -702,6 +717,54 @@ export default {
...
+					await this.$store.dispatch('forwardRichObject', {
+						token: this.token,
+						richObject: {
+							objectId: richObject.id,
+							objectType: richObject.type,
+							metaData: JSON.stringify(richObject),
+							referenceId: '',
+						},
+					})
+					showSuccess(t('spreed', 'Message added to Note to self'))
```
nickvergessen, line 742:
```suggestion
					showSuccess(t('spreed', 'Message forwarded to "Note to self"'))
```

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
DorraJaouad: cc @jancborchardt

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
DorraJaouad: Yes with toast messages addition. For now, the forwarding to note to self is omitting messages with type `{file}` like ForwardMessage.vue . If we want to also be able to forward pictures , files ... either : 
- Post the file link in self to note.
- Process and post it as in NewMessage but change the token to note to self's.

Otherwise, we can move the function to the messageStore to reduce redundancy.

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -616,9 +632,7 @@ export default {
 			}
 		},
 
-		async createConversation(name) {
-			const response = await createPrivateConversation(name)
-			const conversation = response.data.ocs.data
+		addAndOpenConversation(conversation) {
```
Antreesy, line 635: Conversation was already added on backend, so it's just a shortcut for putting it to LeftSidebar and open. Maybe, rename like this?
```suggestion
		openConversation(conversation) {
```
or
```suggestion
		switchToConversation(conversation) {
```

Apart from that, nice reuse

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
Antreesy: For files sharing, it will be touched anyway by https://github.com/nextcloud/spreed/pull/8949, so we can leave it for now
For method, let's reuse it with messagesStore. Toasts could be posted from component, just throw error from store action and add try-catch in Forwarder and MessageButtonsBar

src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
```diff
@@ -594,7 +594,8 @@ export default {
 
 		showModeratorLabel() {
 			return this.isModerator
-				&& ![CONVERSATION.TYPE.ONE_TO_ONE, CONVERSATION.TYPE.ONE_TO_ONE_FORMER, CONVERSATION.TYPE.CHANGELOG].includes(this.conversation.type)
+				&& ![CONVERSATION.TYPE.ONE_TO_ONE, CONVERSATION.TYPE.ONE_TO_ONE_FORMER,
```
Antreesy, line 597: Let's revert it, as there's nothing more changed

src/components/TopBar/CallButton.vue
```diff
@@ -254,6 +254,7 @@ export default {
 
 		showStartCallButton() {
 			return this.callEnabled
+				&& this.conversation.type !== CONVERSATION.TYPE.NOTE_TO_SELF
```
Antreesy, line 257: `canStartCall: true`
Evereything is good here, in normal cases, if false, button is disabled, for this conv.type it should be hidden

Maybe also should be fixed on backend? cc @nickvergessen

src/components/LeftSidebar/ConversationsList/Conversation.vue
nickvergessen: Adjusted in https://github.com/nextcloud/spreed/pull/10606

src/components/ConversationSettings/ConversationSettingsDialog.vue
```diff
@@ -46,7 +47,7 @@
 				<NotificationsSettings :conversation="conversation" />
 			</NcAppSettingsSection>
 
-			<NcAppSettingsSection v-if="canFullModerate"
+			<NcAppSettingsSection v-if="canFullModerate && !isNoteToSelf"
```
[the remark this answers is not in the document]
nickvergessen: No, actually we would like to allow setting up expiration and bots in `Note to self`

src/components/TopBar/CallButton.vue
nickvergessen: Fixed `canStartCall` to be false in https://github.com/nextcloud/spreed/pull/10606
But we should not only disable the button, but still hide it, so this here is still desired

src/components/LeftSidebar/ConversationsList/Conversation.vue
```diff
@@ -220,6 +220,10 @@ export default {
 			return this.item.notificationLevel === PARTICIPANT.NOTIFY.NEVER
 		},
 
+		isNoteToSelf() {
```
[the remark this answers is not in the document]
DorraJaouad: oops, It was leftover from hiding "leave conversation". It is removed now because it will be fixed in backend.

DorraJaouad:
> What about labelling it as New personal note without capital letters, and put on a third place in the list?

I ordered it based on the usage : 
* Create a new conversation - the most common action
* New personal note - taking personal notes is a useful but less common action than starting conversations and more relevant to the user on a daily basis ( it will appear only when the note to self does not exist)
*  Join open conversation - viewing existing conversations is helpful but not where most users will start or use it daily.

cc @nextcloud/designers

src/components/ConversationSettings/ConversationSettingsDialog.vue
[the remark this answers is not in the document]
nickvergessen: the `!isNoteToSelf` check should be moved to the mounting of `ListableSettings` and `LinkShareSettings` so the expiration can still be toggled :)

src/components/ConversationSettings/ConversationSettingsDialog.vue
```diff
@@ -99,7 +100,7 @@
 			<NcAppSettingsSection v-if="canLeaveConversation || canDeleteConversation"
 				id="dangerzone"
 				:title="t('spreed', 'Danger zone')">
-				<LockingSettings :token="token" />
+				<LockingSettings v-if="canFullModerate && !isNoteToSelf" :token="token" />
```
Antreesy, line 103: Pls check that this component doesn't have `canFullModerate` check inside

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
```diff
@@ -702,6 +717,29 @@ export default {
 			this.$emit('update:isReactionsMenuOpen', true)
 		},
 
+		async forwardToNote() {
+			let noteToSelf = this.$store.getters.conversationsList.filter(conversation => conversation.type === CONVERSATION.TYPE.NOTE_TO_SELF)?.[0]
+
+			// If note to self doesn't exist, it will be regenerated
+			if (!noteToSelf) {
```
Antreesy, line 724: I'd move this check to store action also
For example, if token is not provided, default option would be noteToSelf

src/store/messagesStore.js
```diff
@@ -1256,30 +1257,49 @@ const actions = {
...
+			await postRichObjectToConversation(
+				targetToken,
+				{
+					objectId: richObject.id,
+					objectType: richObject.type,
+					metaData: JSON.stringify(richObject),
+					referenceId: '',
+				},
+			)
+			return
```
Antreesy, line 1288:
```suggestion
			return response
```

src/store/messagesStore.js
```diff
@@ -1256,30 +1257,49 @@ const actions = {
...
+		// Overwrite with the target conversation token
+		message.token = targetToken
+
+		if (message.parent) {
+			delete message.parent
+		}
+
+		if (message.message === '{object}' && message.messageParameters.object) {
+			const richObject = message.messageParameters.object
+			await postRichObjectToConversation(
```
Antreesy, line 1279:
```suggestion
			const response = await postRichObjectToConversation(
```

src/store/messagesStore.js
```diff
@@ -1256,30 +1257,49 @@ const actions = {
...
+		// If there are mentions in the message to be forwarded, replace them in the message
+		// text.
+		for (const key in message.messageParameters) {
+			if (key.startsWith('mention')) {
+				const mention = message.messageParameters[key]
+				message.message = message.message.replace(`{${key}}`, `@'${mention.name}'`)
+			}
+		}
+
+		const response = await postNewMessage(message, { silent: false })
```
Antreesy, line 1300: Maybe it should be silent, if posting to yourself? And counter ommited also?

DorraJaouad: Do you mean unread counter? If yes, it won't have it because all messages are coming from the current user.

src/components/ConversationSettings/ConversationSettingsDialog.vue
DorraJaouad: Negative

src/components/RightSidebar/RightSidebar.vue
```diff
@@ -294,10 +302,10 @@ export default {
 			}
 		},
 
-		isOneToOne: {
+		showParticipantsTab: {
```
Antreesy, line 305: guests also fall under this, but don't have 'shared-items' tab. Check please

src/store/messagesStore.js
```diff
@@ -1256,30 +1259,60 @@ const actions = {
 	},
 
 	/**
-	 * Posts a simple text message to a room
+	 * Forwards message to a conversation. By default , the message is forwarded to Note to self.
 	 *
 	 * @param {object} context default store context;
 	 * will be forwarded;
 	 * @param {object} data the wrapping object;
+	 * @param {object} data.targetToken the conversation token to where the message will be forwarded;
```
Antreesy, line 1267:
```suggestion
	 * @param {string} [data.targetToken] the conversation token to where the message will be forwarded;
```

src/store/messagesStore.js
```diff
@@ -1256,30 +1259,60 @@ const actions = {
...
-	 * @param {object} data.richObject the rich object;
-	 */
-	async forwardRichObject(context, { token, richObject }) {
-		const response = await postRichObjectToConversation(token, richObject)
+		// If there are mentions in the message to be forwarded, replace them in the message
+		// text.
+		for (const key in message.messageParameters) {
+			if (key.startsWith('mention')) {
+				const mention = message.messageParameters[key]
+				message.message = message.message.replace(`{${key}}`, `@'${mention.name}'`)
```
Antreesy, line 1309:
```suggestion
				message.message = message.message.replace(`{${key}}`, `@"${mention.id}"`)
```
Or even further, check if `mention.id` includes a space character to wrap with double quotes or not

src/store/messagesStore.spec.js
```diff
@@ -1778,4 +1787,182 @@ describe('messagesStore', () => {
...
+				id: 1,
+				token: TOKEN,
+				message: 'simple text message',
+				messageParameters: {},
+			}
+			conversations = [
+				{
+					token: TOKEN,
+					type: 3,
+					dispalyName: 'conversation 1',
```
Antreesy, line 1816:
```suggestion
					displayName: 'conversation 1',
```

src/store/messagesStore.spec.js
```diff
@@ -1778,4 +1787,182 @@ describe('messagesStore', () => {
...
+		let messageToBeForwarded
+		let targetToken
+		let messageExpected
+
+		beforeEach(() => {
+			localVue = createLocalVue()
+			localVue.use(Vuex)
+
+			testStoreConfig = cloneDeep(storeConfig)
+			// eslint-disable-next-line import/no-named-as-default-member
```
Antreesy, line 1803: Fine for me without comment.
Same as for other tests here

src/store/messagesStore.spec.js
```diff
@@ -1778,4 +1787,182 @@ describe('messagesStore', () => {
...
+				token: TOKEN,
+				message: 'Hello {mention-user1}, {mention-user2}',
+				messageParameters: {
+					'mention-user1': {
+						id: 'taylor',
+						name: 'Taylor',
+						type: 'user',
+					},
+					'mention-user2': {
+						id: 'adam',
```
Antreesy, line 1946: id also can be `'adam driver'` or `'adam@gmail.com'`

src/store/messagesStore.spec.js
```diff
@@ -1778,4 +1787,182 @@ describe('messagesStore', () => {
...
+					},
+				},
+			}
+			targetToken = 'token-2'
+			const mentions = [
+				messageToBeForwarded.messageParameters['mention-user1'],
+				messageToBeForwarded.messageParameters['mention-user2'],
+			]
+			messageExpected = cloneDeep(messageToBeForwarded)
+			messageExpected.message = `Hello @'${mentions[0].name}', @'${mentions[1].name}'`
```
Antreesy, line 1958: Maybe it's better to use a normal string here, like `Hello @taylor, @"adam driver"`, and not reproduce how this store action works
Because test will pass only if it's aligned to action output => it could differ from real behaviour, and mentions won't work at all

src/store/messagesStore.js
DorraJaouad: it should be `mention.name` as it is the display name. The id might differ and not so human-readable ? ( e.g : user1055

src/store/messagesStore.js
```diff
@@ -1256,30 +1259,60 @@ const actions = {
...
+			}
+		}
+
+		const response = await postNewMessage(message, { silent: true })
```
Antreesy, line 1313: Was `false` before.
Maybe it should be true only for 'Note to self'?

src/store/messagesStore.js
nickvergessen: If the Id is used, it would be rendered again.

I think the main reason we used the name is that otherwise if you forward an `@all` into a second room, everyone from the second room will be pinged.

However the quotes you added here break the thing anyway already.

```suggestion
				message.message = message.message.replace(`{${key}}`, `@${mention.name}`)
```

nickvergessen: We can have a look into the handling later, but let's not mix it with this "must merge" feature request.
Keep the previous behaviour for now.

src/store/messagesStore.js
nickvergessen, line 1313: Used to be false, now it's true, meaning if you forward something to a 1-1, the other participant will not be notified.

Unless we add an option for the user to select it, I would stand with above comment also here:
We can have a look into the handling later, but let's not mix it with this "must merge" feature request.
Keep the previous behaviour for now.

```suggestion
		const response = await postNewMessage(message, { silent: false })
```

src/store/messagesStore.js
Antreesy: ATM it will be rendered in different conversation anyway, if `name === id` or `name.split(' ').includes(id)`.
Should be considered in follow-up:

image

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/Forwarder.vue
```diff
@@ -68,7 +68,6 @@
 </template>
```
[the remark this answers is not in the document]
DorraJaouad: Please reference all this in follow-up issue =)

Antreesy (approved):
Seems good as-is, doesn't break the previous behaviour
I've raised a follow-up and RFC issues

PR #10604, opened 2023-09-28
fix(CI): Make sure the "when unrelated" fake summaries work correctly

### LINKED ISSUE #10600: `integration-summary-when-unrelated.yml` is ran when it shouldn't be
Ref
* https://github.com/nextcloud/spreed/pull/10317
* https://github.com/nextcloud/spreed/pull/10598

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths

> When all the path names match patterns in paths-ignore, the workflow will not run. If any path names do not match patterns in paths-ignore, even if some path names match the patterns, the workflow will run.

There was some file that didn't match, so that's why the "when unrelated" ran... that's .... problematic.

"When unrelated" should only run, when none of the matching files was touched.

nickvergessen:
Before they executed when any file outside the defined scope matched. However that means: When a PR touches PHP and JS, both the real and the fake summaries for node, eslint and phpunit got triggered. Since the fake summary is always green it meant one could merge a PR that in the end had red CI on the non-fake summary.

Now the fake summaries are skipped, if any of the files still matches the pattern, instead of being executed whenever a single file did not match it.

Fix #10600

### 🏁 Checklist

### DIFF
--- a/.github/workflows/integration-summary-when-unrelated.yml
+++ b/.github/workflows/integration-summary-when-unrelated.yml
@@ -19,16 +19,79 @@ on:
       - 'composer.json'
       - 'composer.lock'
 
+env:
+  PATHS_IGNORE: '.github/workflows/**;appinfo/**;lib/**;templates/**;tests/integration/**;vendor/**;vendor-bin/**;.php-cs-fixer.dist.php;composer.json;composer.lock'
+
 permissions:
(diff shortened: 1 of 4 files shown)


Antreesy (approved):
Looks fine IMO

PR #10608, opened 2023-09-29
feat(sip-dialout): ☎️ API "SIP dial-out"

### LINKED ISSUE #10410: API "SIP dial-out"
* Overview at https://github.com/nextcloud/spreed/issues/10346

---

- [x] https://github.com/nextcloud/server/pull/40620

nickvergessen:
### ☑️ Resolves

* Fix #10410 

### 🚧 Tasks

- [x] Dependency #10607 
- [x] Docs missing for endpoints the SIP bridge will invoke
- [x] Capability docs missing
- [x] SIP bridge will send a DELETE when the call was denied by the SIP gateway or hangup
- [x] Integration tests for the SIP requests and HPB requests

### 🏁 Checklist


nickvergessen:
- Psalm error extracted to https://github.com/nextcloud/spreed/pull/10632

lib/Config.php
```diff
@@ -139,6 +140,14 @@ public function canUserEnableSIP(IUser $user): bool {
 		return $this->canEnableSIP[$user->getUID()];
 	}
 
+	public function canUserDialOutSIP(IUser $user): bool {
+		if ($this->config->getAppValue('spreed', 'sip_dialout', 'no') === 'no') {
```
nickvergessen, line 144: needs to be documented

appinfo/routes/routesRoomController.php
```diff
@@ -100,6 +100,8 @@
 		['name' => 'Room#getParticipantByDialInPin', 'url' => '/api/{apiVersion}/room/{token}/pin/{pin}', 'verb' => 'GET', 'requirements' => array_merge($requirementsWithToken, [
 			'pin' => '\d{7,32}',
 		])],
+		/** @see \OCA\Talk\Controller\RoomController::getParticipantByDialOutNumber() */
+		['name' => 'Room#getParticipantByDialOutNumber', 'url' => '/api/{apiVersion}/room/{token}/verify-dialout', 'verb' => 'GET', 'requirements' => $requirementsWithToken],
```
fancycode, line 104: Shouldn't this be a `POST`?

lib/Settings/Admin/AdminSettings.php
```diff
@@ -473,6 +473,7 @@ protected function initSIPBridge(): void {
 		$this->initialState->provideInitialState('sip_bridge_groups', $groups);
 		$this->initialState->provideInitialState('sip_bridge_shared_secret', $this->talkConfig->getSIPSharedSecret());
 		$this->initialState->provideInitialState('sip_bridge_dialin_info', $this->talkConfig->getDialInInfo());
+		$this->initialState->provideInitialState('sip_bridge_dialout', $this->talkConfig->isSIPDialOutEnabled()());
```
fancycode, line 476:
```suggestion
		$this->initialState->provideInitialState('sip_bridge_dialout', $this->talkConfig->isSIPDialOutEnabled());
```

lib/Capabilities.php
```diff
@@ -165,6 +165,7 @@ public function getCapabilities(): array {
 				'session-state',
 				'note-to-self',
 				'recording-consent',
+				'sip-support-dialout',
```
nickvergessen, line 168: Needs docs

lib/Controller/RoomController.php
```diff
@@ -1453,6 +1490,55 @@ public function getParticipantByDialInPin(string $pin): DataResponse {
...
+	 *
+	 *  200: Participant created successfully
+	 *  400: Phone number and details could not be confirmed
+	 *  401: SIP request invalid
+	 *  501: SIP dial-out is not configured
+	 */
+	#[PublicPage]
+	#[BruteForceProtection(action: 'talkSipBridgeSecret')]
+	#[RequireRoom]
+	public function getParticipantByDialOutNumber(string $number, array $options = []): DataResponse {
```
nickvergessen, line 1508: Missing from docs

appinfo/routes/routesRoomController.php
nickvergessen: `getParticipantByDialInPin` is also a GET? `createGuestByDialIn` is a POST, but that one also creates something in server (a guest), while `getParticipantByDialOutNumber` just returns something.

fancycode: Well, I was wondering if you want all the parameters in the URL query (and with that in the webserver logs). For privacy reasons it should rather be a `POST` in the  `getParticipantByDialInPin` to not have the PINs in the logs.

But if you want a `GET`, it's fine for me. How should I encode the `options` array for the query parameter? As JSON string?

nickvergessen: > JSON string

Yeah, that would be best for Talk side

nickvergessen: > Well, I was wondering if you want all the parameters in the URL query (and with that in the webserver logs). For privacy reasons it should rather be a POST in the getParticipantByDialInPin to not have the PINs in the logs.

Makes sense. We can introduce a POST in parallel for the existing endpoint and deprecate the GET for a future release, so we are not breaking directly?

fancycode: > Makes sense. We can introduce a POST in parallel for the existing endpoint and deprecate the GET for a future release, so we are not breaking directly?

:+1: but we will need a capability so clients know they can use the `POST` endpoint.

nickvergessen: Capability `sip-support-dialout` added and slightly adjusted the URLs as well
- `POST /room/{token}/verify-dialin` - To replace the existing one for checking a PIN
- `POST /room/{token}/verify-dialout` - To replace this new one for checking a dialout number

nickvergessen:
Failing PHPUnit is due to a regression in Valinor. Let's see if the PR at https://github.com/CuyZ/Valinor/pull/437 is accepted, otherwise we need to adjust/remove the docs on the classes :(

docs/participant.md
```diff
@@ -279,6 +280,73 @@ Note: This is only allowed with validate SIP bridge requests
...
+Note: This is only allowed as validate SIP bridge requests
+
+* Required capability: `sip-support-dialout`
+* Method: `DELETE`
+* Endpoint: `/room/{token}/rejected-dialout`
+* Data:
+
+| field        | type   | Description                                 |
+|--------------|--------|---------------------------------------------|
+| `attendeeId` | int    | The attendee ID of the dial-out participant |
```
fancycode, line 339: In order to not having to parse the `options` to extract the `attendeeId`, I would rather pass the `options` as-is just like it is done for `verify-dialout`.

```suggestion
| `options`    | string | The options as received in the dialout request. |
```

nickvergessen: Makes sense I guess

nickvergessen:
- [ ] add system message for adding participant
- [ ] think of a way to set object type + default room avatar for creates rooms

Antreesy (approved):
Seems working, nothing blocking from my side

PR #10610, opened 2023-09-29
fix(NcAvatar) increase avatar sizes in call

### LINKED ISSUE #10282: follow-up: Reduce amount of component to wrap avatars
We have several options to wrap user's/conversation's/other avatars:
  - ConversationIcon
  - AuthorAvatar
  - AvatarWrapper
  - NcAvatar itself

We need to review first three of the list thoroughly, to check which ones could be replaced or simplified, and wrap the rest of NcAvatar as well, if needed

_Originally posted by @Antreesy in https://github.com/nextcloud/spreed/pull/10262#discussion_r1299367439_

Antreesy:
### ☑️ Resolves

* Fix #10282
  * native NcAvatar is no longer used in components
  * SCSS mixin `avatar-mixin` is no longer used in components
  * Size for all NcAvatar wrappers is now set explicitly - no more reactivity problems
* Size of avatars is increased:
  * in MediaSettings: `128 => 180`
  * in GridView and promoted VideoVue: `128 => 512`
* Common lint fixes
  * some SCSS variables were replaced with CSS variables provided by server 

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image
image | image
image | image

### 🚧 Tasks

- [ ] Visual check
- [ ] Code review

### 🏁 Checklist


src/components/AvatarWrapper/AvatarWrapper.vue
```diff
@@ -184,26 +182,38 @@ export default {
...
+		}
 
-	&--medium {
-		height: 32px;
-		width: 32px;
-		border-radius: 32px;
-		@include avatar-mixin(32px);
+		&.guest {
+			color: #ffffff;
+			background-color: #b9b9b9;
```
DorraJaouad, line 203: I would suggest removing these static colors. 
This is the view from Guest ( display name : pop ) 
image
And this is from a logged in user and Guest is present in the call ( display name : pop) 
image

It would be nice if we use the same style for guest avatar from the user account view in the guest view ( make it always uppercase too)

src/components/CallView/shared/VideoVue.vue
```diff
@@ -361,6 +339,12 @@ export default {
 			}) || {}
 		},
 
+		participantActorType() {
+			return this.participant?.actorType || this.peerData.actorType || this.participantUserId
+				? ATTENDEE.ACTOR_TYPE.USERS
+				: ATTENDEE.ACTOR_TYPE.GUESTS
```
DorraJaouad, line 352: ?

`return this.$store.getters.getActorType()`

Antreesy: Store getter is from actorStore and used only for defining self.
For VideoVue we need to get the type of other participant

Antreesy: Should be fixed though for proper return:
```js
if (this.participant?.actorType) {
  return this.participant.actorType
} else if (this.peerData?.actorType) {
  return this.peerData.actorType
} else {
  return this.participantUserId
    ? ATTENDEE.ACTOR_TYPE.USERS
    : ATTENDEE.ACTOR_TYPE.GUESTS
}
```

src/components/AvatarWrapper/AvatarWrapper.vue
Antreesy: These colors are also used in other places, to identify guest person in participants list or messages, for example:
image

Although I can't reproduce your second screenshot, Maybe, it's related to issue in second comment

DorraJaouad: This identification method might be okay in other places like messages and participant lists because it doesn't own much of "visual weight" but for call view, it occupies more visual surface. 
Alternative : Add "( guest )" to the display name to identify the guest.

Antreesy: Please reference all this in follow-up issue =)

DorraJaouad (approved):
Tested 🦅 
Avatar/space ratio looks better.

Issue #10348, opened 2023-08-24
⏺️ Recording consent - Overview


nickvergessen:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

For compliance reasons with various privacy rights it is really helpful if the operator can prove that everyone agreed that a call was recorded. Therefore they had the idea that users should actively agree that the call gets recorded.

1. When someone joins the room while a call is recorded it is easy: We already show a "warning" in the device dialog that the call is recorded, we could add a checkbox there. The user can only join the room/call when they checked the checkbox.

2. For people already in the room it is a bit more tricky when someone want to start the recording. Some ideas they had:
    *  We could ask the users for consensus, e.g. with a pop up when the moderator starts the recording
    *  When everyone agreed the recording starts directly
    * When not everyone agrees:
        * The moderator can see who agreed/disagreed in the sidebar (e.g. a icon next to each user). 
        * The moderator can remove the user from the room (once the room only consist of people who agreed the recording starts)
        * The moderator can overrule the user (Maybe a important participant you don't want/cant kick out but know that ultimately he will be fine with the recording)

The process would be logged in the database, similar to the ToS-App so that the admin can later prove that everyone agreed to the recording. An admin can then request the list of "agrees" for a given conversation via OCC

Some additional thoughts:

* This process might be to strict for some users. Therefore we would probably need a admin settings "require active approval for call recording" or something similar so that a admin can decide if they want to enforce it or not.
* We could add a room property and when it is "recording consent required" we always show the modal (like when the recording was started before) and add a tick box in front of the warning and disabling the "Join/start call" unless the checkbox was ticked.

---

- [x] API https://github.com/nextcloud/spreed/issues/10408
- [x] Frontend https://github.com/nextcloud/spreed/issues/10409
- [x] Android https://github.com/nextcloud/talk-android/issues/3289
- [ ] iOS https://github.com/nextcloud/talk-ios/issues/1345

marcoambrosini:
[...]

PR #10633, opened 2023-10-04
feat(recording) - add frontend support for recording consent

### LINKED ISSUE #10409: Frontend "Recording consent"
* Overview https://github.com/nextcloud/spreed/issues/10348

---

- [ ] Option to require consent
- [ ] When consent is required
  - [ ] Enforce the media settings dialog 
  - [ ] Show a checkbox that needs ticking before one can "Join call" / "Start call"

Antreesy:
### ☑️ Resolves

* Fix #10409
* Solution is based on MVP + V2 description (https://github.com/nextcloud/spreed/issues/10348#issuecomment-1725333919)
  * Option to require consent globally in Admin settings
  * When consent is required `1`:
    * Enforce the media settings dialog
    * Show a checkbox that needs ticking before moderator/user can start or join call
    * Moderator setting is visible but disabled
  * When consent is optional `2`:
    * Moderator setting is enabled with "Off" and "Required"
    * Disabled, while call is currently on
  * If user accidentally joined without consent, will be blocked by API
* Increase preview size in MediaSettings, align all content elements up to its width

#### Admin settings
image

#### Moderator settings

| Off (hidden) | Required (disabled) | Configurable V2 |
|---|------------|----------|
| image | image | image

#### User view (Media settings)

| Off (hidden) | Required (not checked) | Required (checked) |
|---|------------|----------|
| image | image | image

### 🚧 Tasks

- [ ] Adjust wording / styles
- [ ] Test with recording package
- [ ] Test frontend (enter in AdminSettings -> Recording backend credentials for working signaling server)

### 🏁 Checklist

### DIFF
--- a/src/components/AdminSettings/RecordingServers.vue
+++ b/src/components/AdminSettings/RecordingServers.vue
@@ -65,6 +65,26 @@
 			:label="t('spreed', 'Shared secret')"
 			label-visible
 			@update:value="updateSecret" />
+
+		<template v-if="servers.length && recordingConsentCapability">
+			<h3>{{ t('spreed', 'Recording consent') }}</h3>
+
+			<template v-for="level in recordingConsentOptions">
+				<NcCheckboxRadioSwitch :key="level.value + '_radio'"
+					:value="level.value.toString()"
+					:checked.sync="recordingConsentSelected"
+					name="recording-consent"
+					type="radio"
+					:disabled="loading"
+					@update:checked="setRecordingConsent">
+					{{ level.label }}
+				</NcCheckboxRadioSwitch>
+
+				<p :key="level.value + '_description'" class="consent-description">
+					{{ getRecordingConsentDescription(level.value) }}
+				</p>
+			</template>
+		</template>
 	</section>
 </template>
 
@@ -73,27 +93,45 @@ import debounce from 'debounce'
 
 import Plus from 'vue-material-design-icons/Plus.vue'
 
+import { getCapabilities } from '@nextcloud/capabilities'
 import { showSuccess } from '@nextcloud/dialogs'
 import { formatFileSize } from '@nextcloud/files'
 import { loadState } from '@nextcloud/initial-state'
 
 import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
+import NcCheckboxRadioSwitch from '@nextcloud/vue/dist/Components/NcCheckboxRadioSwitch.js'
 import NcNoteCard from '@nextcloud/vue/dist/Components/NcNoteCard.js'
 import NcTextField from '@nextcloud/vue/dist/Components/NcTextField.js'
 
 import RecordingServer from '../../components/AdminSettings/RecordingServer.vue'
+import { CALL } from '../../constants.js'
 import TransitionWrapper from '../TransitionWrapper.vue'
+
+const recordingConsentCapability = getCapabilities()?.spreed?.features?.includes('recording-consent')
+const recordingConsentOptions = [
+	{ value: CALL.RECORDING_CONSENT.OFF, label: t('spreed', 'Disabled for all calls') },
+	{ value: CALL.RECORDING_CONSENT.REQUIRED, label: t('spreed', 'Enabled for all calls') },
+	{ value: CALL.RECORDING_CONSENT.OPTIONAL, label: t('spreed', 'Configurable on conversation level by moderators') },
+]
 
 export default {
 	name: 'RecordingServers',
 
 	components: {
 		NcButton,
+		NcCheckboxRadioSwitch,
 		NcNoteCard,
 		NcTextField,
 		Plus,
 		RecordingServer,
 		TransitionWrapper,
+	},
+
+	setup() {
+		return {
+			recordingConsentCapability,
+			recordingConsentOptions,
+		}
 	},
 
 	data() {
@@ -103,6 +141,7 @@ export default {
 			uploadLimit: 0,
 			loading: false,
 			saved: false,
+			recordingConsentSelected: loadState('spreed', 'recording_consent').toString(),
 		}
 	},
 
@@ -164,6 +203,27 @@ export default {
 			})
 		},
 
+		setRecordingConsent(value) {
+			this.loading = true
+			OCP.AppConfig.setValue('spreed', 'recording_consent', value, {
+				success: function() {
+					this.loading = false
+				}.bind(this),
+			})
+		},
+
+		getRecordingConsentDescription(value) {
+			switch (value) {
+			case CALL.RECORDING_CONSENT.OPTIONAL:
+				return t('spreed', 'Moderators will be allowed to enable consent on conversation level. The consent to be recorded will be required for each participant before joining every call in this conversation.')
+			case CALL.RECORDING_CONSENT.REQUIRED:
+				return t('spreed', 'The consent to be recorded will be required for each participant before joining every call.')
+			case CALL.RECORDING_CONSENT.OFF:
+			default:
+				return t('spreed', 'The consent to be recorded is not required.')
+			}
+		},
+
 		toggleSave() {
 			this.saved = true
 			setTimeout(() => {
@@ -181,5 +241,15 @@ export default {
 
 .additional-top-margin {
 	margin-top: 10px;
+}
+
+h3 {
+	margin-top: 24px;
+	font-weight: 600;
+}
+
+.consent-description {
+	margin-bottom: 12px;
+	opacity: 0.7;
 }
 </style>
--- a/src/components/ConversationSettings/RecordingConsentSettings.vue
+++ b/src/components/ConversationSettings/RecordingConsentSettings.vue
@@ -1,0 +1,123 @@
+<!--
+  - @copyright Copyright (c) 2023 Maksim Sukharev <antreesy.web@gmail.com>
+  -
+  - @author Maksim Sukharev <antreesy.web@gmail.com>
+  -
+  - @license AGPL-3.0-or-later
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+-->
+
+<template>
+	<div class="app-settings-subsection">
+		<h4 class="app-settings-section__subtitle">
+			{{ t('spreed', 'Recording Consent') }}
+		</h4>
+		<NcCheckboxRadioSwitch v-if="canFullModerate && !isGlobalConsent"
+			type="switch"
+			:checked.sync="recordingConsentSelected"
+			:disabled="disabled"
+			@update:checked="setRecordingConsent">
+			{{ t('spreed', 'Require recording consent before joining call in this conversation') }}
+		</NcCheckboxRadioSwitch>
+		<p v-else-if="isGlobalConsent">
+			{{ t('spreed', 'Recording consent is required for all calls') }}
+		</p>
+		<p v-else>
+			{{ summaryLabel }}
+		</p>
+	</div>
+</template>
+
+<script>
+import { showError, showSuccess } from '@nextcloud/dialogs'
+
+import NcCheckboxRadioSwitch from '@nextcloud/vue/dist/Components/NcCheckboxRadioSwitch.js'
+
+import { CALL } from '../../constants.js'
+import { getCapabilities } from '@nextcloud/capabilities'
+
+const recordingConsent = getCapabilities()?.spreed?.config?.call?.['recording-consent']
+
+export default {
+	name: 'RecordingConsentSettings',
+
+	components: {
+		NcCheckboxRadioSwitch,
+	},
+
+	props: {
+		token: {
+			type: String,
+			default: null,
+		},
+
+		canFullModerate: {
+			type: Boolean,
+			default: true,
+		},
+	},
+
+	data() {
+		return {
+			loading: false,
+			recordingConsentSelected: !!CALL.RECORDING_CONSENT.OFF,
+		}
+	},
+
+	computed: {
+		conversation() {
+			return this.$store.getters.conversation(this.token) || this.$store.getters.dummyConversation
+		},
+
+		isGlobalConsent() {
+			return recordingConsent === CALL.RECORDING_CONSENT.REQUIRED
+		},
+
+		disabled() {
+			return this.loading || this.conversation.hasCall
+		},
+
+		summaryLabel() {
+			return this.conversation.recordingConsent === CALL.RECORDING_CONSENT.REQUIRED
+				? t('spreed', 'Recording consent is required for calls in this conversation')
+				: t('spreed', 'Recording consent is not required for calls in this conversation')
+		},
+	},
+
+	mounted() {
+		this.recordingConsentSelected = !!this.conversation.recordingConsent
+	},
+
+	methods: {
+		async setRecordingConsent(value) {
+			this.loading = true
+			try {
+				await this.$store.dispatch('setRecordingConsent', {
+					token: this.token,
+					state: value ? CALL.RECORDING_CONSENT.REQUIRED : CALL.RECORDING_CONSENT.OFF,
+				})
+				showSuccess(t('spreed', 'Recording consent requirement was updated'))
+			} catch (error) {
+				showError(t('spreed', 'Error occurred while updating recording consent'))
+				console.error(error)
+			}
+			this.loading = false
+		},
+	},
+}
+</script>
+
+<style lang="scss" scoped>
+</style>
--- a/src/components/MediaSettings/MediaSettings.vue
+++ b/src/components/MediaSettings/MediaSettings.vue
@@ -571,12 +571,7 @@ export default {
 
 <style lang="scss" scoped>
 .media-settings {
-	padding: calc(var(--default-grid-baseline) * 4);
-	background-color: var(--color-main-background);
-	overflow-y: auto;
-	overflow-x: hidden;
-	margin: auto;
-	width: 100%;
+	padding: calc(var(--default-grid-baseline) * 5);
 
 	&__title {
 		text-align: center;
@@ -591,8 +586,8 @@ export default {
 		overflow: hidden;
 		border-radius: calc(var(--default-grid-baseline) * 3);
 		background-color: var(--color-loading-dark);
-		height: 300px;
-		width: 400px;
+		width: 100%;
+		aspect-ratio: 4/3;
 	}
 
 	&__toggles-wrapper {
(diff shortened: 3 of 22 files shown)


Antreesy:
cc @marcoambrosini for design review and wording for MVP + V2 solution
cc @danxuliu as it touches signaling, to ensure it won't break anything

src/components/AdminSettings/RecordingServers.vue
```diff
@@ -73,36 +96,55 @@ import debounce from 'debounce'
...
 
 import RecordingServer from '../../components/AdminSettings/RecordingServer.vue'
+import { CALL } from '../../constants.js'
 import TransitionWrapper from '../TransitionWrapper.vue'
 
+const recordingConsentCapability = getCapabilities()?.spreed?.features?.includes('recording-consent')
+const recordingConsentOptions = [
+	{ id: CALL.RECORDING_CONSENT.OFF, label: t('spreed', 'Off') },
+	{ id: CALL.RECORDING_CONSENT.REQUIRED, label: t('spreed', 'Enabled') },
+	{ id: CALL.RECORDING_CONSENT.OPTIONAL, label: t('spreed', 'Optional') },
```
nickvergessen, line 117: Hmm it's not "optional" consent, but basically "on conversation level"
Maybe we find  a better wording ;P

src/components/ConversationSettings/RecordingConsentSettings.vue
```diff
@@ -0,0 +1,133 @@
...
+<template>
+	<div class="app-settings-subsection">
+		<h4 class="app-settings-section__subtitle">
+			{{ t('spreed', 'Recording Consent') }}
+		</h4>
+
+		<label for="recording-consent_input" class="recording-consent__label">
+			{{ t('spreed', 'Require recording consent before joining call in this conversation') }}
+		</label>
+		<NcSelect v-model="recordingConsentSelected"
```
nickvergessen, line 31: Don't show the select when it's enforced on admin level

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -130,6 +130,13 @@
 				type="warning">
 				<p>{{ t('spreed', 'The call is being recorded.') }}</p>
 			</NcNoteCard>
+			<!-- Recording consent warning -->
+			<NcNoteCard v-if="isRecordingConsentRequired" type="warning">
+				<p>{{ t('spreed', 'Your consent to be recorded is required before joining the call.') }}</p>
+				<NcCheckboxRadioSwitch :checked.sync="recordingConsentGiven">
+					{{ t('spreed', 'Give a consent to be recorded') }}
+				</NcCheckboxRadioSwitch>
```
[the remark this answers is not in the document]
Antreesy: Only if you participating as a speaker, so could be either voice/video/screen, or any combination

src/components/ConversationSettings/RecordingConsentSettings.vue
Antreesy: Should we do it the same as for Expiration settings? https://github.com/nextcloud/spreed/pull/10491

nickvergessen: Sounds good.

src/components/MediaSettings/MediaSettings.vue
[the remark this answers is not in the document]
nickvergessen: yeah, but permissions could be given later, so we don't know yet.
But that is why I would love to see a more neutral wording

marcoambrosini (commented):
The wording seems ok, I'd style the component a bit differently, similar to the quote component:
- border instead of background (better text contrast)
- Rounded corners
- Rounded warning bar on the left
- No warning icon
- Same width as the video component above

Quick sketch: 
IMG_8921 2

nickvergessen:
> I'd style the component a bit differently

It's the wode spread NcNoteCard component. If that should be restyled, I'd should be done upstream I guess.
The usecase is exactly what the component was designed for 🙈

src/components/AdminSettings/RecordingServers.vue
DorraJaouad: "Configurable" ?

Antreesy:
@marcoambrosini I'd agree with width aligning, but there are also other components wider than Video preview:
Device selectors:
image
Background selectors:
image

Maybe, we should increase a preview size instead?

danxuliu (changes requested):
There are two cases in which joining a call fails if recording consent is enabled: [switching between breakout rooms](https://github.com/nextcloud/spreed/blob/d833dd3a632eb8f57a63be771eca971e2987e722/src/App.vue#L347-L351) and [requesting the password of a share](https://github.com/nextcloud/spreed/blob/33c82ec006647582dcf33af4023d5959152d7ce4/src/PublicShareAuthSidebar.vue#L160-L164) (video verification). In those cases the recording consent is not asked, so the API prevents the user from joining the call.

## How to test (scenario 1):
- Open Talk settings
- Enable recording consent for all calls
- Create a new group conversation
- Add another user
- Set up breakout rooms with 2 rooms
- Start a call in that conversation; as it is the first time that a call is started the media settings screen will be shown
- Give recording consent
- Start the actual call
- Join the conversation
- Join the call; as it is the first time that a call is started the media settings screen will be shown
- Give recording consent
- Join the actual call
- In the original window, start the breakout rooms session

### Expected result
The user switches to the call in the breakout room (I would say that asking for recording consent again would be overkill and that the recording consent from the parent room should be reused; maybe this could be even done in the API itself)

### Actual result
The user switches to the call in the breakout room, but is immediately kicked out

## How to test (scenario 2):
- Open Talk settings
- Enable recording consent for all calls
- Open the Files app
- Create a link share
- Protect the share with a password
- Enable video verification
- Copy the link
- In a private window, open the link
- Request the password
- Set a name for the guest

### Expected result
The media settings screen (or just a recording consent dialog, but I guess it would be better to reuse the media settings screen) is shown to give recording consent

### Actual result
The user seems to join the call but is immediately kicked out

Independently of that, [the media settings screen should be shown](https://github.com/nextcloud/spreed/blob/f3cd81578ef3c1409701bed95e1138b0b2c6204d/src/components/TopBar/CallButton.vue#L339-L340) even if the user unchecked _Always show preview for this conversation_ when recording consent needs to be given to join a call:
## How to test
- Open Talk settings
- Enable recording consent for all calls, or make it optional and enable it in a specific call
- Create a new conversation, and if recording consent is optional, enable it
- Start a call in that conversation; as it is the first time that a call is started the media settings screen will be shown
- Uncheck _Always show preview for this conversation_
- Give recording consent
- Start the actual call
- Leave the call
- Reload the page
- Start a call

### Expected result
The media settings screen is shown to give recording consent

### Actual result
The user seems to join the call but is immediately kicked out

nickvergessen:
Case 1 (sending to breakout room) I guess we can fix on the API level by ignoring the setting there and basically allow "sending users" there. However when stopping breakout rooms they would join the parent again in which case we can not detect on API level if that is happening, so we would require the client to send the consent flag there again.

nickvergessen:
> However when stopping breakout rooms they would join the parent again in which case we can not detect on API level if that is happening, so we would require the client to send the consent flag there again.

But that is also a bit annoying, because it will be listed as another consent entry, but I guess it's the best step we have atm

nickvergessen:
The reconnection when your permissions are modified is also affected?

danxuliu:
> The reconnection when your permissions are modified is also affected?

Good catch. Yes, reconnections do not work because the consent is not sent again. The `silent` parameter is not taken into account either, I will check _tomorrow_ as that might need a separate fix to be able to backport it (and therefore this pull request would need a rebase on that fix to be able to also fix renegotiations when consent is required).

marcoambrosini (dismissed):
@Antreesy I think the use of selects here is not appropriate.
I think that the admin settings should be radio buttons, with an explanation always shown and not in a warning badge.
On the other hand, for the moderators on a per-conversation basis, we should use a toggle.

danxuliu:
> The `silent` parameter is not taken into account either, I will check _tomorrow_ as that might need a separate fix to be able to backport it (and therefore this pull request would need a rebase on that fix to be able to also fix renegotiations when consent is required).

Fix for `silent` parameter is in https://github.com/nextcloud/spreed/pull/10688 The second commit can be used as a reference on what to change (after rebasing) to send the consent again on reconnections.

Antreesy:
Rebased onto master, change interactive elements and wording
New screenshots attached, review appreciated =)

P.S. only UI changes, logic from Daniel is not touched yet, waiting for 10688 to be merged

Antreesy:
Rebase onto master, fix consent usage for App, FilesSidebarChatView, PublicShareSidebar

PublicShareAuthSidebar needs some adjustments to show MediaSettings before joining the call

src/views/FilesSidebarChatView.vue
```diff
@@ -23,7 +23,7 @@
 	<div class="talk-tab__wrapper">
 		<CallButton class="call-button" />
 		<ChatView />
-		<MediaSettings :initialize-on-mounted="false" />
+		<MediaSettings :initialize-on-mounted="false" :recording-consent-given.sync="recordingConsentGiven" />
```
DorraJaouad, line 26: So why are you adding `recordingConsentGiven` here ?

Antreesy: to be able to connect from files sidebar also, if consent is set to 1 (applied to all calls on instance)
consent may be reused when reconnecting, so should not be stored in MediaSettings component. Also don't want to get a new store for it

src/components/ConversationSettings/RecordingConsentSettings.vue
```diff
@@ -0,0 +1,123 @@
...
+			:disabled="disabled"
+			@update:checked="setRecordingConsent">
+			{{ t('spreed', 'Require recording consent before joining call in this conversation') }}
+		</NcCheckboxRadioSwitch>
+		<p v-else-if="isGlobalConsent">
+			{{ t('spreed', 'Recording consent is required for all calls') }}
+		</p>
+		<p v-else>
+			{{ summaryLabel }}
+		</p>
```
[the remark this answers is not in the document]
Antreesy: Yes, should be in Setup Overview when rebase on top of #10647

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -131,6 +131,21 @@
...
+				</p>
+				<p v-else>
+					{{ t('spreed', 'The call might be recorded.') }}
+				</p>
+				<p>
+					{{ t('spreed', 'The recording might include your voice, video from camera, and screen share. Your consent is required before joining the call.') }}
+				</p>
+				<NcCheckboxRadioSwitch :checked="recordingConsentGiven" @update:checked="setRecordingConsentGiven">
+					{{ t('spreed', 'Give consent to the recording of this call') }}
+				</NcCheckboxRadioSwitch>
```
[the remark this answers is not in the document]
Antreesy: or rewrite color

src/components/ConversationSettings/RecordingConsentSettings.vue
```diff
@@ -0,0 +1,123 @@
...
+	props: {
+		token: {
+			type: String,
+			default: null,
+		},
+
+		canFullModerate: {
+			type: Boolean,
+			default: true,
+		},
```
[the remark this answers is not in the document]
Antreesy: needed for Setup overview

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -591,8 +622,8 @@ export default {
 		overflow: hidden;
 		border-radius: calc(var(--default-grid-baseline) * 3);
 		background-color: var(--color-loading-dark);
-		height: 300px;
-		width: 400px;
+		height: 345px;
+		width: 460px;
```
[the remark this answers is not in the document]
Antreesy: It was aligned. Maybe, depends on whether scrollbar is visible

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -131,6 +131,21 @@
 				{{ t('spreed', 'Start recording immediately with the call') }}
 			</NcCheckboxRadioSwitch>
 
+			<!-- Recording consent warning -->
+			<NcNoteCard v-if="!isInCall && isRecordingConsentRequired" type="warning">
```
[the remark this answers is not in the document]
Antreesy: 1. It could be another moderator
2. Make sense for me to sync checkboxes with `isRecordingFromStart` and `recordingConsentGiven`

danxuliu (dismissed):
The media settings screen is not shown if the user unchecked _Always show preview for this conversation_ when recording consent needs to be given. Please refer to the last _How to test_ section in https://github.com/nextcloud/spreed/pull/10633#pullrequestreview-1668480416

Independently of that, the consent sometimes is sent again and sometimes it is not when using breakout rooms (first _How to test_ section in https://github.com/nextcloud/spreed/pull/10633#pullrequestreview-1668480416). I have not been able to find a pattern, maybe it is related to fetching the room information [to compare the parent room when switching to a different breakout room](https://github.com/nextcloud/spreed/pull/10633/commits/690c423a2c97ce2ef24281352de70701339e1af3#diff-7a7a37b12ee1265d7e27577ec286120d2cbc0940908635e264a2be44ccb9a8a0R232-R237) to clean or not to clean the consent, but I do not know.

When requesting the password of a share it might be good to leave the room if the media settings screen is closed without joining the call. Leaving the call or the conversation should update the sidebar and remove the call view and the chat and [show _This conversation has ended_](https://github.com/nextcloud/spreed/blob/33c82ec006647582dcf33af4023d5959152d7ce4/src/PublicShareAuthSidebar.vue#L28), although it seems that was lost at some point :thinking: As leaving the call does not currently do that I guess it would be fine to leave the conversation when the media settings is closed in a follow up once the existing code was also fixed.

To end in a bright note, the consent is now sent again in forced reconnections :-)

Antreesy:
> ...if the user unchecked _Always show preview for this conversation_ 

I might have lost these code lines... Moved logic from components to the store, to sync it properly between MediaSettings, ConversationSettingsDialog and CallButton (tests included) :green_circle: 


> ...maybe it is related to fetching the room information

I have relied on breakoutRoomsStore, but it works differently for moderators and non-moderators. Fixed :green_circle: 

> Leaving the call or the conversation should update the sidebar and remove the call view and the chat and [show _This conversation has ended_]

with internal signling only, we fetch conversation within an interval, and show context you mentioned. but with HPB it wait for signaling message. Don't know the logic here (cc @nickvergessen), but it seems like when the share owner leaves conversation, it's removed on server side. So, we may track signaling message, that owner left the room, and fetch the conversation within some timeout => and if it was deleted, all should go back to normal => will raise a follow-up issue

DorraJaouad (commented):
In case the consent is configurable on conversation level, I noticed when a moderator switches on the consent requirement while a participant has the conversation open ( on a chat view), it's not updated for them thus there is no consent requested when joining the call.

nickvergessen:
> In case the consent is configurable on conversation level, I noticed when a moderator switches on the consent requirement while a participant has the conversation open ( on a chat view), it's not updated for them thus there is no consent requested when joining the call.

Missing an update trigger via the HPB then, so it's a backend issue.

src/views/FilesSidebarChatView.vue
Antreesy: For "configurable" mode, consent will be switched off for new conversations like this, so not a problem
For "required" mode, I'd say that it's a pre-emptive measure (to not come back to it when we add owners/moderators to such conversations)

src/store/conversationsStore.js
```diff
@@ -565,6 +566,18 @@ const actions = {
 		commit('addConversation', conversation)
 	},
 
+	async setRecordingConsent({ commit, getters }, { token, state }) {
```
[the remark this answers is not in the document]
nickvergessen: Fixed in https://github.com/nextcloud/spreed/pull/10727

nickvergessen:
> In case the consent is configurable on conversation level, I noticed when a moderator switches on the consent requirement while a participant has the conversation open ( on a chat view), it's not updated for them thus there is no consent requested when joining the call.

Also fixed in https://github.com/nextcloud/spreed/pull/10727 now

src/views/FilesSidebarChatView.vue
DorraJaouad: To keep this pre-emptive measure, consent should not be mandatory atm as this might be a GDPR violation https://gdpr.eu/gdpr-consent-requirements/ (see "Consent must be freely given" section)  
> According to [Recital 42](https://gdpr.eu/Recital-42-Burden-of-proof-and-requirements-for-consent/), “Consent should not be regarded as freely given if the data subject has no genuine or free choice or is unable to refuse or withdraw consent without detriment.” 
The one exception is if you need some piece of data from someone to provide them with your service. For example, you may need their credit card information to process a transaction or their mailing address to ship a product.


For those participants who do not wish to consent yet, they want to join the call, you are mandating their consent despite that it does not fall under the one exception because you don't need their consent in this scenario.

Suggested alternative : An opt-in choice to consent or decline without impacting their ability to join the call

nickvergessen: > For those participants who do not wish to consent yet, they want to join the call, you are mandating their consent despite that it does not fall under the one exception because you don't need their consent in this scenario.

This is a customer requirement atm. It is not possible to join without giving consent on API level and this was approved by the customer.

> Suggested alternative : An opt-in choice to consent or decline without impacting their ability to join the call

This can be looked at as a potential follow up, but has quite a long trail (settings needs to be stored on a session based level, recording backend needs access to the user list and filter out peoples audio/video where consent was off, etc). Far out of scope for now, so this is what it will be `¯\_(ツ)_/¯` It's not getting worse then before where the recording was always done without even asking.

nickvergessen: > For those participants who do not wish to consent yet, they want to join the call, you are mandating their consent despite that it does not fall under the one exception because you don't need their consent in this scenario.

That being said, the user can still disable the microphone/camera and will not be recorded. So could say that this is the users choice here and we are basically good again.

Antreesy: It's clear that you're talking about file sharing conversations.
Although there are not any owners / moderators to be able to start the recording, technically it's possible (and we won't return to this code again)
Besides, it's possible to promote user with `occ talk:room:promote [token] [use]`

Antreesy:
Rebased onto master, should work for breakout rooms, and fetch conversation on change
Consent is kept for main view and sidebar integrations (turns out it's possible to promote people to moderators there with occ command at least)

src/views/FilesSidebarChatView.vue
DorraJaouad: Thanks for sharing that @Antreesy . In that case, can end this legal concerns discussion 🧑🏻‍⚖️for now.

DorraJaouad (approved):
Breakout rooms session should be closed when the moderator wants to toggle the consent requirement in the conversation settings . It renders Bad request otherwise ( maybe add a hint in toast ? )

Overall, nothing is blocking  🦅 .

nickvergessen (approved):

danxuliu (commented):
(I did not see that the review was dismissed before actually testing the pull request :facepalm:)

Tested and works :+1: 

A minor detail: when the device selector is shown in _request password_/_video verification_ clicking on _Always show preview for this conversation_ or in the menu to start a silent call has no effect (the checkbox is not modified, and the menu is not opened).

Nevertheless, those options should not be even visible, because every time a password is requested again a new room will be created, so there is no point in remembering _Always show preview for this conversation_, and the call should not be silently started, as the owner of the file should be notified that a password is being requested.

PR #10635, opened 2023-10-05
feat(recording): ⏺️ Add API for recording consent Version2

### LINKED ISSUE #10408: API "Recording consent"
* Overview https://github.com/nextcloud/spreed/issues/10348

---

- [x] OCC command to list / export consent entries for a room
- [x] Deny joining a call when consent is required but not sent on the request
- [x] Endpoint that allows setting the requirement
- [ ] #10726

nickvergessen:
### ☑️ Resolves

* Replacing https://github.com/nextcloud/spreed/pull/10634
* Fix #10408 

### 🚧 Tasks

- [x] Admin config for recording consent to enforce on the whole instance
- [x] Option to allow moderators to require consent on a per conversation level (can only be enabled when no call is on-going)
- [x] API endpoint to change the conversation setting
- [x] Capability for feature and admin config

### ➡️  Followup Tasks
- [ ] Store consent entries in database
- [ ] Add OCC command to list consent per actor or conversation or both

### 🏁 Checklist


Antreesy (approved):
Tested against frontend PR, compatible, works as expected.

Haven't tested usecase with changing consent as moderator, while in the call, but disabled for frontend, and seems to be locked in RoomServise as well

PR #10636, opened 2023-10-05
chore(PR): Split PR templates for UI and API work

nickvergessen:
* Made it a bit easier to remove unneeded things when a PR is only frontend or backend
* Adjusted the list of tasks a bit to match more what @Antreesy is posting all the time :P

> grafik


Antreesy (approved):
:blue_heart:

PR #10637, opened 2023-10-05
feat(MediaSettings) start recording together with call

Antreesy:
### ☑️ Resolves

* Ease usage of recording function for moderators

From moderator perspective:

| Before call start | During call (hidden) |
|---|---|
| Screenshot from 2023-10-05 13-49-41 | Screenshot from 2023-10-05 13-49-51 |

### 🚧 Tasks

- [ ] Test against recording server

### 🏁 Checklist

### DIFF
--- a/src/components/MediaSettings/MediaSettings.vue
+++ b/src/components/MediaSettings/MediaSettings.vue
@@ -163,7 +163,7 @@
 				<!-- Join call -->
 				<CallButton v-if="!isInCall"
 					class="call-button"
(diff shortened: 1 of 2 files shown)


nickvergessen:
> Before call start (checked)

Will cause a jump in height of the dialog? If so I would skip showing the warning for the user toggling the checkbox

DorraJaouad (approved):
Nothing is blocking code-wise.

PR #10640, opened 2023-10-05
fix(messagesStore) - Mentions in forwarded messages

### LINKED ISSUE #10631: RFC: Mentions in forwarded messages
ATM we handle mentions by names here: https://github.com/nextcloud/spreed/blob/77a4405ae0f0e315fa4e412bf04bb8c987fdde5b/src/store/messagesStore.js#L1266

Depending on correlation id/name it will be rendered in different conversation if:
- `name === id`
- `name.split(' ').includes(id)`.

Notification won't be triggered for this person though (or something prevents reproducing it)
When developing, all mention options should be considered:
```
@all
@user1
@"user@2"
@"user 3"
```

See screenshot attached
image

_Originally posted by @Antreesy in https://github.com/nextcloud/spreed/pull/10603#discussion_r1345572096_

DorraJaouad:
### ☑️ Resolves

* Fix #10631

| Before 😞  | 🏚️ After: The actual message | 🏡 After : The forwarded message |
|------------|------------|----------|
| image| image         | image       |

### 🚧 Tasks

- [ ] visual review
- [ ] code review

### 🏁 Checklist

### DIFF
--- a/src/store/messagesStore.js
+++ b/src/store/messagesStore.js
@@ -1306,7 +1306,8 @@ const actions = {
 		for (const key in message.messageParameters) {
 			if (key.startsWith('mention')) {
 				const mention = message.messageParameters[key]
-				message.message = message.message.replace(`{${key}}`, `@${mention.name}`)
+				const mentionString = key.includes('mention-call') ? `**${mention.name}**` : `@"${mention.id}"`
+				message.message = message.message.replace(`{${key}}`, mentionString)
 			}
 		}
 
--- a/src/store/messagesStore.spec.js
+++ b/src/store/messagesStore.spec.js
@@ -1935,7 +1935,7 @@ describe('messagesStore', () => {
 			messageToBeForwarded = {
 				id: 1,
 				token: TOKEN,
-				message: 'Hello {mention-user1}, {mention-user2}',
+				message: 'Hello {mention-user1}, {mention-user2}, and {mention-call1}',
 				messageParameters: {
 					'mention-user1': {
 						id: 'taylor',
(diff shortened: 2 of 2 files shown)


Antreesy (approved):
Tested with `displayName.contains(id)`, `displayName.contains(' ')`, `displayName === id`, and `@all`

image

Person will be notified with the forwarded message again, but should be expected (`silent: false`) and intended (if user forwards some tasks to internal chat, for example)

src/store/messagesStore.js
```diff
@@ -1306,7 +1306,8 @@ const actions = {
 		for (const key in message.messageParameters) {
 			if (key.startsWith('mention')) {
 				const mention = message.messageParameters[key]
-				message.message = message.message.replace(`{${key}}`, `@${mention.name}`)
+				const mentionString = key.includes('mention-call') ? `**${mention.name}**` : `@"${mention.id}"`
```
nickvergessen, line 1309: Not sure it causes issues somewhere, but usually we only quote when a space or slash is contained in the id

nickvergessen (approved):

src/store/messagesStore.js
DorraJaouad: It also works for ids that do not contain spaces of slashes, maybe a specific use case I am not aware of?

nickvergessen: https://github.com/nextcloud-libraries/nextcloud-vue/blob/master/src/components/NcRichContenteditable/NcMentionBubble.vue#L85-L89

Maybe it's only on NcRichContenteditable

nickvergessen:
Backport to 27?

PR #10671, opened 2023-10-10
Use new workflow templates

nickvergessen:
### ☑️ Resolves

* Ref https://github.com/nextcloud/.github/pull/230


.github/workflows/phpunit-pgsql.yml
```diff
@@ -33,9 +21,34 @@ concurrency:
...
+              - 'vendor/**'
+              - 'vendor-bin/**'
+              - '.php-cs-fixer.dist.php'
+              - 'composer.json'
+              - 'composer.lock'
+
   phpunit-pgsql:
     runs-on: ubuntu-latest
 
+    if: needs.changes.outputs.src == 'true'
```
kesselb, line 50:
```suggestion
    needs: changes
    if: needs.changes.outputs.src == 'true'
```

To wait for the changes job, otherwise result is not set.

kesselb: I'm not 100% sure, but for the summary job, we may need "needs"  for both jobs to access the results.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idoutputs

(the downstream jobs part)

nickvergessen: > we may need "needs" for both jobs to access the results.

Yes, I just played around and came to the same result. it's fixed now.

nickvergessen: Here is a run with affected changes:
https://github.com/nextcloud/spreed/actions/runs/6466794906
And here without:
https://github.com/nextcloud/spreed/actions/runs/6466828174

kesselb (approved):

Issue #10498, opened 2023-09-14
Filtered conversation view: don't remove current conversation until I navigate away from it


marcoambrosini:
## Steps to reproduce
1. Filter unread conversation
2. Select an unread conversation
3. Read all the messages

### Expected behavior
The conversation should not disappear from the list once all the messages are read. It should only be cleared when I navigate away from it.

### Actual behavior
The conversation disappears

PR #10679, opened 2023-10-11
Fix(LeftSidebar): Show hint when no filtered results

DorraJaouad:
### ☑️ Resolves

`Hint` component was at the bottom of the `LeftSidebar` because the `ConversationListVirtual` is always present at full height.
* Added utility for conversation
* Refactored filter handling in leftSidebar
* Added captions for unread messages and unread mentions filter
* Add the empty content when there are no results

The filtering logic : 
"Standard Condition" :  if conversation  has call , mentions / unread message and is currently active in the chat view. 
"Special Condition" : if conversation is currently active, the only one in the list and has no flag. 
* At the moment the filter is first applied:

  - Apply Standard Condition.
  - Check Special Condition:
     * If yes, hide the current conversation by returning an empty list. This handles the case of clicking the filter initially and showing empty content if there are 0 unread conversations
      * If no, show the filtered list

* After filters are applied (during subsequent filter navigation):

  - Apply Standard Condition.

| 🏚️ Before | 🏡 After | Rare but possible ( empty list) |
|------------|----------|----------|
| image   | image    |image |

### 🚧 Tasks

- [ ] Code review

### 🏁 Checklist

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -121,16 +121,29 @@
 				<ul class="h-100" :class="{'scroller': isSearching}">
 					<!-- Conversations List -->
 					<template v-if="!isSearching">
-						<li class="h-100">
+						<NcEmptyContent v-if="initialisedConversations && filteredConversationsList.length === 0"
+							:name="t('spreed', 'No matches found')"
+							:description="displayedMessage">
+							<template #icon>
+								<MessageBadge v-if="isFiltered" :size="64" />
+								<MessageOutline v-else :size="64" />
+							</template>
+							<template #action>
+								<NcButton v-if="isFiltered" @click="handleFilter(null)">
+									<template #icon>
+										<FilterRemoveIcon :size="20" />
+									</template>
+									{{ t('spreed', 'Clear filter') }}
+								</NcButton>
+							</template>
+						</NcEmptyContent>
+						<li v-show="filteredConversationsList.length > 0" class="h-100">
 							<ConversationsListVirtual ref="scroller"
 								:conversations="filteredConversationsList"
 								:loading="!initialisedConversations"
 								class="scroller h-100"
 								@scroll.native="debounceHandleScroll" />
 						</li>
-						<Hint v-if="initialisedConversations && filteredConversationsList.length === 0"
-							:hint="t('spreed', 'No matches found')" />
-
 						<NcButton v-if="!preventFindingUnread && lastUnreadMentionBelowViewportIndex !== null"
 							class="unread-mention-button"
 							type="primary"
@@ -252,6 +265,7 @@ import List from 'vue-material-design-icons/FormatListBulleted.vue'
 import MessageBadge from 'vue-material-design-icons/MessageBadge.vue'
 import Note from 'vue-material-design-icons/NoteEditOutline.vue'
 import Plus from 'vue-material-design-icons/Plus.vue'
+import MessageOutline from 'vue-material-design-icons/MessageOutline.vue'
 
 import { showError } from '@nextcloud/dialogs'
 import { emit } from '@nextcloud/event-bus'
@@ -264,6 +278,7 @@ import NcAppNavigationCaption from '@nextcloud/vue/dist/Components/NcAppNavigati
 import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
 import NcListItem from '@nextcloud/vue/dist/Components/NcListItem.js'
 import isMobile from '@nextcloud/vue/dist/Mixins/isMobile.js'
+import NcEmptyContent from '@nextcloud/vue/dist/Components/NcEmptyContent.js'
 
 import ConversationIcon from '../ConversationIcon.vue'
 import Hint from '../Hint.vue'
@@ -287,6 +302,7 @@ import { EventBus } from '../../services/EventBus.js'
 import { talkBroadcastChannel } from '../../services/talkBroadcastChannel.js'
 import CancelableRequest from '../../utils/cancelableRequest.js'
 import { requestTabLeadership } from '../../utils/requestTabLeadership.js'
+import { filterFunction } from '../../utils/conversation.js'
 
 export default {
 
@@ -310,6 +326,7 @@ export default {
 		// Icons
 		AtIcon,
 		MessageBadge,
+		MessageOutline,
 		FilterIcon,
 		FilterRemoveIcon,
 		Plus,
@@ -317,6 +334,7 @@ export default {
 		List,
 		DotsVertical,
 		Note,
+		NcEmptyContent,
 	},
 
 	mixins: [
@@ -366,6 +384,7 @@ export default {
 			isCurrentTabLeader: false,
 			isFocused: false,
 			isFiltered: null,
+			isNavigating: false,
 		}
 	},
 
@@ -386,21 +405,39 @@ export default {
 			}
 		},
 
+		token() {
+			return this.$store.getters.getToken()
+		},
+
+		displayedMessage() {
+			switch (this.isFiltered) {
+			case 'mentions':
+				return t('spreed', 'You have no unread mentions in your inbox.')
+			case 'unread':
+				return t('spreed', 'You have no unread messages in your inbox.')
+			default:
+				return t('spreed', 'Your inbox is empty.')
+			}
+		},
+
 		filteredConversationsList() {
 			if (this.isFocused) {
 				return this.conversationsList
 			}
-
-			if (this.isFiltered === 'unread') {
-				return this.conversationsList.filter(conversation => conversation.unreadMessages > 0
-				|| conversation.hasCall || conversation.token === this.$store.getters.getToken())
-			}
-
-			if (this.isFiltered === 'mentions') {
-				return this.conversationsList.filter(conversation => conversation.unreadMention
-				|| conversation.hasCall
-				|| (conversation.unreadMessages > 0 && (conversation.type === CONVERSATION.TYPE.ONE_TO_ONE || conversation.type === CONVERSATION.TYPE.ONE_TO_ONE_FORMER))
-				|| conversation.token === this.$store.getters.getToken())
+			// applying filters
+			if (this.isFiltered) {
+				let validConversationsCount = 0
+				const filteredConversations = this.conversationsList.filter((conversation) => {
+					const conversationIsValid = filterFunction(this.isFiltered, conversation)
+					if (conversationIsValid) {
+						validConversationsCount++
+					}
+					return conversationIsValid
+						|| conversation.hasCall
+						|| conversation.token === this.token
+				})
+				// return empty if it only includes the current conversation without any flags
+				return validConversationsCount === 0 && !this.isNavigating ? [] : filteredConversations
 			}
 
 			return this.conversationsList
@@ -445,6 +482,14 @@ export default {
 						? t('spreed', 'Circles')
 						: t('spreed', 'Other sources')
 				}
+			}
+		},
+	},
+
+	watch: {
+		token(value) {
+			if (value && this.isFiltered) {
+				this.isNavigating = true
 			}
 		},
 	},
@@ -524,16 +569,16 @@ export default {
 
 		handleFilter(filter) {
 			this.isFiltered = filter
-
 			// Store the active filter
 			if (filter) {
 				BrowserStorage.setItem('filterEnabled', filter)
 			} else {
 				BrowserStorage.removeItem('filterEnabled')
 			}
-
 			// Clear the search input once a filter is active
 			this.searchText = ''
+			// Initiate the navigation status
+			this.isNavigating = false
 		},
 
 		scrollBottomUnread() {
@@ -924,6 +969,11 @@ export default {
 		font-weight: bold;
 	}
 
+}
+
+:deep(.empty-content) {
+	text-align: center;
+	padding: 20% 10px 0;
 }
 
 .settings-button {
--- a/src/utils/conversation.js
+++ b/src/utils/conversation.js
@@ -1,0 +1,59 @@
+/**
+ * @copyright Copyright (c) 2023 <dorra.jaoued7@gmail.com>
+ *
+ * @author Dorra Jaouad <dorra.jaoued7@gmail.com>
+ *
+ * @license AGPL-3.0-or-later
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+import { CONVERSATION } from '../constants.js'
+
+/**
+ * check if the conversation has unread messages
+ *
+ * @param {object} conversation conversation object
+ * @return {boolean}
+ */
+export function hasUnreadMessages(conversation) {
+	return conversation.unreadMessages > 0
+}
+
+/**
+ * check if the conversation has unread mentions
+ *
+ * @param {object} conversation conversation object
+ * @return {boolean}
+ */
+export function hasUnreadMentions(conversation) {
+	return conversation.unreadMention
+		|| conversation.unreadMentionDirect
+		|| (conversation.unreadMessages > 0
+			&& (conversation.type === CONVERSATION.TYPE.ONE_TO_ONE || conversation.type === CONVERSATION.TYPE.ONE_TO_ONE_FORMER))
+}
+
+/**
+ * apply the active filter
+ *
+ * @param {string} filter the filter option
+ * @param {object} conversation conversation object
+ */
+export function filterFunction(filter, conversation) {
+	if (filter === 'unread') {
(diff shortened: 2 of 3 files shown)


src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -121,15 +121,15 @@
...
 						<li class="h-100">
 							<ConversationsListVirtual ref="scroller"
 								:conversations="filteredConversationsList"
 								:loading="!initialisedConversations"
 								class="h-100"
 								@scroll.native="debounceHandleScroll" />
 						</li>
-						<Hint v-if="initialisedConversations && filteredConversationsList.length === 0" :hint="t('spreed', 'No matches found')" />
-
+
```
nickvergessen: >  GitHub Actions / NPM lint
>
> Trailing spaces not allowed



```suggestion

```

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -121,15 +121,15 @@
 				<ul class="scroller h-100">
 					<!-- Conversations List -->
 					<template v-if="!isSearching">
+						<Hint v-if="initialisedConversations && filteredConversationsList.length === 0" :hint="t('spreed', 'No matches found')" />
```
nickvergessen, line 124: Maybe we can include the filter that is currently applied?
And maybe also a button to remove the filter?

Antreesy: We can add applied filter as NcAppNavigationCaption in that case, I guess
But title-name prop change should be considered, as we haven't merged lib yet

DorraJaouad: Should it be kept bold and blue ?
image

Antreesy: I thought about

```
### Unread mentions
No matches found
```

src/views/Dashboard.vue
```diff
@@ -65,7 +65,7 @@ import NcEmptyContent from '@nextcloud/vue/dist/Components/NcEmptyContent.js'
 
 import ConversationIcon from './../components/ConversationIcon.vue'
 
-import { CONVERSATION } from '../constants.js'
+import conversation from '../mixins/conversation.js'
```
Antreesy, line 68: Dashboard entrypoint will be removed with API v2 (see PR from Raimund), so I'd not change anything here, as it will be lost

Antreesy, line 68: Dashboard entrypoint will be removed with API v2 (see PR from Raimund), so I'd not change anything here, as it will be lost

src/mixins/conversation.js
```diff
@@ -0,0 +1,40 @@
+/**
+ * @copyright Copyright (c) 2019
```
Antreesy, line 2: :eyes: 

Let's do it a util instead. We're trying to get rid of mixins in projects, iirc
Besides, if it's not going to be used in Dashboard, we could just keep it as LeftSidebar method

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -926,6 +985,16 @@ export default {
...
 }
 
+.clear-filter__button {
+	border-radius : var(--border-radius-pill);
+	overflow: hidden;
+}
+
+:deep(.empty-content) {
+	text-align: center;
+	margin-top: 10%;
```
Antreesy, line 995: default appearance for empty content is to center it in the sidebar. maybe stick to that?

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -121,6 +121,23 @@
...
 					<template v-if="!isSearching">
+						<NcAppNavigationCaption v-if="isFiltered && filteredConversationsList.length > 0" :name="getFilterCaption" />
+						<NcEmptyContent v-else-if="initialisedConversations && filteredConversationsList.length === 0"
+							:name="t('spreed', 'No matches found')"
+							:description="messageDisplay">
+							<template #icon>
+								<MessageBadge :size="64" />
+							</template>
+							<template #action>
+								<NcActionButton class="clear-filter__button"
```
Antreesy, line 132: Why not use NcButton? It has different appearance and hover feedback at least

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -386,21 +407,59 @@ export default {
...
+
+				return includesCurrentConversationOnly ? [] : filteredConversations
 			}
 
 			if (this.isFiltered === 'mentions') {
-				return this.conversationsList.filter(conversation => conversation.unreadMention
-				|| conversation.hasCall
-				|| (conversation.unreadMessages > 0 && (conversation.type === CONVERSATION.TYPE.ONE_TO_ONE || conversation.type === CONVERSATION.TYPE.ONE_TO_ONE_FORMER))
-				|| conversation.token === this.$store.getters.getToken())
+				const filteredConversations = this.conversationsList.filter(function(conversation) {
```
Antreesy, line 450: Looks like a duplicated logic. Maybe move everything to the method, with this.isFiltered as filtration argument?

src/mixins/conversation.js
DorraJaouad: I thought about making it independent from LeftSidebar so we can have quick methods to check the mention flag and the unread flag in the future without retyping all of those conditions.

src/components/LeftSidebar/LeftSidebar.vue
DorraJaouad: For the leftSidebar .. idk it seemed too low especially that user is expecting something to show at the very top. Setting a default appearance for empty content can be tricky. For square-like shapes or horizontal rectangles, this default is okay but for vertical rectangles, IMO it really depends on the situation.

src/mixins/conversation.js
Antreesy: Then also could be a util, or composable.

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -387,21 +405,41 @@ export default {
...
+								|| conversation.token === this.token
+				}.bind(this))
+
+				// return empty if it only includes the current conversation without any flags
+				const includesCurrentConversationOnly = filteredConversations.length === 1
+					&& filteredConversations.filter(function(conversation) {
+						return !this.filterFunction(conversation) && conversation.token === this.token
+					}.bind(this)).length
+
+				return includesCurrentConversationOnly ? [] : filteredConversations
```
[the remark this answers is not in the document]
Antreesy: Same for the hasCall. Then in filter function we can count amount of `this.filterFunction(conversation) === true`, and if it's equal 0, show EmptyContent. Thus, current conversation and ongoing call will still be always visible

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -121,7 +121,22 @@
 				<ul class="h-100" :class="{'scroller': isSearching}">
 					<!-- Conversations List -->
 					<template v-if="!isSearching">
-						<Hint v-if="initialisedConversations && filteredConversationsList.length === 0" :hint="t('spreed', 'No matches found')" />
+						<NcAppNavigationCaption v-if="isFiltered && filteredConversationsList.length > 0" :name="getFilterCaption" />
+						<NcEmptyContent v-else-if="initialisedConversations && filteredConversationsList.length === 0"
```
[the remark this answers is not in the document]
Antreesy: and add `class: "scroller"` for paddings

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -387,21 +405,41 @@ export default {
...
-				|| conversation.hasCall
-				|| (conversation.unreadMessages > 0 && (conversation.type === CONVERSATION.TYPE.ONE_TO_ONE || conversation.type === CONVERSATION.TYPE.ONE_TO_ONE_FORMER))
-				|| conversation.token === this.$store.getters.getToken())
+			// applying filters
+			if (this.isFiltered) {
+				const filteredConversations = this.conversationsList.filter(function(conversation) {
+					return this.filterFunction(conversation)
+								|| conversation.hasCall
+								|| conversation.token === this.token
+				}.bind(this))
```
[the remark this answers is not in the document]
DorraJaouad: I still need the token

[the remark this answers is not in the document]
Antreesy: Should be available, as it's a computed now, and being used in another computed. Replacing `function(conversation)` with `(conversation) =>` should work

src/components/LeftSidebar/LeftSidebar.vue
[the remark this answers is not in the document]
DorraJaouad: It seems super confusing to have the empty content and the current conversation at the same time. Also, if I filter and find the current conversation, I have a thought that there must be a new message in the conversation that came while it is open.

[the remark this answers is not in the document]
nimishavijay: I would agree with @DorraJaouad here, since the current conversation is open in the main content anyway, we don't need to show it in the results :) Even when we search for a conversation using the search bar, the results don't include the current conversation, so we can follow that same logic for filters too :)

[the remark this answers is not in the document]
nickvergessen: Note this was specifically requested from the design team:
https://github.com/nextcloud/spreed/issues/10498

cc @marcoambrosini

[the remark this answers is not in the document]
nimishavijay: Isn't it marked as read only when you scroll fully to the bottom? So when we open a conversation with some unread messages it would still be unread right? In that case it would show up in the list anyway.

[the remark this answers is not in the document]
Antreesy: Depending on amount of unread messages, it may be instantly triggered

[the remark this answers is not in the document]
DorraJaouad: > Note this was specifically requested from the design team:

This is related to the instant disappearance of conversation when marked read, while navigating in the filtered list.  

So, the current situation is  : 
* If there is at least one unread conversation, the current conversation will show in the list ( regardless of its unread flag) and will be removed when you navigate away. 
* If there are 0 unread conversations, it is an empty list.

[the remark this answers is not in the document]
nimishavijay: Suggestion:

- If there are 0 conversations with unread flag, show empty content
- If there is at least one conversation with unread flag, show list of conversations
- Current conversation is shown if there is an unread flag
- When you navigate away from current conversation, if it still has unread flag it stays in the list, if it has no unread flag it is removed from the list 

Does this make sense?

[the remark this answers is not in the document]
Antreesy: To clarify the usecase we have a discussion around:
- I have 3 unread conversations and Lomeapply the filter to show them = 3
- When I read a first conversation, list will be recomputed. Now it includes current conversation + 2 unread ones = 3
- Read next - previous current removed from list - now I see current + 1 unread = 2
- Read last - should see only current = 1
- :warning: but I see empty content instead = 0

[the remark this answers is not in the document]
nickvergessen: > Read last - should see only current = 1

yes

> ⚠️ but I see empty content instead = 0

You will see it when you click on the talk icon (top bar my main usecase for this PR here, just to have a reminder that it's still filtered) or are removed from the room.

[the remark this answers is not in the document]
Antreesy: [Screencast from 15.10.2023 13:58:13.webm](https://github.com/nextcloud/spreed/assets/93392545/c4b6b6d2-d3fa-47e9-b60b-0d73d24f406e)

Same haapens when you marked last conversation as read manually, or it's triggered by itself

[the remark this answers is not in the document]
DorraJaouad: Then, it should broken down as follows:
* At the moment the filter is first applied:

    - Check if the filtered list contains only the currently active conversation.
    - If so, hide the current conversation by returning an empty list.
    - This handles the case of clicking the filter initially and showing empty content if there are 0 unread conversations

* After filters are applied (during subsequent filter navigation):

    - The current active conversation should always be shown (regardless of its flag).
    - This will solve the disappearance of the last-read conversation.

https://github.com/nextcloud/spreed/assets/84044328/97f83aa0-3279-4cf3-889c-15b5f8bdeff9

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -386,21 +404,39 @@ export default {
...
-				|| conversation.token === this.$store.getters.getToken())
+			// applying filters
+			if (this.isFiltered) {
+				const filteredConversations = this.conversationsList.filter((conversation) => this.filterFunction(conversation)
+					|| conversation.hasCall
+					|| conversation.token === this.token)
+
+				// return empty if it only includes the current conversation without any flags
+				const isNewFilter = this.isPreviousFilter() !== this.isFiltered
+				const hasCurrentConversationOnly = filteredConversations.length === 1
```
Antreesy, line 435: `=== 1` won't work if you have current conversation + conversation in call.
We may count inside of `this.conversationsList.filter()` how many conversations are truthy for `this.filterFunction`, and then check that number (if at least 1 unread won't be there, show `[ ]`)

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -121,16 +121,30 @@
...
+							<template #action>
+								<NcButton @click="handleFilter(null)">
+									<template #icon>
+										<FilterRemoveIcon :size="20" />
+									</template>
+									{{ t('spreed', 'Clear filter') }}
+								</NcButton>
+							</template>
+						</NcEmptyContent>
 						<li class="h-100">
```
Antreesy, line 141:
```suggestion
						<li v-show="filteredConversationsList.length > 0" class="h-100">
```
If there is nothing to show, component is still there, and adds a scroll

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -522,18 +558,32 @@ export default {
...
 				BrowserStorage.removeItem('filterEnabled')
 			}
-
 			// Clear the search input once a filter is active
 			this.searchText = ''
+			// Initiate the previous filter
+			this.previousFilter = null
+		},
+
+		filterFunction(conversation) {
```
Antreesy, line 581: Maybe, move this function to util completely, to have only one entrypoint from another file?

src/utils/conversation.js
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * check if the conversation has unread mentions
+ *
+ * @param {object} conversation conversation object
+ * @return {boolean}
+ */
+export function hasUnreadMentions(conversation) {
+	return conversation.unreadMention
+		|| conversation.unreadMentionDirect
```
Antreesy, line 42: Please find and fix all places with `commit('updateUnreadMessages')` - they are missing setting unreadMentionDirect to false, which cause a bug in your PR

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -121,16 +121,30 @@
 				<ul class="h-100" :class="{'scroller': isSearching}">
 					<!-- Conversations List -->
 					<template v-if="!isSearching">
-						<li class="h-100">
+						<NcAppNavigationCaption v-if="isFiltered && filteredConversationsList.length > 0" :name="getFilterCaption" />
```
Antreesy, line 124: It also creates a scrollbar, because virtual list is rendered.
Check if changing `height: 100%` with `max-height: 100%` in class 'h-100' won't break anything, and I'd suggest to proceed with it

Antreesy (commented):
Don't forget to sqash commit, your first one was completely overwritten

src/utils/conversation.js
DorraJaouad: It breaks the smooth scrolling when conversation is not fully visible in the viewport ( check `ConversationsListVirtual` )

src/components/LeftSidebar/LeftSidebar.vue
DorraJaouad: It breaks the smooth scrolling when conversation is not fully visible in the viewport ( check `ConversationsListVirtual` )

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -386,21 +405,34 @@ export default {
...
-			if (this.isFiltered === 'mentions') {
-				return this.conversationsList.filter(conversation => conversation.unreadMention
-				|| conversation.hasCall
-				|| (conversation.unreadMessages > 0 && (conversation.type === CONVERSATION.TYPE.ONE_TO_ONE || conversation.type === CONVERSATION.TYPE.ONE_TO_ONE_FORMER))
-				|| conversation.token === this.$store.getters.getToken())
+			// applying filters
+			if (this.isFiltered) {
+				const filteredConversations = this.conversationsList.filter((conversation) => filterFunction(this.isFiltered, conversation)
+					|| conversation.hasCall
+					|| conversation.token === this.token)
```
Antreesy, line 431:
```suggestion
let validConversationsCount = 0
				const filteredConversations = this.conversationsList.filter((conversation) => {
					const conversationIsValid = filterFunction(this.isFiltered, conversation)
				if (conversationIsValid) {
					validConversationsCount++
				}
					return conversationIsValid
						|| conversation.hasCall
						|| conversation.token === this.token)
					}
```

Thus we don't need to filter the whole list again to get it's length (L434)

Antreesy (approved):
Left a last note (promise), how we can avoid going through the conversation list twice
Apart from that, looks and works nice!

src/components/LeftSidebar/LeftSidebar.vue
DorraJaouad: Nice workaround 👍🏻

PR #10683, opened 2023-10-11
fix(federation): Error the POST /participant request when federation failed

nickvergessen:
### ☑️ Resolves

* Fix adding a person from a non-existing server, have a cert error, etc.
* Currently there is no user feedback to be aware.

### 🏁 Checklist

### DIFF
--- a/lib/Controller/RoomController.php
+++ b/lib/Controller/RoomController.php
@@ -32,6 +32,7 @@ use InvalidArgumentException;
 use OCA\Talk\Config;
 use OCA\Talk\Events\BeforeRoomsFetchEvent;
 use OCA\Talk\Events\UserEvent;
+use OCA\Talk\Exceptions\CannotReachRemoteException;
 use OCA\Talk\Exceptions\ForbiddenException;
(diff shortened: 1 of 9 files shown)


lib/Controller/RoomController.php
```diff
@@ -1140,7 +1141,11 @@ public function addParticipantToRoom(string $newParticipant, string $source = 'u
 		}
 
 		// add the remaining users in batch
-		$this->participantService->addUsers($this->room, $participantsToAdd, $addedBy);
+		try {
+			$this->participantService->addUsers($this->room, $participantsToAdd, $addedBy);
+		} catch (CannotReachRemoteException $e) {
```
SystemKeeper, line 1146: Could we end up in a situation where only some users were added to the room? Like if the array contains 2 users and the first one throws a `CannotReachRemoteException`, what happens with the second one?

nickvergessen: the array case is only used for groups and circles atm. For normal users, email addresses and federated users the array will always only be a single entry as you have to send 1 API call per target to add 🙈

PR #10692, opened 2023-10-12
feat(federation): Register the OCM resource

nickvergessen:
### ☑️ Resolves

- [x] Depends on https://github.com/nextcloud/server/pull/40885

### 🏁 Checklist

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -84,6 +84,7 @@ use OCA\Talk\Events\SystemMessageSentEvent;
 use OCA\Talk\Events\SystemMessagesMultipleSentEvent;
 use OCA\Talk\Events\UserJoinedRoomEvent;
 use OCA\Talk\Federation\CloudFederationProviderTalk;
+use OCA\Talk\Federation\Proxy\TalkV1\Listener\ResourceTypeRegisterListener;
 use OCA\Talk\Federation\Proxy\TalkV1\Notifier\BeforeRoomDeletedListener as TalkV1BeforeRoomDeletedListener;
 use OCA\Talk\Federation\Proxy\TalkV1\Notifier\CancelRetryOCMListener as TalkV1CancelRetryOCMListener;
 use OCA\Talk\Federation\Proxy\TalkV1\Notifier\MessageSentListener as TalkV1MessageSentListener;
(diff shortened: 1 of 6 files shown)


lib/Federation/BeforeResourceTypesGetListener.php
```diff
@@ -0,0 +1,64 @@
...
+
+		if (!$this->talkConfig->isFederationEnabled()) {
+			return;
+		}
+
+		$resource = $this->provider->createNewResource();
+		$resource->setName(FederationManager::TALK_ROOM_RESOURCE)
+			->setShareTypes($this->talkProvider->getSupportedShareTypes())
+			->setProtocols([
+				'v4' => '/ocs/v2.php/apps/spreed/',
```
nickvergessen, line 60: @ArtificialOwl 

do you happen to know what data this is?
I assume the key is like the "what" and the value is the "where"?

so we should use something like
```suggestion
				'talk-v1' => '/ocs/v2.php/apps/spreed/api/',
```

because it's the first version of talk federation and the subsequent API is in `/ocs/v2.php/apps/spreed/api/*` ?

lib/Federation/ResourceTypeRegisterListener.php
```diff
@@ -0,0 +1,64 @@
...
+
+		if (!$this->talkConfig->isFederationEnabled()) {
+			return;
+		}
+
+		$event->registerResourceType(
+			FederationManager::TALK_ROOM_RESOURCE,
+			$this->talkProvider->getSupportedShareTypes(),
+			[
+				'v4' => '/ocs/v2.php/apps/spreed/',
```
nickvergessen, line 60: @ArtificialOwl 

do you happen to know what data this is?
I assume the key is like the "what" and the value is the "where"?

so we should use something like
```suggestion
				'talk-v1' => '/ocs/v2.php/apps/spreed/api/',
```

because it's the first version of talk federation and the subsequent API is in `/ocs/v2.php/apps/spreed/api/*` ?

ArtificialOwl: From my point of view both are totally acceptable; you can set which ever you want as long as you do not need to follow a specific requirement from the author of the talk app

if you want more details about the ocm api https://cs3org.github.io/OCM-API/

ArtificialOwl (approved):

PR #10697, opened 2023-10-13
[stable25] - fix message grouping for all locales

Antreesy:
Manual partial backport of #10687

To test:
- switch locale to arabic
- open chat
- post some messages
Output:
- messages should be grouped


nickvergessen:
25 is EOL

PR #10712, opened 2023-10-15
fix(messagesStore): clear deleted messages from replies

Antreesy:
### ☑️ Resolves

* Regression from #10296
* If several messages has one as a parent, and parent has been deleted, now we loop through all exisiting messages and clear a parent from each one

[Screencast from 15.10.2023 15:13:11.webm](https://github.com/nextcloud/spreed/assets/93392545/7b87ae59-13d2-44d5-8dde-1b9ade06f94e)

### 🏁 Checklist

### DIFF
--- a/src/store/messagesStore.js
+++ b/src/store/messagesStore.js
@@ -521,6 +521,16 @@ const actions = {
 					messageId: message.parent.id,
 				})
 			}
+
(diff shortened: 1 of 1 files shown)


SystemKeeper (approved):
Tested and works as expected

PR #10714, opened 2023-10-15
fix(useActiveSession): adjust behavior of inactive session 

Antreesy:
### ☑️ Resolves

* Regression from #10544 
* missing clearTimeout was added
* toast message has been kept for cases when tab is visible (otherwise too verbose to see it while switching tabs)

No visual changes

### 🏁 Checklist

### DIFF
--- a/src/composables/useActiveSession.js
+++ b/src/composables/useActiveSession.js
@@ -98,7 +98,10 @@ export function useActiveSession() {
 
 		try {
 			await setSessionState(token.value, SESSION.STATE.INACTIVE)
-			showInfo(t('spreed', 'Session has been marked as inactive'))
+			// Show toast message only when tab is visible on screen
+			if (windowIsVisible.value) {
+				showInfo(t('spreed', 'Session has been marked as inactive'))
+			}
 			console.info('Session has been marked as inactive')
 		} catch (error) {
 			console.error(error)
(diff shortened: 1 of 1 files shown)


src/composables/useActiveSession.js
```diff
@@ -98,7 +98,8 @@ export function useActiveSession() {
 
 		try {
 			await setSessionState(token.value, SESSION.STATE.INACTIVE)
-			showInfo(t('spreed', 'Session has been marked as inactive'))
+			// TODO to many verbose from this state change
```
nickvergessen, line 101:
```suggestion
			// TODO too verbose from this state change
```

nickvergessen: > 'showInfo' is defined but never used

Antreesy: It's still WIP, so will use this atm for testing. But I'm thinking to remove it together with comment

nickvergessen (approved):

ChristophWurst:
> No visual changes

lie :stuck_out_tongue: 

Bildschirmfoto vom 2023-11-13 15-56-11

should this show in production?

Antreesy:
> should this show in production?

As Joas said today, we'll keep it until final release, to be able to catch issues on beta. Yes, it's distracting 😬

nickvergessen:
> Context: I use two screens. If I use Talk on one and my IDE on the other I get to see the toast every now and then. It's a bit distracting.

It already helped finding a bug where it was toggled while being in a call sharing the own screen 😅👍

PR #10728, opened 2023-10-18
Feat/noid/multi file support

nickvergessen:
Split off from #10053  because it has more problems

- [ ] Test that multiple files are unshared when deleting the message
- [ ] These methods should return messages with multiple objects:

https://github.com/nextcloud/spreed/blob/d0d52ce2901c56ef9b03012c029a09a641c21bac/lib/Controller/ChatController.php#L924

https://github.com/nextcloud/spreed/blob/d0d52ce2901c56ef9b03012c029a09a641c21bac/lib/Controller/ChatController.php#L977


nickvergessen:
Decided to solve this in frontend/client side instead

PR #10729, opened 2023-10-18
feat(federation): Send room modifications via OCM notifications to re…

nickvergessen:
…motes

### ☑️ Resolves

* Ref #5723 

### 📺 Samples

- Create a conversation as userA
- Invite remote userB
- Invite remote userC
- Accept invite as userB
- Rename conversation as userA
- As userB check the name in the left sidebar
- As userC check the name in the invite notification

### 🚧 Tasks

- [x] Integration test for the name sample
- [x] Specifically handle invited but not accepted users (caused BFP locally on testing)
- [x] Specify list of supported props and increase it https://github.com/nextcloud/spreed/pull/10731
- [x] Check list of types for new and old values on the events

### 🏁 Checklist

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -59,6 +59,7 @@ use OCA\Talk\Events\RoomDeletedEvent;
 use OCA\Talk\Events\RoomModifiedEvent;
 use OCA\Talk\Events\SystemMessageSentEvent;
 use OCA\Talk\Federation\CloudFederationProviderTalk;
+use OCA\Talk\Federation\Listener as FederationListener;
 use OCA\Talk\Files\Listener as FilesListener;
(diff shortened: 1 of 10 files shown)


ArtificialOwl:
code looks good, tested on real setup and ✔️ 

Small note as I do not think this is related to this PR, but description is empty on remote when inviting a new remote user and require to be updated on main instance so it becomes available on remote instance.

ArtificialOwl (approved):

nickvergessen:
> Small note as I do not think this is related to this PR, but description is empty on remote when inviting a new remote user and require to be updated on main instance so it becomes available on remote instance.

Yeah, same for all other values. I guess name is actually the only relevant one, because that is shown on the invites. All others would be read from the remote instance once the UI is adjusted to support it.

PR #10730, opened 2023-10-18
feat(NewMessageUploadEditor) - caption to file share

### LINKED ISSUE #10414: Frontend "File caption"
- Overview https://github.com/nextcloud/spreed/issues/5354

---

- Draft at https://github.com/nextcloud/spreed/pull/8949

Antreesy:
### ☑️ Resolves

* Fix #10414

🏚️ Before | 🏡 After
---|---
image | Screenshot from 2023-10-18 17-22-50
image | Screenshot from 2023-10-18 17-29-01

### 🚧 Tasks

- [ ] To check
  - [ ] User mentions don't render for me in NewMessage (known bug)
  - [ ] Caught it once, the order of shared files was different, because they are now being loaded in parallel
- [ ] Follow-up
  - [ ] Support multiple file shares (requires a lot of rework in frontend/backend) - draft at #8949

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -457,6 +457,11 @@ export default {
 			type: Array,
 			default: () => { return [] },
 		},
+
+		referenceId: {
+			type: String,
+			default: '',
+		},
 	},
 
 	emits: ['toggle-combined-system-message'],
@@ -609,6 +614,7 @@ export default {
 						props: Object.assign({
 							token: this.token,
 							itemType,
+							referenceId: this.referenceId,
 						}, this.messageParameters[p]),
 					}
 				} else if (type === 'deck-card') {
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -47,11 +47,11 @@ the main body of the message as well as a quote.
 					class="message-body__main__text">
 					<Quote v-if="parent" v-bind="parent" />
 					<div class="single-emoji">
-						{{ message }}
+						{{ renderedMessage }}
 					</div>
 				</div>
 				<div v-else-if="showJoinCallButton" class="message-body__main__text call-started">
-					<NcRichText :text="message"
+					<NcRichText :text="renderedMessage"
 						:arguments="richParameters"
 						autolink
 						dir="auto"
@@ -59,7 +59,7 @@ the main body of the message as well as a quote.
 					<CallButton />
 				</div>
 				<div v-else-if="showResultsButton || isSystemMessage" class="message-body__main__text system-message">
-					<NcRichText :text="message"
+					<NcRichText :text="renderedMessage"
 						:arguments="richParameters"
 						autolink
 						dir="auto"
@@ -72,7 +72,7 @@ the main body of the message as well as a quote.
 						show-as-button />
 				</div>
 				<div v-else-if="isDeletedMessage" class="message-body__main__text deleted-message">
-					<NcRichText :text="message"
+					<NcRichText :text="renderedMessage"
 						:arguments="richParameters"
 						autolink
 						dir="auto"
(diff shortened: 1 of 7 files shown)


src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -502,6 +502,14 @@ export default {
 			return !this.isLastMessage && this.id === this.$store.getters.getVisualLastReadMessageId(this.token)
 		},
 
+		renderedMessage() {
+			if (this.messageParameters?.file && this.message !== '{file}') {
+				return `{file} ${this.message}`
```
nickvergessen, line 507: space? Or should be a new line or something?

Antreesy: new line `\n\n` would be better, as it puts files and caption in different paragraphs
Although it creates a gap under the image, and doesn't look good with one-liners, we could stylize them later

src/components/NewMessage/NewMessage.vue
```diff
@@ -466,6 +481,12 @@ export default {
 				}
 			}
 
+			if (this.upload) {
+				this.$emit('sent', this.text)
+				this.$store.dispatch('setCurrentMessageInput', { token: this.token, text: '' })
```
DorraJaouad, line 486: There should be a smooth scrolling to the bottom after upload is **completely** finished but don't know where to emit the event

src/store/fileUploadStore.js
```diff
@@ -282,31 +270,38 @@ const actions = {
...
+		// If caption is provided, attach to the last temporary message
+		const lastIndex = getters.getUploadsArray(uploadId).at(-1).at(0)
+		for (const [index, uploadedFile] of getters.getUploadsArray(uploadId)) {
 			// mark all files as uploading
 			commit('markFileAsUploading', { uploadId, index })
 			// Store the previously created temporary message
-			const temporaryMessage = state.uploads[uploadId].files[index].temporaryMessage
+			const temporaryMessage = {
+				...uploadedFile.temporaryMessage,
+				message: index === lastIndex ? caption : '{file}',
```
DorraJaouad, line 293: Maybe it's better to store the rendered message format here and use it in `Message` . After all, the message contains a file and usually there is {file} to indicate its presence in `Message.message`

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -568,7 +576,7 @@ export default {
 			let match
 			let emojiStrings = ''
 			let emojiCount = 0
-			const trimmedMessage = this.message.trim()
+			const trimmedMessage = this.renderedMessage.trim()
```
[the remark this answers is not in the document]
Antreesy: Do you mean `containsCodeBlocks()`? then code block will be passed as caption, so `this.message` should contains it. Although we should test that, nice find

Or `<MessageTranslateDialog />`? Then we shouldn't translate {file} part, so passing `this.message` is also fine

src/components/NewMessage/NewMessage.vue
Antreesy: Maybe we could use a preview from temp message behind loading spinner, so height will remain. But should check, how will that work for failed uploads

src/store/fileUploadStore.js
Antreesy: We got either 'caption' or '{file}' in the server response, so temp message should follow the same logic here, IMO

src/components/MessagesList/MessagesGroup/Message/Message.vue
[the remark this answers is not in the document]
DorraJaouad: yes, I meant `containsCodeBlocks()`

src/components/NewMessage/NewMessage.vue
Antreesy: I researched a bit: this has nothing to do with this PR, so I'd raise a follow-up here (or there is probably an issue for this already)

Pushed a new commit to use local images (from blob) and reduce server load and list flickering

PR #10731, opened 2023-10-19
Document and type PHP events

nickvergessen:
### ☑️ Resolves

* Base for https://github.com/nextcloud/spreed/pull/10729

### 🚧 Tasks

- [x] Migrate events to typed events
- [x] Type hint and document the parameters
- [x] Deprecate all other events
- [x] Add CI so we don't forget to document events again :-(

### 🏁 Checklist

### DIFF
--- a/lib/Config.php
+++ b/lib/Config.php
@@ -23,6 +23,7 @@ declare(strict_types=1);
 
 namespace OCA\Talk;
 
+use OCA\Talk\Events\BeforeTurnServersGetEvent;
 use OCA\Talk\Events\GetTurnServersEvent;
 use OCA\Talk\Model\Attendee;
 use OCA\Talk\Service\RecordingService;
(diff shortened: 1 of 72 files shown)


lib/Config.php
```diff
@@ -342,6 +343,8 @@ public function getTurnServers(bool $withEvent = true): array {
 		if ($withEvent) {
 			$event = new GetTurnServersEvent($servers);
 			$this->dispatcher->dispatchTyped($event);
+			$event = new BeforeTurnServersGetEvent($event->getServers());
+			$this->dispatcher->dispatchTyped($event);
```
come-nc, line 347: Are those not the same thing?

come-nc (approved):

lib/Config.php
nickvergessen: `GetTurnServersEvent` is not following the suggested naming schema from https://docs.nextcloud.com/server/latest/developer_manual/basics/events.html#naming-scheme
So replaced it with a "valid" one :)

come-nc: Would it work to have one extend the other so we only dispatch once or does listener not work that way?

nickvergessen: the endpoint is only called once per visitor and I will drop the event once the internal listeners have been migrated, so I don't think it's worth any shenanigans

PR #10756, opened 2023-10-23
fix(event): Strict type the event after clarification with event cons…

nickvergessen:
…umers

### 🏁 Checklist


fancycode (approved):
lgtm

PR #10757, opened 2023-10-23
fix(docs): Add readthedocs config to try fixing the build

nickvergessen:
Before | After
---|---
grafik | grafik

Ref 
https://docs.readthedocs.io/en/stable/config-file/v2.html#supported-settings
https://docs.readthedocs.io/en/stable/config-file/v2.html#mkdocs

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,10 @@
+version: 2
+
+build:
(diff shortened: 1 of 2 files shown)


SystemKeeper (approved):

nickvergessen:
Backporting because "stable" is build from latest tags

PR #10760, opened 2023-10-23
fix(LeftSidebar): Adjust wording to be less email like

nickvergessen:

Filter | 🏚️ Before | 🏡 After
---|---|---
Mentions | Bildschirmfoto vom 2023-10-23 14-37-38 | Bildschirmfoto vom 2023-10-23 14-35-22
Unread | Bildschirmfoto vom 2023-10-23 14-37-33 | Bildschirmfoto vom 2023-10-23 14-35-27
Empty | Bildschirmfoto vom 2023-10-23 14-37-27 | Bildschirmfoto vom 2023-10-23 14-35-15

### 🏁 Checklist

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -122,10 +122,11 @@
 					<!-- Conversations List -->
 					<template v-if="!isSearching">
 						<NcEmptyContent v-if="initialisedConversations && filteredConversationsList.length === 0"
-							:name="t('spreed', 'No matches found')"
-							:description="displayedMessage">
+							:name="emptyContentLabel"
+							:description="emptyContentDescription">
 							<template #icon>
-								<MessageBadge v-if="isFiltered" :size="64" />
+								<AtIcon v-if="isFiltered === 'mentions'" :size="64" />
+								<MessageBadge v-else-if="isFiltered === 'unread'" :size="64" />
 								<MessageOutline v-else :size="64" />
 							</template>
 							<template #action>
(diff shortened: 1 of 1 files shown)


src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -409,14 +410,19 @@ export default {
...
+		},
+
 		displayedMessage() {
 			switch (this.isFiltered) {
 			case 'mentions':
-				return t('spreed', 'You have no unread mentions in your inbox.')
+				return t('spreed', 'You have no unread mentions.')
 			case 'unread':
-				return t('spreed', 'You have no unread messages in your inbox.')
-			default:
```
Antreesy, line 418: It's probably expecting at least default: return ''

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -409,14 +410,19 @@ export default {
 			return this.$store.getters.getToken()
 		},
 
+		displayedHeadline() {
```
Antreesy, line 413:
```suggestion
		emptyContentLabel() {
```

Unlikely there would be a case for wording 'No matches found'. Even so, there is no need in `if {}` check inside the computed

DorraJaouad: Doch, 'No conversations found' is shown `this.conversationsList.length === 0` but otherwise ( for filters, `this.conversationsList.length !== 0` ) , the wording 'No matches found' is applied

nickvergessen: Seems I messed up the comparison, want to show "match" when a filter is applied. So now picked the same logic as for the subline

DorraJaouad (approved):

Antreesy (approved):

PR #10772, opened 2023-10-25
feat(ReactionToaster): Change to animated emojis 🎉

### LINKED ISSUE #10561: Animated reaction emojis
### How to use GitHub

---

The current reaction emojis could be a bit more celebratory. :D 

Animated emojis! This library could help: https://googlefonts.github.io/noto-emoji-animation/

DorraJaouad:
### ☑️ Resolves

* Fix #10561

It was agreed that we keep the simple emojis in popover.

https://github.com/nextcloud/spreed/assets/84044328/d699e336-57ad-4bc0-96fc-c04454f8feaf

### 🚧 Tasks

- [ ] Visual review
- [ ] Code review

### 🏁 Checklist


fancycode:
Nice! Any chance you could use webp images on browsers that support it to reduce the download sizes? For example the `Concerned.gif` here is 4.4 MBytes while the webp from https://googlefonts.github.io/noto-emoji-animation/ is only 967 KBytes (for the same 512x512 resolution).

nickvergessen:
The size is a bit shocking yeah. Maybe there are smaller versions available? We don't need 512px when we only show them in 26px?

That being said, I guess we should make them a bit bigger? cc @jancborchardt

src/components/CallView/shared/ReactionToaster.vue
```diff
@@ -28,7 +28,13 @@
 			:key="toast.seed"
 			class="toast"
 			:style="styled(toast.name, toast.seed)">
-			<span class="toast__reaction">
+			<img v-if="toast.reactionURL"
+				class="toast__reaction-img"
+				:src="toast.reactionURL"
+				:alt="toast.reaction"
+				width="32"
+				height="32">
```
Antreesy, line 36:
```suggestion
			<picture v-if="toast.reactionURL">
				<source :srcset="toast.reactionURL.replace('gif', 'webp')" type="image/webp">
				<img class="toast__reaction-img"
					:src="toast.reactionURL"
					:alt="toast.reaction"
					width="32"
					height="32">
			</picture>
```

That's a rough solution, but should do the trick, if we provide both .webp and .gif

Antreesy (commented):
Managed to reduce gif's size up to 80% by resizing (https://ezgif.com/resize for example)

And we need to think beforehand of fallbacks, how the emojis roster change may impact

src/components/CallView/shared/ReactionToaster.vue
DorraJaouad: I will reduce the gif dimensions to 64px which leads to 91% for Concerned emoji for example. This makes gif have less size than webp version. 

p.s: I tried to resize the webp as it is in 512x512 too, but couldn't find a good resizing tool.

nickvergessen (approved):
Let's get it in? The size can be easily changed afterwards?

jancborchardt:
> That being said, I guess we should make them a bit bigger?

@nickvergessen yeah, they are a bit small currently. We could double the size for sure?

PR #10737, opened 2023-10-20
fix(LeftSidebar): Disable blur event in order to click on a conversation

DorraJaouad:
### ☑️ Resolves

* Fix #10734

### 🖼️ Screenshots / Screencasts

 🏚️ Before  

https://github.com/nextcloud/spreed/assets/84044328/ab3105d8-1aca-492d-adc2-bbe606fe1031   

🏡 After 

   

https://github.com/nextcloud/spreed/assets/84044328/7273a193-46d8-45c7-a7e7-dbe9fc287b92

### 🚧 Tasks

- [ ] Code review
- [ ] Visual review
### 🏁 Checklist

- [x] 🌏 Tested with Chrome, Firefox and Safari or should not be risky to browser differences
- [x] 🖌️ Design was reviewed, approved or inspired by the design team
- [x] ⛑️ Tests are included or not possible
- [x] 📗 User documentation in https://github.com/nextcloud/documentation/tree/master/user_manual/talk has been updated or is not required

PR #10781, opened 2023-10-26
[Stable27] fix(LeftSidebar): Disable blur event in order to click on a conversation

DorraJaouad:
Backport #10737
Changed ref to "container"


DorraJaouad (commented):
Tested ✅

Antreesy (commented):
Shouldn't we use ref container? Scroller is only for virtual scroller, as I see

Antreesy (approved):

PR #10782, opened 2023-10-26
techdebt(context): Use injectFn instead of accessing IServerContainer

nickvergessen:
### 🏁 Checklist

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -94,11 +94,11 @@ use OCA\Talk\Share\Listener as ShareListener;
 use OCA\Talk\Share\RoomShareProvider;
 use OCA\Talk\Signaling\Listener as SignalingListener;
 use OCA\Talk\Status\Listener as StatusListener;
+use OCP\App\IAppManager;
 use OCP\AppFramework\App;
 use OCP\AppFramework\Bootstrap\IBootContext;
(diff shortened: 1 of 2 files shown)


lib/AppInfo/Application.php
```diff
@@ -224,56 +229,51 @@ public function boot(IBootContext $context): void {
...
-		$server->getNavigationManager()->add(static function () use ($server) {
-			/** @var Config $config */
-			$config = $server->get(Config::class);
-			$user = $server->getUserSession()->getUser();
+	public function registerNavigationLink(INavigationManager $navigationManager): void {
+		$navigationManager->add(static function () {
+			$config = Server::get(Config::class);
+			$userSession = Server::get(IUserSession::class);
+			$urlGenerator = Server::get(IURLGenerator::class);
+			$l = Server::get(IFactory::class)->get(self::APP_ID);
```
marcelklehr, line 252: Can't we inject these, too?

marcelklehr: Or maybe this is for performance, because we might not need them?

marcelklehr (approved):

lib/AppInfo/Application.php
nickvergessen: exactly. will avoid setting up the stuff on all occ and api calls

PR #10804, opened 2023-10-28
fix(signaling): Fix pullMessages return type

provokateurin:
## 🛠️ API Checklist

Signaling messages from the DB are strings.

### 🏁 Checklist

### DIFF
--- a/lib/Controller/SignalingController.php
+++ b/lib/Controller/SignalingController.php
@@ -353,7 +353,7 @@ class SignalingController extends OCSController {
 	 * Get signaling messages
 	 *
 	 * @param string $token Token of the room
-	 * @return DataResponse<Http::STATUS_OK|Http::STATUS_NOT_FOUND|Http::STATUS_CONFLICT, array{type: string, data: TalkSignalingSession[]}[], array{}>|DataResponse<Http::STATUS_BAD_REQUEST, string, array{}>
+	 * @return DataResponse<Http::STATUS_OK|Http::STATUS_NOT_FOUND|Http::STATUS_CONFLICT, list<array{type: string, data: TalkSignalingSession[]|string}>, array{}>|DataResponse<Http::STATUS_BAD_REQUEST, string, array{}>
 	 *
 	 * 200: Signaling messages returned
 	 * 400: Getting signaling messages is not possible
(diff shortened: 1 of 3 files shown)


lib/Controller/SignalingController.php
```diff
@@ -353,7 +353,7 @@ public function signaling(string $token, string $messages): DataResponse {
 	 * Get signaling messages
 	 *
 	 * @param string $token Token of the room
-	 * @return DataResponse<Http::STATUS_OK|Http::STATUS_NOT_FOUND|Http::STATUS_CONFLICT, array{type: string, data: TalkSignalingSession[]}[], array{}>|DataResponse<Http::STATUS_BAD_REQUEST, string, array{}>
+	 * @return DataResponse<Http::STATUS_OK|Http::STATUS_NOT_FOUND|Http::STATUS_CONFLICT, array{type: string, data: TalkSignalingSession[]|string}[], array{}>|DataResponse<Http::STATUS_BAD_REQUEST, string, array{}>
```
nickvergessen, line 356: I don't see data bring a string anywhere?

nickvergessen: Let's improve readability:

```suggestion
	 * @return DataResponse<Http::STATUS_OK|Http::STATUS_NOT_FOUND|Http::STATUS_CONFLICT, list<array{type: string, data: TalkSignalingSession[]|string}>, array{}>|DataResponse<Http::STATUS_BAD_REQUEST, string, array{}>
```

provokateurin: It is when it comes from https://github.com/nextcloud/spreed/blob/cd20ba221b582ba95bdbad7e70221eb71569ddc5/lib/Controller/SignalingController.php#L392. I think it should just be an array as well, but it is not because it comes directly from the DB.

lib/Signaling/Messages.php
```diff
@@ -112,7 +112,7 @@ public function addMessageForAllParticipants(Room $room, string $message): void
 	 * "new" message.
 	 *
 	 * @param string $sessionId
-	 * @return array
+	 * @return array{type: string, data: string}[]
```
nickvergessen, line 115:
```suggestion
	 * @return list<array{type: string, data: string}>
```

lib/Controller/SignalingController.php
nickvergessen: right.

lib/Signaling/Messages.php
provokateurin: This doesn't work because psalm will complain about more/less specific return types. `array<...>` works though

nickvergessen: Psalm was just not able to detect it, but it actually always was a list. Anyway, pushed a small commit that fixes it.

nickvergessen: feel free to squash all together now and rebuild openapi

nickvergessen (approved):

PR #10805, opened 2023-10-28
refactor(signaling): Rename sendMessages endpoint for clarification

provokateurin:
## 🛠️ API Checklist

The name of the method is not helpful for consumers of the OpenAPI spec. `sendMessages` correctly communicates what the endpoint does without looking at the description.

### 🏁 Checklist


nickvergessen:
Where is the method name used and where the first line of the description?

provokateurin:
The method name is used in the operationId field and the first line of the doc comment is in the summary field (see the diff in the spec). When generating client code the operationId is used for the method name. With the change clients need to call something like `client.signaling.sendMessages(...)` instead of `client.signaling.signaling(...)` which is more intuitive.

nickvergessen (approved):

PR #10818, opened 2023-10-31
fix(LeftSidebar): revert temp styles and fix conversation styling

### LINKED ISSUE #10443: Conversation item border is cut of when tabbing
### How to use GitHub

---

## Steps to reproduce
1. Click into the search and tab into the conversation list

### Expected behaviour
Border around the conversation

### Actual behaviour
is cut off:
grafik

## Talk app

**Talk app version:**stable27

DorraJaouad:
### ☑️ Resolves

* Fix #10443 again :') ( changes of its PR were reverted [here](https://github.com/nextcloud/spreed/commit/948a640bcd7d60078269b19f5fa0d36b6d61f9fb#diff-7a1abf98ca549db81a418814a97629c796bc85dc32b40f9cffb65ce3ad40d555) ). Another workaround is now used which is basically moving the outline inside the element.
* Fix the unexpected outline on favorite conversations ( border radius is changing) . As I checked, there was no explicit style for the border radius coming from the favorite icon node. Setting `overflow: hidden` solved it though 
* Revert some changes after merging https://github.com/nextcloud-libraries/nextcloud-vue/pull/4628

| 🏚️ Before | 🏡 After |
|------------|----------|
| image       | image       |
| image |image 

### 🚧 Tasks

- [ ] Code review
- [ ] Visual review

### 🏁 Checklist

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -1025,13 +1025,8 @@ export default {
 	padding: 0 !important;
 }
 
-// FIXME upstream https://github.com/nextcloud-libraries/nextcloud-vue/issues/4625
-:deep(.list-item__wrapper--active) {
-	.list-item:hover,
-	.list-item:focus,
-	.list-item:focus-visible,
-	.list-item:active {
-		background-color: var(--color-primary-element-hover);
-	}
+:deep(.list-item) {
+	overflow: hidden;
+	outline-offset: -2px;
 }
 </style>


src/components/LeftSidebar/ConversationsListVirtual.vue
```diff
@@ -151,6 +163,14 @@ export default {
...
 			}
 		},
+
+		adjustScrollerHeight() {
+			const scrollerContainer = this.$refs.scroller.$el.querySelector(
+				'.vue-recycle-scroller__item-wrapper'
+			)
+			const minHeight = this.conversations.length * this.CONVERSATION_ITEM_SIZE + 2
+			scrollerContainer.style.minHeight = minHeight + 'px'
+		},
```
ShGKme, line 173: We should not do this to fix the border...

1. It updates only on conversation list update, but it doesn't work on height change, e.g. window resize. It can be fixed by `window`'s `resize` event or resize observer, but it's not great for performance to do it just for a border.
2. It directly mutates DOM nodes that are rendered and managed by Vue with Virtual DOM. We should never mutate Vue's DOM. It has many ways to break.
3. It also searches for an element with a query selector, which is not a big issue but can be easily broken by a small layout change with a class rename.

ShGKme: Ferdinand just suggested an alternative solution for such issues in `NcCheckboxRadioSwitch`: using [`outline-offset`](https://developer.mozilla.org/en-US/docs/Web/CSS/outline-offset) `-2px`. Then the outline it inside the element, and there is no issue with overflow. Maybe try using it here? Or fix it on the nextcloud/vue side...

ShGKme: Or to try to add a 2px footer in the virtual scroller

DorraJaouad: The `outline-offset` is the optimal solution IMO but would rather change it in nextcloud/vue . If so, I will also add the `overflow: hidden` there too.

ShGKme (approved):
Works, but there is only a native web-browser's outline. Not sure if it is expected.

image

nickvergessen:
Also the last item is cut off again?
Bildschirmfoto vom 2023-11-02 17-12-20

DorraJaouad:
> Works, but there is only a native web-browser's outline. Not sure if it is expected.

image

Native outline seems to be on all elements not only list items. 

>Also the last item is cut off again?

That's the native outline being cut off (extra 1px I guess)

PR #10831, opened 2023-11-03
fix(chat): Fix mentions at the beginning of captions

### LINKED ISSUE #10830: Mentioning in file captions does not work when at the start of the caption
## Steps to reproduce
1. Share a picture and have a mention like "@systemkeeper" as the only input

### Expected behaviour
Mentioning should work

### Actual behaviour
https://github.com/nextcloud/spreed/assets/1580193/9473b772-4265-4818-98ff-497bbb4df352

I see the same when sharing from iOS, so I assumes it's an API issue? It works fine if you input something like "Hey @systemkeeper"

## Talk app

**Talk app version:** main on sermo

nickvergessen:
Set the current message as comment content, so that the message finds mentions which are now part if the message but were not on the original comment, e.g. mentions at the beginning of captions

### ☑️ Resolves

* Fix #10830 

### Before
```behat
    Then user "participant1" sees the following messages in room "public room" with 200 # FeatureContext::userSeesTheFollowingMessagesInRoom()
      | room        | actorType | actorId      | actorDisplayName         | message         | messageParameters |
      | public room | users     | participant1 | participant1-displayname | {mention-user1} | "IGNORE"          |
      Failed asserting that two arrays are equal.
      --- Expected
      +++ Actual
      @@ @@
               'actorType' => 'users'
               'actorId' => 'participant1'
               'actorDisplayName' => 'participant1-displayname'
      -        'message' => '{mention-user1}'
      +        'message' => '@participant2'
               'messageParameters' => '"IGNORE"'
           )
       )

```

### 🏁 Checklist

### DIFF
--- a/lib/Chat/Parser/UserMention.php
+++ b/lib/Chat/Parser/UserMention.php
@@ -94,7 +94,14 @@ class UserMention implements IEventListener {
 
 		$mentionTypeCount = [];
 
+		// Set the current message as comment content, so that the message finds
+		// mentions which are now part of the message, but were not on the original
+		// comment, e.g. mentions at the beginning of captions
+		$originalCommentMessage = $comment->getMessage();
+		$comment->setMessage($message, ChatManager::MAX_CHAT_LENGTH + 10000);
 		$mentions = $comment->getMentions();
+		$comment->setMessage($originalCommentMessage, ChatManager::MAX_CHAT_LENGTH);
+
 		// TODO This can be removed once getMentions() returns sorted results (Nextcloud 21+)
 		usort($mentions, static function (array $m1, array $m2) {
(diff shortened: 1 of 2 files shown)


lib/Chat/Parser/UserMention.php
```diff
@@ -94,7 +94,14 @@ protected function parseMessage(Message $chatMessage): void {
 
 		$mentionTypeCount = [];
 
+		// Set the current message as comment content, so that the message finds
+		// mentions which are now part if the message but were not on the original
```
SystemKeeper, line 98:
```suggestion
		// mentions which are now part of the message but were not on the original
```

SystemKeeper (approved):
Tested on web and on iOS

PR #10832, opened 2023-11-03
fix(reactions): Don't break mobile clients when the reactions list is…

nickvergessen:
… empty

### ☑️ Resolves

* Ref https://github.com/nextcloud/talk-ios/pull/1408

### 🏁 Checklist

### DIFF
--- a/tests/integration/features/bootstrap/FeatureContext.php
+++ b/tests/integration/features/bootstrap/FeatureContext.php
@@ -2391,12 +2391,17 @@ class FeatureContext implements Context, SnippetAcceptingContext {
 		$this->sendRequest('GET', '/apps/spreed/api/' . $apiVersion . '/chat/' . self::$identifierToToken[$identifier] . '/share/overview');
 		$this->assertStatusCode($this->response, $statusCode);
 
-		$overview = $this->getDataFromResponse($this->response);
-		$expected = $formData->getRowsHash();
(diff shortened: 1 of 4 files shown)


tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -3988,6 +4005,13 @@ public function getRoomListWithSpecificUserAgent(string $userAgent, int $statusC
 		$this->assertStatusCode($this->response, $statusCode);
 	}
 
+	protected function assertEmptyArrayIsNotAListButADictionary(?TableNode $formData, string $content) {
+		if (!$formData instanceof TableNode || empty($formData->getHash())) {
+			$data = json_decode($content);
+			Assert::assertIsNotArray($data->ocs->data, 'Response ocs.data should be an "object" to represent a JSON dictionary, not a list-array');
```
nickvergessen, line 4011: Before the patch in `lib` this throws:
```
    Then user "participant1" retrieve reactions "👍" of message "Message 1" in room "room" with 200  # FeatureContext::userRetrieveReactionsOfMessageInRoomWith()
      | actorType | actorId | actorDisplayName | reaction |
      Data should be an "object" to represent a JSON dictionary, not a list
      Failed asserting that Array &0 () is not of type "array".
```

SystemKeeper (approved):
Tested and works:
image

PR #10840, opened 2023-11-06
Fix(MessageButtonsBar): Hide copy message button for files

### LINKED ISSUE #10829: Files without caption allow copying the message
### How to use GitHub

---

## Steps to reproduce
1. Share a file to a conversation
2. See that there's still a "Copy formatted message" button in the context menu

### Expected behaviour
The button should not be available without a caption

### Actual behaviour
The placeholder `{file}` is copied to the clipboard.

## Talk app

**Talk app version:** main, tested on sermo

DorraJaouad:
### ☑️ Resolves

* Fix #10829

| 🏚️ Before | 🏡 After |
|------------|----------|
| image         | image     |

### 🚧 Tasks

- [ ] Code review

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
@@ -83,7 +83,8 @@
 						</template>
 						{{ t('spreed', 'Reply privately') }}
 					</NcActionButton>
-					<NcActionButton close-after-click
+					<NcActionButton v-if="!isFileShareOnly"
+						close-after-click
 						@click.stop="handleCopyMessageText">
 						<template #icon>
 							<ContentCopy :size="20" />
@@ -523,6 +524,10 @@ export default {
 
 		isFileShare() {
 			return Object.keys(Object(this.messageParameters)).some(key => key.startsWith('file'))
+		},
+
+		isFileShareOnly() {
+			return this.isFileShare && this.message === '{file}'
 		},
 
 		isCurrentGuest() {


ShGKme (approved):

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
```diff
@@ -583,32 +596,24 @@ export default {
...
+			const times = {
+				'Later today': laterTodayTime,
+				Tomorrow: tomorrowTime,
+				'This weekend': thisWeekendTime,
+				'Next week': nextWeekTime,
+			}
+
+			return Object.entries(times).map(([key, moment]) => {
+				return {
+					key: key.toLowerCase().replace(' ', '_'),
```
nickvergessen, line 608: this does not match the previous keys? Would it be important?

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
```diff
@@ -583,32 +596,24 @@ export default {
...
+					timestamp: this.getTimestamp(moment),
+					label: t('spreed', '{readableTime} – {timeLocale}', {
+						readableTime: key,
+						timeLocale: moment?.format('LT'),
+					}),
+					ariaLabel: t('spreed', `Set reminder for ${key.toLowerCase()}`),
```
nickvergessen, line 614: This does not work for translations. Please revert to old pattern with fixed strings.

See https://docs.nextcloud.com/server/latest/developer_manual/basics/front-end/l10n.html#important-notes for more information

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
DorraJaouad: It is only used as key in loop and not elsewhere

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
DorraJaouad: The aria label is translatable but I am not sure if the label is translatable. I think that's where the translation broke ? 
image

Anyway, for such cases, how to translate strings with variable part that should be translated ?

nickvergessen: > Anyway, for such cases, how to translate strings with variable part that should be translated ?

This is called string concatenation and does not work in all languages. Just don't use variabled strings. Go back to the predefined array that was here. (It's also unrelated to "Hide copy message button")

DorraJaouad: Yes, it was just an attempt to reduce the redundancy

nickvergessen (approved):

PR #10877, opened 2023-11-13
fix(DAVClient): wrap DAV client creation in a function

### LINKED ISSUE #10851: Cannot upload image/file while still being able to chat
### How to use GitHub

---

## Steps to reproduce
1. go into a room
2. chat
3. try to upload an image

### Expected behaviour
Upload of image should work.

### Actual behaviour
Nothing. In console I see 401: 
<?xml version="1.0" encoding="utf-8"?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>Sabre\DAV\Exception\NotAuthenticated</s:exception>
  <s:message>CSRF check not passed.</s:message>
</d:error>

- Happens on our c.nc.c 
- Already chatted with nickvergessen about it, no meaningful logs on server.
- session cookie looks 1:1 between heartbeat and 401/propfind

Antreesy:
### ☑️ Resolves

* Fix #10851 
* Solution initially was copied from https://github.com/nextcloud/viewer/commit/ab86d7f001327c6bc193a00f72005f3cd6ff0d17, but I've missed that request token can be updated and thus new client object is required

No visual changes

### 🏁 Checklist

### DIFF
--- a/src/services/DavClient.js
+++ b/src/services/DavClient.js
@@ -26,8 +26,8 @@ import { getRequestToken } from '@nextcloud/auth'
 import { generateRemoteUrl } from '@nextcloud/router'
 
 // init webdav client on default dav endpoint
-const client = createClient(generateRemoteUrl('dav'),
-	{ headers: { requesttoken: getRequestToken() || '' } },
-)
-
-export default client
+export const getDavClient = () => {
+	return createClient(generateRemoteUrl('dav'),
+		{ headers: { requesttoken: getRequestToken() || '' } },
+	)
+}
--- a/src/components/MediaSettings/VideoBackgroundEditor.vue
+++ b/src/components/MediaSettings/VideoBackgroundEditor.vue
@@ -92,7 +92,7 @@ import { imagePath, generateUrl } from '@nextcloud/router'
 
 import { VIRTUAL_BACKGROUND } from '../../constants.js'
 import BrowserStorage from '../../services/BrowserStorage.js'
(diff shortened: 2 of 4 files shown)


src/services/DavClient.js
```diff
@@ -26,8 +26,8 @@ import { getRequestToken } from '@nextcloud/auth'
 import { generateRemoteUrl } from '@nextcloud/router'
 
 // init webdav client on default dav endpoint
-const client = createClient(generateRemoteUrl('dav'),
-	{ headers: { requesttoken: getRequestToken() || '' } },
-)
-
-export default client
+export const getDavClient = () => {
```
nickvergessen, line 29: Why get when its creating? :P

Antreesy: To not mix with `import { createClient } from 'webdav'`?

Though it could be renamed, I think that using a `get` prefix is fine here. We're expecting from it to return an object, not to 'create and return'. It might be (and should be) further optimized to not re-create a client each time, unless `requestToken` has ben changed

ShGKme: Agree with Joas, that `get` prefix is usually used for getting something existed, not creating. Alternatives could be
- `buildDavClient`
- `makeDavClient`

But fine to me for consistency with other apps.

ShGKme (approved):
Looks fine.

P.S. there is no callback 😶‍🌫️

PR #10879, opened 2023-11-13
fix(shares): fix issues with file upload

### LINKED ISSUE #10845: Capability `'media-caption'` is not taken into account on frontend
### How to use GitHub

---

## Steps to reproduce
1. Connect to a Talk 17 server from Talk Desktop with Talk 18
2. Send an attachment and add a caption

### Expected behaviour

`media-caption` capability should be taken into account, and there should have been no caption input.

### Actual behaviour

There is a caption input, but the caption is not sent.

## Talk app

**Talk app version:** 17.1.3

**Talk Desktop version:** 0.14.0 with Talk 18.0.0-beta.1

### LINKED ISSUE #10873: File can't be added anymore after it was deleted from the preview dialog
### How to use GitHub

---

## Steps to reproduce
1. Try to send a file in a conversation
2. In the preview dialog, remove the file
3. Try to add the same file again

### Expected behaviour
Should work without an issue

### Actual behaviour
Works for the first time (sometimes) and fails afterwards:

https://github.com/nextcloud/spreed/assets/1580193/a2d924d9-d35b-4875-97af-37ba790f7d06

## Talk app

**Talk app version:** Beta 2 on c.nc.c

Antreesy:
### ☑️ Resolves

* Fix #10873
  * Reset `input[type="file"]` value when upload files to store
* Fix #10845 
  * Talk Desktop client should have an old interface against server <27 (only dismiss-send)
* Fix how `findUniquePath()` function handles filenames with suffixes

| 🏚️ Before | 🏡 After |
|------------|----------|
| image          | Screenshot from 2023-11-13 16-14-35 |

### 🏁 Checklist

### DIFF
--- a/src/store/fileUploadStore.js
+++ b/src/store/fileUploadStore.js
@@ -295,7 +295,7 @@ const actions = {
 			// Store the previously created temporary message
 			const temporaryMessage = {
 				...uploadedFile.temporaryMessage,
-				message: index === lastIndex ? caption : '{file}',
+				message: index === lastIndex && caption ? caption : '{file}',
 			}
 			// Add temporary messages (files) to the messages list
 			dispatch('addTemporaryMessage', temporaryMessage)
--- a/src/components/NewMessage/NewMessageUploadEditor.vue
+++ b/src/components/NewMessage/NewMessageUploadEditor.vue
@@ -67,7 +67,15 @@
 				<AudioPlayer :name="voiceMessageName"
 					:local-url="voiceMessageLocalURL" />
 			</template>
-			<NewMessage v-if="modalContainerId"
+			<div v-if="!supportMediaCaption" class="upload-editor__actions">
+				<NcButton type="tertiary" @click="handleDismiss">
+					{{ t('spreed', 'Dismiss') }}
+				</NcButton>
+				<NcButton type="primary" @click="handleUpload(null)">
+					{{ t('spreed', 'Send') }}
+				</NcButton>
+			</div>
+			<NewMessage v-else-if="modalContainerId"
 				ref="newMessage"
 				role="region"
 				upload
@@ -81,8 +89,9 @@
 </template>
 
 <script>
-
 import Plus from 'vue-material-design-icons/Plus.vue'
+
+import { getCapabilities } from '@nextcloud/capabilities'
 
 import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
 import NcModal from '@nextcloud/vue/dist/Components/NcModal.js'
@@ -91,6 +100,8 @@ import AudioPlayer from '../MessagesList/MessagesGroup/Message/MessagePart/Audio
 import FilePreview from '../MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue'
 import NewMessage from './NewMessage.vue'
 import TransitionWrapper from '../TransitionWrapper.vue'
+
+const supportMediaCaption = getCapabilities()?.spreed?.features?.includes('media-caption')
 
 export default {
 	name: 'NewMessageUploadEditor',
@@ -103,6 +114,12 @@ export default {
 		NcButton,
 		NewMessage,
 		TransitionWrapper,
+	},
+
+	setup() {
+		return {
+			supportMediaCaption,
+		}
 	},
 
 	data() {
@@ -166,7 +183,7 @@ export default {
 
 	watch: {
 		async showModal(show) {
-			if (show) {
+			if (show && this.supportMediaCaption) {
 				await this.getContainerId()
 				this.$nextTick(() => {
 					this.$refs.newMessage?.focusInput()
@@ -251,6 +268,12 @@ export default {
 			outline: 3px dashed var(--color-primary-element);
 			border-radius: var(--border-radius-large);
 		}
+	}
+	&__actions {
+		display: flex;
+		justify-content: flex-end;
+		gap: 4px;
+		padding: 12px 0;
 	}
 }
 
--- a/src/components/NewMessage/NewMessageUploadEditor.vue
+++ b/src/components/NewMessage/NewMessageUploadEditor.vue
@@ -218,6 +218,7 @@ export default {
 		async handleFileInput(event) {
 			const files = Object.values(event.target.files)
 			await this.$store.dispatch('initialiseUpload', { files, token: this.token, uploadId: this.currentUploadId })
+			this.$refs.fileUploadInput.value = null
 		},
 
 		handleRemoveFileFromSelection(id) {


nickvergessen (approved):

SystemKeeper (approved):
Tested and works for me

src/store/fileUploadStore.js
```diff
@@ -217,13 +217,19 @@ const actions = {
...
+		const initialisedUploads = getters.getInitialisedUploads(uploadId).map(upload => upload[1].file)
+
 		for (let i = 0; i < files.length; i++) {
 			const file = files[i]
-
+			if (initialisedUploads.some(upload => upload.name === file.name
+				&& upload.size === file.size
+				&& upload.lastModified === file.lastModified)) {
+				continue
+			}
```
ShGKme, line 232: Why do we need this check? Does it mean it is forbidden to send files with the same name + date + size?  (They still may have different content)

ShGKme: Scenario:
1. I created an `logo.svg` and attached it
2. Then I found that I needed to move one line and change it. There is no size change. 
3. For some reason `lastModified` was the same (suddenly pruned (we have this error on Nextcloud server), or the files were generated).

ShGKme: Isn't the linked issue fixed by clearing input[type=file] value?

Antreesy: I noticed, that after clearing it's possible to add same file twice. Then during the upload first file will be shared normally, but second returns 423 - locked and will be shown as failed. Don't know, which way is better, that or the omitting of duplicates

ShGKme: Issue is caused by this part:

https://github.com/nextcloud/spreed/blob/0727d1987da02129f66c84fa9a9eded45d398e90/src/store/fileUploadStore.js#L313-L322

It makes requests in parallel, so `findUniquePath` is called in parallel and returns the same path for 2 files that are to be uploaded together.

ShGKme: It also fails if 2 files just have the same name, regardless of file size or the last modified.

Antreesy: Will try to find a workaround, because parallel requests allow to save significant amount of time here, while uploading (around 1s for each file)

SystemKeeper: We could check if we try to upload files with the same name? If we do, it will not work anyway. So we need to alter the name before checking for uniqueness with the server?

Antreesy: Seems to be a problem here: when I'm trying to put the file under the same path, it got locked by Dav:
image

Simplest thing i could suggest is to check for duplicates before uploading:
- if there are only unique names, run PROPFIND in parallel
- if no - run consequently

ShGKme: What about check for filename in JS, and if files are equal, and change the name to `{name} (index)` manually? 

There is still a risk that a user has this `{name} (index)` file already, but in this case `findUniquePath` will rename it to `{name} (index) (2)`.

Alternatively, we can add a random prefix.

ShGKme: Or re-request the uniq name in catch for failed files.

Antreesy: > and change the name to `{name} (index)` manually?

findUniquePath checks if path already has a suffix, removes it and only after that perform a search (although it's broken, and for me file `test (10).jpg` resulted in `test (1 (2).jpg`, and `test (6).jpg` resulted in `test ( (2).jpg`.

So, adding an upload timestamp as suffix for duplicates should be enough

nickvergessen: > findUniquePath checks if path already has a suffix, removes it and only after that perform a search (although it's broken, and for me file test (10).jpg resulted in test (1 (2).jpg, and test (6).jpg resulted in test ( (2).jpg.

Then the method needs fixing.

> So, adding an upload timestamp as suffix for duplicates should be enough

Not a fan of working around

Antreesy:
Rebased onto master:
- media caption capability is untouched
- clearing of input value is untouched (backport?)
- fundUniquePath fix is added (backport?)

ShGKme:
I'm still getting errors on uploading files.

Uploading 2 files with the same name results in error `423 Locked`

Uploading `test (1).txt` results in uploading `test ( (2).txt`
image

Antreesy:
> Uploading 2 files with the same name results in error `423 Locked`

I've moved this to the new issue, and haven't fixed here.
 
> Uploading `test (1).txt` results in uploading `test ( (2).txt`

Thanks for noticing. I wonder, fo we need to remove the suffix from the original filename?
```
	// Check if the path ends with ` (n)` and remove suffix if present
	path = path.replace(/ \((\d+)\)$/, '')
```

Antreesy:
Rebased with updated solution:

- no longer omitting the suffix from original filename (so `text (10).jpg` will be uploaded as `text (10).jpg` or `text (10) (2).jpg`
  - thus, there wouldn't be error if uploading `text.jpg` and `text (10).jpg` together
- before uploading, check if there are any duplicate names in the list
  - if all names are unique, should be safe to upload in parallel
  - if not, fall back to upload 1 by 1 (taking longer, but ensures each upload will have unique path)
- before sharing, check if caption was provided
  - if not, the order is probably not important, so we can share files in parallel, and they may end up in chat mixed
  - if yes, fall back to sharing one by one (taking longer, but file with caption will be at the end of list) - should not be relevant then group share will be implemented

src/utils/fileUpload.js
```diff
@@ -45,27 +45,19 @@ const findUniquePath = async function(client, userRoot, path) {
...
-	// Check if the path ends with ` (n)`
-	const suffix = path.match(/ \((\d+)\)$/) ? path.match(/ \((\d+)\)$/) : ''
-	// Initialise a pathwithout suffix variable
-	let pathWithoutSuffix = path
-	if (suffix !== '') {
-		// remove the suffix if any
-		pathWithoutSuffix = path.substring(0, path.length - suffix.length)
-	}
+	// If there is a file extension, remove it from the path string
+	path = path.replace(fileExtension, '')
```
nickvergessen, line 52: This will replace the file extension anywhere in the path.

So:
`Suggestion for new welcome.txt/welcome.txt` will turn into `Suggestion for new welcome/welcome (2).txt` note the missing `.txt` on the folder name

src/utils/fileUpload.js
```diff
@@ -45,27 +45,19 @@ const findUniquePath = async function(client, userRoot, path) {
...
 	}
+
 	// Get the file extension (if any)
 	const fileExtension = getFileExtension(path)
-	// If there's a file extention, remove it from the path string
-	if (fileExtension !== '') {
-		path = path.substring(0, path.length - fileExtension.length)
-	}
-	// Check if the path ends with ` (n)`
-	const suffix = path.match(/ \((\d+)\)$/) ? path.match(/ \((\d+)\)$/) : ''
```
nickvergessen, line 55: no longer checking for this?

So if I download `file (2).txt` and modify it and then upload it again, it turns into `file (2) (2).txt` instead of `file (3).txt`

nickvergessen (changes requested):
Because it has 2 broken aspects now, can we split:
> Fix how findUniquePath() function handles filenames with suffixes

Into a second PR and add tests before adjusting it to ensure the expected behaviour happens?

src/utils/fileUpload.js
Antreesy: Yes, as I'm not sure this is the right way to do this.
- download + upload leads to creating a new file in the system, and we don't have mechanisms for checking, if `file (2).txt` in NC Files and `file (2).txt` from your PC is really the same but modified, or completely different.
- I can revert it, but this should be also considered when checking for duplicate names

src/utils/fileUpload.js
Antreesy: My bad

nickvergessen: the regex does, but your replace does not

src/utils/fileUpload.js
nickvergessen: Well we surely don't want to have `file (2) (2).txt`

So yeah please revert for now

src/utils/fileUpload.js
Antreesy: Will work on it in different PR

src/utils/fileUpload.js
Antreesy: Reverted for this PR

ShGKme (approved):

nickvergessen (approved):

PR #10914, opened 2023-11-16
fix(sip-dialout): Fix the successful dialout case when error is null

nickvergessen:
### ☑️ Resolves

* Fix `Attempt to read property \"message\" on null at …/apps/spreed/lib/Service/ParticipantService.php#1268`

## 🛠️ API Checklist


SystemKeeper (approved):
Makes sense

fancycode (approved):
Makes sense :+1: 

Btw, while you're at it, better check `code` instead of `message`.

PR #10916, opened 2023-11-16
fix(shares): handle shares with duplicate names and caption message

### LINKED ISSUE #10894: File upload fails with 423 for files with non-unique names loaded in parallel
Issue is caused by this part:

https://github.com/nextcloud/spreed/blob/0727d1987da02129f66c84fa9a9eded45d398e90/src/store/fileUploadStore.js#L313-L322

It makes requests in parallel, so `findUniquePath` is called in parallel and returns the same path for 2 files that are to be uploaded together.

_Originally posted by @ShGKme in https://github.com/nextcloud/spreed/pull/10879#discussion_r1391827529_

Antreesy:
### ☑️ Resolves

* Fix #10894 

- refactor `utils/fileUpload.js`
  - add test coverage for it (maybe missed some examples?)
- before uploading, check if there are any duplicate names in the list (`file.txt` and `file (2).txt` are counted)
  - if all names are unique, should be safe to upload in parallel
  - if not, extract duplicates and upload one them after uniques one by one (taking longer :snail:, but ensures each upload will have unique path)
- before sharing, check if caption was provided (should not be relevant then group share will be implemented)
  - if not, the order is probably not important, so we can share all files in parallel, and they may end up in chat mixed
  - if yes, share in parallel all except last one, then share last file with caption (taking longer :snail:, but file with caption will be at the end of list)
 
## 🖌️ UI Checklist

Should not be any visual changes

### 🚧 Tasks

- [x] redo extension check, it should affect only last substring
- [x] remove extra parenthesis from path suffix regex
- [x] add check for duplicates function to utils/fileUpload.js, consider suffixes
- [x] add test coverage
- [ ] To test: Network => Fetch/XHR => run different scenarios

### 🏁 Checklist

### DIFF
--- a/src/store/fileUploadStore.js
+++ b/src/store/fileUploadStore.js
@@ -33,7 +33,12 @@ import {
 	shareFile,
 } from '../services/filesSharingServices.js'
 import { setAttachmentFolder } from '../services/settingsService.js'
-import { findUniquePath, getFileExtension } from '../utils/fileUpload.js'
+import {
+	hasDuplicateUploadNames,
+	findUniquePath,
+	getFileExtension,
+	separateDuplicateUploads,
+} from '../utils/fileUpload.js'
 
 const state = {
 	attachmentFolder: loadState('spreed', 'attachment_folder', ''),
@@ -304,16 +309,13 @@ const actions = {
 			EventBus.$emit('scroll-chat-to-bottom', { force: true })
 		}
 
-		// Iterate again and perform the uploads
-		await Promise.allSettled(getters.getUploadsArray(uploadId).map(async ([index, uploadedFile]) => {
+		const performUpload = async ([index, uploadedFile]) => {
 			// currentFile to be uploaded
 			const currentFile = uploadedFile.file
 			// userRoot path
-			const userRoot = '/files/' + getters.getUserId()
 			const fileName = (currentFile.newName || currentFile.name)
 			// Candidate rest of the path
 			const path = getters.getAttachmentFolder() + '/' + fileName
-			const client = getDavClient()
 			// Get a unique relative path based on the previous path variable
 			const uniquePath = await findUniquePath(client, userRoot, path)
 			try {
@@ -349,10 +351,9 @@ const actions = {
 				commit('markFileAsFailedUpload', { uploadId, index })
 				dispatch('markTemporaryMessageAsFailed', { message: uploadedFile.temporaryMessage, reason })
 			}
-		}))
+		}
 
-		// Share the files, that have successfully been uploaded from the store, to the conversation
-		await Promise.all(getters.getShareableFiles(uploadId).map(async ([index, shareableFile]) => {
+		const performShare = async ([index, shareableFile]) => {
 			const path = shareableFile.sharePath
 			const temporaryMessage = shareableFile.temporaryMessage
 			const metadata = (caption && index === lastIndex)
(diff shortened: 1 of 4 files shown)


src/store/fileUploadStore.js
```diff
@@ -371,7 +367,35 @@ const actions = {
...
+			await Promise.all(uploads.map(upload => performUpload(upload)))
+		}
+
+		const shares = getters.getShareableFiles(uploadId)
+		// Check if caption message for share was provided
+		if (caption) {
+			// Share files one by one to keep their order and put caption on the last message
+			for (const share of shares) {
+				await performShare(share)
+			}
```
DorraJaouad, line 393: What if you async share all files except the last one in the array and then share that last file ?

Antreesy: Good point. Then we could extract last file for share flow and do the same for duplicates in upload flow

DorraJaouad: For duplicate upload, it won't be necessary. The duplication could be in any index ( or indices).

Antreesy: You're not wrong) But if we have 10 unique files and 2 with duplicates, we may upload 10 in parallel, and only 2 consequently.

DorraJaouad: Ah I see, you are also not wrong.

ShGKme (approved):

ShGKme:
Possible optimizations: for duplicates, do not duplicate search for the name. Currently, having `test (20).txt` on the server, uploading **3** `test.txt` makes **69 requests**:
  1. Request `test.txt`..`test (21).txt` to find that 1..20 are taken and 21 is free
  2. Request `test.txt`..`test (21).txt` to find that 1..20 and 21 are still taken. But it is already known from p.1, we should have started from 22.

Antreesy:
@ShGKme sorry, pushed to the same PR, please check again the last commit

src/store/fileUploadStore.js
```diff
@@ -179,6 +190,11 @@ const mutations = {
 		state.currentUploadId = currentUploadId
 	},
 
+	// Set prompt and suffix in case of successful propfind attempt
+	setKnownPathSuffix(state, { promptPath, suffix }) {
+		Vue.set(state.knownPaths, promptPath, suffix)
+	},
```
ShGKme, line 196: Path suffix seams to be some intermediate value to be used only in uploading loop. It doesn't seem to me to be a part of the global reactive application state. Do we need it in the store?

ShGKme: For example, it would break any further possibility to perform parallel uploads of different duplicates.

Antreesy: > it would break any further possibility to perform parallel uploads of different duplicates.

Then we could scope it within only this store action as a local const? Although it has been done to reuse value between uploads (for example, sharing one file twice with different people make 1..21 requests first time and only 22 second time), I see it as a rare use case

src/utils/fileUpload.js
```diff
@@ -38,38 +76,74 @@ const getFileExtension = function(path) {
...
+ * @param {number} knownSuffix The suffix to start looking from
+ * @return {object} The unique path and suffix
  */
-const findUniquePath = async function(client, userRoot, path) {
+const findUniquePath = async function(client, userRoot, path, knownSuffix) {
 	// Return the input path if it doesn't exist in the destination folder
-	if (await client.exists(userRoot + path) === false) {
-		return path
+	if (!knownSuffix && await client.exists(userRoot + path) === false) {
+		return { uniquePath: path, suffix: 1 }
```
DorraJaouad, line 85: For example, if test (6).txt path doesn't exist and there is no known Suffix yet, it will be stored as test (1).txt whereas its suffix should be 6. Either change the path or update suffix with the file's suffix.
As a default process, I am not sure if you want to force a suffix order ( first file is 1, second is 2 etc) regardless of their original suffixes or you want to store them with their original suffixes.  

[ Just a thought] If you are considering the first option, maybe there is a client function to get the list of all filenames that match fileName + ` (suffix = whatever in between)` + fileExtension and get its maximum ? In order to avoid looping.

Antreesy: No, there won't be a file with (1) at all. If we try to upload 4 files with name `file (4).txt`:
- `knownSuffix === undefined`, `client.exists === false` => return `{'file (4).txt', 1}`
- `knownSuffix === 1` => return `{'file (2).txt', 2}`
- `knownSuffix === 2` => return `{'file (3).txt', 3}`
- `knownSuffix === 3` => return `{'file (5).txt', 5}` (cause 4 existing already).

Although I may agree if it keeps following from the original suffix: 4,5,6,7

There is a method to search via webdav client: https://www.npmjs.com/package/webdav#search, but
- `client.search()` is more expensive than `client.exists()`
- it returns `501 (Not Implemented)`, so doesn't work for us =)

DorraJaouad: > .. is more expensive than client.exists()

Yeah, that's what I was skeptical about

>  Although I may agree if it keeps following from the original suffix: 4,5,6,7

I agree too as there is no use of a forced ordering.

DorraJaouad (approved):
Tested and works as expected 🥳

PR #10952, opened 2023-11-20
feat(LeftSidebar): Always show "create a new conversation" item

### LINKED ISSUE #10926: "Create new conversation" only shows up if there is no match for existing rooms
Not sure if it is a bug or a feature request. 

The "Create new conversation" option only shows up if there are no results in the conversations search:

image

But after the room "foo bar" from the example above was created there are legitimate reasons that the user want to create a room "foo" later. But now the "create new conversation" option is not shown:

image

I think the option to create a new room should always be present. The assumption that you only want to create a new room if there is no match is to strong in my opinion and doesn't match real use cases.

DorraJaouad:
### ☑️ Resolves

* Fix #10926

| 🏚️ Before | 🏡 After |
|------------|----------|
| image       | image   |

### 🏁 Checklist

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -169,7 +169,7 @@
 					<!-- Search results -->
 					<template v-else-if="isSearching">
 						<!-- Create a new conversation -->
-						<NcListItem v-if="searchResultsConversationList.length === 0 && canStartConversations"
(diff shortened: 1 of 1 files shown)


Antreesy (approved):
Neat. Backport to 27?

nickvergessen:
> Backport to 27?

No, it's a feature, fancy things only in 28+

PR #10962, opened 2023-11-21
fix(settings): Fix Hosted HPB settings UI

nickvergessen:
### ☑️ Resolves

* Type in the input fields

Regression from https://github.com/nextcloud/spreed/commit/ae7632078357b76aefabea971a960656f68f7ff8 it seems.

🏚️ Before | 🏡 After 
-----------|----------
Bildschirmfoto vom 2023-11-21 15-00-58 | Bildschirmfoto vom 2023-11-21 15-00-26

### 🏁 Checklist

### DIFF
--- a/src/components/AdminSettings/HostedSignalingServer.vue
+++ b/src/components/AdminSettings/HostedSignalingServer.vue
@@ -33,7 +33,7 @@
 		</p>
 
 		<template v-if="!trialAccount.status">
-			<NcTextField v-model="hostedHPBNextcloudUrl"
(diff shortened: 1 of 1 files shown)


nickvergessen:
Needs backporting or did the Vue lib break this recently? @Antreesy

Antreesy (approved):
I'd check this as well for stable26 and stable27.

v-model should work without such problems starting from Vue 3, which we haven't applied yet, and we moved a lot of components to `value.sync` syntax already

PR #10968, opened 2023-11-21
fix(AuthSidebar): Order popups

### LINKED ISSUE #10959: Video verification pops a dialog in the background
### How to use GitHub

---

## Steps to reproduce
1. Share a file as public link
2. Add a password
3. Enable video verification
4. As a guest in a private tab open the link:
Bildschirmfoto vom 2023-11-21 09-50-40
5. Click on "Request password" button:
Bildschirmfoto vom 2023-11-21 09-50-47
6. Close media settings and then see the guest name popup:
Bildschirmfoto vom 2023-11-21 09-50-52

Not sure which behaviour would be expected. Could be the current one or the guest name should be on top of it. But since the call is already ringing the video/mic options also seem important?

cc @DorraJaouad what do you think? Close or resort?

DorraJaouad:
### ☑️ Resolves

* Fix #10959

The condition in MediaSettings for displaying the display name input ( theoretically) should be valid only in `PublicShareAuthSidebar` when the user is a guest because in all other windows, the guest name pop up window is always shown first and cannot be bypassed. 

🏚️ Before (opacity to show both) 
 image 

🏡 After

https://github.com/nextcloud/spreed/assets/84044328/60ca1239-f475-4eaa-832d-0b4231edfaf3

### 🚧 Tasks

TODO
- [x] fix style ( it feels like it is missing padding)
- [x] Double-check if `guestName` ( the removed computed prop) is not needed in MediaSettings

### 🏁 Checklist

### DIFF
--- a/src/PublicShareAuthSidebar.vue
+++ b/src/PublicShareAuthSidebar.vue
@@ -40,7 +40,7 @@
 
 <script>
 import { getCurrentUser } from '@nextcloud/auth'
-import { emit } from '@nextcloud/event-bus'
+import { emit, subscribe, unsubscribe } from '@nextcloud/event-bus'
 import { loadState } from '@nextcloud/initial-state'
 
 import CallView from './components/CallView/CallView.vue'
@@ -54,6 +54,7 @@ import talkHashCheck from './mixins/talkHashCheck.js'
 import { EventBus } from './services/EventBus.js'
 import {
 	leaveConversationSync,
+	setGuestUserName
 } from './services/participantsService.js'
 import { signalingKill } from './utils/webrtc/index.js'
 
@@ -127,23 +128,28 @@ export default {
 	methods: {
 
 		async joinConversation() {
+			const guestStoredName = localStorage.getItem('nick')
+
 			if (getCurrentUser()) {
 				this.$store.dispatch('setCurrentUser', getCurrentUser())
+			} else if (guestStoredName) {
+				this.$store.dispatch('setDisplayName', guestStoredName)
+			} else {
+				subscribe('talk:guest-name:added', this.showGuestMediaSettings)
 			}
 
 			await this.$store.dispatch('joinConversation', { token: this.token })
+
+			// Add guest name to the store, only possible after joining the conversation
+			if (guestStoredName) {
+				await setGuestUserName(this.token, guestStoredName)
+			}
 
 			// Fetching the conversation needs to be done once the user has
 			// joined the conversation (otherwise only limited data would be
 			// received if the user was not a participant of the conversation
 			// yet).
 			await this.fetchCurrentConversation()
-
-			// Joining the call needs to be done once the participant identifier
-			// has been set, which is done once the conversation has been
-			// fetched. MediaSettings are called to set up audio and video devices
-			// and also to give a consent to recording, if set up
-			emit('talk:media-settings:show')
 
 			// FIXME The participant will not be updated with the server data
 			// when the conversation is got again (as "addParticipantOnce" is
@@ -159,6 +165,14 @@ export default {
 				// signaling server is used periodic polling has to be used
 				// instead.
 				this.fetchCurrentConversationIntervalId = window.setInterval(this.fetchCurrentConversation, 30000)
+			}
+
+			if (getCurrentUser() || guestStoredName) {
+				// Joining the call needs to be done once the participant identifier
+				// has been set, which is done once the conversation has been
+				// fetched. MediaSettings are called to set up audio and video devices
+				// and also to give a consent to recording, if set up
+				emit('talk:media-settings:show')
 			}
 		},
 
@@ -187,6 +201,14 @@ export default {
 				this.$store.dispatch('updateToken', '')
 			}
 		},
+
+		async showGuestMediaSettings() {
+			// Guest needs to add their display name right after joining conversation,
+			// before fetching and showing media settings. Then, this latter will be triggered
+			// by the guest name addition event.
+			emit('talk:media-settings:show')
+			unsubscribe('talk:guest-name:added', this.showGuestMediaSettings)
+		}
 	},
 }
 </script>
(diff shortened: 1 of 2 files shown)


src/components/ChatView.vue
```diff
@@ -23,7 +23,7 @@
 		@dragover.prevent="handleDragOver"
 		@dragleave.prevent="handleDragLeave"
 		@drop.prevent="handleDropFiles">
-		<GuestWelcomeWindow v-if="isGuestWithoutDisplayName" :token="token" />
+		<GuestWelcomeWindow v-if="showGuestNameWindow" :token="token" />
```
Antreesy, line 26:
```suggestion
		<GuestWelcomeWindow v-if="showGuestWelcomeWindow" :token="token" />
```

I'd name this computed (and prop) the same way we named a component, for consistensy

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -119,6 +119,14 @@
 				:token="token"
 				@update-background="handleUpdateVirtualBackground" />
 
+			<!-- Setting guest name -->
+			<div> {{ t('spreed','Please enter a display name') }} </div>
+			<NcTextField v-if="isUnknownGuest"
```
Antreesy, line 124:
```suggestion
			<template v-if="isUnknownGuest">
				<p> {{ t('spreed','Please enter a display name') }} </p>
				<NcTextField 
```

Antreesy (commented):
Looks OK, but haven't tested i yet. Ping me, if you want me to take it over

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -119,6 +119,17 @@
 				:token="token"
 				@update-background="handleUpdateVirtualBackground" />
 
+			<!-- Setting guest name -->
+			<template v-if="isUnknownGuest">
+				<label for="guestname-prompt" class="media-settings__guestname-label">
+					{{ t('spreed','Please enter a display name') }}
```
nickvergessen, line 125:
```suggestion
					{{ t('spreed','Enter your name') }}
```

Same wording as in the normal dialog?

marcoambrosini (changes requested):
I don't think we should add this in the media settings. Either we keep it in a separate dialog or we omit it imo

Antreesy:
Another option would be to disable automatic call, so user has to
- enter guest name to "unlock" screen
- click "Start call" manually => MediaSettings won't pop up

Antreesy:
Rebased onto main, split into two commits:
- First one is for omiting dialog for public share component
- Second one is for inserting input into the Media Settings (we could drop it, if come to an agreement)

DorraJaouad:
> Either we keep it in a separate dialog or we omit it imo

I think it's too many popups if we keep them both. If adding the name input is not possible in MediaSettings, we have the following options suggested
A: Remove guest name pop up and guest will always have translatable "Guest" in public share.
B: Disable automatic call.

Which one to go with ?

marcoambrosini:
Ideally we would use the same dialog but have two "pages", mre or less like we did in the first run wizard. If that's too tricky I think that 2 dialogs is the way to go.

nickvergessen:
Could we then somehow change the order so guest name is first modal and then the media settings?

Antreesy:
> so guest name is first modal and then the media settings?

Listen for a guest name to be set, then trigger this step:
https://github.com/nextcloud/spreed/blob/48df3840bb47edb8d227baa313727211aed3d343/src/PublicShareAuthSidebar.vue#L100-L104

DorraJaouad:
The conversation needs to be joined to show the guest pop up. Event listening should happen between joining and fetching the conversation. So trigger the rest of the `joinConversation()` after listening to the guest name event.

src/PublicShareAuthSidebar.vue
```diff
@@ -124,6 +124,14 @@ export default {
 		})
 	},
 
+	mounted() {
+		subscribe('guest-name-added', this.postJoinConversation)
```
nickvergessen, line 128: Every time a guest sets their name, they will now execute postJoin

DorraJaouad: The component is not mounted unless it is called which is with "request password" button click. It is not mounted with Talk app. 
or am I overlooking sthg ?

nickvergessen: yeah no that's all okay, but yeah I missed that guests don't have a rename option there.

Antreesy: event can be emitted only from two components:

image

Technically, there is only one way (and one chance) to submit a guest name from this sidebar. Although it's better to add some guard to it, to ensure we won't trigger it again

Also guest could submit his name earlier (in one of the public chats), so please check, if that name is available

src/PublicShareAuthSidebar.vue
```diff
@@ -187,6 +174,36 @@ export default {
...
+			// joined the conversation (otherwise only limited data would be
+			// received if the user was not a participant of the conversation
+			// yet).
+			await this.fetchCurrentConversation()
+
+			// Joining the call needs to be done once the participant identifier
+			// has been set, which is done once the conversation has been
+			// fetched. MediaSettings are called to set up audio and video devices
+			// and also to give a consent to recording, if set up
+			emit('talk:media-settings:show')
```
Antreesy, line 189: everything apart from this line should be safe to keep in `joinConversation`, as it's only about fetching the conversation, that shouldn't be blocking for unknown guests (to let them know a moderator has started a call, for example)

nickvergessen:
> /home/runner/work/spreed/spreed/src/PublicShareAuthSidebar.vue
Error:   59:34  error  Unable to resolve path to module '../services/participantsService.js'  import/no-unresolved
Error:   59:34  error  "../services/participantsService.js" is not found                      n/no-missing-import

Antreesy (approved):
Tested

PR #10971, opened 2023-11-22
fix(calls): Explicitly block calls in former one-to-one

### LINKED ISSUE #10595: Block calls on API level for `ONE_TO_ONE_FORMER`
> Would it make sense to check for `ONE_TO_ONE_FORMER` here as well?

_Originally posted by @SystemKeeper in https://github.com/nextcloud/spreed/pull/10541#discussion_r1337099752_

nickvergessen:
### ☑️ Resolves

* Fix #10595

### 🚧 Tasks

- Was already blocked indirectly because the room is read only, but now it's more clear and tested.

### 🏁 Checklist


SystemKeeper (approved):
Nice

Issue #10908, opened 2023-11-15
*Close sidebar* focused and visible after page load on mobile


ChristophWurst:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Steps to reproduce
1. Open Talk on mobile browser

### Expected behaviour

I see the conversation list or chat distraction free

### Actual behaviour

![image](https://github.com/nextcloud/spreed/assets/1374172/5be1b621-8f17-4d74-8b79-d901ad5107e1)

This is probably an accessibility *feature* from nextcloud/vue. Maybe the focus after page load should be set elsewhere?

## Talk app

**Talk app version:** (see apps admin page: `/index.php/settings/apps`)

**Custom Signaling server configured:** yes/no and version (see additional admin settings: `/index.php/index.php/settings/admin/talk#signaling_server`)

**Custom TURN server configured:** yes/no (see additional admin settings: `/index.php/settings/admin/talk#turn_server`)

**Custom STUN server configured:** yes/no (see additional admin settings: `/index.php/settings/admin/talk#stun_server`)

## Browser

**Microphone available:** yes/no

**Camera available:** yes/no

**Operating system:** Windows/Ubuntu/...

**Browser name:** Firefox/Chrome/...

**Browser version:** 85/96/...

### Browser log

<details>
```
Insert your browser log here, this could for example include:
a) The javascript console log
b) The network log
c) ...
```

</details>

## Server configuration

**Operating system**: Ubuntu/RedHat/...

**Web server:** Apache/Nginx

**Database:** MySQL/Maria/SQLite/PostgreSQL

**PHP version:** 8.0/8.1/8.2

**Nextcloud Version:** (see admin page)

**List of activated apps:**

<details>

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your server installation folder
```
</details>

**Nextcloud configuration:**

<details>

```
If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your Nextcloud installation folder
```
</details>

### Server log (data/nextcloud.log)
<details>

```
Insert your server log here
```
</details>

Antreesy:
[...]

PR #10973, opened 2023-11-22
fix(search): change conditions for showing trailing button

Antreesy:
### ☑️ Resolves

* Fix issue, described in #10908
> when focusing the search box, we're facing a loop for mobiles / desktops with small screen resolution:
> - Tabbing from input moves the focus to the trailing button
> - At the nextTick, trailing button will be deleted, as system assumes search field to be blurred
> - No element to focus => focus-trap falls back to the first available element => input field again

Workaround from Marco: setting a negative tabindex on the trailing button unless there is a non-empty value

🏚️ Before

https://github.com/nextcloud/spreed/assets/93392545/1223aad1-8e81-488b-a1aa-a9975978d933

🏡 After

https://github.com/nextcloud/spreed/assets/93392545/e874df3e-06d7-4f2e-9266-eeb3d7d312de

### 🏁 Checklist

### DIFF
--- a/src/components/LeftSidebar/SearchBox/SearchBox.vue
+++ b/src/components/LeftSidebar/SearchBox/SearchBox.vue
@@ -25,8 +25,8 @@
 		:label="placeholderText"
 		:show-trailing-button="isFocused"
 		class="search-box"
-		trailing-button-icon="close"
-		v-on="listeners"
+		@focus="handleFocus"
+		@blur="handleBlur"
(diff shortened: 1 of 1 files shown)


marcoambrosini (changes requested):
@Antreesy see https://github.com/nextcloud/spreed/issues/10908#issuecomment-1823571229

marcoambrosini (approved):

Issue #6983, opened 2022-03-09
Performance profiling integration tests


nickvergessen:
Here are some little steps I take to create a query log of integration tests in Talk.

## 🚧 Setup
1. To speed it up and decrease setup time apply the following patch. It will make sure users are not deleted and recreated with each test which heavily improves the speed of tests in case of rerunning and also removes setup queries from your profiling:
```diff
diff --git a/tests/integration/features/bootstrap/FeatureContext.php b/tests/integration/features/bootstrap/FeatureContext.php
index 126e15258..ae89d7015 100644
--- a/tests/integration/features/bootstrap/FeatureContext.php
+++ b/tests/integration/features/bootstrap/FeatureContext.php
@@ -200,15 +200,15 @@ class FeatureContext implements Context, SnippetAcceptingContext {
 	 * @AfterScenario
 	 */
 	public function tearDown() {
-		foreach ($this->createdUsers as $user) {
-			$this->deleteUser($user);
-		}
-		foreach ($this->createdGroups as $group) {
-			$this->deleteGroup($group);
-		}
-		foreach ($this->createdGuestAccountUsers as $user) {
-			$this->deleteGuestUser($user);
-		}
+//		foreach ($this->createdUsers as $user) {
+//			$this->deleteUser($user);
+//		}
+//		foreach ($this->createdGroups as $group) {
+//			$this->deleteGroup($group);
+//		}
+//		foreach ($this->createdGuestAccountUsers as $user) {
+//			$this->deleteGuestUser($user);
+//		}
 	}
 
 	/**
```

2. I normally have additional files in my skeleton folder to have images etc. available when testing. In case of creating e.g. 70 users copying them takes a lot of time, so I recommend to point the skeleton either to the default directory or remove any files but the welcome.txt
3. Make sure you have a Redis/Memcache setup and use it normally and also for file locking as well:
```php
'memcache.locking' => '\\OC\\Memcache\\Redis',
```
4. Run the integration test you want to profile once, so all users and groups are generated
5. Make sure you have no browser tab of the server open, no desktop client connected, xdebug disabled

## 🏎️ Profiling
1. Set the `query_log_file` in your config.php
```sh
'query_log_file' => '/var/log/nextcloud/query.log',
```
3. Run the integration test again
4. Save the result as baseline for your profiling (we have to clear some queries which are happening unrelated increasingly):
```sh
[...]

PR #11014, opened 2023-11-27
fix(calls): Remove a query in a loop when ending call for everyone

nickvergessen:
### ☑️ Resolves

* Did a performance check with #6983 and noticed that there where O(n) queries more

### 🏁 Checklist


SystemKeeper (approved):

nickvergessen:
Failing integration test reported in #11015

PR #11021, opened 2023-11-28
[main] Update nextcloud/ocp dependency

nextcloud-command:
Auto-generated update of [nextcloud/ocp](https://github.com/nextcloud-deps/ocp/) dependency


nickvergessen:
Fails until #11019

PR #11028, opened 2023-11-29
fix(UI): Rename "Note to self" to "Personal notes"

### LINKED ISSUE #10922: Rename "Notes to self"
Screenshot from 2023-11-17 04-00-49

This doesn't sound right 
Other ideas: "My notes" or "Quick notes" or "Talk notes" 
cc @nextcloud/designers

nickvergessen:
### ☑️ Resolves

* Fix #10922 

Has to be merged today, or never.


DorraJaouad (approved):
As it is better than the current wording, it is fine by me.

PR #11034, opened 2023-11-29
fix(CallView): pass method reference instead of result to debounce function

Antreesy:
### ☑️ Resolves

* Fix TypeError
image

### 🏁 Checklist


DorraJaouad (approved):

nickvergessen:
backport?

PR #11030, opened 2023-11-29
fix(UI): Fix missing headlines after vue library changed from title t…

nickvergessen:
…o name

### ☑️ Resolves

* Fix #11029

### 🖼️ Screenshots / Screencasts

Before | After
---|---
Bildschirmfoto vom 2023-11-29 10-58-07 | Bildschirmfoto vom 2023-11-29 10-58-44
Bildschirmfoto vom 2023-11-29 10-58-16 | Bildschirmfoto vom 2023-11-29 10-58-35
Bildschirmfoto vom 2023-11-29 11-04-29 | Bildschirmfoto vom 2023-11-29 11-06-13

### 🏁 Checklist

- [x] 🌏 Tested with Chrome, Firefox and Safari or should not be risky to browser differences
- [x] 🖌️ Design was reviewed, approved or inspired by the design team
- [x] ⛑️ Tests are included or not possible
- [x] 📗 User documentation in https://github.com/nextcloud/documentation/tree/master/user_manual/talk has been updated or is not required

PR #11035, opened 2023-11-29
I should "git add" before I "git commit"

nickvergessen:
Follow up to #11030


Antreesy (approved):
Wanted to suggest restore `title`, but buttons have text content anyway

PR #11038, opened 2023-11-29
fix(call): do not update participants list often when not in the call

Antreesy:
### ☑️ Resolves

* Hopefully reduce a server load during start of the call for large amount of participants:
  * Apply fast update only when user is in the call and has tab visible
  * Increase delay time for fetching from 3 to 5 seconds

To test locally:
- When in chat and someone joins conversation and call within 15s - should be only one request in Networks
- When in call and someone joins conversation and call within 5s - should be only one request in Networks

### 🏁 Checklist


src/mixins/getParticipants.js
```diff
@@ -104,7 +104,7 @@ const getParticipants = {
 			if (!this.fetchingParticipants) {
 				this.cancelableGetParticipants()
 			}
-		}, 3000),
+		}, 5000),
```
DorraJaouad, line 107: Isn't it too much ?

src/mixins/getParticipants.js
```diff
@@ -86,7 +86,7 @@ const getParticipants = {
 				return
 			}
 
-			if (this.$store.getters.windowIsVisible()) {
+			if (this.$store.getters.windowIsVisible() && this.isInCall) {
```
DorraJaouad, line 89: 15s needed to update participants list when chatting. This is too much IMO, people are usually faster than 15s to open conversation and post a message like "Good morning".

DorraJaouad (approved):

DorraJaouad (changes requested):
Again I didn't mean to approve :(

src/mixins/getParticipants.js
nickvergessen: But this adds quite some complexity for missing a person as being online that just entered a chat to drop a one-liner and went off again

DorraJaouad: This PR is intended to improve the high system load during the **start/end** of the call.
It prolonged the update interval from 3s to 5s in the full call. 
However, it also prolonged the update interval from 3s to 15s in the chat. 

I don't see why we should alter the update frequency for chat as there is no issue with that.
Also, the +2s increase in the update interval touches the full call and not only its ends. For the start, maybe we should check the start time of the call and apply this change within  the next 120s. 
The other two conditions won't be very necessary.

This issue of system load has a trade off with the overall page reactivity and update speed, I just think we should be cautious with it as it gives the impression that Talk is slow or the page is frozen.

nickvergessen: > However, it also prolonged the update interval from 3s to 15s in the chat.
>
> I don't see why we should alter the update frequency for chat as there is no issue with that.

Well as mentioned, people are "stuck" in the chat because they help hammering it away with participant updates they don't even remotely care about. They want to join the call, not take down the server with 3 sec update interval of a tab they actually don't care about.

But I agree we shouldn't fully kill the responsiveness.
So maybe instead of `&& this.isInCall` we do something like `&& (this.isInCall || conversationHasNoCall)` for the fast update?
That would keep the chatting part fixed and if there is a call and you are not part of it, it's also less important to be up to date by 3 seconds

Antreesy: Then we proceed with that behavior?

conversation.hasCall | true | false
-- | -- | --
isInCall === true | 3 | -
isInCall === false | 15 | 3

nickvergessen: Sounds like a good improvement for now

DorraJaouad (approved):
Looks better, thanks

Antreesy:
slow and fast debounces are parallel, looking for a way to fix it

src/mixins/getParticipants.js
```diff
@@ -75,7 +75,7 @@ const getParticipants = {
 		 */
 		onJoinedConversation() {
 			this.$nextTick(() => {
-				this.cancelableGetParticipants()
+				this.debounceUpdateParticipants()
```
nickvergessen, line 78: Should be okay to refresh on join once, so you see yourself, or is one-self hard coded already?

Antreesy: User (self) is in the list already. And another fetch would be after participants list refresh (as self has joined), so don't see a reason to do it once more. But with that now we could skip one request

PR #10970, opened 2023-11-22
fix(hosted-hpb): Correctly handle API response when the account expir…

nickvergessen:
…ed or was deleted

### ☑️ Resolves

* Fix #9656 

## 🛠️ API Checklist

- Fixed an issue when the HPB expired (401 can be returned together with an empty response)
- Improved the notifications (and fixed missing translations)

Before | After
---|---
grafik | Bildschirmfoto vom 2023-11-22 12-24-54

### 🏁 Checklist

- [x] ⛑️ Tests (unit and/or integration) are included or not possible
- [x] 📘 API documentation in `docs/` has been updated or is not required
- [x] 🔖 Capability is added or not needed

PR #11045, opened 2023-11-30
[stable27] fix(hosted-hpb): Correctly handle API response when the account expir…

nickvergessen:
Backport #10970


Antreesy (approved):
:seal:

PR #11050, opened 2023-11-30
fix(NcCheckboxRadioSwitch): revert flex-shrink for switch labels

Antreesy:
### ☑️ Resolves

* Patch a regression from  https://github.com/nextcloud-libraries/nextcloud-vue/pull/4760

| 🏚️ Before | 🏡 After |
|------------|----------|
| Screenshot from 2023-11-30 13-30-45          | Screenshot from 2023-11-30 13-32-27        |
image | image

### 🏁 Checklist

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -765,6 +765,11 @@ export default {
 			height: 100vh !important;
 		}
 	}
+
+	// TODO remove after https://github.com/nextcloud-libraries/nextcloud-vue/pull/4760 regression is fixed
+	:deep(.checkbox-radio-switch__text) {
+		flex-shrink: 1;
+	}
 }
 
 </style>


src/App.vue
```diff
@@ -765,6 +765,11 @@ export default {
 			height: 100vh !important;
 		}
 	}
+
+	// TODO remove after https://github.com/nextcloud-libraries/nextcloud-vue/pull/4760 regression is fixed
```
nickvergessen, line 769:
```suggestion
	// TODO remove after https://github.com/nextcloud-libraries/nextcloud-vue/issues/4896 regression is fixed
```

PS you could just send this PR upstream?

Antreesy:
Need a vue-lib release and bump

PR #11060, opened 2023-12-01
fix(SetContacts ): Remove unnecessary scroll

### LINKED ISSUE #10755: ContactSelectionBubble causes unnecessary scroll
### How to use GitHub

---

## Steps to reproduce
1. Start to create new conversation
2. Go to the screen with participants selection
3. Add some people

### Expected behaviour
image

### Actual behaviour
image

## Talk app

**Talk app version:** ?? (styles are untouched since March 2020)

DorraJaouad:
### ☑️ Resolves

* Fix #10755

Close Button has height 44px in `ContactSelectionBubble` whereas selected contacts div goes lower ( 40.8px), causing the extra scroll. 

| 🏚️ Before | 🏡 After |
|------------|----------|
| image  | image  |

### 🏁 Checklist

### DIFF
--- a/src/components/LeftSidebar/NewGroupConversation/SetContacts/ContactSelectionBubble/ContactSelectionBubble.vue
+++ b/src/components/LeftSidebar/NewGroupConversation/SetContacts/ContactSelectionBubble/ContactSelectionBubble.vue
@@ -102,6 +102,7 @@ $bubble-height: 24px;
 	background-color: var(--color-primary-element-light);
 	border-radius: $bubble-height;
 	height: $bubble-height;
+	overflow: hidden;
 	&__avatar {
 		margin-right: 4px;
 	}


Antreesy:
Still issue with 2 lines and more:
image

I remember the issue with 44px, yes =) How about we get rid of NcButton, and put a click event on a whole component?

DorraJaouad:
> Still issue with 2 lines and more:

I thought it is expected with more then one line 

> How about we get rid of NcButton, and put a click event on a whole component?
Hmm, wouldn't that be confusing for the user ?

Antreesy:
> I thought it is expected with more then one line

max-height is 90+px, so 2 lines should be shown without scroll

> Hmm, wouldn't that be confusing for the user ?

I thought we have the same behavior for NcSelect chips as an example, but they have it separately... maybe force a height of NcButton to 22px?

src/components/LeftSidebar/NewGroupConversation/SetContacts/SetContacts.vue
```diff
@@ -295,5 +296,9 @@ export default {
 	flex: 0 240px;
 	flex: 1 0 auto;
 	align-content: flex-start;
+
+	&__bubble {
+		overflow: hidden;
```
Antreesy, line 301: I think, we could put these styles directly to the component. It's local, so there is nothing to upstream =) 

https://github.com/nextcloud/spreed/blob/main/src/components/LeftSidebar/NewGroupConversation/SetContacts/ContactSelectionBubble/ContactSelectionBubble.vue#L98

PR #11063, opened 2023-12-01
fix(chat): expand system messages group if visual unread marker is set on it

### LINKED ISSUE #10758: follow-up(SystemMessageGroup) - improve behaviour
### How to use GitHub

---

Follow-up / Regression fixes for #9777:

- [ ] Showing in LeftSidebar in conversation
- [ ] Fixing the Unread messages marker (may be hidden in the list)
- [x] Disable/remove/rework sorting system messages in the group

Feel free to add items to the list

Antreesy:
### ☑️ Resolves

* Fix #10758

| 🏚️ Before (need to click) | 🏡 After (already open) |
|------------|----------|
Screenshot from 2023-12-01 11-34-31 | Screenshot from 2023-12-01 11-54-59

### 🏁 Checklist


DorraJaouad (commented):
It solved the issue :) but one thing ( not sure about its complexity) :

The `VisualLastReadMessageId` is updated after navigating to the conversation for the second time but it also takes a short period to update. This makes the group uncollapsed for the second time too because it caught the initial value of the last read message Id. 

I think it should not follow the last read message Id but what that id follows in the first place. From first check, `seen` in NewMessage can be a good candidate but I am not sure.

PR #11064, opened 2023-12-01
fix(participants): cancel scheduled request when requesting new one

### LINKED ISSUE #11048: follow-up: slow and fast debounced methods are independent in getParticipants
`this.debounceSlowUpdateParticipants` and `this.debounceFastUpdateParticipants` seem to have a different context, so if both are requested, both will be executed

https://www.npmjs.com/package/debounce library has `clear()` and `flush()` methods, although it either cancel both requests or none of them, depending on usage.

This potentially could save 1+ request to server, depending on conditions

_Originally posted by @Antreesy in https://github.com/nextcloud/spreed/issues/11038#issuecomment-1833406298_

Antreesy:
### ☑️ Resolves

* Fix #11048
  * Upon creation, we save debounced functions with their context. This alows to cancel currently scheduled executions
* Might help for both start/end call

image

### 🚧 Tasks

- [ ] Test different scenarios

Patch for testing:

<details>

```
Index: src/mixins/getParticipants.js
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/mixins/getParticipants.js b/src/mixins/getParticipants.js
--- a/src/mixins/getParticipants.js	(revision 7cd9746d80542c1cc1b244c062fb4185aaded33a)
+++ b/src/mixins/getParticipants.js	(date 1701433028038)
@@ -35,6 +35,7 @@
 			participantsInitialised: false,
 			fetchingParticipants: false,
 			pendingChanges: false,
+			i: 0,
 			debounceFastUpdateParticipants: null,
 			debounceSlowUpdateParticipants: null,
 		}
@@ -48,12 +49,12 @@
 	},
 
 	created() {
-		this.debounceFastUpdateParticipants = debounce(function() {
-			this.cancelableGetParticipants()
+		this.debounceFastUpdateParticipants = debounce(function(i) {
+			this.cancelableGetParticipants(i)
 		}, 3000)
 
-		this.debounceSlowUpdateParticipants = debounce(function() {
-			this.cancelableGetParticipants()
+		this.debounceSlowUpdateParticipants = debounce(function(i) {
+			this.cancelableGetParticipants(i)
 		}, 15000)
 	},
 
@@ -98,6 +99,8 @@
 				return
 			}
 
+			this.i++
+			console.log('update', this.i)
 			// Clear previously requested updates
 			this.debounceFastUpdateParticipants.clear()
 			this.debounceSlowUpdateParticipants.clear()
@@ -105,18 +108,20 @@
 			// this.conversation is provided by component, where mixin is used
 			if (this.$store.getters.windowIsVisible()
 				&& (this.isInCall || !this.conversation?.hasCall)) {
-				this.debounceFastUpdateParticipants()
+				console.log('schedule Fast', this.i)
+				this.debounceFastUpdateParticipants(this.i)
 			} else {
-				this.debounceSlowUpdateParticipants()
+				console.log('schedule Slow', this.i)
+				this.debounceSlowUpdateParticipants(this.i)
 			}
 			this.pendingChanges = false
 		},
 
-		async cancelableGetParticipants() {
+		async cancelableGetParticipants(i) {
 			if (this.fetchingParticipants || this.token === '' || this.isInLobby || !this.isModeratorOrUser) {
 				return
 			}
-
+			console.log('request', i)
 			this.fetchingParticipants = true
 
 			const response = await this.$store.dispatch('fetchParticipants', { token: this.token })
```

</details>

### 🏁 Checklist

### DIFF
--- a/src/mixins/getParticipants.js
+++ b/src/mixins/getParticipants.js
@@ -35,6 +35,8 @@ const getParticipants = {
 			participantsInitialised: false,
 			fetchingParticipants: false,
 			pendingChanges: false,
+			debounceFastUpdateParticipants: null,
+			debounceSlowUpdateParticipants: null,
 		}
 	},
 
@@ -43,6 +45,16 @@ const getParticipants = {
 			type: Boolean,
 			default: true,
 		},
+	},
+
+	created() {
+		this.debounceFastUpdateParticipants = debounce(function() {
+			this.cancelableGetParticipants()
+		}, 3000)
+
+		this.debounceSlowUpdateParticipants = debounce(function() {
+			this.cancelableGetParticipants()
+		}, 15000)
 	},
 
 	methods: {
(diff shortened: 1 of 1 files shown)


DorraJaouad (approved):
Looks good

nickvergessen:
> (but I can't see a case, where participant changing conditions so often)

Webinar with 400 people... someone will always leave, someone will always join

Antreesy:
> someone will always leave, someone will always join

participant updates 'scheduling' depends only on yourself (`this.$store.getters.windowIsVisible() && (this.isInCall || !this.conversation?.hasCall)`, so whatever 400 peoples are doing, you are always be either in slow or fast flow

DorraJaouad:
>  there is also an option flush()

Yeah, this will only speed up the excecution but without blocking the second request

> We will in worst case never update the list

aa right, I think it can be a non-edge case (user switching between tabs while in call multiple times). 

PS: it's never too late for using ✨ flags ✨

Issue #10844, opened 2023-11-07
Include image metadata


SystemKeeper:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

**Is your feature request related to a problem? Please describe.**
The metadata API will be public with Nextcloud 28: https://github.com/nextcloud/server/issues/40676. It would be nice if we could include some metadata for images, e.g. width, height and maybe blurhash.

In talk-ios we need to calculate the height of the image preview based on the aspec ratio of that image. These information are currently only available after we requested a preview, not when we receive the corresponding chat message. If these information were included, we could immediately calculate the correct dimensions and prevent some "jumping" and reloading of the tableView. The blurhash would allow us to immediately display a preview and replace it, once we retrieved the correct preview from the server.

**Describe alternatives you've considered**
Keep it as is.

PR #11093, opened 2023-12-03
Add metadata to file parameters

SystemKeeper:
### ☑️ Resolves

* Fix https://github.com/nextcloud/spreed/issues/10844

I tried to take a look at #10844. Leaving the tests and documentation aside, is this the correct way to solve it or would we need to take another different approach regarding performance?

### 🚧 Tasks

- [x] Correct the comment 
- [x] Needs nextcloud/OCP update after tonight

### 🏁 Checklist

### DIFF
--- a/lib/Chat/Parser/SystemMessage.php
+++ b/lib/Chat/Parser/SystemMessage.php
@@ -34,6 +34,7 @@ use OCA\Talk\Model\Message;
 use OCA\Talk\Participant;
 use OCA\Talk\Room;
 use OCA\Talk\Service\ParticipantService;
+use OCA\Talk\Share\Helper\FilesMetadataCache;
 use OCA\Talk\Share\RoomShareProvider;
 use OCP\Comments\IComment;
 use OCP\EventDispatcher\Event;
@@ -44,7 +45,6 @@ use OCP\Files\IRootFolder;
 use OCP\Files\Node;
 use OCP\Files\NotFoundException;
 use OCP\FilesMetadata\Exceptions\FilesMetadataNotFoundException;
-use OCP\FilesMetadata\IFilesMetadataManager;
 use OCP\IGroup;
 use OCP\IGroupManager;
 use OCP\IL10N;
@@ -87,7 +87,7 @@ class SystemMessage implements IEventListener {
 		protected IRootFolder $rootFolder,
 		protected ICloudIdManager $cloudIdManager,
 		protected IURLGenerator $url,
-		protected IFilesMetadataManager $metadataManager,
+		protected FilesMetadataCache $metadataCache,
 	) {
 	}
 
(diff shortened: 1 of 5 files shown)


lib/Chat/Parser/SystemMessage.php
```diff
@@ -732,6 +735,21 @@ protected function getFileFromShare(?Participant $participant, string $shareId):
 			'preview-available' => $this->previewManager->isAvailable($node) ? 'yes' : 'no',
 		];
 
+		// If a preview is available, check if we can get the dimension of the
```
nickvergessen, line 738: Of the what? Don't leave me hanging 🤔

SystemKeeper: > Of the what? Don't leave me hanging 🤔

Yea, saw it after pushing and added a task above to change it later 😅

nickvergessen:
> Guess we would need to add that to https://github.com/nextcloud/server/blob/master/lib/public/RichObjectStrings/Definitions.php#L316 as well?

I was about to comment that as well, but we have more keys already which are not there (e.g. contact-photo) which are very specific. In most cases the info will not be exposed when when it could be available via other APIs, so not sure it's the best idea, as it could make people assume ah, didn't get data, doesn't work, instead of trying the official API endpoint

SystemKeeper:
Okay, so guess ready to review then :)

tests/php/Chat/Parser/SystemMessageTest.php
```diff
@@ -611,7 +618,7 @@ public function testGetFileFromShareForGuest() {
 		$node = $this->createMock(Node::class);
 		$node->expects($this->once())
 			->method('getId')
-			->willReturn('54');
+			->willReturn(54);
```
nickvergessen, line 621: Just to spill over from chat, this looks fishy and needs to be confirmed.
API docs say int, but it's alerting that string was tested before. Maybe reality of docs and API result diverge

SystemKeeper: If we need to revert that, we need to cast to `int` for the metadata API, as this API requires a `int`.

nickvergessen: > its int it seems
> that's also why we cast it to string on the rich object

lib/Chat/Parser/SystemMessage.php
```diff
@@ -719,19 +722,39 @@ protected function getFileFromShare(?Participant $participant, string $shareId):
...
 			'link' => $url,
 			'etag' => $node->getEtag(),
 			'permissions' => $node->getPermissions(),
 			'mimetype' => $node->getMimeType(),
-			'preview-available' => $this->previewManager->isAvailable($node) ? 'yes' : 'no',
+			'preview-available' => $isPreviewAvailable ? 'yes' : 'no',
 		];
 
+		// If a preview is available, check if we can get the dimensions of the file from the metadata API
+		if ($isPreviewAvailable) {
```
nickvergessen, line 742: To avoid unnecessary DB queries (as per other comment), let's check for the iOS client atm (unless @mahibi  also needs the info for Android)

Antreesy:
Won't there be any good use of this for web? I'd see if we could reserve the space for images beforehand to avoid a chat jumping

PR #11107, opened 2023-12-04
fix(shares): allow to retry failed uploads

### LINKED ISSUE #10768: Handle upload errors of files and allow to retry
### How to use GitHub

---

## Steps to reproduce
1. Upload a file to a conversation

### Expected behaviour
The file should be uploaded, shared and visible to all users

### Actual behaviour
In case the upload was not successful (e.g. erroring out at 503, bad gateway) the temporary placeholder of the file will stay visible to the user, but no file was actually uploaded/shared. It's not clearly visible to the user that the upload failed (you can notice it when looking for the checkmark for example) and there's no option to retry the upload.

This looks like a issue on Talk, while https://github.com/nextcloud/spreed/issues/9681 looks like a issue in Server.

## Talk app

**Talk app version:** 17.1.1 on c.nc.c

Antreesy:
### ☑️ Resolves

* Fix #10768 
  * ATM allows to retry with files marked as 'failed-upload' (in case upload failed when putting on webdav step)

[retry-upload.webm](https://github.com/nextcloud/spreed/assets/93392545/cc9fe7a8-e775-49f5-8acb-6d4f7b6312db)

> [!TIP]
> Add `if (Math.random() < 0.5) throw new Error()` to `performUpload` function to fail some uploads

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -118,8 +118,8 @@ the main body of the message as well as a quote.
 						tabindex="0"
 						@mouseover="showReloadButton = true"
 						@focus="showReloadButton = true"
-						@mouseleave="showReloadButton = true"
-						@blur="showReloadButton = true">
+						@mouseleave="showReloadButton = false"
+						@blur="showReloadButton = false">
 						<NcButton v-if="sendingErrorCanRetry && showReloadButton"
 							:aria-label="sendingErrorIconTooltip"
 							@click="handleRetry">
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -695,6 +695,7 @@ export default {
 
 		sendingErrorCanRetry() {
 			return this.sendingFailure === 'timeout' || this.sendingFailure === 'other'
+				|| this.sendingFailure === 'failed-upload'
 		},
 
 		sendingErrorIconTooltip() {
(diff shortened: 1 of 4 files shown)


src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -810,8 +811,14 @@ export default {
 
 		handleRetry() {
 			if (this.sendingErrorCanRetry) {
-				EventBus.$emit('retry-message', this.id)
-				EventBus.$emit('focus-chat-input')
+				if (this.sendingFailure === 'failed-upload') {
+					const caption = this.renderedMessage !== this.message ? this.message : undefined
+					this.$store.dispatch('retryUploadFiles', { uploadId: this.messageObject.uploadId, caption })
+					this.$store.dispatch('removeTemporaryMessageFromStore', this.messageObject)
```
DorraJaouad, line 817: It would be better if it is in `retryUploadFiles` action and also remove all files from that uploadId not only the one that has the button clicked.

Antreesy: Yeah, I need to test it with multiple files upload

src/store/fileUploadStore.js
```diff
@@ -66,6 +67,16 @@ const getters = {
 			.filter(([_index, uploadedFile]) => uploadedFile.status === 'initialised')
 	},
 
+	getFailedUploads: (state, getters) => (uploadId) => {
+		return getters.getUploadsArray(uploadId)
+			.filter(([_index, uploadedFile]) => uploadedFile.status === 'failedUpload')
```
DorraJaouad, line 72: As I see, you are not using index in any of `getUploadsArray` calls so maybe it's better to change the output format 
```
 return Object.values(state.uploads[uploadId].files)
```

Antreesy: `getUploadsArray` is acting as a middleware to check for existing of files. `index` is used in dependent functions:

```js
for (const [index, uploadedFile] of getters.getInitialisedUploads(uploadId)) {
context.getters.getFailedUploads(uploadId).forEach(([index, file]) => {
const uploads = getters.getUploadingFiles(uploadId)
const shares = getters.getShareableFiles(uploadId)
```

DorraJaouad (approved):
Tested 🦅

DorraJaouad:
Rebased

PR #11119, opened 2023-12-05
fix(activeSession): minor fixes

### LINKED ISSUE #10887: Session is marked as inactive while participating in a call and even when sharing the own screen
### How to use GitHub

---

It's arguable that even in this case it could happen by accident that someone "leaves", but the percentage is too low. So we should consider participants of a call as always active.

Antreesy:
### ☑️ Resolves

* Fix #10887
* always reset timer on window focus change
* do not set inactive state when in call

### 🚧 Tasks

- [ ] To test:
  - [ ] 0. Set INACTIVE_TIME_MS to a lesser value (or wait 3 minutes)
  - [ ] 1. Join call
  - [ ] 1. Unfocus window (switch to another, or start screenshare)
  - [ ] 1. Wait for interval - session should be still active 
  - [ ] 2. Unfocus window
  - [ ] 2. Focus back (within interval, so session still active)
  - [ ] 2. Wait for interval - session should be still active 

### 🏁 Checklist

### DIFF
--- a/src/composables/useActiveSession.js
+++ b/src/composables/useActiveSession.js
@@ -24,6 +24,7 @@ import { showInfo } from '@nextcloud/dialogs'
 
 import { SESSION } from '../constants.js'
 import { setSessionState } from '../services/participantsService.js'
+import { useIsInCall } from './useIsInCall.js'
 import { useStore } from './useStore.js'
 
 const supportSessionState = getCapabilities()?.spreed?.features?.includes('session-state')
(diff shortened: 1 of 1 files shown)


SystemKeeper (approved):
Makes sense

DorraJaouad:
- Unfocus window but keep it shown ( in screen 1) and focus on another tab ( or screen 2) 
-  Move mouse or scroll chat but without focusing on window ( like reading chat from open conversation) 
==> session is marked as inactive

PR #11127, opened 2023-12-05
fix(session): track mouse movement activity

### LINKED ISSUE #11126: Move mouse or scroll chat but without focusing on window still marks session inactive
- Unfocus window but keep it shown ( in screen 1) and focus on another tab ( or screen 2) 
-  Move mouse or scroll chat but without focusing on window ( like reading chat from open conversation) 
==> session is marked as inactive

_Originally posted by @DorraJaouad in https://github.com/nextcloud/spreed/issues/11119#issuecomment-1840490039_

Antreesy:
### ☑️ Resolves

* Fix #11126 
  * `document` => `document.body` fixes behavior on Firefox with event registration
  * `mouseenter` now mark session as active, `mouseleave` restarts a timer
    * moving the mouse and scrolling within tab borders should be covered, as mouse entered the tab by that moment

### 🚧 Tasks

- [ ] To test:
  - [ ] Set INACTIVE_TIME_MS to a lesser value (or wait 3 minutes)
  - [ ] Unfocus window (switch to another, or start screenshare)
  - [ ] Enter mouse cursor onto tab and leave like this
  - [ ] Wait for interval - session should be still active 
  - [ ] Move cursor away from tab
  - [ ] Wait for interval - session should be marked as inactive  

### 🏁 Checklist

### DIFF
--- a/src/composables/useActiveSession.js
+++ b/src/composables/useActiveSession.js
@@ -120,23 +120,27 @@ export function useActiveSession() {
 		if (type === 'focus') {
 			setSessionAsActive()
 
-			document.removeEventListener('mouseenter', handleMouseMove)
-			document.removeEventListener('mouseleave', handleMouseMove)
+			document.body.removeEventListener('mouseenter', handleMouseEnter)
+			document.body.removeEventListener('mouseleave', handleMouseLeave)
 		} else if (type === 'blur') {
 			inactiveTimer.value = setTimeout(() => {
 				setSessionAsInactive()
 			}, INACTIVE_TIME_MS)
 
 			// Listen for mouse events to track activity on tab
-			document.addEventListener('mouseenter', handleMouseMove)
-			document.addEventListener('mouseleave', handleMouseMove)
+			document.body.addEventListener('mouseenter', handleMouseEnter)
+			document.body.addEventListener('mouseleave', handleMouseLeave)
 		}
 	}
 
-	const handleMouseMove = (event) => {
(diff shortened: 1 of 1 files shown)


Antreesy:
cc @nickvergessen for behavior change, @SystemKeeper for checking on Safari

SystemKeeper:
> @SystemKeeper for checking on Safari

Not working in safari it seems:

https://github.com/nextcloud/spreed/assets/1580193/a29e4cf4-a1e5-489d-a9ab-dc57eaefd54a

DorraJaouad:
maybe try `mouseover` and `mouseout` ?

SystemKeeper:
I tried to add:

```
		document.addEventListener('mousemove', (event) => {
			console.log('Move')
		})
```

But I only see "Move" output when the window is focussed in safari, as soon as another window is focussed, no 'Move' is logged.

Scrolling seems to work:

```
		document.getElementsByClassName('scroller').forEach((element) => {
			element.addEventListener('scroll', (event) => {
				console.log('Scroll')
			})
		})
```

but not sure if this also fires when we programmatically scroll.

Antreesy:
Tested as well, seems that MacOS tries to save resources here, on inactive tabs:
- For Safari events on blurred window are disabled
- For Chrome they work, but throttled (caught 1 mousemove event compared to hundreds on active)

DorraJaouad (approved):
As discussed, it is fine to skip Safari for now. 
Tested on Firefox and Chrome ✅

PR #11145, opened 2023-12-07
[stable27] Fix lost audio track on safari

SystemKeeper:
* Manual backport of https://github.com/nextcloud/spreed/pull/11006

It is correct that we use `require` here, since we are still using webpack and haven't moved to esm modules on stable27, right?

### 🏁 Checklist


Antreesy (approved):
Tested (Safari), seems to work (although I've caught a lost audio + frozen camera, then browser reconnected afterwards, so looks unrelated)

> It is correct that we use `require` here 

As I know, it's better to not mix `require` and `import` in one file, so fine to be here like this.

PR #11148, opened 2023-12-07
chore(eslint): enable import/order rule back

Antreesy:
### ☑️ Resolves

* Fix regression from #10622
* Probably minimatch pattern has been changed upstream, so `./..` imports are coming before `../..` now
* storeConfig import messagesStore as well, so there might be a loop while initializing (force disable two rules for Vuex store tests)

## 🖌️ UI Checklist

No visual changes
### 🚧 Tasks

- [x] Smoke test 

### 🏁 Checklist


DorraJaouad (approved):
Tested and checked ✅

PR #11161, opened 2023-12-08
fix(messages): apply ltr styles to system messages

Antreesy:
### ☑️ Resolves

* Fix regression from #10687
* Originates from #11159
* User mention is forced LTR in system messages as well

| 🏚️ Before | 🏡 After |
|------------|----------|

### 🏁 Checklist


DorraJaouad (approved):
Tested

PR #11166, opened 2023-12-08
Fix(Quote): Don't trim the quote when it is not image share with caption

DorraJaouad:
### ☑️ Resolves

- Regression from #11120 
-  Trimming is only applicable when it is an image with a caption

| 🏚️ Before | 🏡 After |
|------------|----------|
| image  | image  |

### 🚧 Tasks

- [ ] Code review

### 🏁 Checklist

### DIFF
--- a/src/components/Quote.vue
+++ b/src/components/Quote.vue
@@ -44,7 +44,6 @@ components.
 			<!-- file preview-->
 			<NcRichText v-if="isFileShareMessage"
 				text="{file}"
-				class="quote__file-preview"
 				:arguments="richParameters" />
 			<!-- text -->
 			<blockquote v-if="!isFileShareWithoutCaption"
@@ -189,6 +188,7 @@ export default {
 						props: Object.assign({
 							token: this.token,
 							smallPreview: true,
+							rowLayout: !this.messageParameters[p].mimetype.startsWith('image/'),
 						}, this.messageParameters[p]),
 					}
 				} else {
@@ -327,9 +327,6 @@ export default {
 		padding: 0 8px 0 8px;
 		position: relative;
 		margin: auto;
-	}
-	&__file-preview {
-		line-height: 0.5 !important;
 	}
 }
 
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
@@ -380,8 +380,8 @@ export default {
 			// Fallback for loading mimeicons (preview for audio files is not provided)
 			if (this.previewAvailable !== 'yes' || this.mimetype.startsWith('audio/')) {
 				return {
-					width: '128px',
-					height: '128px',
+					width: this.smallPreview ? '32px' : '128px',
+					height: this.smallPreview ? '32px' : '128px',
 				}
 			}
 


Antreesy:
Humbly ask to include in the PR:
```
diff --git a/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue b/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue	(revision 33006c3e25bdb97d6b619de4f85f78f973a6498b)
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue	(date 1702329641199)
@@ -380,8 +380,8 @@
 			// Fallback for loading mimeicons (preview for audio files is not provided)
 			if (this.previewAvailable !== 'yes' || this.mimetype.startsWith('audio/')) {
 				return {
-					width: '128px',
-					height: '128px',
+					width: this.smallPreview ? '32px' : '128px',
+					height: this.smallPreview ? '32px' : '128px',
 				}
 			}
``` 
Before | After
--|--
image | image

src/components/Quote.vue
```diff
@@ -328,7 +339,7 @@ export default {
 		position: relative;
 		margin: auto;
 	}
-	&__file-preview {
+	&__image-preview {
 		line-height: 0.5 !important;
```
Antreesy, line 343:
```suggestion
		line-height: 1 !important;
```
For me it looks OK like this for all variations (file and/or caption, image and/or caption). Otherwise everything feels either cropped or too high
So I'm suggesting to drop all other changes :smile:

DorraJaouad: it looks trimmed, no ? 
image

Antreesy: I see a bit, thanks. Maybe 1.2?

DorraJaouad: Still 
Before (line-height: 1.2) | Testing
--|--
image| image|


I was thinking about keeping it to default ( 1.5) but caption looks really far IMO 
Before (line-height: 1.5) | After (line-height: 1)
--|--
image|image

so drop it or keep it ?

Antreesy: Let's keep it, but using existed computed (isFileShare, isFileShareWithoutCaption)

Antreesy:
I know, I'm criticizing it again, but what if we use rowLayout prop, if there is no preview (and line-height 1 to all of them)?
```js
rowLayout: !this.messageParameters[p].mimetype.startsWith('image/'),
```
image

src/components/Quote.vue
```diff
@@ -184,11 +189,13 @@ export default {
 			Object.keys(this.messageParameters).forEach(function(p) {
 				const type = this.messageParameters[p].type
 				if (type === 'file') {
+					this.mimetype = this.messageParameters[p].mimetype
 					richParameters[p] = {
 						component: FilePreview,
 						props: Object.assign({
 							token: this.token,
 							smallPreview: true,
+							rowLayout: !this.mimetype.startsWith('image/'),
```
Antreesy, line 198:
```suggestion
							rowLayout: !this.messageParameters[p].mimetype.startsWith('image/'),
`

data and computed is not needed anymore, I suppose?

DorraJaouad: oh right, forgot to do the cleanup

Antreesy (approved):
Tested

PR #11181, opened 2023-12-10
Add backend url to connection error

SystemKeeper:
### ☑️ Resolves

For easier debugging I suggest we include the used backend url in the error message:
Bildschirmfoto 2023-12-10 um 20 28 13

### 🏁 Checklist


Antreesy (approved):
Looks correct by the code, can't see any case it may break something in signaling

fancycode (commented):
:+1: Makes sense.

PR #11197, opened 2023-12-11
fix(CS): Remove redundant PHPDoc tags

nickvergessen:
### 🏁 Checklist


bigcat88 (approved):
good idea, I need to do the same in my projects, less useless spam

PR #11198, opened 2023-12-11
fix(shares): use `nextcloud/upload` to track upload progress

### LINKED ISSUE #11153: webdav doesn't support upload progress since v5
Reference: https://github.com/perry-mitchell/webdav-client/blob/master/CHANGELOG.md#v500

Antreesy:
### ☑️ Resolves

* Fix #11153 
  * add `nextcloud/upload` library to perform file uploads to server
  * attach uploadManager data to track common upload progress
* Split large `uploadFiles` action into 4 smaller manageable actions
* update NcProgressBar styles (so it won't affects scroller height and chat jumping)

image

https://github.com/nextcloud/spreed/assets/93392545/21a3ad37-cbe9-463e-bcdb-8e93fc620903

### 🚧 TODO

- [x] track upload progress for each single file
- [x] test coverage

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
@@ -79,6 +79,7 @@ import PlayCircleOutline from 'vue-material-design-icons/PlayCircleOutline.vue'
 import { getCapabilities } from '@nextcloud/capabilities'
 import { encodePath } from '@nextcloud/paths'
 import { generateUrl, imagePath, generateRemoteUrl } from '@nextcloud/router'
+import { getUploader } from '@nextcloud/upload'
 
 import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
 import NcProgressBar from '@nextcloud/vue/dist/Components/NcProgressBar.js'
@@ -283,6 +284,7 @@ export default {
 		return {
 			isLoading: true,
 			failed: false,
+			uploadManager: null,
 		}
 	},
 	computed: {
@@ -511,14 +513,29 @@ export default {
 			return this.id.startsWith('temp') && this.index && this.uploadId
 		},
 
+		uploadFile() {
+			return this.$store.getters.getUploadFile(this.uploadId, this.index)
+		},
+
+		upload() {
+			return this.uploadManager?.queue.find(item => item._source.includes(this.uploadFile.sharePath))
+		},
+
 		uploadProgress() {
-			if (this.isTemporaryUpload) {
-				if (this.$store.getters.uploadProgress(this.uploadId, this.index)) {
-					return this.$store.getters.uploadProgress(this.uploadId, this.index)
-				}
+			switch (this.uploadFile?.status) {
+			case 'shared':
+			case 'sharing':
+			case 'successUpload':
+				return 100
+			case 'uploading':
+				return this.upload
+					? this.upload._uploaded / this.upload._size * 100
+					: 100 // file was removed from the upload queue, so considering done
+			case 'pendingUpload':
+			case 'initialised':
+			default:
+				return 0
 			}
-			// likely never reached
-			return 0
 		},
 
 		hasTemporaryImageUrl() {
@@ -534,7 +551,19 @@ export default {
 		},
 	},
 
+	watch: {
+		uploadProgress(value) {
+			if (value === 100) {
+				this.uploadManager = null
+			}
+		},
+	},
+
 	mounted() {
+		if (this.isTemporaryUpload && !this.isUploadEditor) {
+			this.uploadManager = getUploader()
+		}
+
 		const img = new Image()
 		img.onerror = () => {
 			this.isLoading = false
@@ -544,6 +573,10 @@ export default {
 			this.isLoading = false
 		}
 		img.src = this.previewUrl
+	},
+
+	beforeDestroy() {
+		this.uploadManager = null
 	},
 
 	methods: {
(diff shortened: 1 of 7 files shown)


Antreesy:
Possible style variation with circular progress (could be combined with loading spinner):

image

cc @nextcloud/designers

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
```diff
@@ -534,6 +538,31 @@ export default {
 		},
 	},
 
+	watch: {
+		isTemporaryUpload: {
+			immediate: true,
+			handler(value) {
+				this.uploadManager = value ? getUploader() : null
+			}
+		},
```
DorraJaouad, line 547: It is temporary only in the beginning, why don't you call getUploader() when on mounted ? 
and you can remove it once upload is finished

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
```diff
@@ -534,6 +538,31 @@ export default {
...
+			deep: true,
+			handler(value) {
+				if (!value) {
+					return
+				}
+				const progress = value._uploaded / value._size * 100
+				this.uploadProgress = Math.max(this.uploadProgress, progress)
+			}
+		}
+	},
```
DorraJaouad, line 564: This can be computed

```js
   uploadProgress() {
                                if (!this.upload) {
					return 100
				}
				return  this.upload._uploaded / this.upload._size * 100

		}
```

Unless ._uploaded is not progressing in one direction ( gets higher and then lower) or computed can't detect in depth object change here too despite we are using its attributes.

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
Antreesy: Maybe a better idea, also helps to not create the uploadManager for dialog window

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
Antreesy: It could fall back to 0 when upload got removed from the queue, but I will see what could be done

Antreesy: Here is the lifecycle of upload:
```
(File) (fileUpload.status) (upload._progress/upload._size)
FilePreview.vue:379 initialised undefined
FilePreview.vue:379 pendingUpload undefined
FilePreview.vue:379 uploading 0
FilePreview.vue:379 uploading 0.025629074670996455
FilePreview.vue:379 uploading 0.05125814934199291
...
FilePreview.vue:379 uploading 0.9739048374978653
FilePreview.vue:379 uploading 0.9995339121688618
FilePreview.vue:379 uploading 1
FilePreview.vue:379 uploading undefined // progressBar shows 0 at this point, and it takes a while
FilePreview.vue:379 successUpload undefined
FilePreview.vue:379 sharing undefined
FilePreview.vue:379 shared undefined
```

DorraJaouad: We can condition showing progressBar only if uploadProgress is not undefined. This means that this part is not included in the progress state :
`FilePreview.vue:379 uploading undefined // progressBar shows 0 at this point, and it takes a while`

Which makes sense because it is actually uploaded but fileUpload.status hasn't been updated yet ?

Antreesy:
Rebased, updated style of NcProgressBar to `circular`

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
```diff
@@ -50,6 +50,11 @@
 				:class="previewImageClass"
 				alt=""
 				:src="defaultIconUrl">
+			<NcProgressBar v-if="showUploadProgress"
+				class="file-preview__progress"
+				type="circular"
+				:size="56"
+				:value="uploadProgress" />
```
DorraJaouad, line 57: Maybe it is good to reduce the opacity of the file icon behind or add transparent white layer with the progressBar.

Antreesy: Thought about that 👍

Antreesy:
Another proposal for progress circle placement (where the spinner/checkmark is usually):
image

Whic option can we stick with, @marcoambrosini ?

marcoambrosini (commented):
@Antreesy would it be possible to put it to the left of the title? 

In the first option there's not enough contrast with the background and the shapes below make it a bit confusing, while in the second option I fear that the indicator might be too far from the title and icon.

Antreesy:
@marcoambrosini If you're not ready to come up quickly with a solution, we could split out the design changes into follow-up and merge only library + code for now

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
```diff
@@ -648,6 +657,13 @@ export default {
 		transition: outline 0.1s ease-in-out;
 	}
 
+	&__progress {
+		position: absolute;
+		top: 50%;
+		right: 0;
+		transform: translate(100%, -50%);
+	}
+
```
DorraJaouad, line 666: `display: inline-flex` didn't work out ?

Antreesy: Could cause some issues with image resizing

DorraJaouad (approved):
Tested 🦅

PR #11222, opened 2023-12-13
fix: use body element if not the fullscreen

### LINKED ISSUE #8709: Device preview prevents call from starting while editing document

### LINKED ISSUE #9626: File upload dialogue inaccessible when image preview is open during call

Antreesy:
### ☑️ Resolves

* Fix #8709
* Fix #9626
* Replaces #9792

Should be safe to use a `body` element while you're not in the fullscreen (as tested with ViewerOverlayCallView).
Also as long as we won't register a fullscreen state in sidebar (for Files page, for example), modal will end up in `body`, so above the Viewer embedded

>[!WARNING]
>Behaviour change: as NcModal component has more reasons against being re-mounted, then pros, hotkey 'F' to toggle Fulscreen mode is now disabled when any modal is open https://github.com/nextcloud/spreed/pull/11222/commits/368e43ab158577f367fc7b2a94543c443e7afa50

| 🖼️  Window frame | 🏞️ Fullscreen |
|------------|----------|
| Files Sidebar          | -        |
Screenshot from 2023-12-13 14-38-32 | Screenshot from 2023-12-13 14-38-55
Screenshot from 2023-12-13 14-38-22 | Screenshot from 2023-12-13 14-38-46
| Talk main app | - |
Screenshot from 2023-12-13 14-36-49 | Screenshot from 2023-12-13 14-37-17
Screenshot from 2023-12-13 14-36-26 | Screenshot from 2023-12-13 14-36-03
Screenshot from 2023-12-13 14-34-50 | Screenshot from 2023-12-13 14-35-02

### 🚧 Tasks

- [ ] Check possible drawbacks?

### 🏁 Checklist

### DIFF
--- a/src/components/TopBar/TopBarMenu.vue
+++ b/src/components/TopBar/TopBarMenu.vue
@@ -176,6 +176,7 @@ import VideoIcon from 'vue-material-design-icons/Video.vue'
 import GridView from 'vue-material-design-icons/ViewGrid.vue'
 
 import { getCapabilities } from '@nextcloud/capabilities'
+import { showWarning } from '@nextcloud/dialogs'
 import { emit } from '@nextcloud/event-bus'
 
 import NcActionButton from '@nextcloud/vue/dist/Components/NcActionButton.js'
@@ -392,6 +393,16 @@ export default {
 		},
 
 		toggleFullscreen() {
+			// Don't toggle fullscreen if there is an open modal
+			// FIXME won't be needed without Fulscreen API
+			if (Array.from(document.body.childNodes).filter(child => {
+				return child.nodeName === 'DIV' && child.classList.contains('modal-mask')
+					&& window.getComputedStyle(child).display !== 'none'
+			}).length !== 0) {
+				showWarning(t('spreed', 'You need to close a dialog to toggle full screen'))
+				return
+			}
+
 			if (this.isFullscreen) {
 				this.disableFullscreen()
 				this.$store.dispatch('setIsFullscreen', false)
--- a/src/App.vue
+++ b/src/App.vue
@@ -734,6 +734,11 @@ export default {
 .mx-datepicker-main.mx-datepicker-popup {
 	z-index: 10001 !important;
 }
+
+/* FIXME: remove after https://github.com/nextcloud-libraries/nextcloud-vue/pull/4959 is released */
+body .modal-wrapper * {
+	box-sizing: border-box;
+}
 </style>
 
 <style lang="scss" scoped>
--- a/src/components/CallView/shared/ViewerOverlayCallView.vue
+++ b/src/components/CallView/shared/ViewerOverlayCallView.vue
@@ -224,7 +224,7 @@ export default {
 		},
 
 		portalSelector() {
-			return this.$store.getters.isFullscreen() ? '#content-vue' : 'body'
+			return this.$store.getters.getMainContainerSelector()
 		},
 
 		hasLocalScreen() {
--- a/src/components/ChatView.vue
+++ b/src/components/ChatView.vue
@@ -61,6 +61,7 @@
 
 		<!-- Input field -->
 		<NewMessage v-if="containerId"
+			:key="containerId"
 			role="region"
 			:token="token"
 			:container="containerId"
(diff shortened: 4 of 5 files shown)


ShGKme (approved):

DorraJaouad (commented):
2 things to check : 
* Event propagation where modals are involved ( like @event or v-on )
* CSS dependency ( using :deep to affect a modal style), as the modal may be out of scope.

Otherwise, should be fine.

Antreesy:
> * CSS dependence ( using :deep to affect a modal style), as the modal may be out of scope.

Elements inside `#content-vue` have inherited `box-sizing: border-box`.
Others attached to body - depends in each case

Antreesy:
Fixed for NewMessage in ChatView:

It provides container to children components (EmojiPicker, autocomplete, NcActions) when mounted, and won't update anymore.
So we need to listen to fullscreen change and re-mount component accordingly

Antreesy:
Found a drawback: if you go to Fullscreen with open modal/popover, it will stay attached to the `body` element, thus won't be visible. Freshly opened modals works fine, as they get an actual container value before being mounted
Possible solution: attach `:key="container"` to each modal :eyes:

DorraJaouad:
> Possible solution: attach :key="container" to each modal 

As it is causing re-render only when container changes, sounds like a good solution.

Antreesy:
NcActions works fine, as they're re-rendered and re-mounted each time the container is changed
Looks like this function is triggered each time: https://vuejs.org/api/options-rendering.html#render

https://github.com/nextcloud-libraries/nextcloud-vue/blob/ebbe5c689959fc6470ab79352fd8126d3b98aec2/src/components/NcActions/NcActions.vue#L1215-L1221

src/components/TopBar/TopBarMenu.vue
```diff
@@ -391,6 +391,14 @@ export default {
 		},
 
 		toggleFullscreen() {
+			// Don't toggle fullscreen if there is an open modal
+			if (Array.from(document.body.childNodes).filter(child => {
+				return child.nodeName === 'DIV' && child.classList.contains('modal-mask')
+					&& window.getComputedStyle(child).display !== 'none'
```
DorraJaouad, line 397: Looks really non-Vue way because we should tend to query the virtual DOM instead but as modals are now rendered outside Vue hierarchy, this is the shortest way to check their presence.

DorraJaouad: Not sure, but maybe also add a hint why "F" shortkey is not toggling full screen ?

DorraJaouad (approved):
Nothing is blocking, can't find another drawback.

src/components/TopBar/TopBarMenu.vue
Antreesy: Like "showWarning" toast?

DorraJaouad: Exactly

DorraJaouad (approved):
Approved again

DorraJaouad:
Fixup squash and rebased

DorraJaouad:
@nickvergessen backport to stable27 and stable26 ?

nickvergessen:
My take would be no 🤔

nickvergessen:
@DorraJaouad ^ can you take care of the backport while Maksim is out?

PR #11224, opened 2023-12-14
fix(shares): reply to message with attachments

### LINKED ISSUE #11239: Mentions in captions don't trigger notifications
### How to use GitHub

---

## Steps to reproduce
1. Write a message with a mention
2. Drop a file so the message turns into a caption
3. Hit send

### Expected behaviour
- 🔔 Mentioned users have a notification

### Actual behaviour
- 🔕 Mentioned users have no notification
- Also applies to replies after https://github.com/nextcloud/spreed/pull/11224

Antreesy:
### ☑️ Resolves

* allows to reply to other messages with file share
* Fix #11239 

|  🏡 Flow |
|------------|
Screenshot from 2023-12-14 09-41-10
image

### 🏁 Checklist

---

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/lib/Chat/SystemMessage/Listener.php
+++ b/lib/Chat/SystemMessage/Listener.php
@@ -455,12 +455,14 @@ class Listener implements IEventListener {
 			$referenceId = (string) $referenceId;
 		}
 
+		$parent = $parameters['metaData']['replyTo'] ?? null;
+
 		return $this->chatManager->addSystemMessage(
 			$room, $actorType, $actorId,
 			json_encode(['message' => $message, 'parameters' => $parameters]),
 			$this->timeFactory->getDateTime(), $message === 'file_shared',
 			$referenceId,
-			null,
+			$parent,
 			$shouldSkipLastMessageUpdate,
 			$silent,
 		);
--- a/lib/Chat/SystemMessage/Listener.php
+++ b/lib/Chat/SystemMessage/Listener.php
@@ -27,6 +27,7 @@ namespace OCA\Talk\Chat\SystemMessage;
 
 use DateInterval;
 use OCA\Talk\Chat\ChatManager;
+use OCA\Talk\Chat\MessageParser;
 use OCA\Talk\Events\AAttendeeRemovedEvent;
 use OCA\Talk\Events\AParticipantModifiedEvent;
 use OCA\Talk\Events\ARoomModifiedEvent;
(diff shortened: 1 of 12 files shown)


DorraJaouad:
don't forget the OoO
image

DorraJaouad:
> Make this a feature request?

I think so, replying with a picture with caption. For now, when you intend to reply and insert caption with pic, quote is removed and it is send as a normal caption message.

The quote should be removed from NewMessage after sending 
image

nickvergessen (changes requested):
Actually we should just make this work, not hide it

lib/Chat/SystemMessage/Listener.php
```diff
@@ -455,12 +455,14 @@ protected function sendSystemMessage(Room $room, string $message, array $paramet
 			$referenceId = (string) $referenceId;
 		}
 
+		$parent = $parameters['metaData']['replyTo'] ?? null;
```
nickvergessen, line 458: This needs validation, but I will make a commit

nickvergessen (approved):

nickvergessen:
Backporting as we need to also support replying in #11239

Antreesy (commented):
Tested Notifications API patch :green_circle: 

image
image

src/components/NewMessage/NewMessage.vue
```diff
@@ -515,8 +515,9 @@ export default {
 			}
 
 			if (this.upload) {
-				// Clear input content from store
+				// Clear input content from store and remove Quote component
 				this.chatExtrasStore.setChatInput({ token: this.token, text: '' })
+				this.chatExtrasStore.removeParentIdToReply(this.token)
```
DorraJaouad, line 520: Add a check in `removeParentIdToReply()` for the existence of the token item in the store.

DorraJaouad (approved):
Tested 
- Notification 👍🏻 
- Reply with caption 👍🏻 
- Code review 👍🏻

src/components/NewMessage/NewMessage.vue
Antreesy: Vue has an inner guard for it, so nothing will happen if you call `Vue.delete(this.parentToReply, token)` and `this.parentToReply[token]` is undefined

SystemKeeper:
This breaks rendering on talk-ios as a parent message within a file message was not possible until now. Please remember to test/align on these kind of changes :).
CC: @mahibi @rapterjet2004 To check on android as well

nickvergessen:
> Please remember to test/align on these kind of changes

Sorry about that. That's right and I think we brought it up briefly in the chat/call. But should have raised a ticket right away.

PR #11149, opened 2023-12-07
fix(shares): reserve space for file preview while it's loading

Antreesy:
### ☑️ Resolves

* Follow-up for web for #11093
* Fix #11041 

:warning: Metadata requires "Photos" app to be enabled

### 🖼️ Screenshots / Screencasts

| 🏚️ Before | 🏡 After |
|------------|----------|
| Uploaded non-media files preview (text, audio) | - |
| image          | image        |
| Jump after preview loaded | - |
| image | image |
| Double hover on message | - |
| image | image |

#### Using preview metadata

🏚️ Before 

https://github.com/nextcloud/spreed/assets/93392545/0306c018-77f9-41d8-b772-fc4ca127d9dd

🏡 After
Metadata is used as soon as component is mounted. Chat is still jumping sometimes, but feels better anyway

https://github.com/nextcloud/spreed/assets/93392545/0ca46203-ee1f-4d91-88a2-9f698905c5ed

### 🚧 Tasks

- [ ] Test different files / cases

### 🏁 Checklist

- [ ] 🌏 Tested with Chrome, Firefox and Safari or should not be risky to browser differences
- [ ] ⛑️ Tests are included or not possible

PR #11225, opened 2023-12-14
fix(preview): preview size and semantic inside `a` element

skjnldsv:
Fix the play icon alignment and make the whole link semantically valid (no `div` allowed inside `a`)

| Before | After |
|:---------:|:------:|
|||

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
@@ -23,7 +23,8 @@
 -->
 
 <template>
-	<div v-bind="filePreview"
+	<component :is="filePreviewElement"
+		v-bind="filePreviewBinding"
 		:tabindex="wrapperTabIndex"
 		class="file-preview"
 		:class="{ 'file-preview--viewer-available': isViewerAvailable,
@@ -32,7 +33,7 @@
 			'file-preview--row-layout': rowLayout }"
 		@click.exact="handleClick"
 		@keydown.enter="handleClick">
-		<div v-if="!isLoading || fallbackLocalUrl"
+		<span v-if="!isLoading || fallbackLocalUrl"
 			class="image-container"
 			:class="{'playable': isPlayable}">
 			<span v-if="isPlayable && !smallPreview" class="play-video-button">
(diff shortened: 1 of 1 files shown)


skjnldsv:
Ah, this might overlap with https://github.com/nextcloud/spreed/pull/11149
I was working on stable28. it still seems relevant imho, but you can close if you want :)

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
```diff
@@ -648,7 +648,8 @@ export default {
 
 	.image-container {
 		position: relative;
-		display: inline-flex;
+		// Adapt size to children img
+		display: inline-block;
```
Antreesy, line 652: Seems to work fine in both ways

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
```diff
@@ -23,7 +23,7 @@
 -->
 
 <template>
-	<div v-bind="filePreview"
+	<element v-bind="filePreview"
```
Antreesy, line 26: Shouldn't we use `<component />` here?

https://vuejs.org/api/built-in-special-elements#component

Antreesy (approved):

src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
skjnldsv: Sure, this is not really an official standard or recommendation. `component` or `element` is fine for me :shrug:

Antreesy: 👍 Then I'll change a tag name and squash, if you don't mind

nickvergessen: So time to get rid of the `fragment` library we have?

Antreesy: > to get rid

No, it has different usage (although I'm not the fan of it). `<Fragment>` allows to put several elements as a component root:

```vue
<template>
  <Fragment>
    <div class="1"/>
    <div class="2"/> 
  <Fragment>
<template>
```

`<component>` can render different components upon conditions:
```vue
  <component :is="isSystemMessagesGroup ? MessagesSystemGroup : MessagesGroup" />
```

DorraJaouad (approved):
Tested

PR #11261, opened 2023-12-18
Enh(Message): Component splitting and store migration (Reactions only)

DorraJaouad:
### 🚧 Tasks

- [x] Migration of reactions to a new component
- [x] Store migration
- [x] Fix reactions update with receiving messages
- [x] Fix reactions count and reacting participants list synchronization
- [x] Patch reactions
- [x] Tests 😿 

Note: `'reaction'` system message is not coming instantly like other reaction related system messages and sometimes is not even sent and reactions gets refetched upon the regular chat fetch.  That's why you can notice that reactions come all at once as a batch when more than one is sent.

* Bug reported: counter is not in sync with reacting participants list 
* Follow-up: patch reactions

### 🏁 Checklist


Antreesy (commented):
Here goes a first part, with moving the component:

src/store/messagesStore.js
```diff
@@ -1351,12 +1354,15 @@ const actions = {
 			})
 			// The response return an array with the reaction details for this message
 			const response = await addReactionToMessage(token, messageId, selectedEmoji)
-			// We replace the reaction details in the reactions store and wipe the old
-			// values
-			context.dispatch('updateReactions', {
+
+			const actorId = context.getters.getActorId()
+			const actorObject = response.data.ocs.data[selectedEmoji].find(actor => actor.actorId === actorId)
```
Antreesy, line 1359: We receive the same list of actual detailed reactions as a response from:
- `getReactionsDetails`
- `addReactionToMessage`
- `removeReactionFromMessage`

Why don't use `reactionsStore.updateReactions` with a full list instead of patching only one? optimization could be done in the store itself
Plus, someone else could react at the same time, and we will know it from the response (but you're omitting it now)

src/store/messagesStore.js
```diff
@@ -521,10 +522,12 @@ const actions = {
 			const parentInStore = context.getters.message(message.token, message.parent.id)
 			if (Object.keys(parentInStore).length !== 0) {
 				context.commit('addMessage', message.parent)
-				context.dispatch('resetReactions', {
-					token: message.token,
-					messageId: message.parent.id,
-				})
+				const reactionsStore = useReactionsStore()
+				if (message.systemMessage.startsWith('reaction')) {
+					reactionsStore.fetchReactions(message.token, message.parent.id)
```
Antreesy, line 527: If we get 50 reactions on a single message in a row, that means we fetch details 50 times.
- In history conversation - messages come at once, and we fetch them instantly as many times, as there were reactions added/deleted

image

=> Solution here could be comparing new and old reactions object (`JSON.stringify`)

- In currently ongoing conversation - each participant online will receive all details each time somebody reacts. That means - <participants online> * <reactions sent> requests to server, while not everyone is going to need that info (for example, I see amount of reactions, but am not interested, who sent it)

=> Not sure about the right way here, but maybe I'd stick to the old flow with some changes - keep simplified reactions and load them on hover only (but hover on emojis buttons, not the message body)

DorraJaouad: > Solution here could be comparing new and old reactions object (JSON.stringify)

This would be a solution to minimize store update but not fetch requests.

> Not sure about the right way here, but maybe I'd stick to the old flow with some changes - keep simplified reactions and load them on hover only (but hover on emojis buttons, not the message body)

This means make `Reactions` dependent on `Message` again through reactions prop. I thought we are trying to avoid that. Also, this brings back the issue of count is showing a different number then the number of participants in the details list.

Antreesy: > This would be a solution to minimize store update but not fetch requests.

Not if we compare `message.reactions` and `parentInStore.reactions`

> Also, this brings back the issue of count is showing a different number then the number of participants in the details list.

Issue was that within 30 interval between messages polling, messageStore object has old value (`reactions: { 👍: 1, 😃: 1 })`) and by hover we fetching actual details (`reactions: { 👍: [admin], 😃: [bob, alice] })`). Wouldn't it be easier to compare them on hover, and patch message object with `😃: [bob, alice].length`?

DorraJaouad: Is there a possibility to get reaction message sent as reaction_revoked message and not with polling messages process ? This would solve it from the root.

src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -978,7 +827,8 @@ export default {
 	position: relative;
 
 	&:hover .normal-message-body,
-	&:hover .combined-system {
+	&:hover .combined-system,
+	&--hovered  {
```
Antreesy, line 831:
```suggestion
	&--hovered .normal-message-body {
```
Like this probably, for a style consistency

src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.vue
```diff
@@ -0,0 +1,230 @@
...
+					messageId: this.id,
+					selectedEmoji: clickedEmoji,
+				})
+			}
+		},
+
+		getDisplayNameForReaction(reactingParticipant) {
+			const displayName = reactingParticipant.actorDisplayName.trim()
+
+			if (reactingParticipant.actorType === ATTENDEE.ACTOR_TYPE.GUESTS) {
```
Antreesy, line 175: Guest check should be before creating a constant

src/store/messagesStore.js
```diff
@@ -521,10 +522,14 @@ const actions = {
...
 			if (Object.keys(parentInStore).length !== 0) {
 				context.commit('addMessage', message.parent)
-				context.dispatch('resetReactions', {
-					token: message.token,
-					messageId: message.parent.id,
-				})
+				const reactionsStore = useReactionsStore()
+				if (message.actorId !== context.getters.getActorId()) {
+					if (message.systemMessage.startsWith('reaction')) {
+						reactionsStore.fetchReactions(message.token, message.parent.id)
```
Antreesy, line 528: I'm afraid, that is still to many requests we won't use at all.
As I see it, here we could:
- check if new system message contains reactions different than we have now in store (and we don't need to do anything)
- instead of resetting existing details, we can mark them as outdated instead of resetting
  - so we need to fetch them only next time you want to see details

src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.vue
```diff
@@ -0,0 +1,230 @@
...
+-->
+
+<template>
+	<!-- reactions buttons and popover with details -->
+	<div class="reactions-wrapper">
+		<NcPopover v-for="reaction in Object.keys(detailedReactions)"
+			:key="reaction"
+			:delay="200"
+			:focus-trap="false"
+			:triggers="['hover']">
```
Antreesy, line 30:
```vue
@hover="reactionsStore.fetchReactions"
```

Maybe that would be enough?

src/stores/reactions.js
```diff
@@ -0,0 +1,218 @@
...
+		 *
+		 * @param {string} token The conversation token
+		 * @param {number} messageId The id of message
+		 *
+		 */
+		async fetchReactions(token, messageId) {
+			console.debug('getting reactions details')
+			try {
+				const response = await getReactionsDetails(token, messageId)
+				this.updateReactions({
```
Antreesy, line 206: Here we could patch messagesStore with new plain reactions data, if they are outdated. And in that case next fetch won't trigger a detailed reactions reset

src/store/messagesStore.js
DorraJaouad: > check if new system message contains reactions different than we have now in store (and we don't need to do anything)

The problem is that revoke reaction sys msg doesn't have all needed info so far ( emoji is lacking). However, considering our available resources in 'reaction' sys msg ( it includes all we need), we can skip fetch request when we receive the msg which can really minimize requests as adding a reaction is most frequent. So, I can get exactly which actor in which reaction in which message Id I will add, even the timestamp is included. 

> instead of resetting existing details

Here, if you mean in this conditional statement, getting the else condition is only for 'message_deleted', which is needed as the message is deleted so its reactions should be deleted.

src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.vue
DorraJaouad: As the count is now aggregated from reactions store, fetching is mandatory regardless of hover status, otherwise user can't see the reactions until they hover.

src/stores/reactions.js
DorraJaouad: I don't really understand how you see the patch is resetting the store 👾. Reset only happens when no reactions left after removing all emojis ( that means the data from the server comes empty and thus reset is rare to happen)

src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.vue
Antreesy: My intention was to keep both plain and detailed reactions in the store independently. Thus, general overview is always available, and if you want to see details, you simply make a specific request for this

DorraJaouad: So, as an overview : 

- Plain reactions ( aka reactions and their counts) are stored in Message and can be passed as a prop. This to always show up to date reactions count but only fetch details when you hover. A drawback to this way is that the count and the details are synchronized ( you see count outdated but the details are updated because of the recent fetch). However, I suggest a solution for it: Whenever there is a fetch on hover, we can replace the count coming from Message by the calculated length of detailed reactions.

-  In case we still want to get rid of the dependency via props passed from Message component. As I explained somewhere, we use the data from the quote reply of adding a reaction so we don't need to trigger any fetch requests. However, there will be a fetch request for each reaction revoke but there won't be many. Count is calculated by length in a computed prop and we don't need to store it. 

IMO, we can go with the second option.

src/store/messagesStore.js
Antreesy: I'd raise an issue for API here, to include either reaction itself, or ID of message which adds this reaction and is replaced with `reaction_deleted`

src/store/messagesStore.spec.js
```diff
@@ -87,7 +87,7 @@ describe('messagesStore', () => {
 		test('doesn\'t add specific messages to the store', () => {
 			testStoreConfig = cloneDeep(storeConfig)
 			testStoreConfig.modules.pollStore.getters.debounceGetPollData = jest.fn()
-			testStoreConfig.modules.reactionsStore.actions.resetReactions = jest.fn()
+			// testStoreConfig.modules.reactionsStore.actions.resetReactions = jest.fn()
```
Antreesy, line 90: Don't forget to remove or adjust these tests to use Pinia store

src/stores/reactions.js
```diff
@@ -0,0 +1,240 @@
...
+		 *
+		 * @param {object} payload action payload
+		 * @param {string} payload.token The conversation token
+		 * @param {number} payload.messageId The id of message
+		 * @param {string} payload.reaction The reaction to add
+		 * @param {object} payload.actors The users who reacted
+		 *
+		 */
+		addReaction({ token, messageId, reaction, actors }) {
+			if (this.reactions[token]?.[messageId]) {
```
Antreesy, line 77: both `if` possible outcomes don't covered by test. Don't we need to create a message, if we're adding a reaction?

check `this.checkForExistence` method in sharedItems store, maybe we could replicate it for this store as well

src/stores/reactions.js
```diff
@@ -0,0 +1,240 @@
...
+			if (Object.keys(storedReactions).length === 0) {
+				Vue.set(this.reactions[token], messageId, reactionsDetails)
+				return
+			}
+
+			// Handle removed reactions
+			const removedReactions = Object.keys(storedReactions).filter(reaction => {
+				return !reactionsDetails[reaction]
+			})
+			if (Object.keys(removedReactions).length > 0) {
```
Antreesy, line 176: Same, not covered by tests
`[ ].forEach()` shouldn't do anything for empty array, so I guess we could omit this `if` check

src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.vue
```diff
@@ -0,0 +1,246 @@
...
+		plainReactions() {
+			return this.$store.getters.message(this.token, this.id).reactions
+		},
+
+		reactionsSelf() {
+			return this.$store.getters.message(this.token, this.id).reactionsSelf
+		},
+
+		actorId() {
+			return this.$store.getters.getActorId()
```
Antreesy, line 141: Not covered by test, not used in the code

src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.vue
```diff
@@ -0,0 +1,246 @@
...
+		},
+	},
+
+	methods: {
+		fetchReactions() {
```
Antreesy, line 146: Not covered by tests, try https://test-utils.vuejs.org/api/#trigger

Also might be an edge case, but if somehow there is a misalignment `detailedReactions.length != plainReactions`, we should also fetch details

src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.vue
```diff
@@ -0,0 +1,246 @@
...
+			if (!this.hasReactions) {
+				this.reactionsStore.fetchReactions(this.token, this.id)
+			}
+		},
+
+		userHasReacted(reaction) {
+			if (!this.reactionsSelf) {
+				return false
```
Antreesy, line 154: Not covered by tests, maybe don't need this check? Also the output could be `false`, `undefined`, `emoji`, whereas you need only boolean
```js
return this.reactionsSelf?.includes(reaction)
```

src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.vue
```diff
@@ -0,0 +1,246 @@
...
+			}
+			return this.reactionsSelf.find(emoji => emoji === reaction)
+		},
+
+		async handleReactionClick(clickedEmoji) {
+			if (!this.canReact) {
+				showError(t('spreed', 'No permission to post reactions in this conversation'))
+				return
```
Antreesy, line 162: Not covered by tests. try to make a click in `shows reaction buttons with count but without emoji picker when no chat permission`

src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.vue
```diff
@@ -0,0 +1,246 @@
...
+			@after-show="fetchReactions">
+			<template #trigger>
+				<NcButton :type="userHasReacted(reaction) ? 'primary' : 'secondary'"
+					class="reaction-button"
+					@click="handleReactionClick(reaction)">
+					{{ reaction }} {{ reactionsCount(reaction) }}
+				</NcButton>
+			</template>
+
+			<div v-if="hasReactions" class="reaction-details">
```
Antreesy, line 40: Maybe we need a fallback here, to not show an empty popover?
```vue
<span v-else class="loading" />

File: src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.vue [hunk 16]
```diff
@@ -0,0 +1,246 @@
...
+			if (!this.hasReactions) {
+				return ''
+			}
+			const list = this.detailedReactions[reaction]
+			const summary = []
+
+			for (const item in list) {
+				if (list[item].actorType === this.$store.getters.getActorType()
+					&& list[item].actorId === this.$store.getters.getActorId()) {
+					summary.unshift(t('spreed', 'You'))
```
Antreesy, line 209: Not covered by test, thought very frequent to happen. Try https://test-utils.vuejs.org/api/#trigger

src/store/messagesStore.js
```diff
@@ -1351,9 +1355,9 @@ const actions = {
 			})
 			// The response return an array with the reaction details for this message
 			const response = await addReactionToMessage(token, messageId, selectedEmoji)
-			// We replace the reaction details in the reactions store and wipe the old
-			// values
-			context.dispatch('updateReactions', {
+
+			const reactionsStore = useReactionsStore()
```
Antreesy, line 1359: I'd move this and next action fully to reactionsStore. Apart from updating the message object, it doesn't really belong to messagesStore. You can use this workaround for now, until messagesStore isn't migrated to Pinia:
```js
const store = useStore()
store.commit('addReactionToMessage', payload)
```

src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.vue
```diff
@@ -0,0 +1,261 @@
...
+		 * Whether the plain reactions are more than the detailed ones.
+		 */
+		hasMoreReactions() {
+			return this.hasReactions
+					&& Object.keys(this.plainReactions).length > Object.keys(this.detailedReactions).length
+		},
+	},
+
+	watch: {
+		hasMoreReactions(value) {
```
Antreesy, line 149: Could be triggered, when you patched messageStore, but not yet reactionsStore. I'd check only inside `fetchReactions` method

src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.vue
```diff
@@ -0,0 +1,261 @@
...
+		reactionsSelf() {
+			return this.$store.getters.message(this.token, this.id).reactionsSelf
+		},
+
+		/**
+		 * Whether the plain reactions are more than the detailed ones.
+		 */
+		hasMoreReactions() {
+			return this.hasReactions
+					&& Object.keys(this.plainReactions).length > Object.keys(this.detailedReactions).length
```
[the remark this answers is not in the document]
DorraJaouad: If it is less, fetching won't affect the number of plainReactions. It is a different API.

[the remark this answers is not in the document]
Antreesy: If it is less, then you might have missed 'reaction_revoked' somehow, and still need to fetch all details. They will be actual and should overwrite messageStore

src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.vue
```diff
@@ -0,0 +1,254 @@
...
+		reactionsSelf() {
+			return this.$store.getters.message(this.token, this.id).reactionsSelf
+		},
+
+		/**
+		 * Whether the plain reactions are different than the detailed ones.
+		 */
+		hasMoreReactions() {
+			return this.hasReactions
+					&& Object.keys(this.plainReactions).length !== Object.keys(this.detailedReactions).length
```
Antreesy, line 144: We need to apply JSON.stringify here somehow, because you could have `{ '🎄': 2, '🔥': 2, '🔒': 2 }` in plain reactions, and `{ '🎄': 3, '🔥': 3, '🔒': 3 }` in detailed => general length is the same, so it passes your check

And I'd call it `hasOutdatedDetails` or something similar

Antreesy (commented):
Last one hopefully, just noticed some thing to optimize / fix

src/stores/reactions.js
```diff
@@ -0,0 +1,318 @@
...
+					actorType: message.actorType,
+					timestamp: message.timestamp,
+				}
+				this.addActorToReaction({
+					token: message.token,
+					messageId: message.parent.id,
+					reaction: message.message,
+					actor: actorObject,
+				})
+			} else if (message.systemMessage === 'reaction_revoked') {
```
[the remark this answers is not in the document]
DorraJaouad: We can't really use that because it might not be accurate :
* Plain reactions can be `{ '🎄': 2, '🔥': 2, '🔒': 2 }` for example, which is a reflection of ` { '🎄': ['alice','bob'], '🔥': ['jake','bob'], '🔒': ['alice','bob'] }` ( the actual final plainReactions)
* Details reactions that are currently stored are `{ '🎄': ['alice','bob'], '🔥': ['alice','bob'], '🔒': ['alice','bob'] }`. 
* We see that the 'calculated' plain reactions of Detailed reactions are the same with the message plain reactions but in depth, they are different.
* Because of alice revoke reaction and at the same time, we have new reaction coming later from jake, alice won't be removed from the list and we will have` '🔥': ['jake','alice','bob']` along with count is 3 (ofc this will trigger fetch on 'after-show' which will adjust the count and the fetch to 2).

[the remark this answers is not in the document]
Antreesy: Fair point, let's see how it feels and works

src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.vue
Antreesy: `JSON.stringify(this.plainReactions) !== JSON.stringify(this.detailedReactions)` will always be falsy!
plainReactions values are numbers, detailedReactions values - arrays

src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.spec.js
```diff
@@ -242,13 +242,13 @@ describe('Reactions.vue', () => {
 					NcPopover,
 				},
 			})
-			reactionsStored['🎄'].push({ actorDisplayName: 'user3', actorId: 'admin', actorType: 'users' })
-			let response = generateOCSResponse({ payload: reactionsStored })
-			addReactionToMessage.mockResolvedValue(response)
+			const addedReaction = [...reactionsStored['🎄'], { actorDisplayName: 'user3', actorId: 'admin', actorType: 'users' }]
```
Antreesy, line 245:
```suggestion
			const addedReaction = {
				...reactionsStored,
				'🎄': [...reactionsStored['🎄'], { actorDisplayName: 'user3', actorId: 'admin', actorType: 'users' }],
			}
```

Otherwise you're passing incorrect payload

Antreesy (commented):
Apart from that and `hasOutdatedDetails`, see nothing blocking

src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.vue
DorraJaouad: Almost ;)

PR #11283, opened 2023-12-21
fix(mixins): remove unused `participant` mixin

Antreesy:
### ☑️ Resolves

* `participant` mixin has only one computed prop
* most components, where it' imported, don't use it in the code (please double-check it)
* usage in Message/MessageButtonsBar is replacable

### 🏁 Checklist


DorraJaouad (approved):
Scanned , no other usage found.
Evaluated, the alternative way to replace usage is fitting.
Tested, nothing is blocking.

PR #11294, opened 2023-12-26
feat(federations): provide admin setting to enable Federation in Talk

### LINKED ISSUE #11279: 🌐 Federated invitation handling 💌 - Frontend
- Overview at https://github.com/nextcloud/spreed/issues/11229

---

- [x] Administration settings / Talk / Federations:
  - [x] Enable federations in Talk for this instance
  - [x] Allow users on this server to invite users to Talk conversations from all other servers / only trusted servers / only specific groups
- [x] Participants Tab:
  - [x] List remote users provided by `'core/autocomplete/get'` endpoint (should be there, if `settings/admin/sharing` is [configured](https://docs.nextcloud.com/server/latest/admin_manual/configuration_files/federated_cloud_sharing_configuration.html) and other servers provide their users
   - [x] Allow users on this server to invite users to Talk conversations from other servers (works if regex is matching `userid@domain`)
- [x] App level:
  - [x] invite - join conversation when invitation is accepted
    - [x] update participants list on moderator side from system message
  - [x] dismiss -verify that works as expected
  - [x] disinvite - verify that conversation is removed, participant is left and store is purged as for normal disinvite
  - [x] list - get all pending invitations from remore servers, and list them

Antreesy:
### ☑️ Resolves

* Part of #11279
  * Add Admin setting section for Federation (in debug mode only)

image

### 🚧 Tasks

- [x] Check wording

### 🏁 Checklist

---

### 🚧 Tasks

- [ ] Check code changes

### 🏁 Checklist

### DIFF
--- a/lib/Settings/Admin/AdminSettings.php
+++ b/lib/Settings/Admin/AdminSettings.php
@@ -68,6 +68,7 @@ class AdminSettings implements ISettings {
 		$this->initGeneralSettings();
 		$this->initAllowedGroups();
 		$this->initCommands();
+		$this->initFederation();
 		$this->initMatterbridge();
 		$this->initStunServers();
 		$this->initTurnServers();
@@ -107,6 +108,10 @@ class AdminSettings implements ISettings {
 		}, $commands);
 
 		$this->initialState->provideInitialState('commands', $result);
+	}
+
+	protected function initFederation(): void {
+		$this->initialState->provideInitialState('federation_enabled', $this->serverConfig->getAppValue('spreed', 'federation_enabled', 'no'));
 	}
 
 	protected function initMatterbridge(): void {
--- a/src/components/AdminSettings/Federation.vue
+++ b/src/components/AdminSettings/Federation.vue
@@ -1,0 +1,80 @@
+<!--
+ - @copyright Copyright (c) 2024 Maksim Sukharev <antreesy.web@gmail.com>
+ -
+ - @author Maksim Sukharev <antreesy.web@gmail.com>
+ -
+ - @license AGPL-3.0-or-later
+ -
+ - This program is free software: you can redistribute it and/or modify
+ - it under the terms of the GNU Affero General Public License as
+ - published by the Free Software Foundation, either version 3 of the
+ - License, or (at your option) any later version.
+ -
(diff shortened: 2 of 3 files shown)


src/components/AdminSettings/Federations.vue
```diff
@@ -0,0 +1,80 @@
...
+ -
+ - You should have received a copy of the GNU Affero General Public License
+ - along with this program. If not, see <http://www.gnu.org/licenses/>.
+ -
+ -->
+
+<template>
+	<section id="general_settings" class="federations section">
+		<h2>
+			{{ t('spreed', 'Federations') }}
```
nickvergessen, line 26: The thing is called "Federation" (singular there  is no plural of it in English)

nickvergessen (commented):
Should at least hide it when not on debug mode, so we don't forget about it.

Or we default federation to enabled when debug is on, then you also don't need the setting for now, right?

Antreesy:
> Or we default federation to enabled when debug is on, then you also don't need the setting for now, right?

Made changes, now it's visible in debug mode only
I'd keep in configurable anyway. For now it's a good base for putting here new settings, if they'll arise one day.

src/App.vue
```diff
@@ -519,25 +519,38 @@ export default {
...
+					hash: hash ? `#${hash}` : '',
+					params: {
+						token,
+					},
+				})
+
+				event.cancelAction = true
+				break
+			}
+			case 'POST': {
```
nickvergessen, line 545: unrelated?

Antreesy: Supposed to be there for handling notification actions like "Accept invitation", to redirect a user to federated conversation, but API doesn't return a token for that (as we discussed it earlier)

nickvergessen: but this PR is about the admin setting, not about notifications? :D

DorraJaouad: So are you just preparing the handling before the API is ready? you can just stash it or will you wait until API PR is created and merged?

nickvergessen: exactly what Dorra said. The second commit should be a separate PR :)

Antreesy: Fiine

DorraJaouad (approved):
Tested and wording is fine.

PR #11297, opened 2023-12-26
fix(shares): use FilePicker component directly

### LINKED ISSUE #11236: FilePicker doesn't open at FullScreen
### How to use GitHub

---

## Steps to reproduce
1. Open chat
2. Toggle FullScreen
3. Click on "Share from Nextcloud"
image

### Expected behaviour
FilePicker should be visible

### Actual behaviour
FilePicker is mounted by default at `body`, so same issue as with Viewer
We either can:
- set container with https://github.com/nextcloud-libraries/nextcloud-dialogs/blob/4ec4aa2f22e73e7e7a9a32e90af01cb55a7b0181/lib/filepicker-builder.ts#L134C1-L134C1
- migrate to using a component: https://github.com/nextcloud-libraries/nextcloud-dialogs/blob/main/lib/components/FilePicker/FilePicker.vue

:warning:  Haven't tested with stable27 yet

Antreesy:
### ☑️ Resolves

* Fix #11236
  * For chat attachments: added multiselect

`picker.setType` is deprecated, but it was responsible for buttons labels => `Choose` everythere

| 🏚️ Before | 🏡 After |
|------------|----------|

### 🚧 Tasks

- [ ] Hard to reproduce, but sometimes FilePicker dialog appears behind the main dialog (MediaSettings, ConversationSettings)

### 🏁 Checklist

### DIFF
--- a/src/components/SettingsDialog/SettingsDialog.vue
+++ b/src/components/SettingsDialog/SettingsDialog.vue
@@ -38,12 +38,11 @@
 				{{ locationHint }}
 			</h3>
 			<div class="app-settings-section__wrapper">
-				<NcTextField class="app-settings-section__input"
-					:value="attachmentFolder"
-					:disabled="true"
-					@click="selectAttachmentFolder" />
+				<p class="app-settings-section__input" @click="showFilePicker = true">
+					{{ attachmentFolder }}
+				</p>
 				<NcButton type="primary"
-					@click="selectAttachmentFolder">
+					@click="showFilePicker = true">
 					{{ t('spreed', 'Browse …') }}
 				</NcButton>
 			</div>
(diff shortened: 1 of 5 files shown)


src/components/SettingsDialog/SettingsDialog.vue
```diff
@@ -374,13 +387,19 @@ export default {
...
-			cursor: default;
-		}
+		margin: 0;
+		padding: 10px 6px 10px 12px;
+		border: 2px solid var(--color-border-maxcontrast);
+		border-radius: var(--border-radius-large);
+		background-color: var(--color-main-background);
+		font-size: var(--default-font-size);
+		text-overflow: ellipsis;
+		color: var(--color-main-text);
```
DorraJaouad, line 400: not needed

src/components/SettingsDialog/SettingsDialog.vue
```diff
@@ -268,30 +286,25 @@ export default {
...
-					this.attachmentFolderLoading = false
-				})
+			this.attachmentFolderLoading = true
+			try {
+				this.$store.dispatch('setAttachmentFolder', path)
+			} catch (exception) {
+				showError(t('spreed', 'Error while setting attachment folder'))
+			}
+			this.attachmentFolderLoading = false
+			this.showFilePicker = false
```
DorraJaouad, line 307: This is not needed by the way, because close event is emitted

DorraJaouad (approved):
Tested, nothing is blocking.

src/components/SettingsDialog/SettingsDialog.vue
Antreesy: Only color or all of these styles? I tried to fake input appearance, to not import NcTextField component for that

DorraJaouad: Only color

PR #11301, opened 2023-12-27
chore(mixins): remove `video` mixin

Antreesy:
### ☑️ Resolves

* Removes a mixin from video component before migration to Vue 3
* Clean up unused properties, refactor migrated code
* Lint files

No visual changes

### 🏁 Checklist


DorraJaouad:
Why not migrating it to Vue 3's Composition API ?

Antreesy:
> Why not migrating it to Vue 3's Composition API ?

Doesn't make sense:
- data: `mouseover` only needed, and it would be a lot of boilerplate code to achive the same result with composable
- props: couldn't be defined in composable +  it's hard to find them, if they are on a different file
- computed: styles for classes are duplicated in components, couldn't be moved to composable, one-line in template
- methods: after refactoring it's a one-line in template, so same as computed

src/components/CallView/Grid/Grid.vue
```diff
@@ -67,8 +67,6 @@
 									:is-stripe="isStripe"
 									:is-promoted="sharedDatas[callParticipantModel.attributes.peerId].promoted"
 									:is-selected="isSelected(callParticipantModel)"
-									:fit-video="false"
```
DorraJaouad, line 70: As you are not passing it as a prop, remove it from Video.

Antreesy: It could passed, I just don't see a reason to keep this line here, as fit Video prop is false by default

DorraJaouad: oops, found the other instance of the component. nvm

DorraJaouad (approved):
It is good to go.

PR #11328, opened 2024-01-01
Feat(Vue3): Migrate isInLobby mixin to composable.

DorraJaouad:
### ☑️ Resolves

Migrate isInLobby mixin to conversations store.

### 🚧 Tasks

- [ ] Test if everything is working fine.


src/composables/useIsInLobby.js
```diff
@@ -0,0 +1,59 @@
...
+      && (conversation.value.participantType === PARTICIPANT.TYPE.OWNER
+        || conversation.value.participantType === PARTICIPANT.TYPE.MODERATOR
+        || conversation.value.participantType === PARTICIPANT.TYPE.GUEST_MODERATOR)
+	})
+
+	const isModeratorOrUser = computed(() => {
+		return conversation.value && !conversation.value.isDummyConversation && (isModerator.value
+      || conversation.value.participantType === PARTICIPANT.TYPE.USER
+      || conversation.value.participantType === PARTICIPANT.TYPE.USER_SELF_JOINED)
+	}).value
```
Antreesy, line 49:
```suggestion
	})
```
You break reactivity here by adding value (`computed.value` is boolean, so it doesn't return a reference to updated object).
It still can be used in Options API as `this.isModeratorOrUser`, so there will be no extra changes in the component

src/composables/useIsInLobby.js
```diff
@@ -0,0 +1,59 @@
...
+ *
+ */
+export function useIsInLobby() {
+	const store = useStore()
+	const token = computed(() => store.getters.getToken())
+	const conversation = computed(() => store.getters.conversation(token.value))
+
+	const isModerator = computed(() => {
+		return conversation.value
+      && (conversation.value.participantType === PARTICIPANT.TYPE.OWNER
```
Antreesy, line 40:
```suggestion
		const type = conversation.value?.participantType
		return type === PARTICIPANT.TYPE.OWNER
```

to call computed props less
similar could be done for others

src/composables/useIsInLobby.js
Antreesy: cc @ShGKme for confirmation and double-check

src/mixins/getParticipants.js
```diff
@@ -40,6 +38,15 @@ const getParticipants = {
 		}
 	},
 
+	setup() {
```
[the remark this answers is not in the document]
ShGKme: `setup` doesn't work in mixins. Mixins are Options API only and not compatible with Composition API by design (as Composition API was designed to get rid of mixins in Vue).

src/composables/useIsInLobby.js
```diff
@@ -0,0 +1,59 @@
...
+import { useStore } from './useStore.js'
+import { PARTICIPANT, WEBINAR } from '../constants.js'
+
+/**
+ *
+ */
+export function useIsInLobby() {
+	const store = useStore()
+	const token = computed(() => store.getters.getToken())
+	const conversation = computed(() => store.getters.conversation(token.value))
```
ShGKme, line 36: If the only work of this composable is to calculate some values based on a store state, then it should be 3 getters in the store.

It doesn't work with a component instance context, doesn't have any params.

So we don't need to calculate a new copy of these values in each instance as we base on global state anyway.

src/composables/useIsInLobby.js
ShGKme: > to call computed props less

`computed` are not "called". When it is calculated once, it is just a property with a value. It has almost no performance overhead (no more than using any reactive property in Vue).

src/composables/useIsInLobby.js
ShGKme: Yes, returning a value is returning a constant `true/false` value

ShGKme (changes requested):

src/composables/useIsInLobby.js
ShGKme: > to call computed props less

`computed` are not "called". When it is calculated once, it is just a property with a value. It has almost no performance overhead (no more than using any reactive property in Vue).

ShGKme:
Note: I saw a number of PR with Vuex->Pinia and Mixins -> Composables migration with a `Vue 3` scope name. Just a reminder that neither of them is required for Vue 3 migration.

src/composables/useIsInLobby.js
DorraJaouad: Moved to conversations store.

src/store/conversationsStore.js
```diff
@@ -137,6 +137,26 @@ const getters = {
...
+				|| conversation?.participantType === PARTICIPANT.TYPE.USER
+				|| conversation?.participantType === PARTICIPANT.TYPE.USER_SELF_JOINED)
+	},
+	isInLobby: (state, getters) => token => {
+		const conversation = getters.conversation(token)
+		return conversation
+				&& conversation.lobbyState === WEBINAR.LOBBY.NON_MODERATORS
+				&& !getters.isModerator(token)
+				&& (conversation.permissions & PARTICIPANT.PERMISSIONS.LOBBY_IGNORE) === 0
+	},
```
ShGKme, line 159: These getters are always used with only one token value — the value from the store: `this.$store.getters.getToken()`.

In the previous solution with the composable in this PR the token also was got from `store.getters.getToken()`

We may directly the the value from the store and get rid of computing a function as a value instead of the value itself. Returning a function in computed leads to losing caching of the value so than each component instance runs this function again.

DorraJaouad: > We may directly the the value from the store and get rid of computing a function as a value instead of the value itself.

I am not sure if you meant this 
```
isInLobby() {
			return this.$store.getters.isInLobby
		},
```

instead of this 
```
isInLobby() {
			return this.$store.getters.isInLobby()
		},
```

The first approach is function reference so I think it is not what you meant by accessing the value itself ? 
Anyway, Vuex has its caching advantage too. So, despite that computed instances are constantly calling the computed function and executing it,  the value will be returned instantly as it is cached by Vuex.

Antreesy: I guess he meant `context.getters.conversation(context.rootGetters.getToken)` without passing it

ShGKme: > The first approach is function reference so I think it is not what you meant by accessing the value itself ?

I meant, that it should not be a function. Getter in Vuex is the same as a computed in components. $store.getters.foo is a function only, when the getter function computes a new function as a value.

> Anyway, Vuex has its caching advantage too. So, despite that computed instances are constantly calling the computed function and executing it, the value will be returned instantly as it is cached by Vuex.

Yes, but when we use a getter to compute a new getter function, the result of the computing and the cached value is a new function, not the final Boolean value. Then each component instance, that uses this getter, then calls this function to compute the value and cache it instances, not the store. Also we cannot see the value in the Vuex devtools.

ShGKme (approved):
Thanks!

Antreesy (approved):

PR #11122, opened 2023-12-05
fix(shares): handle silent sending and input paste correctly

Antreesy:
### ☑️ Resolves

* Fix #11117 
  * Button for silent send is brought back to upload dialog
  * Following API, silent flag is passed with share metadata
* Fix #11118 
  * Same as for drag-n-drop in dialog, if current upload is existing, we reuse it

### 🚧 Tasks

- [x] To test 11117:
  - [x] as user 2, enable notification in conversation for "All messages"
- [x] To test 11118:
  - [x] Ctrl+C, Ctrl+V

### 🏁 Checklist

- [x] 🌏 Tested with Chrome, Firefox and Safari or should not be risky to browser differences
- [x] ⛑️ Tests are included or not possible

PR #11330, opened 2024-01-02
fix(shares): pass the correct payload to the upload method

Antreesy:
### ☑️ Resolves

* Regression from #11122 
* Fix https://github.com/nextcloud/talk-desktop/issues/473

No visual changes
### 🚧 Tasks

- [ ] Test file upload against server `stable27`
  - Curent step is fixed for me, but local setup doesn't upload files because of CORS probably

### 🏁 Checklist

### DIFF
--- a/src/components/NewMessage/NewMessageUploadEditor.vue
+++ b/src/components/NewMessage/NewMessageUploadEditor.vue
@@ -71,7 +71,7 @@
 				<NcButton type="tertiary" @click="handleDismiss">
 					{{ t('spreed', 'Dismiss') }}
 				</NcButton>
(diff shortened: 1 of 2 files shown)


ShGKme:
@nickvergessen should we backport it to stable27 to support Talk Desktop?

https://github.com/nextcloud/talk-desktop/issues/478

Antreesy:
> backport it to stable27

stable27 doesn't have caption at all, so won't it be fixed with new desktop release for all versions?

ShGKme:
Oh, right. No need to change anything on the server for Talk Desktop. Sorry, tied, The backport just wasn't released yet 🙈

PR #11333, opened 2024-01-03
chore(mixins): migrate device mixin to composable

Antreesy:
### ☑️ Resolves

* Migrate mixin for MediaDevicesPreview and MediaSettings to composable
* Move some code to the component, clean props

No visua changes

### 🚧 Tasks

- [ ] Code review
- [ ] Manual testing

### 🏁 Checklist


src/components/MediaDevicesPreview.vue
```diff
@@ -101,7 +103,91 @@ export default {
...
+			currentThreshold,
+			audioPreviewAvailable,
+			videoPreviewAvailable,
+			audioInputId,
+			videoInputId,
+			audioStream,
+			audioStreamError,
+			videoStream,
+			videoStreamError,
+		} = useDevices(video, true)
```
DorraJaouad, line 120: Shouldn't be called onMounted hook ? Initially, video ref will remain null until "video" element ref is available.

Antreesy: Will check it, but it got called twice or later after mount, so video element is usually available

ShGKme: Composabled must always be called only and only in `setup`s. "composable" is literally "a part of" - a part of composing = setup of the component.

On the moment `useDevices` there is indeed no `video` element. But this is fine.

`video` is not an element, but a reactive variable, (reference to a reactive object - ref). `video` will have a new value (`.value`) when component is moubnted/re-rendered and video element created or changed.

There still might be an issue in `useDevice` side. It should not use `video` initially. But until it is used in `onMounted` or later - it is fine.

ShGKme: A secret note:

<details>

Actually, composables can be used in other places where Vue provides context:
- `computed()`
- `render()`
- Life cycle hooks like mounted

But it should be avoided when possible because it may result in unexpected behavior (breaks in async functions, event handlers, may not be supposed to run many times in one instance).

</details>

ShGKme (approved):
Looks good according to the code, but haven't tested...

src/components/MediaDevicesPreview.vue
Antreesy: UPD: we have guardians to check if video ref is available (defined), so it's safe

DorraJaouad (approved):
Tested, code-wise also good.

PR #11334, opened 2024-01-04
Refactor(Mixins): Change getParticipants mixin to composable

DorraJaouad:
### ☑️ Resolves
* Migrate to composable
* Added a small tweak on joined conversation for one-to-one conversation because it is better to get the status instantly.


src/composables/useGetParticipants.js
```diff
@@ -0,0 +1,135 @@
...
+	const onRouteChange = () => {
+		participantsInitialised.value = store.getters.participantsList(token.value).length > 1
+	}
+
+	const onJoinedConversation = () => {
+		if (oneToOneConversation.value) {
+			cancelableGetParticipants()
+		}
+		nextTick(debounceUpdateParticipants())
+	}
```
Antreesy, line 87:
```suggestion
		} else {
			nextTick(debounceUpdateParticipants)
		}
	}
```
Without `else` you're fetching it twice, and not sure we need parenthesis here for a callback
Apart from that good, conversation should be already joined by the moment

src/composables/useGetParticipants.js
```diff
@@ -0,0 +1,135 @@
...
+	const conversation = computed(() => store.getters.conversation(token.value))
+	const isInCall = useIsInCall()
+	const oneToOneConversation = computed(() => conversation.value.type === CONVERSATION.TYPE.ONE_TO_ONE)
+
+	// External use
+	const participantsInitialised = ref(false)
+	const fetchingParticipants = ref(false)
+	const pendingChanges = ref(false)
+	let debounceFastUpdateParticipants = ref(null)
+	let debounceSlowUpdateParticipants = ref(null)
```
Antreesy, line 50:
```suggestion
	let _debounceFastUpdateParticipants = null
	let _debounceSlowUpdateParticipants = null
```

src/composables/useGetParticipants.js
```diff
@@ -0,0 +1,135 @@
...
+	// Encapsulation
+	const store = useStore()
+	const token = computed(() => store.getters.getToken())
+	const conversation = computed(() => store.getters.conversation(token.value))
+	const isInCall = useIsInCall()
+	const oneToOneConversation = computed(() => conversation.value.type === CONVERSATION.TYPE.ONE_TO_ONE)
+
+	// External use
+	const participantsInitialised = ref(false)
+	const fetchingParticipants = ref(false)
```
Antreesy, line 47:
```suggestion
	let _fetchingParticipants = false
```
It's internal variable, as I see from the code. Shouldn't be a reactive also

src/components/TopBar/TopBar.vue
```diff
@@ -350,9 +358,9 @@ export default {
 			// Group conversations have mixin in RightSidebar, so should work only for one-to-one
 			handler(newValue) {
 				if (newValue) {
-					this.initialiseGetParticipantsMixin()
+					this.initialiseGetParticipants()
```
Antreesy, line 361: you can move this watcher from the `TopBar` component, as you have a getter for `isOneToOneConversation` in the composable. Then you don't need to export so many things =)

src/components/RightSidebar/Participants/ParticipantsTab.vue
```diff
@@ -121,13 +123,25 @@ export default {
...
+			stopGetParticipants,
+		} = useGetParticipants(props.isActive)
 
 		return {
 			sortParticipants,
 			isInCall,
+			participantsInitialised,
+			cancelableGetParticipants,
+			initialiseGetParticipants,
+			pendingChanges,
```
Antreesy, line 143: Same here for watcher on `isActive`. You probably need to make it reactive inside the composable, but thus you don't have to export `pendingChanges` and `debounceUpdateParticipants` (which you forgot btw)

src/composables/useGetParticipants.js
DorraJaouad: `debounceUpdateParticipants` will be called after 3sec the earliest, my intention is to add the initial request at 0sec and keep the debounce update

Antreesy: But why do we need to fetch two participants one more time after 3 seconds?

DorraJaouad: To keep a consistent interval? debounce is on each 3sec.

src/composables/useGetParticipants.js
ShGKme, line 50: Having a `ref` variable defined by `let` is a flag of a possible issue.

`ref` is a reactive variable, object-container for a value. It has reactivity, but it works only while it is used as a const reference for a mutable object. So it must be a `const`, `ref` should never be re-defined.

But here these variables should not be `ref`s, because they don't store a reactive value, they are constant functions.

(See also next comment).

src/composables/useGetParticipants.js
```diff
@@ -0,0 +1,135 @@
...
+	})
+
+	return {
+		pendingChanges,
+		participantsInitialised,
+		initialiseGetParticipants,
+		stopGetParticipants,
+		debounceUpdateParticipants,
+		cancelableGetParticipants,
+	}
```
ShGKme, line 134: Here first `return` return `debounceFastUpdateParticipants` variable with an empty `ref`.
And then in `onMounted` variable is changed to a function.

But `return` happened already, and the result of the composable is a lost empty `ref(null)`.

It would work, if the value is changed here `debounceFastUpdateParticipants.value = debounce`
But debounce function can be defined initially, there is no need to wait for mounting to DOM.

So `debounceFastUpdateParticipants` should be a const function.

ShGKme (changes requested):

src/composables/useGetParticipants.js
ShGKme: `_` prefix is a common practice to define "aka private" properties in JS objects (including component instances). But here it is just a variable, and there is no separation to "private" `debounceFastUpdateParticipants` and "public" `debounceFastUpdateParticipants`. It is a general local variable.

src/composables/useGetParticipants.js
```diff
@@ -0,0 +1,135 @@
...
+
+import { useIsInCall } from './useIsInCall.js'
+import { useStore } from './useStore.js'
+import { CONVERSATION } from '../constants.js'
+import { EventBus } from '../services/EventBus.js'
+
+/**
+ * @param {boolean} isActive whether the participants tab is active
+ */
+export function useGetParticipants(isActive = true) {
```
ShGKme, line 36: I haven't checked if it affects current composable usage, but for a general case for composables development.

If a composable's param is a plain non-reactive value, composable cannot react on it.

`isActive` is get from `props`. Props in theory can change. But there will be no changes in this composable, when `isActive` prop is changed.

It should be a reactive variable `ref`. If we want to support both reactive `ref` and plain value, we can use it in composable via `unref`.

src/composables/useGetParticipants.js
ShGKme: Same here

Antreesy: There is no need to expose them in composable return and use outside, so they seem private to me. Did the same for another PR just to visually distinct them

src/composables/useGetParticipants.js
Antreesy: Should be fine if we'll omit `onMounted` here, and assign it during the initialization?
`debounce` here returns an object with  `.clear` property, so it's not a simple function

src/composables/useGetParticipants.js
Antreesy: So we should do something like this in th composable body?
```js
const isActiveRef = ref(isActive)
```

src/composables/useGetParticipants.js
ShGKme: In JS and Python `_` prefix patter was used where it is not possible to separate private properties and all properties are exposed. So `_` is used to mark exposed property as a private property.

Here variables are private until they are implicitly exposed in `return`. Composable is a function, and variables here are the same local variable as in, for example, services or utils where we also have a local variable and return some of them.

Antreesy: Matter of preferences then. I'm not insisting, and will roll it back in my PR for consistency)

src/composables/useGetParticipants.js
ShGKme: > Should be fine if we'll omit `onMounted` here, and assign it during the initialization?

Yes, and it can be a `const`. We don't need to re-assign it, it is always the same function that exists initially.

> an object with `.clear` property, so it's not a simple function

Still a function =D

Functions have methods and properties (like `toString()` or `bind()`). And JS allows this:

```js
function foo() {}

foo.bar = 'bar'
foo.clear = () => {}
```

src/composables/useGetParticipants.js
DorraJaouad: Thank you for all explanations :)

src/composables/useGetParticipants.js
DorraJaouad: I made it a ref now but I didn't convert it into a plain value in composable in order to watch its change.

ShGKme: > So we should do something like this in th composable body?

No. This would just make a new ref, which initial value is based on initial prop value.

Example

```js
function useA(param) {
  // Not reactive, param is just a plain value
  return computed(() => param * 2)
}

function useB(param) {
  // Not reactive, paramRef is a new variable without relation with param
  const paramRef = ref(param)
  return computed(() => paramRef.value * 2)
}

function useC(paramRef) {
  // Reactive, but requires param to be ref (otherwise there is no .value)
  return computed(() => paramRef.value * 2)
}

function useD(param) {
  // Reactive, unversal - can be used with not reactive param
  // unref(param) = isRef(param) ? param.value : param
  return computed(() => unref(param) * 2)
}

export default defineComponent({
  props: {
    foo: String,
  },
  
  setup(props) {
    // Not reactive, a plain value (String) is passed
    useD(props.foo)
    
    // Reactive
    useD(toRef(props, 'foo'))
    // Reactive
    const { foo } = toRefs(props)
    useD(foo)
    // Reactive, Vue 3.2+
    useD(toRef(() => props.foo))
  },
})
```

src/components/RightSidebar/Participants/ParticipantsTab.vue
```diff
@@ -181,28 +195,22 @@ export default {
 			return this.searchText !== ''
 		},
 		noResults() {
-			return this.searchResults === []
+			return this.searchResults.length === 0
```
ShGKme, line 198: Nice catch 🙈

src/composables/useGetParticipants.js
```diff
@@ -0,0 +1,147 @@
...
+			debounceUpdateParticipants()
+		}
+	})
+
+	return {
+		participantsInitialised,
+		initialiseGetParticipants,
+		stopGetParticipants,
+		cancelableGetParticipants,
+	}
```
ShGKme, line 146: Nice to move `watch` here. It actually makes `initialiseGetParticipants` and `stopGetParticipants` unneeded outside composable, making it even simpler to use.

Only `beforeDestroy` from `ParticipantsTab.vue` should also be moved to composable.

src/composables/useGetParticipants.js
```diff
@@ -0,0 +1,147 @@
...
+	const onRouteChange = () => {
+		participantsInitialised.value = store.getters.participantsList(token.value).length > 1
+	}
+
+	const onJoinedConversation = () => {
+		if (isOneToOneConversation.value) {
+			cancelableGetParticipants()
+		} else {
+			nextTick(debounceUpdateParticipants())
+		}
```
ShGKme, line 85: As @Antreesy noticed above.

Otherwise `debounceUpdateParticipants()` is run immediately, and its return value `undefined` is passed to the `nextTick`.

```suggestion
		} else {
			nextTick(() => debounceUpdateParticipants())
		}
```

src/composables/useGetParticipants.js
DorraJaouad: I kept it defined like this (not a function directly)
```js
const debounceFastUpdateParticipants = debounce(
		cancelableGetParticipants, 3000)
```

Because `.clear()` is a method related to debounce and cannot be called from a wrapping function.
```js
function debounceFastUpdateParticipants() {
   debounce(cancelableGetParticipants, 3000)
}
```

But if we want to keep it as function, this means that .clear should be explicitly defined to `debounceFastUpdateParticipants` like this ? 
```js
debounceFastUpdateParticipants.clear = () => {
  debounce(cancelableGetParticipants, 3000).clear()
}
```

ShGKme: This is a correct defining of a debounced function. `debounce` returns a function, so `debounceFastUpdateParticipants` is a function here.

```js
const debounceFastUpdateParticipants = debounce(
		cancelableGetParticipants, 3000)
```

DorraJaouad:
Squashed fixups and rebased

src/composables/useGetParticipants.js
```diff
@@ -0,0 +1,152 @@
...
+		if (newValue && pendingChanges) {
+			debounceUpdateParticipants()
+		}
+	})
+
+	onBeforeUnmount(
+		() => {
+			stopGetParticipants()
+		}
+	)
```
ShGKme, line 145: nitpick:

```suggestion
	onBeforeUnmount(() => {
		stopGetParticipants()
	})
```

src/composables/useGetParticipants.js
```diff
@@ -0,0 +1,152 @@
...
+
+	return {
+		participantsInitialised,
+		initialiseGetParticipants,
+		cancelableGetParticipants,
+	}
```
ShGKme, line 151: `initialiseGetParticipants` is not needed exposed as well as executing it in `ParticipantsTab`. It is executed on created anyway by the immediate watch above.

Running it manually in the component runs it a second time on the instance.

DorraJaouad: The immediate watch is for one to one conversations only. Group conversations have the composable initialized in `ParticipantsTab` and gets stopped when it is unmounted, one to one conversations have the composable initialized in `TopBar` and gets stopped as soon as you change the conversation to other then one to one. 

It can't be called twice because participants tab is not mounted when isOneToOneConversation is truthy so there is no overlap. 
The only thing that will be called twice is `stopGetParticipants `when going from one to one conv to group conv because `TopBar `and `ParticipantsTab` are both mounted and they execute the immediate watch in case of `!isOneToOneConversation`.

I think the only way to avoid the duplicated call is to move `stopGetParticipants` back to `TopBar` 🤔 .

Antreesy: Or we consider keeping a composable in TopBar only, as it's always mounted for all types of conversations. Then we won't need to invent and care about some conditions to run it in one place and stop in another

DorraJaouad: GetParticipants needs the composable ( `participantsInitialised` and `cancelableGetParticipants`)

Antreesy: Right, but that's only readonly ref and fetch request. So we could keep composable in both, but:
- enable/disable listeners only in TopBar
- migrate and modify sidebarStore, so it contains value of activeTab (then we could use it for fast/slow update)
- detach initialization logic from sidebar, keep only `participantsInitialised` and `cancelableGetParticipants`

DorraJaouad: > migrate and modify sidebarStore, so it contains value of activeTab (then we could use it for fast/slow update)

I don't want to make the pace of update dependent on sidebar because this implies that fast update is limited to only when participants tab is open. One-to-one conversation avatar update will be easily noticeable that it is slow. 

For the migration, sure but follow-up :).

src/components/RightSidebar/Participants/ParticipantsTab.vue
```diff
@@ -121,13 +124,20 @@ export default {
 		},
 	},
 
-	setup() {
+	setup(props) {
+		const { isActive } = toRefs(props)
 		const { sortParticipants } = useSortParticipants()
 		const isInCall = useIsInCall()
+		const {
+			participantsInitialised,
```
Antreesy, line 132: Always false, as you don't fetch it from ParticipantsTab anymore, and don't change refs value
image

Should we move `participantsInitialised` to participantsStore as getter?
> [!TIP]
> Each conversation adds a user participant to the store, but it's a dummy object (not the complete, as we get from server). We could check in getter, if store has only this participant
https://github.com/nextcloud/spreed/blob/38d19fe4082b7cbadaff16af4b5b6a7c537834af/src/store/conversationsStore.js#L281

src/composables/useGetParticipants.js
```diff
@@ -0,0 +1,136 @@
...
+
+	onBeforeUnmount(() => {
+		if (isTopBar) {
+			stopGetParticipants()
+		}
+	})
+
+	return {
+		cancelableGetParticipants,
+	}
```
Antreesy, line 135:
```suggestion
	return cancelableGetParticipants
```

Then you don't need to destructure it in the component

Antreesy (approved):
Tested. Reactivity and methods are working fine, couldn't notice anything broken

src/composables/useGetParticipants.js
DorraJaouad: Thank god

ShGKme: > Then you don't need to destructure it in the component

This is a common practice to return object with values in composables, if there is any chance that this composable will return more values in future (if this is not one-value composable like `isMobile`)

Antreesy: Good timing =) Let's hope we don't need to export something more, the code looks pretty decent and isolated now

ShGKme (approved):

PR #11343, opened 2024-01-06
fix(reactions): add guardian before resetting reactions

Antreesy:
### ☑️ Resolves

* Fix `resetReactions` action, which prevents rendering of following messages
* Add test for `purgeReactionsStore`

| 🏚️ Before | 🏡 After |
|------------|----------|

### 🏁 Checklist

### DIFF
--- a/src/stores/__tests__/reactions.spec.js
+++ b/src/stores/__tests__/reactions.spec.js
@@ -6,7 +6,7 @@ import { showError } from '@nextcloud/dialogs'
 
 import { getReactionsDetails, addReactionToMessage, removeReactionFromMessage } from '../../services/messagesService.js'
 import vuexStore from '../../store/index.js'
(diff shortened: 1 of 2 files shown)


src/stores/__tests__/reactions.spec.js
```diff
@@ -191,6 +191,18 @@ describe('reactionsStore', () => {
 		expect(getReactionsDetails).toHaveBeenCalled()
 		expect(Object.keys(reactionsStore.reactions[token][messageId])).toEqual(['🎄', '🔥'])
 	})
+	it('purges the reactions store', () => {
+		// Arrange
+		reactionsStore.updateReactions({
+			token,
+			messageId,
+			reactionsDetails: reactions
+		})
```
DorraJaouad, line 200: It's already there in beforeEach()

DorraJaouad (approved):
Thanks

PR #11349, opened 2024-01-08
fix(Message): optimize getters and provided props

Antreesy:
### ☑️ Resolves

* Delete unused `scrollerBoundingClientRect` prop
* Move computation of `isLastCallStartedMessage` to store
* Make update of `hasCall` more clear across the app
  * Prevent notifications from updating store if already done

No visual changes

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -247,9 +247,6 @@ const isTranslationAvailable = getCapabilities()?.spreed?.config?.chat?.['has-tr
 	// Fallback for the desktop client when connecting to Talk 17
 	?? getCapabilities()?.spreed?.config?.chat?.translations?.length > 0
 
-/**
- * @property {object} scrollerBoundingClientRect provided by MessageList.vue
- */
 export default {
 	name: 'Message',
 
@@ -271,8 +268,6 @@ export default {
 		UnfoldLess,
 		UnfoldMore,
 	},
-
-	inject: ['scrollerBoundingClientRect'],
 
 	inheritAttrs: false,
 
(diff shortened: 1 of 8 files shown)


src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -508,23 +503,12 @@ export default {
...
-			return this.$store.getters.messagesList(this.token)
-		},
-
 		isLastCallStartedMessage() {
-			// FIXME: remove dependency to messages list and convert to property
-			const messages = this.messagesList
-			// FIXME: don't reverse the whole array as it would create a copy, just do an actual reverse search
-			const lastCallStartedMessage = messages.reverse().find((message) => message.systemMessage === 'call_started')
-			return lastCallStartedMessage ? (this.id === lastCallStartedMessage.id) : false
+			return this.systemMessage === 'call_started' && this.id === this.$store.getters.getLastCallStartedMessageId(this.token)
```
DorraJaouad, line 507: You can return the value of the getter and use `getToken` in the store

src/store/conversationsStore.js
```diff
@@ -216,11 +216,7 @@ const mutations = {
 	},
 
 	overwriteHasCallByChat(state, { token, hasCall }) {
-		if (hasCall) {
-			Vue.set(state.conversations[token], 'hasCallOverwrittenByChat', hasCall)
-		} else {
-			Vue.delete(state.conversations[token], 'hasCallOverwrittenByChat')
-		}
+		Vue.set(state.conversations[token], 'hasCallOverwrittenByChat', hasCall)
 	},
```
DorraJaouad, line 220: So why are you keeping it then ?

Antreesy: good question. I thought to mark the conversation as modified, until we fetch a proper one from server

DorraJaouad (approved):
Tested 🟢

Issue #11103, opened 2023-12-04
"Message deleted by author" not gray enough


moritz-t-w:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Steps to reproduce
1. Open the web interface with darkmode
2. Delete a message
3. Send a message "Message deleted by author"
4. Blindfold yourself and spin around
5. Try to pick the one that's actually deleted

### Expected behaviour
The actually deleted message has a clearly different color, maybe an icon, maybe italicised.
### Actual behaviour
The messages are hardly distinguishable
![Actual deleted message on the top vs message saying "Message deleted by author" below](https://github.com/nextcloud/spreed/assets/90388353/f8c36cd6-2357-498e-a55c-501df0c34fde)

## Talk app
n/a

## Browser

**Microphone available:** n/a

**Camera available:** n/a

**Operating system:** Arch Linux

**Browser name:** Firefox

**Browser version:** latest

### Browser log

n/a

## Server configuration
n/a

### Server log (data/nextcloud.log)
n/a

> I left out a lot of info about the server cause I don't have access. I don't think it's relevant tho.

moritz-t-w:
![image](https://github.com/nextcloud/spreed/assets/90388353/06eba4f3-259f-41fc-86f2-e7f9324c4b8d)
WhatsApp uses an icon to display that a message has been deleted, I think that would be a nice solution, also making it easier to see at a glance instead of reading that it was deleted.

moritz-t-w:
![image](https://github.com/nextcloud/spreed/assets/90388353/cf79470b-7c1b-4d74-9bf5-21fcd7ad23ba)
Here's a mockup

PR #11361, opened 2024-01-09
enh(Message): split out MessageBody component

Antreesy:
### ☑️ Resolves

* Split SFC file into smaller parts
* Optimize methods and getters, rename props
* Mark deleted message additionaly (fix https://github.com/nextcloud/spreed/issues/11103)

* system messages were moved 4px to the right to be aligned with normal messages
* deleted message:

Before | After
-- | --
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -420,10 +420,6 @@ export default {
 					const props = Object.assign({}, this.messageParameters[p])
 					// Add the token to the component props
 					props.token = this.token
-					// The word 'name' is reserved in for the component name in
-					// Vue instances, so we cannot pass that into the component
-					// as a prop, therefore we rename it into pollName
-					props.pollName = this.messageParameters[p].name
 					richParameters[p] = {
 						component: Poll,
 						props,
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -48,6 +48,7 @@
 				:timestamp="timestamp"
 				:is-deleting="isDeleting"
 				:is-temporary="isTemporary"
+				:has-call="conversation.hasCall"
 				:sending-failure="sendingFailure"
 				:show-common-read-icon="showCommonReadIcon"
 				:common-read-icon-tooltip="commonReadIconTooltip"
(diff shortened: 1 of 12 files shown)


src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -41,113 +36,24 @@ the main body of the message as well as a quote.
...
+				:rich-parameters="richParameters"
+				:timestamp="timestamp"
+				:is-deleting="isDeleting"
+				:is-temporary="isTemporary"
+				:has-call="conversation.hasCall"
+				:sending-failure="sendingFailure"
+				:show-common-read-icon="showCommonReadIcon"
+				:common-read-icon-tooltip="commonReadIconTooltip"
+				:show-sent-icon="showSentIcon"
+				:sent-icon-tooltip="sentIconTooltip" />
```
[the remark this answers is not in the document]
Antreesy: This is the minimum :see_no_evil:
`message` here is a string responsible for text, not the object itself

src/components/MessagesList/MessagesGroup/MessagesGroup.vue
```diff
@@ -32,7 +32,7 @@
 		<ul class="messages">
 			<li class="messages__author" aria-level="4">
 				{{ actorDisplayName }}
-				<div v-if="lastEditActorDisplayName">
+				<div v-if="lastEditTimestamp">
```
DorraJaouad, line 35: `lastEditActorDisplayName` is more important to check here

src/components/MessagesList/MessagesGroup/MessagesGroup.vue
```diff
@@ -155,16 +131,18 @@ export default {
 		},
 
 		getLastEditor() {
-			if (this.lastEditActorId === this.actorId && this.lastEditActorType === this.actorType) {
+			if (!this.lastEditTimestamp) {
+				return ''
```
DorraJaouad, line 135: And you are already have a check for `lastEditTimestamp` here

src/components/MessagesList/MessagesGroup/MessagesGroup.vue
Antreesy: I don't want to rely on the displayName, because it could be changed to empty stting -> evaluated to `false`. In that sense, timestamp is more reliable

Antreesy: Although the lastEdit parameters haven't changed for deleted user, but that's on a backend side:
image

DorraJaouad (approved):

PR #11366, opened 2024-01-10
Fix(BreakoutRoomsEditor): Add amount validation

### LINKED ISSUE #11355: Incorrect handling of invalid breakout rooms number
### How to use GitHub

---

## Steps to reproduce
1. In `Configure breakout rooms` dialog enter value outside of allowed range 1-20 (i.e. `0` or `21`).
2. Press `Create rooms`.

### Expected behaviour

Web UI should clearly display allowed rooms range and `Number of breakout rooms` field shouldn't allow entering values outside of allowed range (defined in backend see below).

In case of breakout rooms creating error, no breakout rooms should be created.

### Actual behaviour
Web UI allows `Number of breakout rooms` values outside of range enforced in backend with
```public const MINIMUM_ROOM_AMOUNT = 1;
public const MAXIMUM_ROOM_AMOUNT = 20;
```

When user enters value outside of this range i.e. `0` or `21` and presses `Create rooms`, an error
```
An error occurred while creating breakout rooms
```
is thrown but 1 breakout room is created (and must be removed to allow one to setup correct number of breakout rooms).

## Talk app

v18.0.1

## Browser

Firefox ESR.

DorraJaouad:
### ☑️ Resolves

* Fix #11355

Hint is conditioned because the user will likely pick an amount from 1 to 20, but it will be shown in case they attempt to choose outside that range.

| 🏚️ Before | 🏡 After |
|------------|----------|
| image      | image     |

### 🚧 Tasks

- [ ] Code review
- [ ] Visual review

### 🏁 Checklist

### DIFF
--- a/src/components/BreakoutRoomsEditor/BreakoutRoomsEditor.vue
+++ b/src/components/BreakoutRoomsEditor/BreakoutRoomsEditor.vue
@@ -30,7 +30,11 @@
 			<template v-if="!isEditingParticipants">
 				<div class="breakout-rooms-editor__main">
 					<label class="breakout-rooms-editor__caption" for="room-number">{{ t('spreed', 'Number of breakout rooms') }} </label>
+					<p v-if="isInvalidAmount" class="breakout-rooms-editor__error-hint">
+						{{ t('spreed', 'You can create from 1 to 20 breakout rooms.') }}
+					</p>
 					<NcTextField id="room-number"
+						ref="textField"
 						:value="amount.toString()"
 						class="breakout-rooms-editor__number-input"
 						type="number"
@@ -61,10 +65,16 @@
 					</fieldset>
 				</div>
 				<div class="breakout-rooms-editor__buttons">
-					<NcButton v-if="mode === '2'" type="primary" @click="isEditingParticipants = true">
+					<NcButton v-if="mode === '2'"
+						type="primary"
+						:disabled="isInvalidAmount"
+						@click="isEditingParticipants = true">
 						{{ t('spreed', 'Assign participants to rooms') }}
 					</NcButton>
-					<NcButton v-else type="primary" @click="handleCreateRooms">
+					<NcButton v-else
+						type="primary"
+						:disabled="isInvalidAmount"
+						@click="handleCreateRooms">
 						{{ t('spreed', 'Create rooms') }}
 					</NcButton>
 				</div>
@@ -114,6 +124,7 @@ export default {
 			amount: 2,
 			attendeeMap: '',
 			isEditingParticipants: false,
+			isInvalidAmount: false,
 		}
 	},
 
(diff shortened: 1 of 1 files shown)


src/components/BreakoutRoomsEditor/BreakoutRoomsEditor.vue
```diff
@@ -30,6 +30,9 @@
 			<template v-if="!isEditingParticipants">
 				<div class="breakout-rooms-editor__main">
 					<label class="breakout-rooms-editor__caption" for="room-number">{{ t('spreed', 'Number of breakout rooms') }} </label>
+					<p v-if="isInvalidAmount" class="breakout-rooms-editor__error-hint">
+						{{ t('spreed', 'You can create up to 20 breakout rooms.') }}
+					</p>
 					<NcTextField id="room-number"
```
nickvergessen, line 36: shouldn't the min/max from this component already help with that?

Antreesy: With `min`/`max` we can use a native method `input.checkValidity()`, which returns `true`/`false` in case amount is out of range. But that is not implemented upstream in the vue-library, there I believe it belongs

nickvergessen: yeah that's what I mean. The component should handle this already

DorraJaouad: Yeap, once it is fixed from the upstream we can just replace the boolean `this.isInvalidAmount` with output of the native `checkValidity()` because we still need it for disabling the buttons and maybe keeping the hint 😁.

src/components/BreakoutRoomsEditor/BreakoutRoomsEditor.vue
```diff
@@ -147,6 +157,7 @@ export default {
 		// now it breaks validation. Another option: Create NcNumberField component
 		setAmount(value) {
 			this.amount = parseFloat(value)
+			this.isInvalidAmount = this.amount > 20 || this.amount < 1
```
Antreesy, line 160:
```suggestion
			this.isInvalidAmount = isNaN(this.amount) || !this.$refs.textField.$refs.inputField.$refs.input?.checkValidity()
```
Maybe a bit ugly, but it checks, if input is not empty and not out of range of provided `min`/`max`

Antreesy (commented):
After checking the upstream code, it now make  a little sense for me to literally upstream it :grimacing: 

We still could do at the place (with refs and without pure JS) but to expose `checkValidity()` from lib is not as needed as to expose `focus()`, for example.
In addition to that, NcInput is simple visual / low-functional component, so again, I don't see any point in making it more complicate upstream

src/components/BreakoutRoomsEditor/BreakoutRoomsEditor.vue
DorraJaouad: Literally the same validation as the current one ?

Antreesy: You just got lucky, that `NaN < 1` =)
It's safer, and you define max/min ranges only at the one place (component attrs)

Antreesy (approved):

PR #11370, opened 2024-01-11
fix(messagesStore): avoid usage of `message.token` 

Antreesy:
### ☑️ Resolves

* Fix prior to federation:
  * federated conversation has `token: a` and `remoteToken: b`. When we're fetching something from remote server, we're using `b`, and messages are coming with `token: b` in the body, which doesn't belong to current instance
  * forcing to use local token for all messages actions / mutations
  * remove all extra payload from request (don't pass the whole message, if only token/id is needed)

No visual changes

### 🚧 Tasks

- [x] Manual testing
- [x] Code review
- [ ] Follow-up: migrate store to Pinia (not this PR)

### 🏁 Checklist

### DIFF
--- a/src/store/fileUploadStore.js
+++ b/src/store/fileUploadStore.js
@@ -298,11 +298,12 @@ const actions = {
 	 * @param {object} context.getters the contexts getters object.
 	 * @param {object} context.state the contexts state object.
 	 * @param {object} data the wrapping object
+	 * @param {string} data.token The conversation token
 	 * @param {string} data.uploadId The unique uploadId
 	 * @param {string|null} data.caption The text caption to the media
 	 * @param {object|null} data.options The share options
 	 */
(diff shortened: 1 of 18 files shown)


src/store/fileUploadStore.js
```diff
@@ -321,7 +321,7 @@ const actions = {
 				message: index === lastIndex && caption ? caption : '{file}',
 			}
 			// Add temporary messages (files) to the messages list
-			dispatch('addTemporaryMessage', temporaryMessage)
+			dispatch('addTemporaryMessage', { token: temporaryMessage.token, message: temporaryMessage })
```
DorraJaouad, line 324: Here in `uploadFiles` you kept using token from message but for `retryUploadFiles` , you switched to `getToken` ? 
Pick one :p

DorraJaouad (approved):
Tested on messages, file uploads and breakouts rooms.

PR #11371, opened 2024-01-11
fix(chat): Handle basic editing of captions

nickvergessen:
### ☑️ Resolves

* Ref  #11206 

## 🛠️ API Checklist

- [x] Handle captions
- [x] Prevent empty messages

### 🏁 Checklist

- [x] ⛑️ Tests (unit and/or integration) are included or not possible
- [x] 📘 API documentation in `docs/` has been updated or is not required
- [x] 🔖 Capability is added or not needed

PR #11374, opened 2024-01-11
Feat(messages): Edit messages

### LINKED ISSUE #11205: ✏️ Edit message - Frontend
- See overview at https://github.com/nextcloud/spreed/issues/11203

Feel free to copy tasks from the overview where applicable for your todos

DorraJaouad:
### ☑️ Resolves

* Fix #11205

- [ ] Edit in default input showing the original content as a "quote"
    - [ ] Don't allow changing from message to caption and vice-versa
    - [ ] Don't allow to toggle silent send on
- [ ] System message hidden like with delete?
- [ ] Visible indicator a message was edited (and by whom/when)?
    - [ ] Amend to the author name ` (edited)` / ` (edited by moderator-name)`
    - [ ] Editor+Time in Message menu as subline of timestamp
- [ ] Time limitation like for delete?
    - [ ] Limit editing to X hours (draft 24 hours, check what others do and discuss again)
    - [ ] Set the limit on deletion to infinite
- [ ] What to do about mention notifications (added a mention, removed a mention)?
    - [ ] Should indicate in the UI what happens (only when affected)
    - [ ] Make sure people don't get 1 for the message and 2 for followup edits
- [ ] What about bots?
    - [ ] Can not edit bot messages, moderators can
    - [ ] Bots will not get explicitly notified about previous edits (only with the default system message)?
    - [ ] If a message triggered a bot already it's yolo and will not be retriggered with the new content automatically (bot can do based on the system message?)
        - [ ] Should indicate in the UI what happens (only when affected)

| 🏚️ Before | 🏡 After |
|------------|----------|
| B          | A        |

### 🚧 Tasks

- [ ] ...
- [ ] ...

### 🏁 Checklist


src/components/NewMessage/NewMessage.vue
```diff
@@ -644,14 +645,24 @@ export default {
 		},
 
 		async handleEdit() {
+
+			if (this.messageToEdit.messageParameters.length !== 0) {
```
nickvergessen, line 649: this also matches mentions :-P

src/components/NewMessage/NewMessage.vue
```diff
@@ -644,14 +645,24 @@ export default {
...
+				// Captions editing is not supported yet
```
nickvergessen, line 650: But just remove it, it's coming before the final: https://github.com/nextcloud/spreed/pull/11371 (ready for review actually)

src/components/NewMessage/NewMessage.vue
```diff
@@ -644,14 +645,24 @@ export default {
...
+				return
+			}
+
+			// Submitting an empty message will abort the edit
+			if (this.text.trim() === '') {
+				this.chatExtrasStore.removeMessageIdToEdit(this.token)
```
nickvergessen, line 656: return after this?

src/components/NewMessage/NewMessage.vue
DorraJaouad: As the PR is still draft, I just noted it there in to implement it after that API PR because I saw it ready for review.

src/components/NewMessage/NewMessage.vue
DorraJaouad: still didn't make my last scan on the code :')

src/components/NewMessage/NewMessage.vue
nickvergessen: It's merged now :)

src/components/NewMessage/NewMessage.vue
```diff
@@ -408,7 +433,11 @@ export default {
 
 		isMobileDevice() {
 			return /Android|iPhone|iPad|iPod/i.test(navigator.userAgent)
-		}
+		},
+
+		chatInput() {
```
Antreesy, line 438: If i remember, we should be able to come back to typing normal message after editing something
so we need to have in store state for each: `chatInput` and `chatEditInput`

src/components/Quote.vue
```diff
@@ -249,8 +257,13 @@ export default {
 		},
 	},
 	methods: {
-		handleAbortReply() {
-			this.chatExtrasStore.removeParentIdToReply(this.token)
+		handleAbort() {
+			if (this.editMessage) {
+				this.chatExtrasStore.removeMessageIdToEdit(this.token)
+				this.chatExtrasStore.removeChatInput(this.token)
```
Antreesy, line 263: What i said earlier, we should switch to original chatInput, if it's presented

src/services/messagesService.js
```diff
@@ -178,6 +178,12 @@ const getReactionsDetails = async function(token, messageId) {
 	return axios.get(generateOcsUrl('apps/spreed/api/v1/reaction/{token}/{messageId}', { token, messageId }))
 }
 
+const editMessage = async function({ token, messageId, updatedMessage }) {
+	return axios.put(generateOcsUrl('apps/spreed/api/v1/chat/{token}/{messageId}', { token, messageId }), {
+		message: updatedMessage,
+	})
```
Antreesy, line 184:
```suggestion
const editMessage = async function({ token, messageId, updatedMessage }, options) {
	return axios.put(generateOcsUrl('apps/spreed/api/v1/chat/{token}/{messageId}', { token, messageId }), {
		message: updatedMessage,
	}, options)
```
Could you please provide an options prop? It has to be added for all axios requests to be able to talk to remote servers

src/store/messagesStore.js
```diff
@@ -542,6 +551,31 @@ const actions = {
 			return
 		}
 
+		if (message.parent && message.systemMessage === 'message_edited') {
```
Antreesy, line 554: Maybe we include this code to the block above, to not duplicate lines?

src/store/messagesStore.js
```diff
@@ -542,6 +551,31 @@ const actions = {
...
+			// If parent message is presented in store already, we update it
+			const parentInStore = context.getters.message(message.token, message.parent.id)
+			if (Object.keys(parentInStore).length !== 0) {
+				context.commit('addMessage', message.parent)
+				return
+			}
+
+			// Update message in store
+			const messageUpdated = message.parent
```
Antreesy, line 563: This code below shouldn't be needed, as you just updated parent in the store with all these parameters

Antreesy (commented):
Couple of remarks, while you're still working on it

src/components/MessagesList/MessagesGroup/MessagesGroup.vue
```diff
@@ -32,6 +32,11 @@
 		<ul class="messages">
 			<li class="messages__author" aria-level="4">
 				{{ actorDisplayName }}
+				<div v-if="lastEditActorDisplayName"
+					:aria-label="getLastEditor"
+					:aria-level="4">
+					{{ getLastEditor }}
+				</div>
```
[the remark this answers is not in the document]
DorraJaouad: It's for a11y

src/components/MessagesList/MessagesList.vue
```diff
@@ -358,6 +358,7 @@ export default {
 
 		areGroupsIdentical(group1, group2) {
 			if (group1.messages.length !== group2.messages.length
+				|| JSON.stringify(group1.messages) !== JSON.stringify(group2.messages)
```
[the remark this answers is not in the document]
DorraJaouad: So if there are 3 messages in a group and the second message was edited, it should split the group like
group 1 : message 1
group 2 : message 2 and message 3 ?

or this would split into 3 groups ? 

I think computation cost would still be considered reasonable as there are usually 1 to 10 messages per group but if the idea was to split edited message to new group, then we can skip this comparison.

src/components/NewMessage/NewMessage.vue
```diff
@@ -85,6 +85,9 @@
 				<div v-if="parentMessage" class="new-message-form__quote">
 					<Quote is-new-message-quote v-bind="parentMessage" />
 				</div>
+				<div v-if="messageToEdit" class="new-message-form__quote">
+					<Quote is-new-message-quote edit-message v-bind="messageToEdit" />
```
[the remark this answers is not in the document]
DorraJaouad: I already made a priority in Quote that whatever was updated last should be shown so if you click on reply and you are editing message, reply will be shown instead and vice versa. In other messengers, they do show a pop for confirmation.

src/components/MessagesList/MessagesGroup/MessagesGroup.vue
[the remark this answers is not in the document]
Antreesy: `aria-label` is for interactive elements, here it's just text. And it's already inside the `li` with `aria-level=4`

src/components/NewMessage/NewMessage.vue
```diff
@@ -420,6 +449,26 @@ export default {
...
+		},
+
+		parentMessage(newValue) {
+			if (newValue) {
+				this.chatExtrasStore.removeMessageIdToEdit(this.token)
+				this.chatExtrasStore.removeChatInput(this.token)
+			}
+		},
+
+		chatInput(newValue) {
```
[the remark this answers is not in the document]
DorraJaouad: Partially overlapping ( chatInput watcher is more covering) :
- Token is only changed once if we are talking about opening a given conversation and chatInput can be changed after that ( when editing a message for example or aborting the edit while token was never changed)

chatInput watcher would remain with removing the text overwriting from token watcher

src/components/MessagesList/MessagesList.vue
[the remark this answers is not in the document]
Antreesy: Should be two groups, as in your example :+1:

src/components/NewMessage/NewMessage.vue
DorraJaouad: Yes, it makes sense, I will add it in the description.

src/store/messagesStore.js
DorraJaouad: I thought about it and I put it separately for the sake of clarity. I will translate it there then.

src/components/NewMessage/NewMessage.vue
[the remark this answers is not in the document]
Antreesy: When token is updated, computed `chatInput` will be updated as well accordingly. So we may remove some lines from `token` watcher, I'd say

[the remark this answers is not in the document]
DorraJaouad: this would be outdated as I will add another state for the edit input

src/components/MessagesList/MessagesList.vue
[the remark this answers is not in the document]
nickvergessen: I think having 3 groups would also be okay, so it's clear that the 3rd messages was not edited by the other person.

And then basically the rule would be `if message is edited, never group`

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
```diff
@@ -207,6 +236,24 @@
 						{{ t('spreed', 'Set custom reminder') }}
 					</NcActionButton>
 				</template>
+
+				<template v-else-if="submenu === 'edit-history'">
```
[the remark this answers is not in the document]
DorraJaouad: Editor+Time could not fit in one line (at least not visually appealing) so I switched to a submenu.

src/components/MessagesList/MessagesGroup/MessagesGroup.vue
```diff
@@ -133,6 +141,18 @@ export default {
...
 		},
 
+		getLastEditor() {
+			if (this.lastEditActorId === this.actorId && this.lastEditActorType === this.actorType) {
+				return t('spreed', '(edited)')
+			} else if (this.lastEditActorId === this.$store.getters.getActorId()
+						&& this.lastEditActorType === this.$store.getters.getActorType()) {
+				return t('spreed', '(edited by you)')
+			} else {
+				return t('spreed', '(edited by {user1})', { user1: this.lastEditActorDisplayName })
```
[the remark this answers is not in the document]
DorraJaouad: `'moderator'`  will make more sense actually 😄

src/components/Quote.vue
```diff
@@ -40,6 +40,9 @@ components.
 					:size="AVATAR.SIZE.EXTRA_SMALL"
 					disable-menu />
 				{{ getDisplayName }}
+				<div v-if="editMessage">
+					{{ t('spreed', '(editing)') }}
+				</div>
```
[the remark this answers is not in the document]
DorraJaouad: For the editing hint, it is also possible to replace with a pencil icon. 
Maybe move the 'x' next to the check mark and add a title?

src/components/NewMessage/NewMessage.vue
```diff
@@ -82,8 +82,8 @@
 						</template>
 					</NcButton>
 				</div>
-				<div v-if="parentMessage" class="new-message-form__quote">
-					<Quote is-new-message-quote v-bind="parentMessage" />
+				<div v-if="parentMessage || messageToEdit" class="new-message-form__quote">
+					<Quote is-new-message-quote :edit-message="!!messageToEdit" v-bind="messageToEdit ?? parentMessage" />
```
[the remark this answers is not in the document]
DorraJaouad: It is used to display the 'x' button, if we are omitting this as suggested above, we can set it to false for Editing.

src/components/ChatView.vue
```diff
@@ -161,12 +162,23 @@ export default {
...
+		EventBus.$on('editing-message', () => {
+			this.isEditingMessage = true
+		})
+		EventBus.$on('message-edited', () => {
+			this.isEditingMessage = false
+		})
+	},
+
+	beforeDestroy() {
+		EventBus.$off('editing-message')
```
Antreesy, line 174:
```suggestion
		EventBus.$off('message-editing')
```
let's keep one style for naming

src/components/NewMessage/NewMessage.vue
```diff
@@ -595,6 +655,20 @@ export default {
...
 		},
 
+		async handleEdit() {
+			try {
+				await editMessage({
+					token: this.token,
+					messageId: this.messageToEdit.id,
+					updatedMessage: this.text.trim()
+				})
+				EventBus.$emit('message-edited', this.messageToEdit.id)
```
Antreesy, line 665: ~~Make sure to emit something in case of fail, to switch on drag-n-drop again ~~

UPD: I'd rely on a store here

src/components/ChatView.vue
Antreesy: Or isn't it better to rely on chatExtrasStore? if some value is defined there, we disable drag-n-drop

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
```diff
@@ -560,6 +628,10 @@ export default {
 			return moment(this.timestamp * 1000).format('lll')
 		},
 
+		editedDateTime() {
+			return moment(this.messageObject.lastEditTimestamp * 1000).format('lll')
```
Antreesy, line 632: _Note:_
We'll need to run through these files after one of the PR is merged to avoid conflicts

src/components/MessagesList/MessagesList.vue
```diff
@@ -284,6 +284,7 @@ export default {
 		EventBus.$on('scroll-chat-to-bottom-if-sticky', this.scrollToBottomIfSticky)
 		EventBus.$on('focus-message', this.focusMessage)
 		EventBus.$on('route-change', this.onRouteChange)
+		EventBus.$on('message-edited', this.handleMessageEdited)
```
Antreesy, line 287: Could also watch after store getter instead of EventBus here

src/components/NewMessage/NewMessage.vue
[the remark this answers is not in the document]
Antreesy: Or we rename it to `canCancel` or `canAbort`, which'll make more sense here

src/services/messagesService.js
```diff
@@ -178,6 +178,12 @@ const getReactionsDetails = async function(token, messageId) {
 	return axios.get(generateOcsUrl('apps/spreed/api/v1/reaction/{token}/{messageId}', { token, messageId }))
 }
 
+const editMessage = async function({ token, messageId, updatedMessage }, options) {
```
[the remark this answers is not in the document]
DorraJaouad: This will create confusion at some point because editing messages touches both the original message and the new one so it's better to indicate which one is being passed (although it may seem obvious for now)

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
DorraJaouad: Actual note: Who will deal with conflicts 😛 ?

src/components/MessagesList/MessagesList.vue
DorraJaouad: It might be tricky to watch over a store getter which is probably `messageIdToEdit` because the id can also be removed upon abort too or start a reply message. For the sake of clarity, let's keep the event :3 .

src/components/ChatView.vue
DorraJaouad: This is though can be similarly be a getter watcher 👍🏻

src/components/MessagesList/MessagesList.vue
Antreesy: Or just discard an editing process, because replying is a normal messaging, so you're basically switching from editing. I'm not sure someone is expecting to continue to edit message after sending another one.

src/components/NewMessage/NewMessage.vue
```diff
@@ -595,6 +673,20 @@ export default {
 			}
 		},
 
+		async handleEdit() {
+			try {
+				await editMessage({
+					token: this.token,
+					messageId: this.messageToEdit.id,
+					updatedMessage: this.text.trim()
+				})
```
Antreesy, line 682:
```suggestion
				const response = await editMessage({
					token: this.token,
					messageId: this.messageToEdit.id,
					updatedMessage: this.text.trim()
				})
				this.$store.dispatch('processMessage', { token: this.token, message: response.data.ocs.data })
```
If we patching the store here, it saves us from boilerplate in the MessagesList

src/components/MessagesList/MessagesList.vue
```diff
@@ -627,6 +634,57 @@ export default {
 			}
 		},
 
+		handleMessageEdited(id) {
+			// regroup messagesGroupedByAuthor when a message is edited
```
Antreesy, line 638: ... and a one-liner here:
`this.messagesGroupedByAuthor = this.prepareMessagesGroups(this.messagesList)`

src/components/MessagesList/MessagesList.vue
Antreesy: It makes sense for me now, beacuse we also need to look for external edits

src/components/NewMessage/NewMessage.vue
```diff
@@ -595,6 +673,20 @@ export default {
...
+				EventBus.$emit('message-edited', this.messageToEdit.id)
```
Antreesy, line 683: We need to move it to the messagesStore

src/store/messagesStore.js
```diff
@@ -529,11 +529,16 @@ const actions = {
...
 				|| message.systemMessage === 'reaction_deleted'
-				|| message.systemMessage === 'reaction_revoked')) {
+				|| message.systemMessage === 'reaction_revoked'
+				|| message.systemMessage === 'message_edited')) {
 			// If parent message is presented in store already, we update it
 			const parentInStore = context.getters.message(token, message.parent.id)
 			if (Object.keys(parentInStore).length !== 0) {
 				context.commit('addMessage', { token, message: message.parent })
+				if (message.systemMessage === 'message_edited') {
+					// End of process for edited messages
```
Antreesy, line 539:
```suggestion
					EventBus.$emit('message-edited')
```

This will trigger it for your messages and for external ones

src/store/messagesStore.js
Antreesy: It's still there

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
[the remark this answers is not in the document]
nickvergessen: The thing is a submenu is quit "far away" for the little info (last edit only, no full history, no diff) you get.
Therefore I would vote for the solution outlined above

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
```diff
@@ -63,7 +64,20 @@
...
+					<NcActionButtonGroup v-if="messageObject.lastEditTimestamp">
+						<NcActionText>
+							<template #icon>
+								<ClockEditOutline :size="16" />
+							</template>
+							{{ messageObject.lastEditActorDisplayName }}
+						</NcActionText>
+						<NcActionText>
+							{{ editedDateTime }}
+						</NcActionText>
```
Antreesy, line 78:
```suggestion
						<NcActionText :name="messageObject.lastEditActorDisplayName">
							<template #icon>
								<ClockEditOutline :size="16" />
							{{ editedDateTime }}
						</NcActionText>
```
I guess, that was Joas' idea 
image

src/components/NewMessage/NewMessage.vue
```diff
@@ -382,11 +421,11 @@ export default {
 		},
 
 		showAttachmentsMenu() {
-			return this.canShareFiles && !this.broadcast && !this.upload
+			return this.canShareFiles && !this.broadcast && !this.upload && !this.messageToEdit
```
Antreesy, line 424: It would be good to just disable button, when to hide it. less jumping on the screen

`<NcActions :disabled="this.messageToEdit" />` should do the trick

src/store/messagesStore.js
Antreesy: Don't forget to remove it

Antreesy (approved):
So far good

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
DorraJaouad: Bold is a bit peculiar to the message buttons style :/

PR #11395, opened 2024-01-17
feat(meetings): selectively invite users from lobby

### LINKED ISSUE #8601: Direct option to allow a user to join while the lobby is enabled
> Well the feature is there. You enable the lobby and assign one-by-one the permission to ignore the lobby.
Maybe it's a bit too hidden and we should outline a bit more, who is currently blocked by the lobby and then have a direct button which just adds that permission for them.

_Originally posted by @nickvergessen in https://github.com/nextcloud/spreed/issues/5394#issuecomment-1398157614_

Antreesy:
### ☑️ Resolves

* Fix #8601
* If lobby is enabled, a new button is appeare in participant action to explicitly change lobby permission for that one participant
* `isInLobby` was removed from `chatIdentifier`:
  * chat is mounted/unmounted with param change anyway, so no need to depend on it
  * it leaves outdated cancel requests in the store:
image

>[!IMPORTANT]
> HPB is required to be enabled, to receive signaling messages

In | Out
--|--
image | image

https://github.com/nextcloud/spreed/assets/93392545/fc5ae076-b86d-42c2-98fe-ca038a52e09e

### 🚧 Tasks

- [x] Check wording
- [x] Decide on selected icons

### 🏁 Checklist

### DIFF
--- a/src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
+++ b/src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
@@ -60,6 +60,7 @@
 				<span v-if="showModeratorLabel" class="participant-row__moderator-indicator">({{ t('spreed', 'moderator') }})</span>
 				<span v-if="isBridgeBotUser" class="participant-row__moderator-indicator">({{ t('spreed', 'bot') }})</span>
 				<span v-if="isGuest" class="participant-row__guest-indicator">({{ t('spreed', 'guest') }})</span>
+				<span v-if="!isSelf && isLobbyEnabled && !canSkipLobby" class="participant-row__guest-indicator">({{ t('spreed', 'in the lobby') }})</span>
 			</div>
 
 			<!-- Second line: participant status message if applicable -->
@@ -123,7 +124,8 @@
 		<NcActions v-if="(canBeModerated || canSendCallNotification) && !isSearched"
 			:container="container"
 			:aria-label="participantSettingsAriaLabel"
-			force-menu
+			:inline="showToggleLobbyAction ? 1 : 0"
+			:force-menu="!showToggleLobbyAction"
 			placement="bottom-end"
 			class="participant-row__actions">
 			<template #icon>
@@ -144,6 +146,26 @@
 				</template>
 				{{ attendeePin }}
 			</NcActionText>
+			<!-- Grant or revoke lobby permissions (inline button) -->
+			<template v-if="showToggleLobbyAction">
+				<NcActionButton v-if="canSkipLobby"
+					close-after-click
+					@click="setLobbyPermission(false)">
+					<template #icon>
+						<AccountMinusIcon :size="20" />
+					</template>
+					{{ t('spreed', 'Move back to lobby') }}
+				</NcActionButton>
+				<NcActionButton v-else
+					close-after-click
+					@click="setLobbyPermission(true)">
+					<template #icon>
+						<AccountPlusIcon :size="20" />
+					</template>
+					{{ t('spreed', 'Move to conversation') }}
+				</NcActionButton>
+			</template>
+			<!-- Grant or revoke moderator permissions -->
 			<NcActionButton v-if="canBeDemoted"
 				close-after-click
 				@click="demoteFromModerator">
@@ -292,6 +314,8 @@
 import isEqual from 'lodash/isEqual.js'
 
 import Account from 'vue-material-design-icons/Account.vue'
+import AccountMinusIcon from 'vue-material-design-icons/AccountMinus.vue'
+import AccountPlusIcon from 'vue-material-design-icons/AccountPlus.vue'
 import Bell from 'vue-material-design-icons/Bell.vue'
 import ContentCopy from 'vue-material-design-icons/ContentCopy.vue'
 import Crown from 'vue-material-design-icons/Crown.vue'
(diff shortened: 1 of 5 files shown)


nickvergessen:
I would say it should be even more visible? Directly show it on the participant list next to the ...?

src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
```diff
@@ -935,6 +965,23 @@ export default {
 			}
 		},
 
+		async setLobbyPermission(value) {
+			// Bitwise add or subtract lobby permission for a participant
```
nickvergessen, line 969: Now you did it already, but the API has the option to add/remove single permissions, so you wouldn't need to do this math yourself 😅👍

Antreesy: It could be better to pass a single option without that arithmetic :see_no_evil:

nickvergessen: Then use method=set and only set the permission you want to toggle:
https://nextcloud-talk.readthedocs.io/en/latest/participant/#set-permissions-for-an-attendee

Antreesy:
> I would say it should be even more visible?

Updated code, icons (unlock/lock for account fit more IMO) and screenshots

nickvergessen:
Not sure about the wording of the buttons. cc @jancborchardt for better ideas

jancborchardt:
Improved wording could be:
- Move to conversation
- Move back to lobby

"Move" seems more fitting than "invite" since it is a direct action, right?

Antreesy:
> "Move" seems more fitting than "invite" since it is a direct action, right?

Agree. Maybe `Move user` would be more explicable, but I'm fine with short option too
Participant is already in the conversation, but can't think of other wording, so could proceed with that.

Any thoughts over icons used?

src/components/RightSidebar/Participants/ParticipantsTab.vue
```diff
@@ -203,18 +206,35 @@ export default {
 
 	beforeMount() {
 		EventBus.$on('route-change', this.abortSearch)
+		EventBus.$on('signaling-users-changed', this.updateUsers)
```
DorraJaouad, line 209: Shouldn't `useGetParticipants` handle the update on `'signaling-participant-list-changed'` already ?

src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
```diff
@@ -144,6 +146,26 @@
...
+					<template #icon>
+						<RoomService :size="20" />
+					</template>
+					{{ t('spreed', 'Move back to lobby') }}
+				</NcActionButton>
+				<NcActionButton v-else
+					close-after-click
+					@click="setLobbyPermission(true)">
+					<template #icon>
+						<AccountKeyIcon :size="20" />
```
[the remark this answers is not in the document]
DorraJaouad: IDK, sthg like this
https://pictogrammers.com/library/mdi/icon/door-open/

src/components/RightSidebar/Participants/ParticipantsTab.vue
Antreesy: It doesn't have the required payload: 
https://github.com/nextcloud/spreed/blob/0c7ed6776f1f129e3a1f16794f97c3a1be7f8e85/src/utils/signaling.js#L1514-L1518

src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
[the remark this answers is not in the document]
jancborchardt: Good point. Simple symmetric icons would be good here:
- https://pictogrammers.com/library/mdi/icon/account-plus/
- https://pictogrammers.com/library/mdi/icon/account-remove/ (or account-minus)

jancborchardt:
> > "Move" seems more fitting than "invite" since it is a direct action, right?
> 
> Agree. Maybe `Move user` would be more explicable, but I'm fine with short option too
> Participant is already in the conversation, but can't think of other wording, so could proceed with that.

I would use the short wording. We want to phase out the wording "user", and the text is long enough already. :)


> Any thoughts over icons used?

Yep, wrote as reply in @DorraJaouad's code comment.

nickvergessen:
> We want to phase out the wording "user", and the text is long enough already. :)

Also note this is not only about "users", but also about "guests".
That's why we use the word "participant" e.g. on the remove option. Should use it here too.

Antreesy:
>  Should use it here too.

Or not use it at all, to keep it short (same is done for 'Promote / demote to moderator')

src/components/RightSidebar/Participants/ParticipantsList/Participant/Participant.vue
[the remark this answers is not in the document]
Antreesy: I was concerned, that +- symbol is quite small to distinguish and will blend with others. But combined with participants sorting, invited are always on top, so seems ok

image

src/components/RightSidebar/Participants/ParticipantsTab.vue
DorraJaouad: Fetch requests are mostly in getParticipants composable. Also, there will be duplicated requests when moving to lobby ( request from this method and the other from TopBar's composable).

You need the data only to check if the fetch is needed. However, itis better improvement than only doing from composable because it will not trigger fetch on every user change but only self. 

So I think, this is the best option so far...

Antreesy: > Also, there will be duplicated requests when moving to lobby 

There won't be, as TopBar is unmounted at the same time, and won't make any new requests. Even this last one won't be triggered, because composable has an internal check for `isInLobby`
For the `mounted`, we didn't consider that variant before for composable, but it shouldn't be triggered before user is joined conversation => should be fine as-is, i believe

PR #11396, opened 2024-01-17
fix(breakout): stop breakout rooms on removing

Antreesy:
### ☑️ Resolves

* API part is splitted out from #11394 
* When removing breakout rooms, status at parent is now reset (if create rooms again, they will not be marked as started)
* Updating lastActivity when requesting / dismissing the assistance allows moderators to re-fetch breakout rooms within 30s interval (`?modifiedSince`) and actually see the request (before it was once in 10 requests on a full fetch)

### 🚧 Tasks

- [x] Write tests

### 🏁 Checklist

### DIFF
--- a/lib/Service/BreakoutRoomService.php
+++ b/lib/Service/BreakoutRoomService.php
@@ -36,6 +36,7 @@ use OCA\Talk\Model\BreakoutRoom;
 use OCA\Talk\Participant;
 use OCA\Talk\Room;
 use OCA\Talk\Webinary;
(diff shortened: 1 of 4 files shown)


nickvergessen (approved):

Antreesy:
Tested against frontend :green_circle:

Antreesy:
As something needs to be patched, could we additionaly make sure, that all methods:
- startBreakoutRooms
- stopBreakoutRooms
- removeBreakoutRooms
also update `lastActivity` for the given room?

PR #11426, opened 2024-01-22
fix(attachments): Don't allow selecting shared folders as attachment …

nickvergessen:
…folder

### 🏁 Checklist


Antreesy (approved):
Changes look OK.

Is it possible to trigger it manually for Talk? I tried with new users and `default_attachment_folfer`, but it just creates a normal folder instead of using shared one with the same name

PR #11436, opened 2024-01-24
fix(CallView): reduce computations amount

Antreesy:
### ☑️ Resolves

* Move and rearrange some computed props for the CallView
  * Remove unused v-for and v-if for promoted views (there is always only one model to handle)
  * Unite multiple redundant wrappers
  * One-time things moved to child component
  * loadState and getCapabilities moved from computed props (to call them once)
  * Prioritize selected LocalVideo over Screen

There's still a lot of things processing each time participants change something, but at least some unused computations were eliminated

No visual changes

### 🏁 Checklist

### DIFF
--- a/src/components/CallView/CallView.vue
+++ b/src/components/CallView/CallView.vue
@@ -35,21 +35,18 @@
 			<div id="videos">
 				<template v-if="!isGrid">
 					<!-- Selected video override mode -->
-					<div v-if="showSelectedVideo"
+					<div v-if="showSelectedVideo && selectedCallParticipantModel"
 						class="video__promoted selected-video"
 						:class="{'full-page': isOneToOne}">
-						<template v-for="callParticipantModel in reversedCallParticipantModels">
-							<VideoVue v-if="callParticipantModel.attributes.peerId === selectedVideoPeerId"
-								:key="callParticipantModel.attributes.selectedVideoPeerId"
-								:token="token"
-								:model="callParticipantModel"
-								:shared-data="sharedDatas[selectedVideoPeerId]"
-								:show-talking-highlight="false"
-								:is-grid="true"
-								:is-big="true"
-								:is-one-to-one="isOneToOne"
-								:fit-video="true" />
-						</template>
+						<VideoVue :key="selectedVideoPeerId"
+							:token="token"
+							:model="selectedCallParticipantModel"
+							:shared-data="sharedDatas[selectedVideoPeerId]"
+							:show-talking-highlight="false"
+							:is-one-to-one="isOneToOne"
+							is-grid
+							is-big
+							fit-video />
 					</div>
 					<!-- Screens -->
 					<div v-else-if="showLocalScreen || showRemoteScreen || showSelectedScreen" id="screens">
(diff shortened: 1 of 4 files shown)


src/components/CallView/CallView.vue
```diff
@@ -30,26 +30,23 @@
...
 		<template v-else>
-			<EmptyCallView v-if="!remoteParticipantsCount && !screenSharingActive && !isGrid" :is-sidebar="isSidebar" />
+			<EmptyCallView v-if="!callParticipantModels.length && !screenSharingActive && !isGrid" :is-sidebar="isSidebar" />
 
 			<div id="videos">
 				<template v-if="!isGrid">
 					<!-- Selected video override mode -->
 					<div v-if="showSelectedVideo"
 						class="video__promoted selected-video"
 						:class="{'full-page': isOneToOne}">
```
DorraJaouad, line 40: You can remove this div and add the classes in VideoVue

src/components/CallView/CallView.vue
```diff
@@ -301,26 +273,19 @@ export default {
...
-		gridVideosCapEnforced() {
-			return loadState('spreed', 'grid_videos_limit_enforced')
-		},
-
 		selectedVideoPeerId() {
 			return this.$store.getters.selectedVideoPeerId
 		},
 
+		selectedCallParticipantModel() {
+			if (!this.selectedVideoPeerId) {
```
DorraJaouad, line 281:
```suggestion
			if (!this.selectedVideoPeerId || !this.showSelectedVideo) {
```

src/components/CallView/CallView.vue
```diff
@@ -30,26 +30,23 @@
...
-						</template>
+						<VideoVue v-if="selectedCallParticipantModel"
+							:token="token"
+							:model="selectedCallParticipantModel"
+							:shared-data="sharedDatas[selectedVideoPeerId]"
+							:show-talking-highlight="false"
+							:is-grid="true"
+							:is-big="true"
+							:is-one-to-one="isOneToOne"
+							:fit-video="true" />
```
DorraJaouad, line 49:
```suggestion
							:show-talking-highlight="false"
							is-grid
							is-big
							:is-one-to-one="isOneToOne"
							fit-video />
```

You usually never miss these 🧐

src/components/CallView/CallView.vue
```diff
@@ -30,26 +30,23 @@
...
 					</div>
 					<!-- Screens -->
 					<div v-else-if="showLocalScreen || showRemoteScreen || showSelectedScreen" id="screens">
```
DorraJaouad, line 52: Please check if we used `id="screens"` somewhere. Otherwise, feel free to remove this div too

DorraJaouad (commented):
I liked these changes very much ⭐.  The component is clearer now too.

src/components/CallView/CallView.vue
Antreesy: only for styling, so replaced with a class. Still needed for Screen + presenter overlay components positioning though

src/components/CallView/CallView.vue
Antreesy: it has `position`, `width`, `height` in styles, which depends on html elements nesting. Won't touch it as part of this PR

DorraJaouad: Yes because it relies on the parent div with `id=videos` with position absolute, but there won't be any dependencies from the html elements children I think.

DorraJaouad (commented):
You messed with styles :/
image
image

DorraJaouad (approved):
🦅

PR #11448, opened 2024-01-25
feat(Poll): live polls pop-ups during the call

### LINKED ISSUE #11372: Make polls more visible when "live"
### How to use GitHub

---

During a webinar polls are a useful option to steer the conversation going forward, but only if people vote.

Popup the poll when they happen "live" directly so more people vote.

Mark the poll message visually so they are less likely to be missed

Antreesy:
### ☑️ Resolves

* Fix #11372
* Poll component is splitted into two:
  * `<Poll />` (card/button), to be shown as message part
  * `<PollViewer />` (dialog), detached and moved up on the `<ChatView />` and `<CallView />` level
  * With newly processed messages, dialog is triggered to be open when participant is in call
  * Necessary evil: components code is optimized 

Toast message:
image
Unvoted poll highlight:
image

Talk Main, call view:
Files sidebar (same appearance as with open chat, public share sidebars):

https://github.com/nextcloud/spreed/assets/93392545/d18ebaa5-5a25-4324-aa6b-5cc43dea18df

### 🚧 Tasks

- [ ] Follow-ups:
  - [ ] migrate pollsStore :pineapple: 

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/MessageBody.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/MessageBody.vue
@@ -28,7 +28,7 @@
 			:class="{
 				'system-message': isSystemMessage && !showJoinCallButton,
 				'deleted-message': isDeletedMessage,
-				'call-started': showJoinCallButton,
+				'message-highlighted': showJoinCallButton,
 			}">
 			<!-- Message content / text -->
 			<CancelIcon v-if="isDeletedMessage" :size="16" />
@@ -49,6 +49,7 @@
 		<!-- Normal message body content -->
 		<div v-else
 			class="message-main__text markdown-message"
+			:class="{'message-highlighted': isNewPollMessage }"
 			@mouseover="handleMarkdownMouseOver"
 			@mouseleave="handleMarkdownMouseLeave">
 			<!-- Replied parent message -->
@@ -284,6 +285,14 @@ export default {
 			return this.messageParameters?.file
 		},
 
+		isNewPollMessage() {
+			if (this.messageParameters?.object?.type !== 'talk-poll') {
+				return false
+			}
+
+			return this.isInCall && !!this.$store.getters.getNewPolls[this.messageParameters.object.id]
+		},
+
 		hideDate() {
 			return this.isTemporary || this.isDeleting || !!this.sendingFailure
 		},
@@ -442,7 +451,7 @@ export default {
 			padding: 0 20px;
 		}
 
-		&.call-started {
+		&.message-highlighted {
 			color: var(--color-text-light);
 			background-color: var(--color-primary-element-light);
 			padding: 10px;
--- a/src/components/PollViewer/PollViewer.vue
+++ b/src/components/PollViewer/PollViewer.vue
@@ -1,0 +1,484 @@
+<!--
+  - @copyright Copyright (c) 2024 Maksim Sukharev <antreesy.web@gmail.com>
+  -
+  - @author Marco Ambrosini <marcoambrosini@pm.me>
+  - @author Maksim Sukharev <antreesy.web@gmail.com>
+  -
+  - @license AGPL-3.0-or-later
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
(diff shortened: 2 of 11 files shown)


src/components/PollViewer/PollViewer.vue
```diff
@@ -0,0 +1,425 @@
...
+					<NcActionButton class="critical" @click="endPoll">
+						{{ t('spreed', 'End poll') }}
+						<template #icon>
+							<FileLock :size="20" />
+						</template>
+					</NcActionButton>
+				</NcActions>
+			</div>
+		</div>
+		<LoadingComponent v-else class="poll-modal__loading" />
```
DorraJaouad, line 104: Use this instead https://nextcloud-vue-components.netlify.app/#/Components/NcLoadingIcon
We will be probably removing that soon as we now have it from upstream.

src/components/PollViewer/PollViewer.vue
```diff
@@ -0,0 +1,425 @@
...
+		async submitVote() {
+			this.loading = true
+			await this.$store.dispatch('submitVote', {
+				token: this.token,
+				pollId: this.id,
+				vote: this.voteToSubmit.map(element => +element),
+			})
+			this.modalPage = 'results'
+			this.loading = false
+		},
```
DorraJaouad, line 307: Add error handling here. If there is an error, keep the voting page and showError

DorraJaouad: You can add error handling in other requests where it is needed too.

src/views/MainView.vue
```diff
@@ -9,6 +9,7 @@
 					<CallView :token="token" />
 				</template>
 			</TransitionWrapper>
+			<PollViewer />
```
DorraJaouad, line 12: Maybe just keep it in Poll component ?

Antreesy: What do you mean?
They've asked for a while to be separated, and we need the modal to be mounted on ChatView/CallView level to be able to use them in both

Antreesy:
cc @nextcloud/designers for the input on NC29 feature

marcoambrosini (changes requested):
I don't think that blocking the whole UI like this something we want to do. Joas suggested to make polls more visible in the issue, and I think that's a better solution. Dialogs that block the whole interface should only be shown when triggered by the user.

I will add this to my tasks and come up with a quick mockup soon :)

Antreesy:
What do you think of using a toast message, @marcoambrosini ? Seems less invasive

image

nickvergessen:
> What do you think of using a toast message

Sounds good for when the chat is closed.

But I would still color the system message in the chat view itself

src/components/MessagesList/MessagesGroup/Message/MessagePart/MessageBody.vue
```diff
@@ -284,6 +285,14 @@ export default {
...
 		},
 
+		isNewPollMessage() {
+			if (this.messageParameters?.object?.type !== 'talk-poll') {
+				return false
+			}
+
+			return !!this.$store.getters.getNewPolls[this.messageParameters.object.id]
+		},
+
```
DorraJaouad, line 295: Add a condition to check if it is in call. 
Highlighted polls are those that were not voted yet. If yes, then you should check against vote not against toasts because user can close a toast without voting.

src/components/PollViewer/PollViewer.vue
```diff
@@ -0,0 +1,440 @@
...
+			}
+		},
+
+		setVoteData() {
+			this.voteToSubmit = this.selfHasVoted
+				? this.poll?.votedSelf.map(el => el.toString())
+				: []
+		},
+
+		showPopup({ token, message }) {
```
DorraJaouad, line 291: it is `showToast` now

src/components/PollViewer/PollViewer.vue
```diff
@@ -0,0 +1,440 @@
...
+
+	watch: {
+		modalPage(value) {
+			if (value === 'voting') {
+				this.setVoteData()
+			}
+		},
+
+		id(value) {
+			this.$store.dispatch('hidePollToast', value)
```
DorraJaouad, line 250: Also hide after call all toasts remaining ?

src/components/MessagesList/MessagesGroup/Message/MessagePart/MessageBody.vue
Antreesy: Added.
Closing the toast shouldn't dismiss the highlight, only opening the Poll. Then, it's up to participant to vote or not.

DorraJaouad: Still, user can open a Poll and close it without voting, so highlight behaves like a highlighter to non-opened poll and not to non-voted poll.

Antreesy: Yes, that's intended from my side. We shouldn't oblige user to vote, if he doesn't want to, only to open it and check agenda

DorraJaouad (approved):
Tested. Nothing is blocking

PR #11449, opened 2024-01-25
feat(chat): Store and expose whether a message or call was silent

### LINKED ISSUE #8234: Visual indication for silently sent messages
### Developers

> When I get a silent message I wonder if push notification did not work or if it was sent silently. So I ask the sending person if it was silent or not 😵‍💫

### Users

> But I also for non-developers I see messages like "Oh, I didn't see your message until now" pop up

Once an API is done/considered we should raise follow up issues for:
- [ ] Android
- [ ] iOS

nickvergessen:
### ☑️ Resolves

* Fix #8234 

## 🛠️ API Checklist

grafik

### 🚧 Tasks

- [x] Store flag
- [x] Expose as message property
- [x] Resolve FIXMEs in editMessage code

### 🏁 Checklist


SystemKeeper (approved):
Tested and works 👍

PR #11450, opened 2024-01-25
feat(ConversationIcon): highlight public conversations

Antreesy:
### ☑️ Resolves

* Conversation (NcListItem) now indicates, if conversation is public (available for guests by link)
* Place is details above the unread counter

* To discuss:
  * Icon for federated conversations:
    * Same place?
    * Should overwrite public icon?
    * Which icon to use? https://pictogrammers.com/library/mdi/icon/shield-link-variant/ as option

| :sun_with_face: Light theme | :new_moon: Dark theme |
|------------|----------|
| Default with `--color-primary-element-text` for active            | `--color-text-maxcontrast` for others         |
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/LeftSidebar/ConversationsList/Conversation.vue
+++ b/src/components/LeftSidebar/ConversationsList/Conversation.vue
@@ -46,6 +46,10 @@
 					{{ conversationInformation }}
 				</template>
 			</template>
+			<template v-if="isPublic || isGroup" #details>
+				<LinkIcon v-if="isPublic" :size="16" />
+				<AccountMultipleIcon v-else-if="isGroup" :size="16" />
+			</template>
 			<template v-if="!isSearchResult" #actions>
 				<NcActionButton v-if="canFavorite"
 					:close-after-click="true"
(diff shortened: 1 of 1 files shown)


marcoambrosini (commented):
I think we should avoid adding this information on the avatar @Antreesy 
Let me think about this and will come up with an alternative proposal soon.

Antreesy:
Proposal from Marco:
image

Requires an upstream modification of `NcListItem` component

Antreesy:
Rebased and updated screenshots.
Don't think that group conversations require that 'multiple-account' icon, to not overwhelm the sidebar, but we could discuss that

DorraJaouad (approved):
Code-wise, it's fine ✅ 
We can keep the same icon for federated conversations or it should be changed for some reason ?

marcoambrosini:
I think it would be nice info @Antreesy , shouldn't overwhelm much

marcoambrosini:
@Antreesy could we use the same color as the title for those icons?

PR #11452, opened 2024-01-25
chore(deps): update all debounce methods

Antreesy:
### ☑️ Resolves

* Follow-up to #10933 
* Similar solution to https://github.com/nextcloud/spreed/commit/d55ce12868405f2a20465d4e947977e7428f09e6
* Function creation is modified to get rid of error `debounce called with different context`

| 🏚️ Before | 🏡 After |
|------------|----------|
| image         | image       |

### 🚧 Tasks

- [x] Manual testing:
  - [x] src/App.vue
  - [x] src/components/AdminSettings/AllowedGroups.vue
  - [x] src/components/AdminSettings/RecordingServers.vue
  - [x] src/components/AdminSettings/SIPBridge.vue
  - [x] src/components/AdminSettings/SignalingServers.vue
  - [x] src/components/AdminSettings/StunServers.vue
  - [x] src/components/AdminSettings/TurnServer.vue
  - [x] src/components/AdminSettings/TurnServers.vue
  - [x] src/components/CallView/CallView.vue
  - [x] src/components/CallView/Grid/Grid.vue
  - [x] src/components/LeftSidebar/LeftSidebar.vue
  - [x] src/components/MessagesList/MessagesList.vue (spam messages, then join the call)
  - [x] src/components/NewConversationDialog/NewConversationContactsPage.vue
  - [x] src/components/RightSidebar/Participants/ParticipantsTab.vue
  - [x] src/components/RightSidebar/SharedItems/SharedItemsBrowser.vue
  - [x] src/composables/useArrowNavigation.js
- [ ] Try to reproduce error again

### 🏁 Checklist

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -96,6 +96,7 @@ export default {
 			loading: false,
 			isRefreshingCurrentConversation: false,
 			recordingConsentGiven: false,
+			debounceRefreshCurrentConversation: () => {},
 		}
 	},
 
@@ -256,6 +257,7 @@ export default {
 	},
 
 	beforeDestroy() {
+		this.debounceRefreshCurrentConversation.clear?.()
 		if (!getCurrentUser()) {
 			EventBus.$off('should-refresh-conversations', this.debounceRefreshCurrentConversation)
 		}
@@ -479,6 +481,8 @@ export default {
 	},
 
 	async mounted() {
+		this.debounceRefreshCurrentConversation = debounce(this.refreshCurrentConversation, 3000)
+
 		if (!IS_DESKTOP) {
 			checkBrowser()
 		}
@@ -581,12 +585,6 @@ export default {
 			this.fetchSingleConversation(this.token)
 		},
 
-		debounceRefreshCurrentConversation: debounce(function() {
-			if (!this.isRefreshingCurrentConversation) {
-				this.refreshCurrentConversation()
-			}
-		}, 3000),
-
 		changeWindowVisibility() {
 			this.$store.dispatch('setWindowVisibility', !document.hidden)
 			if (this.windowIsVisible) {
@@ -655,6 +653,9 @@ export default {
 		},
 
 		async fetchSingleConversation(token) {
+			if (this.isRefreshingCurrentConversation) {
+				return
+			}
 			this.isRefreshingCurrentConversation = true
 
 			try {
--- a/src/components/AdminSettings/AllowedGroups.vue
+++ b/src/components/AdminSettings/AllowedGroups.vue
@@ -54,7 +54,7 @@
 				track-by="id"
 				label="displayname"
 				no-wrap
-				@search-change="searchGroup" />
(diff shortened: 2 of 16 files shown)


DorraJaouad (approved):
Tested. Not all of them needed refactoring of debounce function. It is needed for those that get mounted in another component so the debounce got called with the new component context ( like `MessagesList`) but it's fine for consistency purposes :) .

Antreesy:
> Not all of them needed refactoring of debounce function.

Still, we need to create debounce function only once. With an old notation, we create a new function every call, so not sure debounce even worked correctly

danxuliu (commented):
[`this.debounceRefreshCurrentConversation` is assigned as the handler for `should-refresh-conversations` in `beforeMount()`](https://github.com/nextcloud/spreed/blob/4a9952f65b3f348c82c8586dc06f0e66da423217/src/App.vue#L274), so [the original empty handler](https://github.com/nextcloud/spreed/blob/4a9952f65b3f348c82c8586dc06f0e66da423217/src/App.vue#L99) is assigned and [the one set in `mounted()`](https://github.com/nextcloud/spreed/commit/4a9952f65b3f348c82c8586dc06f0e66da423217#diff-7a7a37b12ee1265d7e27577ec286120d2cbc0940908635e264a2be44ccb9a8a0R484) is not the one called (this can be verified by adding something like  `console.log('Original')` to the original handler, joining a conversation as a guest and then changing the conversation name as a moderator).

The result is that after these changes [`should-refresh-conversations` events triggered by the signaling object](https://github.com/nextcloud/spreed/blob/380d81dabfaa47f9ce18b894d4622d8076d9f7e4/src/utils/signaling.js#L793) no longer cause the conversation to be refreshed for guests, so for example if the lobby is disabled the UI is not immediately updated (although fortunately [it only takes at most 30 seconds to update](https://github.com/nextcloud/spreed/blob/f49ee981393a79ae2c2d42f99842df2e8c7bc752/src/App.vue#L263-L272)).

Moreover, as the property was changed in `mounted()` [the cleaned up handler in `beforeDestroy()` is the wrong one](https://github.com/nextcloud/spreed/blob/4a9952f65b3f348c82c8586dc06f0e66da423217/src/App.vue#L262) (although in this specific case there are no leaks, as the whole app is destroyed).

I have not checked if a similar problem could have been introduced in the other changed methods.

Antreesy:
Thanks for noticing! Similar cases are:

CallView: correct order of initialization:
```js
this.debounceFetchPeers = debounce(this.fetchPeers, 1500)
EventBus.on('refresh-peer-list', this.debounceFetchPeers)
```

SearchMessagesTab: also correct order:
```js
const debounceFetchSearchResults = debounce(fetchNewSearchResult, 250)
watch([searchText, fromUser, sinceDate, untilDate], debounceFetchSearchResults)
```

Other debounce assignments seems to be called on-demand by other methods, so should not be affected

Issue #11278, opened 2023-12-19
🌐 Federated invitation handling 💌 - API


nickvergessen:
- Overview at https://github.com/nextcloud/spreed/issues/11229

---

- [x] `FederationController->acceptShare()` could return to user, who accepted a share, a local room (with token. e.t.c), so it could be reused by clients to join room on click  => https://github.com/nextcloud/spreed/pull/11459
- [x] Allow `all users / only moderators / only specific groups` on this server to invite users to Talk conversations from `all other servers / only trusted servers` - discuss?  https://github.com/nextcloud/spreed/pull/11503
- [x] Return conversation name in getInvitation endpoint https://github.com/nextcloud/spreed/pull/11505
- [x] Prevent accepting the same invitation multiple times https://github.com/nextcloud/spreed/pull/11513
- [x] Inviting the same user should not result in a new system message each time https://github.com/nextcloud/spreed/pull/11513
- [x] Split subject and message for push notifications https://github.com/nextcloud/spreed/pull/11513
- [x] The name (rich object) of the person which shared the conversation is missing on the invitation endpoint https://github.com/nextcloud/spreed/pull/11581

nickvergessen:
- [ ] Boolean - Allow incoming
- [ ] Boolean - Allow outgoing
- [ ] Array - List of groups allowed to federate
- [ ] Boolean - Filter federation servers by trusted server list of server

SystemKeeper:
https://github.com/nextcloud/spreed/blob/2bf22e12d217572fb25d071b87a74c785d14b56a/lib/Model/Invitation.php#L72-L83

The endpoint to get federation invitation does not include the conversation name that the user was invited to, but the notification does include that in the rich parameters:

<img width="664" alt="image" src="https://github.com/nextcloud/spreed/assets/1580193/480f8945-2e59-4e7d-9f9d-ffab22bdcc19">

I assume we should add that to the invitation endpoint as well?!

SystemKeeper:
Same invitation can be accepted multiple times:

<img width="672" alt="image" src="https://github.com/nextcloud/spreed/assets/1580193/1021a0a5-cc71-44ef-b4ef-b057d1ceab3d">

SystemKeeper:
Please split subject and message:

https://github.com/nextcloud/spreed/blob/1ae1017d444731bc911939af52a70014a8d0ac20/lib/Federation/CloudFederationProviderTalk.php#L332-L342

As we did for recording:

[...]

PR #11459, opened 2024-01-26
feat(federation): Add room to response when accepting an invite

nickvergessen:
### ☑️ Resolves

* Ref https://github.com/nextcloud/spreed/issues/11278

### 🚧 Tasks

- [x] Add full room object to response when accepting an invite

### 🏁 Checklist


Antreesy (approved):
Changes look OK, and I definitely see a room in response

image

PR #11464, opened 2024-01-26
fix(federation): Send disinvite on remote user removal and remove invite

nickvergessen:
### ☑️ Resolves

* Checkbox from https://github.com/nextcloud/spreed/issues/5723

### 🚧 Tasks

- [x] Send disinvite
- [x] Ensure the remote user is no longer in the/a room and does not have an invite anymore

### 🏁 Checklist


Antreesy:
Couple of things I've noticed:

Should `disinvite` dismiss not accepted notifications with `invite`?
image

For accepted users, disinvite just removes them from the room on remote server. Could we send a notification as well (`admin has unshared a room with you`) to handle it on the frontend and leave the federated room on local server?

nickvergessen:
> Should disinvite dismiss not accepted notifications with invite?

It should, fixing this now

> For accepted users, disinvite just removes them from the room on remote server. Could we send a notification as well (admin has unshared a room with you) to handle it on the frontend and leave the federated room on local server?

We should not send a (push) notification, but we should trigger a signaling message for the user I guess?

Antreesy (approved):
Tested

PR #11499, opened 2024-01-31
fix(Poll): hide poll creation in Note to Self

### LINKED ISSUE #11440: Poll creation should be hidden in Note to self
The option to create a poll is still shown in Web but fails with error 400 when trying to create a poll.

See: 
https://github.com/nextcloud/spreed/blob/931617e836e352b60cef5b252752a6ea40307a2d/lib/Controller/PollController.php#L89-L92

Antreesy:
### ☑️ Resolves

* Fix #11440

image

### 🏁 Checklist


DorraJaouad (approved):
Ah finally, one-line PR 😄

PR #11505, opened 2024-02-01
fix(federation)!: Add room name to invite and fix casing of properties

nickvergessen:
### ☑️ Resolves

* Ref https://github.com/nextcloud/spreed/issues/11278#issuecomment-1920068400

### 🚧 Tasks

- [x] Add room name and document state
- [x] Change from `_` casing to camelCase like on all other APIs

### 🏁 Checklist

### DIFF
--- a/lib/Controller/FederationController.php
+++ b/lib/Controller/FederationController.php
@@ -157,17 +157,25 @@ class FederationController extends OCSController {
 		$invitations = $this->federationManager->getRemoteRoomShares($user);
 
 		/** @var TalkFederationInvite[] $data */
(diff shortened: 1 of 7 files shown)


SystemKeeper:
Just by looking at the code, it is correct that `jsonSerialize` does not contain `roomName`?

nickvergessen:
> Just by looking at the code, it is correct that jsonSerialize does not contain roomName?

It can not, as that is not stored in the invitation table

SystemKeeper (approved):
Tested and works

PR #11507, opened 2024-02-01
Hide 'Failed to poll event' in integration tests

SystemKeeper:
Currently:

image

"Failed to poll event" does not give any information about the test and is just noisy -> hide it from output.

### DIFF
--- a/tests/integration/run.sh
+++ b/tests/integration/run.sh
@@ -26,7 +26,7 @@ PHPPID1=$!
 echo -e "Running on process ID: \033[1;35m$PHPPID1\033[0m"
 
 # Output filtered php server logs
-tail -f phpserver.log | grep --line-buffered -v -E ":[0-9]+ Accepted$" | grep --line-buffered -v -E ":[0-9]+ Closing$" &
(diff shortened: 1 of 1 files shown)


SystemKeeper:
Not sure about that: https://github.com/nextcloud/spreed/actions/runs/7739390969/job/21102764394#step:14:1184 Is this known?

nickvergessen:
But it should never fail to poll? Where is this coming from all a sudden? I only see it since yesterday...

SystemKeeper:
> But it should never fail to poll? Where is this coming from all a sudden? I only see it since yesterday...

Oh, is this new? Then let me investigate... I thought we just missed to exclude this.

nickvergessen:
> Not sure about that: https://github.com/nextcloud/spreed/actions/runs/7739390969/job/21102764394#step:14:1184 Is this known?

No, Talk has green CI only

nickvergessen:
What do we do?

SystemKeeper:
Doesn't seem to be an issue in recent runs. So would close this for now.

PR #11514, opened 2024-02-02
fix(l10n): Don't use the word "Room" in UI strings

nickvergessen:


Antreesy (approved):
Couldn't find anything else in the frontend code

PR #11517, opened 2024-02-02
chore(federation): chatting prerequisite

Antreesy:
### ☑️ Resolves

* Splitted less related things from #11383, as its growing too much
  * Add entry for RightSidebar to add remote participants without OCS_API_VIEWER
  * Add icon to indicate federated conversations
  * move `editMessage` to messagesStore
  * patch messagesService, so all could accept options for `generateOcsUrl` and `axios`

image
image

### 🏁 Checklist


Antreesy:
Sorry, canceled as many checks as possible to give some space for server backports

DorraJaouad (commented):
~~Some of the changes were made in https://github.com/nextcloud/spreed/pull/11383~~
~~Are you planning to split them and revert in the other PR ? ~~

Never mind XD

src/components/NewMessage/NewMessage.vue
```diff
@@ -691,14 +690,16 @@ export default {
 		},
 
 		async handleEdit() {
+			const messageId = this.messageToEdit.id
+			const updatedMessage = this.text.trim()
+			this.chatExtrasStore.removeMessageIdToEdit(this.token)
```
DorraJaouad, line 695: The store should be purged only after successful edit. Otherwise, edit mode needs to remain on

src/store/messagesStore.js
```diff
@@ -628,6 +629,37 @@ const actions = {
...
+	 * @param {string} payload.token The conversation token
+	 * @param {string} payload.messageId The message id
+	 * @param {string} payload.updatedMessage The modified text of the message / file share caption
+	 */
+	async editMessage(context, { token, messageId, updatedMessage }) {
+		const message = Object.assign({}, context.getters.message(token, messageId))
+		context.commit('addMessage', {
+			token,
+			message: { ...message, message: updatedMessage },
+		})
```
DorraJaouad, line 646: Why don't you add it after EditMessage request? so you don't need to restore upon failure and it is only called in successful.

DorraJaouad (approved):
All good

PR #11547, opened 2024-02-07
perf(sharing): Cache the list of tokens per user

nickvergessen:
### 🏁 Checklist


lib/Service/ParticipantService.php
```diff
@@ -512,6 +514,9 @@ public function addUsers(Room $room, array $participants, ?IUser $addedBy = null
 						$this->attendeeMapper->delete($attendee);
 						throw new CannotReachRemoteException();
 					}
+				} elseif ($attendee->getActorType() === Attendee::ACTOR_USERS) {
+					$cache = $this->cacheFactory->createDistributed('talk/usertokens');
```
nickvergessen, line 518: Should go into CachePrefix as a const

PR #11551, opened 2024-02-08
fix(files-sidebar): reduce initial loading size of Files sidebar

### LINKED ISSUE #17908: Files Sidebar integration fails at switching conversations
> When trying to start call sometimes (but it is often). try to close a call and start another call for another file?

- Instance exist from the first call and not cleaned up;
- Triggers this to open new tab: (and prevent navigation):
  https://github.com/nextcloud/spreed/blob/bedc1f5dc2b61f75e675fbfe40b8c00d74dc2124/src/router/router.ts#L116-L120
- old conversation chat is opened from first call;
-- conversation isn't joined -> joinCall fails

_Originally posted by @Antreesy in https://github.com/nextcloud/spreed/issues/17860#issuecomment-4371422486_

---

On switching tabs / Talk integration still mounted, prev.conversation state is still valid -> should be destroyed instead

Antreesy:
### ☑️ Resolves

* Fix #17908
* Changes:
  * Initial sidebar tab registration takes only a loader app (EmptyContent + button) with minimal check
  * When interacted, component triggers async load of main content (chat/call) and replace placeholder
  * Closing the sidebar/tab unmount main content and brings back the loader app
  * Does not conflict with floating call

| 🏚️ Before | 🏡 After |
|------------|----------|
| image         | image    |
| 1.2 MB initial + 0.005 MB (async) | 0.121 MB initial + 1.2 MB (async) | 
| image | image |
| image | image |

### 🚧 Tasks

- [ ] Check properly

### 🏁 Checklist


src/mainFilesSidebarLoader.js
```diff
@@ -34,7 +72,7 @@ const isEnabled = function(fileInfo) {
 	// left; this must be done here because "setFileInfo" will not get
 	// called with the new file if the tab can not be displayed.
 	if (token) {
-		OCA.Talk.store.dispatch('leaveConversation', { token })
+		axios.delete(generateOcsUrl('apps/spreed/api/v4/room/{token}/participants/active', { token }))
```
nickvergessen, line 75: that's skipping call handling, signaling and many more things

nickvergessen (dismissed):

src/mainFilesSidebarLoader.js
```diff
@@ -20,50 +20,80 @@
...
+Object.assign(window.OCA.Talk, {
+	fileInfo: null,
+	loaderTab,
+	isFirstLoad: true,
+	// It might be enough to keep the instance only in the Tab object itself,
+	// without using a shared variable that can be destroyed if a new tab is
+	// mounted and the previous one was not destroyed yet, as the tabs seem to
+	// always be properly destroyed. However, this is how it is done for tabs in
+	// server, so it is done here too just to be safe.
+	tabInstance: null,
```
ShGKme, line 83: Do we need the tab instance as a part of the public API in OCA.Talk?

Antreesy: not neccessary to have it in the public, but this variable should be shared between mainFilesSidebar and mainFilesSidebarLoader, so Sidebar would work correctly with the reference (mount, update, destroy)

Antreesy: I believe, it should work as expected?
```js
const PrivateTalk = {
	tabInstance: null,
}
export default PrivateTalk
```

ShGKme:
Rebased onto `main` and resolved conflicts

nickvergessen:
get in? backport?

Antreesy:
> get in? backport?

Tested once again, works correctly. Most of the JS code (1 Mb in dev) is loaded after sidebar is opened and DOMContentLoaded, so somewhat a boost. I'm fine with both options

src/mainFilesSidebarLoader.js
Antreesy: Aligned with floatingCall - both should not be active simultaneously

src/mainFilesSidebarLoader.js
```diff
@@ -3,14 +3,37 @@
...
+ * Create lightweight loader app for Files Sidebar
+ */
+function newTab() {
+	return defineCustomElement(FilesSidebarLoaderApp, {
+		shadowRoot: false,
+		configureApp(app) {
+			app.use(NextcloudGlobalsVuePlugin)
+		},
+	})
+}
```
ShGKme: nitpick: If we don't need it anywhere else anymore, I'd just put this code to the `onInit`, without a second function.

src/mainFilesSidebar.js
```diff
@@ -24,29 +22,31 @@ __webpack_nonce__ = getCSPNonce()
...
+export function mountApp(container, rootProps, token) {
+	initializeTalk()
+
+	const router = createMemoryRouter()
+
+	const instance = createApp(FilesSidebarTabApp, { ...rootProps, token })
+		.use(store)
+		.use(pinia)
+		.use(router)
+		.use(NextcloudGlobalsVuePlugin)
```
ShGKme: nitpick: checked, `NextcloudGlobalsVuePlugin` is not needed anymore, we can remove it.

rspack.config.js
```diff
@@ -44,10 +44,7 @@ module.exports = defineConfig((env) => {
 			collections: path.join(__dirname, 'src', 'collections.js'),
 			main: path.join(__dirname, 'src', 'main.js'),
 			recording: path.join(__dirname, 'src', 'mainRecording.js'),
-			'files-sidebar': [
-				path.join(__dirname, 'src', 'mainFilesSidebar.js'),
-				path.join(__dirname, 'src', 'mainFilesSidebarLoader.js'),
-			],
```
ShGKme, line 50: `mainFilesSidebar` is not an entrypoint but just a `mountApp` function.

To make it less confusing, I'd:
- Remove `__webpack_*` setting there, it is only needed in entry points
- Rename the file and move it from the root, because it is not a `main` anymore but just a function

src/FilesSidebarTabApp.vue
```diff
@@ -294,94 +206,27 @@ export default {
...
-					this.isTalkSidebarSupportedForFile = false
+		syncLeaveConversation() {
+			console.info('Navigating away, leaving conversation')
+			if (this.token) {
+				this.tokenStore.updateTokenAndFileIdForToken('', null)
+				SessionStorage.removeItem('joined_conversation')
+				// We have to do this synchronously, because in unload and beforeunload
+				// Promises, async and await are prohibited.
+				signalingKill()
+				signalingWebRtcKill()
```
ShGKme, line 201: **issue:** `signalingWebRtcKill` is not imported

ShGKme (approved):

PR #11575, opened 2024-02-10
Feat: Plug-in talk specific filters for unified search

Fenn-CS:
### ☑️ Resolves

* Fix https://github.com/nextcloud/server/issues/42914

Filter | Selector
--|--
image | image

[Screencast from 02-06-2024 11:27:48 AM.webm](https://github.com/nextcloud/server/assets/14317775/4ce91e6d-0d1b-4b94-8a90-b793c98900db)

### 🚧 Tasks

- [ ] ...
- [ ] ...

### 🏁 Checklist

### DIFF
--- a/lib/Search/UnifiedSearchFilterPlugin.php
+++ b/lib/Search/UnifiedSearchFilterPlugin.php
@@ -1,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2024 Fon E. Noel NFEBE <me@nfebe.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Talk\Search;
+
+use OCP\AppFramework\Http\Events\BeforeTemplateRenderedEvent;
+use OCP\EventDispatcher\Event;
+use OCP\EventDispatcher\IEventListener;
+use OCP\IRequest;
+use OCP\Util;
+
+/**
+ * @template-implements IEventListener<Event>
+ */
+class UnifiedSearchFilterPlugin implements IEventListener {
+
+	public function __construct(
+		private IRequest $request,
+	) {
+	}
+
+	public function handle(Event $event): void {
+		if (!($event instanceof BeforeTemplateRenderedEvent)) {
+			return;
+		}
+
+		if (!$event->isLoggedIn()) {
+			return;
+		}
+
+		Util::addScript('spreed', 'talk-search');
+	}
+}
--- a/src/search.js
+++ b/src/search.js
@@ -1,0 +1,100 @@
+/*
+ * @copyright Copyright (c) 2024 Fon E. Noel NFEBE <me@nfebe.com>
+ *
+ * @author Vincent Petry <me@nfebe.com>
+ *
+ * @license AGPL-3.0-or-later
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+import Vue from 'vue'
+
+import { getRequestToken } from '@nextcloud/auth'
+import { emit } from '@nextcloud/event-bus'
+import { translate, translatePlural } from '@nextcloud/l10n'
+import { generateFilePath, imagePath } from '@nextcloud/router'
+
+import '@nextcloud/dialogs/style.css'
+
+(function(OC, OCP, t, n) {
+
+	/**
+	 *
+	 */
+	function init() {
+		if (!OCP.UnifiedSearch) {
+			return
+		}
+		console.debug('Initializing unified search plugin-filters from talk')
+		OCP.UnifiedSearch.registerFilterAction({
+			id: 'talk-message',
+			label: t('spreed', 'In conversation'),
+			icon: imagePath('spreed', 'app.svg'),
+			callback: () => {
+				const container = document.createElement('div')
+				container.id = 'spreed-unified-search-conversation-select'
+				const body = document.getElementById('body-user')
+				body.appendChild(container)
+
+				const RoomSelector = () => import('./components/RoomSelector.vue')
+				const vm = new Vue({
+					el: container,
+					render: h => h(RoomSelector, {
+						props: {
+							dialogTitle: t('spreed', 'Select conversation'),
+							isPlugin: true,
+						},
+					}),
+				})
+
+				vm.$root.$on('close', () => {
+					vm.$el.remove()
+					vm.$destroy()
+				})
+				vm.$root.$on('select', (conversation) => {
+					vm.$el.remove()
+					vm.$destroy()
+
+					emit('nextcloud:unified-search:add-filter', {
+						id: 'talk-message',
+						payload: conversation,
+						filterUpdateText: t('spreed', 'Search in conversation: {conversation}', { conversation: conversation.displayName }),
+						filterParams: { conversation: conversation.token }
+					})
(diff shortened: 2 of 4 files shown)


src/search.js
```diff
@@ -0,0 +1,104 @@
...
+	 *
+	 */
+	function init() {
+		if (!OCA.Core) {
+			return
+		}
+		console.debug('Initializing unified search plugin-filters from talk')
+		OCA.Core.UnifiedSearch.registerFilterAction({
+			id: 'talk-message',
+			label: t('spreed', 'In Conversation'),
```
nickvergessen, line 48:
```suggestion
			label: t('spreed', 'In conversation'),
```

src/search.js
```diff
@@ -0,0 +1,104 @@
...
+
+	/**
+	 *
+	 */
+	function init() {
+		if (!OCA.Core) {
+			return
+		}
+		console.debug('Initializing unified search plugin-filters from talk')
+		OCA.Core.UnifiedSearch.registerFilterAction({
```
nickvergessen, line 46: OCA is for apps, Core is not apps. I didn't check recent JS changes in servers, but should this be OCP.UnifiedSearch?

src/search.js
```diff
@@ -0,0 +1,104 @@
...
+				const container = document.createElement('div')
+				container.id = 'spreed-post-card-to-room-select'
+				const body = document.getElementById('body-user')
+				body.appendChild(container)
+
+				const ComponentVM = Vue.extend(RoomSelector)
+				const vm = new ComponentVM({
+					el: container,
+					propsData: {
+						dialogTitle: t('spreed', 'Select Conversation'),
```
nickvergessen, line 60:
```suggestion
						dialogTitle: t('spreed', 'Select conversation'),
```

src/search.js
```diff
@@ -0,0 +1,104 @@
...
+				})
+				vm.$root.$on('select', (token) => {
+					vm.$el.remove()
+					vm.$destroy()
+					fetchConversation(token).then(response => {
+						const conversation = response.data.ocs.data
+						emit('nextcloud:unified-search:add-filter', {
+							id: 'talk-message',
+							payload: conversation,
+							filterUpdateText: `Search in conversation : ${conversation.displayName}`,
```
nickvergessen, line 77:
```suggestion
							filterUpdateText: `Search in conversation: ${conversation.displayName}`,
```
And should be translated with a normal placeholder

nickvergessen:
@Antreesy can you check for loader stuff so we don't generate the next topic for the performance hackweek 😅

src/search.js
```diff
@@ -0,0 +1,104 @@
...
+	 */
+	function init() {
+		if (!OCA.Core) {
+			return
+		}
+		console.debug('Initializing unified search plugin-filters from talk')
+		OCA.Core.UnifiedSearch.registerFilterAction({
+			id: 'talk-message',
+			label: t('spreed', 'In Conversation'),
+			icon: '/apps/spreed/img/app.svg',
```
Antreesy, line 49: Not sure that path is picked correctly. Might be because I have it in 'apps-extra'?
image

src/search.js
```diff
@@ -0,0 +1,104 @@
...
+			callback: () => {
+				const container = document.createElement('div')
+				container.id = 'spreed-post-card-to-room-select'
```
Antreesy, line 52: may conflicts in Deck app

src/search.js
```diff
@@ -0,0 +1,104 @@
...
+				const body = document.getElementById('body-user')
+				body.appendChild(container)
+
+				const ComponentVM = Vue.extend(RoomSelector)
+				const vm = new ComponentVM({
+					el: container,
+					propsData: {
+						dialogTitle: t('spreed', 'Select Conversation'),
+						showPostableOnly: true,
```
Antreesy, line 61: Not needed, as we might also want to search in read-only conversations

lib/Search/UnifiedSearchFilterPlugin.php
```diff
@@ -0,0 +1,56 @@
...
+			return;
+		}
+
+		if (!$event->isLoggedIn()) {
+			return;
+		}
+
+		// REVIEW: Only add the talk filter if user is on talk app?
+		if (str_starts_with($this->request->getPathInfo(), '/apps/spreed')) {
+			Util::addScript('spreed', 'talk-search');
```
Antreesy, line 53: Script is around 460 kB dev / 270 kB production. Can't see nothing in the bundle analyzer, that we could get rid of (only minor optimizations from our side, maybe)

src/search.js
```diff
@@ -0,0 +1,104 @@
...
+				})
+
+				vm.$root.$on('close', () => {
+					vm.$el.remove()
+					vm.$destroy()
+				})
+				vm.$root.$on('select', (token) => {
+					vm.$el.remove()
+					vm.$destroy()
+					fetchConversation(token).then(response => {
```
Antreesy, line 72: That's an overkill, IMO, as we have a full conversation object in the RoomSelector. But that's on our side, I could provide a fix for that separately

src/components/RoomSelector.vue
```diff
@@ -21,7 +21,7 @@
   -->
 
 <template>
-	<NcModal size="normal"
```
Antreesy, line 24: These changes broke usage of component (and tests for it) in too many places:
- Deck, Maps, Search, Collections integrations
- Talk inner features

I'd suggest you to drop this commit, so we could also work on it separately

Talk | Search | Deck
--|--|--
image | image | image

src/search.js
Antreesy: Might be better, but it returns the same path for me

Fenn-CS: Maybe there is a better of obtaining the app icon from within the app? `imagePath()` looks like it would send a full URL, which would again be passed to `generateUrl` since the other providers from the API send their icons in this format.

lib/Search/UnifiedSearchFilterPlugin.php
Antreesy: Would be nice to have it lazy-loaded, when opening unified search modal

src/search.js
```diff
@@ -0,0 +1,103 @@
...
+						const conversation = response.data.ocs.data
+						emit('nextcloud:unified-search:add-filter', {
```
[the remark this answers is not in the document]
nickvergessen: I guess it should be generic, so we don't need to change server everytime an app starts implementing a filter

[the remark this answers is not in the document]
Fenn-CS: Yes, that is why I changed it. I probably reverted during rebase in the server PR would fix there, so this is fine.

lib/Search/UnifiedSearchFilterPlugin.php
Fenn-CS: Not sure whether this a global backend event for such lazy load to depend on right now.

src/search.js
```diff
@@ -0,0 +1,103 @@
...
+ */
+
+import Vue from 'vue'
+
+import { getRequestToken } from '@nextcloud/auth'
+import { emit } from '@nextcloud/event-bus'
+import { translate, translatePlural } from '@nextcloud/l10n'
+import { generateFilePath } from '@nextcloud/router'
+
+import RoomSelector from './components/RoomSelector.vue'
```
DorraJaouad, line 30: You can dynamically import it instead 
```
const RoomSelector = () => import('./components/RoomSelector.vue')
```

Antreesy: If it could be done inside of init function

DorraJaouad: Ah I assumed init won't be called as the script is loaded on `BeforeTemplateRenderedEvent`. For optimization purposes, if one of these workaround is possible it would be great:
- Lazy load the script itself (  when the unified search modal is opened) 
- Add the script immediately ( when Talk is active) and call the function that initializes the filter action when the unified search modal is opened (and have `RoomSelector` lazy loaded).

Antreesy: You could refer to this commit for component lazy loading: https://github.com/nextcloud/spreed/pull/11589/commits/13de1cc65c6e8876ac57cbc29b4e62270e63c1ba

The benefit is that since you force this script to be loaded on every page where UnifiedSearch is present, we have to minimize its size (e.g. using dynamic import) so that the user doesn't load code every time that they won't use.

src/search.js
```diff
@@ -0,0 +1,103 @@
...
+	/**
+	 *
+	 */
+	function init() {
+		if (!OCP.UnifiedSearch) {
+			return
+		}
+		console.debug('Initializing unified search plugin-filters from talk')
+		OCP.UnifiedSearch.registerFilterAction({
+			id: 'talk-message',
```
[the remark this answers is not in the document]
nickvergessen: It is (I'm the API dev)

Antreesy:
Couldn't get it rendered anymore:
image

Is there anything I'm mising to install on server/ API level?

Good job with dynamic import, btw. That's about collectives plugin size, so should be good for the current release
image

Fenn-CS:
@Antreesy Not sure, can you just pull latest https://github.com/nextcloud/server/pull/43189 and compile and try again?

Fenn-CS:
I am aware of this,  that is why I left this question https://github.com/nextcloud/spreed/pull/11575#discussion_r1489370174 however you are right it is out of scope and we have to figure out why the backend does not respond correctly at the server level.

cc: @Antreesy

src/search.js
Antreesy:
```suggestion
			icon: imagePath('spreed', 'app.svg'),
```

Now works correctly for me with that change: 
image

Antreesy:
Hello! I've rebased the PR and added some changes:
- Updated RoomSelector design
- Changed props
- Changed output (whole conversation instead of token, no need to fetch it)
- Fix icon in filters dropdown

cc @Fenn-CS @DorraJaouad

Fenn-CS (commented):
Updates LGTM

DorraJaouad (approved):
All good
Thanks @Fenn-CS @Antreesy

PR #11583, opened 2024-02-13
Improve room selector modal

Fenn-CS:
### ☑️ Resolves

* Refactor of RoomSelector
  * Use ConversationsListVirtual and ConversationSearchResult for conversations list:
    * improve accessibility, optimize rendering
    * show last message if provided :warning: optionally, disabled atm
    * resolve issue with immediate load of all avatars
  * Test coverage

| 🏚️ Before | 🏡 After |
|------------|----------|
image | image
image | image

Test coverage:
image

### 🚧 Tasks

- [ ] Code review
- [ ] Design review
- [ ] **TODO**:
  - [x] emit full conversation instead of token on 'Select conversation' - https://github.com/nextcloud/spreed/pull/11589
  - [x] Split virtual list into two components => Conversation pulls all dependencies
  - [x] reuse shared computed/methods between ConversationSearchResult / Conversation
  - [ ] check ConversationSearchResult component for reuse in LeftSidebar (need to provide actions)

### DIFF
--- a/src/components/LeftSidebar/ConversationsList/ConversationSearchResult.vue
+++ b/src/components/LeftSidebar/ConversationsList/ConversationSearchResult.vue
@@ -25,8 +25,8 @@
 		:name="item.displayName"
 		:title="item.displayName"
 		:active="item.token === selectedRoom"
-		:bold="!!item.unreadMessages"
-		:counter-number="item.unreadMessages"
+		:bold="exposeMessages && !!item.unreadMessages"
+		:counter-number="exposeMessages ? item.unreadMessages : 0"
 		:counter-type="counterType"
 		@click="onClick">
 		<template #icon>
@@ -39,13 +39,13 @@
 </template>
 
 <script>
-import { inject } from 'vue'
+import { inject, toRefs } from 'vue'
 
 import NcListItem from '@nextcloud/vue/dist/Components/NcListItem.js'
 
 import ConversationIcon from './../../ConversationIcon.vue'
 
-import { CONVERSATION, ATTENDEE } from '../../../constants.js'
+import { useConversationInfo } from '../../../composables/useConversationInfo.js'
 
 export default {
 	name: 'ConversationSearchResult',
@@ -56,6 +56,10 @@ export default {
 	},
 
 	props: {
+		exposeMessages: {
+			type: Boolean,
+			default: false,
+		},
 		item: {
 			type: Object,
 			default() {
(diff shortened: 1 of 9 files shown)


Antreesy:
Thanks! There are couple of things to improve here, so I'll take it over if you don't mind =)

src/components/RoomSelector.vue
```diff
@@ -194,6 +182,9 @@ export default {
 			this.$root.$emit('close')
 			this.$emit('close')
 		},
+		onSelect(item) {
+			this.selectedRoom = item.token
+		},
 		select() {
 			this.$root.$emit('select', this.selectedRoom)
```
DorraJaouad, line 189: Please check why it is needed to be emitted from the root or if EventBus can be an alternative here

src/components/RoomSelector.vue
```diff
@@ -41,23 +41,11 @@
...
-						</li>
-					</ul>
-					<div v-else-if="!loading" class="no-match-message">
-						<h2 class="no-match-title">
-							{{ noMatchFoundTitle }}
-						</h2>
-						<p v-if="noMatchFoundSubtitle" class="subtitle">
-							{{ noMatchFoundSubtitle }}
-						</p>
-					</div>
```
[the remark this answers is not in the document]
Antreesy: To be replaced with NcEmptyContent

src/components/RoomSelector.vue
Antreesy: That was there historically: https://github.com/nextcloud/spreed/blob/0c9599e598ca34e6b0ba27dae3cb319a668be7fd/src/deck.js#L103
If you ask me, I'd prefer this way, cause EventBus spins an extra Vue instance and add some weight to the bundle, which we want to avoid for plugins, if possible

DorraJaouad: Maybe it is not needed in the first place because select here is meant to pass the info to the parent component only ( and from their the select button).

src/components/LeftSidebar/ConversationsList/ConversationSearchResult.vue
```diff
@@ -0,0 +1,198 @@
...
+		class="conversation-item"
+		:class="{'unread-mention-conversation': item.unreadMention}"
+		:data-nav-id="`conversation_${item.token}`"
+		:actions-aria-label="t('spreed', 'Conversation actions')"
+		:bold="!!item.unreadMessages"
+		:counter-number="item.unreadMessages"
+		:counter-type="counterType"
+		@click="onClick">
+		<template #icon>
+			<ConversationIcon :item="item" :hide-favorite="false" :hide-call="false" />
```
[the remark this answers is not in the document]
Antreesy: Done. If you're not a participant (attendeeId === 0), item shouldn't show you anything
Open conv | Joined conv
--|--
image | image

src/components/LeftSidebar/ConversationsList/ConversationSearchResult.vue
```diff
@@ -0,0 +1,198 @@
...
+		</template>
+		<template #subname>
+			<strong v-if="item.unreadMessages"
+				class="subtitle">
+				{{ conversationInformation }}
+			</strong>
+			<template v-else>
+				{{ conversationInformation }}
+			</template>
+		</template>
```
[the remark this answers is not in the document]
Antreesy: v-if should be enough for now. Will create a follow-up later.
But what you wrote seems correct, open conversation omits some information, even if you were a participant before

src/components/RoomSelector.vue
```diff
@@ -297,4 +257,12 @@ li {
 .search-form {
 	margin-bottom: 10px;
 }
+
+.scroller {
+	padding: 0 4px;
```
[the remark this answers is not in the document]
Antreesy: removed

DorraJaouad (commented):
BTW, Conversation can behave like ConversationSearchResult when isSearchResult is truthy and it is currently used in LeftSidebar. It would be better if we align and change to ConversationSearchResult there too and remove the prop from Conversation.

DorraJaouad:
We need design review for adding last message to the conversations list ( cc @szaimen ). It is altering the UX a bit as user is now exposed to more information ( probably more than needed) when they are forwarding a message to a conversation or sharing a deck. Usually ( I mean in other chatting platforms), messages are not shown for such actions (user looks for the conversation name rather then last message).
image

szaimen:
> We need design review for adding last message to the conversations list ( cc @szaimen ). It is altering the UX a bit as user is now exposed to more information ( probably more than needed) when they are forwarding a message to a conversation or sharing a deck. Usually ( I mean in other chatting platforms), messages are not shown for such actions (user looks for the conversation name rather then last message). image

The design looks good to me in general 👍

src/components/RoomSelector.vue
```diff
@@ -203,98 +220,44 @@ export default {
...
+:deep(.modal-wrapper *) {
+	box-sizing: border-box;
 }
 
-.talk-modal {
-	height: 80vh;
+:deep(.modal-wrapper .modal-container) {
+	height: 100%;
+	max-height: 700px !important;
 }
```
DorraJaouad, line 232: I highly suggest to use vh 
image

It is too high for 1920 x 1080 display :/ and also the close button is unclickable because it is overlapping with the main header.

Antreesy: Restored back original state. max-height should be as per default, and don't overlap with header:
```css
:deep(.modal-wrapper .modal-container) {
	height: 700px;
}
```

Antreesy: Restored back original state. max-height should be as per default, and don't overlap with header:
```css
:deep(.modal-wrapper .modal-container) {
	height: 700px;
}
```

DorraJaouad (approved):
Tested 🦅

PR #11586, opened 2024-02-14
Fix/11585/follow up all reactions button

### LINKED ISSUE #11585: Follow up to "All" reactions button
Seeing it the first time in use now, I think we should change 2 things:
grafik

- Hide the button similarly like the "Emoji plus" when the message is not hovered/focused
- To avoid jumping with above change, move the heart to the end between the reactions and the "Emoji plus"

_Originally posted by @nickvergessen in https://github.com/nextcloud/spreed/issues/11508#issuecomment-1942526250_

DorraJaouad:
### ☑️ Resolves

* Fix #11585

| 🏚️ Before | 🏡 After |
|------------|----------|
| image     | image    |

### 🚧 Tasks

- [ ] Visual review
- [ ] Code review

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -46,6 +46,7 @@
 				:id="id"
 				:token="token"
 				:can-react="canReact"
+				:show-controls="isHovered || isFollowUpEmojiPickerOpen"
 				@emoji-picker-toggled="toggleFollowUpEmojiPicker" />
 		</div>
 
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.vue
@@ -23,12 +23,6 @@
 <template>
 	<!-- reactions buttons and popover with details -->
 	<div class="reactions-wrapper">
-		<!-- all reactions button -->
-		<NcButton class="reaction-button"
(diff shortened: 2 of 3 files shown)


src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -46,6 +46,7 @@
 				:id="id"
 				:token="token"
 				:can-react="canReact"
+				:is-hovered="isHovered"
```
Antreesy, line 49:
```suggestion
				:show-controls="isHovered || isFollowUpEmojiPickerOpen"
```

Then it won't dissapear, if you open emoji picker and move cursor out of its borders
And naming could be adjusted a bit

src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.vue
```diff
@@ -57,8 +51,16 @@
 			</div>
 		</NcPopover>
 
+		<!-- all reactions button -->
+		<NcButton v-if="isHovered"
```
Antreesy, line 55:
```suggestion
		<NcButton v-if="hasReactions && showControls"
```

Don't we need to check `hasReactions`?

Antreesy (approved):

src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.vue
DorraJaouad: no because Reactions comp is not shown if there is no reaction.

PR #11617, opened 2024-02-21
fix(search): Hide search providers when not allowed to use Talk

nickvergessen:
## 🛠️ API Checklist

- Similar to https://github.com/nextcloud/spreed/pull/11615
- But for the providers which exist already on stable branches, so backporting it there to 28 (since then null is supported)

### 🏁 Checklist


Antreesy (approved):
Seems legit to me

PR #11624, opened 2024-02-21
Feat(MediaSettings): Add audio test

### LINKED ISSUE #1181: system check
Skype and others provide a way to check the system before call.For example record some voice and play it back. Show the webcam video. Test upstream/downstream speed. Test CPU speed...
It would be great if the use could also start this kind of test to check if everything is ok for a call. This also increases the happyiness of the users with Talk because they realize that the software is not broken but the PC is not working properly.

DorraJaouad:
### ☑️ Resolves

* Fix #1181

| 🏚️ Before | 🏡 After |
|------------|----------|
| image  | image     |

Screencast :

[audio-test.webm](https://github.com/nextcloud/spreed/assets/93392545/9cee50aa-9023-4d64-ab7a-ef206683e8d2)

### 🚧 Tasks

- [ ] Design review
- [ ] Code review

### 🏁 Checklist

### DIFF
--- a/src/components/MediaSettings/MediaSettings.vue
+++ b/src/components/MediaSettings/MediaSettings.vue
@@ -221,10 +221,10 @@ import NcModal from '@nextcloud/vue/dist/Components/NcModal.js'
 import NcNoteCard from '@nextcloud/vue/dist/Components/NcNoteCard.js'
 import Tooltip from '@nextcloud/vue/dist/Directives/Tooltip.js'
 
+import MediaDevicesSelector from './MediaDevicesSelector.vue'
 import VideoBackgroundEditor from './VideoBackgroundEditor.vue'
 import AvatarWrapper from '../AvatarWrapper/AvatarWrapper.vue'
 import VideoBackground from '../CallView/shared/VideoBackground.vue'
-import MediaDevicesSelector from '../MediaDevicesSelector.vue'
 import CallButton from '../TopBar/CallButton.vue'
 import VolumeIndicator from '../VolumeIndicator/VolumeIndicator.vue'
 
--- a/src/components/MediaSettings/MediaSettings.vue
+++ b/src/components/MediaSettings/MediaSettings.vue
@@ -112,6 +112,7 @@
 					:devices="devices"
 					:device-id="videoInputId"
 					@update:deviceId="videoInputId = $event" />
+				<MediaDevicesSpeakerTest />
 			</div>
 
 			<!-- Background selection -->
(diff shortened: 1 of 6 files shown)


DorraJaouad:
> Audio Output: Add the "voice message" option on the media settings screen to allow recording and then playing back? Or a simple "bing" sound for testing?

I used something like bing sound for audio output testing. However, I can also add voice recording as a test for the mic. 
cc @nextcloud/designers  @nickvergessen

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -649,7 +660,18 @@ export default {
...
 			this.$emit('update:recording-consent-given', value)
-		}
+		},
+
+		playTestSound() {
+			if (this.isPlayingTestSound) {
+				return
+			}
+			this.isPlayingTestSound = true
+			this.$store.dispatch('playTestAudio')
```
Antreesy, line 670:
```suggestion
			this.$store.dispatch('playWaitAudio')
```

Antreesy (approved):
I don't mind adding a new sound, just feeling that default call waiting fits better here

marcoambrosini (commented):
Nice @DorraJaouad 
I would left-align this new button so that the speaker icon is vertically aligned with the mic and camera icons, and change the test to "Test speakers".

Also, given that we're adding more elements, could we move the "always show dialog"  setting in the 3-dot menu next to the button as an action checkbox component?

nickvergessen:
> However, I can also add voice recording as a test for the mic.

Exactly that was the idea, it would basically allow you to test both, mic and speaker

img/dingSound.flac
nickvergessen, line 1: Needs a note in the license file

nickvergessen: If not replaced by recording button

DorraJaouad:
> Exactly that was the idea, it would basically allow you to test both, mic and speaker

Right, but some users want just to test speaker and they don't want to randomly say some words XD . That's why I chose the first option. So add both or only voice recorder ?

nickvergessen:
> So add both or only voice recorder ?

I leave that to design cc @marcoambrosini 

As for what to play, any issue with playing an already existing file (less files, more caching, less license issues, etc)

marcoambrosini:
@DorraJaouad @nickvergessen would it be possible to have a third entry with speaker selection? Or at least showing which is the current speaker?

DorraJaouad:
[sinkId ](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/sinkId)can show the audio output being used but read only and it is not supported on Safari. 

Otherwise, [selectAudioOutput](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/selectAudioOutput) allows user to select audio ouput device but it is only supported on Firefox.

Antreesy:
> Or at least showing which is the current speaker?

`navigator.mediaDevices.enumerateDevices()` can list available output devices. Though we couldn't identify, which is being used by the browser (at best it shows `Default`). https://developer.mozilla.org/en-US/docs/Web/API/AudioContext is currently not implemented widely, so there's nothing to do atm.

```js
navigator.mediaDevices.enumerateDevices().then((devices) => {
      devices.forEach((device) => {
        console.log(`${device.kind}: ${device.label} id = ${device.deviceId}`);
      });
    })
```
image

marcoambrosini (approved):
I think that this will suffice for now. I'll open an issue for improvements once this is in

PR #11627, opened 2024-02-21
fix(tests): Increase health-retries for mysql tests

SystemKeeper:
Looks like the mysql container switches to `unhealthy` too fast (could be dependent on how busy the server is). Since github waits longer then our 25 seconds (5s interval * 5 retries), we should increase the retries.
Relevant github code is at https://github.com/actions/runner/blob/d5c7097d2c4f3f40a029124bbf1bc4c9994929f5/src/Runner.Worker/ContainerOperationProvider.cs#L411-L426

### 🏁 Checklist

### DIFF
--- a/.github/workflows/integration-mysql.yml
+++ b/.github/workflows/integration-mysql.yml
@@ -68,7 +68,7 @@ jobs:
           - 4444:3306/tcp
         env:
           MYSQL_ROOT_PASSWORD: rootpassword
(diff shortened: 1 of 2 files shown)


skjnldsv:
right, I can add the workflow permissions :+1:

skjnldsv:
Done, let's try again

PR #11630, opened 2024-02-21
feat(BotInstallEvent): Add "features" field

### LINKED ISSUE #11577: Provide a custom event for AppAPI to receive message to send to bots

Starting from AppAPI `2.0` we support installation of ExApps on the remote servers.
In these cases, we use additional authentication + brute-force protection using `HaProxy` and a password, which is not compatible with the current implementation of TalkBots.

As a temporary solution we implemented such code:
https://github.com/cloud-py-api/app_api/blob/5e492fb4d9102233d0fad03d38c41f5f642d2b70/lib/Controller/TalkBotController.php#L83-L131

Where AppAPI act as a proxy. But it is a temporary solution, and very ugly.

Perhaps a solution would be to add a new value to the BotInstallEvent, such as "callFunctionFlag", and when it is set, the call to provided function occurs instead of a request to the URL.

`url` in this case will be something like: "\OCA\AppAPI\BotMessage"

This will allow PHP applications that installed inside Nextcloud to receive messages directly without authentication.

Implementing a new event: "BotMessage" and PHP applications can listen for it and do with it whenever they want.

Maybe some other alternative will be better and someone has a better idea.

bigcat88:
### ☑️ Resolves

* Fix #11577

### 🏁 Checklist

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -193,6 +193,7 @@ Dispatch this event in your app to install a bot on this server
 
 * Event: `OCA\Talk\Events\BotInstallEvent`
 * Since: 17.0.0
+* Since: 19.0.0 - Features flag was added
 
 ### Bot uninstall
 
--- a/lib/Events/BotInstallEvent.php
+++ b/lib/Events/BotInstallEvent.php
@@ -25,6 +25,7 @@ declare(strict_types=1);
 
 namespace OCA\Talk\Events;
 
+use OCA\Talk\Model\Bot;
 use OCP\EventDispatcher\Event;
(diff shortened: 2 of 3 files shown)


nickvergessen (changes requested):
Please add to https://github.com/nextcloud/spreed/blob/d47ab76a29863269041997e5cd8bcb6f5a05cfdf/docs/events.md#bot-install
```
* Since: 19.0.0 - Features flag was added
```

nickvergessen (approved):

PR #11633, opened 2024-02-22
fix(JEST): use nextcloud/l10n library in tests, omit some console logs

Antreesy:
### ☑️ Resolves

* Fix old pending TODO
* lib is available to be used in JEST testing directly

### 🏁 Checklist


src/stores/__tests__/guestName.spec.js
```diff
@@ -109,7 +109,6 @@ describe('guestNameStore', () => {
 	test('translates default guest name', () => {
 
 		expect(store.getGuestName('token-1', 'actor-id0')).toBe('Guest')
-		expect(global.t).toHaveBeenCalledWith('spreed', 'Guest')
```
DorraJaouad, line 112: Simply remove the test itself ?

src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.spec.js
```diff
@@ -282,6 +282,7 @@ describe('Reactions.vue', () => {
 		test('fetches reactions details when they are not available', async () => {
 			// Arrange
 			reactionsStore.resetReactions(token, messageId)
+			console.debug = jest.fn()
```
DorraJaouad, line 285: Not needed here as the response is mocked and does not produce an error. Same for others ?

Antreesy: Same. It's logging in production, but we don't need it in the jest output

src/stores/__tests__/guestName.spec.js
DorraJaouad: Yes so there is no point to keep the test `translates default guest name`

DorraJaouad (approved):
🐳

PR #11634, opened 2024-02-22
Feat(draft chat input): Store input in browser storage

### LINKED ISSUE #3055: Unsent message gets lost when accidentally closing tab
## Steps to reproduce
1. Write some well thought-out message, don’t send it yet cause it’s in progress
2. Accidentally close the browser tab or whole browser (alternatively have it crash)
3. Reopen it and see the message is gone

### Expected behaviour
Somehow there should be error-prevention here. Several options:
- Easy: If there’s anything typed in the message field and you try to close the window, show one of these browser alerts that you will lose your input.
- Medium: Save the message locally in the browser so it’s restored even on crashes.
- Hardcore (and proper): Have the possibility to have per-conversation drafts which are saved cross-platform so you can continue on your mobile.

DorraJaouad:
### ☑️ Resolves

* Fix #3055 

* Reduced the changes between chat input and test of NewMessage. 

No visual changes. 

### 🚧 Tasks

- [ ] Code review

### 🏁 Checklist


src/components/NewMessage/NewMessage.vue
```diff
@@ -510,10 +516,14 @@ export default {
 		token: {
 			immediate: true,
 			handler(token) {
+				if (!this.chatInput) {
+					this.chatExtrasStore.restoreChatInput(token)
+				}
+
```
Antreesy, line 522:
```suggestion
```

Not needed, as we're doing it in store

src/stores/chatExtras.js
```diff
@@ -130,6 +131,18 @@ export const useChatExtrasStore = defineStore('chatExtras', {
...
+		/**
+		 * Restore chat input from the browser storage and save to store
+		 *
+		 * @param {string} token The conversation token
+		 */
+		restoreChatInput(token) {
+			const chatInput = BrowserStorage.getItem('chatInput_' + token)
+			if (chatInput) {
+				Vue.set(this.chatInput, token, chatInput)
+			}
```
Antreesy, line 143: Otherwise you're removing the stored value before we get them
```js
		getChatInput(token) { // Move from getters to action
			if (this.chatInput[token] === undefined) {
				this.restoreChatInput(token)
			}
			return this.chatInput[token]
		},
```
```suggestion
			Vue.set(this.chatInput, token, chatInput ?? '')
```

src/components/NewMessage/NewMessage.vue
```diff
@@ -333,6 +334,7 @@ export default {
 			clipboardTimeStamp: null,
 			typingInterval: null,
 			wasTypingWithinInterval: false,
+			debouncedUpdateChatInput: debounce(this.updateChatInput, 50)
```
Antreesy, line 337: Not sure methods were defined at this moment, but seems to work. Maybe increase it to 200ms?

src/components/NewMessage/NewMessage.vue
```diff
@@ -581,13 +578,25 @@ export default {
...
 
+		updateChatInput(text) {
+			if (this.messageToEdit) {
+				this.chatExtrasStore.setChatEditInput({
+					token: this.token,
+					text,
+					parameters: this.messageToEdit.messageParameters
+				})
+			} else if (text && text !== this.chatInput) {
+				this.chatExtrasStore.setChatInput({ token: this.token, text})
```
Antreesy, line 589: please lint

src/stores/chatExtras.js
```diff
@@ -130,6 +140,16 @@ export const useChatExtrasStore = defineStore('chatExtras', {
...
 		},
 
+		/**
+		 * Restore chat input from the browser storage and save to store
+		 *
+		 * @param {string} token The conversation token
+		 */
+		restoreChatInput(token) {
+			const chatInput = BrowserStorage.getItem('chatInput_' + token)
+			Vue.set(this.chatInput, token, chatInput ?? '')
```
Antreesy, line 150: Although you has empty strings only from here. Don't produce it here as well?

DorraJaouad: Empty string is needed in the component so we can't render null or undefined, but it will be better if we render it for getChatInput and not store it.

Antreesy (approved):
Rebased onto main
Reverted changes for `removeParentIdToReply`, as breaking replies
Tested with switching, reloading, editing, replies

PR #11587, opened 2024-02-14
feat(federation): Add TalkV1 federation proxy for mentions

nickvergessen:
Decision is to proxy the requests on the own server,
so we can proxy other services (matrix coming to mind)
in the future and the clients still have to only talk to
their own server, to help with CSP and to reduce the efforts

### 🚧 Tasks

**Legend:**
- 🟥 Must have
- 🟨 Should have
- 🟩 Good to have
- ⚫ Not for now

---

- [x] Avatar
    - [x] getAvatar 🟨
    - [x] getAvatarDark 🟨
    - [x] ~~uploadAvatar~~ ⚫ Moderation is not supported in TalkV1
    - [x] ~~emojiAvatar~~ ⚫ Moderation is not supported in TalkV1
    - [x] ~~deleteAvatar~~ ⚫ Moderation is not supported in TalkV1
- [ ] Chat
    - [x] receiveMessages 🟥
    - [x] sendMessage 🟥
    - [x] deleteMessage 🟥
    - [x] editMessage 🟥
    - [x] getMessageContext 🟥
    - [ ] setReadMarker 🟥
    - [ ] setReminder 🟨 
    - [ ] getReminder 🟨 
    - [ ] deleteReminder 🟨 
    - [ ] markUnread 🟨 
    - [x] mentions 🟨 
    - [ ] shareObjectToChat 🟩
    - [ ] getObjectsSharedInRoomOverview 🟩
    - [ ] getObjectsSharedInRoom 🟩
    - [x] ~~clearHistory~~ ⚫ Moderation is not supported in TalkV1
- [ ] Room
    - [x] getParticipants 🟥
    - [x] 🆕 capabilities 🟥
    - [ ] setSessionState 🟨
    - [ ] …

### 🏁 Checklist

- [x] ⛑️ Tests (unit and/or integration) are included or not possible
- [ ] 📘 API documentation in `docs/` has been updated or is not required
- [ ] 🔖 Capability is added or not needed

PR #11636, opened 2024-02-22
fix(MessagesGroup) : remove edit indicator for deleted messages

DorraJaouad:

| 🏚️ Before | 🏡 After |
|------------|----------|
| image     |image      |


nickvergessen:
I think we should do that on API level?

DorraJaouad:
> I think we should do that on API level?

By removing lastEdit* keys ? Yeah, it is also good.

nickvergessen:
Added to https://github.com/nextcloud/spreed/pull/11587

PR #11638, opened 2024-02-23
fix(messagesStore): don't update parent message in the store, if it hasn't been unchanged

Antreesy:
### ☑️ Resolves

* Fix excessive store updates:
  * When first time loading the chat, we already have the parent message up-to-date from the API
  * `message_deleted`, `message_edited`, `reaction...` requested at the same time (not new) contains same parent message without changes, so we don't need to update it again and trigger sideEffect (messagesList re-render)
  * deleteMessage, editMessage, addReaction API requests return parent to update it. When it comes again with message long polling, we also don't need to update it

No visual changes

### 🏁 Checklist

### DIFF
--- a/src/store/messagesStore.js
+++ b/src/store/messagesStore.js
@@ -516,8 +516,7 @@ const actions = {
 	/**
 	 * Adds message to the store.
 	 *
-	 * If the message has a parent message object,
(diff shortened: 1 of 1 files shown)


DorraJaouad (approved):
Makes sense

PR #11649, opened 2024-02-25
refactor(pinia): migrate breakoutRoomsStore to Pinia 🍍

Antreesy:
* Migrate breakoutRoomsStore to Pinia :pineapple: 
  * optimize some methods
  * rename **Action to normal ones

No visual changes

### 🚧 Tasks

- [x] Pinia migration
- [ ] Typescript migration - in follow-up
- [ ] Test coverage - in follow-up

### 🏁 Checklist


src/stores/breakoutRooms.js
```diff
@@ -0,0 +1,307 @@
...
+		 * @param {Array<object>|object} conversationOrArray a single conversation or an array of conversations.
+		 *
+		 */
+		processConversations(token, conversationOrArray) {
+			const conversations = Array.isArray(conversationOrArray) ? conversationOrArray : [conversationOrArray]
+			conversations.forEach(conversation => {
+				if (conversation.objectType === CONVERSATION.OBJECT_TYPE.BREAKOUT_ROOM) {
+					this.addBreakoutRoom(token, conversation)
+				}
+			})
```
DorraJaouad, line 80: Already in `'patchConversations'`

src/stores/breakoutRooms.js
```diff
@@ -0,0 +1,307 @@
...
+			try {
+				await broadcastMessageToBreakoutRooms(token, message)
+			} catch (error) {
+				console.error(error)
+				showError(t('spreed', 'An error occurred while sending a message to the breakout rooms'))
+			}
+		},
+
+		/**
+		 * Update a participants in breakout rooms for a given token.
```
DorraJaouad, line 235:
```suggestion
		 * Update participants in breakout rooms for a given token.
```

src/stores/breakoutRooms.js
```diff
@@ -0,0 +1,307 @@
...
+		 *
+		 * @param {string} token the parent room token.
+		 */
+		async fetchBreakoutRoomsParticipants(token) {
+			try {
+				const response = await getBreakoutRoomsParticipants(token)
```
DorraJaouad, line 241:
```suggestion
				const response = await fetchBreakoutRoomsParticipants(token)
```

Rename this to `fetchBreakoutRoomsParticipants` ?

src/stores/breakoutRooms.js
```diff
@@ -0,0 +1,307 @@
...
+		},
+
+		/**
+		 * Dismiss a notification about raised hand for a breakout room with given token.
+		 *
+		 * @param {string} token the breakout room token.
+		 */
+		async dismissRequestAssistance(token) {
+			try {
+				const response = await resetRequestAssistance(token)
```
DorraJaouad, line 281: same

DorraJaouad (commented):
Testing : 
image
Participants are not in their right room.

Also, is it expected to return room-not-found when reassigning a participant to the room where you are currently joining the call?

src/stores/breakoutRooms.js
Antreesy: Removed

Antreesy:
> Participants are not in their right room

Good catch, they might be received from different sources

> is it expected to return room-not-found when reassigning a participant to the room

Should be expected, as participant doesn't have access rights to the previous room anymore

DorraJaouad:
> Should be expected, as participant doesn't have access rights to the previous room anymore

* Alice is in call in room 1
* John in room 2 and moderator moves him to room 1 
* Alice gets page-not-found

Antreesy:
> * John in room 2 and moderator moves him to room 1
> * Alice gets page-not-found

I'd sat that's on API side. If you check system messages, you'll see that everyone got removed from BR, then only selected participants are added back

DorraJaouad (approved):
As follow-ups are there, this is ready

PR #11650, opened 2024-02-25
fix(MessageForwader): provide value to check where component is being used

Antreesy:
### ☑️ Resolves

* Follow-up to #11616 
  * When check how to handle redirect link, rely on App context, not the window.location or routes
  * Fine to keep as-is for stable releases, but ensure safer check process for future

## 🖌️ UI Checklist

No visual changes

### 🏁 Checklist

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -38,6 +38,7 @@
 
 <script>
 import debounce from 'debounce'
+import { provide } from 'vue'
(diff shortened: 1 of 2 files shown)


ShGKme (approved):
Just a note - provide/inject also works with Options API

DorraJaouad (approved):
Second time, better idea :)

PR #11651, opened 2024-02-25
build: provide Typescript support 

Antreesy:
### ☑️ Resolves

* Reuse existing options from `@nextloud/webpack-vue-config` to compile Typescript files (standalone and Vue SFC scripts)

No visual changes

### 🚧 Tasks

- [ ] Check compile options
- [ ] Check Eslint rules

### 🏁 Checklist

### DIFF
--- a/src/components/VolumeIndicator/VolumeIndicator.vue
+++ b/src/components/VolumeIndicator/VolumeIndicator.vue
@@ -39,7 +39,7 @@
 	</span>
 </template>
 
-<script>
+<script lang="ts">
 import Microphone from 'vue-material-design-icons/Microphone.vue'
 import MicrophoneOff from 'vue-material-design-icons/MicrophoneOff.vue'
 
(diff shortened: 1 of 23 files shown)


tsconfig.json
ShGKme, line 1: For proper typing it should include
```json
  "vueCompilerOptions": {
    "target": 2.7
  }
```

ShGKme: https://github.com/vuejs/language-tools/tree/master/extensions/vscode#usage

src/components/VolumeIndicator/VolumeIndicator.vue
```diff
@@ -129,17 +133,15 @@ export default {
...
 			}
 
-			return (this.size - this.iconOffsetBottom) * this.computeVolumeLevel()
+			return (this.iconSizeWithoutOffset) * this.computeVolumeLevel()
 		},
 	},
 
 	methods: {
-		computeVolumeLevel() {
+		computeVolumeLevel(): number {
```
ShGKme, line 141: Note, TS can determine types in many cases. Explicit type declarations is needed when it cannot be computed and on public interface to specify a contract.

ShGKme (approved):

PR #11660, opened 2024-02-27
Ignore label of data channel when processing received messages.

### LINKED ISSUE #10003: Janus 1.x compatibility problem
### How to use GitHub

---

There is an incompatibility issue using HPB with Janus 1.x (multichannel version).
The signaling server negotiates a Janus data channel, but this channel is not named as expected by Talk.
With Janus 0.x (non-multichannel version) the only possible data channel is named "JanusDataChannel" but with the new 1.x (multichannel version) the channels have not a fixed label anymore.
Talk will keep using only one data channel between any two peers, but these channel is labeled "1" on Janus 1.x.
Because there can be more than one channel between peers, Janus use the channel number as label and these channel will always be labeled "1" on Talk app.

As far as I could find out on the Talk code, there seems to be only two places where the data channel label is checked.
On line 232 of CallParticipantModel.js and on line 1591 of webrtc.js:

https://github.com/nextcloud/spreed/blob/6714e9b73d888c16a23049703ca812918825da6b/src/utils/webrtc/models/CallParticipantModel.js#L232C3-L232C3

https://github.com/nextcloud/spreed/blob/6714e9b73d888c16a23049703ca812918825da6b/src/utils/webrtc/webrtc.js#L1591C3-L1591C3

If you add "1" as a possible label of the data channel Talk will recognize the channel as valid and will understand the messages coming from it.

As Meetecho considers Janus 0.x as "legacy", sooner o later they will stop maintaining that branch so I think that this fix will be needed then.

**Additional context**
https://github.com/meetecho/janus-gateway/issues/2566

fancycode:
This is necessary for support of Janus 1.x which will use the (random) id of the publisher that sent the message as label.

### ☑️ Resolves

* Fix #10003

### Testing

- Setup HPB with Janux 1.x without this PR
- Notice "talking" indication is not working in the browser
- Apply changes from this PR
- Reload and restart call
- "talking" indication is working again

@ivansss @SystemKeeper @mahibi FYI for the mobile apps.

### DIFF
--- a/src/utils/webrtc/models/CallParticipantModel.js
+++ b/src/utils/webrtc/models/CallParticipantModel.js
@@ -228,10 +228,6 @@ CallParticipantModel.prototype = {
 			return
 		}
 
-		if (label !== 'status' && label !== 'JanusDataChannel') {
-			return
-		}
-
 		if (data.type === 'speaking') {
 			this.set('speaking', true)
 		} else if (data.type === 'stoppedSpeaking') {
--- a/src/utils/webrtc/webrtc.js
+++ b/src/utils/webrtc/webrtc.js
@@ -1594,25 +1594,21 @@ export default function initWebRtc(signaling, _callParticipantCollection, _local
 	})
 
 	webrtc.on('channelMessage', function(peer, label, data) {
-		if (label === 'status' || label === 'JanusDataChannel') {
-			if (data.type === 'audioOn') {
-				webrtc.emit('unmute', { id: peer.id, name: 'audio' })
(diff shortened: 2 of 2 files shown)


nickvergessen (approved):

Antreesy (approved):
OK code-wise

fancycode:
Would be nice if you could backport this so we can update Janus at some point and don't have to wait for customers to upgrade to newer versions of NC / Talk.

fancycode:
Thanks for the backports.

PR #11663, opened 2024-02-27
fix(MessagesList): Stop fetching when reaching the chat beginning

DorraJaouad:
### ☑️ Resolves

Old messages were fetched on every scroll event even though messages list weren't updated -> lost requests 

Create a flag to stop fetching when :
* We reach system messages conversation_created and history_cleared
* For chats that don't have such, we wait another request and compare if in both times the list didn't change then we confirm that the very beginning of the chat was reached ( this case applies to chats that have expiration messages, and history cleared after setting it)

No visual changes just spinner doesn't show up when you scroll up at the top.

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -169,6 +169,8 @@ export default {
 			debounceUpdateReadMarkerPosition: () => {},
 
 			debounceHandleScroll: () => {},
+
+			stopFetchingOldMessages: false,
 		}
 	},
 
@@ -268,6 +270,7 @@ export default {
 		token(newToken, oldToken) {
 			// Expire older messages when navigating to another conversation
 			this.$store.dispatch('easeMessageList', { token: oldToken })
+			this.stopFetchingOldMessages = false
 		},
 
 		messagesList: {
@@ -689,6 +692,9 @@ export default {
 		 * @param {boolean} includeLastKnown Include or exclude the last known message in the response
 		 */
 		async getOldMessages(includeLastKnown) {
+			if (this.stopFetchingOldMessages) {
+				return
+			}
 			// Make the request
 			this.loadingOldMessages = true
 			try {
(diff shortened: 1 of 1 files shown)


src/components/MessagesList/MessagesList.vue
```diff
@@ -268,6 +272,7 @@ export default {
 		token(newToken, oldToken) {
 			// Expire older messages when navigating to another conversation
 			this.$store.dispatch('easeMessageList', { token: oldToken })
+			this.stopFetchingOldMessages = false
```
Antreesy, line 275: Wanted to catch you here =)

src/components/MessagesList/MessagesList.vue
```diff
@@ -689,7 +694,11 @@ export default {
...
 		 */
 		async getOldMessages(includeLastKnown) {
+			if (this.stopFetchingOldMessages) {
+				return
+			}
 			// Make the request
+			const prevMessagesList = this.messagesList
 			this.loadingOldMessages = true
 			try {
 				await this.$store.dispatch('fetchMessages', {
```
Antreesy, line 704: we could also get a response here; then:
- check last message in the array `['conversation_created', 'history_cleared'].includes(response.data.ocs.data.at(-1)?.systemMessage)`
- check response status code (200 contains messages. 304 is empty - end of a list)

DorraJaouad: Good point 🦅 , I assumed response code would be 200 for empty list too.

nickvergessen:
> For chats that don't have such, we wait another request and compare if in both times the list didn't change then we confirm that the very beginning of the chat was reached ( this case applies to chats that have expiration messages, and history cleared after setting it)

This can also happen when you have a full page of "hidden system messages" e.g. reactions only

Antreesy (approved):
Tested for these 3 cases, should be enough to cover, and don't break anything

PR #11665, opened 2024-02-28
fix(Conversation): optimize error handling

Antreesy:
### ☑️ Resolves

* Splitting work on conversations system:
  * Template was optimised
  * Error handling / displaying moved from services to the store

### 🚧 Tasks

- [ ] align last message from the chat with last message in the left sidebar

### 🏁 Checklist

### DIFF
--- a/src/components/LeftSidebar/ConversationsList/Conversation.vue
+++ b/src/components/LeftSidebar/ConversationsList/Conversation.vue
@@ -20,133 +20,119 @@
 -->
 
 <template>
-	<Fragment>
-		<NcListItem ref="listItem"
-			:key="item.token"
-			:name="item.displayName"
-			:title="item.displayName"
-			class="conversation-item"
-			:class="{'unread-mention-conversation': item.unreadMention}"
-			:data-nav-id="`conversation_${item.token}`"
-			:actions-aria-label="t('spreed', 'Conversation actions')"
-			:to="to"
-			:bold="!!item.unreadMessages"
-			:counter-number="item.unreadMessages"
-			:counter-type="counterType"
-			@click="onClick">
-			<template #icon>
-				<ConversationIcon :item="item" :hide-favorite="false" :hide-call="false" />
-			</template>
-			<template #subname>
-				<strong v-if="item.unreadMessages"
-					class="subtitle">
-					{{ conversationInformation }}
-				</strong>
-				<template v-else>
-					{{ conversationInformation }}
+	<NcListItem ref="listItem"
+		:key="item.token"
+		:name="item.displayName"
+		:title="item.displayName"
+		:data-nav-id="`conversation_${item.token}`"
+		:actions-aria-label="t('spreed', 'Conversation actions')"
+		:to="to"
+		:bold="!!item.unreadMessages"
+		:counter-number="item.unreadMessages"
+		:counter-type="counterType"
+		@click="onClick">
+		<template #icon>
+			<ConversationIcon :item="item" :hide-favorite="false" :hide-call="false" />
+		</template>
+		<template #subname>
+			{{ conversationInformation }}
+		</template>
+		<template v-if="!isSearchResult" #actions>
+			<NcActionButton v-if="canFavorite"
+				key="toggle-favorite"
+				close-after-click
+				@click="toggleFavoriteConversation">
+				<template #icon>
+					<Star :size="20" :fill-color="!item.isFavorite ? '#FFCC00' : undefined" />
 				</template>
-			</template>
-			<template v-if="!isSearchResult" #actions>
-				<NcActionButton v-if="canFavorite"
-					:close-after-click="true"
-					@click="toggleFavoriteConversation">
-					<template #icon>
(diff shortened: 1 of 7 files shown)


src/components/LeftSidebar/ConversationsList/Conversation.vue
```diff
@@ -20,133 +20,112 @@
...
-					{{ conversationInformation }}
+	<NcListItem ref="listItem"
+		:key="item.token"
+		:name="item.displayName"
+		:title="item.displayName"
+		:data-nav-id="`conversation_${item.token}`"
+		:actions-aria-label="t('spreed', 'Conversation actions')"
+		:to="to"
+		:bold="!isActive && !!item.unreadMessages"
+		:counter-number="!isActive ? item.unreadMessages : 0"
```
DorraJaouad, line 31: If you want to hide the counter when clicked, It would give more sense if you clear the unread messages too. So when you click on another conversation, you won't find the counter is still there ( checked with other apps, they do the same).

src/components/LeftSidebar/ConversationsList/Conversation.vue
```diff
@@ -20,133 +20,112 @@
...
+		:counter-type="counterType"
+		@click="onClick">
+		<template #icon>
+			<ConversationIcon :item="item" :hide-favorite="false" :hide-call="false" />
+		</template>
+		<template #subname>
+			{{ !isActive ? conversationInformation : '' }}
```
DorraJaouad, line 38: I can't digest this tbh. I think it should stay when opened.

src/components/LeftSidebar/ConversationsList/Conversation.vue
```diff
@@ -20,133 +20,112 @@
...
-				<ConversationIcon :item="item" :hide-favorite="false" :hide-call="false" />
-			</template>
-			<template #subname>
-				<strong v-if="item.unreadMessages"
-					class="subtitle">
-					{{ conversationInformation }}
-				</strong>
-				<template v-else>
-					{{ conversationInformation }}
+	<NcListItem ref="listItem"
```
DorraJaouad, line 23: I am not sure how NcListItem is turning into primary color when active at the moment, but maybe we should also bin its `active` prop to `isActive` computed ?

Antreesy: It's internal Router Link path matching. Hard to get it from the Vue lib code

src/components/LeftSidebar/ConversationsList/Conversation.vue
Antreesy: cc @nickvergessen for the input

DorraJaouad: If there are any requests to avoid by hiding conversation Information, I think we can handle it from the front end while it is open and then update the API once it is not active. 
* Unread messages (counter)
* Last message

src/components/LeftSidebar/ConversationsList/Conversation.vue
DorraJaouad: Okay, let's keep it it as-is.

src/components/LeftSidebar/ConversationsList/Conversation.vue
Antreesy: reverted for now

nickvergessen:
> hide subline from active item

Not needed at the moment anymore with #11674 
My PR does not "save data" for now, so lets keep it.

src/services/conversationsService.js
```diff
@@ -81,46 +82,36 @@ const fetchNoteToSelfConversation = async function() {
...
-			}
-		}
-	}
-
-	return axios.get(generateOcsUrl('core/autocomplete/get'), Object.assign(options, {
+		!onlyUsers ? SHARE.TYPE.GROUP : null,
+		!onlyUsers ? SHARE.TYPE.CIRCLE : null,
+		(!onlyUsers && token !== 'new') ? SHARE.TYPE.EMAIL : null,
+		(!onlyUsers && token !== 'new' && loadState('spreed', 'federation_enabled')) ? SHARE.TYPE.REMOTE : null,
+	].filter(type => type !== null)
```
DorraJaouad, line 92: nice

src/components/LeftSidebar/ConversationsList/Conversation.vue
```diff
@@ -20,133 +20,112 @@
...
+				</template>
+				{{ t('spreed', 'Conversation settings') }}
+			</NcActionButton>
+
+			<NcActionButton v-if="item.canLeaveConversation" close-after-click @click="leaveConversation">
+				<template #icon>
+					<ExitToApp :size="16" />
+				</template>
+				{{ t('spreed', 'Leave conversation') }}
+			</NcActionButton>
```
DorraJaouad, line 72: Maybe add `key` to consecutive `NcActionButton` ?

DorraJaouad (approved):
Tested

PR #11664, opened 2024-02-27
feat(federation): Support mentioning federated users

nickvergessen:
### 🚧 Tasks

- [x] Fix mentioning federated users
- [x] Fix parent message actor information
- [x] Convert (federated) users in message parameters

### 🏁 Checklist

- [x] ⛑️ Tests (unit and/or integration) are included or not possible
- [x] 📘 API documentation in `docs/` has been updated or is not required
- [x] 🔖 Capability is added or not needed

PR #11676, opened 2024-02-28
fix(mentions): parse groups and federated user mentions

Antreesy:
### ☑️ Resolves

* Follow-up for #11664
* Requires https://github.com/nextcloud-libraries/nextcloud-vue/pull/5323 for narrow cases
* Allow to reverse-parse federated users and groups when:
  * copy formatted message
  * edit message

image

### 🏁 Checklist

- [x] :desktop_computer: Should not be risky for Desktop client

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -69,13 +69,14 @@ module.exports = {
 	globalSetup: resolve(__dirname, 'jest.global.setup.js'),
 
 	collectCoverageFrom: [
-		'<rootDir>/src/**/*.{js,vue}',
+		'<rootDir>/src/**/*.{js,ts,vue}',
 	],
 
 	testEnvironment: 'jest-environment-jsdom',
 
 	moduleFileExtensions: [
 		'js',
+		'ts',
 		'vue',
 	],
 
--- a/src/components/ConversationSettings/EditableTextField.vue
+++ b/src/components/ConversationSettings/EditableTextField.vue
@@ -87,7 +87,7 @@ import NcRichContenteditable from '@nextcloud/vue/dist/Components/NcRichContente
 import NcRichText from '@nextcloud/vue/dist/Components/NcRichText.js'
 import Tooltip from '@nextcloud/vue/dist/Directives/Tooltip.js'
 
-import { parseSpecialSymbols } from '../../utils/textParse.js'
+import { parseSpecialSymbols } from '../../utils/textParse.ts'
 
 export default {
 	name: 'EditableTextField',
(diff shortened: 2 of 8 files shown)


src/utils/textParse.js
```diff
@@ -32,11 +34,18 @@ function parseMentions(text, parameters) {
 	if (Object.keys(parameters).some(key => key.startsWith('mention'))) {
 		for (const [key, value] of Object.entries(parameters)) {
 			let mention = ''
-			if (value?.type === 'call') {
+
+			if (key.startsWith('mention-call') && value?.type === 'call') {
 				mention = '@all'
-			} else if (value?.type === 'user') {
+			} else if (key.startsWith('mention-federated-user') && value?.type === 'user') {
+				const server = value.server ?? getBaseUrl().split('//').pop()
```
nickvergessen, line 41: `pop()` sounds like a mistake. Let's speak about it tomorrow

DorraJaouad: The url can contain the server of the hosting party so not the one where the user is coming from, right ?

src/utils/textParse.js
```diff
@@ -32,11 +34,18 @@ function parseMentions(text, parameters) {
...
+				mention = `@"federated_user/${value.id}@${server}"`
+			} else if (key.startsWith('mention-group') && value?.type === 'user-group') {
+				mention = `@"group/${value.id}"`
```
DorraJaouad, line 44: This can be broken when group name has spaces
image

maybe change the regex to :
```js
text = text.replace(new RegExp(`\\b{${key}}\\b`, 'g'), mention)
```

src/utils/textParse.js
Antreesy: yes, but we need to omit `https` and keep `http`

nickvergessen: yeah just talked in the office, we need to keep the protocol if it's not `https`

src/utils/textParse.js
Antreesy: will check later

src/utils/textParse.js
DorraJaouad: so as I understood, for {mention-federated-user} and server key does not exists, it means that the user is from the hosting server ? This means that every mention in a federated room is like {mention-federated-user}.

nickvergessen: > for {mention-federated-user} and server key does not exists

it means the user is from your server which is a proxy server and **not** the host of the conversation

nickvergessen: `{mention-federated-user}` is just a placeholder and the name pattern is "agreed on", but it could be any random sting that is `a-z0-9-`

src/utils/textParse.js
Antreesy: Works correctly:
image

DorraJaouad: Only if there is no text after it. Any text that comes after the mention will be included in the quotation.

Antreesy:
amended tests for the util

PR #11678, opened 2024-02-28
Followup/11674/frontend utilizing the return data

nickvergessen:
### ☑️ Resolves

- [ ] Based on https://github.com/nextcloud/spreed/pull/11674
- [ ] Fix node/eslint/js


src/store/conversationsStore.js
```diff
@@ -634,9 +634,8 @@ const actions = {
 			return
 		}
 
-		await setConversationUnread(token)
-		commit('updateUnreadMessages', { token, unreadMessages: 1 })
-		await dispatch('fetchConversation', { token })
+		const response = await setConversationUnread(token)
+		await dispatch('addConversation', response.data.ocs.data)
```
DorraJaouad, line 638: Use `updateConversation` , although it is basically doing the same thing as `addConversation` for now but we can improve it later to patch the changes.

src/store/messagesStore.js
```diff
@@ -823,7 +823,7 @@ const actions = {
 	 */
 	async clearLastReadMessage(context, { token, updateVisually = false }) {
 		const conversation = context.getters.conversations[token]
-		if (!conversation || !conversation.lastMessage) {
+		if (!conversation/* FIXME || !conversation.lastMessage */) {
```
DorraJaouad, line 826: Why was it commented out ? 
It is still needed to not to cause an error of reading undefined. `conversation.lastMessage.id`

src/store/conversationsStore.js
nickvergessen: I have no idea what I'm doing :D

Antreesy:
With https://github.com/nextcloud/spreed/pull/11793 should be not needed to provide an empty/null id

So I'm keeping only response handling

DorraJaouad (approved):
Tested

PR #11696, opened 2024-03-01
Tighten MessageType and ActorType usage

provokateurin:
### ☑️ Resolves

I added the missing message types and tightened the types of the message type and actor type methods to ensure only valid values will be passed. It's not possible to do this with virtual methods (`@psalm-method` exists but doesn't seem to work).

### 🏁 Checklist


provokateurin:
Looks like the default values need to be kept, but that violates the enums and we know that the values will always be set.

nickvergessen:
> we know that the values will always be set.

Only later on. when you do `new X()` and call `getActorType()` on it, it returns the default value...

provokateurin:
This can sadly not be fixed in a proper way because of the empty string as a default value. If it was null then it wouldn't be a problem, but string enums with an empty string as default are bad.

PR #11716, opened 2024-03-03
fix(ChatManager): Add missing verbs

provokateurin:
### ☑️ Resolves

Replaces https://github.com/nextcloud/spreed/pull/11696 because it is not possible to properly type this throughout the whole codebase.

### 🏁 Checklist

### DIFF
--- a/lib/Chat/Parser/SystemMessage.php
+++ b/lib/Chat/Parser/SystemMessage.php
@@ -504,11 +504,11 @@ class SystemMessage implements IEventListener {
 				$parsedMessage = '{file}';
 				$metaData = $parameters['metaData'] ?? [];
 				if (isset($metaData['messageType'])) {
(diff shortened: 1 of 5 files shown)


lib/Chat/Parser/SystemMessage.php
```diff
@@ -490,11 +490,11 @@ protected function parseMessage(Message $chatMessage): void {
 				$metaData = $parameters['metaData'] ?? [];
 				if (isset($metaData['messageType'])) {
 					if ($metaData['messageType'] === 'voice-message') {
```
nickvergessen, line 492: Any clue why you changed only `setMessageType` calls but not the others?

nickvergessen: Also messageType is not the Verb iirc. but will check again

nickvergessen (approved):

PR #11724, opened 2024-03-05
feat(chat): Allow marking chat as read without a last message ID

nickvergessen:
This is required for federated conversations where the clients might not know the hosted message ID but only the proxy ID instead

### 🏁 Checklist

- [x] 🔖 Capability is added or not needed: `chat-read-last`


Antreesy (approved):
Looks good, need to rebase #11678 onto it

PR #11725, opened 2024-03-05
perf: improve performance of resolving recipient shares

icewind1991:
Improves the performance of applying per-user changes to room shares.

I wasn't sure how to handle `getSharesByIds` so I've just disabled the optimization for that code path for now.

See also: https://github.com/nextcloud/server/pull/44004


nickvergessen:
Pushed a commit to fix psalm

nickvergessen (approved):

PR #11727, opened 2024-03-05
fix(messages): minor render enhancements

Antreesy:
### ☑️ Resolves

- assign `viewId` on Options API initialising: chatIdentifier is no longer updated at `mounted()` - no cancelled requests
-  don't check system messages for parents:
    -  weird, but for federated conversations system conversations `getContext` at the moment response is processed, doesn't contain message parent - false check - system message is rendered as-is

🏚️ Before | 🏡 After
-- | --
image | image

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/store/messagesStore.js
+++ b/src/store/messagesStore.js
@@ -529,12 +529,15 @@ const actions = {
 	processMessage(context, { token, message }) {
 		const sharedItemsStore = useSharedItemsStore()
 
-		if (message.parent && message.systemMessage
-			&& (message.systemMessage === 'message_deleted'
+		if (message.systemMessage === 'message_deleted'
 				|| message.systemMessage === 'reaction'
 				|| message.systemMessage === 'reaction_deleted'
 				|| message.systemMessage === 'reaction_revoked'
-				|| message.systemMessage === 'message_edited')) {
+				|| message.systemMessage === 'message_edited') {
+			if (!message.parent) {
+				return
+			}
+
 			// If parent message is presented in store and is different, we update it
 			const parentInStore = context.getters.message(token, message.parent.id)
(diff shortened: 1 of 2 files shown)


Antreesy:
cc @nickvergessen 
please take a look at system messages from API side. Response in devtools has 'parent' fields, but some intermediate result is handled by store

src/store/messagesStore.js
```diff
@@ -529,15 +529,14 @@ const actions = {
...
-		if (message.parent && message.systemMessage
-			&& (message.systemMessage === 'message_deleted'
+		if (message.systemMessage === 'message_deleted'
 				|| message.systemMessage === 'reaction'
 				|| message.systemMessage === 'reaction_deleted'
 				|| message.systemMessage === 'reaction_revoked'
-				|| message.systemMessage === 'message_edited')) {
+				|| message.systemMessage === 'message_edited') {
 			// If parent message is presented in store and is different, we update it
 			const parentInStore = context.getters.message(token, message.parent.id)
```
nickvergessen, line 538: but here you use `message.parent` already?

nickvergessen:
> chatIdentifier is no longer updated at mounted()

This was required for some things when switching from chat UI to call UI. Can not remember the details though

Antreesy:
> Can not remember the details though

Was introduced here: https://github.com/nextcloud/spreed/commit/7503ecffaff8faea7effeeb6bd1c0f343ff0d367
By description, doesn't seem to happen anymore, init it earlier just saves us one re-render and request. Tested with switching between call - chat views

nickvergessen (approved):

PR #11733, opened 2024-03-07
fix(conversation): skip unread marker increasing from notification

Antreesy:
### ☑️ Resolves

* There could be a race condition between room and notification fetch
* If room comes first, unread marker is already updated there, no need to do it again
* For federated rooms, lastMessage.id is not presented, so checked by timestamp

[Screencast from 07.03.2024 12:25:29.webm](https://github.com/nextcloud/spreed/assets/93392545/8bd3a0f8-890a-4d3a-bca5-d91992e51fb8)

[Screencast from 07.03.2024 13:13:34.webm](https://github.com/nextcloud/spreed/assets/93392545/64ca819d-d496-43a4-a51a-5ffff9fde14d)

### 🏁 Checklist

### DIFF
--- a/src/store/conversationsStore.js
+++ b/src/store/conversationsStore.js
@@ -722,6 +722,11 @@ const actions = {
 		}
 
 		const conversation = Object.assign({}, getters.conversations[token])
+		if (conversation.lastMessage.id === parseInt(messageId, 10)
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):
🐘

PR #11739, opened 2024-03-07
fix(federation): Correctly convert message on deleting

nickvergessen:
### 🚧 Tasks

- [x] System message 'message_deleted' parent doesn't transform user in federated conversation 

### 🏁 Checklist


Antreesy (approved):
Tested before

PR #11742, opened 2024-03-07
fix(federation): Don't expose the accessToken anymore to the frontend/clients

nickvergessen:
### 🏁 Checklist


src/stores/federation.js
```diff
@@ -121,7 +121,6 @@ export const useFederationStore = defineStore('federation', {
 			Vue.delete(this.pendingShares[id], 'loading')
 			Vue.set(this.acceptedShares, id, {
 				...this.pendingShares[id],
-				accessToken: conversation.remoteAccessToken,
```
nickvergessen, line 124: cc @Antreesy

Antreesy: Used in the store, bu we don't rely on it

Antreesy (approved):

PR #11769, opened 2024-03-11
fix(ConversationIcon): force light background color for conversation …

DorraJaouad:
…type when using light theme

### ☑️ Resolves

It is aligned with used theme.

🏚️ Before | 🏡 After
-- | --
image | image

### 🚧 Tasks

- [ ] Code review

### 🏁 Checklist

### DIFF
--- a/src/components/TopBar/TopBar.vue
+++ b/src/components/TopBar/TopBar.vue
@@ -20,7 +20,7 @@
 -->
 
 <template>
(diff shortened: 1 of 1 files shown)


Antreesy (approved):
Not sure that important is needed, but let's keep it just in case

DorraJaouad:
> Not sure that important is needed, but let's keep it just in case

Might be needed because colors are already set for the element and parent's style can be skipped.

PR #11777, opened 2024-03-12
fix(federation): Expose local token with the invite so UI can render proxy avatars

### LINKED ISSUE #11722: Federated invitations follow-ups
- [x] use NcRichText / render mention bubble for remote invitee
  - [x] requires access to host server avatar endpoint
  - [x] requires author rich parameters to be passed with notification / invitations endpoint
- [x] show conversation avatar
  - [x] avatar is already proxied from the host server to local room
  - [x] requires local token to be passed with notification / invitations endpoint

image

nickvergessen:
- Fix #11722

### 🚧 Tasks

- [x] Fix missing token so the UI can start the avatar
- [x] Handle invites as "OK" on the avatar endpoint even when not accepted yet

### 🏁 Checklist

### DIFF
--- a/lib/Controller/AEnvironmentAwareController.php
+++ b/lib/Controller/AEnvironmentAwareController.php
@@ -28,6 +28,7 @@ declare(strict_types=1);
 namespace OCA\Talk\Controller;
 
 use OC\AppFramework\Http\Dispatcher;
+use OCA\Talk\Model\Invitation;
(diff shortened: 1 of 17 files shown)


Antreesy:
Will take care of frontend part

nickvergessen:
Fixed the access check, works fine for me now 👍🏼

PR #11779, opened 2024-03-12
fix(MessagesList): remove deleted groups

DorraJaouad:
### ☑️ Resolves

- Removal update was missing for date groups level. So clear history didn't remove past groups by date.

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -412,14 +412,20 @@ export default {
 		},
 
 		softUpdateByDateGroups(oldDateGroups, newDateGroups) {
-			// Check if we have this group in the old list already and it is unchanged
-			Object.keys(newDateGroups).forEach(dateTimestamp => {
-				if (oldDateGroups[dateTimestamp]) {
-					// the group by date has changed, we update its content (groups by author)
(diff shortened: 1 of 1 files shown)


src/components/MessagesList/MessagesList.vue
```diff
@@ -412,14 +412,20 @@ export default {
...
+			dateTimestamps.forEach(dateTimestamp => {
+				if (newDateGroups[dateTimestamp]) {
+					if (oldDateGroups[dateTimestamp]) {
+						// the group by date has changed, we update its content (groups by author)
+						this.softUpdateAuthorGroups(oldDateGroups[dateTimestamp], newDateGroups[dateTimestamp], dateTimestamp)
+					} else {
+						// the group is new
+						this.messagesGroupedByDateByAuthor[dateTimestamp] = newDateGroups[dateTimestamp]
+					}
+				} else if (oldDateGroups[dateTimestamp]) {
```
Antreesy, line 426:
```suggestion
				} else {
```
Not sure we need this check. If we reach this block, timestamp is definitely given by oldDateGroups

Antreesy (approved):

src/components/MessagesList/MessagesList.vue
DorraJaouad: Yeah, it is not needed at the moment

PR #11782, opened 2024-03-12
fix(participants): move interval update of talking time counter to the store

### LINKED ISSUE #11501: Talking time is transferred to other participants
### How to use GitHub

---

## Steps to reproduce
1. Be in call with someone speaking
2. Check "talking time"

### Expected behaviour
Timer should reflect correct individual value

### Actual behaviour
Some of the speaking signals were transferred to wrong participants (See screenshot for  user 3, 4)
image

Might be a regression of #11254

## Talk app

**Talk app version:** 18

Antreesy:
### ☑️ Resolves

* Fix #11501
	* Data in the store are correct and accessible by token + attendeeId
	* Because of virtual scroller in the participant list, interval update in the component can't be used, as it attached to the Vue component object, not the participant data
	* Moved interval update to the store, so each second we update a value there
	* Component rely only on store update and re-rendered once per second (as before), but with correct value

🏚️ Before | 🏡 After
-- | --
Reproducible | Non-reproducible

### 🚧 Tasks

- [ ] speaking don't rely on participantStore, so may be moved to `callExtras` store :pineapple: 
  - [ ] Check if SpeakingStatusHandler class supports Pinia

### 🏁 Checklist

### DIFF
--- a/src/components/RightSidebar/Participants/Participant.vue
+++ b/src/components/RightSidebar/Participants/Participant.vue
@@ -80,7 +80,7 @@
 
 		<!-- Phone participant dial action -->
 		<div v-if="isInCall && canBeModerated && isPhoneActor"
-			id="participantNavigationId"
+			:id="participantNavigationId"
 			class="participant-row__dial-actions">
 			<NcButton v-if="!participant.inCall"
 				type="success"
@@ -464,8 +464,6 @@ export default {
 			isUserNameTooltipVisible: false,
 			isStatusTooltipVisible: false,
 			permissionsEditor: false,
-			speakingInterval: null,
-			timeSpeaking: null,
 			disabled: false,
 		}
 	},
@@ -864,22 +862,17 @@ export default {
 			}
 			return ''
 		},
-	},
-	watch: {
-		isParticipantSpeaking(speaking) {
-			if (speaking) {
-				if (!this.speakingInterval) {
-					this.speakingInterval = setInterval(this.computeElapsedTime, 1000)
-				}
-			} else {
-				if (speaking === undefined) {
-					this.timeSpeaking = 0
-				}
-				clearInterval(this.speakingInterval)
-				this.speakingInterval = null
-			}
-		},
 
+		timeSpeaking() {
+			if (!this.participantSpeakingInformation || this.isParticipantSpeaking === undefined) {
+				return 0
+			}
+
+			return this.participantSpeakingInformation.totalCountedTime
+		},
+	},
+
+	watch: {
 		phoneCallStatus(value) {
 			if (!value || !(value === 'ringing' || value === 'accepted')) {
 				this.disabled = false
@@ -1009,20 +1002,6 @@ export default {
 				}
 			} catch (error) {
 				showError(t('spreed', 'Could not modify permissions for {displayName}', { displayName: this.computedName }))
-			}
-		},
-
-		computeElapsedTime() {
-			if (!this.participantSpeakingInformation) {
-				return null
-			}
-
-			const { speaking, lastTimestamp, totalCountedTime } = this.participantSpeakingInformation
-
-			if (!speaking) {
-				this.timeSpeaking = totalCountedTime
-			} else {
-				this.timeSpeaking = Date.now() - lastTimestamp + totalCountedTime
 			}
 		},
 
--- a/src/components/RightSidebar/Participants/Participant.vue
+++ b/src/components/RightSidebar/Participants/Participant.vue
@@ -704,7 +704,7 @@ export default {
 		},
 
 		participantSpeakingInformation() {
-			return this.$store.getters.getParticipantSpeakingInformation(this.token, this.attendeeId)
+			return this.$store.getters.getParticipantSpeakingInformation(this.attendeeId)
 		},
 
 		isParticipantSpeaking() {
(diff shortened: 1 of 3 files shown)


src/components/RightSidebar/Participants/Participant.vue
```diff
@@ -864,22 +862,20 @@ export default {
...
-				clearInterval(this.speakingInterval)
-				this.speakingInterval = null
+
+		timeSpeaking() {
+			if (!this.participantSpeakingInformation) {
+				return null
+			}
+			if (this.isParticipantSpeaking === undefined) {
+				return 0
 			}
```
DorraJaouad, line 872:
```suggestion
			if (!this.participantSpeakingInformation || !this.isParticipantSpeaking) {
				return null
```

No need to differentiate, boolean value of `timeSpeaking` will be the same

src/store/participantsStore.js
```diff
@@ -414,7 +416,7 @@ const mutations = {
 	 * @param {string} data.attendeeId - the attendee ID of the participant in conversation.
 	 * @param {boolean} data.speaking - whether the participant is speaking or not
 	 */
-	setSpeaking(state, { token, attendeeId, speaking }) {
+	setSpeakingTime(state, { token, attendeeId, speaking }) {
```
DorraJaouad, line 419: I suggest to keep the name and apply the following

```js
	setSpeaking(state, { token, attendeeId, speaking }) {
		// create a dummy object for current call
		if (!state.speaking[token]) {
			Vue.set(state.speaking, token, {})
		}

		if (!state.speaking[token][attendeeId]) {
			Vue.set(state.speaking[token], attendeeId, { speaking, lastTimestamp: 0, totalCountedTime: 0 })
			return
		}
			
		state.speaking[token][attendeeId].speaking = speaking
	},
```

src/store/participantsStore.js
```diff
@@ -433,9 +435,56 @@ const mutations = {
...
+		// create a dummy object for current call
+		if (!state.speakingInterval[token]) {
+			Vue.set(state.speakingInterval, token, {})
+		}
+		if (!state.speakingInterval[token][attendeeId]) {
+			Vue.set(state.speakingInterval[token], attendeeId, interval)
+		} else if (!interval) {
+			clearInterval(state.speakingInterval[token][attendeeId])
+			state.speakingInterval[token][attendeeId] = interval
+		}
```
DorraJaouad, line 462:
```suggestion
		if (!interval) {
			clearInterval(state.speakingInterval[token][attendeeId])
		}

		Vue.set(state.speakingInterval[token], attendeeId, interval)
```

src/store/participantsStore.js
```diff
@@ -433,9 +435,56 @@ const mutations = {
...
+	 */
+	updateTimeSpeaking(state, { token, attendeeId }) {
+		if (!state.speaking[token]?.[attendeeId]) {
+			return
+		}
+		if (!state.speaking[token][attendeeId].speaking) {
+			clearInterval(state.speakingInterval[token][attendeeId])
+			state.speakingInterval[token][attendeeId] = null
+			return
+		}
```
DorraJaouad, line 481:
```suggestion
		if (!state.speaking[token][attendeeId].lastTimestamp) {
			state.speaking[token][attendeeId].lastTimestamp = Date.now()
			return
		}
```

src/store/participantsStore.js
```diff
@@ -1036,8 +1093,19 @@ const actions = {
...
+		if (speaking) {
+			interval = setInterval(() => {
+				context.commit('updateTimeSpeaking', { token, attendeeId })
+			}, 1000)
+		} else {
+			interval = null
+		}
+
+		context.commit('setSpeakingTime', { token, attendeeId, speaking })
+		context.commit('setSpeakingInterval', { token, attendeeId, interval })
```
DorraJaouad, line 1108:
```suggestion
		const interval= speaking ? setInterval(() => {
			context.commit('updateTimeSpeaking', { token, attendeeId })
		})
		: null
		context.commit('setSpeakingTime', { token, attendeeId, speaking })
		context.commit('setSpeakingInterval', { token, attendeeId, interval })
```

DorraJaouad: So, the changes are meant to reduce the redundancy of clearing interval and make each mutation specified to a purpose/object.

src/components/RightSidebar/Participants/Participant.vue
Antreesy: formattedTime expects `number`, and `this..isParticipantSpeking` could be `false`. But good catch anyway

DorraJaouad (commented):
One last time 😬

src/store/participantsStore.js
```diff
@@ -1038,6 +1079,26 @@ const actions = {
 
 	setSpeaking(context, { token, attendeeId, speaking }) {
 		context.commit('setSpeaking', { token, attendeeId, speaking })
+		context.commit('updateTimeSpeaking', { token, attendeeId, speaking, interval: false })
+
+		if (!context.state.speakingInterval) {
```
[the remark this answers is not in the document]
Antreesy: First signal can be `false`, but I'm agree, there's no need to start timers yet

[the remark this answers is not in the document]
DorraJaouad: Yes, I noticed a signal is received from remote participants with speaking false then undefined at the beginning of the call.

PR #11786, opened 2024-03-12
chore(typescript): migrate utils to TS

Antreesy:
### ☑️ Resolves

* Add typings
* Increase test coverage

### 🏁 Checklist


DorraJaouad (approved):
Good

PR #11792, opened 2024-03-13
fix(federation): Respect setReadMarker=1 on proxied ChatController::r…

nickvergessen:
…eceiveMessage()

### 🏁 Checklist


SystemKeeper (approved):
I think we should update the docs that this only works when `lookIntoFuture` is also set

SystemKeeper:
Yea, guess that makes sense in the long run

PR #11799, opened 2024-03-14
fix(federation): Handle federation configs when inviting instead of g…

nickvergessen:
…eneric "Can not reach remote"

## 🛠️ API Checklist

Should allow for better error handling in the frontend in the future.
Returned `error` codes:
- cloudId (is not well formed)
- outgoing (outgoing shares are disabled)
- federation (federation is disabled for the user)
- trusted_servers (the cloud id is not from a trusted server but the restriction for it is enabled)

### 🏁 Checklist


lib/Federation/BackendNotifier.php
```diff
@@ -84,52 +79,25 @@ public function sendRemoteShare(
...
 		/** @var IUser $roomOwner */
 		$roomOwner = $this->userManager->get($roomOwnerAttendee->getActorId());
 
-		$invitedCloudId = $user . '@' . $remote;
-		$remote = $this->prepareRemoteUrl($remote);
+		$remote = $this->prepareRemoteUrl($invitedCloudId->getRemote());
 
 		$share = $this->cloudFederationFactory->getCloudFederationShare(
-			$user . '@' . $remote,
+			$invitedCloudId->getUser() . '@' . $remote,
```
SystemKeeper, line 100: Would it make sense to move that to the interface in the long run?

SystemKeeper, line 100: Would it make sense to move that to the interface in the long run?

lib/Federation/RestrictionValidator.php
```diff
@@ -0,0 +1,89 @@
...
+			throw new \InvalidArgumentException('outgoing');
+		}
+
+		if (!$this->talkConfig->isFederationEnabledForUserId($user)) {
+			$this->logger->debug('Talk federation not allowed for user ' . $user->getUID());
+			throw new \InvalidArgumentException('federation');
+		}
+
+		if ($this->appConfig->getAppValueBool('federation_only_trusted_servers')) {
+			if (!$this->appManager->isEnabledForUser('federation')) {
```
SystemKeeper, line 73: Is it correct to only check this when `federation_only_trusted_servers` is set? When the app is not enabled for the user, it should never work?

SystemKeeper (approved):

lib/Federation/RestrictionValidator.php
nickvergessen: > When the app is not enabled for the user, it should never work?

`cloud_federation_api` is needed to "federate", the federation is for "federating the system address book and having trusted servers" .... :see_no_evil:

PR #11811, opened 2024-03-15
fix(avatars): show Talk-defined avatar thumbnails in modals

Antreesy:
### ☑️ Resolves

* Fix inconsistent avatars in modals
  * Use `modal-mask` class from Vue library instead of `talk-modal`
* Update classes for NcRichContenteditable (`mention-bubble` and `autocomplete-result`)
* Fix dark mode for AvatarWrapper

🏚️ Before | ☀️ Light theme | 🌑 Dark Theme
-- | -- | --
image | image | image
image | image | image
image | image | image

### 🏁 Checklist

### DIFF
--- a/css/icons.css
+++ b/css/icons.css
@@ -139,28 +139,3 @@ body[data-theme-dark] .icon-user-forced-white {
 	background-image: url(../img/app-dark.svg);
 	filter: var(--background-invert-if-dark);
 }
-
-/* To be used in new conversation / invitation handler dialogs */
-.conversation-icon .icon-conversation-group.icon--dark {
-	background-image: url('../img/icon-conversation-group-dark.svg')
-}
-
-.conversation-icon .icon-conversation-group.icon--bright {
-	background-image: url('../img/icon-conversation-group-bright.svg')
-}
-
-.conversation-icon .icon-conversation-public.icon--dark {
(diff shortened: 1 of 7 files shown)


Antreesy:
Fixing poll results now

css/icons.css
```diff
@@ -116,10 +116,20 @@
...
+
+.talk-modal .avatar.icon.icon-contacts.icon--dark,
 .conversation-icon .icon-conversation-group.icon--dark {
 	background-image: url('../img/icon-conversation-group-dark.svg')
 }
 
+.talk-modal .avatar.icon.icon-contacts.icon--bright,
 .conversation-icon .icon-conversation-group.icon--bright {
 	background-image: url('../img/icon-conversation-group-bright.svg')
 }
```
DorraJaouad, line 135: Why don't you omit talk-modal class?  It will be applied on all avatar wrappers at once without the need to specify a parent class, unless there are some avatars that need to stay with the old background ?

Antreesy: Maybe we should reconsider how the whole style list from that class works)

DorraJaouad: Yes, we take the avatar style based on itself only regardless of its parent or where it was rendered.

DorraJaouad (approved):
Tested, everything is good.

PR #11837, opened 2024-03-17
Fix(MessagesList): Limit relative date up to a week.

DorraJaouad:
### ☑️ Resolves

* Relative dates are better shown until a week ago (as it used to be).

🏚️ Before | 🏡 After
-- | --
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -561,8 +561,10 @@ export default {
 				return t('spreed', 'Today')
 			case 1:
 				return t('spreed', 'Yesterday')
+			case 7:
+				return t('spreed', 'A week ago')
 			default:
-				return t('spreed', '{n} days ago', { n: diffDays })
+				return n('spreed', '%n day ago', '%n days ago', diffDays)
 			}
 		},
 
@@ -574,15 +576,22 @@ export default {
 		 */
 		generateDateSeparator(dateTimestamp) {
 			const date = moment.unix(dateTimestamp).startOf('day')
-			// <Today>, <November 11th, 2019>
-			return t('spreed', '{relativeDate}, {absoluteDate}', {
-				relativeDate: this.getRelativePrefix(date),
-				// 'LL' formats a localized date including day of month, month
-				// name and year
-				absoluteDate: date.format('LL'),
-			}, undefined, {
-				escape: false, // French "Today" has a ' in it
-			})
+			// <Today>, <March 18th, 2024>
+			// Relative date is only shown until a week ago
+			if (moment().startOf('day').diff(date, 'days') <= 7) {
+				return t('spreed', '{relativeDate}, {absoluteDate}', {
+					relativeDate: this.getRelativePrefix(date),
+					// 'LL' formats a localized date including day of month, month
+					// name and year
+					absoluteDate: date.format('LL'),
+				}, undefined, {
+					escape: false, // French "Today" has a ' in it
+				})
+			} else {
+				// <March 18th, 2024>
+				return t('spreed', '{absoluteDate}', { absoluteDate: date.format('LL') })
+			}
+
 		},
 
 		/**


src/components/MessagesList/MessagesList.vue
```diff
@@ -561,6 +561,8 @@ export default {
 				return t('spreed', 'Today')
 			case 1:
 				return t('spreed', 'Yesterday')
+			case 7:
+				return t('spreed', 'A week ago')
 			default:
 				return t('spreed', '{n} days ago', { n: diffDays })
```
nickvergessen, line 567:
```suggestion
 				return n('spreed', .... { n: diffDays })
```
Needs to use the plural function

Don't know out of my head how it is in JS

But actually there should be a relative time function already which you can use directly?

nickvergessen (changes requested):

src/components/MessagesList/MessagesList.vue
Antreesy: There is something for it in moment lib

DorraJaouad: > There is something for it in moment lib

Yeah, but it still needs some customization. We need a combination of `date.fromNow()` and `moment().calendar()` because the first shows the past days correctly ( except for yesterday and today, it shows in hours) and the second shows yesterday and tomorrow ( but along with hh:mm).

nickvergessen: Or we simply use what the activity app does? https://github.com/nextcloud/activity/blob/master/src/components/ActivityGroup.vue#L32

DorraJaouad: That's still not a default output from the lib. We are not avoiding the hurdle of custom anyhow.

We keep it as simple as it is now :) ? but it's better to use `date.fromNow()` instead of `t('spreed', '{n} days ago', { n: diffDays })` indeed

src/components/MessagesList/MessagesList.vue
```diff
@@ -574,15 +576,22 @@ export default {
...
-			return t('spreed', '{relativeDate}, {absoluteDate}', {
-				relativeDate: this.getRelativePrefix(date),
-				// 'LL' formats a localized date including day of month, month
-				// name and year
-				absoluteDate: date.format('LL'),
-			}, undefined, {
-				escape: false, // French "Today" has a ' in it
-			})
+			// <Today>, <March 18th, 2024>
+			// Relative date is only shown until a week ago
```
Antreesy, line 580:
```suggestion
			// Relative date is only shown up to a week ago (inclusive)
```

src/components/MessagesList/MessagesList.vue
```diff
@@ -574,15 +576,22 @@ export default {
...
+				return t('spreed', '{relativeDate}, {absoluteDate}', {
+					relativeDate: this.getRelativePrefix(date),
+					// 'LL' formats a localized date including day of month, month
+					// name and year
+					absoluteDate: date.format('LL'),
+				}, undefined, {
+					escape: false, // French "Today" has a ' in it
+				})
+			} else {
+				// <March 18th, 2024>
```
Antreesy, line 591:
```suggestion
				// TRANSLATORS: <March 18th, 2024>
```

src/components/MessagesList/MessagesList.vue
```diff
@@ -574,15 +576,22 @@ export default {
...
-				// name and year
-				absoluteDate: date.format('LL'),
-			}, undefined, {
-				escape: false, // French "Today" has a ' in it
-			})
+			// <Today>, <March 18th, 2024>
+			// Relative date is only shown until a week ago
+			if (moment().startOf('day').diff(date, 'days') <= 7) {
+				return t('spreed', '{relativeDate}, {absoluteDate}', {
+					relativeDate: this.getRelativePrefix(date),
```
Antreesy, line 583:
```suggestion
					relativeDate: this.getRelativePrefix(date, diffDays),
```
To avoid calculation twice

Antreesy (approved):
Works as expected, left a couple of comments

nickvergessen (approved):

PR #11843, opened 2024-03-18
fix(API): Use Room::isFederatedConversation() consistantly

nickvergessen:
### 🏁 Checklist


lib/Chat/SystemMessage/Listener.php
```diff
@@ -401,8 +401,7 @@ protected function fixMimeTypeOfVoiceMessage(ShareCreatedEvent|BeforeDuplicateSh
 			return;
 		}
 		$room = $this->manager->getRoomByToken($share->getSharedWith());
-		if ($room->getRemoteServer() !== '') {
-			// FIXME this should be blocked up front
```
nickvergessen, line 405: Here is the unequal line count of the diff

SystemKeeper (approved):

Issue #11484, opened 2024-01-29
Collapsed messages | minor bugs


Antreesy:
- Duplicated messages won't cleared up from the list:
![image](https://github.com/nextcloud/spreed/assets/93392545/75da5de7-4193-4b13-8c29-8d1abe6345ba)

- Messages won't collapse once unread marker is moved from them, only after reload

## Talk app

**Talk app version:** 18

PR #11876, opened 2024-03-20
fix(MessagesSystemGroup): collapse `actor removed` messages, collapse previously open groups with unread marker

Antreesy:
### ☑️ Resolves

* Ref #11484 
  * message group should be collapsed, as soon as unread marker is moved from it
  * only manually open groups will be opened after list re-render, rest will follow the rules

Rest
* group messages ilke `Administrator removed A, B and 3 other participants` (copy-paster from 'user-added')
* cycle through only existing group keys when soft updating
  * we could remove group just before in previous cycle

Before | After
--|--
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -233,6 +233,17 @@ export default {
 			default: undefined,
 		},
 		/**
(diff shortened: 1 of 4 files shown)


DorraJaouad (approved):
I didn't test unread marker position but logic looks good 🦅.

Tested the rest ✅

PR #11893, opened 2024-03-22
feat(LeftSidebar): rearrange conversation fast creating options

### LINKED ISSUE #11632: "Create new conversation" shows above more useful conversation results in search
### How to use GitHub

---

Currently when searching for a talk room, for example "design" or "design team", you can not press enter or click the first entry below the search (which is usually the best result) to enter the conversation. If you do press enter or pick the first result, a new empty conversation is created.

The "Create new conversation" suggestion of course is nice so it should stay, but it should not take precedence over existing conversations. It would be nice to just be able to press enter or click the first result (usually the most fitting one) to go into the conversation.
**A sensible solution to take these requirements into account seems to be to simply move the "Create new conversation" suggestion to the end of the "Conversations" list in the results.**
(This incidentally also reduces the height and noise of sections – and it further seems that the whitespace between the section headings and the results can be reduced.)

Current | Proposal
-|-
image|image

Antreesy:
### ☑️ Resolves

* Fix #11632
  * Move "Create new conversation" under current conversations
  * Add sublines to indicate that user is starting a new conversation from search results
  * Decrease margin between sections
* Show user status for known users in search results
* Add fast option to create group conversation with federated user from search results
  * API provides entrypoint, would be a good base for follow-up 1-1 federated rooms

🏚️ Before | 🏡 After | 🏚️ Before | 🏡 After
-- | -- | -- | --
image | image | image | image
image | image | image | image
Not available | image | - | -

### 🏁 Checklist

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -182,19 +182,6 @@
 
 				<!-- Search results -->
 				<ul v-else class="h-100 scroller">
-					<!-- Create a new conversation -->
-					<NcListItem v-if="canStartConversations"
-						:name="t('spreed', 'Create a new conversation')"
-						data-nav-id="conversation_create_new"
-						@click="createConversation(searchText)">
-						<template #icon>
-							<ChatPlus :size="30" />
-						</template>
-						<template #subname>
-							{{ searchText }}
-						</template>
-					</NcListItem>
-
 					<!-- Search results: user's conversations -->
 					<NcAppNavigationCaption :name="t('spreed', 'Conversations')" />
 					<Conversation v-for="item of searchResultsConversationList"
@@ -203,6 +190,19 @@
 						:item="item"
 						@click="abortSearch" />
 					<Hint v-if="searchResultsConversationList.length === 0" :hint="t('spreed', 'No matches found')" />
+
+					<!-- Create a new conversation -->
+					<NcListItem v-if="canStartConversations"
+						:name="searchText"
+						data-nav-id="conversation_create_new"
+						@click="createConversation(searchText)">
+						<template #icon>
+							<ChatPlus :size="44" />
+						</template>
+						<template #subname>
+							{{ t('spreed', 'New group conversation') }}
+						</template>
+					</NcListItem>
 
 					<!-- Search results: listed (open) conversations -->
 					<template v-if="!listedConversationsLoading && searchResultsListedConversations.length !== 0">
@@ -225,6 +225,9 @@
 							<template #icon>
 								<ConversationIcon :item="iconData(item)" />
 							</template>
+							<template #subname>
+								{{ t('spreed', 'New private conversation') }}
+							</template>
 						</NcListItem>
 					</template>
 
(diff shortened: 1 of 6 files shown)


src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -255,6 +263,26 @@
...
+								</template>
+							</NcListItem>
+						</template>
+
+						<!-- New conversations: Federated users -->
+						<template v-if="searchResultsFederated.length !== 0">
+							<NcAppNavigationCaption :name="t('spreed', 'Federated users')" />
+							<NcListItem v-for="item of searchResultsFederated"
+								:key="`circle_${item.id}`"
+								:data-nav-id="`circle_${item.id}`"
```
DorraJaouad, line 277:
```suggestion
								:key="`federated_${item.id}`"
								:data-nav-id="`federated_${item.id}`"
```

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -255,6 +263,26 @@
...
+								:name="item.label"
+								@click="createAndJoinConversation(item)">
+								<template #icon>
+									<AvatarWrapper v-bind="iconData(item)" />
```
DorraJaouad, line 281: `conversationIcon` ? like users

Antreesy: It doesn't contain logic for fetching proxy users (and I don't want to add it there)

DorraJaouad:
I think we lost arrow navigation 🙈

Antreesy:
> I think we lost arrow navigation 🙈

Yeah, due to vue-lib changes (again)

Rewrite code, so we don't depend on selector (which could be changed), only on presense of `data-nav-id`

Antreesy:
> I think we lost arrow navigation 🙈

Yeah, due to vue-lib changes (again)

Rewrite code, so we don't depend on selector (which could be changed), only on presense of `data-nav-id`

src/composables/useArrowNavigation.js
```diff
@@ -125,11 +134,21 @@ export function useArrowNavigation(listElementRef, defaultElementRef, selector,
...
+
+		if (itemElement.matches(focusableCondition)) {
+			itemElement.focus()
+		}
+
+		try {
+			itemElement.querySelector(focusableCondition).focus()
+		} catch (e) {
+			console.warn('Nav element does not have any focusable children')
+		}
```
DorraJaouad, line 151:
```suggestion
		if (itemElement.matches(focusableCondition)) {
			itemElement.focus()
		} else { 
			try {
				itemElement.querySelector(focusableCondition).focus()
			} catch (e) {
				console.warn('Nav element does not have any focusable children')
			}
		}

			
```

DorraJaouad: I am not sure if we should do that anyway, navigation should navigate on item level only ? 
and when the element is not focusable, increment index and try again.

Antreesy: That's for the cases you have element like list item (with your approach, you won't find any element to focus):

```html
<li>
  <div>
    <a /> // native focus should be triggered on this
  </div>
</li>
```

New method seems more reliable to me, as it doesn't depends on some external component changes. Although it is visible then trying to navigate through participant list, but that's because items are not tabbable there initially

DorraJaouad (approved):
Tested ✅

PR #11894, opened 2024-03-22
refactor(breakoutRooms): migrate service to typescript

Antreesy:
### ☑️ Resolves

* Follow-up to #11649 
* TODO: Test coverage

### 🏁 Checklist


src/stores/breakoutRooms.ts
```diff
@@ -131,7 +131,8 @@ export const useBreakoutRoomsStore = defineStore('breakoutRooms', {
 				await this.fetchBreakoutRoomsParticipants(token)
 
 				// Open the sidebar and switch to the breakout rooms tab
-				emit('spreed:select-active-sidebar-tab', 'breakout-rooms')
+				// FIXME upstream https://github.com/nextcloud-libraries/nextcloud-event-bus/pull/734
+				emit('spreed:select-active-sidebar-tab', 'breakout-rooms' as unknown as object)
```
DorraJaouad, line 135: but it's not needed, it is always taken as a string.

DorraJaouad (approved):
Types checked and they are correct ✅

src/stores/breakoutRooms.ts
Antreesy: TSC omplains here, as library don't expect a string yet

PR #11900, opened 2024-03-22
fix(federation): Change the Talk-Hash header when overwriting capabil…

### LINKED ISSUE #11865: Overwriting capability setting does not invalide proxy hash
In https://github.com/nextcloud/spreed/pull/11846 we started overwriting some of the remote capabilities. As this is done after setting the proxy-hash, it will not change the provided proxy-hash. That means for everything we introduce/change in later versions of Talk, clients will not refetch the room capabilities.

nickvergessen:
…ities

### ☑️ Resolves

* Fix #11865 

### 🏁 Checklist

### DIFF
--- a/lib/Controller/RoomController.php
+++ b/lib/Controller/RoomController.php
@@ -40,6 +40,7 @@ use OCA\Talk\Exceptions\RoomNotFoundException;
 use OCA\Talk\Exceptions\UnauthorizedException;
 use OCA\Talk\Federation\Authenticator;
 use OCA\Talk\Federation\FederationManager;
+use OCA\Talk\Federation\Proxy\TalkV1\ProxyRequest;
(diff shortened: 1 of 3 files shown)


SystemKeeper:
Why not doing the has calculation at

https://github.com/nextcloud/spreed/blob/455231a6d140f0f2ba0d6794fa3b333c7b9b1a7d/lib/Controller/RoomController.php#L2276-L2278 

?

Also, wouldn't it be more clear if we create an array with our changes (much like in `overwrittenRemoteTalkHash`) and then use the array for the hash update and use `array_merge` for the resulting capabilities ?

nickvergessen:
Because we also deal with the hash in 2 other places and it sounds to risky to rehash in x places

SystemKeeper (approved):

PR #11923, opened 2024-03-25
fix(openapi): Remove long int-lists now that OpenAPI understands int<min,max>

nickvergessen:
### 🏁 Checklist


src/types/openapi/openapi.ts
```diff
@@ -1501,9 +1501,9 @@ export type operations = {
...
+        forcePermissions?: number | null;
```
nickvergessen, line 1506: @Antreesy it seems the TS convertor does not pick up the min and max of integers

Antreesy: TS itself doesn't support ranges. Ideally If that would be done on converter side, in worst case: re-defined in spreed

Antreesy (approved):
Looks better for now from API side. Maybe it should be possible to change configuration for TS transpiler to restore these union types

PR #11934, opened 2024-03-26
chore(MessagesList): Refactor scrolling

DorraJaouad:
## Includes

- Restrict Placeholder to not to create scrolling for the chat when it is there
- Message Highlight is event-driven now instead of ref-oriented, saves some computations
- Adjust fallback visible message to include only actual visible messages (e.g: not reaction system messages)

Follow-up:
- Removed extra steps in scrolling and improved remaining.

## TO-DO
- [ ] ⚠️ Hide list before it is scrolled (when opening a conversation)
- [ ] ⚠️Reverse order of loading so widgets don't cause scrolling jumping
- [ ] 🔴 Empty content is not shown anymore because of loadedMessages flag does not change to true in empty conv
- [ ] 💚 Addition: add a floating button at the bottom in case you are scrolled up and new messages come in ( aka indicator for new messages while opening the chat in top)

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -165,11 +165,6 @@ export default {
 
 			isFocusingMessage: false,
 
-			/**
-			 * Quick edit option to fall back to the loading history and then new messages
-			 */
-			loadChatInLegacyMode: getCapabilities()?.spreed?.config?.chat?.legacy || false,
-
 			destroying: false,
 
 			expirationInterval: null,
@@ -211,7 +206,6 @@ export default {
 
 		showLoadingAnimation() {
 			return !this.$store.getters.isMessageListPopulated(this.token)
-				&& !this.messagesList.length
 		},
 
 		showEmptyContent() {
@@ -661,7 +655,7 @@ export default {
 				if (this.$store.getters.getFirstKnownMessageId(this.token) === null) {
 					let startingMessageId = 0
 					// first time load, initialize important properties
-					if (this.loadChatInLegacyMode || focusMessageId === null) {
+					if (focusMessageId === null) {
 						// Start from unread marker
 						this.$store.dispatch('setFirstKnownMessageId', {
 							token: this.token,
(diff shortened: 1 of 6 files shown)


src/components/MessagesList/MessagesList.vue
```diff
@@ -686,53 +712,21 @@ export default {
...
-						// get history before last read message
-						await this.getOldMessages(true)
-						// at this stage, the read marker will appear at the bottom of the view port since
-						// we haven't fetched the messages that come after it yet
-						// TODO: should we still show a spinner at this stage ?
-
-					} else {
-						// Get chat messages before last read message and after it
-						await this.getMessageContext(startingMessageId)
-						const startingMessageFound = this.focusMessage(startingMessageId, false, focusMessageId !== null)
```
Antreesy, line 699: Not sure it was working before (because of nextTick inside) =P

But that looks like a good fallback (without calling a focus method), I think we should keep it in some form (if context message id was a collapsed sys.message ot reaction, you won't find it in the DOM)
```js
	if (!document.getElementById(`message_${startingMessageId}`)) {
```

src/components/MessagesList/MessagesList.vue
```diff
@@ -615,36 +610,67 @@ export default {
...
+			if (!element) {
+				// Message id doesn't exist
+				console.warn('Message to focus not found in DOM', messageId)
+				return false // element not found
+			}
+
+			console.debug('Scrolling to a focused message programmatically')
+			this.isFocusingMessage = true
+
+			this.$nextTick(async () => {
```
Antreesy, line 632:
```suggestion
```

As you checking the element in the DOM syncronously, you shouldn't wait for the next tick

src/components/MessagesList/MessagesList.vue
```diff
@@ -686,53 +712,21 @@ export default {
...
 				this.isInitialisingMessages = false
 
 				// get new messages
 				await this.lookForNewMessages()
 
-				if (this.loadChatInLegacyMode || focusMessageId === null) {
-					// don't scroll if lookForNewMessages was polling as we don't want
-					// to scroll back to the read marker after receiving new messages later
-					if (!hasScrolled) {
-						this.scrollToFocusedMessage()
```
Antreesy, line 733: Will be handled by messagesList update, OK

src/components/MessagesList/MessagesList.vue
```diff
@@ -615,36 +610,67 @@ export default {
...
 				// scroll to message in URL anchor
-				isFocused = this.focusMessage(focusMessageId, false)
+				isFocused = this.focusMessage(focusMessageId)
+				return
 			}
 
-			if (!isFocused && this.visualLastReadMessageId) {
+			if (this.visualLastReadMessageId) {
 				// scroll to last read message if visible in the current pages
 				isFocused = this.focusMessage(this.visualLastReadMessageId, false, false)
```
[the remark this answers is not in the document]
DorraJaouad: You can see the first unread message, and `readmarker` should be updated based on that.

src/components/MessagesList/MessagesList.vue
DorraJaouad: I think it is not a good fallback, there is still a chance that the message before the last read message is also a hidden system message, it still has the same probability of `not-found` .

src/components/MessagesList/MessagesList.vue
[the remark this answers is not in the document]
Antreesy: Isn't unread marker already at the place by that time?

src/components/MessagesList/MessagesList.vue
Antreesy: Maybe we can rework getter to return only ids of visible messages?

src/components/MessagesList/MessagesList.vue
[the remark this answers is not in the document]
DorraJaouad: image


I consider you saw the message after the marker (if it is fully visible) so it should be updated.

src/components/MessagesList/MessagesList.vue
DorraJaouad: That would make more sense and we don't need all visible messages. We can add a filter of skipping reaction system messages and others that are invisible in getter until we reach a visible one, that's what we render.

Antreesy (approved):

PR #11937, opened 2024-03-26
fix(Quote): restrict overflow content for quotes / replies

Antreesy:
### ☑️ Resolves

* Fix broken styles for Quote components

🏚️ Before | 🏡 After
-- | --
Screenshot from 2024-03-26 15-48-25 | Screenshot from 2024-03-26 15-48-12
Screenshot from 2024-03-26 15-46-59 | Screenshot from 2024-03-26 15-46-47

### 🏁 Checklist


DorraJaouad (approved):
I recommend using
`text-overflow: ellipsis; `

PR #11939, opened 2024-03-26
fix(FilePicker): provide correct container for FilePicker

Antreesy:
### ☑️ Resolves

* Fix cases when FilePicker is shown behind the modal

hard to reproduce

### 🏁 Checklist

### DIFF
--- a/src/components/SettingsDialog/SettingsDialog.vue
+++ b/src/components/SettingsDialog/SettingsDialog.vue
@@ -45,6 +45,15 @@
 					@click="showFilePicker = true">
 					{{ t('spreed', 'Browse …') }}
 				</NcButton>
+				<FilePickerVue v-if="showFilePicker"
+					:name="t('spreed', 'Select location for attachments')"
+					:path="attachmentFolder"
+					container=".app-settings-section__wrapper"
+					:buttons="filePickerButtons"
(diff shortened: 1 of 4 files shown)


src/components/SettingsDialog/SettingsDialog.vue
```diff
@@ -45,6 +45,15 @@
...
 				</NcButton>
+				<FilePickerVue v-if="showFilePicker"
+					:name="t('spreed', 'Select location for attachments')"
+					:path="attachmentFolder"
+					container=".app-settings-section__wrapper"
+					:buttons="filePickerButtons"
+					:multiselect="false"
+					:mimetype-filter="['httpd/unix-directory']"
+					allow-pick-directory
+					@close="showFilePicker = false" />
```
DorraJaouad, line 56: I don't think there is a difference in putting here or in its previous position in DOM

Antreesy: It will be mounted in specified class selector anyway, but I think it would be better to keep them closer with other logic

DorraJaouad (approved):
Tested ✅ 
Though I would suggest to improve container binding later and not to keep so many hard-coded ones 🙈

PR #11941, opened 2024-03-27
fix(NcModal): don't close modals if they were filled

Antreesy:
### ☑️ Resolves

* Fix accidental closing by clicking outside of modals, which require a decent investment of time to fill 

### 🏁 Checklist

### DIFF
--- a/src/components/NewConversationDialog/NewConversationDialog.vue
+++ b/src/components/NewConversationDialog/NewConversationDialog.vue
@@ -24,6 +24,7 @@
 		<!-- New group form -->
 		<NcModal v-show="page !== 2"
 			class="new-group-conversation"
+			:close-on-click-outside="!isFilled"
 			:container="container"
 			@close="closeModal">
 			<h2 class="new-group-conversation__header">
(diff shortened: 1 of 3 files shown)


src/components/NewConversationDialog/NewConversationDialog.vue
```diff
@@ -227,6 +228,11 @@ export default {
 				|| this.conversationName.length > CONVERSATION.MAX_NAME_LENGTH
 				|| this.newConversation.description.length > CONVERSATION.MAX_DESCRIPTION_LENGTH
 		},
+
+		isFilled() {
+			return JSON.stringify(this.newConversation) !== JSON.stringify(NEW_CONVERSATION)
+				|| this.listable !== CONVERSATION.LISTABLE.NONE || this.isAvatarEdited
```
nickvergessen, line 234: description? password?

but I think name should be enough, most people will start with filling that one

Antreesy: evrything you don't see here is included in `this.newConversation` object

nickvergessen (approved):

PR #11943, opened 2024-03-27
Follow-up: chat scrolling refactoring

DorraJaouad:
### ☑️ Resolves

* follow-up to #11934

## TO-DO
- [x] 👷🏻‍♀️ Refactor 
- [x] ⚠️ Fix conversation jumping on opening
- [x] 🔴 Empty content is not shown anymore because of loadedMessages flag does not change to true in empty conv
- [ ] 💚 Addition: add a floating button at the bottom in case you are scrolled up and new messages come in ( aka indicator for new messages while opening the chat in top)

Conversation switching before

https://github.com/nextcloud/spreed/assets/84044328/480f1393-a9ad-4912-8fde-f2e5248ef616

Conversation switching after

https://github.com/nextcloud/spreed/assets/84044328/34e1e285-68e1-456e-a919-6e8ded5537ef

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -78,7 +78,6 @@ import uniqueId from 'lodash/uniqueId.js'
 import Message from 'vue-material-design-icons/Message.vue'
 
 import Axios from '@nextcloud/axios'
-import { getCapabilities } from '@nextcloud/capabilities'
 import { subscribe, unsubscribe } from '@nextcloud/event-bus'
 import moment from '@nextcloud/moment'
 
@@ -241,7 +240,7 @@ export default {
 				return false
 			}
 
-			return !!this.$store.getters.findParticipant(this.token, this.$store.getters.getParticipantIdentifier())
+			return !!this.$store.getters.findParticipant(this.token, this.conversation)
 		},
 
 		isInLobby() {
@@ -644,7 +643,6 @@ export default {
 
 		async handleStartGettingMessagesPreconditions() {
 			if (this.token && this.isParticipant && !this.isInLobby) {
-
 				// prevent sticky mode before we have loaded anything
 				this.isInitialisingMessages = true
 				const focusMessageId = this.getMessageIdFromHash()
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -613,31 +613,33 @@ export default {
 			let isFocused = null
 			if (focusMessageId) {
 				// scroll to message in URL anchor
-				isFocused = this.focusMessage(focusMessageId, false)
+				this.focusMessage(focusMessageId)
+				return
 			}
 
-			if (!isFocused && this.visualLastReadMessageId) {
+			if (this.visualLastReadMessageId) {
 				// scroll to last read message if visible in the current pages
 				isFocused = this.focusMessage(this.visualLastReadMessageId, false, false)
 			}
 
-			// TODO: in case the element is not in a page but does exist in the DB,
-			// we need to scroll up / down to the page where it would exist after
-			// loading said pages
-
 			if (!isFocused) {
-				// if no anchor was present or the message to focus on did not exist,
-				// scroll to bottom
-				this.scrollToBottom({ force: true })
+				// Safeguard: scroll to before last read message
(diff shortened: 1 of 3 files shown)


DorraJaouad:
`ChatIdentifier` is updated on two steps:
1. `{token: 't3qj6xtm', isParticipant: false, isInLobby: false}`
Chat is not scrolled yet because , the condition in `handleStartGettingMessagesPreconditions` is faulty
2. `{token: 't3qj6xtm', isParticipant: true, isInLobby: false}`
this is when the chat is finally scrolled

`isParticipant` is not updated instantly, need to find workarounds to make it available

DorraJaouad:
Focused message from URL doesn't focus correctly sometimes 🥲

src/components/MessagesList/MessagesList.vue
```diff
@@ -643,8 +642,7 @@ export default {
 		},
 
 		async handleStartGettingMessagesPreconditions() {
-			if (this.token && this.isParticipant && !this.isInLobby) {
-
+			if (this.token && !this.isInLobby) {
```
Antreesy, line 648: I would avoid to skip that part, that could lead to some unexpected things.
Though as we knowing some data from conversation, we could do like this (conversation contains the same params as partticipantIdentifier for that conversation, and already known to us):

```js
isParticipant() {
	if (!this.conversation) {
		return false
	}

	return !!this.$store.getters.findParticipant(this.token, this.conversation)
},
```

src/store/messagesStore.js
```diff
@@ -1410,6 +1410,10 @@ const actions = {
 	async easeMessageList(context, { token }) {
 		context.commit('easeMessageList', { token })
 	},
+
+	loadedMessagesOfConversation(context, { token }) {
```
Antreesy, line 1414: Good addition! Please check if we need to clean it up from the store somewhere (like on deleteConversation)

Antreesy (commented):
Checked only the last commits (unrelated to scrolling logic):

src/store/messagesStore.js
DorraJaouad: Nah, it is used to mark the first loading of a conversation so it should be kept there until page reload.

src/components/MessagesList/MessagesList.vue
DorraJaouad: Perfect 😀

src/components/MessagesList/MessagesList.vue
```diff
@@ -302,8 +301,11 @@ export default {
 					this.softUpdateByDateGroups(this.messagesGroupedByDateByAuthor, newGroups)
 				}
 
-				// scroll to bottom if needed
-				this.scrollToBottom({ smooth: true })
+				// Scroll only if there is no focused messages
+				const focusedMessage = this.getMessageIdFromHash()
```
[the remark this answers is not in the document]
DorraJaouad: discussed

src/components/MessagesList/MessagesList.vue
```diff
@@ -613,36 +615,36 @@ export default {
 			let isFocused = null
 			if (focusMessageId) {
 				// scroll to message in URL anchor
-				isFocused = this.focusMessage(focusMessageId, false)
+				return this.focusMessage(focusMessageId)
```
Antreesy, line 618: returned value would be unexpected here
```suggestion
				this.focusMessage(focusMessageId)
				return
```

src/components/MessagesList/MessagesList.vue
```diff
@@ -653,70 +655,27 @@ export default {
...
-						this.$nextTick(() => {
-							this.scrollToFocusedMessage(focusMessageId)
-						})
-					}
-				}
+				this.$nextTick(() => {
+					// basically scrolling to either the last read message or the message in the URL anchor
+					// and there is a fallback to scroll to the bottom if the message is not found
+					this.scrollToFocusedMessage(focusMessageId)
+				})
```
Antreesy, line 672: Should prevent unintentional scrollToBottom
```suggestion
					this.isInitialisingMessages = false
				})
```

Antreesy (approved):
As discussed during pair-review, seems to work much better. Seems that even the opening of every chat at the top is resolved! One more thing to try to reproduce and fix:

PR #11947, opened 2024-03-27
style(MessagesList): extend chat width to 70em

Antreesy:
### ☑️ Resolves

* Set width according to W3C recommendations
* Noticable if chat width starting from 800px (1520px with open LeftSidebar and RightSidebar)

🏚️ Before | image
-- | --
🏡 After | image

### 🏁 Checklist


src/assets/variables.scss
```diff
@@ -20,11 +20,14 @@
...
-//messages list max width
-$messages-list-max-width: 800px;
-
-//message utils width
-$message-utils-width: 100px;
+/** Messages list dimensions:
+ * - text max width: 70em (recommended by W3C standard)
+ * - utils-width: 52px (group avatar with paddings) + 120px (date and checkmark) + 4*4px (date paddings)
+ * - list max width: text width + utils width + 4 + 8 (paddings)
+ */
```
DorraJaouad, line 27: Can you please add totals in px here ? so not to recalculate this every time we check it later :p

DorraJaouad (approved):

src/assets/variables.scss
Antreesy: Done

PR #11990, opened 2024-04-02
fix(SystemMessage): Use link parameter for Circle URL in rich object parameters

provokateurin:
### ☑️ Resolves

According to https://github.com/nextcloud/server/blob/9aa7225ed3e78c2e418c983e65bdc1f1787b9940/lib/public/RichObjectStrings/Definitions.php#L227 only `link` exists and `url` doesn't.

### 🏁 Checklist


provokateurin:
I'm not sure how relevant this fix is and if it should be backported or not.

nickvergessen (approved):
Works :see_no_evil: 
grafik

PR #11991, opened 2024-04-02
fix(OpenAPI): Set rich object size parameter to integer

provokateurin:
### ☑️ Resolves

https://github.com/nextcloud/spreed/blob/1e0bebf4dea14021081f183e35a61a27f6b04de8/lib/Chat/Parser/SystemMessage.php#L803 is always the result of `$node->getSize()` and therefore an integer.

### 🏁 Checklist

### DIFF
--- a/lib/ResponseDefinitions.php
+++ b/lib/ResponseDefinitions.php
@@ -78,7 +78,7 @@ namespace OCA\Talk;
  *     'message-id'?: string,
  *     boardname?: string,
  *     stackname?: string,
- *     size?: string,
+ *     size?: int,
  *     path?: string,
  *     mimetype?: string,
  *     'preview-available'?: 'yes'|'no',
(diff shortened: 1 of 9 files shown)


provokateurin:
Again, I don't know how important this fix is so please decide yourself if a backport is necessary.

nickvergessen:
Im not sure. I think we should cast it to string. Because it is actually float or integer iirc.
And all parameters should be strings as per the definition 🙈 
https://github.com/nextcloud/server/blob/master/lib/public/RichObjectStrings/Definitions.php#L337

provokateurin:
I interpreted the example to be wrong and it should be an integer instead. I guess either way is possible, we just need to know how it was originally intended.
The other thing is the float type which I honestly don't understand. Is this a fix to avoid overflows on 32-bit systems by using double values?

PR #11992, opened 2024-04-02
fix(MessagesSystemGroup): Follow-up correct collapsed property 

DorraJaouad:
### ☑️ Resolves

- regression from  #11876

Both screenshots have last read message is the last system message shown

🏚️ Before | 🏡 After
-- | --
image | image

### 🚧 Tasks

TO-DO next:
- [ ] `MessagesSystemGroup` component should contain only one group (1 combined message and its collapsed messages)
- [ ]  Move `groupIsCollapsed` logic to `MessagesList`, so it can maintain manual collapsed state when switching between conversations.

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesGroup/MessagesSystemGroup.vue
+++ b/src/components/MessagesList/MessagesGroup/MessagesSystemGroup.vue
@@ -42,7 +42,7 @@
 					:key="message.id"
 					v-bind="message"
 					:token="token"
-					is-collapsed-system-message
+					:is-collapsed-system-message="messagesCollapsed.messages?.length > 1"
 					:last-collapsed-message-id="messagesCollapsed.lastId"
 					:next-message-id="getNextMessageId(message)"
 					:previous-message-id="getPrevMessageId(message)" />
@@ -119,7 +119,12 @@ export default {
 			immediate: true,
 			handler(value) {
 				this.messagesGroupedBySystemMessage = this.groupMessages(value)
+				this.updateCollapsedState()
 			},
+		},
+
+		lastReadMessageId() {
+			this.updateCollapsedState()
 		},
 	},
 
@@ -198,6 +203,22 @@ export default {
 			return ''
 		},
 
+		updateCollapsedState() {
+			for (const group of this.messagesGroupedBySystemMessage) {
+				const isLastReadInsideGroup = this.lastReadMessageId >= group.id && this.lastReadMessageId < group.lastId
+				if (isLastReadInsideGroup) {
+					// If the last read message is inside the group, we should show the group expanded
+					group.collapsed = false
+				} else if (this.groupIsCollapsed[group.id] !== undefined) {
+					// If the group was collapsed before, we should keep it collapsed
+					group.collapsed = this.groupIsCollapsed[group.id]
+				} else {
+					// If the group is not collapsed, we should collapse it if it contains more than one message
+					group.collapsed = group.messages.length > 1
+				}
+			}
+		},
+
 		groupMessages(messages) {
 			const groups = []
 			let lastMessage = null
@@ -205,34 +226,29 @@ export default {
 			for (const message of messages) {
 				const groupingType = this.messagesShouldBeGrouped(message, lastMessage)
 				if (!groupingType || forceNextGroup) {
-					groups.push({ id: message.id, lastId: message.id, messages: [message], type: '', collapsed: this.groupIsCollapsed[message.id] ?? true })
+					// Adding a new group
+					groups.push({ id: message.id, lastId: message.id, messages: [message], type: '', collapsed: undefined })
 					forceNextGroup = false
 				} else {
+					// Adding a message to the existing group
+
 					if (groupingType === 'call_reconnected') {
-						groups.push({ id: message.id, lastId: message.id, messages: [groups.at(-1).messages.pop()], type: '', collapsed: this.groupIsCollapsed[message.id] ?? true })
+						groups.push({ id: message.id, lastId: message.id, messages: [groups.at(-1).messages.pop()], type: '', collapsed: undefined })
 						groups.at(-1).lastId = groups.at(-1).messages.at(-1).id
 						forceNextGroup = true
 					}
 					groups.at(-1).messages.push(message)
 					groups.at(-1).lastId = message.id
 					groups.at(-1).type = groupingType
-
-					// Check if last read message is hidden inside the collapsed group, and open it, if so.
-					// Otherwise, combined system message will show a marker
-					const isLastReadInsideGroup = this.lastReadMessageId >= groups.at(-1).id && this.lastReadMessageId < groups.at(-1).lastId
-					if (isLastReadInsideGroup) {
-						groups.at(-1).collapsed = false
-					}
 				}
 				lastMessage = message
 			}
-
 			return groups
 		},
 
-		toggleCollapsed(messages) {
-			this.$set(messages, 'collapsed', !messages.collapsed)
-			this.groupIsCollapsed[messages.id] = !this.groupIsCollapsed[messages.id]
+		toggleCollapsed(group) {
+			this.$set(group, 'collapsed', !group.collapsed)
+			this.groupIsCollapsed[group.id] = group.collapsed
 		},
 
 		getNextMessageId(message) {


src/components/MessagesList/MessagesGroup/MessagesSystemGroup.vue
```diff
@@ -223,6 +230,11 @@ export default {
 					if (isLastReadInsideGroup) {
 						groups.at(-1).collapsed = false
 					}
+
+					// Update collapsed state for the last group if it's not defined yet
+					if (JSON.stringify(message) === JSON.stringify(messages.at(-1)) && groups.at(-1).collapsed === undefined) {
+						groups.at(-1).collapsed = groups.at(-1).messages.length > 1
+					}
```
ShGKme, line 237: Do we need deep equal check via JSON stringify here? Isn't `message` an object from `messages`?

ShGKme: Could you give an example, in which case do we need this `if`?

DorraJaouad: > Do we need deep equal check via JSON stringify here? Isn't message an object from messages?

Yes, for objects comparison ( JS compares by ref) , it will be always faulty here (they are not referencing to the same object) unless you use JSON stringify  which makes them strings (primitive type) to reach the expected comparison. 

> Could you give an example, in which case do we need this if?

It is needed to evaluate the last group as the loop won't go to the first conditional statement in the last message of the last group (given that it has more than 1 message) .  This is if group is not recorded in `groupIsCollapsed` (it was not manually toggled before) , is not `call_reconnected` type and it does not include unread message. It is like a fallback statement so one single message is not marked as collapsed (it can never be collapsed) and it used to be marked collapsed.

One thing to reduce is just putting `groups.at(-1).collapsed = true` instead because condition is passed for the last group that has more than one message.

src/components/MessagesList/MessagesGroup/MessagesSystemGroup.vue
```diff
@@ -205,9 +205,16 @@ export default {
...
+					// Adding a new group
+					groups.push({ id: message.id, lastId: message.id, messages: [message], type: '', collapsed: this.groupIsCollapsed[message.id] })
 					forceNextGroup = false
 				} else {
+					// Adding a message to the existing group
+
 					if (groupingType === 'call_reconnected') {
 						groups.push({ id: message.id, lastId: message.id, messages: [groups.at(-1).messages.pop()], type: '', collapsed: this.groupIsCollapsed[message.id] ?? true })
```
Antreesy, line 219: should we drop `?? true` here as well?

src/components/MessagesList/MessagesGroup/MessagesSystemGroup.vue
```diff
@@ -223,6 +230,11 @@ export default {
...
 						groups.at(-1).collapsed = false
 					}
+
+					// Update collapsed state for the last group if it's not defined yet
+					if (JSON.stringify(message) === JSON.stringify(messages.at(-1)) && groups.at(-1).collapsed === undefined) {
+						groups.at(-1).collapsed = groups.at(-1).messages.length > 1
+					}
 				}
 				lastMessage = message
 			}
```
Antreesy, line 240: Instead of stringifying last messages every, i think it would be more performant to move out all logic to another cycle, there we iterate over compiled group and decide on each one, whether it's supposed to be collapsed
```suggestion
			}
			
			for (const group of groups) {
				// Iterate over all groups and apply collapsed state once (adjust your logic)
				group.collapsed = this.groupIsCollapsed[message.id] ?? <isLastReadInsideGroup || messages.length > 1>
			}
```

src/components/MessagesList/MessagesGroup/MessagesSystemGroup.vue
```diff
@@ -232,7 +244,7 @@ export default {
 
 		toggleCollapsed(messages) {
 			this.$set(messages, 'collapsed', !messages.collapsed)
-			this.groupIsCollapsed[messages.id] = !this.groupIsCollapsed[messages.id]
+			this.groupIsCollapsed[messages.id] = messages.collapsed
```
Antreesy, line 247: should be also the opposite value, like above? `!messages.collapsed`

Antreesy: Apparently, not at the moment:

image

src/components/MessagesList/MessagesGroup/MessagesSystemGroup.vue
```diff
@@ -198,41 +203,52 @@ export default {
 			return ''
 		},
 
+		updateCollapsedState() {
+			for (const group of this.messagesGroupedBySystemMessage) {
```
Antreesy, line 207: I think it would be better to make it without mutations of reactive data. Could also try to do it with `Array.reduce()`:

```suggestion
		updateCollapsedState(groups) {
			return groups.reduce((acc, value) => cb, [])
```

src/components/MessagesList/MessagesGroup/MessagesSystemGroup.vue
```diff
@@ -198,41 +203,52 @@ export default {
...
 		groupMessages(messages) {
 			const groups = []
 			let lastMessage = null
 			let forceNextGroup = false
 			for (const message of messages) {
 				const groupingType = this.messagesShouldBeGrouped(message, lastMessage)
 				if (!groupingType || forceNextGroup) {
-					groups.push({ id: message.id, lastId: message.id, messages: [message], type: '', collapsed: this.groupIsCollapsed[message.id] ?? true })
+					// Adding a new group
+					groups.push({ id: message.id, lastId: message.id, messages: [message], type: '', collapsed: group.messages.length > 1 })
```
Antreesy, line 230: `ReferenceError: group is not defined`. Even if it will be, length here is 1, so doesn't make sense. Just omit it, as you going to iterate over all groups anyway 

```suggestion
					groups.push({ id: message.id, lastId: message.id, messages: [message], type: '' })
```

src/components/MessagesList/MessagesGroup/MessagesSystemGroup.vue
```diff
@@ -198,41 +203,52 @@ export default {
...
 					forceNextGroup = false
 				} else {
+					// Adding a message to the existing group
+
 					if (groupingType === 'call_reconnected') {
-						groups.push({ id: message.id, lastId: message.id, messages: [groups.at(-1).messages.pop()], type: '', collapsed: this.groupIsCollapsed[message.id] ?? true })
+						groups.push({ id: message.id, lastId: message.id, messages: [groups.at(-1).messages.pop()], type: '', collapsed: group.messages.length > 1 })
```
Antreesy, line 236: Same here, only one message yet

src/components/MessagesList/MessagesGroup/MessagesSystemGroup.vue
```diff
@@ -119,8 +119,13 @@ export default {
 			immediate: true,
 			handler(value) {
 				this.messagesGroupedBySystemMessage = this.groupMessages(value)
+				this.updateCollapsedState()
```
Antreesy, line 122: I'd go with something like this, to update reactive array only once
```suggestion
				this.messagesGroupedBySystemMessage = this.updateCollapsedState(this.groupMessages(value))
```

src/components/MessagesList/MessagesGroup/MessagesSystemGroup.vue
```diff
@@ -119,8 +119,13 @@ export default {
...
 			},
 		},
+
+		lastReadMessageId() {
+			this.updateCollapsedState()
```
Antreesy, line 127:
```suggestion
			this.messagesGroupedBySystemMessage = this.updateCollapsedState(this.messagesGroupedBySystemMessage)
```

src/components/MessagesList/MessagesGroup/MessagesSystemGroup.vue
DorraJaouad: I kept it as a hint for the structure of the object.

src/components/MessagesList/MessagesGroup/MessagesSystemGroup.vue
DorraJaouad: the data is reassigned in that case, I am just updating.

Antreesy (approved):
Tested

PR #11996, opened 2024-04-03
Chore(Vue3): Migrate to mitt

DorraJaouad:
* Replace EventBus with mitt.
* As mitt emitter doesn't have `once` method, the latter was replaced like : 
```javascript

const handler = () => {
	function()
	emitter.off('event', handler)
}

emitter.on('event', handler) 

```

no visual changes

### 🏁 Checklist

---


src/services/emitter.js
```diff
@@ -20,6 +20,6 @@
  *
  */
 
-import Vue from 'vue'
+import mitt from 'mitt'
 
-export const EventBus = new Vue()
+export const emitter = mitt()
```
ShGKme, line 25: I think we should keep the name `EventBus`. Although it has a different implementation internally now, it still represent the same thing - the event bus / event hub, not any a general emitter.

ShGKme:
> * As mitt emitter doesn't have `once` method, the latter was replaced like :

Or we can add a new method to the event bus aka
```js
const EventBus = mitt()
EventBus.once = function(type, handler) {
  this.on(type, ...)
}
```

src/services/emitter.js
DorraJaouad: I changed the name to reflect the inner implementation change, and to if EventBus remained, `.once` can be mistakenly used.

DorraJaouad: but will revert the naming as I will define `once` method

ShGKme: > I changed the name to reflect the inner implementation change,

But it should not reflect inner implementation, it should be behind the scenes via "public API".

src/components/LeftSidebar/LeftSidebar.spec.js
```diff
@@ -129,7 +129,7 @@ describe('LeftSidebar.vue', () => {
 
 		test('fetches and renders conversation list initially', async () => {
 			const conversationsReceivedEvent = jest.fn()
-			EventBus.$once('conversations-received', conversationsReceivedEvent)
+			EventBus.on('conversations-received', conversationsReceivedEvent)
```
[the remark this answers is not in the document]
DorraJaouad: Sorry !

src/components/CallView/shared/Screen.vue
```diff
@@ -26,9 +26,10 @@
 			class="screen"
 			:class="screenClass" />
 		<VideoBottomBar v-if="isBig"
-			v-bind="$props"
-			:is-big="true"
-			:is-screen="true"
+			:token="token"
+			:shared-data="sharedData"
+			is-big
+			is-screen
```
ShGKme, line 32: Is it related and `callParticipantModel`, `localMediaModel` are not supposed to be passed?

ShGKme: Makes sense

ShGKme (approved):

PR #12021, opened 2024-04-06
fix: Set all rich object parameter values to string

provokateurin:
### ☑️ Resolves

Replaces https://github.com/nextcloud/spreed/pull/11991.
Only strings are allowed as values.

### 🏁 Checklist


SystemKeeper (approved):
Works fine on iOS

nickvergessen (changes requested):
PS: requires review from all requested persons
So blocking until everyone else approved

Antreesy (approved):
Adjusted frontend component to match new params. Nothing else is blocking

PS: backport to 27 might end up with conflicts, will handle it

DorraJaouad (approved):

mahibi:
all fine for android talk as well, @nickvergessen

PR #12029, opened 2024-04-09
fix(frontpage): Don't bruteforce protect "undefined"

nickvergessen:
It's a misbehaving bug in the frontend but we are yet unsure how it occures

### 🏁 Checklist


DorraJaouad:
> we are yet unsure how it occures

switching between rooms too fast ?

Antreesy:
> index.php/call/undefined

Could we track it somehow? Do we know the method (GET, POST, PUT, DELETE) where it happens?

nickvergessen:
I only see 3 technical options:
1. Dashboard: https://github.com/nextcloud/spreed/blob/f176008dd06034ec4671ddcb16613b849e7c0292/src/views/Dashboard.vue#L102-L106 but it was not accessed by the user today
2. Signaling: https://github.com/nextcloud/spreed/blob/d89d6f24500da801b7dc55650c94071f411b5ca3/src/utils/signaling.js#L315 but the user did not join a call today
3. Talk Hash: https://github.com/nextcloud/spreed/blob/d89d6f24500da801b7dc55650c94071f411b5ca3/src/App.vue#L412-L416 But the hash was not changed 8 times today. There was an update last night, so that could be the main root cause, but then it should be once (per tab?) and even then I would not know why `to.params.token` would be empty

DorraJaouad:
> the frontend endpoint index.php/call/undefined

Maybe switching between rooms too fast leads to `token === undefined` . Because I see not found page rendered so token must be faulty.

nickvergessen:
With fast changing the only thing I can reproduce is ending up on "room not found" `index.php/apps/spreed/not-found` but never on `call/undefined`

nickvergessen:
Today's user confirmed they only worked in mostly the same room (also what I see in the logs), but toggled between shared attachments and the chat frequently to leave comments about it

SystemKeeper:
Does the call always result in a „not-found“ page in talk? Do we know it comes from talk itself or could it be a notification with a wrong action or something?

nickvergessen:
> Does the call always result in a „not-found“ page in talk?

Well the conversation with the token `undefined` does not exist, so any follow up request should lead to that. But it's the frontpage, so it will just load the "home".

> Do we know it comes from talk itself or could it be a notification with a wrong action or something?

Persons didn't report any clicking on notifications. They opened attachments from the chat and closed them again.
I can reproduce the access.log pattern with that description apart from the `call/undefined`

SystemKeeper (approved):

Antreesy (approved):

PR #12032, opened 2024-04-09
enh(LobbySettings): add timezone and relative time

### LINKED ISSUE #12027: Show timezone when setting lobby time
### How to use GitHub

---

## Steps to reproduce
1. Create a conversation
2. Enable the lobby
3. Set a start time

### Expected behaviour
It's clear what the timezone the selected time is (even when it's not changable, so don't need a dropdown or something, just say the current timezone that will be used to "math")

### Actual behaviour
grafik
if it's < 24h even visitors have a better experience as they see a relative "in X hours" message.

## Talk app

**Talk app version:** main

DorraJaouad:
### ☑️ Resolves

* Fix #12027

very soon | soon | not in 24 hours |
-- | -- | -- |
image | image | image |

Lobby screen:
image

### 🏁 Checklist

### DIFF
--- a/src/components/ConversationSettings/LobbySettings.vue
+++ b/src/components/ConversationSettings/LobbySettings.vue
@@ -51,6 +51,12 @@
 					:input-class="['mx-input', { focusable: !lobbyTimerFieldDisabled }]"
 					v-bind="dateTimePickerAttrs"
 					@change="saveLobbyTimer" />
+				<div class="lobby_timer--timezone">
+					{{ getTimeZone }}
+				</div>
+				<div v-if="showRelativeTime" class="lobby_timer--relative">
+					{{ getRelativeTime }}
+				</div>
 			</form>
 		</div>
 	</div>
@@ -64,6 +70,9 @@ import NcDateTimePicker from '@nextcloud/vue/dist/Components/NcDateTimePicker.js
 import NcNoteCard from '@nextcloud/vue/dist/Components/NcNoteCard.js'
 
 import { WEBINAR } from '../../constants.js'
+import { futureRelativeTime } from '../../utils/formattedTime.ts'
+
+const ONE_DAY_IN_MS = 24 * 60 * 60 * 1000
 
 export default {
 	name: 'LobbySettings',
@@ -140,6 +149,24 @@ export default {
 				valueType: 'timestamp',
 			}
 		},
+
+		showRelativeTime() {
+			return this.lobbyTimer
+				&& this.lobbyTimer > Date.now()
+				&& (this.lobbyTimer - Date.now()) < ONE_DAY_IN_MS // less than 24 hours
+		},
+
+		getTimeZone() {
+			if (!this.lobbyTimer) {
+				return ''
+			}
+			const date = new Date(this.lobbyTimer)
+			return t('spreed', `Start time: ${date}`)
+		},
+
+		getRelativeTime() {
+			return futureRelativeTime(this.lobbyTimer)
+		},
 	},
 
 	methods: {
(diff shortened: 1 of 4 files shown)


src/components/ConversationSettings/LobbySettings.vue
```diff
@@ -140,6 +148,33 @@ export default {
...
+			return this.lobbyTimer
+				&& this.lobbyTimer > Date.now()
+				&& (this.lobbyTimer - Date.now()) < ONE_DAY_IN_MS // less than 24 hours
+		},
+
+		getTimeZone() {
+			if (!this.lobbyTimer) {
+				return ''
+			}
+			const date = new Date(this.lobbyTimer)
```
DorraJaouad, line 162: Aligning with the usage of NcDateTimePicker value as in example
image

src/components/ConversationSettings/LobbySettings.vue
```diff
@@ -140,6 +148,33 @@ export default {
...
+			return t('spreed', `Start time: ${date}`)
+		},
+
+		getRelativeTime() {
+			const diff = this.lobbyTimer - Date.now()
+			const hours = Math.ceil(diff / (60 * 60 * 1000))
+			if (hours > 1) {
+				// TRANSLATORS: hint for the time when the meeting starts
+				return n('spreed', 'in less than %n hour', 'in less than %n hours', hours)
```
Antreesy, line 171: Maybe use Math.floor for consistency? What's the usual pattern for these cases?

```suggestion
				return n('spreed', 'in %n hour', 'in %n hours', hours)
```

Antreesy (commented):
Works good.
But I'd expect to see similar relative numbers in moderator dialog, and lobby waiting page. Maybe we should align the appearance in both places?

image

src/components/ConversationSettings/LobbySettings.vue
DorraJaouad: I avoided floor because it is misleading IMO: 

left 1h59min -> shows in 1 hour.

Antreesy: Same would be in opposite to 1h02min. But anyway, as long as it's consistent in both places (here and at LobbyScreen.vue), I don't mind

jancborchardt: We should be exact to minutes here. Can't we say "In 1 hour 24 minutes"?

jancborchardt: Short form "In 1h 24m" also works if space is low.

src/components/LobbyScreen.vue
```diff
@@ -94,7 +96,7 @@ export default {
 			if (diff > -45000 && diff < 45000) {
 				return t('spreed', 'The meeting will start soon')
 			}
-			return this.timerInMoment.fromNow()
+			return futureRelativeTime(this.timerInMoment)
```
Antreesy, line 99: You should pass ms, not the Moment object
```suggestion
			return futureRelativeTime(this.timerInMoment.valueOf())
```

Yet it still doesn't show minutes after update (from 6 minutes to 5), can't get why: 
image

Antreesy (approved):
Works good. Not sure that missing minutes issue is related to changes, as computed value is correct

ShGKme:
> Not sure that missing minutes issue is related to changes, as computed value is correct

It is managed by this server code. If we live-relative-timestampwant to render time manually by Talk, then we need to remove this `live-relative-timestamp` class. Otherwise, I'd say, we should use `nextcloud-vue` component.

https://github.com/nextcloud/server/blob/3ad4bbb09692b180971e024ffeaab1d81fe77e4e/core/src/init.js#L42-L50

Antreesy (approved):

PR #12107, opened 2024-04-15
style: update CSS specificity for fallback mention icons

Antreesy:
### ☑️ Resolves

* Fix regression from #11811
* For BG color specifity was increased, so it follows:
  * server/component styles (0.2.0);
  * light theme in-Talk styles (0.3.0);
  * dark theme in-Talk styles (0.4.1);
  * system dark theme in-Talk styles (0.3.0 `@media`);
* Increase size for mention bubble (see screenshots)

☀️ Light theme | 🌑 Dark Theme
-- | --
image | image

Before | After
-- | --
image | image

### 🏁 Checklist

### DIFF
--- a/css/icons.css
+++ b/css/icons.css
@@ -92,29 +92,46 @@
  * .app-Talk rules above.
  * "forced-white" needs to be included in the class name as the Avatar does
  * not accept several classes. */
-.user-bubble__avatar .icon-group-forced-white,
-.user-bubble__avatar .icon-user-forced-white,
-.autocomplete-result .icon-group-forced-white,
-.autocomplete-result .icon-user-forced-white,
-.mention-bubble .icon-group-forced-white,
-.mention-bubble .icon-user-forced-white {
-	background-color: var(--color-text-maxcontrast-default) !important;
-}
-
-body[data-theme-dark] .icon-group-forced-white,
-body[data-theme-dark] .icon-user-forced-white {
-	background-color: #3B3B3B !important;
-}
-
-.user-bubble__avatar .icon-group-forced-white,
-.user-bubble__avatar .icon-user-forced-white {
-	background-size: 75%;
-}
-
+.user-bubble__avatar .icon-group-forced-white.avatar-class-icon,
+.user-bubble__avatar .icon-user-forced-white.avatar-class-icon,
 .autocomplete-result .icon-group-forced-white.autocomplete-result__icon--,
 .autocomplete-result .icon-user-forced-white.autocomplete-result__icon--,
 .mention-bubble .icon-group-forced-white.mention-bubble__icon--,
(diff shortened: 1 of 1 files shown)


nickvergessen:
For me the color is still different:
grafik

Even more obvious when mentioning:
grafik

css/icons.css
```diff
@@ -92,18 +92,18 @@
...
-.mention-bubble .icon-group-forced-white,
-.mention-bubble .icon-user-forced-white {
-	background-color: var(--color-text-maxcontrast-default) !important;
+.user-bubble__avatar .icon-group-forced-white.avatar-class-icon,
+.user-bubble__avatar .icon-user-forced-white.avatar-class-icon,
+.autocomplete-result .icon-group-forced-white.autocomplete-result__icon--,
+.autocomplete-result .icon-user-forced-white.autocomplete-result__icon--,
+.mention-bubble .icon-group-forced-white.mention-bubble__icon--,
+.mention-bubble .icon-user-forced-white.mention-bubble__icon-- {
+	background-color: var(--color-text-maxcontrast-default);
```
nickvergessen, line 101: this color here seems wrong. It seems to be from the intermediate time when we used the server color scheme.
But we reverted that so that avatars are dark on dark mode again

Antreesy: Same `#6B6B6B` and `#3B3B3B` we're using at PHP. But would be better to use directly, to not rely on server values

css/icons.css
```diff
@@ -92,29 +92,28 @@
...
 
-.user-bubble__avatar .icon-group-forced-white,
-.user-bubble__avatar .icon-user-forced-white {
+body[data-theme-dark] .user-bubble__avatar .icon-group-forced-white.avatar-class-icon,
+body[data-theme-dark] .user-bubble__avatar .icon-user-forced-white.avatar-class-icon,
+body[data-theme-dark] .autocomplete-result .icon-group-forced-white.autocomplete-result__icon--,
+body[data-theme-dark] .autocomplete-result .icon-user-forced-white.autocomplete-result__icon--,
+body[data-theme-dark] .mention-bubble .icon-group-forced-white.mention-bubble__icon--,
+body[data-theme-dark] .mention-bubble .icon-user-forced-white.mention-bubble__icon-- {
+	background-color: #3B3B3B;
```
nickvergessen, line 111: right sidebar is 3B with dark mode, but the group mention in the chat is still 6B on dark-mode.

After some time I then finally found it...
`body[data-theme-dark]` does not match.... My theme is empty as I use "system default" instead of hardcoded dark mode

nickvergessen:
It had broken the "force bright theme" with "dark os".
I added a patch to fix that.

nickvergessen (approved):

PR #12108, opened 2024-04-15
fix(media): rewrite media preferences functions

### LINKED ISSUE #12064: preferred media devices follow-up
- [x] `this._fallbackAudioInputId` can be omitted given that we have the list now, but let's do it next time.
- [x] add test coverage
- [x] check, what else could affect and select wrong audio input
- [ ] Add showWarning for changes / fallback / preferences (outside of modal) - coudl be also handled by https://github.com/nextcloud/spreed/issues/7025
- [ ] check why audio/video might be disabled on device change during call - https://github.com/nextcloud/spreed/issues/7025
- [x] replace console.log 

_Originally posted by @DorraJaouad in https://github.com/nextcloud/spreed/pull/11936#discussion_r1561141865_

Antreesy:
### ☑️ Resolves

* Closes #12064
* rewrite media preferences functions
  - fix registering only last device when populate with several
  - fix selecting last available device in list when join (instead of last used | first available)
  - drop 'promote' option as unused (we register promoted device in another place)
  - keep default devices in preferences list
  - remove fallback id logic
  - move console.log from function for testing purposes

### 🚧 Tasks

- [ ] Check behavior without granted permissions

### 🏁 Checklist

### DIFF
--- a/src/services/mediaDevicePreferences.ts
+++ b/src/services/mediaDevicePreferences.ts
@@ -74,16 +74,14 @@ ${videoInputList.map(getDeviceString).join('\n')}
 /**
  * Get the first available device from the preference list.
  *
- * Returns id of device from the list / provided fallback id / 'default' id
+ * Returns id of device from the list
  *
  * @param devices list of available devices
  * @param inputList list of registered audio/video devices in order of preference
- * @param [fallbackId] id of currently selected input
- * @return first available (plugged) device id or fallback
+ * @return {string|undefined} first available (plugged) device id
  */
-function getFirstAvailableMediaDevice(devices: MediaDeviceInfo[], inputList: MediaDeviceInfo[], fallbackId?: string): string | undefined {
-	const availableDevices = devices.map(device => device.deviceId).filter(id => id !== 'default')
-	return inputList.find(device => availableDevices.includes(device.deviceId))?.deviceId ?? fallbackId
+function getFirstAvailableMediaDevice(devices: MediaDeviceInfo[], inputList: MediaDeviceInfo[]): string | undefined {
+	return inputList.find(device => devices.some(d => d.kind === device.kind && d.deviceId === device.deviceId))?.deviceId
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/services/mediaDevicePreferences.ts
```diff
@@ -63,10 +71,10 @@ ${videoInputList.map(getDeviceString).join('\n')}
...
  * @param inputList list of registered audio/video devices in order of preference
  * @param [fallbackId] id of currently selected input
- * @return first available (plugged) device id or fallback
+ * @return {string|undefined} first available (plugged) device id or fallback
  */
 function getFirstAvailableMediaDevice(devices: MediaDeviceInfo[], inputList: MediaDeviceInfo[], fallbackId?: string): string | undefined {
-	const availableDevices = devices.map(device => device.deviceId).filter(id => id !== 'default')
+	const availableDevices = devices.map(device => device.deviceId).filter(id => id !== DEFAULT_ID)
 	return inputList.find(device => availableDevices.includes(device.deviceId))?.deviceId ?? fallbackId
 }
```
DorraJaouad, line 79: fallbackId is just an intermediary and can be skipped, here you have a fall back and also in the function all itself. 
- We should have default device in the list 
- Do the loop:
    - If a plugged device found, done
    - If not ( including default), return the value that indicates that no device is selected ( I am not sure if it's `null` or `undefined`) 
    - remove the fallback on the function call 
    `this.attributes.audioInputId = getFirstAvailableMediaDevice(devices, this._preferenceAudioInputList, this._fallbackAudioInputId) ?? 'default'`

and by having all devices in the list,  it means that we have the fallback there too.

src/services/mediaDevicePreferences.ts
```diff
@@ -105,26 +104,30 @@ function registerNewMediaDevice(device: MediaDeviceInfo, devicesList: MediaDevic
...
 		return null
 	}
 
+	const newInputList = inputList.slice()
 	let deviceToPromote = null
-	if (currentDevicePosition === -1 && inputId !== 'default' && inputId !== null) {
+
+	if (currentDevicePosition === -1) {
 		// If device was not registered in preferences list, get it from devices list
 		deviceToPromote = devices.find(device => device.deviceId === inputId)
```
DorraJaouad, line 132: devices and preferences list should be have the same length as populating the pref list loops over devices.

src/utils/webrtc/MediaDevicesManager.js
```diff
@@ -283,13 +283,13 @@ MediaDevicesManager.prototype = {
...
+			console.info(listMediaDevices(
 				devices,
 				this.attributes.audioInputId,
 				this.attributes.videoInputId,
 				this._preferenceAudioInputList,
 				this._preferenceVideoInputList,
-			)
+			))
 		})
 	},
```
DorraJaouad, line 294: nit: `_populatePreferences` should not update the browser storage because it doesn't reflect preferences yet, it follows the order of registration

DorraJaouad (commented):
Part 1

src/services/mediaDevicePreferences.ts
Antreesy: prefs list could be longer, if you used some device before that

src/utils/webrtc/MediaDevicesManager.js
Antreesy: This is something to think about: if we won't save device at this moment, and won't join a call to update prefs, device will not be saved anywhere (though it's not really needed, as 'default' will be higher in the list, and be used as a fallback value). But if we pick a device later from available, it will appear in the list eventually.
So might be fine, i think?

src/services/mediaDevicePreferences.ts
Antreesy: Only if we keep `_populatePreferences` =)

DorraJaouad: > Only if we keep _populatePreferences =)

Yes, it is kept, only browser storage part is skipped

Antreesy: As written below, we'd miss new devices in the list in some cases, so I'd keep browser storage part

nickvergessen:
There seems to be an issue now with Chromium at least.

After resetting permissions and reloading the page:
grafik


It's only asking for my camera permission after several seconds and even after granting it does not allow to enable the camera.
Browser console is yelling: `Error getting video stream: OverconstrainedError:`

nickvergessen:
If I reload after accepting permissions it seems to be working properly

DorraJaouad (approved):
Tested in different senarios, ✅ 

When user selects nothing (None) , preferences won't be applied anymore as we expect the user doesn't want to have any device set anymore.

When user manually selects in preview and without joining call, we assume was just experimenting devices and we do not update preferences.

PR #12133, opened 2024-04-17
fix(PublicShareSidebar): fix prop reactivity

### LINKED ISSUE #12132: polishing of sidebar in public view
just tested on nc29.0.0b1
- [x] The icons in system dark mode in the header look broken:
  image
- [x] Also the sidebar toggle button does not seem to work... https://github.com/nextcloud/spreed/pull/12133

DorraJaouad:
### ☑️ Resolves

* Fix #12132

Vue was not tracking the object changes in computed ( though in prop, it was changed), which means it's not reactive. 

### 🏁 Checklist

### DIFF
--- a/src/mainPublicShareSidebar.js
+++ b/src/mainPublicShareSidebar.js
@@ -19,7 +19,7 @@
  */
 
 import { createPinia, PiniaVuePlugin } from 'pinia'
-import Vue from 'vue'
+import Vue, { reactive } from 'vue'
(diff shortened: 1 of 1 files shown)


ShGKme (approved):

nickvergessen:
Fixes the button, but the wrong color of the icon is still a thing :see_no_evil: 
The file name however is also wrong (using the same variable, but directly instead of via a button).

PR #12158, opened 2024-04-19
fix(MessageBody): restrict checkbox editing in one-to-one conversation

DorraJaouad:
### ☑️ Resolves

Checkboxes should be disabled in other user messages for one to one conversation.

image

### 🏁 Checklist


nickvergessen:
Also the normal edit option should be hidden in 1-1
At least it's not working :P

Antreesy (approved):
Tested (see screenshot)

Follow-up: there's a lot of common checks between Message parts. We either need to localize them in HOC or extract into composable

nickvergessen (approved):

Issue #11994, opened 2024-04-03
False "All messages have expired or have been deleted." when switching conversations


ChristophWurst:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Steps to reproduce
1. Be in multiple conversations
2. Switch conversation

### Expected behaviour

See the latest messages

### Actual behaviour

See an empty content saying that all messages have been deleted or have expired.

![image](https://github.com/nextcloud/spreed/assets/1374172/70fe4083-1cc3-448f-ac96-3c1345d1862f)

After a page reload the messages are "back".

## Talk app

**Talk app version:** (see apps admin page: `/index.php/settings/apps`)

**Custom Signaling server configured:** yes/no and version (see additional admin settings: `/index.php/index.php/settings/admin/talk#signaling_server`)

**Custom TURN server configured:** yes/no (see additional admin settings: `/index.php/settings/admin/talk#turn_server`)

**Custom STUN server configured:** yes/no (see additional admin settings: `/index.php/settings/admin/talk#stun_server`)

## Browser

**Microphone available:** yes/no

**Camera available:** yes/no

**Operating system:** Windows/Ubuntu/...

**Browser name:** Firefox/Chrome/...

**Browser version:** 85/96/...

### Browser log

<details>
```
Insert your browser log here, this could for example include:
a) The javascript console log
b) The network log
c) ...
```

</details>

## Server configuration

**Operating system**: Ubuntu/RedHat/...

**Web server:** Apache/Nginx

**Database:** MySQL/Maria/SQLite/PostgreSQL

**PHP version:** 8.0/8.1/8.2

**Nextcloud Version:** (see admin page)

**List of activated apps:**

<details>

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your server installation folder
```
</details>

**Nextcloud configuration:**

<details>

```
If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your Nextcloud installation folder
```
</details>

### Server log (data/nextcloud.log)
<details>

```
Insert your server log here
```
</details>

nickvergessen:
Can you reproduce this with browser console open?
[...]

PR #12178, opened 2024-04-22
fix(messagesStore): compare message ids as numeric values

Antreesy:
### ☑️ Resolves

* Ref #11994
  * Not sure if that fixes this issue, but should fix another:

```js
const messageIds = Object.keys(state.messages[token])
// Array<string>
...
const messagesToRemove = messageIds.sort().reverse().slice(199)
// Sort strings by characters, so '100' < '99'
...
if (state.firstKnown[token] && messagesToRemove.includes(state.firstKnown[token]))
// state.firstKnown[token] is number, so Array<string>.includes(number) is always false
```

Because of that, firstKnown is not updated after list, and after opening the chat again, user couldn't load old messages 

### 🏁 Checklist

### DIFF
--- a/src/store/messagesStore.js
+++ b/src/store/messagesStore.js
@@ -504,15 +504,15 @@ const mutations = {
 			return
 		}
 
-		const messagesToRemove = messageIds.sort().reverse().slice(199)
+		const messagesToRemove = messageIds.sort((a, b) => b - a).slice(199)
 		const newFirstKnown = messagesToRemove.shift()
 
 		messagesToRemove.forEach((messageId) => {
 			Vue.delete(state.messages[token], messageId)
 		})
 
-		if (state.firstKnown[token] && messagesToRemove.includes(state.firstKnown[token])) {
-			Vue.set(state.firstKnown, token, newFirstKnown)
+		if (state.firstKnown[token] && messagesToRemove.includes(state.firstKnown[token].toString())) {
+			Vue.set(state.firstKnown, token, +newFirstKnown)
 		}
 	},
 }


src/store/messagesStore.js
```diff
@@ -504,7 +504,7 @@ const mutations = {
 			return
 		}
 
-		const messagesToRemove = messageIds.sort().reverse().slice(199)
+		const messagesToRemove = messageIds.map(parseInt).sort((a, b) => b - a).slice(199)
```
ShGKme, line 507:
```js
messageIds.map(parseInt)
```
is the same as
```js
messageIds.map((id, index) => parseInt(number=id, radix=index))
```

So `['9', '10', '11', '12'].map(parseInt)` doesn't return `[9, 10, 11, 12]`. It returns `[9, NaN, 3, 5]` using index as radix.

ShGKme: `parseInt` is not needed, `b - a` already casts to numbers

ShGKme (changes requested):

src/store/messagesStore.js
Antreesy: Yeah, but we need to compare it later on line 514 (though it would be easier to cast a firstKnown as a string there)

src/store/messagesStore.js
```diff
@@ -504,15 +504,15 @@ const mutations = {
...
 		const newFirstKnown = messagesToRemove.shift()
 
 		messagesToRemove.forEach((messageId) => {
 			Vue.delete(state.messages[token], messageId)
 		})
 
-		if (state.firstKnown[token] && messagesToRemove.includes(state.firstKnown[token])) {
-			Vue.set(state.firstKnown, token, newFirstKnown)
+		if (state.firstKnown[token] && messagesToRemove.includes(state.firstKnown[token].toString())) {
+			Vue.set(state.firstKnown, token, +newFirstKnown)
```
ShGKme, line 515: I haven't changed everything, but is it safe to change the type of `state.firstKnown`?

Antreesy: `state.firstKnown[token]` is number
`newFirstKnown` is string

ShGKme: Yes, now it is a `number`.

But before this PR it was a `string`. And it is a global state, so I guess, it is used in many places.

Can you confirm that it is safe to change the global state here, changing the type of `state.firstKnown[token]` from `string` to `number`?

ShGKme: > before this code was never reached =/

Yes, but having only changed code I can only see that it was supposed to be as as string before and now as a number. So I can either go and check all the usage of `state.firstKnown` to be sure it is actually supposed to be a number, or ask if you have id done before PR 😀

ShGKme (approved):

PR #12205, opened 2024-04-24
fix(MessagesList): handle special case in scrolling

DorraJaouad:
### ☑️ Resolves

We use last read message to scroll to when opening conversation. When it is collapsed, there is nothing to scroll to and the conversation remains shown at the top. 

* Scroll to its nearest visible parent div
* Adjust the other safeguards

🏚️ Before | 🏡 After
-- | --

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -621,18 +621,21 @@ export default {
 			}
 
 			if (!isFocused) {
-				// Safeguard: scroll to before last read message
+				// Safeguard 1: scroll to first visible message before the read marker
 				const fallbackLastReadMessageId = this.$store.getters.getFirstDisplayableMessageIdBeforeReadMarker(this.token, this.visualLastReadMessageId)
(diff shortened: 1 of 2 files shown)


src/components/MessagesList/MessagesList.vue
```diff
@@ -621,18 +621,16 @@ export default {
 			}
 
 			if (!isFocused) {
-				// Safeguard: scroll to before last read message
+				// Safeguard 1: scroll to first visible message before the read marker
 				const fallbackLastReadMessageId = this.$store.getters.getFirstDisplayableMessageIdBeforeReadMarker(this.token, this.visualLastReadMessageId)
 				if (fallbackLastReadMessageId) {
 					isFocused = this.focusMessage(fallbackLastReadMessageId, false, false)
-					this.$store.dispatch('setVisualLastReadMessageId', {
```
Antreesy, line 628: Not sure if that would break anything, maybe keep it as-is

Antreesy (approved):

Issue #10680, opened 2023-10-11
[RFC] Capability handling in federated conversations


nickvergessen:
When a remote conversation is used by a local user, some capabilities need to be checked/respected from the remote server, while others need to be considered from the local server.
Here is a list with the current capabilities and my understanding where they should be read from. Having said that, some capabilities can be considered to "not diverge" at the moment, e.g. features that are available. But when a new API version would ever hit that requires us to remove a capability, the Server would say again whether remote or local controls it.

### `features` Flags

Feature | Server | Comment
---|---|---
`audio` | ▶️ Remote
`video` | ▶️ Remote
`chat-v2` | ▶️ Remote
`conversation-v4` | ▶️ Remote
`guest-signaling` | ▶️ Remote
`empty-group-room` | ▶️ Remote
`guest-display-names` | ▶️ Remote
`multi-room-users` | ▶️ Remote
`favorites` | 🏠 Local
`last-room-activity` | ▶️ Remote
`no-ping` | ▶️ Remote
`system-messages` | ▶️ Remote
`delete-messages` | ▶️ Remote
`mention-flag` | ▶️ Remote
`in-call-flags` | ▶️ Remote
`conversation-call-flags` | ▶️ Remote
`notification-levels` | ▶️ Remote
`invite-groups-and-mails` | ▶️ Remote
`locked-one-to-one-rooms` | ▶️ Remote
`read-only-rooms` | ▶️ Remote
`listable-rooms` | 🏠 Local
`chat-read-marker` | ▶️ Remote
`chat-unread` | ▶️ Remote
`webinary-lobby` | ▶️ Remote
`start-call-flag` | ▶️ Remote
`chat-replies` | ▶️ Remote
`circles-support` | ▶️ Remote
`force-mute` | ▶️ Remote
`sip-support` | ▶️ Remote
`sip-support-nopin` | ▶️ Remote
`chat-read-status` | ▶️ Remote
`phonebook-search` | 🏠 Local
`raise-hand` | ▶️ Remote
`room-description` | ▶️ Remote
`rich-object-sharing` | ▶️ Remote
`temp-user-avatar-api` | 🏠 Local
`geo-location-sharing` | ▶️ Remote
`voice-message-sharing` | ▶️ Remote
`signaling-v3` | ▶️ Remote
`publishing-permissions` | ▶️ Remote
`clear-history` | ▶️ Remote
`direct-mention-flag` | ▶️ Remote
`notification-calls` | ▶️ Remote
[...]

PR #12208, opened 2024-04-25
fix(capabilities): implement a wrapper for capabilities

### LINKED ISSUE #11850: Check capabilities of remote server in federated conversations
ATM only 29-29 can be federated, but we need to check in future if new features are supported by both servers

- [x] Storing / reading:
  - [x] federationsStore -> capabilities -> per token
- [x] Using:
  - [x] useCapabilities wrapper to check room and decide: -> local capabilities or remote

Antreesy:
### ☑️ Resolves

* Fix #11850
* Ref #10680

No visual changes

### 🚧 Tasks

- [x] Implement logic to check if room is federated, to fetch remote capabilities
  - [x] Think of a way to store/ retrive remote capabilities
  As it's a request to server, it should be asyncronous. That means we must in some case to wait for remote capabilities or re-render application based on response (which is now isn't happen and requires a reload, if capability has changed
  Reload as we do for dirty hash
- [x] Implement helpers like `hasFeature` to replace long `getCapabilities()?.spreed?.features?.includes('capability')`
- [x] Distinguish between whether to consider local/remote capability (see ref issue)

### 🏁 Checklist


DorraJaouad (commented):
You can use store instead of browser storage. The purpose of keeping the last previous fetched capabilities is to check in case the new fetch renders a different list (on the same page) -> store is enough for that.

src/store/participantsStore.js
```diff
@@ -852,6 +853,11 @@ const actions = {
 				sessionId: response.data.ocs.data.sessionId,
 			})
 
+			if (response.data.ocs.data.remoteServer) {
+				// fetch and store remote capabilities for federated conversation
+				await setRemoteCapabilities(token)
+			}
+
```
DorraJaouad, line 860: what if the conversation is already open and capabilities have changed ?

Antreesy: We have now a check for local requests (dirty hash), might be required to implement the same for remote conversations

Antreesy:
> You can use store instead of browser storage.

Might not work well for some files (as we have components, which are using Vue, and files like service/webrtc
Looks like we shoud handle each case differently..

`--` not needed
`🟢` any user
`🟡` moderator
`🔴` administrator


File | check once for instance | check on created/setup | check on token update
-- | -- | -- | --
AdminSettings/RecordingServers.vue | 🔴 | -- | --
views/AdminSettings.vue | 🔴 | -- | --
src/views/FlowPostToConversation.vue | 🔴 | -- | --
CallView/CallView.vue | -- | 🟢 (rendered on each call) | --
ChatView.vue | -- | 🟢 | 🟢
ConversationIcon.vue | -- | 🟢 | 🟢
BasicInfo.vue | -- | 🟢 | 🟢
ConversationSettingsDialog.vue | -- | 🟡 | 🟡
NotificationsSettings.vue | -- | 🟢 | 🟢
RecordingConsentSettings.vue | -- | 🟡 | 🟡
LeftSidebar.vue | -- | 🟢 (SIP, local only) | --
MediaSettings.vue | -- | 🟢 | 🟢
VideoBackgroundEditor.vue | -- | 🟢 | 🟢
Message.vue | -- | 🟢 (translation, local only -> ChatView and provide) | --
MessageButtonsBar.vue | -- | 🟢 (rendered on each switch) | --
FilePreview.vue | -- | 🟢 | 🟢
MessageBody.vue | -- | 🟢 (rendered on each switch) | --
NewConversationSetupPage.vue | -- | 🟢 (local, avatar) | --
NewMessage.vue | -- | 🟢 | 🟢
NewMessageUploadEditor.vue | -- | 🟢 | 🟢
Participant.vue | -- | 🟢 (rendered on each switch) | --
ParticipantsTab.vue | -- | 🟢 (SIP) | 🟢
RightSidebar.vue | -- | 🟢 | 🟢
RoomSelector.vue | -- | 🟢 | 🟢
SettingsDialog.vue | -- | 🟢 (local) | --
CallButton.vue | -- | 🟢 | 🟢
TopBar.vue | -- | 🟢 | 🟢
TopBarMenu.vue | -- | 🟢 | 🟢
src/composables/useActiveSession.js | -- | depends on where state is | --
src/services/conversationsService.js | 🟢 should be local? | -- | --
src/store/messagesStore.js | -- | -- | 🟢 (rendered on each switch)
src/utils/webrtc/index.js | 🟢 local for now | -- | --

src/services/CapabilitiesManager.ts
```diff
@@ -0,0 +1,97 @@
...
+			? this.localCapabilities
+			: this.remoteCapabilities[token]
+	}
+
+	public hasTalkFeature(token: 'local'|string, feature: string): boolean {
+		if (this.localCapabilities.spreed['features-local'].includes(feature)) {
+			return this.localCapabilities?.spreed?.features?.includes(feature) ?? false
+		}
+		return this.getCapabilities(token)?.spreed?.features?.includes(feature) ?? false
+	}
```
DorraJaouad, line 45: If the feature is local, we don't need to check remote server at all because those features are usually not affecting the chat, call etc (e.g: note-to-self)
If the feature is not flagged as local, both remote and local capabilities should be checked using `AND` (if any of the server doesn't have that feature, it should not be there).

Note: we need to be careful when flagging feature/config with local, I noticed that `config => attachments => allowed` is local whereas it is affecting the chat.

src/services/CapabilitiesManager.ts
```diff
@@ -0,0 +1,97 @@
...
+
+type Config = Capabilities['spreed']['config']
+
+class CapabilitiesManager implements ICapabilitiesManager {
+
+	public localCapabilities
+
+	public remoteCapabilities
+
+	static instance: ICapabilitiesManager
```
ShGKme, line 27: This property is never used

src/services/CapabilitiesManager.ts
```diff
@@ -0,0 +1,97 @@
...
+			this.remoteCapabilities[token].hash = joinRoomResponse.headers['x-nextcloud-talk-proxy-hash']
+			BrowserStorage.setItem('remoteCapabilities', JSON.stringify(this.remoteCapabilities))
+
+			// As normal capabilities update, requires a reload to take effect
+			showError(t('spreed', 'Nextcloud Talk Federation was updated, please reload the page'), {
+				timeout: TOAST_PERMANENT_TIMEOUT,
+			})
+		} catch (error) {
+			console.error(error)
+		}
```
ShGKme, line 78: In case of error, this method finishes successfully continuing joining the chat. Maybe instead of logging exception, do not catch it at all here?

src/services/CapabilitiesManager.ts
```diff
@@ -0,0 +1,97 @@
...
+			return this.localCapabilities?.spreed?.features?.includes(feature) ?? false
+		}
+		return this.getCapabilities(token)?.spreed?.features?.includes(feature) ?? false
+	}
+
+	public getTalkConfig(token: 'local'|string, key1: keyof Config, key2: keyof Config[keyof Config]) {
+		return this.localCapabilities.spreed['config-local'][key1][key2]
+			? this.localCapabilities?.spreed?.config?.[key1]?.[key2]
+			: getCapabilities(token)?.spreed?.config?.[key1]?.[key2]
+	}
```
ShGKme, line 51: Isn't it supposed to be `this.getCapabilities` instead of `getCapabilities`?

src/services/CapabilitiesManager.ts
```diff
@@ -0,0 +1,97 @@
...
+		}
+
+		return JSON.parse(remoteCapabilities) as ICapabilitiesManager['remoteCapabilities']
+	}
+
+}
+
+const capabilitiesManager = new CapabilitiesManager()
+
+export const getCapabilities = capabilitiesManager.getCapabilities.bind(capabilitiesManager)
```
ShGKme, line 94: I'd think about another name to not mess with `@nextcloud/capabilities`, because this `getCapabilities` has a bit different interface

ShGKme (commented):
Looks very cool!

Just some implementation nitpicks and questions.

src/components/RightSidebar/Participants/ParticipantsTab.vue
```diff
@@ -184,6 +179,10 @@ export default {
 			return this.$store.getters.getUserId()
 		},
 		canAddPhones() {
+			const canModerateSipDialOut = hasTalkFeature(this.token, 'sip-support-dialout')
+					&& getTalkConfig(this.token, 'call', 'sip-enabled')
+					&& getTalkConfig(this.token, 'call', 'sip-dialout-enabled')
+					&& getTalkConfig(this.token, 'call', 'can-enable-sip')
```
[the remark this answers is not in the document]
ShGKme: I guess here computed is used because `this.token` can change?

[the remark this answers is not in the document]
Antreesy: Yes, and RightSidebar is always mounted, so we have to check in on conversation switch
Don't think it's worth to make a composable out of it

src/services/CapabilitiesManager.ts
```diff
@@ -0,0 +1,97 @@
...
+export const getTalkConfig = capabilitiesManager.getTalkConfig.bind(capabilitiesManager)
+export const hasTalkFeature = capabilitiesManager.hasTalkFeature.bind(capabilitiesManager)
+export const setRemoteCapabilities = capabilitiesManager.setRemoteCapabilities.bind(capabilitiesManager)
```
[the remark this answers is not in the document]
Antreesy: Nope, just tried it out. Functional approach should work the same way here

src/services/CapabilitiesManager.ts
Antreesy: I'd agree that feature (and also config) should be respected from both local / remote, if that's related to the conversation; remote user shouldn't have more possibilities, than he already has locally, and more than a local user:

Remote cap. | Local cap. | Resulting | Example
-- | -- | -- | --
🔴 | 🔴 | 🔴 | Not conflicting
🟢 | 🟢 | 🟢 | Not conflicting
🟢 | 🔴 | 🔴 | "Moderator" from 29 can't ban participants, because he isn't aware of the feature 
🔴 | 🟢 | 🔴 | Can't upload attachments, if not allowed on remote server

src/services/CapabilitiesManager.ts
```diff
@@ -0,0 +1,85 @@
...
+ * @param key1 top-level key (e.g. 'attachments')
+ * @param key2 second-level key (e.g. 'allowed')
+ */
+export function getTalkConfig(token: string = 'local', key1: keyof Config, key2: keyof Config[keyof Config]) {
+	if (token === 'local' || !remoteCapabilities[token] || localCapabilities.spreed['config-local'][key1][key2]) {
+		return localCapabilities?.spreed?.config?.[key1]?.[key2]
+	}
+
+	return remoteCapabilities[token]?.spreed?.config?.[key1]?.[key2]
+}
```
ShGKme, line 45: Missing return type

src/services/CapabilitiesManager.ts
DorraJaouad: > But for the config, i think remote should be respected, unless we define it in config-local: as we have numeric and string values there, not only boolean

For config, it's a bit tricky, maybe min(remote, local) for numeric config and for not measurable numeric (like integers to reflect modes), use remote, right  ?
For now, `chat-read-status` (local) which should be following remote server, no ?

src/services/CapabilitiesManager.ts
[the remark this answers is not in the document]
DorraJaouad: > If token === 'local' but localCapabilities.spreed['config-local'][key1][key2] is false, then it returns remoteCapabilities.local?

If `token === 'local'`, it means we strictly check local config. Same for `localCapabilities.spreed['config-local'][key1][key2]`.  One of them being truthy is enough I think.

src/services/CapabilitiesManager.ts
Antreesy: Or we treat each config individually for now.
chat-read-status differs per user and set up by him, if I'm not mistaken, so should be from local

DorraJaouad: > * `chat-read-status` (local) - On conversation API v3 and the chat API the last common read message is exposed which can be used to update the "read status" flag of own chat messages. The info should be shown only when the user also shares their read status. The user's value can be found in `config => chat => read-privacy`.

That means it is first configured by server (expose  the last common read message) and then enabled/disabled by user.

src/services/CapabilitiesManager.ts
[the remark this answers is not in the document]
Antreesy: Rewrote it a bit:
- First source of truth would be `features-local` and `config-local`, so we can define it on backend side for each individually;
- Then check for explicit 'local' or having a remote ones loaded

ShGKme (approved):
Looks good to me ✨

src/services/CapabilitiesManager.ts
Antreesy: `chat-read-status` isn't used in frontend code, so consider this out of scope of this PR. Let's stick with:
- features should be respected on both
- configs should be respected from remote, unless we don't define them as local || unless we don't set code to respect both individually for each case

Speaking of which, @nickvergessen :
`config => attachments => allowed`, shouldn't we respect both local and remote?
`features => avatar`, aren't we proxying the same local endpoint for avatar, so only local is enough?

nickvergessen: yeah the docs might be from before we decided to proxy avatars.

As for attachments, I guess we should check both, but we can clarify once we implement them

Antreesy:
Rebased onto main and resolved conflicts (message components and composables)

src/components/ConversationSettings/ConversationSettingsDialog.vue
```diff
@@ -181,7 +176,7 @@ export default {
 		},
 
 		showMediaSettingsToggle() {
-			return (!supportFederationV1 || !this.conversation.remoteServer)
+			return (!hasTalkFeature(this.token, 'federation-v1') || !this.conversation.remoteServer)
```
DorraJaouad, line 179: This is not dynamic

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -215,13 +214,11 @@ import { useDevices } from '../../composables/useDevices.js'
 import { useIsInCall } from '../../composables/useIsInCall.js'
 import { AVATAR, CALL, PARTICIPANT, VIRTUAL_BACKGROUND } from '../../constants.js'
 import BrowserStorage from '../../services/BrowserStorage.js'
+import { getTalkConfig } from '../../services/CapabilitiesManager.ts'
 import { useGuestNameStore } from '../../stores/guestName.js'
 import { useSettingsStore } from '../../stores/settings.js'
 import { localMediaModel } from '../../utils/webrtc/index.js'
 
-const recordingEnabled = getCapabilities()?.spreed?.config?.call?.recording || false
-const recordingConsent = getCapabilities()?.spreed?.config?.call?.['recording-consent']
```
DorraJaouad, line 223: one is per server level and one per conversation afaik ? (static and dynamic)

src/components/ConversationSettings/ConversationSettingsDialog.vue
```diff
@@ -218,19 +213,20 @@ export default {
...
-			return getCapabilities()?.spreed?.features?.includes('bots-v1')
+		supportBotsV1() {
+			return hasTalkFeature(this.token, 'bots-v1')
 		},
 
 		canConfigureBreakoutRooms() {
-			const breakoutRoomsEnabled = getCapabilities()?.spreed?.config?.call?.['breakout-rooms'] || false
 			return this.canFullModerate
-				&& breakoutRoomsEnabled
+				&& (getTalkConfig(this.token, 'call', 'breakout-rooms') || false)
```
DorraJaouad, line 222: `|| false` are not needed in Boolean evaluation, it will return a falsy value `undefined`

DorraJaouad (approved):
Tested 🦅

src/components/ConversationSettings/ConversationSettingsDialog.vue
Antreesy: Component is always mounted, and we can call it from any route (LeftSidebar -> Conversation -> Conversation Settings), so we should rely on token of this component

src/components/MediaSettings/MediaSettings.vue
Antreesy: `recordingEnabled` - if recording server is available (should be from remote)
`recordingConsent` - if consent is enabled by administrator (should be from remote)
`conversation.recordingConsent` - if enabled per conversation

src/components/ConversationSettings/ConversationSettingsDialog.vue
Antreesy: Decided not to change that. Though we could improve it later, I don't see a need for it (if works, don't touch)

PR #12214, opened 2024-04-26
Refactor: Move common message related computed props into composable

DorraJaouad:
### ☑️ Resolves

- Fix condition in `NewMessage` so it filters messages withing 24 hours (according to the current editing capability)
- Move ` isEditable` logic to composable.
- Move more common computed props
- There were missing oneToOne props in some components O.o 

### 🏁 Checklist


Antreesy:
I thought about it as a bigger thing, like `useConversationInfo` for several things at once.
Now you compute several things twice for MessageButtonsBar (`isMyMsg`, `isConversationReadOnly` and so on)
I think we could extract to composable and return from it more than `isEditable` only

src/components/RightSidebar/BreakoutRooms/BreakoutRoomsTab.vue
```diff
@@ -111,6 +111,11 @@ export default {
 			return !this.isOneToOne && (this.canFullModerate || this.mainConversation.participantType === PARTICIPANT.TYPE.GUEST_MODERATOR)
 		},
 
+		isOneToOne() {
```
Antreesy, line 114: No wonder we haven't noticed it before :see_no_evil:

DorraJaouad: It slipped from the radar

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
```diff
@@ -566,27 +561,10 @@ export default {
...
 		isCurrentGuest() {
 			return this.$store.getters.isActorGuest()
 		},
 
-		isMyMsg() {
-			return this.actorId === this.$store.getters.getActorId()
-				&& this.actorType === this.$store.getters.getActorType()
-		},
-
 		isConversationReadOnly() {
```
Antreesy, line 568: Already have it in composable

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
```diff
@@ -515,24 +528,6 @@ export default {
...
-		isEditable() {
-			if (!canEditMessage || !this.isModifiable || this.isObjectShare
-					|| ((!this.$store.getters.isModerator || this.isOneToOne) && !this.isMyMsg)) {
-				return false
-			}
-
-			return (moment(this.timestamp * 1000).add(1, 'd')) > moment()
-		},
-
 		isDeleteable() {
```
Antreesy, line 531: I'd consider move that to composable or at least simplify them with `isOneToOne`, `getters.isModerator` and so on

src/composables/useMessageInfo.js
```diff
@@ -0,0 +1,85 @@
+/*
+ * @copyright Copyright (c) 2024 Dorra Jaouad <dorra.jaoued1@gmail.com>
```
Antreesy, line 2: Rebase & replace with SPDX headers

src/composables/useMessageInfo.js
```diff
@@ -0,0 +1,85 @@
...
+		&& conversation.value.participantType !== PARTICIPANT.TYPE.GUEST)
+
+	const isObjectShare = computed(() => Object.keys(Object(message.value.messageParameters)).some(key => key.startsWith('object')))
+
+	const isMyMsg = computed(() =>
+		message.value.actorId === store.getters.getActorId()
+		&& message.value.actorType === store.getters.getActorType()
+	)
+
+	const canEditMessage = getCapabilities()?.spreed?.features?.includes('edit-messages')
```
Antreesy, line 60: That's better to move out of composable (after import), so it gets called only once

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.spec.js
```diff
@@ -96,13 +97,20 @@ describe('MessageButtonsBar.vue', () => {
...
+		let useMessageInfoSpy
 
 		beforeEach(() => {
 			store = new Store(testStoreConfig)
 
 			injected = {
 				getMessagesListScroller: jest.fn(),
 			}
+
+			useMessageInfoSpy = jest.spyOn(useMessageInfoModule, 'useMessageInfo')
```
Antreesy, line 109: Same here

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.spec.js
```diff
@@ -313,13 +330,19 @@ describe('MessageButtonsBar.vue', () => {
 			})
 
 			test('shows delete action on other people messages for moderators', () => {
+				useMessageInfoSpy.mockReturnValue({
+					isDeleteable: () => true,
+				})
 				messageProps.actorId = 'another-user'
 				conversationProps.type = CONVERSATION.TYPE.GROUP
 				conversationProps.participantType = PARTICIPANT.TYPE.MODERATOR
```
Antreesy, line 338: Apart from `PARTICIPANT.TYPE.GUEST_MODERATOR`, that wasn't allowed before

src/composables/useConversationInfo.js
```diff
@@ -114,8 +114,23 @@ export function useConversationInfo({
...
+		return [CONVERSATION.TYPE.ONE_TO_ONE, CONVERSATION.TYPE.ONE_TO_ONE_FORMER].includes(item.value.type)
+	})
+
+	const isConversationReadOnly = computed(() => {
+		return item.value.readOnly === CONVERSATION.STATE.READ_ONLY
+	})
+
+	const isModifiable = computed(() =>
+		!isConversationReadOnly.value
+		&& item.value.participantType !== PARTICIPANT.TYPE.GUEST)
```
Antreesy, line 127:
```suggestion
		&& item.value.participantType !== PARTICIPANT.TYPE.GUEST
		&& item.value.participantType !== PARTICIPANT.TYPE.GUEST_MODERATOR)
```

src/components/MessagesList/MessagesGroup/Message/Message.spec.js
```diff
@@ -414,6 +424,11 @@ describe('Message.vue', () => {
 						type: 'file',
 					},
 				}
+
+				useMessageInfoSpy.mockReturnValue({
```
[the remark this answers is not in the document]
DorraJaouad: I prefer not to rely on passed props of the component as it may miss one of the needed keys like `timestamp`. 

> get rid of v-bind and set all props to Message/MessageBody/e.t.c explicitly (as a whole object) - could be done separately

If composable has a way to get it without extending component props, it is good as-is ? 

and regarding testing, we can move mocking level to store getter instead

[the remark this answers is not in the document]
Antreesy: With the recent changed (order of v-bind) it will overwrite passed token in Vue 3.
 Ref: https://v3-migration.vuejs.org/breaking-changes/v-bind
 So it would be better to replace `<Message :token="token" v-bind="message" />` to `<Message :token="token" :message="message" />` and so on, but just for Next RC before migration to Vue 3 (as was discussed with @ShGKme)

Composable itself is good, but in Vue3 we'll face the issue in federated conversations, as `props.token` would be invalid for the local instance

We shouldn't mock our own modules, if they provide us logic, which we're testing. It's better to add complete message object to the store instead, if needed (`addMessage` or `processMessage`)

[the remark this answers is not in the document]
DorraJaouad: > With the recent changed (order of v-bind) it will overwrite passed token in Vue 3.
Ref: [v3-migration.vuejs.org/breaking-changes/v-bind](https://v3-migration.vuejs.org/breaking-changes/v-bind)
So it would be better to replace <Message :token="token" v-bind="message" /> to <Message :token="token" :message="message" />

it turns out that token is now the same after [this](https://github.com/nextcloud/spreed/commit/ce6b67d9214f21c55119f434887ff06c3685b827) so both orders are the same. 

>  We should be able to refactor composables anytime, create or delete composables without the need to rewrite all the tests. It makes further refactoring very painful and doesn't help to test the component, it's only testing internal implementation.

That's why I mocked the composable output in the test instead of adding it to the store ( so composable can fetch from the store). 
If I understood correctly, composables are to be mocked in the UI test as we don't need their implementation testing, rather their output.

src/components/CallView/shared/EmptyCallView.vue
```diff
@@ -78,7 +78,8 @@ export default {
 		},
 
 		isOneToOneConversation() {
-			return this.conversation && this.conversation.type === CONVERSATION.TYPE.ONE_TO_ONE
+			return this.conversation?.type === CONVERSATION.TYPE.ONE_TO_ONE
+				|| this.conversation?.type === CONVERSATION.TYPE.ONE_TO_ONE_FORMER
```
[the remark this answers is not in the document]
DorraJaouad: `EmptyCallView` is open when there is a call thus when a conversation is open.
and here I was just making the` isOneToOne` prop consistent between all components.

src/components/MessagesList/MessagesGroup/Message/MessagePart/MessageBody.vue
```diff
@@ -245,9 +244,15 @@ export default {
 		},
 	},
 
-	setup() {
+	setup(props) {
+		const {
+			isEditable,
+			isFileShare,
+		} = useMessageInfo(props.token, props.id)
```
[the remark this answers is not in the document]
DorraJaouad: `props.token` and  `props.id` are not needed to be reactive, they only have one value each during the lifetime of a component `MessageBody` in particular.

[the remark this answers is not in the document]
ShGKme: > `props.token` and `props.id` are not needed to be reactive, they only have one value each during the lifetime of a component `MessageBody` in particular.

Is it a hard requirement of a component, e.g., it is unusable with update and must be recreated?

Anyway, it is a good practice to not lose reactivity even if it is not needed in a specific case.

src/components/MessagesList/MessagesGroup/Message/Message.spec.js
```diff
@@ -337,6 +339,14 @@ describe('Message.vue', () => {
 				return messageEl
 			}
 
+			beforeEach(() => {
+				useMessageInfoSpy = jest.spyOn(useMessageInfoModule, 'useMessageInfo')
+			})
+
+			afterEach(() => {
+				useMessageInfoSpy.mockRestore()
+			})
```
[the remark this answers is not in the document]
DorraJaouad: It will be more boilerplace, as it needs to mock return for all props from composable to every suite, at least 
`TypeError: Cannot destructure property 'isEditable' of '(0 , _useMessageInfo.useMessageInfo)(...)' as it is undefined.`

src/components/MessagesList/MessagesGroup/Message/Message.spec.js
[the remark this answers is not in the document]
ShGKme: Composables are a part of a component, not its interface.

It isn't important for a component's user if they are implemented via reusable composables or not. Tests should test how a component is used by other components and apps, not it is implemented internally.

This component renders a message using params from props and data from global state. Then we should test it with a specific prop in a specific global state as an integration test.

Currently, if we refactor it again, for example, remove, change or create a new composable, or refactor composable, we would need to rewrite component tests again.

If we break the composable, the test will pass even if it actually doesn't work (at least without unit tests on the composable).

In these cases, tests create more problems than they solve. They don't prevent all the issues, but they require doing an additional job on internal changes.

src/components/MessagesList/MessagesGroup/Message/MessagePart/MessageBody.vue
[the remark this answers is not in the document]
DorraJaouad: >  it is unusable with update and must be recreated?

Yes, as:
* If token has changed, MessagesList will be re-mounted
* If id has changed, Message component ( where MessageBody is used) will be re-mounted.

I thought it's a good practice not to abuse reactivity :p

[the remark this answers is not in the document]
ShGKme: > * If token has changed, MessagesList will be re-mounted

This is not a requirement of `Message.vue` usage or its restriction. This is only its today's usage in another component.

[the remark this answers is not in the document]
ShGKme: And it increases the chances that these lines will be copy/pasted or used as an example for another component with probably another composable, where it is important to not lose reactivity.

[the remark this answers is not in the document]
ShGKme: If you are worried about performance - there is almost zero performance impact.

`toRef` doesn't create "new reactivity" here. We just providing a reference to `props.token` instead of providing the value of `props.token. It creates a small object with a `value` property to provide getter.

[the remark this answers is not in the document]
DorraJaouad: I made them reactive

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.spec.js
```diff
@@ -264,19 +278,10 @@ describe('MessageButtonsBar.vue', () => {
 
 			/**
 			 * @param {boolean} visible Whether or not the delete action is visible
-			 * @param {Date} mockDate The message date (deletion only works within 6h)
 			 */
-			function testDeleteMessageVisible(visible, mockDate) {
+			function testDeleteMessageVisible(visible) {
 				store = new Store(testStoreConfig)
```
Antreesy, line 283: you don't rely on store in these tests, so not needed here, probably

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.spec.js
```diff
@@ -291,53 +296,18 @@ describe('MessageButtonsBar.vue', () => {
...
+			test('hides delete action when it cannot be deleted', async () => {
 				testDeleteMessageVisible(false)
 			})
 
-			test('show delete action for file messages', () => {
+			test('show delete action when it can be deleted', () => {
+				useMessageInfoSpy.mockReturnValue({
+					isDeleteable: () => true,
+				})
 				messageProps.message = '{file}'
```
Antreesy, line 307: same for this, default message should be enough

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
```diff
@@ -296,7 +298,6 @@ import { parseMentions } from '../../../../../utils/textParse.ts'
 const EmojiIndex = new EmojiIndexFactory(data)
 const supportReminders = getCapabilities()?.spreed?.features?.includes('remind-me-later')
 const canEditMessage = getCapabilities()?.spreed?.features?.includes('edit-messages')
```
Antreesy, line 300: Only usage of this can be replaced with `isEditable`

src/components/MessagesList/MessagesGroup/Message/Message.spec.js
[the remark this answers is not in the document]
Antreesy: `store.dispatch('processMessage', { token: TOKEN, message: messageProps })` works for that file without mocking, if you put it inside `function renderRichObject`

src/composables/__tests__/useMessageInfo.spec.js
```diff
@@ -0,0 +1,228 @@
...
+		// Assert
+		expect(result.isDeleteable.value).toBe(false)
+	})
+
+	test('message is not deleteable when the conversation is read-only', () => {
+		// Arrange
+		mockConversationInfo = {
+			isOneToOneConversation: computed(() => false),
+			isConversationReadOnly: computed(() => true),
+			isConversationModifiable: computed(() => true),
```
Antreesy, line 102: You might want to add tests for `isConversationModifiable: computed(() => false)`

src/composables/__tests__/useMessageInfo.spec.js
```diff
@@ -0,0 +1,228 @@
...
+	})
+
+	test('file message is deleteable', () => {
+		// Arrange
+		messageProps.message = '{file}'
+		messageProps.messageParameters.file = {}
+		// Act
+		const result = useMessageInfo(messageProps.token, messageProps.id)
+		// Assert
+		expect(result.isFileShare.value).toBe(true)
```
Antreesy, line 118: `		expect(result.isFileShareWithoutCaption.value).toBe(true)`

src/composables/useMessageInfo.js
```diff
@@ -0,0 +1,81 @@
...
+const canEditMessage = getCapabilities()?.spreed?.features?.includes('edit-messages')
+
+/**
+ * Check whether the user can edit the message or not
+ *
+ * @param {import('vue').Ref} tokenRef conversation token
+ * @param {import('vue').Ref} messageIdRef message id to edit
+ * @return {import('vue').ComputedRef<boolean>}
+ */
+export function useMessageInfo(tokenRef = ref(null), messageIdRef = ref(null)) {
```
Antreesy, line 24: Cases where [token, id, conversation, message] are `undefined` and `null` are not handled, right? :eyes:

Antreesy (commented):
Looks good, but I'd cover some more cases with new tests, to be sure composable is working as expected

Antreesy (approved):
Tested, a good addition to the codebase :seal:

PR #12240, opened 2024-04-29
test(teams): Add a test to check if "Shared with team" allows chatting

nickvergessen:
### 🚧 Tasks

- [x] Add test to show the problem
- [x] Requires https://github.com/nextcloud/circles/issues/1575

### 🏁 Checklist

### DIFF
--- a/.github/workflows/integration-mysql.yml
+++ b/.github/workflows/integration-mysql.yml
@@ -56,6 +56,7 @@ jobs:
         php-versions: ['8.2']
         server-versions: ['master']
         guests-versions: ['master']
+        circles-versions: ['master']
         call-summary-bot-versions: ['main']
         notifications-versions: ['master']
 
@@ -92,6 +93,13 @@ jobs:
           repository: nextcloud/call_summary_bot
           path: apps/call_summary_bot
           ref: ${{ matrix.call-summary-bot-versions }}
+
+      - name: Checkout circles app
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+        with:
+          repository: nextcloud/circles
+          path: apps/circles
+          ref: ${{ matrix.circles-versions }}
 
       - name: Checkout guests app
         uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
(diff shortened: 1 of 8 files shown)


.github/workflows/integration-oci.yml
```diff
@@ -105,6 +106,13 @@ jobs:
           path: apps/call_summary_bot
           ref: ${{ matrix.call-summary-bot-versions }}
 
+      - name: Checkout circles app
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
```
SystemKeeper, line 110: Any reason to use a different checkout version here then for the other checkouts?

.github/workflows/integration-pgsql.yml
```diff
@@ -96,6 +97,13 @@ jobs:
           path: apps/call_summary_bot
           ref: ${{ matrix.call-summary-bot-versions }}
 
+      - name: Checkout circles app
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
```
SystemKeeper, line 101: Same as above

.github/workflows/integration-sqlite.yml
```diff
@@ -84,6 +85,13 @@ jobs:
           path: apps/call_summary_bot
           ref: ${{ matrix.call-summary-bot-versions }}
 
+      - name: Checkout circles app
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
```
SystemKeeper, line 89: Same as above

.github/workflows/integration-mysql.yml
```diff
@@ -93,6 +94,13 @@ jobs:
           path: apps/call_summary_bot
           ref: ${{ matrix.call-summary-bot-versions }}
 
+      - name: Checkout circles app
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
```
SystemKeeper, line 98: And here

.github/workflows/integration-oci.yml
nickvergessen: Copy pasted but others were updated since

SystemKeeper (approved):
Looks good!

PR #12244, opened 2024-04-29
test(chat): Add a test that editing old messages is blocked

nickvergessen:
### 🏁 Checklist


SystemKeeper:
Would still be nice if we could execute that through spreadcheats

SystemKeeper (approved):

PR #12257, opened 2024-05-03
fix(LeftSidebar): small glitch on sidebar scroll

ShGKme:
### ☑️ Resolves

* On scroll with debounce delay we add a 1px border
* It visually moves all the list after the scroll is finished or during the scroll
* When scroll position is 1px - it results in an infinite glitch

🏚️ Before | 🏡 After
-- | --
before | after
border-bug | image

### 🚧 Tasks

- [x] Add constant border to not change height during scrolling

### 🏁 Checklist

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -1014,9 +1014,10 @@ export default {
 	display: flex;
 	padding: 8px 4px 8px 12px;
 	align-items: center;
+	border-bottom: 1px solid transparent;
 
 	&--scrolled-down {
-		border-bottom: 1px solid var(--color-placeholder-dark);
+		border-bottom-color: var(--color-placeholder-dark);
 	}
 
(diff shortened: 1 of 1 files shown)


DorraJaouad (approved):

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -1014,6 +1014,7 @@ export default {
 	display: flex;
 	padding: 8px 4px 8px 12px;
 	align-items: center;
+	border-bottom: 1px solid transparent;
 
 	&--scrolled-down {
 		border-bottom: 1px solid var(--color-placeholder-dark);
```
Antreesy, line 1020: That also would be a nice addition to change specific parameter only
```suggestion
		border-bottom-color: var(--color-placeholder-dark);
```

Antreesy (approved):

ShGKme:
- Squashed

PR #12259, opened 2024-05-03
⛔ feat(ban): Contract for banning attendees and IP/ranges

nickvergessen:
### 🚧 Tasks

- [x] Allows to send a ban
- [x] Get a list of bans
- [x] Remove a ban
- [x] Experience a ban
    - [x] If conversation is named `ban user`, the user with id `banned` is banned (can not join)
    - [x] If conversation is named `ban guest`, all guests are banned (can not join)
- [x] *forgot to do:* Feature capability: `ban-v1`

### 🏁 Checklist

### DIFF
--- a/lib/Controller/BanController.php
+++ b/lib/Controller/BanController.php
@@ -1,0 +1,123 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Talk\Controller;
+
+use OCA\Talk\Middleware\Attribute\RequireModeratorParticipant;
+use OCA\Talk\Model\Attendee;
+use OCA\Talk\ResponseDefinitions;
+use OCP\AppFramework\Http;
+use OCP\AppFramework\Http\Attribute\PublicPage;
+use OCP\AppFramework\Http\DataResponse;
+use OCP\IRequest;
+
+/**
+ * @psalm-import-type TalkBan from ResponseDefinitions
+ */
+class BanController extends AEnvironmentAwareController {
+	public function __construct(
+		string $appName,
+		IRequest $request,
+	) {
+		parent::__construct($appName, $request);
+	}
+
+	/**
+	 * Ban an actor or IP address
+	 *
+	 * Required capability: `ban-v1`
+	 *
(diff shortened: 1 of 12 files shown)


lib/Controller/BanController.php
```diff
@@ -0,0 +1,126 @@
...
+	/**
+	 * Ban an actor or IP address
+	 *
+	 * Required capability: `ban-v1`
+	 *
+	 * @param string $token Conversation token
+	 * @param 'users'|'groups'|'circles'|'emails'|'federated_users'|'phones'|'ip' $actorType Type of actor to ban, or `ip` when banning a clients remote address
+	 * @psalm-param Attendee::ACTOR_*|'ip' $actorType Type of actor to ban, or `ip` when banning a clients remote address
+	 * @param string $actorId Actor ID or the IP address or range in case of type `ip`
+	 * @param string $internalNote Optional internal note
```
Antreesy, line 40: image

lib/Controller/BanController.php
```diff
@@ -0,0 +1,126 @@
...
+
+
+		return new DataResponse(
+			[
+				'id' => random_int(1, 1337),
+				'actorType' => $this->participant->getAttendee()->getActorType(),
+				'actorId' => $this->participant->getAttendee()->getActorId(),
+				'bannedType' => $actorType,
+				'bannedId' => $actorId,
+				'bannedTime' => time(),
```
Antreesy, line 63: do we need later in addition expand it to banStartTime and banEndTime (in case for temporal bans)?

Antreesy (approved):
Overall fine, but I'm struggling now to imagine how it would be used later on (from moderator perspective)

lib/Controller/BanController.php
nickvergessen: We don't support temporary bans for now 🤷

lib/Controller/BanController.php
```diff
@@ -0,0 +1,126 @@
...
+	 * @param string $actorId Actor ID or the IP address or range in case of type `ip`
+	 * @param string $internalNote Optional internal note
+	 * @return DataResponse<Http::STATUS_OK, TalkBan, array{}>|DataResponse<Http::STATUS_BAD_REQUEST, array{error: string}, array{}>
+	 *
+	 * 200: Ban successfully
+	 * 400: Actor information is invalid
+	 */
+	#[PublicPage]
+	#[RequireModeratorParticipant]
+	public function banActor(string $token, string $actorType, string $actorId, string $internalNote): DataResponse {
```
nickvergessen, line 48:
```suggestion
	public function banActor(string $token, string $actorType, string $actorId, string $internalNote = ''): DataResponse {
```

PR #12261, opened 2024-05-03
fix(LeftSidebar): get rid of server styles for settings button

ShGKme:
### ☑️ Resolves

* Settings button is a nix of global legacy server styles and `@nextcloud/vue`
  * It's style is a bit different from other buttons
  * It has double outline on focus
* Options 1: `secondary` - same as other similar apps - Mail, Calendar
* Options 2: `tertiary` - close to previous design
* Also, added a top padding

🏚️ Before | 🏡 After (secondary) | 🏡 After (tertiary)
-- | -- | --
image | image | image
image | image | image
image | . | image

### 🏁 Checklist

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -280,12 +280,13 @@
 		</template>
 
 		<template #footer>
-			<div id="app-settings">
-				<div id="app-settings-header">
-					<NcButton class="settings-button" @click="showSettings">
-						{{ t('spreed', 'Talk settings') }}
-					</NcButton>
-				</div>
+			<div class="left-sidebar__settings-button-container">
+				<NcButton type="tertiary" wide @click="showSettings">
+					<template #icon>
+						<Cog :size="20" />
+					</template>
+					{{ t('spreed', 'Talk settings') }}
+				</NcButton>
 			</div>
 		</template>
 	</NcAppNavigation>
@@ -298,6 +299,7 @@ import { ref } from 'vue'
 import AccountMultiplePlus from 'vue-material-design-icons/AccountMultiplePlus.vue'
 import AtIcon from 'vue-material-design-icons/At.vue'
 import ChatPlus from 'vue-material-design-icons/ChatPlus.vue'
+import Cog from 'vue-material-design-icons/Cog.vue'
 import FilterIcon from 'vue-material-design-icons/Filter.vue'
 import FilterRemoveIcon from 'vue-material-design-icons/FilterRemove.vue'
 import List from 'vue-material-design-icons/FormatListBulleted.vue'
@@ -393,6 +395,7 @@ export default {
 		Phone,
 		Plus,
 		ChatPlus,
+		Cog,
 		List,
 		Note,
 		NcEmptyContent,
@@ -1087,13 +1090,13 @@ export default {
 	}
 }
 
+.left-sidebar__settings-button-container {
+	padding: calc(2 * var(--default-grid-baseline));
+}
+
 :deep(.empty-content) {
 	text-align: center;
 	padding: 20% 10px 0;
-}
-
-.settings-button {
-	justify-content: flex-start !important;
 }
 
 :deep(.app-navigation__list) {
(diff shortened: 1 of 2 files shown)


src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -1084,15 +1087,15 @@ export default {
 	}
 }
 
+.left-sidebar__settings-button-container {
```
Antreesy, line 1090: You might now revert this commit: [9b21e9dc9d14bd8549086861ee663349ffffde8d](https://github.com/nextcloud/spreed/commit/9b21e9dc9d14bd8549086861ee663349ffffde8d)

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -280,12 +280,13 @@
...
 
 		<template #footer>
-			<div id="app-settings">
-				<div id="app-settings-header">
-					<NcButton class="settings-button" @click="showSettings">
-						{{ t('spreed', 'Talk settings') }}
-					</NcButton>
-				</div>
+			<div class="left-sidebar__settings-button-container">
+				<NcButton type="tertiary" wide @click="showSettings">
```
Antreesy, line 284: I'm fine with current state, as it draws much less attention (and looks better :yum: ):
Talk | Mail | Calendar
-- | -- | --
image | image | image

Antreesy (approved):

src/components/LeftSidebar/LeftSidebar.vue
nickvergessen: Should this be handled in/via https://github.com/nextcloud-libraries/nextcloud-vue/blob/master/src/components/NcAppNavigationSettings/NcAppNavigationSettings.vue or a new component, to avoid all the custom code/handling in X apps?

ShGKme: > Should this be handled in/via https://github.com/nextcloud-libraries/nextcloud-vue/blob/master/src/components/NcAppNavigationSettings/NcAppNavigationSettings.vue or a new component, to avoid all the custom code/handling in X apps?

I though about it but I'm not sure about it.

This cannot be handled via `NcAppNavigationSettings` (at least without making it a 2-in-1 component), because `NcAppNavigationSettings` is about "deprecated" collapsible settings, not a button.

This could be a new component, but its usage would probably have more variability than commons:
- In some apps, `AppNavigation` is fulfilled with navigation items and the settings button should look like a navigation item (Files, Photos)
- In some apps it would be secondary (Mail)
- In some apps it would be tertiary (Talk)

The only common is that it is a wide `NcButton`.

But we can discuss it with designers in `@nextcloud/vue`

ShGKme: > I'm fine with current state, as it draws much less attention (and looks better 😋 ):

The current state - in `main` or in this PR (tertiary)?

nickvergessen: > In some apps it would be secondary (Mail)
In some apps it would be tertiary (Talk)

This is exactly why I think it should be a common component.
I'm not sure the difference is intentional.

Antreesy: Talk and Mail use simple NcButton as a trigger for opening a dialog
Calendar has special component with "accordion" feature: https://github.com/nextcloud-libraries/nextcloud-vue/blob/master/src/components/NcAppNavigationSettings/NcAppNavigationSettings.vue

So I'd say, it would be better to summon design team here, to come to a conclusion on all three apps, then apply resulting styles on Mail and Vue-lib for Calendar

Antreesy: cc @marcoambrosini

marcoambrosini: The accordion should go asap, in favor of the settings dialog with proper explanations of the several settings.
I like both the secondary y and tertiary versions

Antreesy (approved):
Still good. Let's merge it for NC30 for now?

Antreesy:
Rebased, added style fixes for SearchBox (replace magic numbers with variables)

PR #12283, opened 2024-05-05
fix: throw error when target has DND on

### LINKED ISSUE #11282: Notify about DND status when trying to resend call notification
### How to use GitHub

---

## Steps to reproduce
1. Have a user that is on DND
2. Send them the call notification again

### Expected behaviour
Either it works or I know it didn't

### Actual behaviour
Moderator is not notified that no push notification was send.

skalidindi53:
### ☑️ Resolves

* Fix #11282 

### 🚧 Tasks

- Throws a new error if target is on DND

### DIFF
--- a/lib/Service/ParticipantService.php
+++ b/lib/Service/ParticipantService.php
@@ -1181,6 +1181,11 @@ class ParticipantService {
 			throw new DoesNotExistException('Room mismatch');
 		}
 
+		$userStatus = $this->userManager->getUserStatus(array ($attendee->getActorId()));
+		if ($userStatus[0] === IUser::DND){
+			throw new \InvalidArgumentException('User is in do not disturb mode, failed to send notification');
+		}
+
 		$sessions = $this->sessionMapper->findByAttendeeId($targetAttendeeId);
 		foreach ($sessions as $session) {
 			if ($session->getInCall() !== Participant::FLAG_DISCONNECTED) {
--- a/lib/Service/ParticipantService.php
+++ b/lib/Service/ParticipantService.php
@@ -71,6 +71,8 @@ use OCP\IUser;
 use OCP\IUserManager;
 use OCP\Security\ISecureRandom;
 use OCP\Server;
+use OCP\UserStatus\IManager as IUserStatusManager;
+use OCP\UserStatus\IUserStatus;
 
 class ParticipantService {
 
@@ -95,6 +97,7 @@ class ParticipantService {
 		private BackendNotifier $backendNotifier,
 		private ITimeFactory $timeFactory,
 		private ICacheFactory $cacheFactory,
+		private IUserStatusManager $userStatusManager,
 	) {
 	}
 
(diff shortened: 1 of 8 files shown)


lib/Service/ParticipantService.php
```diff
@@ -1181,6 +1184,16 @@ public function sendCallNotificationForAttendee(Room $room, Participant $current
 			throw new DoesNotExistException('Room mismatch');
 		}
 
+		$userStatus = $this->userStatusManager->getUserStatuses(array($attendee->getActorId()));
+
+		if (empty($userStatus)) {
+			throw new \InvalidArgumentException('User not found');
+		}
+
+		if ($userStatus[0]->getStatus() === IUserStatus::DND){
```
nickvergessen, line 1193: The user having no status is default and okay when they are "offline".
So it should not fail in this case and just continue without checking the user status.
So let's combine both checks to:
```suggestion
		if (isset($userStatus[0]) && $userStatus[0]->getStatus() === IUserStatus::DND) {
```

lib/Service/ParticipantService.php
```diff
@@ -1181,6 +1184,16 @@ public function sendCallNotificationForAttendee(Room $room, Participant $current
 			throw new DoesNotExistException('Room mismatch');
 		}
 
+		$userStatus = $this->userStatusManager->getUserStatuses(array($attendee->getActorId()));
```
nickvergessen, line 1187: *new array syntax*
```suggestion
		$userStatus = $this->userStatusManager->getUserStatuses([$attendee->getActorId()]);
```

lib/Service/ParticipantService.php
```diff
@@ -1181,6 +1184,16 @@ public function sendCallNotificationForAttendee(Room $room, Participant $current
...
+
+		if (empty($userStatus)) {
+			throw new \InvalidArgumentException('User not found');
+		}
+
+		if ($userStatus[0]->getStatus() === IUserStatus::DND){
+			throw new \InvalidArgumentException('User is on do not disturb mode, failed to send notification');
```
nickvergessen, line 1194: I would throw:
```suggestion
			throw new \InvalidArgumentException('status');
```

So you have the reason in the correct format available as `$e->getMessage()` for the last part of https://github.com/nextcloud/spreed/issues/11282#issuecomment-2089819787

skalidindi53: Thank you for the help, just updated the code with the suggested changes. Also, I'm a little confused with which controller should I add the `return new DataResponse(['error' => 'status'], Http::STATUS_BAD_REQUEST);`. Is it the CallController.php?

lib/Service/ParticipantService.php
```diff
@@ -1181,6 +1184,11 @@ public function sendCallNotificationForAttendee(Room $room, Participant $current
 			throw new DoesNotExistException('Room mismatch');
 		}
 
+		$userStatus = $this->userStatusManager->getUserStatuses([$attendee->getActorId()]);
+		if (isset($userStatus[0]) && $userStatus[0]->getStatus() === IUserStatus::DND) {
```
nickvergessen, line 1188: You can't know that, but the docs of the user status manager are lying. It's actually a map (userid being the key) not a list:
```suggestion
		if (isset($userStatus[$attendee->getActorId()]) && $userStatus[$attendee->getActorId()]->getStatus() === IUserStatus::DND) {
```

nickvergessen: Docs fixed in https://github.com/nextcloud/server/pull/45219

skalidindi53:
> Maybe you can also duplicate the test in
> 
> https://github.com/nextcloud/spreed/blob/0d1fc13e36e40a6cd8dfa80915ea77f06b835f80/tests/integration/features/callapi/notifications.feature#L63-L78
> 
> add a check for the the expected status code (and message) when the user status was set to `dnd` using
> 
> https://github.com/nextcloud/spreed/blob/e5327997bc3c54924bd8d3a157930dd517eb3cd2/tests/integration/features/bootstrap/FeatureContext.php#L4315

Added the test case as well

nickvergessen (approved):

PR #12288, opened 2024-05-06
fix(LobbySettings): convert to string before passing it to translation

DorraJaouad:
### ☑️ Resolves

It seems that placeholders in [translation](https://github.com/nextcloud-libraries/nextcloud-l10n/blob/main/lib/translation.ts#L70) are not converted to strings, at least to minimize [unhandled cases](https://github.com/nextcloud-libraries/nextcloud-l10n/blob/e0b16b54fc946273e31e99e0b16d5e85b4ff9560/lib/translation.ts#L88)  :')  

🏚️ Before | 🏡 After
-- | --
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/ConversationSettings/LobbySettings.vue
+++ b/src/components/ConversationSettings/LobbySettings.vue
@@ -145,7 +145,7 @@ export default {
 				return ''
 			}
 			const date = new Date(this.lobbyTimer)
-			return t('spreed', 'Start time: {date}', { date })
+			return t('spreed', 'Start time: {date}', { date: date.toString() })
 		},
 
 		getRelativeTime() {


src/components/ConversationSettings/LobbySettings.vue
```diff
@@ -145,7 +145,7 @@ export default {
 				return ''
 			}
 			const date = new Date(this.lobbyTimer)
-			return t('spreed', 'Start time: {date}', { date })
+			return t('spreed', 'Start time: {time}', { time: date.toString() })
```
nickvergessen, line 148:
```suggestion
			return t('spreed', 'Start time: {date}', { date: date.toString() })
```
To not create a new string they have to translate?

nickvergessen (approved):

PR #12289, opened 2024-05-07
fix(chat): focus submit on upload attachments without caption

ShGKme:
### ☑️ Resolves

* Fix https://github.com/nextcloud/talk-desktop/issues/651
* Regression of https://github.com/nextcloud/spreed/pull/10730 for old servers:
  * When captions aren't supported submit button should be focus as it was before adding captions
  * Otherwise it is focused on "Close" button and `Enter` results in dismissing sending files
* Removed optional chaining `?.` to not miss error here
* Note, `getContainerId` did not returned actual promise from `nextTick`. It word because of `await` just as `await Promise.resolve()` but it is not guaranteed

🏚️ Before | 🏡 After
-- | --
image | image
image | image

### 🚧 Tasks

- [x] Focus `submitButton` if there is no input

### 🏁 Checklist

### DIFF
--- a/src/components/NewMessage/NewMessageUploadEditor.vue
+++ b/src/components/NewMessage/NewMessageUploadEditor.vue
@@ -55,7 +55,7 @@
 				<NcButton type="tertiary" @click="handleDismiss">
 					{{ t('spreed', 'Dismiss') }}
 				</NcButton>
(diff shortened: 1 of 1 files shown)


DorraJaouad (approved):
Tested ✅

Antreesy (approved):
LGTM

PR #12300, opened 2024-05-07
fix(CallView): light improvement

DorraJaouad:
### ☑️ Resolves

* Fix #…

- When joining a call in group conversation, it is set to grid view even for "connecting" view which makes the tile looks weird. I enforce a speaker view when connecting so to keep a consistent view for all joiners and to avoid having an expanded local video tile.
- Stop following button lost restoring the previous view state because of some watchers enforcing the temporary view (speaker view). I omitted those as `startPresentation` already handles it. 

🏚️ Before | 🏡 After
-- | --
image |  image |

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist


src/components/CallView/CallView.vue
```diff
@@ -13,10 +13,10 @@
 			:local-shared-data="localSharedData" />
 
 		<template v-else>
-			<EmptyCallView v-if="!callParticipantModels.length && !screenSharingActive && !isGrid" :is-sidebar="isSidebar" />
+			<EmptyCallView v-if="!callParticipantModels.length && !screenSharingActive" :is-sidebar="isSidebar" />
 
 			<div id="videos">
-				<div v-if="!isGrid" class="video__promoted" :class="{'full-page': showFullPage}">
+				<div v-if="!isGrid || !callParticipantModels.length" class="video__promoted" :class="{'full-page': showFullPage}">
```
DorraJaouad, line 19: This is because stripe relies on `video__promoted` class and it is needed for the fake speaker view in group conversation  (`isGrid` remains truthy) too. 

Another alternative is to condition style `flex-direction: column` for `#videos` and not to show this video promoted div but there is a high risk of breaking something.

src/components/CallView/CallView.vue
```diff
@@ -397,18 +397,6 @@ export default {
 			}
 		},
 
-		showSelectedVideo(newVal) {
-			if (newVal) {
-				this.$store.dispatch('setCallViewMode', { isGrid: false })
-			}
-		},
-
-		showSelectedScreen(newVal) {
```
Antreesy, line 406: Maybe unrelated to current changes:
it is handled by code somewhere else, but it will still break, if you change any parameter with this dispatch;
https://github.com/nextcloud/spreed/blob/1ccbf67da231c37801c1406b10000144125aec35/src/store/callViewStore.js#L159-L162

For example:
`<Grid>` - start screenshare - `<Speaker>` - end screenshare - `<Grid>`
`<Grid>` - start screenshare - `<Speaker>` - open stripe `<Speaker>` - end screenshare - `<Speaker>` :warning: 

IMO, `lastIsGrid` and `lastIsStripeOpen` should only be changed, if `isGrid` or `isStripeOpen` were passed with a payload, and are different from the previous state

src/components/CallView/CallView.vue
Antreesy: I agree that this small change would be sufficient (though it might still jump from speaker to grid view with the first joined participant)

Antreesy (approved):

src/components/CallView/CallView.vue
DorraJaouad: Agree, and for your example, the main reason is that `clearLast` was passed as default truthy when opening the stripe.

src/components/CallView/CallView.vue
DorraJaouad: yeah, but in both cases, it will have the same risk of view jumping from gid to speaker. Though, it's not visible at the moment.

src/components/CallView/CallView.vue
DorraJaouad: There is a test `callViewStore › call view mode and presentation › switching modes during presentation does not resets it after it ends`

Is it meant not to reset it ?

Antreesy: Good catch, same could be added for:
- `stopPresentation` (will be overwritten anyway)
- `changeView` (same as handleClickStripeCollapse)

> Is it meant not to reset it ?

Looks like test is not checking anything. Plus isGrid and isStripeOpen can't be both true at the same time. Should be a series of test with different initial values, like `start presentation switches off grid view and restores when it ends`

DorraJaouad: > stopPresentation (will be overwritten anyway)

Here, we are using the stored state as current state so it makes sense to empty the stored one ?

> changeView (same as handleClickStripeCollapse)

If user is updating their preferences, there is no need to keep the previous state (like resetting preferences). 

> Plus isGrid and isStripeOpen can't be both true at the same time.

I also noticed this abnormal logic. 
What cannot be the same value are isStripe and isGrid but isStripeOpen is independent from isGrid (which shouldn't be) and can have the same value.

DorraJaouad: > Looks like test is not checking anything

regardless of the implementation, the title of the test implies that it is expected not to restore it. If we agree on restoring after closing presentation, I will change that test as we are changing the behavior.

Antreesy: > If user is updating their preferences, there is no need to keep the previous state (like resetting preferences).

We also don't have to modify it twice in one call, `true` -> `null` -> `false`, that's what I'm trying to avoid

> > stopPresentation (will be overwritten anyway)

That's actually the only place where we're using lastIsGrid and lastIsStripeOpen values :warning:

DorraJaouad: Okay, let's keep this PR for fixing issues. Refactoring (removing unnecessary updates) will come next.

Antreesy (approved):
As it's aligned with tests, let's go with it for now.
I'd still consider to discuss all possible variants and their outputs as follow-up (like with a table we did for scrolling)

ShGKme (dismissed):
The only thing that looks weird now, is that on the empty call view the switch does nothing. You click on "Grid view" or "Speaker view" and nothing got changed. But I don't see any alternatives.

image

ShGKme:
1. On the call, start following somebody and open stripe
2. Start screensharing from another participant
3. Stop screensharing
4. No you are on the grid view instead of the previous "Following + open stripe"

Is that expected?

Antreesy:
> But I don't see any alternatives.

Hide an option, while there's nothing to place on grid?

src/components/CallView/CallView.vue
```diff
@@ -646,9 +646,8 @@ export default {
 			if (this.isSidebar) {
 				return
 			}
-			this.$store.dispatch('startPresentation')
 			this.$store.dispatch('selectedVideoPeerId', peerId)
-			this.isLocalVideoSelected = false
+			this.$store.dispatch('startPresentation')
```
Antreesy, line 650: Couldn't reproduce an issue here, but `isLocalVideoSelected` was unused in the code, and method now is aligned with `handleClickLocalVideo` by behavior :+1:

src/store/callViewStore.js
```diff
@@ -15,6 +15,7 @@ const state = {
 	isViewerOverlay: false,
 	isGrid: false,
 	isStripeOpen: true,
+	isEmptyCallView: false,
```
Antreesy, line 18: Shouldn't it be true by default? We don't have any participants models by the time we just have joined the call, so don't need to immediately rewrite it

Antreesy: And set back to true after leaving the call (to prepare app for the next call)

DorraJaouad:
> TypeError: Cannot read properties of null (reading 'attributes')
    at VueComponent.shouldShowPresenterOverlay

This was unrelated, but fixed as it is involved in "light improvement"  :p

src/components/CallView/CallView.vue
DorraJaouad: There was no issue, just to align with `handleClickLocalVideo`

src/components/TopBar/TopBarMenu.vue
```diff
@@ -65,7 +65,7 @@
 			</template>
 
 			<!-- Call layout switcher -->
-			<NcActionButton v-if="showActions && isInCall"
+			<NcActionButton v-if="showCallLayoutSwitch"
```
Antreesy, line 68:
```suggestion
			<NcActionButton v-if="$store.getters.isEmptyCallView"
```
And just move to the template above?

Antreesy (approved):
Last nitpick from me, can't tell what else we could enhance atm. Behaves much better, in my opinion

src/components/CallView/CallView.vue
```diff
@@ -322,13 +322,17 @@ export default {
 		},
 
 		shouldShowPresenterOverlay() {
-			return this.shownRemoteScreenCallParticipantModel.attributes.videoAvailable || this.isModelWithVideo(this.shownRemoteScreenCallParticipantModel)
+			return this.shownRemoteScreenCallParticipantModel?.attributes?.videoAvailable || this.isModelWithVideo(this.shownRemoteScreenCallParticipantModel)
```
ShGKme, line 325: Can you confirm that:
1. This is expected `shownRemoteScreenCallParticipantModel` to be nullish here (in which case)
2. Even with `shownRemoteScreenCallParticipantModel`, it is possible that `shownRemoteScreenCallParticipantModel.attributes` is nullish

DorraJaouad:
```js
		shownRemoteScreenCallParticipantModel() {
			if (!this.shownRemoteScreenPeerId) {
				return null
			}
			return this.callParticipantModels.find(callParticipantModel => {
				return callParticipantModel.attributes.peerId === this.shownRemoteScreenPeerId
			})
		},
```

1. It can be nullish when no other user screensharing.
2. Well, here it shouldn't be nullish, just added it to avoid extra exception. If it is null or undefined, it will be shown in `shownRemoteScreenCallParticipantModel` already.

ShGKme: > just added it to avoid extra exception

We should avoid using optional chaining for such purpose. While it "removes" the error from the console and an exception from the function, it does not fix the real issue if any. It only hides it.

If there is a high level issue, it will be harder to find it in future, because instead of an exception for a specific line, we have everything working, just incorrect in further computing.

If there is no issues, then additional check is just useless, adding cognitive complexity.

Noting that, because we have a lot juch optional changing usage.

PR #12031, opened 2024-04-09
perf(call): Debug starting a big call takes long

nickvergessen:
## 🛠️ API Checklist

- 29 retry of #11208 

### 🏁 Checklist

- [ ] ⛑️ Tests (unit and/or integration) are included or not possible
- [ ] 📘 API documentation in `docs/` has been updated or is not required
- [ ] 🔖 Capability is added or not needed

PR #11208, opened 2023-12-12
[stable28] Debug starting a big call takes long

nickvergessen:
```json
{"reqId":"j9UxMXl1CcVRQo3cCNkj","level":2,"time":"2023-12-12T14:01:37+00:00","remoteAddr":"2001:…","user":"me","app":"no app in context","method":"POST","url":"/ocs/v2.php/apps/spreed/api/v4/call/…","message":"Debugging step #0: 1702389697.4765","userAgent":"Mozilla/5.0 (Linux) Nextcloud-Talk v0.19.0","version":"28.0.0.11","data":[]}
{"reqId":"j9UxMXl1CcVRQo3cCNkj","level":2,"time":"2023-12-12T14:01:37+00:00","remoteAddr":"2001:…","user":"me","app":"no app in context","method":"POST","url":"/ocs/v2.php/apps/spreed/api/v4/call/…","message":"Debugging step #1: 1702389697.478","userAgent":"Mozilla/5.0 (Linux) Nextcloud-Talk v0.19.0","version":"28.0.0.11","data":[]}
{"reqId":"j9UxMXl1CcVRQo3cCNkj","level":2,"time":"2023-12-12T14:01:37+00:00","remoteAddr":"2001:…","user":"me","app":"no app in context","method":"POST","url":"/ocs/v2.php/apps/spreed/api/v4/call/…","message":"Debugging step #3: 1702389697.4782","userAgent":"Mozilla/5.0 (Linux) Nextcloud-Talk v0.19.0","version":"28.0.0.11","data":[]}
{"reqId":"j9UxMXl1CcVRQo3cCNkj","level":2,"time":"2023-12-12T14:01:37+00:00","remoteAddr":"2001:…","user":"me","app":"no app in context","method":"POST","url":"/ocs/v2.php/apps/spreed/api/v4/call/…","message":"Debugging step #4: 1702389697.4783","userAgent":"Mozilla/5.0 (Linux) Nextcloud-Talk v0.19.0","version":"28.0.0.11","data":[]}
{"reqId":"j9UxMXl1CcVRQo3cCNkj","level":2,"time":"2023-12-12T14:01:37+00:00","remoteAddr":"2001:…","user":"me","app":"no app in context","method":"POST","url":"/ocs/v2.php/apps/spreed/api/v4/call/…","message":"Debugging step #10: 1702389697.4785","userAgent":"Mozilla/5.0 (Linux) Nextcloud-Talk v0.19.0","version":"28.0.0.11","data":[]}
[...]

PR #12303, opened 2024-05-07
fix(notifications): Preparse call notifications for improved performance

nickvergessen:
Reducing the roundtrips in the notification providers by only parsing each language once

Previous PRs:
- https://github.com/nextcloud/spreed/pull/11208
- https://github.com/nextcloud/spreed/pull/12031

### 🏁 Checklist

### DIFF
--- a/lib/Notification/Listener.php
+++ b/lib/Notification/Listener.php
@@ -24,10 +24,12 @@ use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\EventDispatcher\Event;
 use OCP\EventDispatcher\IEventDispatcher;
 use OCP\EventDispatcher\IEventListener;
+use OCP\IConfig;
 use OCP\IDBConnection;
 use OCP\IUser;
 use OCP\IUserSession;
 use OCP\Notification\IManager;
+use OCP\Notification\INotification;
 use Psr\Log\LoggerInterface;
 
 /**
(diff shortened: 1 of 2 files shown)


lib/Notification/Listener.php
```diff
@@ -273,17 +279,52 @@ protected function sendCallNotifications(Room $room): void {
...
+					if (!isset($this->preparedCallNotifications[$languageCode])) {
+						$translatedNotification = clone $notification;
+
+						$this->notificationManager->setPreparingPushNotification(true);
+						$this->preparedCallNotifications[$languageCode] = $this->notificationProvider->prepare($translatedNotification, $languageCode);
+						$this->notificationManager->setPreparingPushNotification(false);
+						$userNotification = $translatedNotification;
+					} else {
+						$userNotification = $this->preparedCallNotifications[$languageCode];
+					}
```
Antreesy, line 320: Would it be the same? 
```suggestion
					}
					$userNotification = $this->preparedCallNotifications[$languageCode];
```

lib/Notification/Listener.php
```diff
@@ -273,17 +279,52 @@ protected function sendCallNotifications(Room $room): void {
...
+					// Get the settings for this particular user, then check if we have notifications to email them
+					$languageCode = $userLanguages[$userId] ?? $fallbackLang;
+
+					if (!isset($this->preparedCallNotifications[$languageCode])) {
+						$translatedNotification = clone $notification;
+
+						$this->notificationManager->setPreparingPushNotification(true);
+						$this->preparedCallNotifications[$languageCode] = $this->notificationProvider->prepare($translatedNotification, $languageCode);
+						$this->notificationManager->setPreparingPushNotification(false);
+						$userNotification = $translatedNotification;
```
Antreesy, line 317: Because I'm not sure, that `$this->notificationProvider->prepare()` will mutate `$translatedNotification`

lib/Notification/Listener.php
nickvergessen: Right, fixed

Antreesy (approved):

lib/Notification/Listener.php
SystemKeeper: I guess that means no?

nickvergessen: The object is mutated, that is exactly why this was needed and I couldn't reuse `$notification` directly.
But yeah improved the flow now to remove the else

PR #12311, opened 2024-05-08
feat(capabilities): Expose which capabilities should be considered lo…

nickvergessen:
…cal vs federated

### ☑️ Resolves

* Ref #10680

### 🏁 Checklist


docs/capabilities.md
```diff
@@ -56,19 +56,19 @@
...
 
 ## 11.0
-* `chat-read-status` - On conversation API v3 and the chat API the last common read message is exposed which can be used to update the "read status" flag of own chat messages. The info should be shown only when the user also shares their read status. The user's value can be found in `config => chat => read-privacy`.
-* `listable-rooms` - Conversations can searched for even when not joined. A "listable" attribute set on rooms defines the scope of who can find it.
-* `phonebook-search` - Is present when the server has the endpoint to search for phone numbers to find matches in the accounts list
+* `chat-read-status` (local) - On conversation API v3 and the chat API the last common read message is exposed which can be used to update the "read status" flag of own chat messages. The info should be shown only when the user also shares their read status. The user's value can be found in `config => chat => read-privacy`.
```
SystemKeeper, line 59: According to the original issue this was remote and in iOS we are using the remote one at the moment.

docs/capabilities.md
```diff
@@ -114,26 +114,26 @@
 * `chat-keep-notifications` - Whether messages can be retrieved without marking notifications as read
 
 ## 17
-* `avatar` - Avatar of conversation
+* `avatar` (local) - Avatar of conversation
```
SystemKeeper, line 117: There was an open question in the issue if it should be remote, since we load the image from there? Current implementation on iOS uses the local capability, though.

SystemKeeper (commented):
Maybe we should also mark `config => signaling => hello-v2-token-key` as local for now?

SystemKeeper (approved):

Antreesy (approved):
Shouldn't break anything, and would be helpful for detecting local-only features on a frontend. Thanks!

nickvergessen:
Added federation configs and signaling configs to it, as they are also both "local" at the moment

PR #12318, opened 2024-05-10
fix(messageStore): adjust getters and tests

Antreesy:
### ☑️ Resolves

* required for #12208
  * jest fails on rootGetters, also tests are done without respecting new capability
  * not a part of mentioned PR, so splitted

No visual changes, but might fix some visual glitches, while server response hasn't come

### 🏁 Checklist


src/services/messagesService.ts
```diff
@@ -192,7 +193,7 @@ const postRichObjectToConversation = async function(token: string, { objectType,
  * @param lastReadMessage id of the last read message to set
  * @param options request options
  */
-const updateLastReadMessage = async function(token: string, lastReadMessage: setReadMarkerParams, options: object): setReadMarkerResponse {
+const updateLastReadMessage = async function(token: string, lastReadMessage?: number|null, options?: object): setReadMarkerResponse {
```
DorraJaouad, line 196:
```suggestion
const updateLastReadMessage = async function(token: string, lastReadMessage?: number|null|undefined, options?: object): setReadMarkerResponse {
```

?

src/store/messagesStore.js
```diff
@@ -833,6 +833,7 @@ const actions = {
 			context.dispatch('updateLastReadMessage', { token, id: null, updateVisually })
 			return
 		}
+		// federated conversations don't proxying lastMessage id
```
DorraJaouad, line 836:
```suggestion
		// federated conversations don't proxy lastMessage id
```

src/store/messagesStore.js
```diff
@@ -861,10 +862,12 @@ const actions = {
 			return
 		}
 
+		const idToUpdate = (id === null) ? conversation.lastMessage.id : id
```
DorraJaouad, line 865: For fed conversations; there is no `conversation.lastMessage.id`, so if it is null, it should stay `null` ?

DorraJaouad: and the capability `markAsReadWithoutLast`, was it added for feds only or for all conversations in v19 ?

src/services/messagesService.ts
Antreesy: `lastReadMessage?:` is already undefined

DorraJaouad: Ah okay, I thought it's better to specify it too in the possible types

src/store/messagesStore.js
Antreesy: It's applied since v19.
`idToUpdate` needed only for immediate frontend handling, and for federated ones, null wouldn't be correct. Best approach would be to get last id from messagesList

DorraJaouad: ok but that will render an error of undefined for federated conversations, should it get the id of the last message in `messagesList` ?

DorraJaouad: >  Best approach would be to get last id from messagesList

agree

Antreesy: As it's only a visual change, I've decided to skip one step for federated conversations.
- LeftSidebar will be updated anyway after server response received
- Visual marker will be updated, only if chat is loaded -> last message id is known

DorraJaouad (approved):

PR #12319, opened 2024-05-10
fix(federation): Don't send notifications for most system messages in…

### LINKED ISSUE #12307: Edit in a federated conversation moves conversation to the top again
### How to use GitHub

---

## Steps to reproduce
1. A conversation hosted on server "S" adds a user from server "C" as a federated user
2. Messages are exchanged and the conversation is not the top of the list anymore
3. A user from server "S" edits a message

### Expected behaviour
The conversation should not be moved to the top of the conversation list again (this seems to be the case for non-federated conversations)

### Actual behaviour
The conversation is moved to the top again, although it is still marked as read. 
Not sure if we need to take into account if it was the last message because of soft-updating, but in the observed case it was not the last message.

## Talk app

**Talk app version:** 19.0.0

nickvergessen:
… federation

### ☑️ Resolves

* Fix #12307 

### 🚧 Tasks

- [x] Don't create push notifications
- [x] Don't bump the last activity

### 🏁 Checklist

### DIFF
--- a/lib/Notification/FederationChatNotifier.php
+++ b/lib/Notification/FederationChatNotifier.php
@@ -63,9 +63,11 @@ class FederationChatNotifier {
 			}
 		} elseif ($participant->getAttendee()->getNotificationLevel() === Participant::NOTIFY_ALWAYS
 			|| ($defaultLevel === Participant::NOTIFY_ALWAYS && $participant->getAttendee()->getNotificationLevel() === Participant::NOTIFY_DEFAULT)) {
-			$notification = $this->createNotification($room, $message, 'chat');
-			$notification->setUser($participant->getAttendee()->getActorId());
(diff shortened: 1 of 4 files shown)


lib/Notification/FederationChatNotifier.php
```diff
@@ -111,6 +113,14 @@ protected function isMentionedAll(Room $room, ProxyCacheMessage $message): bool
 		return false;
 	}
 
+	protected function isUserMessageOrRelevantSystemMessage(?string $systemMessage): bool {
+		return $systemMessage === null
+			|| $systemMessage === ''
+			|| $systemMessage === 'object_shared'
+			|| $systemMessage === 'poll_closed'
+			|| $systemMessage === 'file_shared';
```
SystemKeeper, line 121: `reaction`?

PR #12338, opened 2024-05-13
fix(settings): don't allow guests when breakout rooms are set

Antreesy:
### ☑️ Resolves

* keep error handling in the store only
* disable option in dialog window

🏚️ Before | 🏡 After
-- | --
image | unreachable
image | Screenshot from 2024-05-13 11-14-07

### 🏁 Checklist

### DIFF
--- a/src/components/ConversationSettings/LinkShareSettings.vue
+++ b/src/components/ConversationSettings/LinkShareSettings.vue
@@ -10,8 +10,11 @@
 		</h4>
 
 		<template v-if="canModerate">
+			<p v-if="hasBreakoutRooms" class="app-settings-section__hint">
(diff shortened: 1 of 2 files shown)


src/components/ConversationSettings/LinkShareSettings.vue
```diff
@@ -11,7 +11,7 @@
 
 		<template v-if="canModerate">
 			<NcCheckboxRadioSwitch :checked="isSharedPublicly"
-				:disabled="isSaving"
+				:disabled="isDisabled"
```
DorraJaouad, line 14: Now I think we should add a hint here , why it is disabled.

DorraJaouad (approved):
🦅

PR #12361, opened 2024-05-15
feat(CallView): introduce dynamic order

### LINKED ISSUE #11393: Dynamic arrangement of tiles in a call in grid mode
### How to use GitHub

---

In grid mode in a Talk call, if the sheer number of participants make the new tiles move to new pages navigable through arrow buttons, we don't see who's talking if that someone is on a different page. It gets harder if more than 2 people are talking and they are on different pages (grid pages) so keeping track becomes cumbersome.

A dynamic grid which bumps a speaking tile to the first grid page. It might help to reduce the size of the tiles a bit for large number of participants to reduce tile swaps.

The alternative which I use and is present in the Talk app is the "Speaker view" but quick switches of different speakers makes almost half the screen change colours very quickly and is not ideal.  
The tile swap can also be ported to the bottom participant list in this view as well but not really necessary.

**Additional context**

DorraJaouad:
### ☑️ Resolves

* Fix #11393

- [x] After following, go back to the view type that was on just before (done)
- [x] Default order :
1. user is screensharing
2. user is speaking
3. user has speaking rights and video with video ON
4. user has speaking rights and video with video OFF
5. user with no permissions

There is no differentiation between guests, users and moderators (hopefully in v2 😇)

## Tiles dynamic order

- [x] in Stripe (so it is Speaker view) and in grid view, last speaker should be placed first in their category (1. and 2.)
- [x] Same for screensharing
- [x] Re-ordering is only applicable when the speaker is not in the first page 

> [!TIP]
> For testing with a few tiles, make this condition always falsy
> and I suggest to review changes all at once and not by commit, it's messy..

```js
// if model is already in the first page, do nothing
	if (false && this.orderedVideos.slice(0, this.slots).find(video => video.attributes.peerId === model.attributes.peerId)) {
		return
	}
```

## Additional (follow-up): 

- [ ] Webinar mask: when it is switched on by a moderator the tiles of attendees who don't have permissions will be hidden (showing only total number in a tile).
- [ ] Styling the low-num-of-participants call as they are in one page and space should be minimized.
- [ ] When joining the call, It should be speaker view for the first tile and if another tile was added, it becomes a grid

🏚️ Before | 🏡 After
-- | --

### 🚧 Tasks

> [!CAUTION]
> This section for the side effects that I need to handle

- [x] waiting for participants to be fetched (around 3sec) is noticeable when joining, maybe I need to add to expand empty view for that case or add another view as tiles placeholders loading (we know how many participants at that time) 
- [x] skip any kind of ordering when user is guest as they don't have participants list thus, who has mic permissions/ moderator etc
### 🏁 Checklist

### DIFF
--- a/src/components/CallView/Grid/Grid.vue
+++ b/src/components/CallView/Grid/Grid.vue
@@ -152,6 +152,8 @@ import LocalVideo from '../shared/LocalVideo.vue'
 import VideoBottomBar from '../shared/VideoBottomBar.vue'
 import VideoVue from '../shared/VideoVue.vue'
 
+import { PARTICIPANT, ATTENDEE } from '../../../constants.js'
+
 // Max number of videos per page. `0`, the default value, means no cap
 const videosCap = parseInt(loadState('spreed', 'grid_videos_limit'), 10) || 0
 const videosCapEnforced = loadState('spreed', 'grid_videos_limit_enforced') || false
@@ -268,6 +270,9 @@ export default {
 			// Timer for the videos bottom bar
 			showVideoOverlayTimer: null,
 			debounceMakeGrid: () => {},
+			tempPromotedModels: [],
+			unpromoteSpeakerTimer: {},
+			promotedHistoryMask: [],
 		}
 	},
 
@@ -317,11 +322,11 @@ export default {
 			const slots = (this.videosCap && this.videosCapEnforced) ? Math.min(this.videosCap, this.slots) : this.slots
 
 			// Slice the `videos` array to display the current page of videos
-			if (((this.currentPage + 1) * slots) >= this.videos.length) {
-				return this.videos.slice(this.currentPage * slots)
+			if (((this.currentPage + 1) * slots) >= this.orderedVideos.length) {
+				return this.orderedVideos.slice(this.currentPage * slots)
 			}
 
-			return this.videos.slice(this.currentPage * slots, (this.currentPage + 1) * slots)
+			return this.orderedVideos.slice(this.currentPage * slots, (this.currentPage + 1) * slots)
 		},
 
 		isLessThanTwoVideos() {
@@ -420,7 +425,7 @@ export default {
 		// Hides or displays the `grid-navigation next` button
 		hasNextPage() {
 			if (this.displayedVideos.length !== 0 && this.hasPagination) {
-				return this.displayedVideos.at(-1) !== this.videos.at(-1)
+				return this.displayedVideos.at(-1) !== this.orderedVideos.at(-1)
 			} else {
 				return false
 			}
@@ -429,7 +434,7 @@ export default {
 		// Hides or displays the `grid-navigation previous` button
 		hasPreviousPage() {
 			if (this.displayedVideos.length !== 0 && this.hasPagination) {
-				return this.displayedVideos[0] !== this.videos[0]
+				return this.displayedVideos[0] !== this.orderedVideos[0]
 			} else {
 				return false
 			}
@@ -478,6 +483,70 @@ export default {
 		stripeOpen() {
 			return this.$store.getters.isStripeOpen && !this.isRecording
 		},
+
+		participantsInitialised() {
+			return this.$store.getters.participantsInitialised(this.token)
+		},
+
+		isGuestNonModerator() {
+			return this.$store.getters.getActorType() === ATTENDEE.ACTOR_TYPE.GUESTS
+				&& this.$store.getters.conversation(this.token).participantType !== PARTICIPANT.TYPE.GUEST_MODERATOR
+		},
+
+		orderedVideos() {
+			// Dynamic ordering is not possible for guests because
+			// participants store is not initialized
+			if (this.isGuestNonModerator) {
+				return this.videos
+			}
+
+			if (!this.participantsInitialised) {
+				return []
+			}
+
+			const objectMap = {
+				modelsWithScreenshare: [],
+				modelsTempPromoted: [],
+				modelsWithVideoEnabled: [],
+				modelsWithAudioOnly: [],
+				modelsWithNoPermissions: [],
+			}
+			const screensSet = new Set(this.screens)
+			const tempPromotedModelsSet = new Set(this.tempPromotedModels.map(model => model.attributes.nextcloudSessionId))
+			const videoTilesMap = new Map()
+			const audioTilesMap = new Map()
+
+			this.callParticipantModels.forEach((model) => {
+				if (screensSet.has(model.attributes.peerId)) {
+					objectMap.modelsWithScreenshare.push(model)
+				} else if (tempPromotedModelsSet.has(model.attributes.nextcloudSessionId)) {
+					objectMap.modelsTempPromoted.push(model)
+				} else if (this.isModelWithVideo(model)) {
+					videoTilesMap.set(model.attributes.nextcloudSessionId, model)
+				} else if (this.isModelWithAudio(model)) {
+					audioTilesMap.set(model.attributes.nextcloudSessionId, model)
+				} else {
+					objectMap.modelsWithNoPermissions.push(model)
+				}
+			})
+
+			objectMap.modelsWithVideoEnabled = this.getOrderedTiles(videoTilesMap, this.promotedHistoryMask)
+			objectMap.modelsWithAudioOnly = this.getOrderedTiles(audioTilesMap, this.promotedHistoryMask)
+
+			return [...objectMap.modelsWithScreenshare,
+				...objectMap.modelsTempPromoted,
+				...objectMap.modelsWithVideoEnabled,
+				...objectMap.modelsWithAudioOnly,
+				...objectMap.modelsWithNoPermissions]
+		},
+
+		speakers() {
+			return this.callParticipantModels.filter(model => model.attributes.speaking)
+		},
+
+		speakersWithAudioOff() {
+			return this.tempPromotedModels.filter(model => !model.attributes.audioAvailable)
+		},
 	},
 
 	watch: {
@@ -522,6 +591,24 @@ export default {
 			if (this.currentPage >= this.numberOfPages) {
 				this.currentPage = Math.max(0, this.numberOfPages - 1)
 			}
+		},
+
+		speakers(models) {
+			models.forEach(model => {
+				this.promoteSpeaker(model)
+				clearTimeout(this.unpromoteSpeakerTimer[model.attributes.nextcloudSessionId])
+			})
+		},
+
+		speakersWithAudioOff(newModels, oldModels) {
+			newModels.forEach(speaker => {
+				if (oldModels.includes(speaker)) {
+					return
+				}
+				this.unpromoteSpeakerTimer[speaker.attributes.nextcloudSessionId] = setTimeout(() => {
+					this.unpromoteSpeaker(speaker)
+				}, 10000)
+			})
 		},
 	},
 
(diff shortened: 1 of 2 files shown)


src/components/CallView/Grid/Grid.vue
```diff
@@ -522,6 +547,32 @@ export default {
...
 		},
+
+		callParticipantModels: {
+			immediate: true,
+			handler(newModels, oldModels) {
+				if (!this.participantsInitialised) {
+					return
+				}
+				// TODO: patch the array with the new models
+				this.patchOrderedVideos(newModels)
```
DorraJaouad, line 558: I hope we didn't lose any reactivity on attributes that we need in `displayedVideos` 😇

src/components/CallView/Grid/Grid.vue
```diff
@@ -882,7 +882,7 @@ export default {
 			if (this.orderedVideos.slice(0, this.slots).find(video => video.attributes.peerId === model.attributes.peerId)) {
 				return
 			}
-			const participant = this.$store.getters.getParticipantByPeerId(this.token, model.attributes.peerId)
+			const participant = this.$store.getters.getParticipantByPeerId(this.token, model.attributes.nextcloudSessionId)
```
DorraJaouad, line 885: it's more reliable

src/components/CallView/Grid/Grid.vue
```diff
@@ -828,6 +922,82 @@ export default {
...
+					this.tilesWithAudioMask.splice(audioIndex, 1)
+				}
+
+				this.videosTempPromoted.unshift(model)
+			}
+		},
+
+		getOrderedTiles(tiles, orderMask) {
+			// Create a map for quick lookup
+			const tilesMap = new Map(tiles.map(tile => [tile.attributes.nextcloudSessionId, tile]))
```
Antreesy, line 989: Since we iterate the models in computed one by one, we can provide a Map from there, and don't recalculate it again in method:
```suggestion
		getOrderedTiles(tilesMap, orderMask) {
```

Antreesy (commented):
Since it was almost completely rewritten and differs from the very first commit, maybe rebase and re-commit again?)

src/components/CallView/Grid/Grid.vue
```diff
@@ -478,6 +485,65 @@ export default {
...
+			}
+			// Dynamic ordering is not possible for guests because
+			// participants store is not initialized
+			if (this.actorType === ATTENDEE.ACTOR_TYPE.GUESTS) {
+				return this.videos
+			}
+
+			return [...this.modelsWithScreenshare, ...this.videosTempPromoted, ...this.modelsWithVideoEnabled, ...this.modelsWithAudioOnly, ...this.modelsWithNoPermissions]
+		},
+
```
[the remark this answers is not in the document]
Antreesy: And if return would look like this, maybe we don't need `orderedVideos` and just modify `videos` instead?

src/components/CallView/Grid/Grid.vue
```diff
@@ -478,6 +485,65 @@ export default {
...
+		orderedVideos() {
+			if (!this.participantsInitialised) {
+				return []
+			}
+			// Dynamic ordering is not possible for guests because
+			// participants store is not initialized
+			if (this.actorType === ATTENDEE.ACTOR_TYPE.GUESTS) {
+				return this.videos
+			}
```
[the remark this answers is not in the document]
DorraJaouad: Yeah, it was returning [] for guests. Guest check should come before store check
If we return videos if store is not initialized, then random order will be shown for a few secs for users and then show the order.

src/components/CallView/Grid/Grid.vue
```diff
@@ -828,6 +922,82 @@ export default {
 			return callParticipantModel.attributes.peerId === this.$store.getters.selectedVideoPeerId
 		},
 
+		isModelWithVideo(callParticipantModel) {
+			return callParticipantModel.attributes.videoAvailable
+				&& this.sharedDatas[callParticipantModel.attributes.peerId].remoteVideoBlocker.isVideoEnabled()
```
Antreesy, line 927: That re-sorts the list, if user-receiver disables video of other participant. I think, only ones that disabled by sender should be falsy

src/components/CallView/Grid/Grid.vue
```diff
@@ -828,6 +922,82 @@ export default {
...
+
+		promoteSpeaker(model) {
+			const id = model.attributes.nextcloudSessionId
+
+			// if model is already in the first page, do nothing
+			if (this.orderedVideos.slice(0, this.slots).find(video => video.attributes.nextcloudSessionId === model.attributes.nextcloudSessionId)) {
+				return
+			}
+
+			if (this.modelsWithScreenshare.includes(model)) {
```
Antreesy, line 966:
```suggestion
			if (this.screens.includes(model.attributes.peerId)) {
```

src/components/CallView/Grid/Grid.vue
```diff
@@ -828,6 +928,81 @@ export default {
...
+
+		getOrderedTiles(tilesMap, orderMask) {
+			// Get the ordered tiles
+			const orderedTiles = orderMask
+				.filter(id => tilesMap.has(id))
+				.map(id => tilesMap.get(id))
+
+			// Now the rest of allTiles that are not in the orderedTiles
+			const rest = Array.from(tilesMap.values())
+				.filter(tile => !orderedTiles.includes(tile))
```
Antreesy, line 1002: Let's try to do it in one iteration?
```suggestion
			const { orderedTiles, rest } = orderMask.reduce((acc, id) => {
				if (tilesMap.has(id)) {
					acc.orderedTiles.push(tilesMap.get(id))
				} else {
					acc.rest.push(tilesMap.get(id))
				}
				return acc
			}, { orderedTiles: [], rest: [] })
```

Antreesy: You're right, that wouldn't work here

src/components/CallView/Grid/Grid.vue
```diff
@@ -268,6 +270,10 @@ export default {
 			// Timer for the videos bottom bar
 			showVideoOverlayTimer: null,
 			debounceMakeGrid: () => {},
+			videosTempPromoted: [],
```
Antreesy, line 273: Since it's based on speaking, shall we rename it to:
- `promotedModels`
- `tempPromotedModels`
- e.t.c?

src/components/CallView/Grid/Grid.vue
```diff
@@ -478,6 +484,73 @@ export default {
...
+				...objectMap.modelsWithNoPermissions]
+		},
+
+		speakers() {
+			return this.callParticipantModels.filter(model => model.attributes.speaking)
+		},
+
+		speakersWithAudioOff() {
+			if (this.videosTempPromoted.length === 0) {
+				return
```
Antreesy, line 550:
```suggestion
				return []
```

To be consistent with return type.
Another option: drop early return. `[ ].filter()` will return `[ ]` anyway

src/components/CallView/Grid/Grid.vue
```diff
@@ -523,6 +596,34 @@ export default {
...
+		speakers(speakers) {
+			if (speakers.length === 0) {
+				return
+			}
+			speakers.forEach(speaker => {
+				this.promoteSpeaker(speaker)
+				if (this.unpromoteSpeakerTimer[speaker.attributes.nextcloudSessionId]) {
+					clearTimeout(this.unpromoteSpeakerTimer[speaker.attributes.nextcloudSessionId])
+				}
+			})
```
Antreesy, line 609: `[ ].forEach()` does nothing
`clearTimeout(undefined)` won't throw an error

```suggestion
			speakers.forEach(speaker => {
				this.promoteSpeaker(speaker)
				clearTimeout(this.unpromoteSpeakerTimer[speaker.attributes.nextcloudSessionId])
			})
```

src/components/CallView/Grid/Grid.vue
```diff
@@ -523,6 +596,34 @@ export default {
...
+		speakersWithAudioOff(newModels, oldModels) {
+			if (!newModels || newModels.length === 0) {
+				return
+			}
+
+			const addedModels = !oldModels || oldModels.length === 0 ? newModels : newModels.filter(model => !oldModels.includes(model))
+			if (addedModels.length === 0) {
+				return
+			}
+			addedModels.forEach(speaker => {
```
Antreesy, line 621: And if type is consistent:

```suggestion
			newModels.forEach(speaker => {
				if(!oldModels.includes(speaker)) {
					return
				}
```

src/components/CallView/Grid/Grid.vue
Antreesy: And maybe keep consistent naming? I'm confused between `speaker` and `model`

src/components/CallView/Grid/Grid.vue
```diff
@@ -828,6 +929,88 @@ export default {
...
+			if (!this.videosTempPromoted.includes(model)) {
+				const videoIndex = this.tilesWithVideoMask.indexOf(id)
+				if (videoIndex !== -1) {
+					this.tilesWithVideoMask.splice(videoIndex, 1)
+				}
+				const audioIndex = this.tilesWithAudioMask.indexOf(id)
+				if (audioIndex !== -1) {
+					this.tilesWithAudioMask.splice(audioIndex, 1)
+				}
+
```
Antreesy, line 990: I see nothing wrong, if we at the same time promote model and add it to the mask array. Tile would not be duplicated, since we filter them in `orderedVideos`, but we remain the order of promotion and won't provoke additional reactive values updates.
```suggestion
				if (this.isModelWithVideo(model)) {
					this.tilesWithVideoMask.unshift(id)
				} else {
					this.tilesWithAudioMask.unshift(id)
				}
```

src/components/CallView/Grid/Grid.vue
```diff
@@ -828,6 +929,88 @@ export default {
...
+			if (this.isModelWithVideo(model)) {
+				if (!this.tilesWithVideoMask.includes(id)) {
+					this.tilesWithVideoMask.unshift(id)
+				}
+			} else {
+				if (!this.tilesWithAudioMask.includes(id)) {
+					this.tilesWithAudioMask.unshift(id)
+				}
+			}
```
Antreesy, line 963: if we added a tile in the mask array during promotion, we already know, where to put it with the next re-render, just need to remove from tempPromoted 
```suggestion
```

src/components/CallView/Grid/Grid.vue
```diff
@@ -268,6 +270,10 @@ export default {
 			// Timer for the videos bottom bar
 			showVideoOverlayTimer: null,
 			debounceMakeGrid: () => {},
+			videosTempPromoted: [],
+			unpromoteSpeakerTimer: {},
+			tilesWithVideoMask: [],
+			tilesWithAudioMask: [],
```
Antreesy, line 276: If I was promoted with video enabled, then stopped speaking, and disable video, that would drop me after everyone who was promoted with audio only (and in reverse), without remaining the historical order
My suggestion is to unite both masks into a single queue, `First in - First out`, regardless of whether your video was on when you spoke, or not.

```suggestion
			promotedHistoryMask: [],
```

Shouldn't affect `getOrderedTiles()`, as it has two separate arrays (video / audio) to sort provided

src/components/CallView/Grid/Grid.vue
DorraJaouad: We still need to remove it from its current position if it exists. in mask.

src/components/CallView/Grid/Grid.vue
DorraJaouad: If we add the model to the mask during unpromotion, it will drop you in audios, first place.

src/components/CallView/Grid/Grid.vue
```diff
@@ -979,16 +954,23 @@ export default {
...
+				// add model to the beginning of the orderedVideos in its category
+				if (this.isModelWithVideo(model)) {
+					if (!this.promotedHistoryMask.includes(id)) {
+						this.promotedHistoryMask.unshift(id)
+					}
+				} else {
+					if (!this.promotedHistoryMask.includes(id)) {
+						this.promotedHistoryMask.unshift(id)
+					}
+				}
```
Antreesy, line 973: You wanted to write this, right? =P

from Line 958-960 we already know it's not in the array, or remove it

```suggestion
				this.promotedHistoryMask.unshift(id)
```

src/components/CallView/Grid/Grid.vue
```diff
@@ -828,6 +915,76 @@ export default {
...
+			}
+
+			this.tempPromotedModels.splice(index, 1)
+		},
+
+		promoteSpeaker(model) {
+			const id = model.attributes.nextcloudSessionId
+
+			// if model is already in the first page, do nothing
+			if (this.orderedVideos.slice(0, this.slots).find(video => video.attributes.nextcloudSessionId === model.attributes.nextcloudSessionId)) {
```
Antreesy, line 945:
```suggestion
			if (this.orderedVideos.slice(0, this.slots).find(video => video.attributes.nextcloudSessionId === id)) {
```

Antreesy (approved):
Tested in call, works reliable.
Let's merge it to `main` branch and see later how it works during beta release

DorraJaouad:
Tested and no issues, works as expected.

PR #12368, opened 2024-05-16
fix(vue3): use value from v-model (instead of event.target.value)

Antreesy:
### ☑️ Resolves

* Ref #12366
  * To test - create a new conversation with description and passport via dialog window

### 🏁 Checklist

### DIFF
--- a/src/components/NewConversationDialog/NewConversationSetupPage.vue
+++ b/src/components/NewConversationDialog/NewConversationSetupPage.vue
@@ -113,8 +113,8 @@ export default {
 			get() {
 				return this.newConversation.displayName
 			},
-			set(event) {
-				this.updateNewConversation({ displayName: event.target.value })
+			set(value) {
+				this.updateNewConversation({ displayName: value })
 			},
 		},
 
@@ -122,8 +122,8 @@ export default {
 			get() {
 				return this.newConversation.description
 			},
-			set(event) {
-				this.updateNewConversation({ description: event.target.value })
+			set(value) {
+				this.updateNewConversation({ description: value })
 			},
 		},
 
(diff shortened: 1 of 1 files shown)


src/App.vue
```diff
@@ -13,7 +13,7 @@
 			<router-view />
 		</NcAppContent>
 		<RightSidebar :is-in-call="isInCall" />
-		<MediaSettings :recording-consent-given.sync="recordingConsentGiven" />
+		<MediaSettings v-model.recording-consent-given="recordingConsentGiven" />
```
ShGKme, line 16: Incorrect syntax

src/components/BreakoutRoomsEditor/SelectableParticipant.vue
```diff
@@ -44,21 +44,21 @@ export default {
 			required: true,
 		},
 
-		checked: {
+		modelValue: {
```
ShGKme, line 47: It is not a part of a breaking change 👀

src/components/TopBar/TopBar.vue
```diff
@@ -4,7 +4,7 @@
 -->
 
 <template>
-	<div class="top-bar" :style="topBarStyle" :data-theme-dark="isInCall">
+	<div class="top-bar" :style="topBarStyle" :data-theme-dark="isInCall ? true : undefined">
```
ShGKme, line 7: And this is far from `.sync` modifier changes 👀

Antreesy: :shrug:

ShGKme:
Converting `:prop.sync` syntax to `v-model:prop` is covered automatically by ESLint in https://github.com/nextcloud/spreed/pull/12378

So we can use this PR to fix other stuff like theming.

Antreesy:
Rebased, only fixes are kept

ShGKme (approved):

PR #12378, opened 2024-05-16
chore(vue3): eslint

ShGKme:
### ☑️ Resolves

* Ref: https://github.com/nextcloud/spreed/issues/9448
* Adjusted the config
* Run fixable rules one by one
* Only removed `$listeners` left
* Also covers:
  * https://github.com/nextcloud/spreed/pull/12376
  * https://github.com/nextcloud/spreed/pull/12368

PR #12380, opened 2024-05-16
fix(vue3): temporary pass class as a prop to MessageButtonsBar

ShGKme:
### ☑️ Resolves

* Chained on: https://github.com/nextcloud/spreed/pull/12378
* In Vue 3 with `inheritAttrs: false` classes are also not inherited as attributes
* Temporary added a prop for it, but it is better to fix it on `main` and to not provide message object as a banch of properties with `v-bind="message"`

🏚️ Before | 🏡 After
-- | --
image | image

### 🏁 Checklist

---

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist


src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
```diff
@@ -5,7 +5,7 @@
 
 <template>
 	<!-- Message Actions -->
-	<div v-click-outside="handleClickOutside">
+	<div v-click-outside="handleClickOutside" :class="$props['class']">
```
DorraJaouad, line 8: can't be just `:class="class"` ?

DorraJaouad (approved):
I also agree to sift v-bind object to only what's declared as `props`, so we can omit using `inheritAttrs: false` 

Not needed and they are passed as `$attrs`
```
isCollapsedSystemMessage
isCombinedSystemMessage
isCombinedSystemMessageCollapsed
isTemporary
lastCollapsedMessageId
markdown
nextMessageId
parent
reactions
reactionsSelf
referenceId
sendingFailure
systemMessage
```

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
ShGKme: No, because `class` is reserved work in JavaScript, e.g., you cannot create a variable called `class` or `function`. But you can create an object property with any name.

Antreesy:
Superseded by #12395

PR #12381, opened 2024-05-16
chore(vue3): migrate async component registration and remove $children

ShGKme:
### ☑️ Resolves

* Ref: https://github.com/nextcloud/spreed/issues/12366
* [Async components now require defineAsyncComponent method to be created](https://v3-migration.vuejs.org/breaking-changes/async-components.html)
* [$children instance property](https://v3-migration.vuejs.org/breaking-changes/children.html)
* Note: sidebar is still half broken but can be fixed separately on `main`. This PR only fixes breaking changes.
  * Styles has been lost
  * Reactivity is lost

🏚️ Before | 🏡 After
-- | --
image | image

### DIFF
--- a/src/FilesSidebarCallViewApp.vue
+++ b/src/FilesSidebarCallViewApp.vue
@@ -15,6 +15,8 @@
 </template>
 
 <script>
+import { defineComponent, defineAsyncComponent, h } from 'vue'
+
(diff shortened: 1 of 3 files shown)


DorraJaouad (approved):
Tested

Issue #10358, opened 2023-08-25
Improve talk mobile, narrow screens and sidebar experience.


marcoambrosini:
- [ ] Collapse the start call button into an icon-only button
- [ ] Create a row for avatar, name, and date above a `messageGroup` and make text full-width

<img width="318" alt="Screenshot 2023-08-25 at 16 45 35" src="https://github.com/nextcloud/spreed/assets/26852655/8c550b08-f96b-4543-9249-15822f9bb694">

cc @nextcloud/designers @nickvergessen this is a quick big win

nickvergessen:
There are many things wrong on mobile atm, but it's not on the roadmap as we have mobile clients

Ref 
* https://github.com/nextcloud/spreed/issues/10091
* https://github.com/nextcloud/spreed/issues/9936
* https://github.com/nextcloud/spreed/issues/6969

marcoambrosini:
@nickvergessen This is also about the desktop experience, especially in the sidebar during calls. Or on small laptops with the two sidebars open.

Wireframe of the idea above:
<img width="400" alt="Screenshot 2023-08-25 at 16 56 27" src="https://github.com/nextcloud/spreed/assets/26852655/581d42cd-fc64-4d15-b0f7-7fd09f5c9136">

Antreesy:
It would work nice IMO, if we decide to proceed with showing a timestamp and checkmark for the whole message group instead of each message (messages are combined in group if they were posted within 5-minutes interval, and it's still available for each message in buttons bar)

Antreesy:
For bringing this up a bit:
* could #11212 be an option?

PR #12383, opened 2024-05-17
feat(mobile): use button without text if on mobile

szaimen:
For https://github.com/nextcloud/spreed/issues/10358

| Before | After |
|---|---|
| image | image |

<details>
<summary>For my own testing</summary>

```
docker run -it --rm ^
--name nextcloud-easy-test1 ^
-p 8443:443 ^
-e TALK_BRANCH=enh/noid/shrink-button ^
--volume="nextcloud_easy_test_npm_cache_volume:/var/www/.npm" ^
ghcr.io/szaimen/nextcloud-easy-test:latest

```

</details>

### DIFF
--- a/src/components/TopBar/CallButton.vue
+++ b/src/components/TopBar/CallButton.vue
@@ -14,6 +14,7 @@
 				autoHide: false,
 				html: true
 			}"
+			:aria-label="startCallLabel"
 			:disabled="startCallButtonDisabled || loading"
 			:type="startCallButtonType"
 			@click="handleClick">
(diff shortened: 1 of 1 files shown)


src/components/TopBar/CallButton.vue
```diff
@@ -14,6 +14,7 @@
 				autoHide: false,
 				html: true
 			}"
+			:aria-label="isMobile ? startCallLabel : undefined"
```
Antreesy, line 17:
```suggestion
			:aria-label="startCallLabel"
```

a11y team passed by and said "Both visible label and aria-label could be used together, if they have the save content"

Antreesy (commented):
Rest is good, IMO. Let's wait for other's feedback

src/components/TopBar/CallButton.vue
szaimen: done

src/components/TopBar/CallButton.vue
```diff
@@ -22,31 +23,40 @@
 				<VideoOutlineIcon v-else-if="silentCall" :size="20" />
 				<VideoIcon v-else :size="20" />
 			</template>
-			{{ startCallLabel }}
+			<template v-if="!isMobile">
+				{{ startCallLabel }}
+			</template>
```
ShGKme, line 28: Probably this should be passed as a conditional scoped slot instead of always having a slot with conditional content to avoid some rendering issues in `NcActions`. But I need to re-check.
```suggestion
			<template v-if="!isMobile" #default>
				{{ startCallLabel }}
			</template>
```

szaimen: done

nimishavijay (approved):
Nice!

marcoambrosini (approved):

szaimen:
Merge?

DorraJaouad (approved):
LGTM

PR #12386, opened 2024-05-17
fix(OpenAPI): Add missing reactionsSelf to ChatMessage

provokateurin:
Psalm probably didn't catch the missing field because the logic for setting it is too complicated and it can't figure out what is happening.
https://github.com/nextcloud/spreed/blob/704a0727ddc1c798a1f2ba8120cb17316a9995c6/lib/Controller/ChatController.php#L680-L694

### 🏁 Checklist


Antreesy (approved):
Makes sense, looking at TS output

provokateurin:
@Antreesy do you know why typescript wasn't complaining that a non-existent property of an object was used?

Issue #11454, opened 2024-01-26
Can't select message text box


miaulalala:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

When using FF Mobile 122.0 (Build #2015998391), I cannot select the textbox yo type a new message.

## Steps to reproduce
1. Open any chat 
2. Switch to another chat
3. Try selecting the message text box to type

### Expected behaviour
Tell us what should happen

The keyboard should open, the cursor should appear and I should be able to send a message

### Actual behaviour
Tell us what happens instead

No keyboard or cursor

## Talk app

**Talk app version:** (see apps admin page: `/index.php/settings/apps`) c.nc

**Custom Signaling server configured:** yes/no and version (see additional admin settings: `/index.php/index.php/settings/admin/talk#signaling_server`) c.nc

**Custom TURN server configured:** yes/no (see additional admin settings: `/index.php/settings/admin/talk#turn_server`) c.nc

**Custom STUN server configured:** yes/no (see additional admin settings: `/index.php/settings/admin/talk#stun_server`) c.nc

## Browser

**Microphone available:** yes

**Camera available:** yes/no

**Operating system:** Android

**Browser name:** Firefox Mobile 122.0 (Build #2015998391)

**Browser version:** 85/96/...

### Browser log

<details>
```
Insert your browser log here, this could for example include:
a) The javascript console log
b) The network log
c) ...
```

</details>

## Server configuration

**Operating system**: Ubuntu/RedHat/...

**Web server:** Apache/Nginx

**Database:** MySQL/Maria/SQLite/PostgreSQL

**PHP version:** 8.0/8.1/8.2

**Nextcloud Version:** (see admin page)

**List of activated apps:**

<details>

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your server installation folder
```
</details>

**Nextcloud configuration:**

<details>

```
If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your Nextcloud installation folder
```
</details>

### Server log (data/nextcloud.log)
<details>

```
Insert your server log here
```
</details>

Antreesy:
[...]

PR #12387, opened 2024-05-17
fix(sidebar): use open state to remove always working focus trap on mobile

ShGKme:
### ☑️ Resolves

* Fix: https://github.com/nextcloud/spreed/issues/11454
* Bump `@nextcloud/vue` from 8.11.3 to 8.13.0
* Using new `open` state of `NcAppSidebar` with built-in toggle button
* Adjusted animations

🏚️ Before | 🏡 After
-- | --
image | image

### 🚧 Tasks

- [ ] Style changes

### 🏁 Checklist

### DIFF
--- a/src/components/RightSidebar/RightSidebar.vue
+++ b/src/components/RightSidebar/RightSidebar.vue
@@ -4,13 +4,27 @@
 -->
 
 <template>
-	<NcAppSidebar v-show="opened"
+	<NcAppSidebar v-if="token"
+		:open="opened"
 		:name="conversation.displayName"
 		:title="conversation.displayName"
 		:active.sync="activeTab"
 		:class="'active-tab-' + activeTab"
+		:toggle-classes="{ 'chat-button-sidebar-toggle': isInCall }"
+		:toggle-attrs="isInCall ? inCallToggleAttrs : undefined"
+		@update:open="handleUpdateOpen"
+		@update:active="handleUpdateActive"
 		@closed="handleClosed"
 		@close="handleClose">
+		<!-- Use a custom icon when sidebar is used for chat messages during the call -->
+		<template v-if="isInCall" #toggle-icon>
+			<MessageText :size="20" />
+			<NcCounterBubble v-if="unreadMessagesCounter > 0"
+				class="chat-button__unread-messages-counter"
+				:type="hasUnreadMentions ? 'highlighted' : 'outlined'">
+				{{ unreadMessagesCounter }}
+			</NcCounterBubble>
+		</template>
 		<template #description>
 			<LobbyStatus v-if="canFullModerate && hasLobbyEnabled" :token="token" />
 		</template>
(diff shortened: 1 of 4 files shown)


Antreesy (approved):
Works as expected. Though there are still too many styles changes came from library (follow-up?)

ShGKme:
> Though there are still too many styles changes came from library (follow-up?)

What do you mean? Are they from the library or new server styles?

ShGKme:
`@nextcloud/vue@8` doesn't suppose to have any breaking change at least for supported 28 and 29.

Styles on 30 are broken in many places also from server styles, it's in progress

Antreesy:
Yeah, server styles, I have compared at `main` only. Tested with desktop + server 29, so far looks the same

ShGKme:
- Rebased onto `main`
- Minor fixes in fixup:
  - Correctly use conditional slot
  - Pass all the previous attrs to the button like title and aria-label.

src/components/RightSidebar/RightSidebar.vue
```diff
@@ -156,7 +175,7 @@ export default {
 			return this.$store.getters.getSidebarStatus
 		},
 		opened() {
-			return !!this.token && !this.isInLobby && this.show
+			return !this.isInLobby && this.show
```
Antreesy, line 178: Makes sense because of `v-if="token"`

src/components/RightSidebar/RightSidebar.vue
```diff
@@ -4,14 +4,27 @@
...
+		:toggle-attrs="isInCall ? inCallToggleAttrs : undefined"
+		@update:open="handleUpdateOpen"
 		@update:active="handleUpdateActive"
 		@closed="handleClosed"
 		@close="handleClose">
+		<!-- Use a custom icon when sidebar is used for chat messages during the call -->
+		<template v-if="isInCall" #toggle-icon>
+			<MessageText :size="20" />
+			<NcCounterBubble v-if="unreadMessagesCounter > 0"
+				class="chat-button__unread-messages-counter"
```
Antreesy, line 23: image

we don't need to bury class under :deep selector, I think

ShGKme: For `chat-button__unread-messages-counter` deep is not needed, but for the `chat-button` itself it is needed.

Let me check with 29 styles...

ShGKme: Working on a regression

ShGKme:
- Rebased onto `main` and resolved conflicts
- Fixed styles. Required, because with new Teleport fix in `NcAppSidebar`, the toggle is no longer inside the sidebar

image

src/components/RightSidebar/RightSidebar.vue
ShGKme: Thanks for noticing the problem

ShGKme:
Note, screenshot is made on stable29 styles

Antreesy (approved):
Tested against stable29

on main ubread marker is still messy, but because of reducing the clickable area 44 -> 34

DorraJaouad:
On main :
message (shouldn't be lower ? It doesn't look much readable)
image
mention (at all) 
image
direct mention 
image

Color should be applied only on message notif
```
.chat-button__unread-messages-counter {
	color: var(--color-primary-element);
}
```
Besides, styles should be aligned with conversation counter styles ?

ShGKme:
> On main

@DorraJaouad Is it much different from what was before? Styles on main, including `@nextcloud/vue` are not adjusted to new theming.

src/components/RightSidebar/RightSidebar.vue
```diff
@@ -391,4 +484,23 @@ export default {
 	height: 100%;
 }
 
+.chat-button__unread-messages-counter {
+	position: absolute;
+	bottom: 2px;
+	right: 2px;
+	pointer-events: none;
+	color: var(--color-primary-element);
```
Antreesy, line 492: On stable29 | With a patch
-- | --
image | image

```suggestion

	&.counter-bubble__counter--highlighted {
		color: var(--color-primary-text);
	}
```

Antreesy: This hasn't changed since Feb.2023 https://github.com/nextcloud/spreed/commit/d3a062e25ea770dc62df4957beb58f487fe3fb98, and not related to this PR changes, so could be fixed in follow-up

Antreesy:
Rebased on `main`, squashed fixups and provide a fix for 'highlighted' type:
image

DorraJaouad (approved):
Tested

PR #12395, opened 2024-05-20
fix(MessageButtonsBar): pass props only

DorraJaouad:
### ☑️ Resolves

* Related to #12380 and to avoid passing class as a prop in vue 3 

Tested with Vue3:
image

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -4,9 +4,9 @@
 -->
 
 <template>
-	<li :id="`message_${id}`"
+	<li :id="`message_${message.id}`"
 		ref="message"
-		:data-message-id="id"
+		:data-message-id="message.id"
 		:data-seen="seen"
 		:data-next-message-id="nextMessageId"
 		:data-previous-message-id="previousMessageId"
@@ -23,12 +23,13 @@
 			<MessageBody :rich-parameters="richParameters"
 				:is-deleting="isDeleting"
 				:has-call="conversation.hasCall"
-				v-bind="{...$props, ...readInfoProps}" />
+				:message="message"
+				v-bind="readInfoProps" />
 
 			<!-- reactions buttons and popover with details -->
-			<Reactions v-if="Object.keys(reactions).length"
-				:id="id"
-				:token="token"
+			<Reactions v-if="Object.keys(message.reactions).length"
+				:id="message.id"
+				:token="message.token"
 				:can-react="canReact"
 				:show-controls="isHovered || isFollowUpEmojiPickerOpen"
 				@emoji-picker-toggled="toggleFollowUpEmojiPicker" />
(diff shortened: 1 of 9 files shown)


src/components/MessagesList/MessagesGroup/Message/Message.vue
```diff
@@ -440,6 +440,23 @@ export default {
...
+				timestamp: this.timestamp,
+				id: this.id,
+				isReplyable: this.isReplyable,
+				token: this.token,
+				messageType: this.messageType,
+				previousMessageId: this.previousMessageId,
+				lastEditActorDisplayName: this.lastEditActorDisplayName,
+				lastEditTimestamp: this.lastEditTimestamp,
+			}
+		},
```
ShGKme, line 458: If we already consider it an atomic object, then it makes sense to pass it as a single prop.

DorraJaouad: Not really, there is `isReplyable` and it doesn't cover the whole message object that we commonly use.

Antreesy: > there is `isReplyable`

Which is a part of message object. The only thing missing from it is `previousMessageId`, and its usage looks minor, so I'd pass it separately and think of a way to not use it in the component at all (by modifying store action?)

ShGKme: If we consider some specific props to not be a part of the `message` object we can pass them separately, keeping all the other properties a part of the `message` object.

So, if we want to refactor it, let's in both `Message` and `MessageButtonsBar` add a prop for `message` object with all the properties from the original `message` object (from store = API) and keep only additional properties as separated props.

Antreesy: Refactored

DorraJaouad:
Changes lGTM 👍🏻

Antreesy (approved):
Approving on behalf of Dorra :eagle: . Tested:
- readInfo properties on children
- callback actions (Deck app)
- polls, replies, markdown, failures, captions, system messages, edited / deleted messages. rich params, buttonbar actions

src/components/MessagesList/MessagesGroup/Message/MessagePart/MessageBody.vue
```diff
@@ -156,101 +156,36 @@ export default {
...
 	},
 
 	setup(props) {
-		const { token, id } = toRefs(props)
+		const { message } = toRefs(props)
 		const {
 			isEditable,
 			isFileShare,
-		} = useMessageInfo(token, id)
+		} = useMessageInfo(message.value.token, message.value.id)
```
ShGKme, line 188: This is reactivity loosing

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
```diff
@@ -477,7 +397,7 @@ export default {
 			isFileShareWithoutCaption,
 			isConversationReadOnly,
 			isConversationModifiable,
-		} = useMessageInfo(token, id)
+		} = useMessageInfo(message.value.token, message.value.id)
```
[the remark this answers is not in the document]
ShGKme: Still reactivity loosing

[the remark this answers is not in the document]
Antreesy: It wasn't reactive initially.. 
But now we could passed a whole message object and make token and I'd truly reactive

PR #12432, opened 2024-05-28
fix(attachments): allow null expireDate in createShare

ShGKme:
### ☑️ Resolves

* Fix regression from: https://github.com/nextcloud/server/pull/44485
* `$expireDate` is now `null` by default and can be not only `string`
* Or is it better to fix on server side?

### 🚧 Tasks

- [x] Make `expireDate` nullable and check for null on setting date

### 🏁 Checklist

### DIFF
--- a/lib/Share/Helper/ShareAPIController.php
+++ b/lib/Share/Helper/ShareAPIController.php
@@ -78,14 +78,14 @@ class ShareAPIController {
 	 * @param IShare $share
 	 * @param string $shareWith
 	 * @param int $permissions
-	 * @param string $expireDate
+	 * @param ?string $expireDate
 	 * @throws OCSNotFoundException
 	 */
-	public function createShare(IShare $share, string $shareWith, int $permissions, string $expireDate): void {
+	public function createShare(IShare $share, string $shareWith, int $permissions, ?string $expireDate = null): void {
 		$share->setSharedWith($shareWith);
 		$share->setPermissions($permissions);
 
-		if ($expireDate !== '') {
+		if ($expireDate !== null && $expireDate !== '') {
 			try {
 				$expireDateTime = $this->parseDate($expireDate);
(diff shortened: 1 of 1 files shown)


SystemKeeper:
> * Or is it better to fix on server side?

If I understand the changes correctly, the public interface always allowed it to be null?

https://github.com/nextcloud/server/pull/44485/files#diff-10411333a1207c164cd88b80b0df9d1987227d1bfb4ae215454b34fc28eb48aa

Which I would then say needs to be fixed in talk?

lib/Share/Helper/ShareAPIController.php
```diff
@@ -78,14 +78,14 @@ public function formatShare(IShare $share): array {
...
+	 * @param ?string $expireDate
 	 * @throws OCSNotFoundException
 	 */
-	public function createShare(IShare $share, string $shareWith, int $permissions, string $expireDate): void {
+	public function createShare(IShare $share, string $shareWith, int $permissions, ?string $expireDate = null): void {
 		$share->setSharedWith($shareWith);
 		$share->setPermissions($permissions);
 
-		if ($expireDate !== '') {
+		if ($expireDate !== null && $expireDate !== '') {
```
Fenn-CS, line 88: `empty` returns true for both `null` and empty string (`''`).

```suggestion
		if (!empty($expireDate)) {
```

Fenn-CS (commented):
Looks ok so far.

lib/Share/Helper/ShareAPIController.php
ShGKme: Yes, but I wanted to keep it separated to explicitly show 2 different cases, because `''` and `null` have different semantics here

ShGKme:
- Closing in flavor of https://github.com/nextcloud/server/pull/45562

PR #12460, opened 2024-06-05
fix(chat): refactor Quote component, fix reactivity

Antreesy:
### ☑️ Resolves

* Follow-up https://github.com/nextcloud/spreed/pull/12395#discussion_r1618771274

### 🏁 Checklist

### DIFF
--- a/src/composables/useMessageInfo.js
+++ b/src/composables/useMessageInfo.js
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-import { computed, unref, ref } from 'vue'
+import { computed, ref } from 'vue'
 
 import { getCapabilities } from '@nextcloud/capabilities'
 import moment from '@nextcloud/moment'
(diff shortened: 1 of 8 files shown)


src/composables/useMessageInfo.js
```diff
@@ -17,19 +17,13 @@ const canEditMessage = getCapabilities()?.spreed?.features?.includes('edit-messa
...
- * @param {import('vue').Ref} tokenRef conversation token
- * @param {import('vue').Ref} messageIdRef message id to edit
- * @return {import('vue').ComputedRef<boolean>}
+ * @param {import('vue').Ref} message message object
+ * @return {Record<string, import('vue').ComputedRef<boolean>>}
  */
-export function useMessageInfo(tokenRef = ref(null), messageIdRef = ref(null)) {
-	const token = unref(tokenRef)
-	const messageId = unref(messageIdRef)
+export function useMessageInfo(message = ref({})) {
```
ShGKme, line 23: I don't think, the default value, especially an empty object which is not a special case for a message object, makes sense here.

`if (...!message.value)` also doesn’t work for `useMessageInfo()` with `{}` value.

ShGKme: > Maybe, check for any key present in object?

We can check for `id` property only.

But in a follow-up (not right now) I'd refactor it and replace this `{}` value with explicit `null`.

ShGKme (approved):
Thanks!

DorraJaouad (approved):
tested

PR #12461, opened 2024-06-05
[vue3] fix: fix unmount issues and call reactivity

Antreesy:
### ☑️ Resolves

* Reactivity doesn't work correctly for callParticipantModels, this leads to missed updates / re-renders, which worked in Vue2

### 🚧 Tasks

- [x] replace ref with reactive
  - :warning:  doesn't work deeply (or works, but flacky), if we push models later in the reactive array, and then update some properties null -> object (stream, peer, speaking)
- [x] Minor TypeErrors in console.log
- [x] Participant join with video
- [ ] Participant join with screenshare (don't see)
- [ ] Participant and speaking indicator (don't see)
- [ ] Participant stops screenshare (main peer video is missing for several seconds)

### 🏁 Checklist

### DIFF
--- a/src/components/CallView/CallView.vue
+++ b/src/components/CallView/CallView.vue
@@ -376,9 +376,16 @@ export default {
 		},
 
 		screens() {
+			console.log('screens watcher')
 			this._setScreenVisible()
 
 		},
+		// screens: {
+		// 	deep: true,
+		// 	handler() {
+		// 		this._setScreenVisible()
+		// 	}
+		// },
 
 		callParticipantModelsWithScreen(newValue, previousValue) {
 			// Everytime a new screen is shared, switch to promoted view
@@ -455,6 +462,7 @@ export default {
 		 * @param {Array} models the array of CallParticipantModels
 		 */
 		updateDataFromCallParticipantModels(models) {
+			console.log('updateDataFromCallParticipantModels', models)
 			const addedModels = models.filter(model => !this.sharedDatas[model.attributes.peerId])
 			const removedModelIds = Object.keys(this.sharedDatas).filter(sharedDataId => models.find(model => model.attributes.peerId === sharedDataId) === undefined)
 
@@ -464,15 +472,15 @@ export default {
 				delete this.sharedDatas[removedModelId]
 
 				this.speakingUnwatchers[removedModelId]()
-				// Not reactive, but not a problem
+				// FIXME Not reactive, but not a problem?
 				delete this.speakingUnwatchers[removedModelId]
 
 				this.screenUnwatchers[removedModelId]()
-				// Not reactive, but not a problem
+				// FIXME Not reactive, but not a problem?
 				delete this.screenUnwatchers[removedModelId]
 
 				this.raisedHandUnwatchers[removedModelId]()
-				// Not reactive, but not a problem
+				// FIXME Not reactive, but not a problem?
 				delete this.raisedHandUnwatchers[removedModelId]
 
 				const index = this.speakers.findIndex(speaker => speaker.id === removedModelId)
@@ -490,7 +498,7 @@ export default {
 
 				this.sharedDatas[addedModel.attributes.peerId] = sharedData
 
-				// Not reactive, but not a problem
+				// FIXME Not reactive, but not a problem?
 				this.speakingUnwatchers[addedModel.attributes.peerId] = this.$watch(function() {
 					return addedModel.attributes.speaking
 				}, function(speaking) {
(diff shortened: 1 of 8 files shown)


src/utils/webrtc/models/CallParticipantCollection.js
```diff
@@ -15,36 +15,35 @@ export default function CallParticipantCollection() {
 
 	this._superEmitterMixin()
 
-	// FIXME: use reactive instead of ref after migration to vue 3
-	this.callParticipantModels = ref([])
+	this.callParticipantModels = reactive([])
```
ShGKme, line 18: Did it cause a problem anywhere or just refactoring?

src/utils/webrtc/models/CallParticipantModel.js
```diff
@@ -121,8 +122,16 @@ CallParticipantModel.prototype = {
 		this._trigger('change:' + key, [value])
 	},
 
+	// Compare Object and Proxy(Object), whether they refer to the same object
+	_isSameProxy(obj1, obj2) {
+		if (typeof obj1 !== 'object' || obj1 === null || typeof obj2 !== 'object' || obj2 === null) {
+			return obj1 === obj2
+		}
+		return Reflect.getPrototypeOf(obj1) === Reflect.getPrototypeOf(obj2)
+	},
```
ShGKme, line 131: Vue does not create a proxy twice for the same object...
Also, Vue has `toRaw` function to get the original object from Proxy.

Doesn't `return toRaw(obj1) === toRaw(obj2)` work here?

src/utils/webrtc/models/CallParticipantModel.js
```diff
@@ -27,7 +28,7 @@ export default function CallParticipantModel(options) {
 
 	this._superEmitterMixin()
 
-	this.attributes = {
+	this.attributes = reactive({
```
ShGKme, line 31: This is what we missed...

src/utils/webrtc/models/CallParticipantCollection.js
Antreesy: I thought it could be an initial problem, but as we're touching it anyway, just refactored

ShGKme: > Also `reactive` works for nested objects, and `ref` doesn't, right?

No, it does. `ref` creates a fully reactive variable, and for objects, it returns value as `reactive(_value)`. In general, there is nothing `ref` cannot do what we can with `reactive` and `reactive` is only better if we want to avoid writhing `.value`.

src/components/CallView/CallView.vue
```diff
@@ -360,9 +360,12 @@ export default {
...
 
-		speakers(value) {
-			if (value) {
-				this._setPromotedParticipant()
+		speakers: {
+			deep: true,
+			handler(value) {
+				if (value) {
+					this._setPromotedParticipant()
+				}
```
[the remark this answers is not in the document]
ShGKme: In Vue 3 watch on an array works without `deep` only for re-assigning the property, like `this.speakers = newValue`. All the array mutations, by `[index]` or using methods like `push()` won't trigger array watch without `deep`.

So, in Vue 3 watch for arrays works exactly the same as for other kinds of objects:
- without `deep` for immutable change only = re-assignment value
- with `deep` for mutations

In Vue 3 array mutation with methods worked without `deep`.

src/utils/webrtc/models/CallParticipantModel.js
ShGKme: Oh, I missed that we store Peer in reactive data. Probably we should store it as a non-reactive, raw value using `toRaw` to not have proxy created for Peer at all.

ShGKme:
> replace ref with reactive
> * ⚠️  doesn't work deeply (or works, but flacky), if we push models later in the reactive array, and then update some properties null -> object (stream, peer, speaking)

I don't understand this change. Object in the `ref` is the same `reactive` (`ref` wraps object value to `reactive`)

ShGKme:
Changing `ref` to `reactive` allows to work without `.value` (could fix a case where `.value` was missing).

But it also makes it impossible to assign a new object/array.

Antreesy:
> doesn't work deeply

I don't remember writing it anymore. Probably issue was with the model.attributes

> But it also makes it impossible to assign a new object/array.

But that's not the case for us, since we're creating colleaction once? `const callParticipantCollection = new CallParticipantCollection()`

DorraJaouad:
Tested: no issues so far

src/components/CallView/CallView.vue
```diff
@@ -429,10 +429,11 @@ export default {
...
 		},
 
-		speakers(value) {
-			if (value) {
+		speakers: {
+			deep: true,
+			handler() {
 				this._setPromotedParticipant()
-			}
+			},
```
ShGKme, line 436: Wasn't it the bug?

ShGKme (approved):

DorraJaouad:
Maybe also move `mediaDevicesManager` reactive object to the module instead of `useDevices` composable? so other usage will also be reactive

Antreesy:
> Maybe also move `mediaDevicesManager` reactive object to the module instead of `useDevices` composable? so other usage will also be reactive

It looks more complicated than it was for callParticipantsModels. But feel free to follow-up)

PR #12465, opened 2024-06-06
[stable29] fix(MessagesList): keep token consistent during async methods

Antreesy:
### :warning: Issue

* fetching old messages (starting from "conversation_created") happens then param is missing - request is `api/v1/chat/{token}?setReadMarker=0&lookIntoFuture=1&limit=100&includeLastKnown=0&markNotificationsAsRead=0` instead of `api/v1/chat/{token}?setReadMarker=0&lookIntoFuture=1&lastKnownMessageId={id}&limit=100&includeLastKnown=0&markNotificationsAsRead=0`

### ☑️ Resolves

Component methods were using the actual token.
- If catch a moment during a conversation switch (while request for messageContext is pending), context for conv1 will be cancelled, but lookForNewMessages for conv2 will be triggered (before preconditions for conv1), resulting not known and missing lastKnownMessageId parameter. 
- With explicitly passing a token through a methods, we ensure it won't leak while switch
-  If request was cancelled. restore initial state for the component and store

image

### 🚧 Tasks

- [ ]  Manual test with long request delay

### 🏁 Checklist

### DIFF
--- a/src/store/messagesStore.js
+++ b/src/store/messagesStore.js
@@ -1121,6 +1121,12 @@ const actions = {
 	async lookForNewMessages(context, { token, lastKnownMessageId, requestId, requestOptions }) {
 		context.dispatch('cancelLookForNewMessages', { requestId })
 
+		if (!lastKnownMessageId) {
+			// if param is null | undefined, it won't be included in the request query
+			console.warn('Trying to load messages without the required parameter')
+			return
+		}
+
 		// Get a new cancelable request function and cancel function pair
 		const { request, cancel } = CancelableRequest(lookForNewMessages)
 
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -661,7 +661,14 @@ export default {
 					this.$store.dispatch('setLastKnownMessageId', { token, id: startingMessageId })
 
 					// Get chat messages before last read message and after it
-					await this.getMessageContext(token, startingMessageId)
+					try {
+						await this.getMessageContext(token, startingMessageId)
+					} catch (exception) {
+						// Request was cancelled, stop getting preconditions and restore initial state
+						this.$store.dispatch('setFirstKnownMessageId', { token, id: null })
+						this.$store.dispatch('setLastKnownMessageId', { token, id: null })
+						return
+					}
 				}
 
 				this.$nextTick(() => {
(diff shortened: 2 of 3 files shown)


src/store/messagesStore.js
```diff
@@ -1112,6 +1112,21 @@ const actions = {
...
+				token,
+				lastKnownMessageId,
+				messagesList: context.getters.messagesList(token),
+				firstKnown: context.getters.getFirstKnownMessageId(token),
+				lastKnown: context.getters.getLastKnownMessageId(token),
+				realFirstKnown: context.getters.messagesList(token)?.at(0)?.id,
+				realLastKnown: context.getters.messagesList(token)?.at(-1)?.id,
+				lastReadMessageId: context.getters.conversation(token)?.lastReadMessage,
+			})
+		}
```
ShGKme, line 1128: Better to log copy or serialized value. Otherwise there is a chance that when you check the log, you have a new value already by the same object reference.

Antreesy: That might affect only messagesList as array; plain variables should be persistent here (and I want to check them, actually)

Antreesy:
Occurrences of ''dispatch('setLastKnownMessageid''' in Project:
- MessagesList.vue  (2 usages found)
  - 646 `this.$store.dispatch('setLastKnownMessageId', { token: this.token, id: startingMessageId })` - refactored on main, but logic is unchanged; id from hash or lastReadMessage, then load context
  - 1169 `this.$store.dispatch('setLastKnownMessageId', {` - same but for old messages; id from ahsh then load context
- messagesStore.js  (3 usages found)
  - 951 `context.dispatch('setLastKnownMessageId', {` - for old messages (lookIntoFuture = 0), updates only with truthy value, but could be string? :warning:
  - 1040 `context.dispatch('setLastKnownMessageId', {` - for load context: updates if wasn't defined before OR if new one is bigger (could be issue with numbers as strings again) - at least context id will be saved :warning: 
  - 1182 `context.dispatch('setLastKnownMessageId', {` - get value from header `parseInt(response.headers['x-chat-last-given'], 10)` - might fail here? :warning:

ShGKme (approved):

Antreesy:
For testing step by step:
- First commit: for pure test - if switch conversations too fast, `context` will be cancelled for conv1, `lookForNewMessages` will be triggered with token for conv2 - we don't know anything yet and see an error
- Second commit: same token for all requests, no error anymore, but conv1 will be marked as loaded with what comes from `lookForNewMessages`
- Third commit - if context request was cancelled, restore state for conv1, to load context again next time

Could be easier to test with a desktop client against c.nc or another real server (to have more chances to reproduce), or mock a delay in responce in another way

ShGKme:
> Could be easier to test with a desktop client against c.nc or another real server (to have more chances to reproduce)

What is the difference?

Antreesy:
> What is the difference?

Since we know the issue, doesn't matter really. Just need to catch the moment when `getContext` request is pending

Antreesy:
Rebased and aligned with main branch. As code changes are identical, considering it approved :eyes:

PR #12467, opened 2024-06-07
fix(viewer): fix creation of file from template

Antreesy:
### ☑️ Resolves

- pass permissions to Viewer as a string instead of Bitmask
- add visual indicator of loading

🏚️ Before | 🏡 After
-- | --
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/NewMessage/NewMessageNewFileDialog.vue
+++ b/src/components/NewMessage/NewMessageNewFileDialog.vue
@@ -40,7 +40,11 @@
 
 				<div class="new-text-file__buttons">
 					<NcButton type="primary"
+						:disabled="loading"
 						@click="handleCreateNewFile">
+						<template v-if="loading" #icon>
+							<NcLoadingIcon />
+						</template>
 						{{ t('spreed', 'Create file') }}
 					</NcButton>
 				</div>
@@ -53,6 +57,7 @@
 import { showError } from '@nextcloud/dialogs'
 
 import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
+import NcLoadingIcon from '@nextcloud/vue/dist/Components/NcLoadingIcon.js'
 import NcModal from '@nextcloud/vue/dist/Components/NcModal.js'
 import NcTextField from '@nextcloud/vue/dist/Components/NcTextField.js'
 
@@ -66,9 +71,10 @@ export default {
 
 	components: {
 		NcButton,
-		NewMessageTemplatePreview,
+		NcLoadingIcon,
 		NcModal,
 		NcTextField,
+		NewMessageTemplatePreview,
 	},
 
 	props: {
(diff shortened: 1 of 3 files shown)


src/components/NewMessage/NewMessageNewFileDialog.vue
```diff
@@ -212,7 +220,9 @@ export default {
 			}
 
 			await shareFile(filePath, this.token, '', '')
+			this.loading = false
 
+			fileData = this.generateViewerObject(fileData)
```
DorraJaouad, line 225: It would be better if it is handled in `openViewer()` as it is related to it.

Antreesy:
follow-up for Nc30:
- do not use v-bind on FilePreview

src/components/NewMessage/NewMessageNewFileDialog.vue
```diff
@@ -212,6 +220,7 @@ export default {
 			}
 
 			await shareFile(filePath, this.token, '', '')
+			this.loading = false
```
DorraJaouad, line 223: There is an early return when there is an error so it will stay loading forever.

DorraJaouad (commented):
🦅

src/components/NewMessage/NewMessageNewFileDialog.vue
Antreesy: right, will add there as well

DorraJaouad (approved):

src/composables/useViewer.js
```diff
@@ -105,13 +105,13 @@ export function useViewer() {
...
-		fileid: parseInt(file.id, 10),
-		filename: generateAbsolutePath(file.path),
-		basename: file.name,
-		mime: file.mimetype,
-		hasPreview: file.previewAvailable === 'yes' || file['preview-available'] === 'yes',
+		fileid: file.fileid ?? parseInt(file.id, 10),
+		filename: file.filename ?? generateAbsolutePath(file.path),
+		basename: file.basename ?? file.name,
+		mime: file.mime ?? file.mimetype,
+		hasPreview: file.hasPreview ?? (file.previewAvailable === 'yes' || file['preview-available'] === 'yes'),
```
ShGKme, line 112: Could you describe this change?

Antreesy: Object structure differs between file in message.messageParameters and for example, Share API response (mime instead of mimetype, e.g.)
Apart from it, both objects contains permissions as numeric value, and Viewer expects a string like 'RW'

ShGKme: If we explicitly have 2 different and specific data structures here, I'd insist on 2 explicit code branches with `if` that constructs Viewer data for the first and for the second case.

Otherwise we potentially have 32 different results here, and it is not clear from the code why we have a backup value for each property.

PR #12496, opened 2024-06-11
Removed the background job and checks

### LINKED ISSUE #12428: Remove `reference_id` check
### How to use GitHub

---

We have checks in place to verify that the `reference_id` column on `oc_comments` exists, e.g.:

https://github.com/nextcloud/spreed/blob/ada70487a306751accb280a475b1f123a56d4e74/lib/BackgroundJob/CheckReferenceIdColumn.php#L51-L57

https://github.com/nextcloud/spreed/blob/ada70487a306751accb280a475b1f123a56d4e74/lib/Capabilities.php#L250-L252

Since Nextcloud 29 the `reference_id` column was added to a migration, making it mandatory for all installations:

https://github.com/nextcloud/server/blob/ae4a6e8d4454b7896d0e40f367096a26d999dbf0/core/Migrations/Version29000Date20231213104850.php#L49-L55

Therefore I suggest that we remove the BackgroundJob and the checks.

skalidindi53:
### ☑️ Resolves

* Fix #12428 

### 🚧 Tasks

- [x] Remove background job and checks related to CheckReferenceIdColumn

### 🏁 Checklist


lib/Capabilities.php
```diff
@@ -247,10 +247,6 @@ public function getCapabilities(): array {
 			$capabilities['config']['signaling']['hello-v2-token-key'] = $pubKey;
 		}
 
-		if ($this->serverConfig->getAppValue('spreed', 'has_reference_id', 'no') === 'yes') {
-			$capabilities['features'][] = 'chat-reference-id';
```
nickvergessen, line 251: The capability needs to be always on, so add it to the hardcoded feature list 🤔

SystemKeeper:
@skalidindi53 Failing tests are fixed in https://github.com/nextcloud/spreed/pull/12506. If you rebase on latest main, you should have green CI here :)

lib/Capabilities.php
nickvergessen: Now after thinking about it, maybe we should adjust the docs in https://github.com/nextcloud/spreed/blob/fb8a73c011802429138d5702ec1041fdfb81f780/docs/capabilities.md#L52 to point out that it's always set since version X

nickvergessen (approved):
Looks good otherwise, just needs a rebase for CI

PR #12509, opened 2024-06-13
feat(call): auto lower hand on speaking for 3s

### LINKED ISSUE #12399: Raised hand should be removed if participant speaks
If a participants raises his hand, the raised hand should be automatically removed once the person starts speaking. At the moment the hand is raised until the user removes it manually.

sanskar-soni-9:
### ☑️ Resolves

* Fix #12399 
  * If hand is raised, and participant started to speak, 3-seconds timer is started;
  * If person still speaks, remove the hand after timeout;
  * If person stopped speak sooner (e.g. unintended noise registration, or mic check) - stop the timer, decrease the timeout
  * Thus, speaking time is accumulated, so 3 intervals with speaking each 1 second long would also remove the raised hand

🏡 After

https://github.com/nextcloud/spreed/assets/88997516/adb5654a-bd38-467e-aad9-d270d797898d

### 🏁 Checklist

### DIFF
--- a/src/components/TopBar/TopBarMenu.vue
+++ b/src/components/TopBar/TopBarMenu.vue
@@ -178,6 +178,8 @@ import { useBreakoutRoomsStore } from '../../stores/breakoutRooms.ts'
 import { generateAbsoluteUrl } from '../../utils/handleUrl.ts'
 import { callParticipantCollection } from '../../utils/webrtc/index.js'
 
+const AUTO_LOWER_HAND_THRESHOLD = 3000
+
 export default {
 	name: 'TopBarMenu',
 
@@ -251,6 +253,9 @@ export default {
 	data() {
 		return {
 			boundaryElement: document.querySelector('.main-view'),
+			lowerHandTimeout: null,
+			speakingTimestamp: null,
+			lowerHandDelay: AUTO_LOWER_HAND_THRESHOLD,
 		}
 	},
 
@@ -377,7 +382,39 @@ export default {
 
 		showCallLayoutSwitch() {
 			return !this.$store.getters.isEmptyCallView
-		}
+		},
+
+	},
+
+	watch: {
+
+		'model.attributes.speaking'(speaking) {
+			// user stops speaking in lowerHandTimeout
+			if (this.lowerHandTimeout !== null && !speaking) {
+				this.lowerHandDelay = Math.max(0, this.lowerHandDelay - (Date.now() - this.speakingTimestamp))
+				clearTimeout(this.lowerHandTimeout)
+				this.lowerHandTimeout = null
+
+				return
+			}
+
+			// user is not speaking OR timeout is already running OR hand is not raised
+			if (!speaking || this.lowerHandTimeout !== null || !this.isHandRaised) {
+				return
+			}
+
+			this.speakingTimestamp = Date.now()
+			this.lowerHandTimeout = setTimeout(() => {
+				this.lowerHandTimeout = null
+				this.speakingTimestamp = null
+				this.lowerHandDelay = AUTO_LOWER_HAND_THRESHOLD
+
+				if (this.isHandRaised) {
+					this.toggleHandRaised()
+				}
+			}, this.lowerHandDelay)
+		},
+
 	},
 
 	methods: {


src/components/CallView/shared/LocalAudioControlButton.vue
```diff
@@ -148,6 +154,26 @@ export default {
...
+						if (this.showMicrophoneOn) {
+							this.model.toggleHandRaised(false)
+							this.$store.dispatch(
+								'setParticipantHandRaised',
+								{
+									sessionId: this.$store.getters.getSessionId(),
+									raisedHand: this.model.attributes.raisedHand,
+								}
+							)
+						}
```
DorraJaouad, line 174: Lowering/raising hand has a particular process for breakout rooms. I suggest to use the method in `TopBarMenu`
https://github.com/nextcloud/spreed/blob/main/src/components/TopBar/TopBarMenu.vue#L453-L478

danxuliu (commented):
@sanskar-soni-9 Thanks for your contribution!

As the logic was added in `LocalAudioControlButton` it is currently applied only when the audio is unmuted from the microphone button. However, it is possible to [unmute by holding the spacebar](https://github.com/nextcloud/spreed/blob/59b8f51fd053f12e47c64a165b24fd844d5876e1/src/components/TopBar/TopBarMediaControls.vue#L410-L432) (in a "walkie-talkie" way), which would not lower the hand.

To be honest I do not know if it would be better to be consistent and lower the hand also in that case, or keep it like it is now, as holding the spacebar is typically used for short messages which may not warrant automatically lowering the hand. @DorraJaouad @Antreesy Opinions? **Edit:** Oops, I saw now @ShGKme message.

If it is kept like it is now, although the logic seems to be fine (I have not tested it) I would prefer to move it to its own method and call it from `toggleAudio`. Also it is a style preference, of course, but I would prefer to use a "return early" approach (I am not one of those who prefer a single exit point in a function :-) ), so I would prefer if:
```
if (this.model.attributes.raisedHand?.state) {
	this.lowerHandTimeout = setTimeout(() => {
		if (this.showMicrophoneOn) {
			...
```
was written as:
```
if (!this.model.attributes.raisedHand?.state) {
	return
}

this.lowerHandTimeout = setTimeout(() => {
	if (!this.showMicrophoneOn) {
		return
	}

	...
```

sanskar-soni-9:
Thank you very much for detailed review, I will definitely fix it all and ready it for review when done :)

sanskar-soni-9:
Hey, I made it for speaking now, I tried to cover all the needs only time it fails is when user raises the hand while already speaking, in this case it will only work when next speaking = true event will be triggered.

So I will really like reviews like the previous one and if its not the ideal solution then maybe a help with pointing to the right direction :)

DorraJaouad (commented):
Looks good overall :)

src/components/TopBar/TopBarMenu.vue
```diff
@@ -377,7 +378,29 @@ export default {
...
+			if (!speaking || this.lowerHandTimeout !== null || !this.isHandRaised) {
+				return
+			}
+
+			this.lowerHandTimeout = window.setTimeout(() => {
+				this.lowerHandTimeout = null
+
+				if (!this.model.attributes.speaking || !this.isHandRaised) {
+					return
+				}
```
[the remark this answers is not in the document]
sanskar-soni-9: I agree with `!this.model.attributes.speaking` point, and I think making it total count of 3 sec speaking will be better as in case of 3 sec of first speaking it can be triggered by any false alarm like background noise so having a threshold for 3 secs will be better

src/components/TopBar/TopBarMenu.vue
```diff
@@ -377,7 +382,40 @@ export default {
...
+		},
+
+	},
+
+	watch: {
+
+		'model.attributes.speaking'(speaking) {
+			// user stops speaking in lowerHandTimeout
+			if (this.lowerHandTimeout !== null && !speaking) {
+				this.lowerHandDelay = AUTO_LOWER_HAND_THRESHOLD - (Date.now() - this.speakingTimestamp)
```
Antreesy, line 394: I'm afraid of getting a negative value here. Although it may be treated as 0 by `setTimeout` function, I'd prefer to wrap it with `Math.max`. Also, the value is not accumulating properly (which was intended, as I get from discussion):

```suggestion
				this.lowerHandDelay = Math.max(0, this.lowerHandDelay - (Date.now() - this.speakingTimestamp))
```

Before | After
-- | --
Screenshot from 2024-06-18 11-02-09 |  image

src/components/TopBar/TopBarMenu.vue
```diff
@@ -377,7 +382,40 @@ export default {
...
+				return
+			}
+
+			// user is not speaking OR timeout is already running OR hand is not raised
+			if (!speaking || this.lowerHandTimeout !== null || !this.isHandRaised) {
+				return
+			}
+
+			this.speakingTimestamp = Date.now()
+			this.lowerHandTimeout = window.setTimeout(() => {
```
Antreesy, line 407:
```suggestion
			this.lowerHandTimeout = setTimeout(() => {
```

Antreesy (commented):
Tested, works nicely. Left a pair of comments:

sanskar-soni-9:
thanks @Antreesy, I will add all these changes

Antreesy (approved):
Good job, thanks for the contribution! 🎉

Antreesy (approved):
Good job, thanks for the contribution! 🎉

PR #12510, opened 2024-06-13
feat: show out of office replacement

hamza221:
### ☑️ Resolves

- Part of https://github.com/nextcloud/server/issues/40398
- [x]  requires https://github.com/nextcloud/server/pull/45766

🏚️ Before | 🏡 After
-- | --
image |image

### 🚧 Tasks

- [x] ...

### 🏁 Checklist

### DIFF
--- a/src/components/NewMessage/NewMessageAbsenceInfo.vue
+++ b/src/components/NewMessage/NewMessageAbsenceInfo.vue
@@ -15,6 +15,15 @@
 				disable-tooltip />
 		</template>
 		<h4 class="absence-reminder__caption">{{ userAbsenceCaption }}</h4>
+		<div v-if="userAbsence.replacementUserId" class="absence-reminder__replacement">
+			<!-- TRANSLATORS An acting person during the period of absence of the main contact -->
+			<p>{{ t('spreed','Replacement: ') }}</p>
+			<NcUserBubble class="absence-reminder__replacement__bubble"
+				:title="t('spreed','Open conversation')"
+				:display-name="userAbsence.replacementUserDisplayName"
+				:user="userAbsence.replacementUserId"
+				@click="openConversationWithReplacementUser" />
+		</div>
 		<NcButton v-if="userAbsenceMessage && isTextMoreThanOneLine"
 			class="absence-reminder__button"
 			type="tertiary"
@@ -34,6 +43,7 @@ import { t } from '@nextcloud/l10n'
 
 import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
 import NcNoteCard from '@nextcloud/vue/dist/Components/NcNoteCard.js'
+import NcUserBubble from '@nextcloud/vue/dist/Components/NcUserBubble.js'
 
 import AvatarWrapper from '../AvatarWrapper/AvatarWrapper.vue'
 
@@ -47,6 +57,7 @@ export default {
 		ChevronUp,
 		NcButton,
 		NcNoteCard,
+		NcUserBubble,
 	},
 
 	props: {
@@ -107,6 +118,15 @@ export default {
 		setIsTextMoreThanOneLine() {
 			this.isTextMoreThanOneLine = this.$refs.absenceMessage?.scrollHeight > this.$refs.absenceMessage?.clientHeight
 		},
+
+		async openConversationWithReplacementUser() {
+			this.$router.push({
+				name: 'root',
+				query: {
+					callUser: this.userAbsence.replacementUserId,
+				}
+			}).catch(err => console.debug(`Error while pushing the new conversation's route: ${err}`))
+		},
 	},
 }
 </script>
@@ -121,6 +141,14 @@ export default {
 	&__caption {
 		font-weight: bold;
 		margin: 5px 44px 5px 0;
+	}
+
+	&__replacement {
+		display: flex;
+
+		&__bubble {
+			padding: 3px;
+		}
 	}
 
 	&__message {


src/services/participantsService.js
```diff
@@ -252,6 +253,10 @@ const getUserAbsence = async (userId) => {
 	return axios.get(generateOcsUrl('/apps/dav/api/v1/outOfOffice/{userId}/now', { userId }))
 }
 
+const getReplacementUser = async (userId) => {
+	return axios.post(generateUrl('contactsmenu/findOne'), `shareType=0&shareWith=${userId}`)
```
nickvergessen, line 257: This is restricted by sharing restrictions.
Can't we simply return the user ID and display name on the ooo endpoint?

hamza221: added

src/components/NewMessage/NewMessageAbsenceInfo.vue
```diff
@@ -15,6 +15,13 @@
 				disable-tooltip />
 		</template>
 		<h4 class="absence-reminder__caption">{{ userAbsenceCaption }}</h4>
+		<div v-if="hasReplacementUser" class="absence-reminder__replacement">
```
Antreesy, line 18: We could avoid creating a computed property here, if it's just about checking for truthy value
```suggestion
		<div v-if="userAbsence.replacementUserId" class="absence-reminder__replacement">
```

src/components/NewMessage/NewMessageAbsenceInfo.vue
```diff
@@ -15,6 +15,13 @@
...
+			<p>{{ t('spreed','Replacement: ') }}</p>
```
Antreesy, line 19: For plain, one-word lines it is better to leave an explanation for translators
```suggestion
			// TRANSLATORS An acting person during the period of absence of the main contact
			<p>{{ t('spreed','Replacement: ') }}</p>
```

src/components/NewMessage/NewMessageAbsenceInfo.vue
```diff
@@ -15,6 +15,13 @@
...
+			<NcUserBubble class="absence-reminder__replacement__bubble"
+				:display-name="userAbsence.replacementUserDn"
+				:user-id="userAbsence.replacementUserId"
+				@click="openConversationWithReplacementUser(userAbsence.replacementUserId)" />
```
Antreesy, line 23: Since you're not using an argument in the method:
```suggestion
				@click="openConversationWithReplacementUser" />
```

src/components/NewMessage/NewMessageAbsenceInfo.vue
```diff
@@ -74,7 +83,9 @@ export default {
 		token() {
 			return this.$store.getters.getToken()
 		},
-
+		hasReplacementUser() {
+			return this.userAbsence.replacementUserId !== null && this.userAbsence.replacementUserId !== ''
+		},
```
Antreesy, line 88:
```suggestion
```

Antreesy (commented):
Changes looks good! Let's wait for server PR to be merged, then we'll test it again
When possible, please provide a screenshot, if design team would like to check it out

src/components/NewMessage/NewMessageAbsenceInfo.vue
```diff
@@ -15,6 +15,14 @@
...
 		</template>
 		<h4 class="absence-reminder__caption">{{ userAbsenceCaption }}</h4>
+		<div v-if="userAbsence.replacementUserId" class="absence-reminder__replacement">
+			<!-- TRANSLATORS An acting person during the period of absence of the main contact -->
+			<p>{{ t('spreed','Replacement: ') }}</p>
+			<NcUserBubble class="absence-reminder__replacement__bubble"
+				:display-name="userAbsence.replacementUserDisplayName"
+				:user-id="userAbsence.replacementUserId"
+				@click="openConversationWithReplacementUser" />
+		</div>
```
DorraJaouad, line 25: There is no indication on UI that's clickable, I suggest to add `cursor : pointer` and a label like `chat with {user}`

src/components/NewMessage/NewMessageAbsenceInfo.vue
```diff
@@ -15,6 +15,14 @@
 				disable-tooltip />
 		</template>
 		<h4 class="absence-reminder__caption">{{ userAbsenceCaption }}</h4>
+		<div v-if="userAbsence.replacementUserId" class="absence-reminder__replacement">
+			<!-- TRANSLATORS An acting person during the period of absence of the main contact -->
+			<p>{{ t('spreed','Replacement: ') }}</p>
+			<NcUserBubble class="absence-reminder__replacement__bubble"
+				:display-name="userAbsence.replacementUserDisplayName"
+				:user-id="userAbsence.replacementUserId"
```
DorraJaouad, line 23:
```suggestion
				:user="userAbsence.replacementUserId"
```

DorraJaouad (commented):
LGTM code-wise. However, the UI doesn't look much accessible cc @nextcloud/designers 

image


image

src/components/NewMessage/NewMessageAbsenceInfo.vue
hamza221: I think that should be fixed upstream

Antreesy: Made a PR for cursor, will be fixed eventually: https://github.com/nextcloud-libraries/nextcloud-vue/pull/5708
But `title` attribute should be added at Talk app level. We could use something of existing translations, maybe?
- Private conversation
- Join conversation
- Open conversation
- Go to conversation

nickvergessen:
@DorraJaouad @Antreesy please review

src/components/NewMessage/NewMessageAbsenceInfo.vue
```diff
@@ -107,6 +118,14 @@ export default {
...
 			this.isTextMoreThanOneLine = this.$refs.absenceMessage?.scrollHeight > this.$refs.absenceMessage?.clientHeight
 		},
+
+		async openConversationWithReplacementUser() {
+			const conversation = await this.$store.dispatch('createOneToOneConversation', this.userAbsence.replacementUserId)
+			this.$router.push({
+				name: 'conversation',
+				params: { token: conversation.token },
+			}).catch(err => console.debug(`Error while pushing the new conversation's route: ${err}`))
+		},
```
ShGKme, line 128: We don't need to manually create conversation here in the store, it is supported via `callUser` links. So you can push to `root` with `{ query: { callUser: userid } }`

ShGKme: Making `this.$store.dispatch` is not possible for true links anyway if we want to make it a valid link.

DorraJaouad (approved):
Tested, thanks!

PR #12522, opened 2024-06-14
chore: audit JS dependencies

Antreesy:
### ☑️ Resolves

Audit JS dependencies

🏚️ Before | 🏡 After
-- | --
image | image

### 🏁 Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -91,6 +91,7 @@
       "version": "2.2.0",
       "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.2.0.tgz",
       "integrity": "sha512-qRmjj8nj9qmLTQXXmaR1cck3UXSRMPrbsLJAasZpF+t3riI71BXed5ebIOYwQntykeZuhjsdweEc9BxH5Jc26w==",
+      "dev": true,
       "dependencies": {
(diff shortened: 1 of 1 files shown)


DorraJaouad (approved):

Antreesy:
Audit bumps @nextcloud/capabilities to 1.2.0, which has `"npm": "^10.0.0"` against stable28 `"npm": "^9.0.0`, so I'm not so sure that we could do that

DorraJaouad:
I think it's fine? there are also a couple of deps that have `"npm": "^10.0.0"` and everything looks okay :p. 
Or we can just bump npm too ?

Antreesy:
> When doing dependency resolution, make sure that all the packages being installed will work with the running version of node.

Would be better to match the server. But I see in package-lock.json, that some other nextcloud libraries are already bumped to npm:^10 version, so one more wouldn't do more harm

ShGKme:
> which has `"npm": "^10.0.0"` against stable28 `"npm": "^9.0.0`, so I'm not so sure that we could do that

We have too strict `engines` requirements which does not reflect actual requirement. We don't use anything from new `npm` versions, so it should be fine.

ShGKme (approved):

PR #12540, opened 2024-06-18
fix: pass both arguments to the method

Antreesy:
### ☑️ Resolves

* Fix regression from #12492
  * first argument (token) was added to the method

🏚️ Before | 🏡 After
-- | --
image | image


DorraJaouad (approved):
:o

PR #12542, opened 2024-06-18
fix(federation): update counter after invites store update

Antreesy:
### ☑️ Resolves

Ref #12340: we should remove update the object records count after updating themselves.


DorraJaouad (approved):
LGTM

PR #12554, opened 2024-06-20
[stable29] perf(sharing): Use getFirstNodeById() which is more performant

nickvergessen:
As we don't care which node we get for rendering the message
As suggested by icewind

### 🏁 Checklist


lib/Chat/Parser/SystemMessage.php
```diff
@@ -756,30 +756,32 @@ protected function getFileFromShare(Room $room, ?Participant $participant, strin
...
-					$name = $node->getName();
-					$size = $node->getSize();
-					$path = $pathSegments[3] ?? $name;
 				}
+
+				$fullPath = $node->getPath();
+				$pathSegments = explode('/', $fullPath, 4);
+				$name = $node->getName();
+				$size = $node->getSize();
+				$path = $pathSegments[3] ?? $name;
```
icewind1991, line 784: `$userFolder->getRelativePath($fullPath)` is the "correct" way to do this.

icewind1991 (approved):

PR #12556, opened 2024-06-21
perf(participants): update participants session status from signaling

Antreesy:
### ☑️ Resolves

* Attempt to reduce the number of participants list updates

> [!IMPORTANT]
> 🍍Stores all sessions in Pinia store:
> - internal: by nextcloud sessionId
> - external: by signaling server sessionId
> If no attendee was mapped to the session, fetch participants and patch session later
>
> ♻️ Update participant object with data from signaling:
> - permissions
> - nextcloud sessionId
> - inCall
> - participantType
> - lastPing
> - displayName

> [!IMPORTANT]
> 🚧 Experimental feature: set config to enable:
> `occ config:app:set spreed experiments_users --value=1`

<details>
<summary> Examples of signaling messages inside: (was changed in Talk 20, have some new fields now) </summary>

### INTERNAL JOINED

```json
[
    [
        {
            "userId": "alice",
            "roomId": 7,
            "lastPing": 1718882327,
            "sessionId": "4eQPZvqBSnQ1AUb6DgO50JUgneyhx7K31c0qNZ7+M3gHEeU365NU0/1+jGueeO5oEboyJNG0cCA7xXkWxrK9dQzbNuY2ZjXdYhePoyh4j1M3h2F78DUocHvKF/I+Q56PNBT8QZts+XaswE2qW2gkm7pgr+3yD3+qaEKOZtptuE+m1JaKP7E4LGoE7ES8aD4GV+tKIgONhcurl1PNPzgMWj7NmWskJ44P/ZR3K0EYCoERCF6DLLcDXvldy4S4qqj",
            "inCall": 0,
            "participantPermissions": 246
        },
        {
            "userId": "antreesy", // Current user
            "roomId": 7,
            "lastPing": 1718882332,
            "sessionId": "0uf9N3nvaWKxz1Gn4mG5oHqnKu4N8XabhWHOSv+kmuPt8RFyK9tEk82+itf+iGflPyPzfeZr5SObW5PXDbXOO0y/wjt3O/xgoJrfI3qK7+7xmtz0WYZj0zdYaQXsv9jsEDpU6/QPXXMEoD9r+Alv4GlrzPwi0IMVo/Z4wtes2OM/jBXRhAU0AM7uwMlLfXS5ztA9QZRk8nT8sooBTiD0/D4y1rTkiXRBiX21YHNHVwVQfs38vOPnqxnNATX1Byh",
            "inCall": 0,
            "participantPermissions": 254
        }
    ]
]
```

### INTERNAL LEFT (`alice`)

```json
[
    [
        {
            "userId": "antreesy", // Current user
            "roomId": 7,
            "lastPing": 1718882333,
            "sessionId": "0uf9N3nvaWKxz1Gn4mG5oHqnKu4N8XabhWHOSv+kmuPt8RFyK9tEk82+itf+iGflPyPzfeZr5SObW5PXDbXOO0y/wjt3O/xgoJrfI3qK7+7xmtz0WYZj0zdYaQXsv9jsEDpU6/QPXXMEoD9r+Alv4GlrzPwi0IMVo/Z4wtes2OM/jBXRhAU0AM7uwMlLfXS5ztA9QZRk8nT8sooBTiD0/D4y1rTkiXRBiX21YHNHVwVQfs38vOPnqxnNATX1Byh",
            "inCall": 0,
            "participantPermissions": 254
        }
    ]
]
```

### EXTERNAL JOINED

```json
[
    [
        {
            "sessionid": "Li3jtK-0B4FW3m4eJDLajKQ7asilwKlxnUzc91Mnlrt8MVFnWnJQNlFpSFRvbkwxWGxoc3NxS19JMjNwUl9fLWNVb1FTSER4eFFnaHFFbUkyZFFWNWVzVFFRMGlMX003cVdaeWNGWXhDbWxlelR0UE5pZG1iTkcwZVdkVUJoZ2o4ZDRva2I5VERNSlcycW5iazFyMnVlX3JndkJkT1FKYzI2MGpIcHV4aUlhQWRHcGVVNWxYX2NTc0ZFMExvbGNJMFRWU19rU3I2fDI4OTE4ODgxNzE=",
            "userid": "bob",
            "user": {
                "displayname": "Bob Odenkirk"
            },
            "roomsessionid": "RnObY28zWT6p0tF84Jwlm0Xr8miMdUaCN0XBiqPuWdATkeVPxo7G7OFuIFkYft67vhDPXBpgeKRN4VUvvP8CTnNDRcsb9Xi+GDmFcV0jUxsIfZp/5PQuHpaZuWh7udfnmMc8nwC8rQWEDuH14G2OAvyJNtt5jQXeu08vt+h/gHBMboaETsEYgZZwNSSjmZHY5i/MvFu6gYHTLwojtrxIZp0iiyQmSDwPv8yjOLNxLV/eBRIySeeQzfFnDOgOMjl" // Nextcloud sessionId
        },
        {
            "sessionid": "hymgxogGWSihSgwbrtU58KVvZE073Hn0wSZM8mQDz2J8UzRMdl9IMVZDRVU2WXJpZDFCTm9SNENIMFhyNlNfM1Q3R0QxekkwLXNuMkNRenRCMDZ3ZE9Nb3EwVk5vaW1xOVdXem96Zk1DektJMmdJV2dLLWRwVW9YWjZTNE1pN2ZGT2RHeEdUVXFDa2RjRF9LZlNNQ3lHN3VuUlFXZl9qd2ZJNnI0a3VvZllWVWNYWDUzSVZxRGd4ZW1saTFEOTVDUWpXMV9QQU1RfDUyODA4ODgxNzE=",
            "userid": "alice",
            "user": {
                "displayname": "Alice Adair"
            },
            "roomsessionid": "ZljR8uqzJQYbi66LcM2z0sVFkNRtraNxQ8IfkDB0HIw1CB+lmeL8KXBO2OAoKOIoPW/enIsmsCs2fQPLP4g5rI8olyUrL3wv9bKMEJ1IdVjtaUbMXuA0GNqfzbxwr2+JfukhSecmjB7Rnf8OZGN76BXZtuXd90UKCOepoia4jXNZtg7ca4lZMpNZlWmMUueZd6IWMVvO/0Xoy4Nu/ePwTtB31DhCyIwfkwqHE6/3RlD6NHAeHKYwRv96R8IuWdN" // Nextcloud sessionId
        }
    ]
]
```

### EXTERNAL LEFT (`bob`)

```json
[
    [
        "Li3jtK-0B4FW3m4eJDLajKQ7asilwKlxnUzc91Mnlrt8MVFnWnJQNlFpSFRvbkwxWGxoc3NxS19JMjNwUl9fLWNVb1FTSER4eFFnaHFFbUkyZFFWNWVzVFFRMGlMX003cVdaeWNGWXhDbWxlelR0UE5pZG1iTkcwZVdkVUJoZ2o4ZDRva2I5VERNSlcycW5iazFyMnVlX3JndkJkT1FKYzI2MGpIcHV4aUlhQWRHcGVVNWxYX2NTc0ZFMExvbGNJMFRWU19rU3I2fDI4OTE4ODgxNzE="
    ]
]
```

### EXTERNAL CHANGED (`alice` joined call)

```json
[
    [
        {
            "lastPing": 1718962151,
            "sessionId": "cuFDIw8ygQQ29NMcr49-6WHTW7MZ4BCSdr16Vd6uvop8aXRYREVOYko2Mzg5LTBZM0NjSVB5REV0OEE2UXpMaEFxeklMVldEMEZyOGYwTHJMcmlfYjUwVlF4OXlrMU5iTEJlR0hvR1V4a2puZFdFeGlpcmhQZ1Q4R2JzdXFNbTNqMEIxTlF3QnNJWG9seHBvVFhTLTNqM0s1UUh3ekRrcWpTck9xaU9BWTUxY3dBcncxLW9Od1lhX09lME1vNWVuZmJzZExJY2lBfDI0MTI2OTgxNzE=",
            "nextcloudSessionId": "p8XiddHgLezUngrcg7GdkkTvrb+9kXzMNB2hd26YdImC5OmrTu3pPsZQ2atCvNEffYdMqyRTWuYt0OdIeZvFqGFztWNEveN5LJkXhlWiBElEedr6xZdJlDjwwnQn9O+pO3J8xivDyHrmvjDM+IwlLp/8etHLMxmLXauNHHfN1RlRXo/tlBu2AWGxqnFl54xJM8ip+4reyZckt5zQXucB3odU4x3VRP25G2NLkNDd0+42gwocfBPcxmO65Cr2IxX",
            "participantType": 3,
            "participantPermissions": 246,
            "userId": "alice",
            "inCall": 7
        }
    ]
]
```

### EXTERNAL END CALL FOR ALL

```json
undefined
```

</details>

[Screencast from 21.06.2024 13:04:46.webm](https://github.com/nextcloud/spreed/assets/93392545/b335c013-2a76-42de-ac57-436fce2dd968)

### 🚧 Tasks

- [x] Internal signaling
- [x] External signaling
- [ ] Reduce list updates -> follow-up

### 🏁 Checklist

### DIFF
--- a/src/composables/useGetParticipants.js
+++ b/src/composables/useGetParticipants.js
@@ -47,6 +47,7 @@ export function useGetParticipants(isActive = ref(true), isTopBar = true) {
 			EventBus.on('signaling-users-joined', handleUsersUpdated)
 			EventBus.on('signaling-users-changed', handleUsersUpdated)
 			EventBus.on('signaling-users-left', handleUsersLeft)
+			EventBus.on('signaling-all-users-changed-in-call-to-disconnected', handleUsersDisconnected)
 		}
 
 		// FIXME this works only temporary until signaling is fixed to be only on the calls
@@ -65,7 +66,9 @@ export function useGetParticipants(isActive = ref(true), isTopBar = true) {
 	const handleUsersLeft = ([sessionIds]) => {
 		sessionStore.updateSessionsLeft(token.value, sessionIds)
 	}
-
+	const handleUsersDisconnected = () => {
+		sessionStore.updateParticipantsDisconnectedFromStandaloneSignaling(token.value)
+	}
 	/**
 	 * Stop the get participants listeners
 	 *
@@ -76,6 +79,7 @@ export function useGetParticipants(isActive = ref(true), isTopBar = true) {
 		EventBus.off('signaling-users-joined', handleUsersUpdated)
 		EventBus.off('signaling-users-changed', handleUsersUpdated)
 		EventBus.off('signaling-users-left', handleUsersLeft)
+		EventBus.off('signaling-all-users-changed-in-call-to-disconnected', handleUsersDisconnected)
 		EventBus.off('signaling-participant-list-changed', throttleUpdateParticipants)
 		unsubscribe('guest-promoted', onJoinedConversation)
 	}
--- a/src/composables/useGetParticipants.js
+++ b/src/composables/useGetParticipants.js
@@ -9,8 +9,12 @@ import { subscribe, unsubscribe } from '@nextcloud/event-bus'
 import { useDocumentVisibility } from './useDocumentVisibility.ts'
 import { useIsInCall } from './useIsInCall.js'
 import { useStore } from './useStore.js'
-import { CONVERSATION } from '../constants.ts'
+import { CONFIG, CONVERSATION } from '../constants.ts'
+import { getTalkConfig } from '../services/CapabilitiesManager.ts'
 import { EventBus } from '../services/EventBus.ts'
+import { useSessionStore } from '../stores/session.ts'
+
+const experimentalUpdateParticipants = (getTalkConfig('local', 'experiments', 'enabled') ?? 0) & CONFIG.EXPERIMENTAL.UPDATE_PARTICIPANTS
 
 /**
  * @param {import('vue').Ref} isActive whether the participants tab is active
@@ -19,6 +23,7 @@ import { EventBus } from '../services/EventBus.ts'
 export function useGetParticipants(isActive = ref(true), isTopBar = true) {
 
 	// Encapsulation
+	const sessionStore = useSessionStore()
 	const store = useStore()
 	const token = computed(() => store.getters.getToken())
 	const conversation = computed(() => store.getters.conversation(token.value))
@@ -37,6 +42,12 @@ export function useGetParticipants(isActive = ref(true), isTopBar = true) {
 	 */
 	function initialiseGetParticipants() {
 		EventBus.on('joined-conversation', onJoinedConversation)
+		if (experimentalUpdateParticipants) {
+			EventBus.on('signaling-users-in-room', handleUsersUpdated)
+			EventBus.on('signaling-users-joined', handleUsersUpdated)
+			EventBus.on('signaling-users-changed', handleUsersUpdated)
+			EventBus.on('signaling-users-left', handleUsersLeft)
+		}
 
 		// FIXME this works only temporary until signaling is fixed to be only on the calls
 		// Then we have to search for another solution. Maybe the room list which we update
@@ -45,12 +56,26 @@ export function useGetParticipants(isActive = ref(true), isTopBar = true) {
 		subscribe('guest-promoted', onJoinedConversation)
 	}
 
+	const handleUsersUpdated = async ([users]) => {
+		if (sessionStore.updateSessions(token.value, users)) {
+			throttleUpdateParticipants()
+		}
+	}
+
+	const handleUsersLeft = ([sessionIds]) => {
+		sessionStore.updateSessionsLeft(token.value, sessionIds)
+	}
+
 	/**
 	 * Stop the get participants listeners
 	 *
 	 */
 	function stopGetParticipants() {
 		EventBus.off('joined-conversation', onJoinedConversation)
+		EventBus.off('signaling-users-in-room', handleUsersUpdated)
+		EventBus.off('signaling-users-joined', handleUsersUpdated)
+		EventBus.off('signaling-users-changed', handleUsersUpdated)
+		EventBus.off('signaling-users-left', handleUsersLeft)
 		EventBus.off('signaling-participant-list-changed', throttleUpdateParticipants)
 		unsubscribe('guest-promoted', onJoinedConversation)
 	}
--- a/src/store/participantsStore.js
+++ b/src/store/participantsStore.js
@@ -37,6 +37,7 @@ import SessionStorage from '../services/SessionStorage.js'
 import { talkBroadcastChannel } from '../services/talkBroadcastChannel.js'
 import { useCallViewStore } from '../stores/callView.ts'
 import { useGuestNameStore } from '../stores/guestName.js'
+import { useSessionStore } from '../stores/session.ts'
 import CancelableRequest from '../utils/cancelableRequest.js'
 import { convertToUnix } from '../utils/formattedTime.ts'
 import { messagePleaseTryToReload } from '../utils/talkDesktopUtils.ts'
@@ -753,6 +754,7 @@ const actions = {
 	 */
 	async patchParticipants(context, { token, newParticipants, hasUserStatuses }) {
 		const guestNameStore = useGuestNameStore()
+		const sessionStore = useSessionStore()
 
 		const currentParticipants = context.state.attendees[token]
 		for (const attendeeId of Object.keys(Object(currentParticipants))) {
@@ -769,6 +771,17 @@ const actions = {
 				if (hasUserStatuses) {
 					emitUserStatusUpdated(participant)
 				}
+			}
+
+			// Heal unknown sessions from participants request
+			// If session.inCall is undefined, this is the best attempt to get the data; but in that case
+			// it will be the same for different sessions, otherwise we trust signaling messages
+			const sessionsToUpdate = sessionStore.orphanSessions.filter(session => participant.sessionIds.includes(session.sessionId))
+			for (const session of sessionsToUpdate) {
+				sessionStore.updateSession(session.signalingSessionId, {
+					attendeeId: participant.attendeeId,
+					inCall: session.inCall ?? participant.inCall,
+				})
 			}
 
 			if (participant.participantType === PARTICIPANT.TYPE.GUEST
--- a/src/stores/session.ts
+++ b/src/stores/session.ts
@@ -99,6 +99,10 @@ export const useSessionStore = defineStore('session', {
 				return acc | (session.inCall ?? 0)
 			}, 0)
 		},
+
+		orphanSessions: (state) => {
+			return Object.values(state.sessions).filter(session => !session.attendeeId)
+		},
 	},
 
 	actions: {
@@ -110,6 +114,15 @@ export const useSessionStore = defineStore('session', {
 		deleteSession(signalingSessionId: string) {
 			if (this.sessions[signalingSessionId]) {
 				Vue.delete(this.sessions, signalingSessionId)
+			}
+		},
+
+		updateSession(signalingSessionId: string, updatedData: Partial<Session>) {
+			if (this.sessions[signalingSessionId]) {
+				Vue.set(this.sessions, signalingSessionId, {
+					...this.sessions[signalingSessionId],
+					...updatedData,
+				})
 			}
 		},
 
--- a/src/stores/session.ts
+++ b/src/stores/session.ts
@@ -38,6 +38,11 @@ type ParticipantUpdatePayload = {
 	sessionIds?: string[],
 }
 
+type ParticipantStandaloneJoinPayload = {
+	displayName: string,
+	sessionIds: string[],
+}
+
 type State = {
 	sessions: Record<string, Session>,
 }
@@ -81,6 +86,15 @@ export const useSessionStore = defineStore('session', {
 			if (signalingSessionId) {
 				return state.sessions[signalingSessionId]
 			}
+		},
+
+		getAttendeeInCall: (state) => (attendeeId: number) => {
+			return Object.values(state.sessions).reduce((acc, session) => {
+				if (session.attendeeId !== attendeeId) {
+					return acc
+				}
+				return acc | (session.inCall ?? 0)
+			}, 0)
 		},
 	},
 
@@ -181,7 +195,7 @@ export const useSessionStore = defineStore('session', {
 				}
 
 				if (isStandaloneSignalingJoinSession(user)) {
-					// this.updateParticipantJoinedFromStandaloneSignaling(token, user)
+					this.updateParticipantJoinedFromStandaloneSignaling(token, session.attendeeId, user)
 				} else if (isStandaloneSignalingUpdateSession(user)) {
 					// this.updateParticipantChangedFromStandaloneSignaling(token, user)
 				}
@@ -210,8 +224,7 @@ export const useSessionStore = defineStore('session', {
 		 */
 		updateSessionsLeft(token: string, sessionIds: StandaloneSignalingLeaveSession[]) {
 			for (const sessionId of sessionIds) {
-				// this.updateParticipantLeftFromStandaloneSignaling(token, sessionId)
-				this.deleteSession(sessionId)
+				this.updateParticipantLeftFromStandaloneSignaling(token, sessionId)
 			}
 		},
 
@@ -266,6 +279,67 @@ export const useSessionStore = defineStore('session', {
 					})
 				}
 			}
+		},
+
+		/**
+		 * Update participants joined in store according to signaling message
+		 *
+		 * @param token - Conversation token
+		 * @param attendeeId - Attendee ID
+		 * @param user - Users payload from signaling message
+		 */
+		updateParticipantJoinedFromStandaloneSignaling(token: string, attendeeId: number, user: StandaloneSignalingJoinSession) {
+			/** Exclude internal clients (phone) from update */
+			if (!user.roomsessionid || (user.user && 'callid' in user.user)) {
+				return
+			}
+
+			const participant = store.getters.getParticipant(token, attendeeId) as Participant | null
+			if (!participant) {
+				return
+			}
+
+			const updatedData: ParticipantStandaloneJoinPayload = {
+				displayName: user.user?.displayname ?? participant.displayName,
+				sessionIds: [...new Set([...participant.sessionIds, user.roomsessionid])],
+			}
+
+			store.commit('updateParticipant', { token, attendeeId, updatedData })
+		},
+
+		/**
+		 * Update participant left in store according to signaling message
+		 *
+		 * @param token - Conversation token
+		 * @param signalingSessionId - Disconnected signaling session
+		 */
+		updateParticipantLeftFromStandaloneSignaling(token: string, signalingSessionId: string) {
+			const session = this.getSession(signalingSessionId)
+			this.deleteSession(signalingSessionId)
+			const attendeeId = session?.attendeeId
+			if (!attendeeId) {
+				// Skip participant update
+				return
+			}
+
+			const participant = store.getters.getParticipant(token, attendeeId) as Participant | null
+			if (!participant) {
+				return
+			}
+
+			// If attendee has another sessions left, need to compute remaining inCall value
+			const sessionIds = participant.sessionIds.filter((id: string) => id !== session.sessionId)
+			const inCall = sessionIds.length
+				? this.getAttendeeInCall(attendeeId)
+				: PARTICIPANT.CALL_FLAG.DISCONNECTED
+			store.commit('updateParticipant', {
+				token,
+				attendeeId,
+				updatedData: {
+					sessionIds,
+					inCall,
+				},
+			})
 		},
 	},
 })
--- a/src/stores/session.ts
+++ b/src/stores/session.ts
@@ -6,7 +6,7 @@
 import { defineStore } from 'pinia'
 import Vue from 'vue'
 
-import { ATTENDEE } from '../constants.ts'
+import { ATTENDEE, PARTICIPANT } from '../constants.ts'
 import store from '../store/index.js'
 import type {
 	InternalSignalingSession,
@@ -28,6 +28,15 @@ type SignalingSessionPayload =
 	| InternalSignalingSession
 	| StandaloneSignalingJoinSession
 	| StandaloneSignalingUpdateSession
+
+type ParticipantUpdatePayload = {
+	displayName?: string,
+	inCall: number,
+	lastPing: number,
+	permissions: number,
+	participantType?: number,
+	sessionIds?: string[],
+}
 
 type State = {
 	sessions: Record<string, Session>,
@@ -179,7 +188,7 @@ export const useSessionStore = defineStore('session', {
 			}
 
 			if (isInternalSignalingSession(users)) {
-				// this.updateParticipantsFromInternalSignaling(token, users)
+				this.updateParticipantsFromInternalSignaling(token, users)
 
 				// Internal signaling server always returns all current sessions,
 				// so if some participants are missing - they are 'offline'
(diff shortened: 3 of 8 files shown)


src/composables/useGetParticipants.js
```diff
@@ -34,30 +31,56 @@ export function useGetParticipants(isActive = ref(true), isTopBar = true) {
...
 		// Then we have to search for another solution. Maybe the room list which we update
 		// periodically gets a hash of all online sessions?
 		EventBus.on('signaling-participant-list-changed', debounceUpdateParticipants)
 		subscribe('guest-promoted', onJoinedConversation)
 	}
 
+	const updateUsersFromInternalSignaling = async ([participants]) => {
+		const hasUnknownSessions = await store.dispatch('updateParticipantsFromInternalSignaling', { token: token.value, participants })
+		if (hasUnknownSessions) {
+			debounceUpdateParticipants()
```
Antreesy, line 49: RFC: Maybe we need to update immediately here?

Should happen in two cases:
- we don't have a list yet, so we already fetching it, (then request will be cancelled);
- a new participant (guest) joined and not in the list yet (would be a disaster for webinars, where new people are joining every time)

Antreesy: Kept as-is, so will be updated in 3-15s
- [ ] add trigger to cancel this request (if everyone was found in very first server response, but data wasn't there yet)

src/stores/signaling.ts
```diff
@@ -0,0 +1,308 @@
...
+		 *
+		 * @param signalingSessionIds disconnected signaling sessions;
+		 */
+		updateParticipantsLeftFromStandaloneSignaling(signalingSessionIds: string[]) {
+			for (const signalingSessionId of signalingSessionIds) {
+				const { token, attendeeId, sessionId } = this.getSignalingSession(signalingSessionId)
+				const attendee = store.getters.getParticipant(token, attendeeId)
+				if (!attendee) {
+					continue
+				}
```
DorraJaouad, line 229: This is kind of an error/ unexpected behavior?

src/stores/signaling.ts
```diff
@@ -0,0 +1,308 @@
...
+		 */
+		updateParticipantsChangedFromStandaloneSignaling(token: string, participants: StandaloneSignalingChangePayload) {
+			const guestNameStore = useGuestNameStore()
+			const attendeeUsersToUpdate: StandaloneUpdatePayload = {}
+
+			for (const participant of participants) {
+				const { token, attendeeId } = this.getSignalingSession(participant.sessionId)
+				if (!attendeeId) {
+					continue
+				}
```
DorraJaouad, line 256: If we weren't able to find `attendeeId` in signaling store for some reason, at least we try to find it from participants store as a safeguard.

src/stores/signaling.ts
```diff
@@ -0,0 +1,308 @@
...
+							? attendee.actorType !== ATTENDEE.ACTOR_TYPE.GUESTS && attendee.actorId === participant.userId
+							: attendee.actorType === ATTENDEE.ACTOR_TYPE.GUESTS && attendee.sessionIds.includes(participant.sessionId)
+					})?.attendeeId
+
+				this.addSignalingSession({
+					attendeeId,
+					token,
+					signalingSessionId: participant.sessionId,
+					sessionId: participant.sessionId,
+				})
```
DorraJaouad, line 116: You can refactor into a method for this logic ;)

src/composables/useGetParticipants.js
```diff
@@ -34,48 +32,89 @@ export function useGetParticipants(isActive = ref(true), isTopBar = true) {
...
+	}
+	const updateUsersLeftFromStandaloneSignaling = ([signalingSessionIds]) => {
+		const signalingStore = useSignalingStore()
+		signalingStore.updateParticipantsLeftFromStandaloneSignaling(signalingSessionIds)
+	}
+	const updateUsersChangedFromStandaloneSignaling = ([participants]) => {
+		const signalingStore = useSignalingStore()
+		signalingStore.updateParticipantsChangedFromStandaloneSignaling(token.value, participants)
+	}
+	const updateUsersCallDisconnectedFromStandaloneSignaling = (payload) => {
```
DorraJaouad, line 69:
```suggestion
	const updateUsersCallDisconnectedFromStandaloneSignaling = () => {
```

src/stores/signaling.ts
```diff
@@ -0,0 +1,308 @@
...
+				sessionIdsToUpdate[attendeeId].push(participant.roomsessionid)
+			}
+
+			for (const [attendeeId, sessionIds] of Object.entries(sessionIdsToUpdate)) {
+				store.commit('updateParticipant', {
+					token,
+					attendeeId,
+					updatedData: { sessionIds },
+				})
+			}
```
DorraJaouad, line 213: As we have display name, maybe add it too ?

src/composables/useGetParticipants.js
DorraJaouad: I agree to keep it as-is.

As we know from signaling message what attendees are joining, what about create a temporary participant with the info available and data will be completed once fetched ?

Antreesy: we don't have enough information to show a participant from signaling until fetch. And we got all the actual info from it, there's no need to complete it later.
Though we can consider cache participants list 🤔

src/stores/signaling.ts
```diff
@@ -0,0 +1,308 @@
...
+					continue
+				}
+
+				const attendee = store.getters.getParticipant(token, attendeeId)
+
+				if (!sessionIdsToUpdate[attendeeId]) {
+					sessionIdsToUpdate[attendeeId] = [...attendee.sessionIds]
+				}
+				// Participant might join from several devices
+				sessionIdsToUpdate[attendeeId].push(participant.roomsessionid)
```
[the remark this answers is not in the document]
Antreesy: 0) `sessionIdsToUpdate[attendeeId]` is `undefined` in else block, we need to initialize first
1) If attendee already has some sessionIds, we should not lose them
2) If attendee already has this exact sessionId, we shoud not duplicate it

Should be:
```js
if (!sessionIdsToUpdate[attendeeId]) {
	sessionIdsToUpdate[attendeeId] = [...attendee.sessionIds]
}
// Participant might join from several devices
if (!sessionIdsToUpdate[attendeeId].includes(participant.roomsessionid)) {
	sessionIdsToUpdate[attendeeId].push(participant.roomsessionid)
}
```

src/stores/signaling.ts
Antreesy: With refactored method should be doable

Antreesy:
Call stats:
image

src/stores/signaling.ts
```diff
@@ -0,0 +1,343 @@
...
+					hasUnknownSessions = true
+					continue
+				}
+
+				const attendee = store.getters.getParticipant(token, attendeeId)
+
+				if (!attendeeUsersToUpdate[attendeeId]) {
+					attendeeUsersToUpdate[attendeeId] = { sessionIds: [...attendee.sessionIds] }
+				}
+				if (participant.user.displayname) {
```
Antreesy, line 228:
```suggestion
				if (participant.user?.displayname) {
```
index.js:2 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'displayname')

src/stores/signaling.ts
```diff
@@ -0,0 +1,343 @@
...
+				const session = this.getSignalingSession(signalingSessionId)
+				if (!session) {
+					continue
+				}
+				this.deleteSignalingSession(signalingSessionId)
+
+				const { token, attendeeId, sessionId } = session
+				const attendee = store.getters.getParticipant(token, attendeeId)
+				const updatedData : { sessionIds: [], inCall?: number } = {
+					sessionIds: attendee.sessionIds.filter((id: string) => id !== sessionId)
```
Antreesy, line 264: When actor leaves conversation
```suggestion
					sessionIds: attendee?.sessionIds.filter((id: string) => id !== sessionId)
```

src/composables/useGetParticipants.js
Antreesy: Will split to the follow-up PR

Antreesy:
I'm going to keep only session store and participants update (+ test coverage) in this PR

⚠️ No performance changes or reducing amount of requests

src/stores/session.ts
```diff
@@ -0,0 +1,375 @@
...
+			if (!participant) {
+				return
+			}
+
+			const updatedData = {
+				attendeeId,
+				displayName: user.displayName ?? participant.displayName,
+				participantType: user.participantType,
+				permissions: user.participantPermissions,
+				inCall: Math.max(participant.inCall, user.inCall),
```
DorraJaouad, line 342: This signaling "change" can also come when disconnecting from the call?

DorraJaouad (commented):
I haven't tested yet, first, we need to be on the same theoretical page

src/stores/session.ts
```diff
@@ -0,0 +1,375 @@
...
+					continue
+				}
+
+				if (!participantsToUpdate[attendeeId]) {
+					// Prepare payload to update object in participants store
+					participantsToUpdate[attendeeId] = {
+						attendeeId,
+						inCall: user.inCall,
+						lastPing: user.lastPing,
+						permissions: user.participantPermissions,
```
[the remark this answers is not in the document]
Antreesy: not the permissions, as they're user-defined, but for `user.inCall` and `user.lastPing`, which are client-defined, we should take the highest value, as it's done in updateParticipantChangedFromStandaloneSignaling()

src/stores/session.ts
```diff
@@ -0,0 +1,375 @@
...
+		 * @param sessionIdsLeft - Disconnected signaling sessions
+		 */
+		updateParticipantsLeftFromStandaloneSignaling(token: string, sessionIdsLeft: string[]) {
+			for (const sessionId of sessionIdsLeft) {
+				const session = this.getSession(sessionId)
+				const attendeeId = session?.attendeeId
+				if (!attendeeId) {
+					// Skip participant update
+					continue
+				}
```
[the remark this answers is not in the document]
Antreesy: yes, orphan session without participant in the store - nothing to update. Session itself will be deleted afterwards

src/stores/session.ts
```diff
@@ -0,0 +1,375 @@
...
+
+				const participant = store.getters.getParticipant(token, attendeeId) as Participant | null
+				if (!participant) {
+					continue
+				}
```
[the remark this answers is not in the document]
Antreesy: I can't think of a good case. Let's assume someone was in the conversation and left as participant - and you fetched updated participant's list earlier than session was closed - there is nothing to update anymore.

src/stores/session.ts
Antreesy: Yes, you decrease `inCall: 7` to `inCall: 0`...
should be highest lastPing, and actual inCall. For one session it's not a problem, but for several - maybe we need to store this data as well

fancycode: Drive-by review: `inCall` contains bit flags, so you could use `participant.inCall | user.inCall` to combine them, depending on what you want to use them for.

See https://github.com/nextcloud/spreed/blob/772b90dd70cde2c473bceff3b8b1a3f0e85aa3bb/src/constants.ts#L148-L154

Antreesy: Yes, might be useful for increasing the number =)
But the other way around it's a bit more complicated:

Participant | Session 1 | Session 2 | Action | Result
-- | -- | -- | -- | --
inCall: 3 | inCall: 3 | inCall: 0 | session 2 joins with video (5 or 7) | inCall: 7 🟢 
inCall: 7 | inCall: 3 | inCall: 7 | session 2 leaves (0) | inCall: 7 ❌ should be 3 now

fancycode: Isn't this also the case with the current `max` approach and will also stay at `7`?

You could add code in `updateParticipantsLeftFromStandaloneSignaling` to calculate the correct `inCall` value from the remaining sessions of a participant.

Antreesy: Resolved, thanks again!

nickvergessen:
Also tested with SIP and it looks okay

src/stores/session.ts
```diff
@@ -0,0 +1,398 @@
...
+						inCall: user.inCall,
+						lastPing: user.lastPing,
+						permissions: user.participantPermissions,
+						sessionIds: [user.sessionId],
+					}
+				} else {
+					// Payload already exists, but participant also joined from another device
+					participantsToUpdate[attendeeId].sessionIds!.push(user.sessionId)
+					participantsToUpdate[attendeeId].inCall = participantsToUpdate[attendeeId].inCall | user.inCall
+					participantsToUpdate[attendeeId].lastPing = Math.max(participantsToUpdate[attendeeId].lastPing, user.lastPing)
```
DorraJaouad, line 262: But permissions can also change, no?

src/store/participantsStore.js
```diff
@@ -775,6 +778,17 @@ const actions = {
 				}
 			}
 
+			// Heal unknown sessions from participants request
+			// If session.inCall is null, this is the best attempt to get the data; but in that case
+			// it will be the same for different sessions, otherwise we trust signaling messages
+			const sessionsToUpdate = sessions.filter(session => participant.sessionIds.includes(session.sessionId))
```
DorraJaouad, line 784: This is not about healing sessions anymore, it will basically update all sessions.

src/store/participantsStore.js
```diff
@@ -775,6 +778,17 @@ const actions = {
...
+			for (const session of sessionsToUpdate) {
+				sessionStore.updateSession(session.signalingSessionId, {
+					attendeeId: participant.attendeeId,
+					inCall: (session.inCall === null) ? participant.inCall : session.inCall,
```
DorraJaouad, line 788:
```suggestion
					inCall: session.inCall ?? participant.inCall,
					
```

even it is 0, it's fine

src/stores/session.ts
```diff
@@ -0,0 +1,398 @@
...
+				if (session.attendeeId !== attendeeId) {
+					return acc
+				}
+				return acc | (session.inCall ?? 0)
+			}, 0)
+		},
+
+		orphanSessions: (state) => {
+			return Object.values(state.sessions).filter(session => !session.attendeeId)
+		},
```
DorraJaouad, line 107: Left over

src/stores/session.ts
```diff
@@ -0,0 +1,398 @@
...
+			}
+
+			const knownSession = this.getSession(signalingSessionId)
+			if (knownSession) {
+				return knownSession
+			}
+
+			let sessionId: Session['sessionId']
+			let attendee: Participant | null
+			let inCall: Session['inCall'] = null
```
DorraJaouad, line 144: There is no need for default null here

src/stores/session.ts
```diff
@@ -0,0 +1,398 @@
...
+		 *
+		 * @param token - Conversation token
+		 * @param users - Users payload from signaling message
+		 */
+		updateParticipantsFromInternalSignaling(token: string, users: InternalSignalingSession[]) {
+			const participantsToUpdate: Record<string, ParticipantUpdatePayload> = {}
+
+			for (const user of users) {
+				const session = this.getSession(user.sessionId)
+				this.updateSession(user.sessionId, { inCall: user.inCall })
```
DorraJaouad, line 242: You are updating a bit down in this loop anyway. Perhaps you put here to save from the skip when there is no attendeeId (there is no point because it's an orphan session then and we are not updating any other information).

src/stores/session.ts
```diff
@@ -0,0 +1,398 @@
...
+		 * Update participants changes in store according to signaling message
+		 *
+		 * @param token - Conversation token
+		 * @param attendeeId - Attendee ID
+		 * @param user - Users payload from signaling message
+		 */
+		updateParticipantChangedFromStandaloneSignaling(token: string, attendeeId: number, user: StandaloneSignalingUpdateSession) {
+			const guestNameStore = useGuestNameStore()
+
+			this.updateSession(user.sessionId, { inCall: user.inCall })
```
DorraJaouad, line 353: Maybe after participant existence check?

src/store/participantsStore.js
Antreesy: I think, we can still limit it to only orphan sessions here for checking (only if attendeeId is undefined, inCall can be null, otherwise we take fallback value from participants store). But need to check

src/stores/session.ts
Antreesy: Can change null => undefined for this param, should work the same

src/stores/session.ts
Antreesy: We updating later only participantsStore, not sessionStore. inCall is the only parameter that can change at Session, so I would update it as well (for the sake of having actual data only. It makes no sense now as internal signaling message always contains all message and we update participant inCall with bitwise sum every loop - not like with external signaling, where we have only changed data)

src/stores/session.ts
Antreesy: Internal signaling message contains all live sessions, so at every session from the same participant permissions would be exactly the same (can also do a bitwise sum, but here if actually makes no sense)

src/stores/session.ts
Antreesy: Same as for internal, we need actual data in this store

Antreesy:
Addresssed reviews, adjust tests for new changes, rebased and reshuffled commits

Antreesy:
Rebased onto main, no changes

Antreesy:
Adjusted naming for `throttleUpdateParticipants()`

src/store/participantsStore.js
```diff
@@ -775,6 +777,17 @@ const actions = {
...
+			// Heal unknown sessions from participants request
+			// If session.inCall is undefined, this is the best attempt to get the data; but in that case
+			// it will be the same for different sessions, otherwise we trust signaling messages
+			const sessionsToUpdate = sessionStore.orphanSessions.filter(session => participant.sessionIds.includes(session.sessionId))
+			for (const session of sessionsToUpdate) {
+				sessionStore.updateSession(session.signalingSessionId, {
+					attendeeId: participant.attendeeId,
+					inCall: session.inCall ?? participant.inCall,
+				})
+			}
```
DorraJaouad, line 789: You need to check that `sessionsToUpdate` is not empty

src/stores/session.ts
```diff
@@ -0,0 +1,404 @@
...
+					...this.sessions[signalingSessionId],
+					...updatedData,
+				})
+			}
+		},
+
+		findOrCreateSession(token: string, user: SignalingSessionPayload): Session | null {
+			const signalingSessionId = isStandaloneSignalingJoinSession(user) ? user.sessionid : user.sessionId
+			if (!signalingSessionId) {
+				throw new Error('Can not define sessionId from the payload')
```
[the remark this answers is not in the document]
Antreesy: to see `undefined` or `null`?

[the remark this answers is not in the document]
DorraJaouad: no, to show what attribute actually missing or expecting

src/stores/session.ts
```diff
@@ -0,0 +1,404 @@
...
+					sessionId,
+					actorId: user.actorId,
+					actorType: user.actorType,
+				}) as Participant | null
+				inCall = user.inCall
+			}
+
+			if (!sessionId) {
+				return null
+			}
```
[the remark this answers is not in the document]
Antreesy: It's nextcloud session, not signaling session.
recording or dialout phone number do not have it

src/stores/session.ts
```diff
@@ -0,0 +1,404 @@
...
+		 * Delete sessions in store and update participants objects
+		 *
+		 * @param token - Conversation token
+		 * @param sessionIds - Left session ids from signaling message
+		 */
+		updateSessionsLeft(token: string, sessionIds: StandaloneSignalingLeaveSession[]) {
+			for (const sessionId of sessionIds) {
+				this.updateParticipantLeftFromStandaloneSignaling(token, sessionId)
+			}
+		},
```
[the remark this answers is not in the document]
Antreesy: `updateSessionsLeft` is 'public' store action used in composable, which passes everything given my HPB.
`updateParticipantLeftFromStandaloneSignaling` focuses on a single id. I intentionally make it that way, so it also has less nested code =P

src/stores/session.ts
```diff
@@ -0,0 +1,404 @@
...
+			const participant = store.getters.getParticipant(token, attendeeId) as Participant | null
+			if (!participant) {
+				return
+			}
+
+			// If attendee has another sessions left, need to compute remaining inCall value
+			const sessionIds = participant.sessionIds.filter((id: string) => id !== session.sessionId)
+			const inCall = sessionIds.length
+				? this.getAttendeeInCall(attendeeId)
+				: PARTICIPANT.CALL_FLAG.DISCONNECTED
```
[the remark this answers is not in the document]
Antreesy: > we don't have info about inCall status only in `changed` signaling

That's the mandatory field there 👀 
In worst case of data missing - store is populated from participant store. so both session1 and session2 will be 'in call'. When session2 is left, `this.getAttendeeInCall(attendeeId)` will return a known call state from session1

src/store/participantsStore.js
Antreesy: it's still an array.
`for (const item of []) { /** some code */}` just won't be executed

src/composables/useGetParticipants.js
```diff
@@ -45,12 +56,32 @@ export function useGetParticipants(isActive = ref(true), isTopBar = true) {
 		subscribe('guest-promoted', onJoinedConversation)
 	}
 
+	const handleUsersUpdated = async ([users]) => {
+		const sessionStore = useSessionStore()
```
ShGKme, line 60: `useSessionStore` is a composable and in theory should be only called in the setup context: in the component's `setup` or in other composables.

It still works here because Pinia tries to simplify it by storing an active pinia instance internally. It works fine while:
- This code is executed later than the pinia instance is created and activated in the Vue app
- There is only one pinia
  - We cannot guarantee it. Any library may create a new Pinia instance

To avoid possible problems here, we should explicitly provide the pinia instance:
- Create a new root module for pinia like we have for router/vuex and export created pinia
- Use this pinia instance in pinia plugin
- Pass it to the composable

And we should make sure that these stores don't use any composables

ShGKme: In general, I don't see at the moment why sessions are stored in the Pinia store (UI global reactive state).

This store is never used as a state (this state is not used in UI/computeds, anything reactivity or Vue related). It is only used as an internal variable, and a store is used as a module with functions.

But fine by me, as removing it is a large change.

Antreesy: Will move `const sessionStore = useSessionStore()` up to setup context of `useGetParticipants`

I would keep it as all other stores, cause it makes easier at the moment to use as a stepstone for future changes (e.g. if we'll move call-related logic from Vuex participantsStore to here: `peers` state and `phones` state is also updated from signaling messages (see `processTransientCallStatus`)

Antreesy: image

PR #12558, opened 2024-06-21
feat: Added database migration for banning users

### LINKED ISSUE #12292: ⛔ Banning users and guests - API
- Overview at https://github.com/nextcloud/spreed/issues/12291

---

- [x] Deliver contract up front https://github.com/nextcloud/spreed/pull/12259
- [x] Implement database model
- [x] Implement Blocking the join
- [x] Add integration tests
- [x] https://github.com/nextcloud/spreed/issues/12670

skalidindi53:
### ☑️ Resolves

* Fix #12292 

### 🚧 Tasks

- [x] Create the API contracts, already done with https://github.com/nextcloud/spreed/pull/12259
- [x] Create the database migration in lib/Migration/ (create a skeleton file with matching name by using `occ migrations:generate spreed 20000`)
   1. Check the contract for needed fields
   2. Get inspired from `Version18000Date20230808120823` how to define the `oc_talk_bans` table, columns and indixes
- [x] Create the entity and basic mapper in lib/Model/ (Reminder and ReminderMapper look like they have all data types you need aswell). The base class QBMapper has quite some methods and involves magic methods, so it can be a bit confusing to work with in the beginning.
- [x] Create a service (or put the methods into the BanMapper) to handle validating, creating, retrieving and deleting bans
- [x] Use the service or mapper in the controller and other code blocks that were touched for the contract in step 1
- [x] Write integration tests (I'd put them into `tests/integration/features/conversation-1/ban.feature`)
- [x] Ensure all queries are covered by indexes 

### 🏁 Checklist

### DIFF
--- a/appinfo/routes/routesBanController.php
+++ b/appinfo/routes/routesBanController.php
@@ -17,6 +17,6 @@ return [
 		/** @see \OCA\Talk\Controller\BanController::listBans() */
 		['name' => 'Ban#listBans', 'url' => '/api/{apiVersion}/ban/{token}', 'verb' => 'GET', 'requirements' => $requirements],
 		/** @see \OCA\Talk\Controller\BanController::unbanActor() */
-		['name' => 'Ban#unbanActor', 'url' => '/api/{apiVersion}/ban/{token}', 'verb' => 'DELETE', 'requirements' => $requirements],
+		['name' => 'Ban#unbanActor', 'url' => '/api/{apiVersion}/ban/{token}/{banId}', 'verb' => 'DELETE', 'requirements' => $requirements],
 	],
 ];
--- a/lib/Controller/BanController.php
+++ b/lib/Controller/BanController.php
@@ -48,17 +48,16 @@ class BanController extends AEnvironmentAwareController {
 	#[RequireModeratorParticipant]
 	public function banActor(string $actorType, string $actorId, string $internalNote = ''): DataResponse {
 		try {
-			$attendee = $this->participant->getAttendee();
-			$roomId = $attendee->getRoomId();
-			$bannedId = $attendee->getActorId();
-			$bannedType = $attendee->getActorType();
+			$moderator = $this->participant->getAttendee();
+			$moderatorActorType = $moderator->getActorType();
+			$moderatorActorId = $moderator->getActorId();
 
 			$ban = $this->banService->createBan(
+				$moderatorActorId,
+				$moderatorActorType,
+				$this->room->getId(),
 				$actorId,
 				$actorType,
-				$roomId,
-				$bannedId,
-				$bannedType,
 				null,
 				$internalNote
 			);
--- a/lib/Controller/BanController.php
+++ b/lib/Controller/BanController.php
@@ -111,19 +111,4 @@ class BanController extends AEnvironmentAwareController {
 		$this->banService->findAndDeleteBanByIdForRoom($banId, $this->room->getId());
 		return new DataResponse([], Http::STATUS_OK);
 	}
-
-	/**
-	 * @psalm-return TalkBan
-	 */
-	protected function randomBan(string $actorType, string $actorId): array {
-		return [
-			'id' => random_int(1, 1337),
-			'actorType' => $this->participant->getAttendee()->getActorType(),
-			'actorId' => $this->participant->getAttendee()->getActorId(),
-			'bannedType' => $actorType,
-			'bannedId' => $actorId,
-			'bannedTime' => random_int(1514747958, 1714747958),
-			'internalNote' => '#NOTE#' . $actorType . '#' . $actorId . '#' . sha1($actorType . $actorId),
-		];
-	}
 }
--- a/lib/Controller/BanController.php
+++ b/lib/Controller/BanController.php
@@ -104,7 +104,7 @@ class BanController extends AEnvironmentAwareController {
 	#[PublicPage]
 	#[RequireModeratorParticipant]
 	public function unbanActor(int $banId): DataResponse {
-		$this->banService->findAndDeleteBanById($banId);
+		$this->banService->findAndDeleteBanByIdForRoom($banId, $this->room->getId());
 		return new DataResponse([], Http::STATUS_OK);
 	}
 
--- a/lib/Controller/BanController.php
+++ b/lib/Controller/BanController.php
@@ -103,16 +103,8 @@ class BanController extends AEnvironmentAwareController {
 	 */
 	#[PublicPage]
 	#[RequireModeratorParticipant]
-	public function unbanActor(): DataResponse {
-		$banId = $this->request->getParam('id');
-		if ($banId === null) {
-			return new DataResponse([
-				'error' => 'Missing ban ID',
-			], Http::STATUS_BAD_REQUEST);
-		}
-		echo "Ban ID!!!: $banId\n";
-		$this->banService->findAndDeleteBanById((int)$banId);
-		echo "Unbanned successfully\n";
+	public function unbanActor(int $banId): DataResponse {
+		$this->banService->findAndDeleteBanById($banId);
 		return new DataResponse([], Http::STATUS_OK);
 	}
 
--- a/lib/Controller/BanController.php
+++ b/lib/Controller/BanController.php
@@ -11,6 +11,7 @@ namespace OCA\Talk\Controller;
 
 use OCA\Talk\Middleware\Attribute\RequireModeratorParticipant;
 use OCA\Talk\Model\Attendee;
+use OCA\Talk\Model\Ban;
 use OCA\Talk\ResponseDefinitions;
 use OCA\Talk\Service\BanService;
 use OCP\AppFramework\Http;
@@ -79,7 +80,7 @@ class BanController extends AEnvironmentAwareController {
 	 *
 	 * Required capability: `ban-v1`
 	 *
-	 * @return DataResponse<Http::STATUS_OK, list<TalkBan>, array{}>
+	 * @return DataResponse<Http::STATUS_OK, TalkBan[], array{}>
 	 *
 	 * 200: List all bans
 	 */
@@ -87,11 +88,7 @@ class BanController extends AEnvironmentAwareController {
 	#[RequireModeratorParticipant]
 	public function listBans(): DataResponse {
 		$bans = $this->banService->getBansForRoom($this->room->getId());
-		$result = array_map(function ($ban) {
-			return $ban->jsonSerialize();
-		}, $bans);
-
-		/** @psalm-var list<array{actorId: string, actorType: string, bannedId: string, bannedTime: int, bannedType: string, id: int, internalNote: string}> $result */
+		$result = array_map(static fn (Ban $ban): array => $ban->jsonSerialize(), $bans);
 		return new DataResponse($result, Http::STATUS_OK);
 	}
 
--- a/lib/Controller/BanController.php
+++ b/lib/Controller/BanController.php
@@ -12,6 +12,7 @@ namespace OCA\Talk\Controller;
 use OCA\Talk\Middleware\Attribute\RequireModeratorParticipant;
 use OCA\Talk\Model\Attendee;
 use OCA\Talk\ResponseDefinitions;
+use OCA\Talk\Service\BanService;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\Attribute\PublicPage;
 use OCP\AppFramework\Http\DataResponse;
@@ -24,6 +25,7 @@ class BanController extends AEnvironmentAwareController {
 	public function __construct(
 		string $appName,
 		IRequest $request,
+		protected BanService $banService
 	) {
 		parent::__construct($appName, $request);
 	}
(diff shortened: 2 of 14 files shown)


lib/Migration/Version20000Date20240621150333.php
```diff
@@ -0,0 +1,60 @@
...
+		$schema = $schemaClosure();
+
+		if (!$schema->hasTable('oc_talk_bans')) {
+			$table = $schema->createTable('oc_talk_bans');
+			$table->addColumn('id', Types::BIGINT, [
+				'autoincrement' => true,
+				'notnull' => true,
+				'length' => 20,
+			]);
+			$table->addColumn('user_id', Types::STRING, [
```
nickvergessen, line 35: We need actor type and actor id to identify participants

lib/Migration/Version20000Date20240621150333.php
```diff
@@ -0,0 +1,60 @@
...
+	 * @param IOutput $output
+	 * @param Closure(): ISchemaWrapper $schemaClosure
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if (!$schema->hasTable('oc_talk_bans')) {
```
nickvergessen, line 28: oc_ is automatically added by the migration code, sorry for.the confusion 😅

lib/Migration/Version20000Date20240621150333.php
```diff
@@ -0,0 +1,60 @@
...
+				'length' => 20,
+			]);
+			$table->addColumn('user_id', Types::STRING, [
+				'length' => 64,
+			]);
+			$table->addColumn('room_id', Types::BIGINT, [
+				'notnull' => true,
+				'unsigned' => true,
+			]);
+			$table->addColumn('banned_by', Types::STRING, [
```
nickvergessen, line 42: Also for banned by we should store type and id, so that guest moderators can ban in the future

lib/Migration/Version20000Date20240621150333.php
```diff
@@ -0,0 +1,60 @@
...
+			$table->addColumn('banned_at', Types::DATETIME, [
+				'notnull' => false,
+			]);
+			$table->addColumn('reason', Types::TEXT, [
+				'notnull' => false,
+			]);
+
+			$table->setPrimaryKey(['id']);
+			$table->addUniqueIndex(['user_id', 'room_id'], 'talk_ban_user_room'); //A user should not be banned from the same room more than once
+			$table->addIndex(['banned_at'], 'talk_ban_banned_at');
```
nickvergessen, line 54: For indexes it might be better to first write the mapper/service, then we know which queries we have

nickvergessen (commented):
And you need to bump the last digit in the version of appinfo/info.xml

lib/Model/Ban.php
```diff
@@ -0,0 +1,59 @@
...
+ * @method string getBannedByActorId()
+ * @method void setBannedByActorType(string $bannedByActorType)
+ * @method string getBannedByActorType()
+ * @method void setBannedAt(\DateTime $bannedAt)
+ * @method \DateTime getBannedAt()
+ * @method void setReason(string $reason)
+ * @method string getReason()
+ */
+class Ban extends Entity implements \JsonSerializable {
+	protected $actorId = '';
```
nickvergessen, line 30:
```suggestion
	protected string $actorId = '';
```

lib/Model/BanMapper.php
```diff
@@ -0,0 +1,53 @@
...
+		$query->select('*')
+			->from($this->getTableName())
+			->where($query->expr()->eq('actor_id', $query->createNamedParameter($actorId, IQueryBuilder::PARAM_STR)))
+			->andWhere($query->expr()->eq('actor_type', $query->createNamedParameter($actorType, IQueryBuilder::PARAM_STR)))
+			->andWhere($query->expr()->eq('room_id', $query->createNamedParameter($roomId, IQueryBuilder::PARAM_INT)));
+
+		return $this->findEntity($query);
+	}
+
+	public function deleteBanForActorAndRoom(string $actorId, string $actorType, int $roomId): void {
```
nickvergessen, line 44: Could also findForActorAndRoom and then call delete with it.
In case we add events to the ban service in the future we need the instance anyway.

lib/Service/BanService.php
```diff
@@ -0,0 +1,66 @@
...
+	public function createBan(string $actorId, string $actorType, int $roomId, string $bannedByActorId, string $bannedByActorType, ?DateTime $bannedAt, ?string $reason): Ban {
+		$this->validateBanData($actorId, $actorType, $roomId, $bannedByActorId, $bannedByActorType, $bannedAt, $reason);
+
+		$ban = new Ban();
+		$ban->setActorId($actorId);
+		$ban->setActorType($actorType);
+		$ban->setRoomId($roomId);
+		$ban->setBannedByActorId($bannedByActorId);
+		$ban->setBannedByActorType($bannedByActorType);
+		$ban->setBannedAt($bannedAt ?? new DateTime());
```
nickvergessen, line 47: Because it is not easily testable, we have `OCP\AppFramework\Utility\ITimeFactory::getDateTime()` to generate "now"

lib/Model/BanMapper.php
skalidindi53: Will do

lib/Model/BanMapper.php
```diff
@@ -0,0 +1,62 @@
...
+		$query = $this->db->getQueryBuilder();
+		$query->delete($this->getTableName())
+			->where($query->expr()->eq('actor_id', $query->createNamedParameter($actorId, IQueryBuilder::PARAM_STR)))
+			->andWhere($query->expr()->eq('actor_type', $query->createNamedParameter($actorType, IQueryBuilder::PARAM_STR)))
+			->andWhere($query->expr()->eq('room_id', $query->createNamedParameter($roomId, IQueryBuilder::PARAM_INT)));
+
+		$query->executeStatement();
+	}
+
+	public function findAndDeleteBanForActorAndRoom(string $actorId, string $actorType, int $roomId): void {
```
nickvergessen, line 54: I'd do that in the service

lib/Controller/BanController.php
```diff
@@ -45,25 +47,39 @@ public function __construct(
...
+			return new DataResponse(
+				[
+					'id' => $ban->getId(),
+					'actorType' => $ban->getActorType(),
+					'actorId' => $ban->getActorId(),
+					'bannedType' => $ban->getBannedByActorType(),
+					'bannedId' => $ban->getBannedByActorId(),
+					'bannedTime' => $ban->getBannedAt()->getTimestamp(),
+					'internalNote' => $ban->getInternalNote(),
+				],
```
nickvergessen, line 75: You added `jsonSerialize()` to the Ban model. That function should return what you need here, if it does not you should adjust it.
Same in the list endpoint.

lib/Controller/BanController.php
```diff
@@ -102,7 +104,16 @@ public function listBans(): DataResponse {
 	 */
 	#[PublicPage]
 	#[RequireModeratorParticipant]
-	public function unbanActor(int $banId): DataResponse {
+	public function unbanActor(): DataResponse {
```
nickvergessen, line 107: banId should be taken from the URL
But I just noticed I missed the parameter there:
https://github.com/nextcloud/spreed/blob/main/appinfo/routes/routesBanController.php#L20
should be:
```
['name' => 'Ban#unbanActor', 'url' => '/api/{apiVersion}/ban/{token}/{banId}', 'verb' => 'DELETE',…
```
there

lib/Model/BanMapper.php
```diff
@@ -0,0 +1,72 @@
...
+use OCP\IDBConnection;
+
+/**
+ * @method Ban mapRowToEntity(array $row)
+ * @method Ban findEntity(IQueryBuilder $query)
+ * @method Ban[] findEntities(IQueryBuilder $query)
+ * @template-extends QBMapper<Ban>
+ */
+class BanMapper extends QBMapper {
+	use TTransactional;
```
nickvergessen, line 24:
```suggestion
```

You are not using it, so we don't need it?

lib/Model/BanMapper.php
```diff
@@ -0,0 +1,72 @@
...
+	public function findByRoomId(int $roomId): array {
+		$query = $this->db->getQueryBuilder();
+		$query->select('*')
+			->from($this->getTableName())
+			->where($query->expr()->eq('room_id', $query->createNamedParameter($roomId, IQueryBuilder::PARAM_INT)));
+
+		return $this->findEntities($query);
+	}
+
+	public function findByBanId(int $banId): Ban {
```
nickvergessen, line 53: document DoesNotExistException

appinfo/routes/routesBanController.php
```diff
@@ -17,6 +17,6 @@
 		/** @see \OCA\Talk\Controller\BanController::listBans() */
 		['name' => 'Ban#listBans', 'url' => '/api/{apiVersion}/ban/{token}', 'verb' => 'GET', 'requirements' => $requirements],
 		/** @see \OCA\Talk\Controller\BanController::unbanActor() */
-		['name' => 'Ban#unbanActor', 'url' => '/api/{apiVersion}/ban/{token}', 'verb' => 'DELETE', 'requirements' => $requirements],
+		['name' => 'Ban#unbanActor', 'url' => '/api/{apiVersion}/ban/{token}/{banId}', 'verb' => 'DELETE', 'requirements' => $requirements],
```
Antreesy, line 20: Don't forget about endpoint here:
https://github.com/nextcloud/spreed/blob/04b1d46a41affedd2065d98674beface67e437a3/src/services/banService.ts#L45-L50

```js
const unbanActor = async function(token: string, banId: unbanActorParams['banId'], options?: object): unbanActorResponse {
	return axios.delete(generateOcsUrl('/apps/spreed/api/v1/ban/{token}/{banId}', { token, banId }, options), options)
}
```

And pay attention, if there would be openapi changes because of that

nickvergessen: > if there would be openapi changes because of that

There will be and have to be :( I'm sorry about that.

lib/Migration/Version20000Date20240621150333.php
```diff
@@ -0,0 +1,70 @@
...
+			]);
+			$table->addColumn('banned_time', Types::DATETIME, [
+				'notnull' => false,
+			]);
+			$table->addColumn('internal_note', Types::TEXT, [
+				'notnull' => false,
+			]);
+
+			$table->setPrimaryKey(['id']);
+			// $table->addUniqueIndex(['user_id', 'room_id'], 'talk_ban_user_room'); //A user should not be banned from the same room more than once
```
nickvergessen, line 63: This index makes sense for uniqueness and before your query on join is matching that.
```suggestion
			$table->addUniqueIndex(['actor_type', 'actor_id', 'room_id'], 'talk_ban_user_room'); //A user should not be banned from the same room more than once
```

lib/Migration/Version20000Date20240621150333.php
```diff
@@ -0,0 +1,70 @@
...
+			// $table->addIndex(['banned_at'], 'talk_ban_banned_at');
```
nickvergessen, line 64: We don't need time, but `room_id` only, to list the bans in a room

lib/Service/BanService.php
```diff
@@ -0,0 +1,100 @@
...
+	 * Validate the ban data.
+	 */
+	private function validateBanData(string $actorId, string $actorType, int $roomId, string $bannedId, string $bannedType, ?DateTime $bannedTime, ?string $internalNote): void {
+		if (empty($actorId) || empty($actorType) || empty($roomId) || empty($bannedId) || empty($bannedType)) {
+			throw new \InvalidArgumentException("Invalid ban data provided.");
+		}
+
+		if ($bannedTime !== null && !$bannedTime instanceof DateTime) {
+			throw new \InvalidArgumentException("Invalid date format for bannedTime.");
+		}
```
nickvergessen, line 35: Should expend this and have an exception for each of them with their own "key" (not "Invalid ban data provided", but `invalid_banned`, `invalid_note`, …) (where it makes sense, e.g. roomId can never be invalid and neither can "actor" as it's not provided but generated inside our code)

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -1455,6 +1467,112 @@ public function userRemovesAttendeeFromRoom(string $user, string $actorType, str
...
+		}
+
+		$bans = $this->getDataFromResponse($this->response);
+		foreach ($bans as &$key) {
+			unset($key['id'], $key['roomId'], $key['bannedTime']);
+		}
+
+		var_dump($bans);
+		var_dump($tableNode->getColumnsHash());
+
```
nickvergessen, line 1539:
```suggestion
```

lib/Service/BanService.php
```diff
@@ -0,0 +1,101 @@
...
+	/**
+	 * Retrieve a ban by its ID and delete it.
+	 */
+	public function findAndDeleteBanById(int $banId): void {
+		try {
+			$ban = $this->banMapper->findByBanId($banId);
+			$this->banMapper->delete($ban);
+		} catch (DoesNotExistException $e) {
+			// Ban does not exist
+			echo "Ban does not exist.\n";
```
nickvergessen, line 98: Debug code

Antreesy:
https://github.com/nextcloud/spreed/blob/ec73a36edcdfc76757e141a09284ed26733dc286/package.json#L13

nickvergessen (approved):

PR #12575, opened 2024-06-24
fix(MessageButtonsBar): Hide reply button if participant has no chat permission

provokateurin:
### ☑️ Resolves

The button was displayed, even when the user had no permissions to send a message to the chat.

### 🏁 Checklist


Antreesy (approved):
Good catch!

PR #12578, opened 2024-06-24
[stable29] chore(deps): bump and fix nexcloud dependencies

Antreesy:
### ☑️ Resolves

* bump `nexcloud/capabilities` from v1.1.0 to v1.2.0
  * this cleans up extraneous prod.dependencies from the lib (`babel`, `jest`, e.t.c)
* fix `nexcloud/vue` to v8.11.3 (exact)
  * this prevents an accidental library bump by dependabot
* audit the rest dependencies

### 🏁 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "@nextcloud/paths": "^2.1.0",
     "@nextcloud/router": "^3.0.1",
     "@nextcloud/upload": "^1.0.5",
-    "@nextcloud/vue": "^8.11.3",
+    "@nextcloud/vue": "8.11.3",
     "crypto-js": "^4.2.0",
     "debounce": "^2.0.0",
     "emoji-regex": "^10.3.0",
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "@nextcloud/auth": "^2.2.1",
     "@nextcloud/axios": "^2.4.0",
     "@nextcloud/browser-storage": "^0.3.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
     "@nextcloud/auth": "^2.2.1",
     "@nextcloud/axios": "^2.4.0",
     "@nextcloud/browser-storage": "^0.3.0",
-    "@nextcloud/capabilities": "^1.1.0",
+    "@nextcloud/capabilities": "1.2.0",
```
DorraJaouad, line 31: I don't see a reason to freeze future dependabot bumps

ShGKme: > this cleans up extraneous prod.dependencies from the lib (babel, jest, e.t.c)

I didn't understand.

It has no such dependencies, hasn't it?

ShGKme: And 1.2.0 is the latest version

Antreesy: > It has no such dependencies, hasn't it?

See here: https://github.com/nextcloud-libraries/nextcloud-capabilities/commit/185933db7fad4f26a499450b3c220c78041923e1

Yes, it's the latest ATM, I can set it back as `^1.2.0`.

ShGKme: > Yes, it's the latest ATM, I can set it back as `^1.2.0`.

Let's do it, then auto update works. With `@nextcloud/vue` it makes sense because of a regression.

DorraJaouad (approved):

PR #12584, opened 2024-06-26
fix(MessagesList): fix unread marker when it is attached to the last message in collapsed group

DorraJaouad:
### ☑️ Resolves

There is a special workaround when last read message is the last message in a collapsed group: we set the unread marker to the combined system message instead #11876. Thus, when checking `data-seen` attribute, it should check the attribute of the combined system message.

🏚️ Before | 
Conversations cannot be cleared from unread messages if the unread marker is attached to a combined system message.

🏡 After
Conversations can be cleared from unread messages.

### 🏁 Checklist


src/components/MessagesList/MessagesList.vue
```diff
@@ -967,6 +967,16 @@ export default {
...
 				el = el.closest('.message')
+				if (el === null || el.offsetParent === null) {
+					// Exception: when the message remains not visible
+					// e.g: it is the last message in collapsed group
+					// unread marker is set to the combined group.
+					// Look for the unread marker itself
+					el = document.querySelector('.new-message-marker')
+					if (el) {
+						el = el.closest('.message')
+					}
```
Antreesy, line 978:
```suggestion
					} else {
						// If we don't have a marker, would it be safe to assume everything is read and return the last message in the list?
						// If it is visible, then we read everything, and `lastReadMessageElement` exists
					}
```

DorraJaouad: > // If we don't have a marker, would it be safe to assume everything is read and return the last message in the list?

The only acceptable case where there is no marker, is that last read message is the last message in the conversation (which is first thing we check before updating read marker). Otherwise, if there is no marker and the last read message is not the last message, it means there is an issue, we should probably log ? 

> // If it is visible, then we read everything, and `lastReadMessageElement` exists

?

Antreesy: We probably should

Antreesy: We probably should

PR #12586, opened 2024-06-26
fix(sidebar): tab is not shown after switching a conversation

ShGKme:
### ☑️ Resolves

0. Open Nextcloud Talk on `/apps/spreed`
1. Open a chat with one tab, for example, one-to-one with only **Shared Media** tab
2. Switch to a conversation with only one another tab, for example, Federation chat with only **Participants** tab
3. See no tab

Follow-up: add note to docs in `nextcloud-vue`.

### Why

`NcAppSidebar` registers tabs when they are created or destroyed.

But in this case, with Vue optimizing rendering, no tab has been destroyed or created. Instead, the only tab has been reused: it has received new content and `id` changed from `shared-media` to `participants`.

Supporting tab's `id` change in `NcAppSidebar` is not really possible and risky.

A solution - identify `NcAppSidebarTab`s with `key`, so Vue knows that `shared-items` and `participants` or other tabs are actually different tabs and doesn't try to reuse tabs.

The same solution is required in other cases where Vue has to identify nodes: KeepAlive, Transition, `v-for`.

### 🏁 Checklist

### DIFF
--- a/src/components/RightSidebar/RightSidebar.vue
+++ b/src/components/RightSidebar/RightSidebar.vue
@@ -17,6 +17,7 @@
 		</template>
 		<NcAppSidebarTab v-if="isInCall"
 			id="chat"
+			key="chat"
 			:order="1"
 			:name="t('spreed', 'Chat')">
 			<template #icon>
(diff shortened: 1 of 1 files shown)


ShGKme:
Thanks for noticing the issue @Antreesy

DorraJaouad (approved):
tested, thanks for the explanation.

Antreesy (approved):

PR #12607, opened 2024-06-28
refactor(sidebar): stop relying on internal classes and rename class

ShGKme:
### ☑️ Resolves

* https://github.com/nextcloud/spreed/pull/12387

No visual changes

### 🚧 Tasks

- [x] Don't rely on `NcCounterBubble` internal class
- [x] Rename class back to name from #12387 (probably was overwritten during force push)

### 🏁 Checklist


Antreesy (approved):
Yep, also a good approach

DorraJaouad (approved):

PR #12632, opened 2024-07-02
fix(chat): promptly update messages list in sidebar, store and chat

Antreesy:
### ☑️ Resolves

* Update deleted messages in LefSidebar
* Drop last edit information for deleted messages (edited before)
* Simplify soft group update in MessagesList
  * keep simple conditions: 1 `group no longer exists`, 2 `group did not exist before`, 3 `group is different from previous version`
  * re-sort conditions for groups identical comparison (to make early return on less complex operations)

> [!TIP]
> Case 2: delete edited message
> Before | After
> -- | -- 
> image | image

> [!TIP]
> Case 3: drop last commit, post message 1, post message 2, edit message 2, delete message 2
> Before | After
> -- | -- 
> image | image

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -406,49 +406,31 @@ export default {
 		},
 
 		softUpdateAuthorGroups(oldGroups, newGroups, dateTimestamp) {
-			const newGroupIdSet = new Set(this.messagesList.map(message => message.id))
-			const newGroupsKeys = Object.keys(newGroups)
+			const groupIds = new Set([...Object.keys(oldGroups), ...Object.keys(newGroups)])
 
-			Object.entries(oldGroups).forEach(([id, oldGroup]) => {
-				if (newGroupsKeys.includes(id)) {
-					// Group is presented, will be handled in the next loop
-					return
-				}
-
-				if (oldGroup.messages.some(message => !newGroupIdSet.has(message.id))) {
-					// Delete groups of normal and temporary messages,
-					// if at least one message from the group is no longer in the store
+			groupIds.forEach(id => {
+				if (oldGroups[id] && !newGroups[id]) {
+					// group no longer exists, remove
 					this.$delete(this.messagesGroupedByDateByAuthor[dateTimestamp], id)
-				}
-			})
-			Object.entries(newGroups).forEach(([id, newGroup]) => {
-				if (!oldGroups[id]) {
-					const oldId = Object.keys(oldGroups)
-						.find(key => +id < +key && oldGroups[key].nextMessageId <= newGroup.nextMessageId)
-					if (oldId) {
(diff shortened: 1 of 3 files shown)


src/components/MessagesList/MessagesList.vue
```diff
@@ -406,49 +406,38 @@ export default {
...
-					this.$set(this.messagesGroupedByDateByAuthor[dateTimestamp], id, newGroup)
-				} else if (!this.areGroupsIdentical(newGroup, oldGroups[id])) {
-					// newGroup includes oldGroup and more recent messages
-					this.$set(this.messagesGroupedByDateByAuthor[dateTimestamp], id, newGroup)
+				} else if (newGroups[id] && !oldGroups[id]) {
+					// group did not exist before, add
+					this.$set(this.messagesGroupedByDateByAuthor[dateTimestamp], id, newGroups[id])
+				} else if (!this.areGroupsIdentical(newGroups[id], oldGroups[id])) {
+					// group exists in both, update
+					this.$set(this.messagesGroupedByDateByAuthor[dateTimestamp], id, newGroups[id])
```
DorraJaouad, line 420: You can merge these two together :)

src/components/MessagesList/MessagesList.vue
```diff
@@ -406,49 +406,38 @@ export default {
...
 
-			// Check for temporary messages, replaced with messages from server
-			return group1.messages.every((message, index) => group2.messages[index].id === message.id)
+			// Compare ids (temporary messages, replaced with messages from server)
+			if (!group1.messages.every((message, index) => group2.messages[index].id === message.id)) {
+				return false
+			}
+
+			// Compare stringified arrays of messages
+			return JSON.stringify(group1.messages) === JSON.stringify(group2.messages)
```
DorraJaouad, line 440: Could `JSON.stringify` comparison be enough to also spot if there are some temp messages replaced from server ( ids are not the same so it will return false)?

Antreesy: it would be enough. But:
- JSON conversion is complex and requires expensive computations
- after conversion, we need to compare two strings character by character (minimum converted string length for single message is ~400 characters)
- with id comparison, we check only N pairs of values - make sense to try it first

src/components/MessagesList/MessagesList.vue
Antreesy: I know, just wanted to make steps clear

src/components/MessagesList/MessagesList.vue
DorraJaouad: Yes but temp messages are usually  in 1 group and JSON conversion would thus be applied on n-1 items while 1 group is exited with id comparison.
The question is if n-1 id comparison pairs are cheaper than 1 JSON coversion/comparison.

DorraJaouad (approved):
Tested 🦅 , looks good!

PR #12633, opened 2024-07-03
fix(settings): drop local setting for background blur filter, use provided by server

Antreesy:
### ☑️ Resolves

* Follow-up to #10464 
* Ref https://github.com/nextcloud/server/pull/45395
* Keep checkbox and translation for the reference, add link to server theming settings

> [!TIP]
> To discard state, use `occ user:setting <user_id> theming force_enable_blur_filter ''`

🏚️ Before | 🏡 After
-- | --
image | image

### 🏁 Checklist

---

### 🚧 Tasks

- [ ] Check TInitialState change

### 🏁 Checklist

### DIFF
--- a/src/components/CallView/CallView.vue
+++ b/src/components/CallView/CallView.vue
@@ -4,7 +4,7 @@
 -->
 
 <template>
-	<div id="call-container" :class="{ 'blurred': isBackgroundBlurred }">
+	<div id="call-container">
 		<ViewerOverlayCallView v-if="isViewerOverlay"
 			:token="token"
 			:model="promotedParticipantModel"
@@ -134,7 +134,6 @@ import VideoVue from './shared/VideoVue.vue'
 import ViewerOverlayCallView from './shared/ViewerOverlayCallView.vue'
 
 import { SIMULCAST } from '../../constants.js'
-import BrowserStorage from '../../services/BrowserStorage.js'
 import { fetchPeers } from '../../services/callsService.js'
 import { getTalkConfig } from '../../services/CapabilitiesManager.ts'
 import { EventBus } from '../../services/EventBus.js'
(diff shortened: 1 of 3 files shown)


src/components/CallView/CallView.vue
```diff
@@ -417,7 +415,6 @@ export default {
 		// Ensure that data is properly initialized before mounting the
 		// subviews.
 		this.updateDataFromCallParticipantModels(this.callParticipantModels)
-		this.isBackgroundBlurred = BrowserStorage.getItem('background-blurred') !== 'false'
```
nickvergessen, line 420: should we still read this? Otherwise there will be a performance regression on "update"

Antreesy: I would say, we can read it once, and if value exists in the storage, force update theming on server via new API, then drop value from storage?

nickvergessen: sounds good to me

src/components/SettingsDialog/SettingsDialog.vue
```diff
@@ -263,11 +277,12 @@ export default {
 
 	created() {
 		const blurred = BrowserStorage.getItem('background-blurred')
-		if (blurred === null) {
-			BrowserStorage.setItem('background-blurred', 'true')
+		if (blurred === 'false' && isBackgroundBlurred === '') {
```
DorraJaouad, line 280: maybe we should follow any value (`true` or `false`) that was set in browser storage as it is user's preference.

Antreesy: `true` was a default value, so unlikely user even ticked that. If we force apply that to server, no one will get this enhancement.

Antreesy:
Putting a backport label, as server backports are not merged yet

DorraJaouad (approved):
LGTM

DorraJaouad (approved):
LGTM

nickvergessen:
Backport label is for 29?
I think server only has it in 30?

Antreesy:
Was backported for stable29: https://github.com/nextcloud/server/pull/46258

PR #12639, opened 2024-07-04
fix(CallView): freeze promoted video on hover

### LINKED ISSUE #12631: Speaker view should not change when hovering icons
### How to use GitHub

---

## Steps to reproduce
grafik

DorraJaouad:
### ☑️ Resolves

* Fix #12631

🏚️ Before (hovered)| 🏡 After (hovered)
-- | --
image | image

### 🏁 Checklist

---

### DIFF
--- a/src/components/CallView/CallView.vue
+++ b/src/components/CallView/CallView.vue
@@ -76,7 +76,8 @@
 						fit-video
 						is-big
 						:is-one-to-one="isOneToOne"
-						:is-sidebar="isSidebar" />
+						:is-sidebar="isSidebar"
+						@force-promote-video="forcePromotedModel = $event" />
 				</div>
 
 				<!-- Stripe or fullscreen grid depending on `isGrid` -->
@@ -192,12 +193,13 @@ export default {
 			},
 			showPresenterOverlay: true,
 			debounceFetchPeers: () => {},
+			forcePromotedModel: null,
 		}
 	},
 
 	computed: {
 		promotedParticipantModel() {
-			return this.callParticipantModels.find((callParticipantModel) => this.sharedDatas[callParticipantModel.attributes.peerId].promoted)
+			return this.forcePromotedModel ?? this.callParticipantModels.find((callParticipantModel) => this.sharedDatas[callParticipantModel.attributes.peerId].promoted)
 		},
 
 		callParticipantModels() {
--- a/src/components/CallView/shared/VideoBottomBar.vue
+++ b/src/components/CallView/shared/VideoBottomBar.vue
@@ -4,7 +4,10 @@
 -->
 
 <template>
-	<div class="wrapper" :class="{'wrapper--big': isBig}">
+	<div class="wrapper"
+		:class="{'wrapper--big': isBig}"
+		@mouseover.stop="mouseover = true"
+		@mouseleave.stop="mouseover = false">
 		<TransitionWrapper name="fade">
 			<div v-if="showRaiseHandIndicator" class="status-indicator raiseHandIndicator">
 				<HandBackLeft :size="18" fill-color="#ffffff" />
@@ -176,6 +179,14 @@ export default {
 		},
 	},
 
+	emits: ['bottom-bar-hover'],
+
+	data() {
+		return {
+			mouseover: false,
+		}
+	},
+
 	computed: {
 		connectionStateFailedNoRestart() {
 			return this.model.attributes.connectionState === ConnectionState.FAILED_NO_RESTART
@@ -247,6 +258,15 @@ export default {
 		},
 	},
 
+	watch: {
+		mouseover(value) {
+			if (!this.isBig) {
+				return
+			}
+			this.$emit('bottom-bar-hover', value)
+		}
+	},
+
 	methods: {
 		t,
 		forceMute() {
@@ -278,14 +298,23 @@ export default {
 	position: absolute;
 	bottom: 0;
 	width: 100%;
-	padding: 0 12px 8px 12px;
+	padding: 0 calc(var(--default-grid-baseline) * 2) var(--default-grid-baseline);
 	z-index: 1;
 
 	&--big {
 		justify-content: center;
+		margin: 0 var(--default-clickable-area); // grid collapse button
+		width: calc(100% - var(--default-clickable-area) * 2);
 		& .bottom-bar {
 			width: unset;
+			padding: calc(var(--default-grid-baseline) * 2);
+
+			&:hover {
+				background-color: rgba(0, 0, 0, 0.2);
+				border-radius: var(--border-radius-large);
+			}
 		}
+
 		& .participant-name {
 			margin-right: 0;
 		}
@@ -297,7 +326,6 @@ export default {
 	align-items: center;
 	gap: 8px;
 	width: 100%;
-	height: 44px;
 
 	& .media-indicators {
 		display: flex;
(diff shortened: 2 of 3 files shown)


src/components/CallView/shared/VideoBottomBar.vue
```diff
@@ -11,7 +11,10 @@
 			</div>
 		</TransitionWrapper>
 
-		<div v-if="!isSidebar" class="bottom-bar">
+		<div v-if="!isSidebar"
+			class="bottom-bar"
```
Antreesy, line 15: That's quite a small area to catch:
image

I thought about freezing the whole promoted VideoVue on hover. It already has a logic for that (and styles, but they're disable for `isBig`, and I'd revert it to provide feedback)

src/components/CallView/shared/VideoVue.vue
```diff
@@ -588,6 +589,10 @@ export default {
 				// Fallback for Firefox
 				?? this.$refs.video.videoWidth / this.$refs.video.videoHeight
 		},
+
+		handleHoverEvent(event) {
+			this.$emit('force-promote-video', { promoted: event, model: this.model })
+		}
```
Antreesy, line 595: Should be enough?

```suggestion
	watch: {
		mouseover(value) {
			if (!this.isBig) {
				return
			}
			this.$emit('force-promote-video', value ? this.model : null )
		}
	},
```

src/components/CallView/shared/VideoBottomBar.vue
DorraJaouad: That would be a large area from the screen and probability that the mouse is just there is high

Antreesy: So fine? If user sees that it's hovered
Or we add some complex logic with timeout, like video players have

DorraJaouad: > So fine? If user sees that it's hovered

That means, it could be easily hovered randomly (user left the mouse there) and now the actual speaker doesn't show up anymore.

src/components/CallView/shared/VideoBottomBar.vue
```diff
@@ -278,17 +298,28 @@ export default {
...
 		& .bottom-bar {
 			width: unset;
+			padding: 0 4px;
 		}
+
 		& .participant-name {
 			margin-right: 0;
 		}
+
+		&:hover {
```
Antreesy, line 317: Feels a bit counter-intuitive, so the whole line is now freezing the model, but feedback is only on smaller part.
Maybe, we keep `@mouseover` on the bottom-bar, as it was initially, and increase it's padding?

wrapper | bottombar
-- | --
image | image
image | image

src/components/CallView/shared/VideoVue.vue
```diff
@@ -588,6 +589,10 @@ export default {
 				// Fallback for Firefox
 				?? this.$refs.video.videoWidth / this.$refs.video.videoHeight
 		},
+
+		handleHoverEvent(event) {
+			this.$emit('force-promote-video', { promoted: event, model: this.model })
```
Antreesy, line 594:
```suggestion
		handleHoverEvent(value) {
			this.$emit('force-promote-video', value ? this.model : null)
```

src/components/CallView/CallView.vue
```diff
@@ -76,7 +76,8 @@
 						fit-video
 						is-big
 						:is-one-to-one="isOneToOne"
-						:is-sidebar="isSidebar" />
+						:is-sidebar="isSidebar"
+						@force-promote-video="updateForcePromoted" />
```
Antreesy, line 80:
```suggestion
						@force-promote-video="forcePromotedModel = $event" />
```

src/components/CallView/shared/VideoBottomBar.vue
```diff
@@ -247,6 +258,15 @@ export default {
 		},
 	},
 
+	watch: {
+		mouseover(value) {
+			if (!this.isBig) {
+				return
+			}
```
Antreesy, line 265: theoretically, this can be and should be checked on a VideoVue level (if we ever wanted to apply a freeze on a Stripe/Grid)

src/components/CallView/shared/VideoBottomBar.vue
```diff
@@ -278,17 +298,28 @@ export default {
...
+	padding: 0 12px 4px 12px;
 	z-index: 1;
 
 	&--big {
 		justify-content: center;
+		margin: 0 var(--default-clickable-area); // grid collapse button
+		width: calc(100% - var(--default-clickable-area) * 2);
 		& .bottom-bar {
 			width: unset;
+			padding: 0 4px;
```
Antreesy, line 310:
```suggestion
			padding: calc(var(--default-grid-baseline) * 2);
```

src/components/CallView/shared/VideoBottomBar.vue
DorraJaouad: It was indented, so to have more area to apply freeze as you mentioned it's a small area initially and user can miss it. I will revert to the initial state.

src/components/CallView/shared/VideoBottomBar.vue
DorraJaouad: I wouldn't apply the freeze too much on the call view ? It's not much needed in grid or stripe as much as in speaker view because in the latter, tile will change on each speaking signal coming in chronologically whereas in the other view, the move is only on the first speaking signal of each speaker and they remain in the same position in first page until they mute their mic. 

Also, it can happen that user lays the cursor on the bottom bar of a tile for a while and many changes took place (a lot of participants spoke and order changed). Once they move it away, the view can jump oddly.

src/components/CallView/shared/VideoBottomBar.vue
```diff
@@ -4,14 +4,18 @@
...
+		:class="{'wrapper--big': isBig}"
+		@mouseover.stop="mouseover = true"
+		@mouseleave.stop="mouseover = false">
 		<TransitionWrapper name="fade">
 			<div v-if="showRaiseHandIndicator" class="status-indicator raiseHandIndicator">
 				<HandBackLeft :size="18" fill-color="#ffffff" />
 			</div>
 		</TransitionWrapper>
 
-		<div v-if="!isSidebar" class="bottom-bar">
```
Antreesy, line 14: revert to one-line or bring back mouseover listeners here?
Still weird for me to have this 50px-high freeze line, but it's your call =)

src/components/CallView/shared/VideoBottomBar.vue
```diff
@@ -278,14 +299,23 @@ export default {
 	position: absolute;
 	bottom: 0;
 	width: 100%;
-	padding: 0 12px 8px 12px;
+	padding: 0 12px 4px 12px;
```
Antreesy, line 302: variables. and maybe reduce padding on the sides?

```suggestion
	padding: 0 calc(var(--default-grid-baseline) * 2) var(--default-grid-baseline);
```

Antreesy (approved):

src/components/CallView/shared/VideoBottomBar.vue
DorraJaouad: so you are suggesting to reduce the height here or what ? I am a bit confused 🙈

Antreesy: Just revert a change on this line

marcoambrosini (approved):

PR #12655, opened 2024-07-09
fix(directive): replace v-observe-visibility with vIntersectionObserver

Antreesy:
### ☑️ Resolves

* Drop dependency lib `vue-observe-visibility`, as it's using the same browser API
  * `vue-virtual-scroller` still has it in deps, but different version anyway

No changes

### 🚧 Tasks

- [ ] Polls might be fetched several times, from component and messageStore
  - [ ] Follow-up when fixing pollStore

### 🏁 Checklist


DorraJaouad (approved):
Tested ✅

PR #12658, opened 2024-07-09
fix: refactor messages list update in watchers

DorraJaouad:

🏚️ Before | 🏡 After
-- | --
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/ChatView.vue
+++ b/src/components/ChatView.vue
@@ -32,7 +32,7 @@
 
 		<div class="scroll-to-bottom">
 			<TransitionWrapper name="fade">
-				<NcButton v-show="!isChatScrolledToBottom"
+				<NcButton v-show="!isChatScrolledToBottom && !isLoadingChat"
 					type="secondary"
 					:aria-label="t('spreed', 'Scroll to bottom')"
 					class="scroll-to-bottom__button"
@@ -159,7 +159,11 @@ export default {
 
 		container() {
 			return this.$store.getters.getMainContainerSelector()
-		}
+		},
+
+		isLoadingChat() {
+			return !this.$store.getters.isMessageListPopulated(this.token)
+		},
 	},
 
 	watch: {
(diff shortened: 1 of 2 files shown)


src/components/MessagesList/MessagesList.vue
```diff
@@ -266,21 +266,16 @@ export default {
 			// Expire older messages when navigating to another conversation
 			this.$store.dispatch('easeMessageList', { token: oldToken })
 			this.stopFetchingOldMessages = false
-			this.messagesGroupedByDateByAuthor = this.prepareMessagesGroups(this.messagesList)
 			this.stickyDate = null
 			this.checkSticky()
```
Antreesy, line 270: Now we didn't update the messagesList from there, so should also move this?

DorraJaouad: so to check on each messagesList changes ? I don't think it's needed because scroll event is actually triggered when checkSticky needs to be called (e.g: not called on added message when div is not scrollable so there is no sticky date sep but called on added message, if message length changed etc..)

I'd keep it here.

src/components/ChatView.vue
```diff
@@ -105,7 +105,7 @@ export default {
 
 	data() {
 		return {
-			isChatScrolledToBottom: true,
+			isChatScrolledToBottom: false,
```
Antreesy, line 108: Initial loading:
image

Let's hide, if messages aren't initialized yet? or keep as `true`? it will be true anyway in most cases after chat loaded, what are exceptions?

Antreesy (commented):
Rest is good

src/components/ChatView.vue
DorraJaouad: Yeah, in most cases it is scrolled to bottom except for message hash in link or unread messages above the bottom view. Will set it to true again.

Antreesy (approved):

PR #12671, opened 2024-07-10
fix: Moderators no longer able to ban themselves or other moderators

### LINKED ISSUE #12670: fix: Moderators are able to ban themselves and other moderators
### How to use GitHub

---

### Expected behaviour
Moderators should not be able to ban themselves and other moderators

### Actual behaviour
Moderators are able to ban themselves and other moderators

skalidindi53:
### ☑️ Resolves

* Fix #12670 

### 🚧 Tasks

- [x] Moderators no longer able to ban themselves or other moderators

### 🏁 Checklist

### DIFF
--- a/tests/integration/features/bootstrap/FeatureContext.php
+++ b/tests/integration/features/bootstrap/FeatureContext.php
@@ -1838,6 +1838,8 @@ class FeatureContext implements Context, SnippetAcceptingContext {
 		$this->assertStatusCode($this->response, $statusCode);
 	}
 
+
+
 	/**
 	 * @Then /^user "([^"]*)" (promotes|demotes) "([^"]*)" in room "([^"]*)" with (\d+) \((v4)\)$/
 	 *
--- a/lib/Service/BanService.php
+++ b/lib/Service/BanService.php
@@ -9,6 +9,7 @@ declare(strict_types=1);
 namespace OCA\Talk\Service;
 
 use DateTime;
+use OCA\Talk\Manager;
 use OCA\Talk\Model\Ban;
 use OCA\Talk\Model\BanMapper;
 use OCP\AppFramework\Db\DoesNotExistException;
@@ -17,6 +18,8 @@ class BanService {
 
 	public function __construct(
 		protected BanMapper $banMapper,
+		protected Manager $manager,
+		protected ParticipantService $participantService,
 	) {
 	}
 
@@ -32,6 +35,12 @@ class BanService {
 
 		if (empty($internalNote)) {
 			throw new \InvalidArgumentException('internalNote');
+		}
+
+		$bannedParticipant = $this->participantService->getParticipantByActor($this->manager->getRoomById($roomId), $bannedType, $bannedId);
+
+		if ($bannedParticipant->hasModeratorPermissions()) {
+			throw new \InvalidArgumentException('moderator');
 		}
 
 		$ban = new Ban();
--- a/tests/integration/features/conversation-1/ban.feature
+++ b/tests/integration/features/conversation-1/ban.feature
@@ -63,27 +63,35 @@ Feature: conversation/ban
         | roomType | 3 |
         | roomName | room |
         And user "participant2" joins room "room" with 200 (v4)
-        And user "participant1" bans user "participant3" from room "room" with 200 (v1)
+        And user "participant1" bans user "participant3" from room "room" with 404 (v1)
             | internalNote | BannedInvalid |
 
-    # Scenario: Moderator trying to ban moderator
-    #     Given user "participant1" creates room "room" (v4)
-    #     | roomType | 3 |
-    #     | roomName | room |
-    #     And user "participant1" joins room "room" with 200 (v4)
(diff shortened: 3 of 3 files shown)


tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -1487,6 +1489,18 @@ public function userBansUserFromRoom(string $user, string $actorType, string $ac
...
+		if($user === $actorId) {
+			return;
+		}
+
+		//If actorId is a moderator
+		if(isset(self::$promotedParticipants[$identifier])) {
+			if($actorId === self::$promotedParticipants[$identifier]) {
+				return;
+			}
+		}
```
nickvergessen, line 1502: this must not be here. You should block this in the actual BanController/BanService

lib/Controller/BanController.php
```diff
@@ -55,6 +58,17 @@ public function banActor(string $actorType, string $actorId, string $internalNot
...
 
+			$bannedParticipant = $this->participantService->getParticipantByActor($this->room, $actorType, $actorId);
+			
+			if ($bannedParticipant->getAttendee()->getParticipantType() === Participant::MODERATOR) {
+				throw new \InvalidArgumentException('Cannot ban a moderator');
+			}
+
+			if ($actorId === $moderatorActorId && $actorType === $moderatorActorType) {
+				throw new \InvalidArgumentException('Cannot ban yourself');
+			}
```
nickvergessen, line 69: Should be in the banService with `throw new \InvalidArgumentException('yourself')` and `'moderator'`

If you would keep it here, it should not throw but return a 400 data response instead, but that would just duplicate the work and allow to make mistakes if the services is invoked in another way at some point, so it's better to have the check in the service.

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -63,6 +63,8 @@ class FeatureContext implements Context, SnippetAcceptingContext {
 	protected static array $createdTeams = [];
 	/** @var array<string, int> */
 	protected static array $userToBanId;
+	/** @var array<string, string> */
+	protected static array $promotedParticipants = []; //list to keep track of participants promoted to moderator
```
nickvergessen, line 67:
```suggestion
```

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -1838,6 +1840,9 @@ public function userAddAttendeeToRoom(string $user, string $newType, string $new
 		$this->assertStatusCode($this->response, $statusCode);
 	}
 
+
+
+
```
nickvergessen, line 1845:
```suggestion
```

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -1866,6 +1871,7 @@ public function userPromoteDemoteInRoom(string $user, string $isPromotion, strin
 			'/apps/spreed/api/' . $apiVersion . '/room/' . self::$identifierToToken[$identifier] . '/moderators',
 			new TableNode($requestParameters)
 		);
+
```
nickvergessen, line 1874:
```suggestion
```

lib/Controller/BanController.php
```diff
@@ -55,6 +58,17 @@ public function banActor(string $actorType, string $actorId, string $internalNot
...
 			$moderatorActorId = $moderator->getActorId();
 
+			$bannedParticipant = $this->participantService->getParticipantByActor($this->room, $actorType, $actorId);
+			
+			if ($bannedParticipant->getAttendee()->getParticipantType() === Participant::MODERATOR) {
+				throw new \InvalidArgumentException('Cannot ban a moderator');
+			}
+
+			if ($actorId === $moderatorActorId && $actorType === $moderatorActorType) {
+				throw new \InvalidArgumentException('Cannot ban yourself');
```
nickvergessen, line 68: Just as a note, if you check the error message in the integration test, you will notice this never triggers, as you yourself are a moderator so the previous check does it :P

lib/Controller/BanController.php
```diff
@@ -55,6 +58,17 @@ public function banActor(string $actorType, string $actorId, string $internalNot
 			$moderatorActorType = $moderator->getActorType();
 			$moderatorActorId = $moderator->getActorId();
 
+			$bannedParticipant = $this->participantService->getParticipantByActor($this->room, $actorType, $actorId);
+			
+			if ($bannedParticipant->getAttendee()->getParticipantType() === Participant::MODERATOR) {
```
nickvergessen, line 63:
```suggestion
			if ($bannedParticipant->hasModeratorPermissions()) {
```

nickvergessen (changes requested):
Otherwise it looks good :)

lib/Controller/BanController.php
skalidindi53: Yeah I also thought the same but yesterday I was able to see that it was able catch this check. so I just left it But your right today it never triggers so I just moved it up.

nickvergessen (approved):

PR #12691, opened 2024-07-12
Ban guests by ip

nickvergessen:
### ☑️ Resolves

* Fix #…

### 🚧 Tasks

- [x] Ban IP when a banned guest was detected
- [x] Allow empty note

### 🏁 Checklist

### DIFF
--- a/lib/Model/BanMapper.php
+++ b/lib/Model/BanMapper.php
@@ -39,12 +39,16 @@ class BanMapper extends QBMapper {
 		return $this->findEntity($query);
 	}
 
-	public function findByRoomId(int $roomId): array {
+	public function findByRoomId(int $roomId, ?string $bannedActorType = null): array {
 		$query = $this->db->getQueryBuilder();
 		$query->select('*')
 			->from($this->getTableName())
 			->where($query->expr()->eq('room_id', $query->createNamedParameter($roomId, IQueryBuilder::PARAM_INT)))
 			->orderBy('id', 'ASC');
+
+		if ($bannedActorType !== null) {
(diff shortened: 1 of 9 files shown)


lib/Model/BanMapper.php
```diff
@@ -28,12 +28,12 @@ public function __construct(IDBConnection $db) {
...
 	 */
-	public function findForActorAndRoom(string $actorId, string $actorType, int $roomId): Ban {
+	public function findForBannedActorAndRoom(string $actorId, string $actorType, int $roomId): Ban {
 		$query = $this->db->getQueryBuilder();
 		$query->select('*')
 			->from($this->getTableName())
-			->where($query->expr()->eq('actor_id', $query->createNamedParameter($actorId, IQueryBuilder::PARAM_STR)))
-			->andWhere($query->expr()->eq('actor_type', $query->createNamedParameter($actorType, IQueryBuilder::PARAM_STR)))
+			->where($query->expr()->eq('banned_id', $query->createNamedParameter($actorId, IQueryBuilder::PARAM_STR)))
+			->andWhere($query->expr()->eq('banned_type', $query->createNamedParameter($actorType, IQueryBuilder::PARAM_STR)))
```
SystemKeeper, line 36: Maybe it's just my personal preference, but is `banned_id` and `banned_type` really self explanatory ? Wouldn't `banned_actor_id` and `banned_actor_type` make more sense? Or too long?

nickvergessen: That would work. we are currently considering whether we should change actor_type, … to moderator_type, … as both devs made a mistake with that already.
But could also then just go for banned_actor_type, … and moderator_actor_type, …

skalidindi53 (approved):

PR #12697, opened 2024-07-12
fix(chat): keep message parents in the store after update

Antreesy:
### ☑️ Resolves

Fix lost replies
  - a fix from e815ef2dd0cd37e8fde5a3f8cff652dc1277970b works for plain values, but we don't propagate nested parents to parents
  - see backend reference:
https://github.com/nextcloud/spreed/blob/1ccbf67da231c37801c1406b10000144125aec35/lib/Controller/ChatController.php#L551-L552

🏚️ Before | 🏡 After
-- | --
reply-before | reply-after

### 🏁 Checklist

### DIFF
--- a/src/store/messagesStore.js
+++ b/src/store/messagesStore.js
@@ -290,7 +290,14 @@ const mutations = {
 		if (!state.messages[token]) {
 			Vue.set(state.messages, token, {})
 		}
-		Vue.set(state.messages[token], message.id, message)
+		// TODO split adding and updating message in the store to different actions
+		// message.parent doesn't contain grand-parent, so we should keep it
+		// when updating message in store from new message.parent object
+		const storedMessage = state.messages[token][message.id]
+		const preparedMessage = !message.parent && storedMessage?.parent
+			? { ...message, parent: storedMessage.parent }
+			: message
+		Vue.set(state.messages[token], message.id, preparedMessage)
 	},
 	/**
 	 * Deletes a message from the store.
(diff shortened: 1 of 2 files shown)


src/store/messagesStore.js
```diff
@@ -290,7 +290,14 @@ const mutations = {
...
 			Vue.set(state.messages, token, {})
 		}
-		Vue.set(state.messages[token], message.id, message)
+		// TODO split adding and updating message in the store to different actions
+		// message.parent doesn't contain grand-parent, so we should keep it
+		// when updating message in store from new message.parent object
+		const storedMessage = state.messages[token][message.id]
+		const preparedMessage = !message.parent && storedMessage?.parent
+			? { ...message, parent: storedMessage.parent }
+			: message
```
DorraJaouad, line 299: let's revert it back ? this just solves if parent existed or not, earlier it could solve any attr missing in parent of a system message.

Antreesy: Then we would have "edited" deleted message, because we didn't cleanup lastEdit properties.
Best would be to check with API team, whether we're always passing  all properties apart from parent, or not

DorraJaouad: or if it is possible to add that parent property to the message (without its grand parents if any :p)

Antreesy: Only nested parents are ignored (to prevent infinite loop, I guess) 
https://github.com/nextcloud/spreed/blob/1ccbf67da231c37801c1406b10000144125aec35/lib/Controller/ChatController.php#L551-L552
So should be good with current changes?

DorraJaouad: We won't have infinite loop, if we only add parent up to second level.

DorraJaouad (approved):

PR #12715, opened 2024-07-15
fix(federation): Fix message parameters after adjusting https handling

nickvergessen:
### ☑️ Resolves

* Follow-up 3 to #12687 

## 🛠️ API Checklist

- Mention a federated user (with https) or reply to one

### Expected
- User has a notification

### Actual
- No notification

### 🏁 Checklist

### DIFF
--- a/lib/Federation/Proxy/TalkV1/UserConverter.php
+++ b/lib/Federation/Proxy/TalkV1/UserConverter.php
@@ -38,7 +38,7 @@ class UserConverter {
 	public function convertTypeAndId(Room $room, string $type, string $id): array {
 		if ($type === Attendee::ACTOR_USERS) {
 			$type = Attendee::ACTOR_FEDERATED_USERS;
-			$id .= '@' . $room->getRemoteServer();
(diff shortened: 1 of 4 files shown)


lib/Federation/Proxy/TalkV1/UserConverter.php
```diff
@@ -195,4 +195,11 @@ protected function getLocalParticipants(Room $room): array {
 
 		return $this->participantsPerRoom[$room->getToken()];
 	}
+
+	protected function createCloudIdFromUserIdAndFullServerUrl(string $userId, string $serverUrl): string {
+		if (str_starts_with($serverUrl, 'https://')) {
+			$serverUrl = substr($serverUrl, strlen('https://'));
+		}
+		return $userId . '@' . $serverUrl;
+	}
```
ShGKme, line 204: So server URL either with HTTPS or without protocol at all, right?

ShGKme (approved):

lib/Federation/Proxy/TalkV1/UserConverter.php
nickvergessen: in the cloudId https is removed, otherwise it has to be with protocol in all cases

PR #12723, opened 2024-07-15
fix(ban): Limit query for participants when they actually can be a moderator

nickvergessen:
### 🚧 Tasks

- Allow empty internal note (its marked optional)
- Use existing room object (save a query)
- Skip moderator check when not needed
- Allow banning former and future users

### 🏁 Checklist

### DIFF
--- a/tests/integration/features/conversation-1/ban.feature
+++ b/tests/integration/features/conversation-1/ban.feature
@@ -63,7 +63,7 @@ Feature: conversation/ban
         | roomType | 3 |
         | roomName | room |
         And user "participant2" joins room "room" with 200 (v4)
-        And user "participant1" bans user "participant3" from room "room" with 404 (v1)
+        And user "participant1" bans user "participant3" from room "room" with 200 (v1)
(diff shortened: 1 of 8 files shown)


tests/integration/features/conversation-1/ban.feature
```diff
@@ -63,7 +63,7 @@ Feature: conversation/ban
         | roomType | 3 |
         | roomName | room |
         And user "participant2" joins room "room" with 200 (v4)
-        And user "participant1" bans user "participant3" from room "room" with 404 (v1)
+        And user "participant1" bans user "participant3" from room "room" with 200 (v1)
```
skalidindi53, line 66: Just have a question on why this would be a response with 200, are we allowing the moderator to ban people who are not in the room?

nickvergessen: Yeah, as the UI has sending the ban parallel to the delete.
Also at some point we might add UI to ban IPs directly and they would also never be participants

skalidindi53 (approved):

PR #12752, opened 2024-07-19
fix(sharing): Don't repopulate query builder object

nickvergessen:
### ☑️ Resolves

* Fix 
```
Using where() on non-empty WHERE part, please verify it is intentional to not call andWhere() or orWhere() instead. Otherwise consider creating a new query builder object or call resetQueryPart('where') first.
```

### 🏁 Checklist


SystemKeeper (approved):
Tested and works

PR #12753, opened 2024-07-19
fix(Quote): simplify styles for ellipsized text

Antreesy:
### ☑️ Resolves

* Fix https://github.com/nextcloud/spreed/issues/12757
* Regression from f513569945d88eee3bdaa8d743165b1038a99279
* simplify styles from 2020 👴🏽 

🏚️ Before | 🏡 After
-- | --
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/Quote.vue
+++ b/src/components/Quote.vue
@@ -37,11 +37,8 @@ components.
 				text="{file}"
 				:arguments="richParameters" />
 			<!-- Message text -->
-			<blockquote v-if="!isFileShareWithoutCaption"
-				class="quote__main__text">
-				<p dir="auto">
-					{{ shortenedQuoteMessage }}
-				</p>
+			<blockquote v-if="!isFileShareWithoutCaption" dir="auto" class="quote__main__text">
+				{{ shortenedQuoteMessage }}
 			</blockquote>
 		</div>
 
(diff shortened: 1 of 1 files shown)


ShGKme (commented):
The problem is that the text has a lot of blank lines in the beginning and in the end. And that is rendering issue. There is no extra line in the data and they are added during compile time.
```html
<p>
      {{ shortenedQuoteMessage }}
</p>
```

compiles to 
```js
"\n				" + _vm._s(_vm.shortenedQuoteMessage) + "\n			"
```

This PR fixes the issue, but then we have no ability to show 2-3 or more lines in preview. Always only one line.

src/components/Quote.vue
```diff
@@ -280,18 +275,10 @@ export default {
 
 		&__text {
 			color: var(--color-text-maxcontrast);
-			white-space: pre-wrap;
```
ShGKme, line 283: I guess that is the problem

Antreesy:
> we have no ability to show 2-3 or more lines in preview. Always only one line.

We migrated with -webkit to showing one-line anyway, so I don't know it's worth to keep this complex styles anymore

Antreesy:
Was introduced with f513569945d88eee3bdaa8d743165b1038a99279

ShGKme (approved):

PR #12756, opened 2024-07-19
fix(federation): Make sure the returned unread message count is alway…

nickvergessen:
…s an int (not null)

### ☑️ Resolves

- Empty bubble with message count

grafik

### 🏁 Checklist

### DIFF
--- a/lib/Model/Attendee.php
+++ b/lib/Model/Attendee.php
@@ -99,74 +99,29 @@ class Attendee extends Entity {
 	public const PERMISSIONS_MODIFY_REMOVE = 'remove';
 	public const PERMISSIONS_MODIFY_ADD = 'add';
 
-	/** @var int */
-	protected $roomId;
-
-	/** @var string */
-	protected $actorType;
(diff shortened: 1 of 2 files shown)


lib/Model/Attendee.php
```diff
@@ -99,74 +99,29 @@ class Attendee extends Entity {
...
+	protected int $lastMentionDirect = 0;
+	protected int $readPrivacy = 0;
+	protected int $permissions = 0;
+	protected string $accessToken = '';
+	protected string $remoteId = '';
+	protected string $invitedCloudId = '';
+	protected ?string $phoneNumber = null;
+	protected ?string $callId = null;
+	protected int $state = 0;
+	protected int $unreadMessages = 0;
```
Antreesy, line 123: Tested with this line only - works

The rest is breaking most of the things for now, even POST participants returns 500. Should also be the reason for integration tests to fail?

Antreesy (approved):

Issue #10484, opened 2023-09-09
APCu is missing on OCI/MySQL/sqlite tests


SystemKeeper:
https://github.com/nextcloud/spreed/blob/14f77fcb51485025250aa094ebccba9e8c456496/.github/workflows/integration-pgsql.yml#L131-L132

We only enabled APCu on the Postgres tests. Not sure if this is an oversight or if there was a reason for not doing it on all integration tests?

PR #12758, opened 2024-07-19
fix(ci): Use apcu in all integration tests

SystemKeeper:
## 🛠️ API Checklist

- Fixes https://github.com/nextcloud/spreed/issues/10484

### 🏁 Checklist


nickvergessen (approved):
Doesn't really matter as they only run when touching the files (was about to add migration files there as well)

Issue #10486, opened 2023-09-10
Show error if recording server is configured, but no signaling server


SystemKeeper:
https://github.com/nextcloud/spreed/blob/14f77fcb51485025250aa094ebccba9e8c456496/src/components/AdminSettings/RecordingServers.vue#L30-L32

Like the upload warning, we should show a warning in case a recording server is configured, but no signaling server.

nickvergessen:
Should be an error, because it actually doesn't even work?

PR #12771, opened 2024-07-21
fix(AdminSettings): Require HPB for recording server configuration

SystemKeeper:
### ☑️ Resolves

* Fixes https://github.com/nextcloud/spreed/issues/10486
Same as we currently do for the SIP configuration: https://github.com/nextcloud/spreed/blob/main/src/components/AdminSettings/SIPBridge.vue.
* Add emitter to promprtly update all sections (see screencast)

Changes are best reviewed as https://github.com/nextcloud/spreed/pull/12771/files?diff=unified&w=1

🏚️ Before | 🏡 After
-- | --
Bildschirmfoto 2024-07-21 um 10 30 06 | Bildschirmfoto 2024-07-21 um 10 30 18

[Screencast from 22.07.2024 20:44:11.webm](https://github.com/user-attachments/assets/99d619f6-793d-4582-b2bd-0dbe28008a9f)

### 🏁 Checklist

### DIFF
--- a/src/components/AdminSettings/RecordingServers.vue
+++ b/src/components/AdminSettings/RecordingServers.vue
@@ -98,6 +98,7 @@ import TransitionWrapper from '../UIShared/TransitionWrapper.vue'
 
 import { CALL } from '../../constants.js'
 import { hasTalkFeature } from '../../services/CapabilitiesManager.ts'
+import { EventBus } from '../../services/EventBus.js'
 
 const recordingConsentCapability = hasTalkFeature('local', 'recording-consent')
 const recordingConsentOptions = [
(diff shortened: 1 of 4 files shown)


src/components/AdminSettings/RecordingServers.vue
```diff
@@ -150,6 +157,11 @@ export default {
 		this.uploadLimit = parseInt(state.uploadLimit, 10)
 	},
 
+	mounted() {
```
Antreesy, line 160: We need to provide an updated value, if we add/remove a signaling server in another component. Otherwise an admin should reload the page to unlock recording =)
I could take it from here, if you want

SystemKeeper: Yes, see my comment :-) 
Same in SIP currently. I would be happy if you can take it from here, as I fail to see an easy way to do that (without creating a store or using props in the parent component).

Antreesy (approved):

PR #12778, opened 2024-07-22
chore: Used the getRelativePath function to correctly get the path

### LINKED ISSUE #12676: Use `IUserFolder::getRelativePath($fullPath)`
> `$userFolder->getRelativePath($fullPath)` is the "correct" way to do this.

_Originally posted by @icewind1991 in https://github.com/nextcloud/spreed/pull/12554#discussion_r1648579142_

skalidindi53:
### ☑️ Resolves

* Fix #12676 

### 🚧 Tasks

- [x] Used getRelativePath function to properly/correctly get the path

### 🏁 Checklist

### DIFF
--- a/lib/Chat/Parser/SystemMessage.php
+++ b/lib/Chat/Parser/SystemMessage.php
@@ -763,10 +763,9 @@ class SystemMessage implements IEventListener {
 				}
 
 				$fullPath = $node->getPath();
-				$pathSegments = explode('/', $fullPath, 4);
 				$name = $node->getName();
 				$size = $node->getSize();
-				$path = $pathSegments[3] ?? $name;
+				$path = "Shared/" . ($userFolder->getRelativePath($fullPath) ?? $name);
 			} else {
 				$node = $share->getNode();
 				$name = $node->getName();


lib/Chat/Parser/SystemMessage.php
```diff
@@ -763,18 +763,15 @@ protected function getFileFromShare(Room $room, ?Participant $participant, strin
...
 			} else {
 				$node = $share->getNode();
 				$name = $node->getName();
 				$size = $node->getSize();
 
 				$fullPath = $node->getPath();
-				$pathSegments = explode('/', $fullPath, 4);
-				$path = $pathSegments[3] ?? $name;
+				$userFolder = $this->rootFolder->getUserFolder($participant->getAttendee()->getActorId());
+				$userFolder->getRelativePath($fullPath);
```
skalidindi53, line 774: Just added it here or should I only have one instance of $userFolder, initialized after the first if statement

lib/Chat/Parser/SystemMessage.php
```diff
@@ -763,10 +763,9 @@ protected function getFileFromShare(Room $room, ?Participant $participant, strin
 				}
 
 				$fullPath = $node->getPath();
-				$pathSegments = explode('/', $fullPath, 4);
 				$name = $node->getName();
 				$size = $node->getSize();
-				$path = $pathSegments[3] ?? $name;
+				$path = "Shared/" . ($userFolder->getRelativePath($fullPath) ?? $name);
```
SystemKeeper, line 768: Where does the `Shared/` come from?

skalidindi53: Yeah I don't think it's right, it's hardcoded. Not sure why but getRelatviePath is getting null here and path is getting set to "Shared/" + name. I was trying to figure out what I'm missing Friday but wasn't successful.

SystemKeeper: What is `$fullPath` and `$userFolder->getPath()` in your case?

This is what happens on a quick test for me:
```
$fullPath = "/admin/files/Talk/IMG_0005.jpg"
$userFolder->getRelativePath($fullPath) = "/Talk/IMG_0005.jpg"
```

which is what I would expected here (as `$pathSegments[3]` was also "/Talk/IMG_0005.jpg" in my case).

skalidindi53: This is the test case that it's failing for me:
`$fullPath = "/user/files/Shared/different"
$userFolder->getRelativePath($fullPath) = "NULL"`

The path is supposed to "Shared/different" not NULL

nickvergessen: This "cleanup" is killing way too much time. We'll just close the ticket.

PR #12780, opened 2024-07-23
fix(openapi): Update openapi extractor to latest

nickvergessen:


nickvergessen:
@Antreesy it uses requestBody now instead of query parameters. Can you check it doesn't break the frontend?

nickvergessen:
Let's wait for https://github.com/nextcloud/spreed/pull/12618 to go in first to avoid rebasing with the remote repository which is a bit of pain

Antreesy (approved):

PR #12783, opened 2024-07-23
feat(openapi): Add more type/int limitations

nickvergessen:
- Would something like this be desired @provokateurin 
Is a break like this okay?
- Is it a good/okay idea to ease reusing as a API developer to introduce a subtype for things like CALL_FLAGS or is that confusing?


provokateurin (approved):
This is a nice improvement, everything should be documented as strictly as possible.
It's not even a breaking change, since the API is the same and only the documentation for it has changed. To the end user this is not a problem, except for the case where they generate code from it, but then they have to deal with it themselves and it doesn't concern these changes.

provokateurin:
Also adding extra type aliases is helpful, otherwise code generators will need to come up with names and they can be ugly (and ofc then they are re-usable which also helps a lot).

nickvergessen:
Okay thanks, then I will check if there are more things to adjust and then start fixing false-negative psalm complaints

provokateurin:
The psalm complaints might be legit though

nickvergessen:
No, we just don't have get*() declare that they return the consts…
So yeah "legit" by psalm, but not "broken code".

nickvergessen:
Pushed the current state, but https://github.com/nextcloud/spreed/blob/1ccbf67da231c37801c1406b10000144125aec35/lib/Model/BotServer.php#L15-L18 is a bit annoying. It does not allow to specify `non-empty-string` and therefore psalm afterwards complains. Similar problem with all int fields

lib/ResponseDefinitions.php
```diff
@@ -164,140 +170,140 @@
...
- *     participantFlags: int,
- *     participantType: int,
- *     permissions: int,
- *     readOnly: int,
- *     recordingConsent: int,
+ *     participantFlags: TalkCallFlags,
+ *     participantType: TalkParticipantTypes,
+ *     permissions: TalkPermissions,
+ *     readOnly: 0|1,
+ *     recordingConsent: 0|1|2,
```
nickvergessen, line 267:
```suggestion
 *     recordingConsent: 0|1,
```

lib/Model/Attendee.php
```diff
@@ -59,7 +59,8 @@
  * @method int getLastAttendeeActivity()
  */
 class Attendee extends Entity {
-	public const ACTOR_USERS = 'users';
+	public const ACTOR_USERS = 'users';lib/Chat/SystemMessage/Listener.php
```
Antreesy, line 62: Should be like this?

nickvergessen: No, it's copy paste fail, but psalm is complaining in this PR anyway and fixing it is only possible with a lot of duplicated docs

Issue #12800, opened 2024-07-24
Adjust UI according to the design review


ShGKme:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Conversation and composing

* [x] Top bar title is cut off @Antreesy 
  https://github.com/nextcloud/spreed/pull/12851
  https://github.com/nextcloud/spreed/pull/12875
* [x] Date marker "Today, July 23, 2024" border radius should be reduced to make it similar to the button @Antreesy 
  - https://github.com/nextcloud/spreed/pull/12810/

## List view

* [ ] ~~#13923~~
* [x] Margin on the left of list item is too much and on the right is too little --> it should nicely aligned with the left of the search bar and right aligned with the new conversation button (upstream issue?) @Antreesy 
  - https://github.com/nextcloud/spreed/pull/12810/
* [ ] ~~Settings button should be left aligned and not bold like in Files (visually)~~
  - https://github.com/nextcloud-libraries/nextcloud-vue/pull/5866
* [x] Counter bubble should have min-width such that width is never lesser than height (even when counter bubble has "1" it should be a circle) @ShGKme 
  - https://github.com/nextcloud-libraries/nextcloud-vue/pull/5905

## Sidebar

* [x] Participant list: @Antreesy 
  - Issues:
    - [x] Too much padding on left and right of each participant is too much
    - [x] Vertical space can also be reduced
    - [x] border radius should be reduced
    - [x] avatar focus should be circle
  - Migrate to `NcListItem`? https://github.com/nextcloud-libraries/nextcloud-vue/pull/5870
* [x] Media tab: thumbnails are top left aligned their containers @Antreesy https://github.com/nextcloud/spreed/pull/12810/

## Calling

* [x] Media settings:
  - [x] border radius on bg should be consistent
    - @Antreesy https://github.com/nextcloud/spreed/pull/12810/
  - [x] "Test devices" button should be left aligned with select components
    - @Antreesy https://github.com/nextcloud/spreed/pull/12810/
  - [x] Height changes when you change between devices and background tabs --> render both tabs and make one not visible @ShGKme 
    - https://github.com/nextcloud/spreed/pull/12815
* [ ] Ongoing call:
[...]

PR #12810, opened 2024-07-25
fix(design): adjustments for Nextcloud 30 - v3

Antreesy:
### ☑️ Resolves

* Ref #12800

🏚️ Before | 🏡 After
-- | --
image | image
image | image
image | image
image | image
image | image

## Conversation and composing

* [ ] Top bar title is cut off - dropped for now
* [x] Date marker "Today, July 23, 2024" border radius should be reduced to make it similar to the button

## List view

* [x] Margin on the left of list item is too much and on the right is too little --> it should nicely aligned with the left of the search bar and right aligned with the new conversation button (upstream issue?)

## Sidebar

* [x] Media tab: thumbnails are top left aligned their containers

## Calling

* [x] Media settings:
  - [x] border radius on bg should be consistent
  - [x] "Test devices" button should be left aligned with select components

### 🏁 Checklist

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -180,7 +180,7 @@
 					data-nav-id="conversation_create_new"
 					@click="createConversation(searchText)">
 					<template #icon>
-						<ChatPlus :size="44" />
+						<ChatPlus :size="AVATAR.SIZE.TWO_LINES" />
 					</template>
 					<template #subname>
 						{{ t('spreed', 'New group conversation') }}
@@ -206,7 +206,7 @@
 						:name="item.label"
 						@click="createAndJoinConversation(item)">
 						<template #icon>
-							<AvatarWrapper v-bind="iconData(item)" />
+							<AvatarWrapper v-bind="iconData(item)" :size="AVATAR.SIZE.TWO_LINES" />
 						</template>
 						<template #subname>
 							{{ t('spreed', 'New private conversation') }}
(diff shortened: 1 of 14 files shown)


src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -1086,7 +1086,9 @@ export default {
 	padding: 0 !important;
 }
 
+// Overwrite NcListItem styles to reduce a component height
 :deep(.list-item) {
+	padding: var(--default-grid-baseline) calc(var(--default-grid-baseline) * 2);
```
ShGKme, line 1091: Issue on upstream?

src/components/MediaSettings/MediaDevicesSelector.vue
ShGKme, line 1: Let's also make the refresh button `tertiary` here? Same problem as on AdminSettings page

src/components/LeftSidebar/LeftSidebar.vue
Antreesy: I'm not sure if it wasn't an intended design change for compatibility, but can upstream of course

src/components/TopBar/TopBar.vue
```diff
@@ -347,6 +347,7 @@ export default {
 		width: 100%;
 		overflow: hidden;
 		min-height: var(--default-clickable-area);
+		font-size: 13px;
```
ShGKme, line 350: This makes the text smaller than default even when there is only a title without a description.
And there is some additional space if we reduce the padding.

@jancborchardt What do you think is better here?

Also, we can stop showing the description here and show it in the sidebar. It is one line for a long (potentially rich) text anyway.

src/components/LeftSidebar/LeftSidebar.vue
ShGKme: I'd say, if Talk needs non-standard list item design now, then something is wrong.

src/components/TopBar/TopBar.vue
Antreesy: Reverted changes, will provide this suggestion as a separate PR

ShGKme (approved):
Looks good everywhere, fixes all the issues

PR #12815, opened 2024-07-25
fix(MediaSettings): jumping on devices/backgrounds switch

ShGKme:
### ☑️ Resolves

* A part of: https://github.com/nextcloud/spreed/issues/12800

🏚️ Before | 🏡 After
-- | --
before-media | after-media

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist


ShGKme:
@jancborchardt 

We can easily remove the jumping when switching between tabs. But there is still a jump on opening a tab for the first time.

What do you think?

jancborchardt (commented):
> We can easily remove the jumping when switching between tabs.

Nice, looks great! :)

> But there is still a jump on opening a tab for the first time.

Can we have a quick vertical size increase / decrease animation on opening the tab initially (and closing it as well)? We can also use either --animation-slow or --animation-quick for that, whatever looks nicer.

jancborchardt:
And btw, the first time you switch to the "Backgrounds" tab, there is no left/right animation. Is that an artifact in the screen recording or a bug?

ShGKme:
> Can we have a quick vertical size increase / decrease animation on opening the tab initially (and closing it as well)?

Sure

> And btw, the first time you switch to the "Backgrounds" tab, there is no left/right animation. Is that an artifact in the screen recording or a bug?

If the switch happens too fast after opening the first tab, there is a small glitch — a browser needs some time to load and render quite huge images... Will have a look if we can prevent it.

Also, on gif it's a bit more noticeable than in reality

ShGKme:
@jancborchardt Updated gif. What do you think?

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -74,46 +74,28 @@
 				</div>
 			</div>
 
-			<!-- Tabs -->
-			<div class="media-settings__call-preferences">
```
Antreesy, line 78: do not forget to remove style leftovers

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -310,13 +306,15 @@ export default {
 			stopDevices,
 			virtualBackground,
 			model: localMediaModel,
+			tabs,
+			isVirtualBackgroundAvailable,
```
Antreesy, line 310: we don't need it outside of setup, right?

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -310,13 +306,15 @@ export default {
...
 		}
 	},
 
 	data() {
 		return {
 			modal: false,
-			tabContent: 'none',
+			tabContent: undefined,
```
Antreesy, line 317: We got a browser storage check in showModal() method to automatically open devices tab. Doesn't seem to work now:

```js
	if (!BrowserStorage.getItem('audioInputDevicePreferred') || !BrowserStorage.getItem('videoInputDevicePreferred')) {
		this.tabContent = 'devices'
	}
```

src/components/MediaSettings/MediaSettingsTabs.vue
```diff
@@ -0,0 +1,139 @@
...
+    tabs: TabDefinition[],
+    active?: string,
+}>()
+
+const emit = defineEmits<{
+    (event: 'update:active', value?: string): void
+}>()
+
+/** Whether the tab panel is open */
+const isOpen = ref(false)
```
Antreesy, line 30: That solves auto-opening, please check it doesn't break anything else =)
```suggestion
const isOpen = ref(!!props.active)
```

src/components/MediaSettings/MediaSettings.vue
ShGKme: No, can be removed

ShGKme: If it was about `isVirtualBackgroundAvailable`

src/components/MediaSettings/TransitionExpandDown.vue
```diff
@@ -0,0 +1,49 @@
+<!--
+  - SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<script setup lang="ts">
```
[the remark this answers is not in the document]
ShGKme: CI is green... What doesn't work?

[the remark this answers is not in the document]
Antreesy: All good, just pointing out we aren't handling .vue files with tsconfig

src/components/MediaSettings/MediaSettings.vue
ShGKme: Fixed

src/components/MediaSettings/MediaSettings.vue
ShGKme: Fixed

src/components/MediaSettings/MediaSettings.vue
ShGKme: Fixed

src/components/MediaSettings/MediaSettingsTabs.vue
ShGKme: Added support for open state change from parent via falsy `active` (2-ways)

src/components/MediaSettings/TransitionExpandDown.vue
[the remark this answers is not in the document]
ShGKme: Can have a look at it a separated PR

Antreesy (approved):
Looks good!

ShGKme:
Rebased and squashed

jancborchardt (commented):
@ShGKme this looks so damns smooth now, awesome work! :astonished:

PR #12819, opened 2024-07-25
fix(Participant): migrate to NcListItem

Antreesy:
### ☑️ Resolves

* Ref #12800 
* Requires https://github.com/nextcloud-libraries/nextcloud-vue/pull/5870

🏚️ Before | 🏡 After
-- | --
image | image
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/RightSidebar/Participants/Participant.vue
+++ b/src/components/RightSidebar/Participants/Participant.vue
@@ -4,36 +4,33 @@
 -->
 
 <template>
-	<component :is="tag"
+	<NcListItem :name="computedName"
 		:data-nav-id="participantNavigationId"
 		class="participant"
 		:class="{
 			'participant--offline': isOffline,
-			'participant--searched': isSearched,
-			'participant--selected': isSelected }"
+			'participant--searched': isSearched }"
+		:active="isSelected"
 		:aria-label="participantAriaLabel"
-		:role="isSearched ? 'listitem' : undefined"
-		:tabindex="0"
+		:actions-aria-label="participantSettingsAriaLabel"
+		force-display-actions
 		@click="handleClick"
 		@keydown.enter="handleClick">
 		<!-- Participant's avatar -->
-		<AvatarWrapper :id="computedId"
-			:token="isSearched ? 'new' : token"
-			:name="computedName"
-			:source="participant.source || participant.actorType"
-			:disable-menu="isSearched"
-			disable-tooltip
-			:show-user-status="showUserStatus"
-			:preloaded-user-status="preloadedUserStatus"
-			:highlighted="isSpeakingStatusAvailable && isParticipantSpeaking"
-			:offline="isOffline" />
+		<template #icon>
+			<AvatarWrapper :id="computedId"
+				:token="isSearched ? 'new' : token"
+				:name="computedName"
+				:source="participant.source || participant.actorType"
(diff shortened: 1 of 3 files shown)


nickvergessen:
That's quite "in the face" for a "list of selectable items"...

I had already mentioned back then when the components where created that it might not be the best idea to combine the items in the right sidebar with the attendee picking in the convo creation dialog.
I'd still be fine with splitting that. They operate on total different data set, have different data and are inside different context.

Antreesy:
> I'd still be fine with splitting that

Could be a good first step, before upstream PR is here. Search result is simple enough to migrate it right now

src/components/RightSidebar/Participants/Participant.vue
```diff
@@ -4,154 +4,128 @@
 -->
 
 <template>
-	<component :is="tag"
+	<NcListItem :name="computedName"
```
DorraJaouad, line 7:
```suggestion
	<NcListItem 
```

as you are using name slot

src/components/RightSidebar/Participants/Participant.vue
```diff
@@ -1176,110 +1124,84 @@ export default {
...
+	}
+
+	&--offline &__user-name {
+		color: var(--color-text-maxcontrast);
+	}
 
-	&.isSearched {
+	&--searched :deep(.list-item) {
 		cursor: pointer;
 	}
```
DorraJaouad, line 1146: It is not applied and participant is now selectable in list, is it expected ?
participants

src/components/RightSidebar/Participants/Participant.vue
```diff
@@ -1176,110 +1124,84 @@ export default {
...
-			padding-right: 5px;
-			/** call icon on the right most column */
-			width: calc(100% - 90px);
-		}
+			&:hover,
+			&:focus-within,
+			&:has(:focus-visible),
+			&:has(:active) {
+				background-color: var(--color-primary-element-light-hover);
+			}
```
DorraJaouad, line 1158: As shown above, I am not sure if it is expected to have background-color on click focus, it seems like you can do some action on it from a header menu

DorraJaouad (commented):
It is almost ready 🌟

src/components/RightSidebar/Participants/Participant.vue
Antreesy: Prop is required, so test will scream about it

src/components/RightSidebar/Participants/Participant.vue
Antreesy: styles fixed.
It's expected as it is interactive element `<a/>`. we have it before anyway with `tabindex="0"`, just without visual feedback

src/components/RightSidebar/Participants/Participant.vue
ShGKme: Better to be fixed in upstream. CUrrently `NcListItem` always supposes item to be a link.

ShGKme:
Follow-ups:
- With additional padding it is not inline with search
  image
- Structure is not a valid HTML
  image
- Placeholder is different from list item (an old issue? cc @ShGKme)
  image
- Old a11y issue - on creation Participant represent a checkbox but it is a link
  image

PR #12801, opened 2024-07-24
fix: Banned user cannot be added to the room

skalidindi53:
### ☑️ Resolves

* Fix #12793 

### 🚧 Tasks

- [x] Banned user cannot be added to the room

### 🏁 Checklist

- [x] ⛑️ Tests (unit and/or integration) are included or not possible
- [ ] 📘 API documentation in `docs/` has been updated or is not required
- [ ] 🔖 Capability is added or not needed

PR #12821, opened 2024-07-25
fix: When banning a user they are also removed from the room

### LINKED ISSUE #12804: Banning should remove the participant server-side
Since banning is now a dedicated endpoint and not tied to removing the participant, we should automatically remove the participant from the conversation when the participant was banned. Currently these are separate 2 requests done by the clients.

skalidindi53:
### ☑️ Resolves

* Fix #12804 

### 🚧 Tasks

- [x] Removes users from the room when they are banned

### 🏁 Checklist

### DIFF
--- a/lib/Service/BanService.php
+++ b/lib/Service/BanService.php
@@ -9,6 +9,7 @@ declare(strict_types=1);
 namespace OCA\Talk\Service;
 
 use DateTime;
+use OCA\Talk\Events\AAttendeeRemovedEvent;
 use OCA\Talk\Exceptions\ForbiddenException;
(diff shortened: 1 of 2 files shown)


nickvergessen (commented):
Should then have a test what happens when adding someone after they get banned.

skalidindi53:
> Should then have a test what happens when adding someone after they get banned.

I do have that test in PR #12801 but the response code might change  from 403 to 404 now that we also just remove the participant from the room right away. Could you approve/review that PR first? I'll rebase this branch to main and makes changes as needed

nickvergessen (approved):

PR #12823, opened 2024-07-26
fix(MessageList): rewrite CSS layout of chat

Antreesy:
### ☑️ Resolves

* Ref #12800

🏚️ Before | 🏡 After |
-- | -- |
image | image
image | image

> [!TIP]
> For visual diff use https://www.diffchecker.com/image-compare/

### 🏁 Checklist

### DIFF
--- a/src/assets/variables.scss
+++ b/src/assets/variables.scss
@@ -4,14 +4,17 @@
  */
 
 /** Messages list dimensions:
- * - text max width: 70em (recommended by W3C standard) = 1050px
- * - utils-width: 52px (group avatar with paddings) + 120px (date and checkmark) + 4*4px (date paddings) = 188px
- * - list max width: text width + utils width + 4 + 8 (paddings) = 1250px
+ * - text max width: 1050px (70em recommended by W3C standard)
+ * - avatar width: 32px (AVATAR.SIZE.SMALL) + 16px (paddings) = 48px
+ * - info width: 8ch(~68px) (timestamp) + 40px (checkmark with paddings) = ~108px
+ * - list max width: 48px (avatar width) + 1058px (text width with paddings) + ~108px (info width) = ~1214px
+ * - input max width: ~1214px (list max width) - 100px (send button) = ~1114px
  */
 $messages-text-max-width: calc(70 * var(--default-font-size));
-$message-utils-width: calc(52px + 4 * var(--default-grid-baseline) + 120px);
-$messages-list-max-width: calc($messages-text-max-width + $message-utils-width + 3 * var(--default-grid-baseline));
-$message-input-max-width: calc($messages-list-max-width - 100px);
+$messages-avatar-width: calc(32px + 4 * var(--default-grid-baseline));
+$messages-info-width: calc(8ch + var(--clickable-area-small, 24px) + 4 * var(--default-grid-baseline));
+$messages-list-max-width: calc($messages-avatar-width + $messages-text-max-width + 2 * var(--default-grid-baseline) + $messages-info-width);
+$messages-input-max-width: calc($messages-list-max-width - 100px);
 
 // background color of call container
 $color-call-background: rgba(34, 34, 34, 0.8);
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/MessageBody.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/MessageBody.vue
@@ -69,7 +69,7 @@
 
 		<!-- Additional message info-->
 		<div v-if="!isDeletedMessage" class="message-main__info">
-			<span :class="['date', { 'date--hidden': hideDate }]" :title="messageDate">{{ messageTime }}</span>
+			<span class="date" :class="{ 'date--hidden': hideDate }" :title="messageDate">{{ messageTime }}</span>
 
 			<!-- Message delivery status indicators -->
 			<div v-if="message.sendingFailure"
@@ -83,6 +83,7 @@
 				@mouseleave="showReloadButton = false"
 				@blur="showReloadButton = false">
 				<NcButton v-if="sendingErrorCanRetry && showReloadButton"
+					size="small"
 					:aria-label="sendingErrorIconTooltip"
 					@click="handleRetry">
 					<template #icon>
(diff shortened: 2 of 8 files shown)


nimishavijay (commented):
Nice! Looks pretty compact. I'd suggest that these changes be implemented when there's a lack of horizontal space, since we discussed this mostly in the context of the chat during calls. 

How about these values are enforced only when the width of the screen falls below the max line width (which I believe is around 700px)?

Antreesy:
> only when the width of the screen falls below the max line width

Some general rework was needed here anyway. Applied reduced padding for sidebar chats only (that is Files integration and during calls)

ShGKme:
With read marker in Sidebar looks too close.

All the other places look good, but I'd also test on c.nc.nc via the desktop client to have a look at more real word chats.

image

nickvergessen:
> Applied reduced padding for sidebar chats only (that is Files integration and during calls)

Also test the "video verification"

Antreesy:
> Also test the "video verification"

It's broken in another way.
- Fix is here: https://github.com/nextcloud/spreed/pull/12853
- Added screenshots with new chat layout

src/components/MessagesList/MessagesGroup/Message/MessagePart/MessageBody.vue
```diff
@@ -447,16 +449,14 @@ export default {
 @import '../../../../../assets/variables';
 
 .message-main {
-	display: flex;
+	display: grid;
+	grid-template-columns: minmax(0, $messages-text-max-width) 1fr;
 	justify-content: space-between;
```
DorraJaouad, line 454: It's not consistent for all messages
image

src/components/MessagesList/MessagesList.vue
```diff
@@ -21,11 +21,13 @@
...
 			:ref="`dateGroup-${token}`"
 			:data-date-timestamp="dateTimestamp"
-			:class="{'has-sticky': dateTimestamp === stickyDate}">
+			class="scroller__content"
+			:class="{ 'has-sticky': dateTimestamp === stickyDate }">
 			<li class="messages-group__date">
 				<span class="messages-group__date-text" role="heading" aria-level="3">
 					{{ dateSeparatorLabels[dateTimestamp] }}
 				</span>
+				<span class="messages-group__date-hidden">{{ getHiddenTime(dateTimestamp) }}</span>
```
DorraJaouad, line 30: Just a waste of resources I'd say :p. You can set to static time or string if possible as it will always be 12:00 or we use margin.

src/components/MessagesList/MessagesGroup/Message/MessagePart/MessageBody.vue
```diff
@@ -524,15 +524,15 @@ export default {
...
 	}
 }
 
 .message-status {
-	width: var(--default-clickable-area);
-	height: 24px;
+	width: $message-status-icon-width;
+	height: var(--default-line-height);
 	display: flex;
 	justify-content: center;
```
Antreesy, line 537: For @ShGKme comment:
```suggestion
	justify-content: flex-end;
```
B | A
-- | --
image | image

src/components/MessagesList/MessagesGroup/Message/MessagePart/MessageBody.vue
```diff
@@ -83,6 +83,8 @@
 				@mouseleave="showReloadButton = false"
 				@blur="showReloadButton = false">
 				<NcButton v-if="sendingErrorCanRetry && showReloadButton"
+					type="tertiary"
+					size="small"
```
DorraJaouad, line 87: Its height has extra px to the message height

https://github.com/user-attachments/assets/0dc8d555-d714-482a-985f-4a9d40c5a935

src/components/MessagesList/MessagesGroup/MessagesGroup.vue
```diff
@@ -134,24 +134,24 @@ export default {
...
 	width: 100%;
 	min-width: 0;
 
 	&__avatar {
 		position: sticky;
 		top: 0;
-		height: 52px;
-		width: 52px;
-		padding: 18px 10px 10px 10px;
+		padding-inline: var(--chat-spacing, 8px);
```
DorraJaouad, line 145: IMO it would be better if it's 10px at least. It looks really close to the edge in main view.

ShGKme: Agree, also looks too close for me in the main window on a small screen.

Maybe even 12px to make it on the grid

src/components/MessagesList/MessagesGroup/Message/MessagePart/MessageBody.vue
Antreesy: adressed with 8ch for now

src/components/MessagesList/MessagesGroup/MessagesGroup.vue
Antreesy: keep it as 8px, so 8+32+8 for the complete width. Does 2px to the edge reaaly make that difference?

Antreesy:
Dropped most of the conditional size changes to establish correct grid layout first. Let's discuss this part, then continue with call view

src/components/MessagesList/MessagesGroup/Message/MessagePart/MessageBody.vue
```diff
@@ -531,8 +529,8 @@ export default {
 }
 
 .message-status {
-	width: var(--default-clickable-area);
-	height: 24px;
+	width: 24px;
```
DorraJaouad, line 532:
```suggestion
	width: var(--clickable-area-small);
```

src/assets/variables.scss
```diff
@@ -9,8 +9,9 @@
  * - list max width: text width + utils width + 4 + 8 (paddings) = 1250px
  */
 $messages-text-max-width: calc(70 * var(--default-font-size));
-$message-utils-width: calc(52px + 4 * var(--default-grid-baseline) + 120px);
-$messages-list-max-width: calc($messages-text-max-width + $message-utils-width + 3 * var(--default-grid-baseline));
+$message-avatar-width: calc(32px + 4 * var(--default-grid-baseline));
```
DorraJaouad, line 12:
```suggestion
$message-avatar-width: calc(var(--avatar-size) + 4 * var(--default-grid-baseline)); 
```

Antreesy: We don't have that in CSS.. But I'll leave a comment in variables stylesheet

DorraJaouad (approved):
Good to go 🦅

PR #12651, opened 2024-07-08
fix(chat): wrap Quote in RouterLink to handle in-app navigation

Antreesy:
### ☑️ Resolves

* Fix faulty dialog trigger when navigating to quoted message in the same chat
  *  Replace `<a>` with `<RouterLink>`

### 🖼️ Screenshots / Screencasts

🏚️ Before 

[Screencast from 08.07.2024 17:06:36.webm](https://github.com/nextcloud/spreed/assets/93392545/ba95b2bf-f535-4f17-8c18-4571ecb78992)

 🏡 After

[Screencast from 08.07.2024 17:03:48.webm](https://github.com/nextcloud/spreed/assets/93392545/7a111dc0-8de7-46b6-9e85-585d2cacebc6)

### 🏁 Checklist

- [x] 🌏 Tested with Chrome, Firefox and Safari or should not be risky to browser differences
- [x] 🖥️ Tested with Desktop client or should not be risky for it 
- [x] 🖌️ Design was reviewed, approved or inspired by the design team
- [ ] ⛑️ Tests are included or not possible
- [ ] 📗 User documentation in https://github.com/nextcloud/documentation/tree/master/user_manual/talk has been updated or is not required

PR #12824, opened 2024-07-26
fix(Quote): prevent button click event from propagation

Antreesy:
### ☑️ Resolves

* Follow up to #12651
* button is wrapped in `<a>` link, so bubbling click event doesn't get captured by router-link, and causes watning dialog to be shown (my humble assumption)

🏚️ Before | 🏡 After
-- | --
image | See URL hash is unchanged

### 🏁 Checklist

### DIFF
--- a/src/components/Quote.vue
+++ b/src/components/Quote.vue
@@ -9,7 +9,8 @@ components.
 </docs>
 
 <template>
-	<router-link :to="{ hash, params: { skipLeaveWarning: true } }"
+	<component :is="component.tag"
+		:to="component.link"
 		class="quote"
 		:class="{'quote-own-message': isOwnMessageQuoted}"
 		@click.native="handleQuoteClick">
@@ -45,13 +46,14 @@ components.
 		<NcButton v-if="canCancel"
 			class="quote__close"
 			type="tertiary"
+			:title="cancelQuoteLabel"
 			:aria-label="cancelQuoteLabel"
 			@click="handleAbort">
 			<template #icon>
 				<Close :size="20" />
 			</template>
 		</NcButton>
-	</router-link>
+	</component>
 </template>
 
 <script>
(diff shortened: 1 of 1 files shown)


ShGKme (approved):
Should fix the problem, but having an interactive element (button) inside another interactive element is invalid and an a11y issue.

Antreesy:
> is invalid and an a11y issue.

we have that only in one place, and Quote is used there only as a visual element, so, could be altered with `<component :is`

src/components/Quote.vue
```diff
@@ -9,9 +9,9 @@ components.
 </docs>
 
 <template>
-	<router-link :to="{ hash, params: { skipLeaveWarning: true } }"
-		class="quote"
-		:class="{'quote-own-message': isOwnMessageQuoted}"
+	<component :is="component.tag"
+		:to="component.link"
+		:class="['quote', { 'quote-own-message': isOwnMessageQuoted }]"
```
ShGKme, line 14: To separate static actual element classes and dynamically declared. (Saw it in many places recently)

```suggestion
		class="quote"
		:class="{ 'quote-own-message': isOwnMessageQuoted }"
```

src/components/Quote.vue
```diff
@@ -127,6 +128,12 @@ export default {
 	},
 
 	computed: {
+		component() {
+			return this.canCancel || this.editMessage
+				? { tag: 'div', link: undefined }
```
ShGKme, line 133: Why use div on edit? Seems useful to navigate to the message being edited.

src/components/Quote.vue
Antreesy: I found it more readable 🙈 
Reverted

src/components/Quote.vue
Antreesy: You're right, edit mode doesn't have interactive elements inside

PR #12839, opened 2024-07-29
fix(AdminSettings): remove PHP template

ShGKme:
### ☑️ Resolves

* Likely from legacy, during admin settings loading, half of the settings are half-loaded
* On slow connection it feels weird because the settings are unusable
* On fast connection it looks like a layout shift glitch

Replaced with a placeholder

🏚️ Before | 🏡 After
-- | --
image | image
before-admin-settings-load | after-admin-settings-load

### 🏁 Checklist

### DIFF
--- a/css/talk-admin-settings.css
+++ b/css/talk-admin-settings.css
@@ -1,0 +1,49 @@
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+.settings-section-placeholder {
+	--settings-section-placeholder-header-height: 30px;
+	--settings-section-placeholder-line-height: var(--default-line-height);
+	--settings-section-placeholder-padding: 1em;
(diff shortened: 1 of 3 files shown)


nickvergessen:
The other day I also planned to raise an issue about this. We are missing quite some sections there. I also get it that we can't keep them up to date all the time.
But is a blank page really better? Should it have a inline decorative content (loading animation or something?) instead?

ShGKme:
> But is a blank page really better?

I think it's better than some template settings. They don't work, don't correspond to rendered settings ("limit to groups" are replaced by "Matterbridge"), and cause a visual glitch (only the General settings header stays in place). And this is what we have in all apps with client-rendered settings...

But we can add something common

ShGKme:
Updated

nickvergessen (approved):
Feels much better

Antreesy (approved):
Looks good, but indentation is messy now in changed files

ShGKme:
Squashed with fixed tabs

nickvergessen:
`features/federation/chat.feature:73` can be ignored. Saw it elsewhere and am checking it locally today

PR #12840, opened 2024-07-29
fix(AdminSettings): migrate to NcSettingsSection

ShGKme:
### ☑️ Resolves

* Migrate to Nextcloud common `NcSettingsSection` for design consistency
  * Also, add doc links
* Missing features:
  * Description slot for description with links or multi-line description
  * `BETA` badge

🏚️ Before | 🏡 After
-- | --

### 🏁 Checklist

### DIFF
--- a/src/components/AdminSettings/Federation.vue
+++ b/src/components/AdminSettings/Federation.vue
@@ -4,11 +4,11 @@
 -->
 
 <template>
-	<section id="federation_settings" class="federation section">
-		<h2>
-			{{ t('spreed', 'Federation') }}
-			<small>{{ t('spreed', 'Beta') }}</small>
-		</h2>
+	<NcSettingsSection id="federation_settings"
+		class="federation"
+		:name="t('spreed', 'Federation')"
+		doc-url="https://docs.nextcloud.com/server/latest/user_manual/en/talk/advanced_features.html#federated-conversation">
+		<!-- <small>{{ t('spreed', 'Beta') }}</small> -->
 
 		<NcCheckboxRadioSwitch :checked="isFederationEnabled"
 			:disabled="loading"
@@ -76,7 +76,7 @@
 				</NcButton>
 			</div>
 		</template>
-	</section>
+	</NcSettingsSection>
 </template>
 
 <script>
@@ -90,6 +90,7 @@ import { generateOcsUrl, generateUrl } from '@nextcloud/router'
 import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
 import NcCheckboxRadioSwitch from '@nextcloud/vue/dist/Components/NcCheckboxRadioSwitch.js'
 import NcSelect from '@nextcloud/vue/dist/Components/NcSelect.js'
+import NcSettingsSection from '@nextcloud/vue/dist/Components/NcSettingsSection.js'
 
 const FEDERATION_ENABLED = loadState('spreed', 'federation_enabled', false)
 const FEDERATION_INCOMING_ENABLED = loadState('spreed', 'federation_incoming_enabled', true)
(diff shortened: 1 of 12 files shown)


src/components/AdminSettings/Federation.vue
```diff
@@ -4,11 +4,11 @@
...
-	<section id="federation_settings" class="federation section">
-		<h2>
-			{{ t('spreed', 'Federation') }}
-			<small>{{ t('spreed', 'Beta') }}</small>
-		</h2>
+	<NcSettingsSection id="federation_settings"
+		class="federation"
+		:name="t('spreed', 'Federation')"
+		doc-url="https://docs.nextcloud.com/server/latest/user_manual/en/talk/advanced_features.html#federated-conversation">
+		<!-- <small>{{ t('spreed', 'Beta') }}</small> -->
```
nickvergessen, line 11: Add to name then?

src/components/AdminSettings/MatterbridgeIntegration.vue
```diff
@@ -4,11 +4,11 @@
...
-	<section id="matterbridge_settings" class="matterbridge section">
-		<h2>
-			{{ t('spreed', 'Matterbridge integration') }}
-			<small>{{ t('spreed', 'Beta') }}</small>
-		</h2>
+	<NcSettingsSection id="matterbridge_settings"
+		class="matterbridge"
+		:name="t('spreed', 'Matterbridge integration')"
+		doc-url="https://nextcloud-talk.readthedocs.io/en/latest/matterbridge/">
+		<!-- <small>{{ t('spreed', 'Beta') }}</small> -->
```
nickvergessen, line 11: :see_no_evil:

src/components/AdminSettings/StunServers.vue
```diff
@@ -4,15 +4,10 @@
...
-		</h2>
-
-		<p class="settings-hint">
-			{{ t('spreed', 'A STUN server is used to determine the public IP address of participants behind a router.') }}
-		</p>
-
+	<NcSettingsSection id="stun_server"
+		:name="t('spreed', 'STUN servers')"
+		:description="t('spreed', 'A STUN server is used to determine the public IP address of participants behind a router.')"
+		doc-url="https://nextcloud-talk.readthedocs.io/en/latest/TURN/">
```
nickvergessen, line 10: stun is not turn?

src/components/AdminSettings/MatterbridgeIntegration.vue
ShGKme: > * Missing features:
>   * BETA badge

Requires some adjustments on `NcSettingsSection` 😶

src/components/AdminSettings/Federation.vue
ShGKme: Then it loses all the "badge" styles

src/components/AdminSettings/StunServers.vue
ShGKme: It is not, but on the TURN page (the "configuration/overview" page) there is a small paragraph describing what is STUN and when it is necessary.

> * Talk tries to establish a direct [peer-to-peer (P2P)](https://en.wikipedia.org/wiki/Peer-to-peer) connection, thus on connections beyond the local network (behind a [NAT](https://en.wikipedia.org/wiki/Network_address_translation) or router), clients do not only need to know each other's public IP, but the participants local IPs as well. Processing this, is the job of a [STUN](https://en.wikipedia.org/wiki/STUN) server. As there is one preconfigured for Nextcloud Talk that is operated by Nextcloud GmbH, for this case nothing else needs to be done.

Better to link to a Wiki article directly?

nickvergessen: I would say either it should be a useful installation/setup/configuration manual or we don't add a link

PR #12846, opened 2024-07-29
feat(RightSidebar): enforce tab navigation, show chat description in sidebar

Antreesy:
### ☑️ Resolves

* Ref #12800
* Follow-up of discussion in https://github.com/nextcloud/spreed/pull/12810#discussion_r1691396919

🏚️ Before | 🏡 After
-- | --
Group chat | --
image | image
image | image
Federated chat | --
image | image
image | image
1-1 chat | --
image | image
image | image

### 🚧 Tasks

- [x] Top bar title is cut off - keep it one line only

### 🏁 Checklist

### DIFF
--- a/src/components/RightSidebar/RightSidebar.vue
+++ b/src/components/RightSidebar/RightSidebar.vue
@@ -12,6 +12,7 @@
 		:class="'active-tab-' + activeTab"
 		:toggle-classes="{ 'chat-button-sidebar-toggle': isInCall }"
 		:toggle-attrs="isInCall ? inCallToggleAttrs : undefined"
+		force-tabs
 		@update:open="handleUpdateOpen"
 		@update:active="handleUpdateActive"
 		@closed="handleClosed"
@@ -27,6 +28,12 @@
 			</NcCounterBubble>
 		</template>
 		<template #description>
+			<NcRichText v-if="conversation.description"
+				class="rich-text-description"
+				:text="conversation.description"
+				dir="auto"
+				autolink
+				use-extended-markdown />
 			<LobbyStatus v-if="canFullModerate && hasLobbyEnabled" :token="token" />
 		</template>
 		<NcAppSidebarTab v-if="isInCall"
@@ -119,6 +126,7 @@ import NcAppSidebar from '@nextcloud/vue/dist/Components/NcAppSidebar.js'
 import NcAppSidebarTab from '@nextcloud/vue/dist/Components/NcAppSidebarTab.js'
 import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
 import NcCounterBubble from '@nextcloud/vue/dist/Components/NcCounterBubble.js'
+import NcRichText from '@nextcloud/vue/dist/Components/NcRichText.js'
 
 import BreakoutRoomsTab from './BreakoutRooms/BreakoutRoomsTab.vue'
 import LobbyStatus from './LobbyStatus.vue'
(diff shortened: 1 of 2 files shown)


src/components/TopBar/TopBar.vue
```diff
@@ -332,38 +312,31 @@ export default {
...
-	.description {
-		overflow: hidden;
-		text-overflow: ellipsis;
-		max-width: fit-content;
-		&__in-chat {
-			color: var(--color-text-maxcontrast);
-		}
+
+	&__status {
+		padding-inline: var(--default-grid-baseline);
```
ShGKme, line 337: Maybe `2 * --default-grid-baseline`?

src/components/RightSidebar/RightSidebar.vue
```diff
@@ -27,6 +28,12 @@
 			</NcCounterBubble>
 		</template>
 		<template #description>
+			<NcRichText v-if="conversation.description"
+				class="rich-text-description"
+				:text="conversation.description"
+				dir="auto"
+				autolink
+				use-extended-markdown />
```
ShGKme, line 36: Nice!

src/components/RightSidebar/RightSidebar.vue
```diff
@@ -12,6 +12,7 @@
 		:class="'active-tab-' + activeTab"
 		:toggle-classes="{ 'chat-button-sidebar-toggle': isInCall }"
 		:toggle-attrs="isInCall ? inCallToggleAttrs : undefined"
+		force-tabs
```
ShGKme, line 15: Not sure about forced tabs... IMO, looked good without navigation when there was only on tab.

src/components/RightSidebar/RightSidebar.vue
```diff
@@ -27,6 +28,12 @@
 			</NcCounterBubble>
 		</template>
 		<template #description>
+			<NcRichText v-if="conversation.description"
```
nickvergessen, line 31: needs a max height somehow

marcoambrosini:
@Antreesy what is the issue we're trying to solve here? At first glance I don't see improvements and only potential drawbacks

Antreesy:
@marcoambrosini check the initial PR and discussion: https://github.com/nextcloud/spreed/pull/12810#discussion_r1691396919
Was suggested either decrease the font size in TopBar to fit two lines in 34px, or find another way, like move the description out of it

jancborchardt (changes requested):
@Antreesy I agree with @marcoambrosini that this is mostly a potential drawback – the description was intentionally moved to the subline to be always visible, which is super helpful.

We can go for a third option: Increase the height of the header bar (aka not decrease it so much) to be able to fit 2 lines of text, to also be symmetric to any of the entries in the left navigation.

ShGKme:
> We can go for a third option: Increase the height of the header bar (aka not decrease it so much) to be able to fit 2 lines of text, to also be symmetric to any of the entries in the left navigation.

@jancborchardt Then it won't be in line with the navigation toggle, navigation search, sidebar toggle, and sidebar title, which intentionally was forced to be the same everywhere in components.

Antreesy:
See https://github.com/nextcloud/spreed/pull/12851
Looks okay for me with reduced padding and default line-height (at least text won't be trimmed)

marcoambrosini:
I would downsize tue subline to 13px 
Screenshot 2024-07-30 at 13 13 57

We should create a new variable for it: `--font-size-small`

PR #12851, opened 2024-07-30
style(TopBar): fit two-lines with normal line-height

Antreesy:
### ☑️ Resolves

* Ref #12800

🏚️ Before | 🏡 After
-- | --
🏚️ Before | image
🏡 After | image
🏚️ Before  | image
🏡 After | image

### 🏁 Checklist


nickvergessen:
Can you have rooms and descriptions with q, p and y ? :P

Antreesy:
> Can you have rooms and descriptions with q, p and y ? :P

They're not clipped at my machine anyway 🙈

src/components/TopBar/TopBar.vue
```diff
@@ -303,11 +303,12 @@ export default {
 	flex-wrap: wrap;
 	z-index: 10;
 	gap: 3px;
-	align-items: flex-start;
+	align-items: center;
 	justify-content: flex-end;
-	padding: calc(2 * var(--default-grid-baseline));
+	min-height: calc(2 * var(--default-line-height) + 3px);
+	padding-block: 1px;
```
DorraJaouad, line 309: let's make it 2 so height can match search box in LS ?

DorraJaouad (approved):
Looks good, letters like p y q are not cropped. No side effects on top bar in call.

src/components/TopBar/TopBar.vue
```diff
@@ -358,6 +359,7 @@ export default {
 		text-overflow: ellipsis;
 	}
 	.description {
+		font-size: var(--font-size-small, 13px);
```
marcoambrosini, line 362: Shouldn't the fallback here be the --default-font-size?

marcoambrosini (approved):

src/components/TopBar/TopBar.vue
Antreesy: You're right. adjusted. Here's the diff from stable29:

image

Header is thinner now by 7px, but still looks decent (and only desktop with older versions will be affected)

src/components/TopBar/TopBar.vue
```diff
@@ -303,17 +303,21 @@ export default {
 
 <style lang="scss" scoped>
 .top-bar {
+	--border-width: 1px;
+	// hardcoded 1.5 value of line-height for compatibility with older versions
+	--text-height: calc(1.5 * (var(--default-font-size) + var(--font-size-small, 15px)));
```
ShGKme, line 308: Why default small size to 15px? Do we want to backport it to 29?

Antreesy: Font size shouldn't change for Desktop client against 29, am I right?

ShGKme: It will use server styles required for the current Talk version.

Otherwise it's more painful and we would need support not only different backend versions but also styles on frontend

Antreesy: Please adjust in your follow up, when possible. I'm not getting it fully still 🤔

ShGKme: Was 15px added here only for the Desktop client?

Antreesy: Yes. Otherwise I don't see why we adjusted styles for stable versions before

ShGKme: We adjusted in libraries like `@nextcloud/vue` because it is used on multiple versions.

Did we do it on Talk?

Antreesy: https://github.com/nextcloud/spreed/commit/42e919117191e1bc266740991133a8166710fa2d

ShGKme: This PR is not backported and adapts to new styles and variables, which is necessary for Talk web on 30 🤔

Antreesy: I meant he iconSize, which depends on server styles and not the Talk version

https://github.com/nextcloud/spreed/commit/42e919117191e1bc266740991133a8166710fa2d#diff-841f2c9a1004d8fafe1e1f6fe44412b062860ea277a4ddc071068f7f2a162e7dR156-R157

ShGKme: It's better to get it from variable anyway.

PR #12854, opened 2024-07-30
fix(Reactions): integrate reactions into grid CSS layout

Antreesy:
### ☑️ Resolves

* Based on top of  #12823 
* make reactions wrapper a part of message grid, so it is aware of available width (date deducted)
* rewrite styles to default 'small' NcButton
* add thumbnails to reserve space for appearing icons (to avoid chat jumping)

🏚️ Before | 🏡 After |
-- | -- |
react-before | react-after

### 🏁 Checklist

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.vue
@@ -15,7 +15,7 @@
 			@after-show="fetchReactions">
 			<template #trigger>
 				<NcButton :type="userHasReacted(reaction) ? 'primary' : 'secondary'"
-					class="reaction-button"
+					size="small"
 					@click="handleReactionClick(reaction)">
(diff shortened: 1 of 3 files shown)


src/components/MessagesList/MessagesGroup/Message/MessagePart/Reactions.vue
```diff
@@ -274,20 +276,24 @@ export default {
 </script>
 <style lang="scss" scoped>
 .reactions-wrapper {
+	--minimal-button-width: 48px;
```
ShGKme, line 279: Can it be calculated?

Antreesy: if we can do it for width of `emoji character`, `whitespace`, `digit 1-9`, and add button paddings. Otherwise I don't see any options to do it =/

DorraJaouad (approved):
All good 🦅

PR #12859, opened 2024-07-31
fix(cs): Use single quotes for simple strings

nickvergessen:
### ☑️ Resolves

* Found by https://github.com/nextcloud/coding-standard/issues/23

### 🏁 Checklist

### DIFF
--- a/lib/MatterbridgeManager.php
+++ b/lib/MatterbridgeManager.php
@@ -342,7 +342,7 @@ class MatterbridgeManager {
 					$serverUrl = $part['server'];
 				} else {
 					$serverUrl = preg_replace('/\/+$/', '', $this->urlGenerator->getAbsoluteURL(''));
-					$content .= "	SeparateDisplayName = true" ."\n";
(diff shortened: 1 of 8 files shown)


lib/MatterbridgeManager.php
```diff
@@ -342,7 +342,7 @@ private function generateConfig(array $bridge): string {
 					$serverUrl = $part['server'];
 				} else {
 					$serverUrl = preg_replace('/\/+$/', '', $this->urlGenerator->getAbsoluteURL(''));
-					$content .= "	SeparateDisplayName = true" ."\n";
+					$content .= '	SeparateDisplayName = true' ."\n";
```
Antreesy, line 345: `"\n"` is fine for standard, or also should be changed?

Antreesy (approved):

lib/MatterbridgeManager.php
SystemKeeper: It needs to be, otherwise it’s not a new line anymore

PR #12861, opened 2024-07-31
fix(capabilities): store remote capabilities based on remoteServer

Antreesy:
### ☑️ Resolves

* Fix duplicated requests and reloads, if different federated rooms belong to the same remote server

Added some logs for visibility:
- for clean storage or new server - works as before, asking to reload the page
- for next, if server matches - update token map, cache new token for future
- in Network - no new requests for 'capabilities'

[Screencast from 31.07.2024 13:50:16.webm](https://github.com/user-attachments/assets/17f8c52c-e07f-433b-a30f-91dc1d6fd5a1)

### 🏁 Checklist

### DIFF
--- a/src/services/CapabilitiesManager.ts
+++ b/src/services/CapabilitiesManager.ts
@@ -130,10 +130,25 @@ export async function setRemoteCapabilities(joinRoomResponse: JoinRoomFullRespon
  * Restores capabilities from BrowserStorage
  */
 function restoreRemoteCapabilities(): RemoteCapabilities {
-	const remoteCapabilities = BrowserStorage.getItem('remoteCapabilities')
-	if (!remoteCapabilities?.length) {
+	const storageValue = BrowserStorage.getItem('remoteCapabilities')
+	if (!storageValue) {
 		return {}
 	}
+	const remoteCapabilities = JSON.parse(storageValue) as RemoteCapabilities
 
-	return JSON.parse(remoteCapabilities) as RemoteCapabilities
+	// Migration step for capabilities based on token
+	let hasMigrated = false
+	Object.keys(remoteCapabilities).forEach(key => {
+		const remoteServer = remoteTokenMap[key]
+		if (remoteServer) {
+			remoteCapabilities[remoteServer] = remoteCapabilities[key]
+			delete remoteCapabilities[key]
+			hasMigrated = true
+		}
+	})
+	if (hasMigrated) {
+		BrowserStorage.setItem('remoteCapabilities', JSON.stringify(remoteCapabilities))
+	}
+
+	return remoteCapabilities
 }
--- a/src/services/CapabilitiesManager.ts
+++ b/src/services/CapabilitiesManager.ts
@@ -10,13 +10,34 @@ import { t } from '@nextcloud/l10n'
 import { getRemoteCapabilities } from './federationService.ts'
 import BrowserStorage from '../services/BrowserStorage.js'
 import { useTalkHashStore } from '../stores/talkHash.js'
-import type { Capabilities, JoinRoomFullResponse } from '../types'
+import type { Capabilities, Conversation, JoinRoomFullResponse } from '../types'
 
 type Config = Capabilities['spreed']['config']
-type RemoteCapabilities = Record<string, Capabilities & Partial<{ hash: string }>>
+type RemoteCapability = Capabilities & Partial<{ hash: string }>
+type RemoteCapabilities = Record<string, RemoteCapability>
+type TokenMap = Record<string, string|undefined|null>
+
+let remoteTokenMap: TokenMap = generateTokenMap()
 
 const localCapabilities: Capabilities = _getCapabilities() as Capabilities
 const remoteCapabilities: RemoteCapabilities = restoreRemoteCapabilities()
+
+/**
+ * Generate new token map based on remoteCapabilities and cachedConversation
+ */
+function generateTokenMap() {
+	const tokenMap: TokenMap = {}
+	const storageValue = BrowserStorage.getItem('cachedConversations')
+	if (!storageValue?.length) {
+		return {}
+	}
+	const cachedConversations = JSON.parse(storageValue) as Conversation[]
+	cachedConversations.forEach(conversation => {
+		tokenMap[conversation.token] = conversation.remoteServer || null
+	})
+
+	return tokenMap
+}
 
 /**
  * Check whether the feature is presented (in case of federation - on both servers)
@@ -25,12 +46,13 @@ const remoteCapabilities: RemoteCapabilities = restoreRemoteCapabilities()
  */
 export function hasTalkFeature(token: string = 'local', feature: string): boolean {
 	const hasLocalTalkFeature = localCapabilities?.spreed?.features?.includes(feature) ?? false
+	const remoteCapabilities = getRemoteCapability(token)
 	if (localCapabilities?.spreed?.['features-local']?.includes(feature)) {
 		return hasLocalTalkFeature
-	} else if (token === 'local' || !remoteCapabilities[token]) {
+	} else if (token === 'local' || !remoteCapabilities) {
 		return hasLocalTalkFeature
 	} else {
-		return hasLocalTalkFeature && (remoteCapabilities[token]?.spreed?.features?.includes(feature) ?? false)
+		return hasLocalTalkFeature && (remoteCapabilities?.spreed?.features?.includes(feature) ?? false)
 	}
 }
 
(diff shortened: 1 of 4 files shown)


src/services/CapabilitiesManager.ts
```diff
@@ -41,14 +57,42 @@ export function hasTalkFeature(token: string = 'local', feature: string): boolea
...
+	if (remoteCapabilities[remoteTokenMap[token]]) {
+		return remoteCapabilities[remoteTokenMap[token]]
+	}
+
+	const cachedConversations = BrowserStorage.getItem('cachedConversations')
+	if (!cachedConversations?.length) {
+		return null
 	}
+
+	const remoteServer = JSON.parse(cachedConversations)?.[token]?.remoteServer
```
DorraJaouad, line 85: `cachedConversations` is an array not keyed by `token`

src/services/CapabilitiesManager.ts
```diff
@@ -41,14 +57,42 @@ export function hasTalkFeature(token: string = 'local', feature: string): boolea
...
+
+	if (remoteServer && remoteCapabilities[remoteServer]) {
+		console.debug(`Reuse remote capabilities from another conversation (same remote server ${remoteServer})`)
+		remoteCapabilities[remoteServer].tokens = [...new Set((remoteCapabilities[remoteServer].tokens || []).concat(token))]
+		BrowserStorage.setItem('remoteCapabilities', JSON.stringify(remoteCapabilities))
+		remoteTokenMap = generateTokenMap()
+		return remoteCapabilities[remoteServer]
+	}
```
DorraJaouad, line 93: Given that we searched in cached conversations using the current conversation token, we are basically looking in the same conversation, it is almost equal to 
```
if (remoteCapabilities[remoteTokenMap[token]]) {
		return remoteCapabilities[remoteTokenMap[token]]
	}
```
except that it is guaranteed to have server name from cached conversations.
if server ever existed from another conversation, it will be present in `remoteCapabilities` anyway

DorraJaouad: For updating token in tokens, maybe use simple existence check before appending the item? and we don't need to generate the map again as we can only add the new token.

src/services/CapabilitiesManager.ts
```diff
@@ -73,8 +118,9 @@ export async function setRemoteCapabilities(joinRoomResponse: JoinRoomFullRespon
 		return
 	}
 
-	remoteCapabilities[token] = { spreed: response.data.ocs.data }
-	remoteCapabilities[token].hash = joinRoomResponse.headers['x-nextcloud-talk-proxy-hash']
+	remoteCapabilities[remoteServer] = { spreed: response.data.ocs.data }
+	remoteCapabilities[remoteServer].hash = joinRoomResponse.headers['x-nextcloud-talk-proxy-hash']
+	remoteCapabilities[remoteServer].tokens = [...new Set((remoteCapabilities[remoteServer].tokens || []).concat(token))]
```
DorraJaouad, line 123: You rewrote `remoteCapabilities[remoteServer]` here, so no need to soft update `.tokens` unless you didn't mean to overwrite it :p

src/services/CapabilitiesManager.ts
```diff
@@ -41,14 +57,42 @@ export function hasTalkFeature(token: string = 'local', feature: string): boolea
...
+}
+
+/**
+ * Returns capability for specified token (if already matches from one of remote servers)
+ * @param token token of the conversation
+ */
+function getRemoteCapability(token: string): RemoteCapability | null {
+	if (remoteCapabilities[remoteTokenMap[token]]) {
+		return remoteCapabilities[remoteTokenMap[token]]
+	}
```
DorraJaouad, line 78:
```
if (remoteTokenMap[token] && remoteCapabilities[remoteTokenMap[token]]) {
		return remoteCapabilities[remoteTokenMap[token]]
	}

```
It may not be generated yet.

src/services/CapabilitiesManager.ts
```diff
@@ -41,14 +57,42 @@ export function hasTalkFeature(token: string = 'local', feature: string): boolea
...
-		return remoteCapabilities[token]?.spreed?.config?.[key1]?.[key2]
+		return remoteCapabilities?.spreed?.config?.[key1]?.[key2]
+	}
+}
+
+/**
+ * Returns capability for specified token (if already matches from one of remote servers)
+ * @param token token of the conversation
+ */
+function getRemoteCapability(token: string): RemoteCapability | null {
```
DorraJaouad, line 75: We can pass remoteServer here (from the functions where this was called), right?

DorraJaouad (commented):
First commit

src/services/CapabilitiesManager.ts
Antreesy: It's 75 places in the code, and not every place has `this.conversation` to get remoteServer

src/services/CapabilitiesManager.ts
Antreesy: It should be generated, when the file is imported/called first time

src/services/CapabilitiesManager.ts
Antreesy: reworked creating of remoteTokenMap, should be appropriate now

src/services/CapabilitiesManager.ts
DorraJaouad: and what about creating the conversation? its token won't be there yet

src/services/CapabilitiesManager.ts
DorraJaouad: but we can get conversation by calling the store here directly?

Antreesy: We can't, as it can be called before store is initialized. Might become better lately with Pinia

src/services/CapabilitiesManager.ts
```diff
@@ -10,27 +10,59 @@ import { t } from '@nextcloud/l10n'
...
+	if (!storageValue?.length) {
+		return {}
+	}
+	const cachedConversations = JSON.parse(storageValue) as Conversation[]
+	cachedConversations.forEach(conversation => {
+		tokenMap[conversation.token] = conversation.remoteServer || null
+	})
+
+	return tokenMap
+}
```
DorraJaouad, line 40: Why not keep storing tokens in 'remoteCapabilities' storage. It will only parse fed conversations tokens.

src/services/CapabilitiesManager.ts
```diff
@@ -73,9 +129,10 @@ export async function setRemoteCapabilities(joinRoomResponse: JoinRoomFullRespon
 		return
 	}
 
-	remoteCapabilities[token] = { spreed: response.data.ocs.data }
-	remoteCapabilities[token].hash = joinRoomResponse.headers['x-nextcloud-talk-proxy-hash']
+	remoteCapabilities[remoteServer] = { spreed: response.data.ocs.data }
+	remoteCapabilities[remoteServer].hash = joinRoomResponse.headers['x-nextcloud-talk-proxy-hash']
```
DorraJaouad, line 133:
```suggestion
	remoteCapabilities[remoteServer].spreed = response.data.ocs.data
	remoteCapabilities[remoteServer].hash = joinRoomResponse.headers['x-nextcloud-talk-proxy-hash']
	remoteCapabilities[remoteServer].tokens.append(token) //maybe check existence before
```

This way, if a new conversation joined from already know remote server, it will just update it and not overwrite known tokens list. It will later be stored in browser storage and used in initialization, instead of going through the whole list of conversations (local and remote)

src/services/CapabilitiesManager.ts
Antreesy: to distinguigh local (`null`), remote (`string`) and unknown (`undefined`) conversations

src/services/CapabilitiesManager.ts
Antreesy: New conversation won't go that far, if `hash` is equal to `x-nextcloud-talk-proxy-hash`.
I decided to drop storing tokens list in each remoteCaps in favor of tokenMap

src/services/CapabilitiesManager.ts
```diff
@@ -41,25 +73,49 @@ export function hasTalkFeature(token: string = 'local', feature: string): boolea
...
+ */
+function getRemoteCapability(token: string, callIndex: number = 0): RemoteCapability | null {
+	const remoteServer = remoteTokenMap[token]
+	if (!token || token === 'local' || remoteServer === null) {
+		// Local or no conversation opened
+		return null
+	} else if (remoteServer === undefined && callIndex < 3) {
+		// Unknown conversation, attempt to get remoteServer from cache
+		remoteTokenMap = generateTokenMap()
+		return getRemoteCapability(token, callIndex + 1)
```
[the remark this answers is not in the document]
Antreesy: Will change it. We need to run it maximum 1 time to update map from cached conversations. If it's not there - fallback to local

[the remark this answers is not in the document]
Antreesy: Will change it. We need to run it maximum 1 time to update map from cached conversations. If it's not there - fallback to local

DorraJaouad:
to also change `isFederationEnabled` from state to config 
`capabilities?.spreed?.config?.federation?.enabled`

nickvergessen:
Whats the state here?

Antreesy:
> Whats the state here?

Ready IMO, waiting for reviews

DorraJaouad (approved):
LGTM

PR #12869, opened 2024-07-31
Fix conversations layout

marcoambrosini:
- [x] Revert commit 12ec9babcb3f3bef763db380bfc1c32ab271f947
- [x] Reduce space between elements to 4px (outline)
- [x] Return avatar size to 44px

Screenshot 2024-07-31 at 18 01 44

### DIFF
--- a/src/components/ConversationIcon.vue
+++ b/src/components/ConversationIcon.vue
@@ -129,7 +129,7 @@ export default {
 
 		size: {
 			type: Number,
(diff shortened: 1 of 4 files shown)


Antreesy:
Rebased and revert previous commit, which was supposed to adjust height.
Now avatars should be reduced to initial 44px, and we can discuss changes later from there.

ShGKme (approved):

DorraJaouad (approved):

PR #12875, opened 2024-08-01
fix(TopBar): align lines

ShGKme:
### ☑️ Resolves

- With the default `line-height` one-line is 1.5 height of font
- In a limited spaces and with different font sizes, it looks a bit not aligned for me
  - The gap between lines is higher than between line and bottom border
  - Small text may look too small
- Using `normal` (about 1.2) line height ensures that the text is not cut off but also has no extra height even with normal font-size

🏚️ Before | 🏡 After
-- | --
image | image
image | image
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/TopBar/TopBar.vue
+++ b/src/components/TopBar/TopBar.vue
@@ -303,15 +303,13 @@ export default {
 <style lang="scss" scoped>
 .top-bar {
 	--border-width: 1px;
-	// hardcoded 1.5 value of line-height for compatibility with older versions
-	--text-height: calc(1.5 * (var(--default-font-size) + var(--font-size-small, 15px)));
 	display: flex;
 	flex-wrap: wrap;
 	z-index: 10;
 	gap: 3px;
 	align-items: center;
 	justify-content: flex-end;
-	min-height: calc(2 * var(--default-grid-baseline) + var(--text-height) + var(--border-width));
+	min-height: calc(var(--border-width) + 2 * (2 * var(--default-grid-baseline)) + var(--default-clickable-area));
 	padding-block: var(--default-grid-baseline);
 	// Reserve space for the sidebar toggle button
 	padding-inline: calc(2 * var(--default-grid-baseline)) calc(2 * var(--default-grid-baseline) + var(--app-sidebar-offset, 0));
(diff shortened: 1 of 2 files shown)


marcoambrosini:
I don't think this is ok for accessibility. Could check that?

ShGKme:
> I don't think this is ok for accessibility. Could check that?

With any solution it changes the top bar height, no matter what line height is set initially, if yuo mean this check.

image

ShGKme:
Updated to normal font-size with normal line-height everywhere

Antreesy (commented):
On Linux (evaluates to 22px):

B | A
-- | --
image | image

Diff:
image

ShGKme:
- Changed `normal` to `1.2` to make sure, it's the same in all browsers and OSs.
- Removed `--text-height` to make min-height based on similar elements size

Antreesy (approved):

src/components/TopBar/TopBar.vue
```diff
@@ -303,15 +303,13 @@ export default {
...
-	// hardcoded 1.5 value of line-height for compatibility with older versions
-	--text-height: calc(1.5 * (var(--default-font-size) + var(--font-size-small, 15px)));
 	display: flex;
 	flex-wrap: wrap;
 	z-index: 10;
 	gap: 3px;
 	align-items: center;
 	justify-content: flex-end;
-	min-height: calc(2 * var(--default-grid-baseline) + var(--text-height) + var(--border-width));
+	min-height: calc(var(--border-width) + 2 * (2 * var(--default-grid-baseline)) + var(--default-clickable-area));
```
Antreesy, line 312: Call Container uses magic number 60 or close, need to be updated to the same value there too

ShGKme: Done in a hotfix

Antreesy: Ok, but I thought, we're going for variables)

ShGKme:
Follow-up: use CSS variables in call view

src/components/TopBar/TopBar.vue
ShGKme: I found it more complex and risky. `CallView` is used in different entrypoints and not directly related to `TopBar`

PR #12890, opened 2024-08-02
feat(Call): show presenter overlay for local screenshare

DorraJaouad:
### ☑️ Resolves

Popover on one of the buttons (video or screenshare) couldn't fit because there are popovers or menu below. 

image

### 🚧 Tasks

- [x] Change the sentence ? and/or its location?

### 🏁 Checklist

### DIFF
--- a/src/components/CallView/CallView.vue
+++ b/src/components/CallView/CallView.vue
@@ -52,21 +52,21 @@
 						:shared-data="localSharedData"
 						is-big />
 					<!-- Remote or selected screen -->
-					<template v-else-if="showRemoteScreen || showSelectedScreen">
-						<Screen v-if="shownRemoteScreenCallParticipantModel"
-							:key="`screen-${shownRemoteScreenPeerId}`"
-							:token="token"
-							:call-participant-model="shownRemoteScreenCallParticipantModel"
-							:shared-data="sharedDatas[shownRemoteScreenPeerId]"
-							is-big />
-						<!-- presenter overlay -->
-						<PresenterOverlay v-if="shouldShowPresenterOverlay"
-							:token="token"
-							:model="shownRemoteScreenCallParticipantModel"
-							:shared-data="sharedDatas[shownRemoteScreenPeerId]"
-							:is-collapsed="!showPresenterOverlay"
-							@click="toggleShowPresenterOverlay" />
-					</template>
+					<Screen v-else-if="(showRemoteScreen || showSelectedScreen) && shownRemoteScreenCallParticipantModel"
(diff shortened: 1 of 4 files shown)


nickvergessen:
Since the dropdown is opening and bright white and this is also were you interact with, I think the bottom note will be missed.
Are we trying to solve a ticket with this?

Antreesy:
For the location - it would fit better as a popover to Video toggle
image

DorraJaouad:
> Since the dropdown is opening and bright white and this is also were you interact with, I think the bottom note will be missed.
Are we trying to solve a ticket with this?

> For the location - it would fit better as a popover to Video toggle


In video toggle (tooltip) and in screenshare (dropdown), adding a popover there can make it crowded IMO. Popover will overlap one of them.

It's not a ticket but a suggestion discussed previously.

DorraJaouad:
> I think the bottom note will be missed.

I am also a bit skeptical, what about top left ?

nickvergessen:
> It's not a ticket but a suggestion discussed previously.

Then let's focus on the tickets assigned to the beta milestone for now?

nickvergessen:
Other idea: make the own video floating just like for all other participants?

DorraJaouad:
> Other idea: make the own video floating just like for all other participants?

Sounds good but I am not sure about floating part, your own video should be different, maybe put it on bottom right in rectangular shape? 
image

it disappears if:
- You click on the video
- You open stripe

and instead, a small button will be shown (like presenter overlay)

wdy ? cc @nextcloud/designers

Antreesy:
> your own video should be different,

would be better to keep it the same, so other participants see what you see (otherwise your head might be cut-off by presenter overlay circle without you knowing)

DorraJaouad:
> would be better to keep it the same, so other participants see what you see (otherwise your head might be cut-off by presenter overlay circle without you knowing)

There won't be other presenter overlay with your own screenshare

> Could someone please clarify the issue?

When a user shares their screen while having camera on, it's not visible in the UI that people see both your video and screen. 
We thought about writing a hint (see PR description, video at the bottom left) indicating that or we keep showing self video (either floating like presenter overlay or fixed at bottom right).

ShGKme:
> or we keep showing self video

👍

jancborchardt (commented):
Yeah, we should keep showing one’s own video like others see it.

Additionally for the note I think @Antreesy’s suggestion from https://github.com/nextcloud/spreed/pull/12890#issuecomment-2265311917 makes most sense, as this is where the clickable elements are. I also missed the bottom left note in the video.

Antreesy (approved):
Tested

Issue #12912, opened 2024-08-07
Make more obvious that you can add Federated Cloud ID accounts


jancborchardt:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Feature request

**Which Nextcloud Version are you currently using:** (see administration page)
Latest master of v30.

**Is your feature request related to a problem? Please describe.**
Currently it is not very clear that you can add people to conversations via Federated Cloud ID, as noted by @karlitschek 
![image](https://github.com/user-attachments/assets/4f7ffb7a-a11e-4dc5-83d1-66a469146985)

**Describe the solution you'd like**
In Files, the input at least says:
> Name, email or Federated Cloud ID …

The placeholder can be adjusted to be similar:
> Search or add participants via name or Federated Cloud ID …

Maybe even something shorter works.
![image](https://github.com/user-attachments/assets/bb5523e2-a19c-4a50-99aa-c5c0f7b2d449)

**Describe alternatives you've considered**
To make it even more visible, in addition to adjusting the placeholder, we could add a short info-paragraph below the search field saying something like:
> Participants can now also be added via Federated Cloud ID. Find yours in the sharing settings.

("sharing settings" is a link to: `settings/user/sharing`)

ShGKme:
> To make it even more visible, in addition to adjusting the placeholder, we could add a short info-paragraph below the search field saying something like:

A permanent note between the search and the participants list? Won't it be too much?

ShGKme:
> The placeholder can be adjusted to be similar:
> 
  > > Search or add participants via name or Federated Cloud ID …

This is quite a huge line even in English. And even bigger in other languages like German

Small | Large
---|---
![image](https://github.com/user-attachments/assets/a2a69d4b-7271-4728-bf8c-623071ae2169) | ![image](https://github.com/user-attachments/assets/2c258457-e6d4-4a2a-b58e-b795cb20265c)


> Name, email or Federated Cloud ID …

In this case, though, it's less obvious than you can use search for adding participants.

karlitschek:
So this longer description text takes a lot of space and is too short to properly explain it to users at the same time. Can we do something with a mouseover. or make the input line bigger with more explanation on click?

[...]

PR #12916, opened 2024-08-07
fix: make more obvious that you can add Federated Cloud ID accounts

ShGKme:
### ☑️ Resolves

* Fix https://github.com/nextcloud/spreed/issues/12912

🏚️ Before | 🏡 After
-- | --
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/RightSidebar/Participants/ParticipantsTab.vue
+++ b/src/components/RightSidebar/Participants/ParticipantsTab.vue
@@ -12,9 +12,17 @@
 				:value.sync="searchText"
 				:is-focused.sync="isFocused"
 				:placeholder-text="searchBoxPlaceholder"
+				:aria-describedby="showSearchBoxDescription && searchBoxDescriptionId"
 				@input="handleInput"
 				@keydown.enter="addParticipants(participantPhoneItem)"
 				@abort-search="abortSearch" />
+			<div v-if="showSearchBoxDescription"
+				:id="searchBoxDescriptionId"
+				:title="searchBoxDescription"
+				class="search-form__description">
+				<IconInformationOutline :size="20" />
+				<span class="hidden-visually">{{ searchBoxDescription }}</span>
+			</div>
 			<DialpadPanel v-if="canAddPhones"
 				:value.sync="searchText"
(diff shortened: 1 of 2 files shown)


SystemKeeper:
Should we make that dependent on federation enabled/allowed?

ShGKme:
> Should we make that dependent on federation enabled/allowed?

Yes, but currently I don't know what's the best way to show the text. It's too long for this field.

nickvergessen:
@jancborchardt @marcoambrosini any idea?

nickvergessen:
Ah, input is in https://github.com/nextcloud/spreed/issues/12912#issuecomment-2275576321

nickvergessen:
@ShGKme would be nice to get this in before tomorrows RC1 which is the translation string freeze

src/components/RightSidebar/Participants/ParticipantsTab.vue
```diff
@@ -12,9 +12,17 @@
...
+				:aria-describedby="showSearchBoxDescription && searchBoxDescriptionId"
 				@input="handleInput"
 				@keydown.enter="addParticipants(participantPhoneItem)"
 				@abort-search="abortSearch" />
+			<div v-if="showSearchBoxDescription"
+				:id="searchBoxDescriptionId"
+				:title="searchBoxDescription"
+				class="search-form__description">
+				<IconInformationOutline :size="20" />
+				<span class="hidden-visually">{{ searchBoxDescription }}</span>
```
Antreesy, line 24: Wouldn't aria-label be sufficient here? If not, I'm afraid, we might accidentally delete it in future, if it won't be specifically commented

Antreesy (approved):
Also looks fine with Dialpad panel
image

jancborchardt (approved):

src/components/RightSidebar/Participants/ParticipantsTab.vue
ShGKme: Alternative is to remove the text here and set it as `aria-desription` on the input. But then I'm afraid it's even simpler to lose, because it would have no direct relation with the element.

PR #12918, opened 2024-08-07
fix(MediaSettings): provide accessible heading

ShGKme:
### ☑️ Resolves

image

### 🏁 Checklist

### DIFF
--- a/src/composables/useId.ts
+++ b/src/composables/useId.ts
@@ -1,0 +1,13 @@
+/*
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+/**
+ * Generate a random string id for HTML RefIDs
+ * @todo Replace with real useId composable from Vue 3.5 after upgrade
+ * @return {string} String with 6 characters of digits and letters
+ */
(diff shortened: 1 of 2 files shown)


src/composables/useId.ts
```diff
@@ -0,0 +1,13 @@
...
+ */
+
+/**
+ * Generate a random string id for HTML RefIDs
+ * @todo Replace with real useId composable from Vue 3.5 after upgrade
+ * @return {string} String with 6 characters of digits and letters
+ */
+export function useId(): string {
+	return Math.random().toString(36).slice(2, 9)
+}
```
DorraJaouad, line 13: Wouldn't it be regenerated on closing/open mediaSettings (`NcModal` is re-rendered base on boolean `modal`) ?

ShGKme: On mounting/unmounting the modal itself — no.

On mounting/unmounting `MediaSettings` component — yes, because it's created on `MediaSettings` creating. But we do it only once.

But it is fine anyway. It only must be uniq on the page.

DorraJaouad (approved):

Antreesy:
Don't we need it for all NcModal components?

ShGKme:
> Don't we need it for all NcModal components?

Yes, or use high-level `NcDialog` component

PR #12919, opened 2024-08-07
fix: show specific error

### LINKED ISSUE #12914: Windows compatibilty: upload does not show correct error
- enable windows compatible filenames (in admin -> settings)
- try to upload a forbidden file, e.g. CON
- see without any real error message, only "error while uploading file"
image

DorraJaouad:
### ☑️ Resolves

* Fix #12914

🏚️ Before | 🏡 After
-- | --
image | image

### 🏁 Checklist

### DIFF
--- a/src/store/fileUploadStore.js
+++ b/src/store/fileUploadStore.js
@@ -349,7 +349,20 @@ const actions = {
 				context.commit('markFileAsPendingUpload', { uploadId, index, sharePath: uniquePath })
 			} catch (exception) {
 				console.error(`Error while uploading file "${fileName}":` + exception.message, fileName)
-				showError(t('spreed', 'Error while uploading file "{fileName}"', { fileName }))
+				if (exception.response) {
+					try {
+						const responseText = await exception.response.text()
+						const parser = new DOMParser()
+						const xmlDoc = parser.parseFromString(responseText, 'application/xml')
+						const messageElement = xmlDoc.getElementsByTagName('s:message')[0]
+						if (messageElement) {
+							const errorMessage = messageElement.textContent
+							showError(errorMessage)
+						}
+					} catch (parseError) {
+						showError(t('spreed', 'Error while uploading file "{fileName}"', { fileName }))
+					}
+				}
 				// Mark the upload as failed in the store
 				context.commit('markFileAsFailedUpload', { uploadId, index })
 				const { id } = uploadedFile.temporaryMessage


src/store/fileUploadStore.js
```diff
@@ -349,7 +349,20 @@ const actions = {
...
 				console.error(`Error while uploading file "${fileName}":` + exception.message, fileName)
-				showError(t('spreed', 'Error while uploading file "{fileName}"', { fileName }))
+				if (exception.response) {
+					try {
+						const responseText = await exception.response.text()
+						const parser = new DOMParser()
+						const xmlDoc = parser.parseFromString(responseText, 'application/xml')
+						const messageElement = xmlDoc.getElementsByTagName('s:message')[0]
+						if (messageElement) {
+							const errorMessage = messageElement.textContent
```
ShGKme, line 359: This part parses a specific XML response and should be a part of the transport application layer, not a global UI state module (store). Let's move this logic out from Vuex. So in the UI part we know that the process was not successful and there is a specific error.

Also, we can consider it as a duplicated (= unavailable) name and proceed to uploading with a different name.

src/store/fileUploadStore.js
```diff
@@ -349,7 +350,12 @@ const actions = {
 				context.commit('markFileAsPendingUpload', { uploadId, index, sharePath: uniquePath })
 			} catch (exception) {
 				console.error(`Error while uploading file "${fileName}":` + exception.message, fileName)
-				showError(t('spreed', 'Error while uploading file "{fileName}"', { fileName }))
+				if (exception.response) {
+					const message = await parseUploadError(exception)
+					if (message) {
+						showError(message)
```
nickvergessen, line 356: is it in user language already?

DorraJaouad: Yes, it's translated in server.

Antreesy (approved):
Looks good, tested with reserved word

src/store/fileUploadStore.js
```diff
@@ -349,7 +350,12 @@ const actions = {
...
+					}
```
nickvergessen, line 357: Hmm should we:
```suggestion
					} else {
						showError(t('spreed', 'Error while uploading file "{fileName}"', { fileName }))
					}
```

Antreesy: yep, would be as before for unhandled cases

DorraJaouad: oops, I removed it by mistake

PR #12926, opened 2024-08-08
fix(TopBar): hand raised button animation

ShGKme:
### ☑️ Resolves

* Ref: https://github.com/nextcloud/spreed/issues/12800

animation

### 🏁 Checklist


DorraJaouad (commented):
The raised hand animation looks good, I like it :D. 
Rergarding media settings (not related to this PR but to https://github.com/nextcloud/spreed/pull/12815), there is a space created in devices selector to keep the same height with the other tab. It seems a bit off.  Maybe a delimiter to indicate that it slides and that space is not random would be good there. 
image

ShGKme:
> there is a space created in devices selector to keep the same height with the other tab.

Making it height-stable was the goal from the design review 🙈

> Maybe a delimiter to indicate that it slides and that space is not random would be good there.

🤔 Can try in a separate PR (not related to raised hand)

DorraJaouad:
> Making it height-stable was the goal from the design review 🙈

Yes, I want to indicate that in the UI too and not remove it

DorraJaouad (approved):
✨

ShGKme:
@DorraJaouad Like this?

image

Issue #10774, opened 2023-10-26
Conversation icon background color stays after switching OS dark/light mode


juliushaertl:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Steps to reproduce
1. Have OS in light mode
2. Have talk open
3. Switch to OS in dark mode

Also happens the other way round

### Expected behaviour

The background of the conversation icons in the left sidebar changes

This is how it looks after a page reload:

![Screenshot 2023-10-26 at 09 01 20](https://github.com/nextcloud/spreed/assets/3404133/956218c2-abb7-41df-80a2-1d1d93c741f0)

### Actual behaviour

The background stays

![Screenshot 2023-10-26 at 09 01 09](https://github.com/nextcloud/spreed/assets/3404133/3d02e427-6aa6-41d4-857b-46af3d2841b5)

Reproduced on our instance

SystemKeeper:
> the image is cached in browser cache and the URL can not change. Press ctrl+f5 or wait max. 24h

But we do have different endpoints for that: 
https://github.com/nextcloud/spreed/blob/ca1ad2d43c525256fccce4f8940ece4436e935f8/docs/avatar.md?plain=1#L70-L92

The check is also correctly working:

https://github.com/nextcloud/spreed/blob/ca1ad2d43c525256fccce4f8940ece4436e935f8/src/utils/isDarkTheme.js#L32-L32

![image](https://github.com/nextcloud/spreed/assets/1580193/b688bc2c-d299-4dd3-8a83-ed6f9835be46)

But the conversation list does not take that into account

Used here:

https://github.com/nextcloud/spreed/blob/ca1ad2d43c525256fccce4f8940ece4436e935f8/src/components/ConversationIcon.vue#L220

But I guess since `isDarkTheme` is a function and therefore not reactive, it changes only after a reload.

DorraJaouad:
Currently, we are updating the theme based on a CSS variable `'--background-invert-if-dark'` . Since CSS variables can't change without reloading the page, this means the theme check will only happen on load.

https://github.com/nextcloud/spreed/blob/cd20ba221b582ba95bdbad7e70221eb71569ddc5/src/utils/isDarkTheme.js#L32

PR #12940, opened 2024-08-12
feat(frontend): handle dark them change without reload

ShGKme:
### ☑️ Resolves

* Fix https://github.com/nextcloud/spreed/issues/10774
* Not much actually on web, but became more useful on Talk Desktop with theme settings

🏚️ Before | 🏡 After
-- | --

### 🚧 Tasks

WIP

- [x] Add `useIsDarkTheme` composable
  - Only works on global theme change, but not on a random element
- [x] Use composable instead of constant
- [x] Remount avatars on them change with `key`
  - Components doesn't support dark theme change internally
- [ ] If everything works well - add composable to `@nextcloud/vue`

### 🏁 Checklist


Antreesy:
Mentions styles are also hardcoded atm:
image

src/components/UIShared/LoadingPlaceholder.vue
```diff
@@ -21,10 +20,6 @@
 <script>
 import { AVATAR } from '../../constants.js'
 
-const bodyStyles = window.getComputedStyle(document.body)
```
Antreesy, line 24: Now I'm also wondering

src/utils/isDarkTheme.ts
```diff
@@ -0,0 +1,30 @@
...
+
+	// There is no theme? Fallback to the light theme
+	return false
+}
+
+/**
+ * Was Dark Theme enabled on the page load
+ * @type {boolean}
+ */
+export const isDarkTheme = checkIfDarkTheme()
```
Antreesy, line 30: Seems to be not used in project anymore? In worst case we can call a function (if used outside of Vue component)

Antreesy (approved):
Nice, works in browser as well!

ShGKme:
> data-theme-default is not presented on document.body

It should... Can you show an example?

ShGKme:
Rebased to resolve conflict.

Added a fixup tp remove unused variable.

Antreesy:
> > data-theme-default is not presented on document.body
> 
> It should... Can you show an example?

Just first-time logged-in users only affected. Upstream to server/theming?

image

ShGKme:
Rebased, squashed, no new changes

PR #12956, opened 2024-08-13
fix(avatar): decrease avatar size from 44px to 36px

Antreesy:
### ☑️ Resolves

* Part of design changes to Nc30
* User status is reduced in `@nextcloud/vue@master` https://github.com/nextcloud-libraries/nextcloud-vue/pull/5959

🏚️ Before | 🏡 After
-- | --
image | image
image | image
image | image
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -1089,7 +1089,9 @@ export default {
 	padding: 0 !important;
 }
 
+// Overwrite NcListItem styles
 :deep(.list-item) {
+	line-height: 1.2;
 	overflow: hidden;
 	outline-offset: -2px;
 }
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -1094,5 +1094,15 @@ export default {
 	line-height: 1.2;
 	overflow: hidden;
 	outline-offset: -2px;
+
+	// FIXME clean up after nextcloud/vue release
+	.avatardiv .avatardiv__user-status {
+		right: -2px !important;
+		bottom: -2px !important;
+		min-height: 14px !important;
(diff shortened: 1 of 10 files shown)


src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -1088,8 +1089,17 @@ export default {
 	padding: 0 !important;
 }
 
+// Overwrite NcListItem styles
 :deep(.list-item) {
+	padding-block: calc(1.5 * var(--default-grid-baseline)) !important;
```
ShGKme, line 1094: Why change default padding?

Antreesy: Was done before user status size adjustment, to compensate spacing around it. Will check, if --list-item-height can be adjusted instead

ShGKme: > `--list-item-height: 2lh;`

If we want elements inside the list item to have `line-height: 1.2` and adjust paddings (height) to their lines height, can we make an entire list item (without deep) having `line-height: 1.2`?

Then it works without overrinding styles and there is no extra padding, line height `2lh` is still correct.

Antreesy: Sure, but then it will evaluate to 36px. Unless we're also decrease the avatar size to 36px, it will overflow the content

ShGKme: I missed, why avatar must be 38px? Not 36px, inline with the text content.

Antreesy: No reason, tbh, let me push and make screenshots with 36px 🦭

marcoambrosini (commented):
The position of the status icon also needs to change, likely 2px up and 2px left

nickvergessen (approved):

PR #12969, opened 2024-08-14
fix(call): Fix race condition of "End call for everyone" and a parall…

nickvergessen:
…el "Leave call"

## 🛠️ API Checklist

The following patch will simulate a heavier installation and can be used to reliable produce the issue before

```diff
diff --git a/lib/Controller/CallController.php b/lib/Controller/CallController.php
index 07eb3b62f..89d83c30a 100644
--- a/lib/Controller/CallController.php
+++ b/lib/Controller/CallController.php
@@ -415,8 +415,10 @@ class CallController extends AEnvironmentAwareController {
                        return $response;
                }
 
+               sleep(10);
                if ($all && $this->participant->hasModeratorPermissions()) {
                        $this->participantService->endCallForEveryone($this->room, $this->participant);
+                       sleep(10);
                        $this->roomService->resetActiveSince($this->room, $this->participant, true);
                } else {
                        $this->participantService->changeInCall($this->room, $this->participant, Participant::FLAG_DISCONNECTED);
```

### 🏁 Checklist

### DIFF
--- a/lib/Controller/CallController.php
+++ b/lib/Controller/CallController.php
@@ -423,8 +423,14 @@ class CallController extends AEnvironmentAwareController {
 		}
 
 		if ($all && $this->participant->hasModeratorPermissions()) {
+			$result = $this->roomService->resetActiveSinceInDatabaseOnly($this->room);
+			if (!$result) {
+				// Someone else won the race condition, make sure this user disconnects directly and then return
+				$this->participantService->changeInCall($this->room, $this->participant, Participant::FLAG_DISCONNECTED);
+				return new DataResponse();
+			}
 			$this->participantService->endCallForEveryone($this->room, $this->participant);
-			$this->roomService->resetActiveSince($this->room, $this->participant, true);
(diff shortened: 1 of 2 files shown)


lib/Controller/CallController.php
```diff
@@ -416,8 +416,14 @@ public function leaveCall(bool $all = false): DataResponse {
 		}
 
 		if ($all && $this->participant->hasModeratorPermissions()) {
+			$result = $this->roomService->resetActiveSinceInDatabaseOnly($this->room);
+			if (!$result) {
+				// Someone else won the race condition, make sure this user disconnects directly and then return
+				// FIXME $this->participantService->changeInCall($this->room, $this->participant, Participant::FLAG_DISCONNECTED);
```
nickvergessen, line 422: this looks like a good idea, but it makes the user being missed in the call summary. TO be investigated

nickvergessen: Can no longer reproduce. Maybe it was something else :see_no_evil:

PR #12970, opened 2024-08-14
fix: migrate NcModal to NcDialog

Antreesy:
### ☑️ Resolves

* Follow-up to #12918
* Not a complete work
* For easy cases: migrate from NcModal to NcDialog

🏚️ Before | 🏡 After
-- | --
image | image
image | image
image | image
image | image
image | image

### 🚧 Tasks

- [ ] TODO:
- [ ] migrate rest to NcDialog

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/BreakoutRoomsEditor/SendMessageDialog.vue
+++ b/src/components/BreakoutRoomsEditor/SendMessageDialog.vue
@@ -4,32 +4,32 @@
 -->
 
 <template>
-	<NcModal ref="modal"
+	<NcDialog ref="modal"
+		:name="dialogTitle"
 		:container="container"
-		v-on="$listeners">
-		<div class="send-message-dialog">
-			<h2 class="send-message-dialog__title">
-				{{ dialogTitle }}
-			</h2>
-			<NewMessage v-if="modalContainerId"
-				:key="modalContainerId"
-				ref="newMessage"
-				role="region"
-				:token="token"
-				:container="modalContainerId"
-				:aria-label="t('spreed', 'Post message')"
-				:broadcast="broadcast"
-				@sent="handleMessageSent"
-				@failure="handleMessageFailure" />
-		</div>
-	</NcModal>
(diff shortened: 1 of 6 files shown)


ShGKme (approved):

ShGKme:
Didn't test, but looks fine

src/components/BreakoutRoomsEditor/SendMessageDialog.vue
```diff
@@ -4,40 +4,40 @@
...
-				:aria-label="t('spreed', 'Post message')"
-				:broadcast="broadcast"
-				@sent="handleMessageSent"
-				@failure="handleMessageFailure" />
-		</div>
-	</NcModal>
+		close-on-click-outside
+		size="normal"
+		v-on="$listeners"
+		@update:open="$emit('close')">
```
DorraJaouad, line 13: isn't it more accurate to use `@closing` event here ?

Antreesy: From nextcloud/vue - I think we should use the one after transition ends:

```
		/**
		 * Handle closing the dialog, optional out transition did not run yet
		 */
		const handleClosing = () => {
			showModal.value = false
			/**
			 * Emitted when the dialog is closing, so the out transition did not finish yet
			 */
			emit('closing')
		}

		/**
		 * Handle dialog closed (out transition finished)
		 */
		const handleClosed = () => {
			showModal.value = true
			/**
			 * Emitted then the dialog is fully closed and the out transition run
			 */
			emit('update:open', false)
		}
```

DorraJaouad: Alright, because `@update:open` is not only restricted to emit `false` value (for now, yes but it might change).

DorraJaouad (approved):

PR #13009, opened 2024-08-15
feat: Added blurhash to image metadata

### LINKED ISSUE #12351: Add blurhash to image metadata
### How to use GitHub

---

## Feature request

For showing a preview of an image while it is still loading a blurhash can be used. It is not currently exposed in the metadata though. This was supposed to happen with https://github.com/nextcloud/spreed/issues/10844, but it wasn't added in https://github.com/nextcloud/spreed/pull/11093.

skalidindi53:
### ☑️ Resolves

* Fix #12351 

### 🚧 Tasks

- [x] Added blurhash to image metadata

### 🏁 Checklist

### DIFF
--- a/lib/Chat/Parser/SystemMessage.php
+++ b/lib/Chat/Parser/SystemMessage.php
@@ -812,7 +812,7 @@ class SystemMessage implements IEventListener {
 		// If a preview is available, check if we can get the dimensions of the file from the metadata API
 		if ($isPreviewAvailable && str_starts_with($node->getMimeType(), 'image/')) {
 			try {
-				$sizeMetadata = $this->metadataCache->getMetadataPhotosSizeForFileId($fileId);
+				$sizeMetadata = $this->metadataCache->getImageMetadataForFileId($fileId);
 				if (isset($sizeMetadata['width'], $sizeMetadata['height'])) {
 					$data['width'] = (string) $sizeMetadata['width'];
 					$data['height'] = (string) $sizeMetadata['height'];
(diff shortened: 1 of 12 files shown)


lib/Chat/Parser/SystemMessage.php
```diff
@@ -817,6 +817,10 @@ protected function getFileFromShare(Room $room, ?Participant $participant, strin
 					$data['width'] = (string) $sizeMetadata['width'];
 					$data['height'] = (string) $sizeMetadata['height'];
 				}
+
+				if (isset($sizeMetadata['blurhash'])) {
+					$data['blurhash'] = $sizeMetadata['blurhash'];
```
nickvergessen, line 822: Hmm we should add it to https://github.com/nextcloud/server/blob/master/lib/public/RichObjectStrings/Definitions.php#L294 and https://github.com/nextcloud/spreed/blob/main/lib/ResponseDefinitions.php#L93 I guess

nickvergessen:
@SystemKeeper can/want to give this a try?

nickvergessen (approved):

SystemKeeper:
> @SystemKeeper can/want to give this a try?

The PR works fine for me and correctly returns the blurhash. 

Would it make sense to rename size-related methods/variables like `getMetadataPhotosSizeForFileId` to `getImageMetadataForFileId` or the like to make it clearer that this is not only about size?

One thing I noticed is that blurhash generation has since been moved to a background job: https://github.com/nextcloud/server/blob/560282a47bf3f9dca4653f89d530a6e1dd144f5f/lib/private/Blurhash/Listener/GenerateBlurhashMetadata.php#L57-L61

So when we upload images, a blurhash will only be available after the next background run, which might limit the usefulness a bit...

Example from ios:

SystemKeeper (approved):

nickvergessen:
> One thing I noticed is that blurhash generation has since been moved to a background job

Hmpf, makes sense performancewise, but yeah kind of kills the feature

PR #13010, opened 2024-08-16
fix(avatar): increase avatar size to 40px

Antreesy:

🏚️ Before | 🏡 After
-- | --
image | image
image | image
image | image
image | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [x] Talk Desktop

### DIFF
--- a/src/components/LeftSidebar/ConversationsList/ConversationsListVirtual.vue
+++ b/src/components/LeftSidebar/ConversationsList/ConversationsListVirtual.vue
@@ -24,14 +24,16 @@ import { RecycleScroller } from 'vue-virtual-scroller'
 import Conversation from './Conversation.vue'
 import LoadingPlaceholder from '../../UIShared/LoadingPlaceholder.vue'
 
+import { AVATAR } from '../../../constants.js'
+
 import 'vue-virtual-scroller/dist/vue-virtual-scroller.css'
 
 /* Consider:
- * 36 = 2 * 1.2 * var(--default-font-size) - two lines of text
- * 8 = 2 * var(--default-grid-baseline) - item padding
- * 4 = var(--default-grid-baseline) - item outline (collapsed)
+ * avatar size (and two lines of text)
+ * list-item padding
+ * list-item__wrapper padding
  */
-const CONVERSATION_ITEM_SIZE = 48
+const CONVERSATION_ITEM_SIZE = AVATAR.SIZE.DEFAULT + 2 * 4 + 2 * 2
 
 export default {
 	name: 'ConversationsListVirtual',
--- a/src/components/LeftSidebar/InvitationHandler.vue
+++ b/src/components/LeftSidebar/InvitationHandler.vue
@@ -228,14 +228,14 @@ export default {
 			padding-left: 4px;
 
 			&__name {
-				line-height: 1.2;
+				line-height: 20px;
 				font-weight: bold;
 				color: var(--color-main-text);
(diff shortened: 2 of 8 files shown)


src/components/LeftSidebar/ConversationsList/ConversationsListVirtual.vue
```diff
@@ -27,11 +27,11 @@ import LoadingPlaceholder from '../../UIShared/LoadingPlaceholder.vue'
 import 'vue-virtual-scroller/dist/vue-virtual-scroller.css'
 
 /* Consider:
- * 36 = 2 * 1.2 * var(--default-font-size) - two lines of text
+ * 39 = 2 * 1.3 * var(--default-font-size) - two lines of text
  * 8 = 2 * var(--default-grid-baseline) - item padding
  * 4 = var(--default-grid-baseline) - item outline (collapsed)
  */
-const CONVERSATION_ITEM_SIZE = 48
+const CONVERSATION_ITEM_SIZE = 51
```
ShGKme, line 34: To not check that comment is true.
Or with variables.
```suggestion
const CONVERSATION_ITEM_SIZE = 2 * 1.3 * 15 + 2 * 4 + 4
```

src/components/LeftSidebar/InvitationHandler.vue
```diff
@@ -228,14 +228,14 @@ export default {
 			padding-left: 4px;
 
 			&__name {
-				line-height: 1.2;
+				line-height: 1.3;
```
ShGKme, line 231: `line-height` is inherited and not overwritten in `NcListItem`. We can put it directly on the `NcListItem` without `deep` overrides and without specifying each element manually with internal classes.

src/constants.js
```diff
@@ -273,7 +273,7 @@ export const AVATAR = {
 	SIZE: {
 		EXTRA_SMALL: 22,
 		SMALL: 32,
-		DEFAULT: 36,
+		DEFAULT: 39,
```
ShGKme, line 276: Is it actually the "default" size? It seems that it is the size of avatar in list items, defined by the list item height (= conversations list and participants list).

src/components/LeftSidebar/ConversationsList/ConversationsListVirtual.vue
ShGKme: If we have a constant for avatar size, then it's SIZE + 2 * 4 + 4

src/components/LeftSidebar/InvitationHandler.vue
Antreesy: Not in that case (not a NcListItem), but I got the idea

src/constants.js
Antreesy: which is also most widely used in project

src/components/UIShared/LoadingPlaceholder.vue
```diff
@@ -19,6 +19,8 @@
 </template>
 
 <script>
+import { AVATAR } from '../../constants.js'
```
DorraJaouad, line 22: 👍🏻

DorraJaouad (approved):
LGTM (avatar size)

src/constants.js
ShGKme: But this size is 39 not because we want the default size to be 39 wherever it's used, but because we want it exactly in the list items

PR #13017, opened 2024-08-16
fix(SelectableParticipant): restyle component

Antreesy:
### ☑️ Resolves

* Fix appearance of component
	* visual - should look like NcListItem
	* a11y - should be an input with type: checkbox

🏚️ Before | 🏡 After
-- | --
image | image

### 🚧 Tasks

- [ ] Follow-up: use component in ParticipantsList (if used as a search result)

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/BreakoutRoomsEditor/SelectableParticipant.vue
+++ b/src/components/BreakoutRoomsEditor/SelectableParticipant.vue
@@ -4,35 +4,46 @@
 -->
 
 <template>
-	<div tabindex="0"
-		class="selectable-participant">
-		<input id="participant.attendeeId"
-			v-model="modelProxy"
+	<label class="selectable-participant">
+		<input v-model="modelProxy"
 			:value="participant.attendeeId"
 			type="checkbox"
-			name="participant.attendeeId">
+			class="selectable-participant__checkbox">
 		<!-- Participant's avatar -->
 		<AvatarWrapper :id="participant.actorId"
 			:name="participant.displayName"
 			:source="participant.source || participant.actorType"
 			disable-menu
 			disable-tooltip
-			show-user-status
-			show-user-status-compact />
-		<div>
-			{{ participant.displayName }}
-		</div>
-	</div>
+			show-user-status />
+
+		<span class="selectable-participant__content">
+			<span class="selectable-participant__content-name">
+				{{ participant.displayName }}
+			</span>
+			<span v-if="participantStatus"
+				class="selectable-participant__content-subname">
+				{{ participantStatus }}
+			</span>
+		</span>
+
+		<IconCheck class="selectable-participant__check-icon" :size="20" />
+	</label>
 </template>
 
 <script>
+import IconCheck from 'vue-material-design-icons/Check.vue'
+
 import AvatarWrapper from '../AvatarWrapper/AvatarWrapper.vue'
+
+import { getStatusMessage } from '../../utils/userStatus.js'
 
 export default {
 	name: 'SelectableParticipant',
 
 	components: {
 		AvatarWrapper,
+		IconCheck,
 	},
 
 	props: {
@@ -61,6 +72,10 @@ export default {
 				this.$emit('update:checked', value)
 			},
 		},
+
+		participantStatus() {
+			return getStatusMessage(this.participant)
+		},
 	},
 }
 </script>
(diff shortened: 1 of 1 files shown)


src/components/BreakoutRoomsEditor/SelectableParticipant.vue
```diff
@@ -4,35 +4,47 @@
 -->
 
 <template>
-	<div tabindex="0"
-		class="selectable-participant">
-		<input id="participant.attendeeId"
+	<label :for="participant.attendeeId"
+		class="selectable-participant"
+		:class="{ 'selectable-participant--selected': selected }">
+		<input :id="participant.attendeeId"
```
ShGKme, line 10: It's a but dangerous to set `:id="participant.attendeeId"`, risky to duplicate IDs on the page.

But `for/id` are not needed here at all, because `input` is a descendant of the `label` here.

src/components/BreakoutRoomsEditor/SelectableParticipant.vue
```diff
@@ -4,35 +4,47 @@
...
+			<span class="selectable-participant__content-name">
+				{{ participant.displayName }}
+			</span>
+			<span v-if="participant.statusMessage"
+				class="selectable-participant__content-subname">
+				{{ participant.statusIcon + ' ' + participant.statusMessage }}
+			</span>
+		</span>
+
+		<IconCheck v-if="selected" class="selectable-participant__check-icon" :size="20" />
```
ShGKme, line 33: It's sad, we cannot use Cannot we use `NcListItem` here? Because of the check mark? Even with extra actions slot?

src/components/BreakoutRoomsEditor/SelectableParticipant.vue
```diff
@@ -61,6 +73,10 @@ export default {
 				this.$emit('update:checked', value)
 			},
 		},
+
+		selected() {
+			return this.checked.includes(this.participant.attendeeId)
+		},
```
ShGKme, line 79: Or we can show it using CSS from `:checked` input

src/components/BreakoutRoomsEditor/SelectableParticipant.vue
```diff
@@ -4,35 +4,47 @@
...
-		<input id="participant.attendeeId"
+	<label :for="participant.attendeeId"
+		class="selectable-participant"
+		:class="{ 'selectable-participant--selected': selected }">
+		<input :id="participant.attendeeId"
 			v-model="modelProxy"
 			:value="participant.attendeeId"
 			type="checkbox"
-			name="participant.attendeeId">
+			class="hidden-visually">
```
ShGKme, line 14: We probably need a new class here, but this `hidden-visually` is not valid for a11y for interactive element. It is only valid for visually hidden text description.

It makes an input a part of the page but located in a random far place. Side effects:
- When using virtual cursor from assistive technologies, it may move cursor to the actual checkbox position on the page (completely random place)
- When focusing to the element it may:
  - if we are lucky and we are in a scrollable container - just don't scroll
  - if we are not and we are not in a scroll non-scrollable container - scroll the container and broke it
  
To reproduce:
1. Have many users
2. Make it small
3. Move via Tab

Expected - scrolls moves with you

Actual - it only moves when we focus 

Actual | Expected
---|---
bug | no-bug

ShGKme: A correct solution is to create a new class with similar styles, but:
- Make `label`: `position: relative`, so the `input` is inside the label
- Don't make `input` moved away with `top: 10000px`, just place it somewhere in the list item in a meaningful place
- Add `z-index: -1` to visually hide it.

Additionally we could make it reasonable size in a specific position to mimic real checkbox, but in this element there is no place for it. No visual element that looks like a checkbox.

To, putting it in the `label` is enough I'd say.

ShGKme (changes requested):
Thanks! Looks better and more accessible. but some changes are required:

- A problem with ID
- A problem with input position

src/components/BreakoutRoomsEditor/SelectableParticipant.vue
Antreesy: I thought the problem was in the `<a>` tag? Otherwise we can try to hide input in extra slot, I think

Antreesy: Postponing until NcListItem is a11y-aproppriate to do that. Then we could migrate this component as well

src/components/BreakoutRoomsEditor/SelectableParticipant.vue
```diff
@@ -113,6 +103,28 @@ export default {
...
+			content: "";
+			/* short arm */
+			width: 5px;
+			border-bottom: var(--border-width-input-focused) solid var(--color-main-text);
+			/* long arm */
+			height: 11px;
+			border-right: var(--border-width-input-focused) solid var(--color-main-text);
+			/* rotate to make it a check icon */
+			transform: translateY(-50%) rotate(45deg);
+		}
```
ShGKme, line 120: By CSS I meant, just show/hide the icon using `:checked` and `display: none` CSS, not fully implement an icon 🙈

Antreesy: 🙈

src/components/BreakoutRoomsEditor/SelectableParticipant.vue
DorraJaouad: Yet, we are using NcListItem in selectable participants in new conversation dialog. Can we use the same ?

Antreesy: Want to do it with separate pr, to split the changes

ShGKme (approved):
Thanks!

PR #13030, opened 2024-08-19
fix: correct userGroups intersection check for federation access

sanskar-soni-9:
Fixed `isFederatedEnabledForUser`, which was returning true when the intersection of `allowedGroups` and `userGroups` was empty. Now it correctly returns true when the intersection is not empty.

### 🏁 Checklist

### DIFF
--- a/lib/Config.php
+++ b/lib/Config.php
@@ -108,7 +108,7 @@ class Config {
 		}
 
 		$userGroups = $this->groupManager->getUserGroupIds($user);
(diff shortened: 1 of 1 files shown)


sanskar-soni-9:
also it returns `true` when there are no `allowedGroups` is that intended?

nickvergessen:
> also it returns true when there are no allowedGroups is that intended?

Yeah, that is basically the default

nickvergessen (approved):

PR #13086, opened 2024-08-22
 fix(conversations): Remove call-permissions

### LINKED ISSUE #13012: Remove call permissions

nickvergessen:
### ☑️ Resolves

* Based on #13072
* Fix #13012 

The permission only existed on API level and it's too complicated
to figure out a good UX for it. So the functionality to change it
are removed here. It's still kept in the database and response to
avoid breakage of existing clients

### 🏁 Checklist


danxuliu (commented):
[`forcePermissions` parameter in `joinCall`](https://github.com/nextcloud/spreed/blob/4d73f9bdb6a4fbcada24571fa84eb4ef021287f0/lib/Controller/CallController.php#L119) should be deprecated, as [it is used to set the now deprecated call permissions](https://github.com/nextcloud/spreed/blob/4d73f9bdb6a4fbcada24571fa84eb4ef021287f0/lib/Controller/CallController.php#L169-L171).

Same for [the whole `setAllAttendeesPermissions` endpoint](https://github.com/nextcloud/spreed/blob/77efdf6b189736a85930f6e47793bde9c8f636d7/lib/Controller/RoomController.php#L2174-L2175) (which it is in the docs, although it would be good to mark it too in the code documentation so it is picked by OpenAPI).

nickvergessen:
> [forcePermissions parameter in joinCall](https://github.com/nextcloud/spreed/blob/4d73f9bdb6a4fbcada24571fa84eb4ef021287f0/lib/Controller/CallController.php#L119)

It never made it to the normal docs :D So I will remove it, but not deprecate it (as additional send parameters don't break anyway)

danxuliu (approved):
Code looks good :+1:

PR #13092, opened 2024-08-22
fix(federation): Propagate permission changes to federated servers

### LINKED ISSUE #12953: Federated users are not aware of changed permissions

nickvergessen:
### ☑️ Resolves

* Replaces #12979
* Based on #13086
* Fix #12953  

## 🛠️ API Checklist

Modifications to original PR:
- 1 commit: Only propagate to the affected participant and extended the test with something that failed before and now passes
- 2 commit: Dropped as the permissions is now "simple" like all other room properties as the chained and other parts got dropped with #13086 
- 3 commit: Removed references to call permissions
- 4 commit: Removed references to call permissions

### 🏁 Checklist


danxuliu (approved):
I will rebase and also fix the problem from my comment once https://github.com/nextcloud/spreed/pull/13086 is merged.

**Edit:** force pushed with `[skip-ci]` in the last commit in case the base PR is merged while I am way.

PR #13111, opened 2024-08-23
fix(userStatus): migrate util to Typescript

Antreesy:
### ☑️ Resolves

* Prerequisite to #12850
* Allow to use util with search results

image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/BreakoutRoomsEditor/SelectableParticipant.vue
+++ b/src/components/BreakoutRoomsEditor/SelectableParticipant.vue
@@ -15,6 +15,7 @@
 			:source="participant.source || participant.actorType"
 			disable-menu
 			disable-tooltip
+			:preloaded-user-status="preloadedUserStatus"
 			show-user-status />
 
 		<span class="selectable-participant__content">
@@ -36,7 +37,7 @@ import IconCheck from 'vue-material-design-icons/Check.vue'
 
 import AvatarWrapper from '../AvatarWrapper/AvatarWrapper.vue'
 
-import { getStatusMessage } from '../../utils/userStatus.js'
+import { getPreloadedUserStatus, getStatusMessage } from '../../utils/userStatus.js'
 
 export default {
 	name: 'SelectableParticipant',
@@ -71,6 +72,10 @@ export default {
 			set(value) {
 				this.$emit('update:checked', value)
 			},
+		},
+
+		preloadedUserStatus() {
+			return getPreloadedUserStatus(this.participant)
 		},
 
 		participantStatus() {
--- a/src/components/BreakoutRoomsEditor/SelectableParticipant.vue
+++ b/src/components/BreakoutRoomsEditor/SelectableParticipant.vue
@@ -37,7 +37,7 @@ import IconCheck from 'vue-material-design-icons/Check.vue'
 
 import AvatarWrapper from '../AvatarWrapper/AvatarWrapper.vue'
 
-import { getPreloadedUserStatus, getStatusMessage } from '../../utils/userStatus.js'
+import { getPreloadedUserStatus, getStatusMessage } from '../../utils/userStatus.ts'
 
 export default {
 	name: 'SelectableParticipant',
--- a/src/components/RightSidebar/Participants/Participant.vue
+++ b/src/components/RightSidebar/Participants/Participant.vue
@@ -371,7 +371,7 @@ import {
 import { hasTalkFeature } from '../../../services/CapabilitiesManager.ts'
 import { formattedTime } from '../../../utils/formattedTime.ts'
 import { readableNumber } from '../../../utils/readableNumber.ts'
-import { getPreloadedUserStatus, getStatusMessage } from '../../../utils/userStatus.js'
+import { getPreloadedUserStatus, getStatusMessage } from '../../../utils/userStatus.ts'
 
 export default {
 	name: 'Participant',
--- a/src/components/RightSidebar/Participants/Participant.vue
+++ b/src/components/RightSidebar/Participants/Participant.vue
@@ -371,7 +371,7 @@ import {
 import { hasTalkFeature } from '../../../services/CapabilitiesManager.ts'
 import { formattedTime } from '../../../utils/formattedTime.ts'
 import { readableNumber } from '../../../utils/readableNumber.ts'
-import { getStatusMessage } from '../../../utils/userStatus.js'
+import { getPreloadedUserStatus, getStatusMessage } from '../../../utils/userStatus.js'
 
 export default {
 	name: 'Participant',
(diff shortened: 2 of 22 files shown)


src/utils/userStatus.ts
```diff
@@ -0,0 +1,66 @@
...
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+import { t } from '@nextcloud/l10n'
+
+import type {
+	Conversation,
+	Participant,
+	ParticipantSearchResult,
+	ParticipantStatus,
+} from '../types'
```
ShGKme, line 13:
```suggestion
} from '../types/index.ts'
```

src/utils/userStatus.ts
```diff
@@ -0,0 +1,66 @@
...
+		return userData.status as ParticipantStatus
+	} else {
+		return {
+			status: (userData as Participant).status,
+			message: (userData as Participant).statusMessage,
+			icon: (userData as Participant).statusIcon,
+			clearAt: (userData as Participant).statusClearAt,
+		}
+	}
+}
```
ShGKme, line 32: Using so many type assertions makes TS not useful because it doesn't check anything anymore

src/utils/userStatus.ts
```diff
@@ -0,0 +1,66 @@
...
+type AnyDataWithUserStatus = Conversation | Participant | ParticipantSearchResult
+
+/**
+ * Parse user status data from conversation, participant or search result to a single type object
+ * @param userData current user data
+ */
+function parseUserStatus(userData: AnyDataWithUserStatus): ParticipantStatus {
+	if (typeof userData.status === 'object') {
+		return userData.status as ParticipantStatus
+	} else {
```
ShGKme, line 24: No need for type assertion here. It has correct actual type

```suggestion
	if (typeof userData.status === 'object') {
		return userData.status
	} else {
```

src/utils/userStatus.ts
ShGKme: For example, it allows the function with `ParticipantSearchResult`.

ParticipantSearchResult has `status: ParticipantStatus | string`.

So it does to else and `status` can be `string`. Type assertion hides this case. Why? Is it an issue and a lost case, or there is something hidden that guarantees that `ParticipantSearchResult` has `status` only as `ParticipantStatus` but not string?

Antreesy: We have this JS check in component for similar purposes:
```js
if (Object.prototype.hasOwnProperty.call(this.participant, 'statusMessage')) {
	// We preloaded the status when via participants API
	return {
		status: this.participant.status || null,
		message: this.participant.statusMessage || null,
		icon: this.participant.statusIcon || null,
	}
}
if (Object.prototype.hasOwnProperty.call(this.participant, 'status')) {
	// We preloaded the status when via search API
	return {
		status: this.participant.status.status || null,
		message: this.participant.status.message || null,
		icon: this.participant.status.icon || null,
	}
}
```
I wonder if it's better to prepare an object for TS util before calling it?

ShGKme: The question is, is it actually guaranteed (by some logic) that here function is called only with `ParticipantSearchResult` with object status and never with the string one?

For example, if we know when `ParticipantSearchResult` has object status and when string.

If yes - we should adjust the type.

If no - we need to add an explicit check like `'statusMessage' in userData`.

Antreesy: Yes, we have:
- status: string (offline, dnd, away) from partcipiants API / conversation API (statusMessage is present)
- status: object from search API (for users) (status.message is present)
- status string (empty '') from search API (e.g. for groups) (no other fields)

ShGKme: > * status string (empty '') from search API (e.g. for groups) (no other fields)

But `ParticipantSearchResult` in new type says `status: ParticipantStatus | string,`

Antreesy: then I'll change to ''

src/utils/userStatus.ts
Antreesy: 😕 Either we:
- keep it as-is (Typescript works and doesn't complain for me)
- or adjust eslint and change in all occurences (~14 for now):
image

ShGKme: > * or adjust eslint

This message is not from ESLint, but from PhpStorm with Node.js (`require`) style import.

image

ShGKme: It works, because both Webpack and TS by default uses old import processing - like it is not ESM but CJS (`require` from Node.js). But in web and native Nodejs ESM, `imports` must have exact path. This is also why modern tools like Vite ask to use full path even there is a support to resolve extensions or index.

src/components/RightSidebar/Participants/Participant.vue
```diff
@@ -844,17 +844,19 @@ export default {
...
 					message: this.participant.statusMessage || null,
 					icon: this.participant.statusIcon || null,
+					clearAt: this.participant.statusClearAt || null,
 				}
 			}
 			if (Object.prototype.hasOwnProperty.call(this.participant, 'status')) {
 				// We preloaded the status when via search API
 				return {
-					status: this.participant.status.status || null,
+					status: this.participant.status.status || '',
```
Antreesy, line 856: NcAvatar expects `null` if there's nothing for `hasStatus`, so it's better to revert this

src/components/BreakoutRoomsEditor/SelectableParticipant.vue
```diff
@@ -15,6 +15,7 @@
 			:source="participant.source || participant.actorType"
 			disable-menu
 			disable-tooltip
+			:preloaded-user-status="preloadedUserStatus"
```
DorraJaouad, line 18: It's still a computed prop in `Participant`

Antreesy: Yes, but I'll get rid of it in a follow-up

Antreesy:
- rebased
- unify using of `getPreloadedUserStatus`
- add types

src/utils/userStatus.ts
```diff
@@ -0,0 +1,74 @@
...
+	}
+	return undefined
+}
+
+/**
+ * Generate full status message for user according to its status data
+ *
+ * @param userData user data (formatted via getPreloadedUserStatus() )
+ */
+export function getStatusMessage(userData?: ParticipantStatus | ''): string {
```
DorraJaouad, line 49: We are using this only with `getPreloadedUserStatus()`, it's better then to integrate them together from here.

Antreesy: true, there was an issue before in TS part. Now all good

DorraJaouad (approved):
Tested 🦅

PR #13121, opened 2024-08-23
fix(fileUpload): don't duplicate file templates in store

Antreesy:
### ☑️ Resolves

* Fix duplicating of templates

🏚️ Before | 🏡 After
-- | --
image | Screenshot from 2024-08-23 18-29-30

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [x] Talk Desktop

### DIFF
--- a/src/store/fileUploadStore.js
+++ b/src/store/fileUploadStore.js
@@ -204,7 +204,7 @@ const mutations = {
 		state.fileTemplatesInitialised = true
 	},
 
-	markFileTemplatesInitialisedForGuests(state) {
+	markFileTemplatesInitialised(state) {
 		state.fileTemplatesInitialised = true
 	},
 }
@@ -547,9 +547,15 @@ const actions = {
 	},
 
 	async getFileTemplates({ commit, getters }) {
+		if (getters.fileTemplates.length) {
+			console.debug('Skip file templates setup as already done')
+			commit('markFileTemplatesInitialised')
+			return
+		}
+
 		if (getters.getUserId() === null) {
 			console.debug('Skip file templates setup for participants that are not logged in')
-			commit('markFileTemplatesInitialisedForGuests')
+			commit('markFileTemplatesInitialised')
 			return
 		}
 
--- a/src/store/fileUploadStore.js
+++ b/src/store/fileUploadStore.js
@@ -102,7 +102,7 @@ const getters = {
 		return state.fileTemplatesInitialised
 	},
 
-	getFileTemplates: (state) => () => {
+	fileTemplates: (state) => {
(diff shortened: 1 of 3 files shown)


src/store/fileUploadStore.js
```diff
@@ -200,11 +200,12 @@ const mutations = {
 	},
 
 	storeFilesTemplates(state, { template }) {
-		state.fileTemplates.push(template)
+		const id = `${template.app}-${template.extension.replace('.', '')}`
+		Vue.set(state.fileTemplates, id, template)
```
ShGKme, line 204: Is it guaranteed to be a uniq pair?

Antreesy: Here's the server response, i don't see better candidates:
```json
 {
                "app": "text",
                "label": "New text file",
                "extension": ".md",
                "iconClass": null,
                "iconSvgInline": "\u003Csvg xmlns=\"http:\/\/www.w3.org\/2000\/svg\" width=\"24\" height=\"24\" viewBox=\"0 0 24 24\" fill=\"#000000\"\u003E\n\t\u003Cpath d=\"M0 0h24v24H0z\" fill=\"none\" \/\u003E\n\t\u003Cpath fill=\"#969696\"\n\t\td=\"M19 3H5c-1.1 0-2 .9-2 2v14c0 1.1.9 2 2 2h14c1.1 0 2-.9 2-2V5c0-1.1-.9-2-2-2zm-5 14H7v-2h7v2zm3-4H7v-2h10v2zm0-4H7V7h10v2z\" \/\u003E\n\u003C\/svg\u003E\n",
                "mimetypes": [
                    "text\/markdown",
                    "text\/plain"
                ],
                "ratio": 1,
                "actionLabel": "Create new text file",
                "templates": [...]
            },
            ```

DorraJaouad: It seems good 👍🏻

ShGKme: Checked the server code with @Antreesy in the office. It doesn't look like there cannot be 2 different files from one app with the same extension. Though I don't have an example from an app, I can imagine, let's say, creating Vue Options API and Vue Composition API file with `.vue` extension from a single app =D Filename as well as action label can be different.

So, I'd either use an entire object as a key (as a serialized, JSON string).

Or use a different method to prevent the problem. For example, replacing an array instead of pushing. This should also prevent from potential broken order, if we don't sort manually.

ShGKme (approved):

DorraJaouad (approved):
It makes sense 👍🏻 
I couldn't reproduce it though

PR #13124, opened 2024-08-23
fix(Avatar): Return dark avatar URL if the user has a dark theme selected

provokateurin:
In places like the Dashboard the light user avatar is always shown regardless of the user theme. This is quite a visual distraction to me. AFAICT this fix also affects other places where a room avatar is shown.

Unfortunately it is only possible to detect the dark theme if the user configured it explicitly. If the user only has the `default` aka system theme, then only the browser can now the final theme.

The only real fix would be to extend the places where an URL is required and provide one for light and one for dark theme separately, such that the frontend can decide which to pick (but that would need to be implemented both in the clients and every dashboard widget and so on and is probably not worth the effort).

### 🏁 Checklist

### DIFF
--- a/lib/Service/AvatarService.php
+++ b/lib/Service/AvatarService.php
@@ -12,6 +12,7 @@ namespace OCA\Talk\Service;
 use InvalidArgumentException;
 use OC\Files\Filesystem;
 use OCA\Talk\Room;
+use OCA\Theming\Service\ThemesService;
 use OCP\Files\IAppData;
 use OCP\Files\NotFoundException;
 use OCP\Files\SimpleFS\InMemoryFile;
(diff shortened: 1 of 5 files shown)


nickvergessen:
Ref https://github.com/nextcloud/server/issues/37272

provokateurin:
Yeah, I also mentioned that in the initial comment. I don't think anyone will work on fixing that soon and this PR at least improves the situation a little bit. Is there anything against merging this?

nickvergessen:
> Is there anything against merging this?

Yeah, it only fixes for forced dark mode and it will make Talk behave diffrent from anyone else, and e.g. break mobile and desktop apps which don't use the server's dark mode setting.
Let's just add the dark=1 optional parameter to those endpoints at the conf (or right now?) and then each provider can be fixed until 31 release

provokateurin:
Sure, let's take a look at it together at the conf :)

PR #13133, opened 2024-08-25
feat(error_message): Display proper message on error on join call

### LINKED ISSUE #13077: Error on "Join call" is not shown to user
### How to use GitHub

---

- Extracting from https://github.com/nextcloud/spreed/issues/13045

## Steps to reproduce
1. As a user try to join a call
2. Receive a 400 response, see patch below for hardcoded 400

```diff
diff --git a/lib/Controller/CallController.php b/lib/Controller/CallController.php
index 373337913..c8a6fe00f 100644
--- a/lib/Controller/CallController.php
+++ b/lib/Controller/CallController.php
@@ -136,6 +136,7 @@ class CallController extends AEnvironmentAwareController {
        #[RequireParticipant]
        #[RequireReadWriteConversation]
        public function joinCall(?int $flags = null, ?int $forcePermissions = null, bool $silent = false, bool $recordingConsent = false): DataResponse {
+               return new DataResponse(['error' => 'consent'], Http::STATUS_BAD_REQUEST);
                try {
                        $this->validateRecordingConsent($recordingConsent);
                } catch (\InvalidArgumentException) {
```

### Expected behaviour
You see an/the error message

### Actual behaviour
Browser warning about leaving the page:
grafik

Souptik2001:
### ☑️ Resolves

* Fix #13077

🏚️ Before | 🏡 After
-- | --
Screenshot 2024-08-25 at 12 46 30 PM | Screenshot 2024-08-25 at 12 50 25 PM

🏡 After (Screencast) -

https://github.com/user-attachments/assets/346f9108-b57f-4e1d-ab16-6eab2c47ed3e

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop
- [ ] ⛑️ Tests are included or not possible (**:warning: I have yet not added test for participant store. Once the code is confirmed, I will add it, to avoid repeated changes in test cases**)


SystemKeeper:
Should the message be dependent on the error?

Ref https://github.com/nextcloud/talk-ios/blob/81ecc92ada12ae88f9a2f2aa6c297fa5295111d8/NextcloudTalk/NCCallController.m#L230-L257

nickvergessen:
While that is a good starting point, 400 can also have different error messages, (list pending to be documented: consent, session, type, are known atm)

src/utils/signaling.js
```diff
@@ -281,16 +280,9 @@ Signaling.Base.prototype.joinCall = function(token, flags, silent, recordingCons
...
-					// will then show the proper error message to the user.
-					window.location = generateUrl('call/' + token)
-				} else {
-					// TODO: Is it true, reload is equal to generateUrl('call/' + token) here?
-					// Or can we always just reload the page?
-					window.location.reload()
-				}
+				store.commit('connectionFailed', {
+					token,
+				})
```
Antreesy, line 285: If we can get the meaningful error message from `axios.post` request and pass it to here, then it would be perfect.

Even if we get `error => 'consent'`, should be possible to interprete response and show relevant text as iOS does it

src/store/participantsStore.js
```diff
@@ -90,6 +92,9 @@ const getters = {
 	isConnecting: (state) => (token) => {
 		return !!(state.connecting[token] && Object.keys(state.connecting[token]).length > 0)
 	},
+	connectionFailed: (state) => (token) => {
+		return (state.connectionFailed[token] && Object.keys(state.connectionFailed[token]).length > 0) ?? false
```
Antreesy, line 96: With further planned migration to pinia, we could simply return the state value from the store. Even now, there's no need to transform it to boolean

```suggestion
		return state.connectionFailed[token]
```

src/utils/signaling.js
```diff
@@ -279,18 +278,12 @@ Signaling.Base.prototype.joinCall = function(token, flags, silent, recordingCons
...
-					// will then show the proper error message to the user.
-					window.location = generateUrl('call/' + token)
-				} else {
-					// TODO: Is it true, reload is equal to generateUrl('call/' + token) here?
-					// Or can we always just reload the page?
-					window.location.reload()
-				}
+				store.commit('connectionFailed', {
+					token,
+					payload: e.response?.data?.ocs?.data,
```
Antreesy, line 285: If we know it's going to be an error, we can pass it straight away

```suggestion
					payload: e.response?.data?.ocs?.data?.error
```

src/store/participantsStore.js
```diff
@@ -56,6 +56,20 @@ function emitUserStatusUpdated(participant) {
...
+ * @param {*} code Error message code
+ *
+ * @return {string} Error message
+ */
+function getConnectionFailedErrorMessage(code) {
+	if (code === 'consent') {
+		return t('spreed', 'Recording consent is required')
+	}
+	return t('spreed', 'Oops, something went wrong!')
+}
```
Antreesy, line 71: I'm not sure this belongs to the store. Maybe we'll move it to EmptyCallView component (in computed `message`, where it initially was)?

src/components/CallView/shared/EmptyCallView.vue
```diff
@@ -122,6 +132,9 @@ export default {
 		},
 
 		title() {
+			if (this.connectionFailed) {
+				return t('spreed', 'Connection Failed')
```
Antreesy, line 136:
```suggestion
				return t('spreed', 'Connection failed')
```

src/store/participantsStore.js
```diff
@@ -56,6 +56,20 @@ function emitUserStatusUpdated(participant) {
...
+ *
+ * @param {*} code Error message code
+ *
+ * @return {string} Error message
+ */
+function getConnectionFailedErrorMessage(code) {
+	if (code === 'consent') {
+		return t('spreed', 'Recording consent is required')
+	}
+	return t('spreed', 'Oops, something went wrong!')
```
Antreesy, line 70: I'd keep it more formal, and easier to translate to other languages
```suggestion
	return t('spreed', 'Something went wrong')
```

Antreesy (commented):
Almost there! Just a couple of comments, and I believe, we can merge it:

src/store/participantsStore.js
Souptik2001: Yes makes sense. 👍  I just copied it from Microsoft Teams. 😅

src/store/participantsStore.js
Souptik2001: @Antreesy Did you mean to write this logic directly in the `message` function of `EmptyCallView` component, like this? - https://github.com/nextcloud/spreed/pull/13133/files#diff-3d0a2e872c65aa3113ac800fa38a10dbe0d15f932a9ebbd9c14adb34f3018dc7R158-R164

src/utils/signaling.js
```diff
@@ -279,18 +278,12 @@ Signaling.Base.prototype.joinCall = function(token, flags, silent, recordingCons
...
-					// will then show the proper error message to the user.
-					window.location = generateUrl('call/' + token)
-				} else {
-					// TODO: Is it true, reload is equal to generateUrl('call/' + token) here?
-					// Or can we always just reload the page?
-					window.location.reload()
-				}
+				store.commit('connectionFailed', {
+					token,
+					payload: e.response?.data?.ocs?.data?.error,
```
nickvergessen, line 285: should also log the erorr in the console, so it's retrievable by the user?

Souptik2001: @nickvergessen Just the error code is enough?

Something like this - 

```JS
console.error('Connection failed, reason: ', e.response?.data?.ocs?.data?.error)
```

nickvergessen: yes

nickvergessen:
Something more on the general side, maybe @Antreesy can answer:
When joining the conversation fails, why do we still toggle to the CallView and not stay on the DeviceChecker (or bring that up in case it was not shown)? Feels a bit weird as the next thing the user has to do is "Leave call" although they never managed to join in first place, to then try again.

Antreesy:
> When joining the conversation fails

Joining the call, you meant?

It was like this since the code appeared. Just tried to change places, but it feels weird (matter of habit, maybe): several seconds nothings happens, then callView appears.

>  stay on the DeviceChecker (or bring that up)

Make sense for some cases (e.g. recording consent isn't given), but in most cases we still need to reload?

Even if we'll change the order, still can find the use to this PR changes (e.g. for failures during the call)

src/store/participantsStore.js
Antreesy: Yeah, that's more than enough =)

Antreesy (approved):

PR #13135, opened 2024-08-26
docs: Fix values for used recording consent options

### LINKED ISSUE #13085: Recording consent property is inconsistent
### How to use GitHub

---

API on room level should only return 0/1, but seems to be able to also return 2. 

## Talk app

**Talk app version:** 20 RC2

nickvergessen:
### ☑️ Resolves

* Fix #13085 
* The other part was a misunderstanding. I thought as we return the talkConfig value here it could be 2, but the 2 is the first part of the check which will return the room option: https://github.com/nextcloud/spreed/blob/956a1ab00ffb25d87c0f5192333c9b66e56a63fe/lib/Service/RoomFormatter.php#L141
* OpenAPI for this part is in https://github.com/nextcloud/spreed/pull/12783#discussion_r1730962029

### 🏁 Checklist


Antreesy (commented):
LGTM

Antreesy (approved):

PR #13149, opened 2024-08-27
fix(federation): ensure federation invites appear in store

### LINKED ISSUE #13146: Can not accept federated invites in notification
### How to use GitHub

---

## Steps to reproduce
1. Receive a federated invitation
2. Load the talk frontpage (`index.php/apps/spreed/`) and do **NOT** open the "Pending invitations"
3. Open the notification area in top right
4. Click on "Accept" or "Decline"

### Expected behaviour
Works :accept: 

### Actual behaviour
```
TypeError: conversation is undefined
    interceptNotificationActions App.vue:419
    emit index.mjs:50
    emit index.mjs:48
    emit _plugin-vue2_normalizer-BJx_e0Vj.chunk.mjs:7
    k0 _plugin-vue2_normalizer-BJx_e0Vj.chunk.mjs:7
    onClickActionButton NotificationsApp-Sd_-k7u9.chunk.mjs:29
```

After some debugging I found out that `this.pendingShares` is empty if you never opened the "Pending invitations". Then:
https://github.com/nextcloud/spreed/blob/82450e5c7017c8ae874c9b9d02ff3f90a60e7046/src/stores/federation.ts#L106-L108
returns without anything and => `conversation is undefined`.

I know I reported this before (or was it only in the chat) but @Antreesy could not reproduce. But now I assume he simply had opened the pending invitations list before accepting on the notification

Antreesy:
### ☑️ Resolves

* Fix #13146
  * Works now after reload / accept from notification
  * If already in the store - early return, so doesn't affect performance

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -543,6 +543,7 @@ export default {
 				if (event.notification.objectType === 'remote_talk_share') {
 					try {
 						event.cancelAction = true
+						this.federationStore.addInvitationFromNotification(event.notification)
 						const conversation = await this.federationStore.acceptShare(event.notification.objectId)
 						if (conversation.token) {
 							this.$store.dispatch('addConversation', conversation)
@@ -559,7 +560,8 @@ export default {
 				if (event.notification.objectType === 'remote_talk_share') {
 					try {
 						event.cancelAction = true
-						this.federationStore.rejectShare(event.notification.objectId)
+						this.federationStore.addInvitationFromNotification(event.notification)
+						await this.federationStore.rejectShare(event.notification.objectId)
 					} catch (error) {
 						console.error(error)
 					}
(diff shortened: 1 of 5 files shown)


src/App.vue
```diff
@@ -559,6 +560,7 @@ export default {
 				if (event.notification.objectType === 'remote_talk_share') {
 					try {
 						event.cancelAction = true
+						this.federationStore.addInvitationFromNotification(event.notification)
 						this.federationStore.rejectShare(event.notification.objectId)
```
nickvergessen, line 564: await btw?

nickvergessen (approved):
The notification is not fading at the moment. I think this needs some more work on the notifications app and then here.
We need a way to "cancelAction but with success" and that should then still call https://github.com/nextcloud/notifications/blob/535046031341e5caaa2a256c4e9da9345741713c/src/Components/Action.vue#L127-L129

PR #13266, opened 2024-09-11
chore(SettingsAPI): replace reactive with ref to remove Vue 2 warn

ShGKme:
Changes nothing in how it works.

Gets rid of the annoying Vue 2 warning about `reactive(array)` behavior on Vue 2 vs Vue 3 in `watch` (which we even don't use).

image

### DIFF
--- a/src/services/SettingsAPI.ts
+++ b/src/services/SettingsAPI.ts
@@ -3,7 +3,8 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-import { readonly, shallowReactive } from 'vue'
+import { readonly, ref, markRaw } from 'vue'
+import type { Ref } from 'vue'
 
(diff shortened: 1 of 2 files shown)


DorraJaouad (approved):

src/services/SettingsAPI.ts
```diff
@@ -22,24 +23,25 @@ type TalkSettingsSection = {
 	element: string
 }
 
-const customSettingsSections: TalkSettingsSection[] = shallowReactive([])
+// TODO: use shallowReactive instead of ref + markRaw in Vue 3 (see file commit history)
+const customSettingsSections: Ref<TalkSettingsSection[]> = ref([])
```
ShGKme, line 27: `shallowReactive` has no real advantages except that it allows to get rid of `markRaw` in just one function.

PR #13269, opened 2024-09-11
fix(avatar): don't use 'user/avatar' endpoint for generating a temp avatar

### LINKED ISSUE #13208: Set conversation avatar updates account avatar instead
### How to use GitHub

---

Quite an edge case, but still happens under conditions described below.

## Steps to reproduce
1. Upload square picture via Files app (width = height)
2. Try to set conversation avatar "from Nextcloud"
3. See result

### Expected behaviour
Cropper should open (to show how picture would look like, and maybe crop some more)

### Actual behaviour

Frontend logic copied from server settings / AvatarSection:
https://github.com/nextcloud/spreed/blob/bd84203ddd2672fc017fa9883a315a0da2e20445/src/components/ConversationSettings/ConversationAvatarEditor.vue#L278-L288

If image is square, cropping isn't requested, and image is set by default endpoint => as avatar image
https://github.com/nextcloud/server/blob/af6de04e9e141466dc229e444ff3f146f4a34765/core/Controller/AvatarController.php#L243-L254

## Talk app

**Talk app version:** (main, but given server/frontend code is unchanged, also valid for 17)

Antreesy:
### ☑️ Resolves

* Fix #13208

🏚️ Before | 🏡 After
-- | --
image | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [x] Talk Desktop

### DIFF
--- a/src/components/ConversationSettings/ConversationAvatarEditor.vue
+++ b/src/components/ConversationSettings/ConversationAvatarEditor.vue
@@ -110,8 +110,6 @@ import Folder from 'vue-material-design-icons/Folder.vue'
 import Palette from 'vue-material-design-icons/Palette.vue'
 import Upload from 'vue-material-design-icons/Upload.vue'
 
-import { getRequestToken } from '@nextcloud/auth'
-import axios from '@nextcloud/axios'
 import { showError } from '@nextcloud/dialogs'
 import { FilePickerVue } from '@nextcloud/dialogs/filepicker.js'
 import { t } from '@nextcloud/l10n'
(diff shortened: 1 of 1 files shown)


src/components/ConversationSettings/ConversationAvatarEditor.vue
```diff
@@ -268,24 +268,17 @@ export default {
...
-					this.$refs.cropper.replace(tempAvatar)
-					this.showCropper = true
-				} else {
-					showError(data.data.message)
-					this.cancel()
-				}
+				const tempAvatar = generateUrl(`/core/preview?fileId=${fileid}&x=512&y=512&a=1`)
+				this.loading = false
+				this.$refs.cropper.replace(tempAvatar)
+				this.showCropper = true
```
DorraJaouad, line 281: Do we still need loading flags here?

DorraJaouad (approved):
Tested

src/components/ConversationSettings/ConversationAvatarEditor.vue
Antreesy: Looks like a no, we don't wait for server response here (even for preview load)

DorraJaouad (approved):
Tested

nickvergessen (approved):

PR #13290, opened 2024-09-13
fix(changelog): Don't check the user changelog twice

nickvergessen:
### ☑️ Resolves

* Related
grafik

### Explanation

There was a simple race condition previously with 2 requests R1 and R2:

- R1: handle()
- R1: userHasNewChangelog() compares X0 with count()
- R2: handle()
- R2: userHasNewChangelog() compares X0 with count()
- R1: updateChangelog()
- R1: getChangelogForUser() reads X0
- R1: setUserValue() sets X1 with precondition X0
- R2: updateChangelog()
- R2: getChangelogForUser() reads X1 :boom: 
- R2: setUserValue() sets X1 with precondition X1

### 🏁 Checklist

### DIFF
--- a/lib/Chat/Changelog/Listener.php
+++ b/lib/Chat/Changelog/Listener.php
@@ -32,10 +32,6 @@ class Listener implements IEventListener {
 			return;
 		}
 
-		if (!$this->manager->userHasNewChangelog($event->getUserId())) {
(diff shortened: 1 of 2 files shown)


Antreesy (approved):
LGTM

PR #13302, opened 2024-09-14
fix: move page visibility check to composable

Antreesy:
### ☑️ Resolves

* Move a value handling from the store

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -59,6 +59,7 @@
 <script>
 import debounce from 'debounce'
 import uniqueId from 'lodash/uniqueId.js'
+import { computed } from 'vue'
 
 import Message from 'vue-material-design-icons/Message.vue'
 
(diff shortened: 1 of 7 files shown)


src/components/MessagesList/MessagesList.vue
```diff
@@ -121,10 +123,13 @@ export default {
 
 	emits: ['update:is-chat-scrolled-to-bottom'],
 
-	setup() {
+	setup(props) {
+		const isWindowVisible = useWindowVisibility()
+		const isChatVisible = computed(() => isWindowVisible.value && props.isVisible)
```
DorraJaouad, line 128: It could be a normal computed

DorraJaouad (approved):
Perfect!

src/store/windowVisibilityStore.js
ShGKme, line 1: I'd rename to `useDocumentVisibility`, because visibility is a `document` property (it is checked and subscribed on the document).

And this composable is usually called `useDocumentVisibility` in util libraries.

PR #13311, opened 2024-09-15
fix: migrate fullscreen store and related functions/listeners to composable

Antreesy:

No visual changes

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


src/composables/useDocumentFullscreen.ts
```diff
@@ -0,0 +1,74 @@
...
+	})
+
+	return readonly(isFullscreen)
+}
+
+/**
+ * Enable a fullscreen with Fullscreen API
+ */
+export async function enableFullscreen() {
+	const element = document.getElementById('content-vue')
```
DorraJaouad, line 42:
```suggestion
	const element = document.getElementById('content-vue')
	
	if (!element) {
		return
	}
```

src/composables/useDocumentFullscreen.ts
```diff
@@ -0,0 +1,74 @@
...
+	} else if (element.msRequestFullscreen) {
+		await element.msRequestFullscreen()
+	}
+}
+
+/**
+ * Disable a fullscreen
+ */
+export async function disableFullscreen() {
+	if (document.exitFullscreen) {
```
DorraJaouad, line 59:
```suggestion
	if (document?.exitFullscreen) {
```

src/composables/useDocumentFullscreen.ts
```diff
@@ -0,0 +1,74 @@
...
+	const changeIsFullscreen = () => {
+		isFullscreen.value = document.fullscreenElement !== null
+	}
+
+	onBeforeMount(() => {
+		document.addEventListener('fullscreenchange', changeIsFullscreen)
+		document.addEventListener('mozfullscreenchange', changeIsFullscreen)
+		document.addEventListener('MSFullscreenChange', changeIsFullscreen)
+		document.addEventListener('webkitfullscreenchange', changeIsFullscreen)
+	})
```
ShGKme, line 26: Same as
```suggestion
document.addEventListener('fullscreenchange', changeIsFullscreen)
document.addEventListener('mozfullscreenchange', changeIsFullscreen)
document.addEventListener('MSFullscreenChange', changeIsFullscreen)
document.addEventListener('webkitfullscreenchange', changeIsFullscreen)
```

src/composables/useDocumentFullscreen.ts
```diff
@@ -0,0 +1,74 @@
...
+
+	if (element.requestFullscreen) {
+		await element.requestFullscreen()
+	} else if (element.webkitRequestFullscreen) {
+		await element.webkitRequestFullscreen(Element.ALLOW_KEYBOARD_INPUT)
+	} else if (element.mozRequestFullScreen) {
+		await element.mozRequestFullScreen()
+	} else if (element.msRequestFullscreen) {
+		await element.msRequestFullscreen()
+	}
```
ShGKme, line 52: We only need `webkit` prefix for Safari now

https://caniuse.com/?search=requestFullscreen

src/composables/useDocumentFullscreen.ts
ShGKme: `document` should be defined. We don't have SSR and we won't use this composable in a worker

src/composables/useDocumentFullscreen.ts
ShGKme, line 42: In the current implementation, it enabled Fullscreen specifically on `#content-vue`, not a general fullscreen

ShGKme (approved):

src/composables/useDocumentFullscreen.ts
Antreesy: Will be changed in follow-up

ShGKme (approved):

Issue #13160, opened 2024-08-28
follow-up(CallView): connection state visual enhancements


Antreesy:
Some follow-ups for the CallView:
- [x] replace `<div class="iconClass" />` with MDI icons used across the project: https://pictogrammers.com/library/mdi/icon/alert-octagon/
  - [x] Make sure color is consistent on light-dark theme
- [x] add statusCode from response to `connectionFailed` state (Dorra)
  - [x] apply translations from iOS repo
  - [ ] check qualityWarningTooltip for reuse at the same place https://github.com/nextcloud/spreed/blob/main/src/components/TopBar/TopBarMediaControls.vue#L323
- [x] Think about https://github.com/nextcloud/spreed/pull/13133#issuecomment-2315134339: intermediate state while connecting / do not switch to CallView at all

_Originally posted by @Antreesy in https://github.com/nextcloud/spreed/issues/13133#issuecomment-2315209527_

DorraJaouad:
>  intermediate state while connecting / do not switch to CallView at all

We can put a modal showing loading status until we receive a response from join call request. If it's success, modal goes away as the call view appears and if it's an error, the modal stays there showing the error message. wdyt ?

DorraJaouad:
@Antreesy do we close this?

PR #13329, opened 2024-09-17
fix(Call): introduce intermediary phase before joining call

DorraJaouad:
* Part of #13160

Before

https://github.com/user-attachments/assets/85d9db4c-d9fc-4ad1-bb9c-5c4ddf458121

After

https://github.com/user-attachments/assets/e108bb3f-2321-481f-a6d9-4d14a8e34bf9

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/CallView/CallFailedDialog.vue
+++ b/src/components/CallView/CallFailedDialog.vue
@@ -1,0 +1,72 @@
+<!--
+  - SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<script setup lang="ts">
+import { computed } from 'vue'
+
+import IconAlertOctagon from 'vue-material-design-icons/AlertOctagon.vue'
+
+import { t } from '@nextcloud/l10n'
+
+import NcEmptyContent from '@nextcloud/vue/dist/Components/NcEmptyContent.js'
+import NcModal from '@nextcloud/vue/dist/Components/NcModal.js'
+
+import { useStore } from '../../composables/useStore.js'
+
+const store = useStore()
+
+const props = defineProps({
+	token: {
+		type: String,
+		required: true,
+	},
+})
+
+const STATUS_ERRORS = {
+	400: t('spreed', 'Recording consent is required'),
+	403: t('spreed', 'This conversation is read-only'),
+	404: t('spreed', 'Conversation not found or not joined'),
+	412: t('spreed', "Lobby is still active and you're not a moderator"),
+}
+const userId = computed(() => store.getters.userId)
+const connectionFailed = computed(() => store.getters.connectionFailed(props.token))
+const connectionFailedDialogId = `connection-failed-${userId.value}`
+const message = computed(() => {
+	if (connectionFailed.value) {
+		const statusCode = connectionFailed.value?.meta?.statuscode
+		if (STATUS_ERRORS[statusCode]) {
+			return STATUS_ERRORS[statusCode]
+		}
+		if (connectionFailed.value?.data?.error) {
+			return connectionFailed.value.data.error
+		}
+
+		return t('spreed', 'Please try to reload the page')
+	} else {
+        return ''
+    }
+})
+
+/**
+ *
+ */
+function clearConnectionFailedError() {
+	store.dispatch('clearConnectionFailed', props.token)
+}
+
+</script>
+
+<template>
+	<NcModal class="connection-failed__modal"
+		:label-id="connectionFailedDialogId"
+		@close="clearConnectionFailedError">
+		<NcEmptyContent :name="t('spreed', 'Connection failed')"
+			:description="message">
+			<template #icon>
+				<IconAlertOctagon />
+			</template>
+		</NcEmptyContent>
+	</NcModal>
+</template>
--- a/src/components/CallView/CallFailedDialog.vue
+++ b/src/components/CallView/CallFailedDialog.vue
@@ -45,8 +45,8 @@ const message = computed(() => {
 
 		return t('spreed', 'Please try to reload the page')
 	} else {
-        return ''
-    }
+		return ''
+	}
 })
 
 /**
--- a/src/components/CallView/CallFailedDialog.vue
+++ b/src/components/CallView/CallFailedDialog.vue
@@ -30,23 +30,22 @@ const STATUS_ERRORS = {
 	404: t('spreed', 'Conversation not found or not joined'),
 	412: t('spreed', "Lobby is still active and you're not a moderator"),
 }
-const userId = computed(() => store.getters.userId)
 const connectionFailed = computed(() => store.getters.connectionFailed(props.token))
-const connectionFailedDialogId = `connection-failed-${userId.value}`
+const connectionFailedDialogId = `connection-failed-${props.token}`
 const message = computed(() => {
-	if (connectionFailed.value) {
-		const statusCode = connectionFailed.value?.meta?.statuscode
-		if (STATUS_ERRORS[statusCode]) {
-			return STATUS_ERRORS[statusCode]
-		}
-		if (connectionFailed.value?.data?.error) {
-			return connectionFailed.value.data.error
-		}
-
-		return t('spreed', 'Please try to reload the page')
-	} else {
+	if (!connectionFailed.value) {
 		return ''
 	}
+
+	const statusCode = connectionFailed.value?.meta?.statuscode
+	if (STATUS_ERRORS[statusCode]) {
+		return STATUS_ERRORS[statusCode]
+	}
+	if (connectionFailed.value?.data?.error) {
+		return connectionFailed.value.data.error
+	}
+
+	return t('spreed', 'Please try to reload the page')
 })
 
 /**
@@ -59,8 +58,7 @@ function clearConnectionFailedError() {
 </script>
 
 <template>
-	<NcModal class="connection-failed__modal"
-		:label-id="connectionFailedDialogId"
+	<NcModal :label-id="connectionFailedDialogId"
 		@close="clearConnectionFailedError">
 		<NcEmptyContent :name="t('spreed', 'Connection failed')"
 			:description="message">
--- a/src/services/callsService.js
+++ b/src/services/callsService.js
@@ -32,11 +32,7 @@ import {
  * @return {number} The actual flags based on the available media
  */
 const joinCall = async function(token, flags, silent, recordingConsent) {
-	try {
-		return await signalingJoinCall(token, flags, silent, recordingConsent)
-	} catch (error) {
-		console.debug('Error while joining call: ', error)
-	}
+	return await signalingJoinCall(token, flags, silent, recordingConsent)
 }
 
 /**
--- a/src/services/callsService.js
+++ b/src/services/callsService.js
@@ -29,10 +29,10 @@ import {
  * @param {boolean} silent Whether the call should trigger a notifications and
  * sound for other participants or not
  * @param {boolean} recordingConsent Whether the participant gave their consent to be recorded
- * @return {number} The actual flags based on the available media
+ * @return {Promise<number>} The actual flags based on the available media
  */
 const joinCall = async function(token, flags, silent, recordingConsent) {
-	return await signalingJoinCall(token, flags, silent, recordingConsent)
+	return signalingJoinCall(token, flags, silent, recordingConsent)
 }
 
 /**
--- a/src/store/participantsStore.js
+++ b/src/store/participantsStore.js
@@ -821,10 +821,19 @@ const actions = {
 			return
 		}
 
-		commit('setInCall', {
-			token,
-			sessionId: participantIdentifier.sessionId,
-			flags,
+		// Preparing the event listener for the signaling-join-call event
+		EventBus.once('signaling-join-call', () => {
+			commit('setInCall', {
+				token,
+				sessionId: participantIdentifier.sessionId,
+				flags,
+			})
+			commit('finishedConnecting', { token, sessionId: participantIdentifier.sessionId })
+		})
+
+		// Preparing the event listener for the signaling-join-call-failed event
+		EventBus.once('signaling-join-call-failed', () => {
+			commit('finishedConnecting', { token, sessionId: participantIdentifier.sessionId })
 		})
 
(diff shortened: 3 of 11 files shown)


src/store/participantsStore.js
```diff
@@ -788,28 +792,31 @@ const actions = {
...
+		// Preparing the event listener for the signaling-join-call-failed event
+		EventBus.once('signaling-join-call-failed', () => {
 			commit('finishedConnecting', { token, sessionId: participantIdentifier.sessionId })
 		})
 
-		setTimeout(() => {
-			// If by accident we never receive a users list, just switch to
-			// "Waiting for others to join the call …" after some seconds.
-			commit('finishedConnecting', { token, sessionId: participantIdentifier.sessionId })
-		}, 10000)
```
DorraJaouad, line 812: It's guaranteed to receive one of the above signaling events as we are the base server

src/store/participantsStore.js
```diff
@@ -321,27 +321,29 @@ const mutations = {
 			if (state.inCall[token] && state.inCall[token][sessionId]) {
 				Vue.delete(state.inCall[token], sessionId)
 			}
-
-			if (state.connecting[token] && state.connecting[token][sessionId]) {
```
Antreesy, line 325: I think we can keep this resetting (in case it survives somehow). Maybe via `commit('finishedConnecting')`?

src/components/CallView/CallFailedDialog.vue
```diff
@@ -0,0 +1,70 @@
...
+import { computed } from 'vue'
+
+import IconAlertOctagon from 'vue-material-design-icons/AlertOctagon.vue'
+
+import { t } from '@nextcloud/l10n'
+
+import NcEmptyContent from '@nextcloud/vue/dist/Components/NcEmptyContent.js'
+import NcModal from '@nextcloud/vue/dist/Components/NcModal.js'
+
+import store from '../../store/index.js'
```
Antreesy, line 16: we have useStore composable, it should be better to use it here

src/services/callsService.js
```diff
@@ -36,6 +36,7 @@ const joinCall = async function(token, flags, silent, recordingConsent) {
 		return await signalingJoinCall(token, flags, silent, recordingConsent)
 	} catch (error) {
 		console.debug('Error while joining call: ', error)
+		throw error
```
Antreesy, line 39: maybe remove try-catch block here, since it's not needed?

Antreesy (commented):
Works good in general, but I'd test it in some real calls before merging

Antreesy:
- [ ] NcLoadingIcon isn't shown
- [ ] CallButon keeps disabled when leave a call, unless reload

DorraJaouad:
> * [ ]  NcLoadingIcon isn't shown
> * [ ]  CallButon keeps disabled when leave a call, unless reload

It was because `isConnecting` state is reinitiated at the time where it should be truthy

src/components/CallView/CallFailedDialog.vue
```diff
@@ -0,0 +1,72 @@
...
+
+<script setup lang="ts">
+import { computed } from 'vue'
+
+import IconAlertOctagon from 'vue-material-design-icons/AlertOctagon.vue'
+
+import { t } from '@nextcloud/l10n'
+
+import NcEmptyContent from '@nextcloud/vue/dist/Components/NcEmptyContent.js'
+import NcModal from '@nextcloud/vue/dist/Components/NcModal.js'
```
Antreesy, line 14: `NcDialog` should be used maybe?

DorraJaouad: let's keep it with `NcEmptyContent`, the layout looks better with the icon :p

src/store/participantsStore.js
danxuliu: Note that `finishedConnecting` does not mean `It was possible to join the call`, but `After joining the call, the list of participants in the call was received`. Therefore this change breaks its behaviour.

After joining a call without HPB _Connecting_ was shown in the call view until the list of participants in the call was received. If there were no other participants in the call _Connecting_ was replaced by _Waiting for others to join the call_, and if there were other participants then a connection was established with them. Now, if there are other participants, _Waiting for others to join the call_ will be briefly shown until the participant list is received, which is not correct.

If there is HPB, however, the _Connecting_ state was wrong, as `usersInRoom` is never triggered in that case (probably `usersChanged` should have been used instead) :shrug: So _Connecting_ was shown for ten seconds when there were no other participants in the call.

DorraJaouad: Fixed, yet one issue is that I couldn't find a signaling event on `base` level for receiving participants list. 
`usersInRoom` works for internal signaling
However, there was no event triggered in external signaling. Am I missing one or I can just create a new one?

danxuliu: You could listen to [`usersChanged`](https://github.com/nextcloud/spreed/blob/157e1751be3e571c961e0e28d1141f1b33473c18/src/utils/signaling.js#L1500). This is the counterpart of `usersInRoom` for the external signaling server, and it is in fact [the one used to check which participants are in the call](https://github.com/nextcloud/spreed/blob/32ba79dec837d3c2f33015ccb2522bc4b4534b46/src/utils/webrtc/webrtc.js#L521) to connect to them (similarly to [what is done with `usersInRoom`](https://github.com/nextcloud/spreed/blob/32ba79dec837d3c2f33015ccb2522bc4b4534b46/src/utils/webrtc/webrtc.js#L563)).

A little detail to keep in mind. While it would be really unusual, it is technically possible that a previous participant update is received after joining the call and before receiving the updated list of participants that includes you. So while in general it would be enough to wait for the first `usersInRoom`/`usersChanged` event received after joining the call the best would be to wait for the first that includes your own session with [an `inCall` flag different than disconnected](https://github.com/nextcloud/spreed/blob/9c30404613fb2694e2a1c60bb2da1f572f332874/src/constants.js#L133).

But keep in mind that when the external signaling server is used there are two different session IDs, the Nextcloud session ID (probably the one that you should check) and the signaling session ID. The `participants->update` message sent by the external signaling server (which becomes the `usersChanged` event) should include both, while the `usersInRoom` message will only have `sessionId` (because in that case there is just one).

DorraJaouad: From testing, I spotted that **updated** `usersChanged` comes before `joinCall` :')

danxuliu: That is [expected](https://github.com/nextcloud/spreed/blob/32ba79dec837d3c2f33015ccb2522bc4b4534b46/src/utils/webrtc/webrtc.js#L575-L581) (unfortunate, but expected :-P ). The join call request sent to the Nextcloud server causes the signaling server to send a signaling message with the updated participant list. However, as the signaling message is sent through the WebSocket in some cases it can be received (which causes [the `usersChanged` event to be emitted](https://github.com/nextcloud/spreed/blob/24d2865eb3fda780e5f36439ad1f2497c42d4f59/src/utils/signaling.js#L1500)) before the HTTP request response for the join call request is received (which causes [the `joinCall` event to be emitted](https://github.com/nextcloud/spreed/blob/24d2865eb3fda780e5f36439ad1f2497c42d4f59/src/utils/signaling.js#L277)).

What you could do is, rather than finish connecting once the updated participant list is received, finish connecting once the updated participant list is received if `joinCall` was already received, and finish connecting once `joinCall` is received if the updated participant list was already received between starting to join the call and actually joining the call.

DorraJaouad: image

Antreesy (commented):
Some thoughts from me

src/store/participantsStore.spec.js
```diff
@@ -572,23 +572,70 @@ describe('participantsStore', () => {
...
+			EventBus.emit('signaling-users-in-room', [participants])
+			expect(store.getters.isInCall(TOKEN)).toBe(true)
+			expect(store.getters.isConnecting(TOKEN)).toBe(false)
+		}
+
+		const emitFailureAndAssert = (reason) => {
+			EventBus.emit('signaling-join-call-failed', [TOKEN, reason])
+			expect(store.getters.isInCall(TOKEN)).toBe(false)
+			expect(store.getters.isConnecting(TOKEN)).toBe(false)
+		}
```
Antreesy, line 602:
```suggestion
			expect(store.getters.connectionFailed(TOKEN)).toBe(reason)
		}
```

src/store/participantsStore.js
```diff
@@ -338,30 +344,51 @@ const mutations = {
...
+		if (!state.connecting[token]) {
+			Vue.set(state.connecting, token, {})
+		}
+		Vue.set(state.connecting[token], sessionId, flags)
+	},
+
 	finishedConnecting(state, { token, sessionId }) {
 		if (state.connecting[token] && state.connecting[token][sessionId]) {
 			Vue.delete(state.connecting[token], sessionId)
+			if (!Object.keys(state.connecting[token]).length) {
```
Antreesy, line 389: Maybe we don't need to store them by token at all, since we can have only one session per tab?
But something for the follow-up, when migrate to pinia

src/store/participantsStore.js
```diff
@@ -821,31 +863,91 @@ const actions = {
 			return
 		}
 
-		commit('setInCall', {
-			token,
-			sessionId: participantIdentifier.sessionId,
-			flags,
-		})
+		let isParticipantsListReceived = null
```
Antreesy, line 866: Consistent boolean
```suggestion
		let isParticipantsListReceived = false
```

src/store/participantsStore.js
```diff
@@ -821,31 +863,91 @@ const actions = {
...
+					isParticipantsListReceived = true
+					commit('connecting', { token, sessionId, flags })
+					return
+				}
+				finishConnecting()
+			}
+		}
+
+		const finishConnecting = () => {
+			isParticipantsListReceived = null
```
Antreesy, line 918: Maybe we don't need to reassign it back to null/false, since it's purpose is completed, and we won't change it later anymore in this function scope
```suggestion
```

src/store/participantsStore.js
```diff
@@ -821,31 +863,91 @@ const actions = {
...
+					isParticipantsListReceived = true
+					commit('connecting', { token, sessionId, flags })
+					return
+				}
+				finishConnecting()
+			}
+		}
+
+		const handleUsersChanged = (payload) => {
+			const participant = payload[0].find(p => p.nextcloudSessionId === sessionId)
```
Antreesy, line 906: do not copy-paste
```suggestion
		const handleUsersChanged = (payload) => {
			// EXTERNAL server event
			handleParticipantsListReceived(payload, 'nextcloudSessionId')
		}
		const handleParticipantsListReceived = (payload, key) => {
			const participant = payload[0].find(p => p[key] === sessionId)
```

Antreesy (commented):
The rest seems fine

src/store/participantsStore.spec.js
```diff
@@ -572,27 +572,109 @@ describe('participantsStore', () => {
...
+				attendeeId: 1,
+				sessionId: 'session-id-1',
+				inCall: actualFlags,
+				participantType: PARTICIPANT.TYPE.USER,
+			}])
+		}
 
-			// Finished connecting to the call
-			EventBus.emit('signaling-users-in-room')
+		const finishConnectingWithUsers = (participants) => {
```
Antreesy, line 593: make the same as `assertParticipantsListReceivedBeforeJoiningCall = (event, payload)` to test also `'joins call with external signaling'`?

src/store/participantsStore.spec.js
```diff
@@ -572,27 +572,109 @@ describe('participantsStore', () => {
...
+			}]])
+			expect(store.getters.isInCall(TOKEN)).toBe(true)
+			expect(store.getters.isConnecting(TOKEN)).toBe(false)
+			expect(store.getters.isJoiningCall(TOKEN)).toBe(false)
+		})
+
+		test('joins call but it fails', async () => {
+			assertInitialCallState()
+			emitFailureAndAssert('error reason')
+		})
```
Antreesy, line 646: Error handling is here, so you don't receive flags
```suggestion
		})
		
		test('trying to join call but it fails', async () => {
			emitFailureAndAssert('error reason')
		})
```

src/store/participantsStore.js
```diff
@@ -809,8 +833,23 @@ const actions = {
...
+		// 2B. Wait for the users list (signaling-users-changed) EXTERNAL server event
+		// In case of failure, we receive a signaling-join-call-failed event
+
+		// Exception 1: We may receive the users list before the signaling-join-call event
+		// In this case, we use the isParticipantsListReceived flag to handle this case
+
+		// Exception 2: We may receive the users list in a second event of signaling-users-changed or signaling-users-in-room
+		// In this case, we always check if the list is the updated one (it has the current participant in the call)
+
+		const { sessionId } = participantIdentifier
```
Antreesy, line 850: Just in case so we don't stuck here
```suggestion
		const { sessionId } = participantIdentifier ?? {}
```

src/store/participantsStore.js
```diff
@@ -821,31 +860,86 @@ const actions = {
...
+			commit('finishedConnecting', { token, sessionId })
+			commit('finishedJoiningCall', { token, sessionId })
+			EventBus.off('signaling-join-call', handleJoinCall)
+			EventBus.off('signaling-join-call-failed', handleJoinCallFailed)
+			EventBus.off('signaling-users-in-room', handleUsersInRoom)
+			EventBus.off('signaling-users-changed', handleUsersChanged)
+		}
+
+		// Fallback in case we never receive the users list after joining the call
 		setTimeout(() => {
```
Antreesy, line 920:
```suggestion
		const connectingTiemout = setTimeout(() => {
```

src/store/participantsStore.js
```diff
@@ -821,31 +860,86 @@ const actions = {
...
+
+		const handleUsersInRoom = (payload) => {
+			handleParticipantsListReceived(payload, 'sessionId')
+		}
+
+		const handleUsersChanged = (payload) => {
+			handleParticipantsListReceived(payload, 'nextcloudSessionId')
+		}
+
+		const finishConnecting = () => {
```
Antreesy, line 910:
```suggestion
		const finishConnecting = () => {
			clearTimeout(connectingTimeout)
```

Antreesy (approved):
Acceptable

PR #13356, opened 2024-09-19
fix: migrate sounds store to Pinia 🍍 

Antreesy:
### ☑️ Resolves

* Fix nothing
* Break nohing 

No visual changes

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


src/stores/sounds.js
```diff
@@ -0,0 +1,94 @@
...
+import { loadState } from '@nextcloud/initial-state'
+import { generateFilePath } from '@nextcloud/router'
+
+import BrowserStorage from '../services/BrowserStorage.js'
+import { setPlaySounds } from '../services/settingsService.js'
+
+/**
+ * Get play sounds option (from server for user or from browser storage for guest)
+ * @param {boolean} value whether sounds should be played
+ */
```
DorraJaouad, line 19:
```suggestion
/**
 * Get play sounds option (from server for user or from browser storage for guest)
 */
```
and did you mean to put it under `hasUserAccount` ?

src/stores/sounds.js
```diff
@@ -0,0 +1,94 @@
...
+			if (!this.audioObjectsCreated) {
+				this.initAudioObjects()
+			}
+			this.audioObjects[key].play()
+		},
+
+		pauseAudio(key) {
+			if (!this.audioObjectsCreated) {
+				this.initAudioObjects()
+			}
```
DorraJaouad, line 61:
```suggestion
			if (!this.audioObjectsCreated) {
				return
			}
```

If it wasn't initiated before, it means it wasn't needed and maybe it's not needed after pausing

src/components/MediaSettings/MediaDevicesSpeakerTest.vue
```diff
@@ -61,17 +69,17 @@ export default {
...
 			this.isPlayingTestSound = true
-			this.$store.dispatch('playWaitAudio').then((response) => {
-				response.addEventListener('ended', () => {
-					this.isPlayingTestSound = false
-				})
-			})
+			this.soundsStore.playAudio('wait')
+			this.soundsStore.audioObjects.wait.addEventListener('ended', () => {
+				this.isPlayingTestSound = false
+			}, { once: true })
```
DorraJaouad, line 82: I wonder if we need error handling here in case playing doesn't work (in that case, we should not set the event listener)

DorraJaouad (commented):
Nice refactoring 🦅

src/stores/sounds.js
Antreesy: I'd keep the first one. We are initiating objects only if CallButton is clicked, so it will act as fallback, if for some reason it wouldn't be registered

DorraJaouad (approved):
Tested ✅

PR #13358, opened 2024-09-20
feat: archive conversations (frontend)

Antreesy:
### ☑️ Resolves

* Fix #…

🏚️ Before | 🏡 After
-- | --

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


src/utils/conversation.js
```diff
@@ -27,6 +27,30 @@ export function hasUnreadMentions(conversation) {
...
+ * - non-archived
+ * - archived, but have an unread message with notification triggered
+ *
+ * @param {object} conversation conversation object
+ * @return {boolean}
+ */
+export function isDefault(conversation) {
+	return !isArchived(conversation)
+		|| (conversation.notificationLevel === PARTICIPANT.NOTIFY.ALWAYS && hasUnreadMessages(conversation))
+		|| (conversation.notificationLevel === PARTICIPANT.NOTIFY.MENTION && hasUnreadMentions(conversation))
```
DorraJaouad, line 51: notification level can be `ALWAYS` or (`DEFAULT` and it is 1-1 conv) for unread messages
can be `ALWAYS` or `DEFAULT` or `MENTION` for unread mentions

src/store/conversationsStore.js
```diff
@@ -904,7 +927,7 @@ const actions = {
 	async setNotificationLevel({ commit }, { token, notificationLevel }) {
 		try {
 			await setNotificationLevel(token, notificationLevel)
-			commit('setNotificationLevel', { token, notificationLevel })
+			commit('setNotificationLevel', { token, notificationLevel: +notificationLevel })
```
DorraJaouad, line 930: Was there an issue you noticed?

src/store/conversationsStore.js
```diff
@@ -525,6 +528,26 @@ const actions = {
...
+		}
+
+		try {
+			if (isArchived) {
+				await unarchiveConversation(token)
+			} else {
+				await archiveConversation(token)
+			}
+
+			const conversation = Object.assign({}, getters.conversations[token], { isArchived: !isArchived })
```
DorraJaouad, line 543:
```suggestion
			const conversation = {
				...getters.conversations[token],
				isArchived: !isArchived
			}
```

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -45,6 +45,15 @@
 							{{ t('spreed', 'Filter unread messages') }}
 						</NcActionButton>
 
+						<NcActionButton close-after-click
```
DorraJaouad, line 48: Capability support check?

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -277,6 +287,25 @@
...
+				</NcButton>
+				<NcButton v-else
+					type="tertiary"
+					wide
+					@click="handleFilter('archived')">
+					<template #icon>
+						<IconArchive :size="20" />
+					</template>
+					{{ t('spreed', 'Archived conversations') }}
+				</NcButton>
```
DorraJaouad, line 307: I am not sure if we need another clear filter button here.
Yet, Archived conversations should be shown ONLY if there are archived ones indeed.

I think you put filter button here so it undoes what archived conversation button do? in that case, we show `return to conversations` or sthg like that (we don't show to user this feature as a filter but as another page/tab in left sidebar ;) ). Also, we won't show it in filters action.

DorraJaouad (commented):
Functionality is good but a few notes about how we will implement it in LS.

src/utils/conversation.js
SystemKeeper: I think we discussed to not make it dependant on the level for now? Direct mention will show it again, regardless of notification setting ?

Antreesy:
Short summary after the live discussion:
- [ ] When archiving, change 'notifications level' -> direct only (if was wider)
- [ ] When archiving, change 'should notify about call' -> false
- [ ] remove an option from filter dropdown (footer button is always present on screen)
- [ ] make 'leave conversation' a dialog with 'archive instead' option
- [ ] capabilities check (when filtering too)
- [ ] conversation settings dialog option (danger zone)

Antreesy:
> I guess we should do this on API level?

Makes sense

src/utils/conversation.js
Antreesy: DEFAULT is backend-only for unassigned yet conversation (by default, it's ALWAYS for 1-1 chats and MENTION for group chats, so we don't see and don't consider it in frontend

See `hasUnreadMessages()` and `hasUnreadMentions()` methods above, they should include these cases correctly

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -277,6 +278,25 @@
...
+				<NcButton v-else-if="supportsArchive"
+					type="tertiary"
+					wide
+					@click="handleFilter('archived')">
+					<template #icon>
+						<IconArchive :size="20" />
+					</template>
+					{{ t('spreed', 'Archived conversations') }}
+				</NcButton>
+
```
DorraJaouad, line 299: I think it should not be seen when you don't have any archived conversations. 
P.S: also check if filter !== 'archived' in render condition so to not remove them when you are already on archived conversations page

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -496,6 +522,8 @@ export default {
 				return t('spreed', 'You have no unread mentions.')
 			case 'unread':
 				return t('spreed', 'You have no unread messages.')
+			case 'archived':
+				return t('spreed', 'You have no archived conversations.')
```
DorraJaouad, line 526: Filter traces are seen in:
* Highlighted filter button
* Clear filter button
image

src/components/LeftSidebar/ConversationsList/Conversation.vue
```diff
@@ -96,12 +107,29 @@
 		</template>
 
 		<!-- confirmation required to delete conversation -->
-		<template v-if="isDialogOpen" #extra>
```
DorraJaouad, line 99: `<!-- confirmation required to delete conversation or leave conversation -->`

src/components/ConversationSettings/DangerZone.vue
```diff
@@ -4,78 +4,109 @@
 -->
 
 <template>
-	<div>
+	<div class="danger-zone">
```
DorraJaouad, line 7: image
Total margin top and bottom should be the opposite around notecard ?

src/components/LeftSidebar/LeftSidebar.vue
Antreesy: Maybe it's better to implement separately from filters then

src/components/ConversationSettings/DangerZone.vue
ShGKme: I'd even put it into a separate section.

Leave/Delete conversation is destructive operations that, as said, cannot be undone. Archive is just a UI filter.

Antreesy: Then I'll move it to Personal section

src/components/ConversationSettings/ConversationSettingsDialog.vue
```diff
@@ -30,6 +30,15 @@
...
+				<NcCheckboxRadioSwitch v-if="supportsArchive"
+					type="switch"
+					:checked="isArchived"
+					@update:checked="toggleArchiveConversation">
+					{{ t('spreed', 'Archive conversation') }}
+				</NcCheckboxRadioSwitch>
+				<p>
+					{{ t('spreed', 'Archived conversations are hidden from the list. They will still be shown, if you have been mentioned directly.') }}
+				</p>
 				<NotificationsSettings v-if="!isGuest" :conversation="conversation" />
```
DorraJaouad, line 42: Well, putting them next to each other highlight that notification settings need to be updated for conversation settings :p cc @nickvergessen 
image

However, do we still need to show the toggle button as disabled as notifications setting is standard for archived conversations?

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -136,6 +136,7 @@
 					<template #icon>
 						<AtIcon v-if="isFiltered === 'mentions'" :size="64" />
 						<MessageBadge v-else-if="isFiltered === 'unread'" :size="64" />
+						<IconArchive v-else-if="showArchived" :size="64" />
```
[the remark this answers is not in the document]
Antreesy: It's shown when you unarchived all of your conversation. So make sense for me to keep it

src/components/ConversationSettings/ConversationSettingsDialog.vue
Antreesy: we can set the appropriate value for notification level, but not lock it completely. In theory, if needed, I can still set `Notify about calls` after archiving, so I'll receive a notification and should also see it in the list

src/components/ConversationSettings/ConversationSettingsDialog.vue
```diff
@@ -30,6 +30,15 @@
 				<p v-if="recordingConsentRequired">
 					{{ t('spreed', 'The consent to be recorded will be required for each participant before joining every call.') }}
 				</p>
+				<NcCheckboxRadioSwitch v-if="supportsArchive"
+					type="switch"
+					:checked="isArchived"
+					@update:checked="toggleArchiveConversation">
+					{{ t('spreed', 'Archive conversation') }}
+				</NcCheckboxRadioSwitch>
+				<p>
```
nickvergessen, line 39: Should by styled like `conversation-permissions-editor__hint` is on the same modal

src/components/ConversationSettings/ConversationSettingsDialog.vue
nickvergessen: > UPD: notifications are disabled by https://github.com/nextcloud/spreed/commit/e5d37ded241c6ebc27abc937bcce1fc1b943f1a0, but we can at least show it in the list, if enabed by user

We can change this. It was just my first thought. But we can change it now that we agreed that the API would toggle the other setting(s)?

src/components/ConversationSettings/DangerZone.vue
```diff
@@ -14,9 +14,26 @@
...
 					{{ t('spreed', 'Leave conversation') }}
 				</NcButton>
+				<NcDialog class="danger-zone__dialog"
+					:open.sync="isLeaveConversationDialogOpen"
+					:name="t('spreed','Leave conversation')"
+					:message="leaveConversationDialogMessage"
+					container=".danger-zone">
+					<template #actions>
+						<NcButton v-if="supportsArchive && !conversation.isArchived" type="secondary" @click="toggleArchiveConversation">
+							{{ t('spreed', 'Archive instead') }}
```
nickvergessen, line 27: Explain what it changes?

src/components/ConversationSettings/ConversationSettingsDialog.vue
Antreesy: Sounds good to us, since endpoint returns updated Room object

src/components/ConversationSettings/DangerZone.vue
Antreesy: Copy-paste from archive toggle should be fine here as well

Antreesy: Nah, too much text. Placing 'Archive' option in Danger zone should be enough to explain for both
image

src/components/LeftSidebar/ConversationsList/Conversation.vue
```diff
@@ -95,13 +106,35 @@
...
+				<template #actions>
+					<NcButton v-if="supportsArchive && !item.isArchived" type="secondary" @click="toggleArchiveConversation">
+						{{ t('spreed', 'Archive instead') }}
+					</NcButton>
+					<NcButton type="tertiary" @click="isLeaveDialogOpen = false">
+						{{ t('spreed', 'No') }}
+					</NcButton>
+					<NcButton type="warning" @click="leaveConversation">
+						{{ t('spreed', 'Yes') }}
+					</NcButton>
```
DorraJaouad, line 129: image

flip "no" and "archive instead" ?

src/components/ConversationSettings/ConversationSettingsDialog.vue
```diff
@@ -84,6 +84,18 @@
...
+					{{ t('spreed', 'Archive conversation') }}
+				</h4>
+				<NcCheckboxRadioSwitch v-if="supportsArchive"
+					type="switch"
+					:checked="isArchived"
+					@update:checked="toggleArchiveConversation">
+					{{ t('spreed', 'Archive conversation') }}
+				</NcCheckboxRadioSwitch>
+				<p class="app-settings-section__hint">
+					{{ t('spreed', 'Archived conversations are hidden from the conversation list by default. They will still be shown, if you have been mentioned directly, when you search for the conversation name or filter your conversation list for archived conversations.') }}
```
DorraJaouad, line 97:
```suggestion
					{{ t('spreed', 'Archived conversations are hidden from the conversation list by default. They will still be shown when you search for the conversation name or open archived conversations list.') }}
```

Because mentioning depends on user notification settings

src/components/LeftSidebar/ConversationsList/Conversation.vue
```diff
@@ -95,13 +106,35 @@
...
+				:name="t('spreed', 'Leave conversation')">
+				<template #default>
+					<p>{{ dialogLeaveMessage }}</p>
+					<p v-if="supportsArchive && !item.isArchived">
+						{{ t('spreed', 'You can archive this conversation instead.') }}
+					</p>
+				</template>
+				<template #actions>
+					<NcButton v-if="supportsArchive && !item.isArchived" type="secondary" @click="toggleArchiveConversation">
+						{{ t('spreed', 'Archive instead') }}
```
DorraJaouad, line 122:
```suggestion
						{{ t('spreed', 'Archive conversation') }}
```

Better and aligning with `DangerZone`

DorraJaouad (approved):
Tested, works good 🦅 , just a few comments on the UI.

src/components/ConversationSettings/ConversationSettingsDialog.vue
Antreesy: Dropping line for now to not create extra work for translators. Will continue in follow-up

PR #13360, opened 2024-09-20
feat(audioPlayer): Auto play voice messages which are grouped together

### LINKED ISSUE #13199: Let audio recordings play after each other
### How to use GitHub

---

## Feature request

**Which Nextcloud Version are you currently using:** (see administration page)
29.0.6 (RC)

No,  there is no problem.

On, for example Whatsapp, when you send multiple voice memo's they keep playing after each other so you don't need to press them all. In nextcloud, you have to keep clicking on play. Maybe it's a nice idea to make them just play after each other automatically.

None.

**Additional context**
None.

Souptik2001:
### ☑️ Resolves

* Fix #13199 

**Note:**  In Whatsapp (whose example is given in the issue), the voice messages auto-play only when they are grouped together. Therefore I tried to replicate that only. If there's a message between two voice messages, it will not be auto-played.
That being said, its very easy to change this behaviour. We just need to remove this `since` check over [here](https://github.com/nextcloud/spreed/pull/13360/files#diff-8a7e5104736c054cb13539ee9b0721cbb5ad4bfa747633f51e9794f08d6a1ce6R94).

Before 

https://github.com/user-attachments/assets/5b899ae6-1e0e-4563-9097-e3282b8a0c28

After

https://github.com/user-attachments/assets/968525ea-9758-4b0f-bb63-a9b447c4557e

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/AudioPlayer.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/AudioPlayer.vue
@@ -4,9 +4,11 @@
 -->
 
 <template>
-	<audio class="audio-player"
+	<audio ref="audioPlayer"
+		class="audio-player"
 		controls
-		:src="fileURL">
+		:src="fileURL"
+		@ended="handleEnded">
 		{{ t('spreed', 'Your browser does not support playing audio files') }}
 	</audio>
 </template>
@@ -15,6 +17,8 @@
 import { t } from '@nextcloud/l10n'
 import { encodePath } from '@nextcloud/paths'
 import { generateRemoteUrl } from '@nextcloud/router'
+
+import { EventBus } from '../../../../../services/EventBus.js'
 
 export default {
 	name: 'AudioPlayer',
@@ -47,6 +51,17 @@ export default {
 			type: String,
 			default: '',
 		},
+		/**
+		 * Message ID.
+		 */
+		messageId: {
+			type: Number,
+			default: 0,
+		},
+		nextMessageId: {
+			type: Number,
+			default: 0,
+		},
 	},
 
 	computed: {
@@ -72,8 +87,37 @@ export default {
 		},
 	},
 
+	mounted() {
+		EventBus.on('audio-player-ended', this.autoPlay)
+	},
+
+	beforeDestroy() {
+		EventBus.off('audio-player-ended', this.autoPlay)
+	},
+
 	methods: {
 		t,
+
+		handleEnded() {
+			if (!this.nextMessageId) {
+				return
+			}
+
+			EventBus.emit('audio-player-ended', this.nextMessageId)
+		},
+
+		/**
+		 * Autoplay the audio message as soon as previous one was played
+		 *
+		 * @param {number} messageId Message ID to play.
+		 */
+		autoPlay(messageId) {
+			if (messageId !== this.messageId) {
+				return
+			}
+
+			this.$refs.audioPlayer?.play()
+		},
 	},
 }
 </script>
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -276,6 +276,8 @@ export default {
 						component: FilePreview,
 						props: {
 							token: this.message.token,
+							messageId: this.message.id,
+							nextMessageId: this.nextMessageId,
 							itemType,
 							referenceId: this.message.referenceId,
 							file: this.message.messageParameters[p],
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/FilePreview.vue
@@ -115,6 +115,16 @@ export default {
 			required: true,
 		},
 
+		messageId: {
+			type: [String, Number],
+			default: 0,
+		},
+
+		nextMessageId: {
+			type: [String, Number],
+			default: 0,
+		},
+
 		/**
 		 * File object
 		 */
@@ -245,6 +255,8 @@ export default {
 					name: this.file.name,
 					path: this.file.path,
 					link: this.file.link,
+					messageId: Number(this.messageId),
+					nextMessageId: Number(this.nextMessageId),
 				}
 			}
 			return {


src/components/MessagesList/MessagesGroup/Message/MessagePart/AudioPlayer.vue
```diff
@@ -72,6 +80,37 @@ export default {
...
+			if (!this.$refs.audioPlayer) {
+				return
+			}
+
+			const voiceMessages = this.$store.getters.getVoiceMessages()
+
+			const previouslyPlayedVoiceMessage = voiceMessages[e.detail.id]
+			const thisVoiceMessage = voiceMessages[previouslyPlayedVoiceMessage.nextId]
+
+			if (previouslyPlayedVoiceMessage.nextId === this.id && thisVoiceMessage && thisVoiceMessage.since === 0) {
```
Souptik2001, line 94: In Whatsapp (whose example is given in the issue), the voice messages auto-play only when they are grouped together. Therefore I tried to replicate that only. If there's a message between two voice messages, it will not be auto-played.

Therefore this `thisVoiceMessage.since === 0` condition is given. If we don't want this behaviour we can remove this and it will auto-lay irrespective of they are grouped or not.

src/components/MessagesList/MessagesGroup/Message/MessagePart/AudioPlayer.vue
```diff
@@ -47,6 +48,13 @@ export default {
 			type: String,
 			default: '',
 		},
+		/**
+		 * File ID.
+		 */
+		id: {
```
Antreesy, line 54:
```suggestion
		messageId: {
```
I'd stay consistent with naming

src/components/MessagesList/MessagesGroup/Message/MessagePart/AudioPlayer.vue
```diff
@@ -72,6 +80,37 @@ export default {
 		},
 	},
 
+	mounted() {
+		document.addEventListener('audioPlayerEnded', (e) => {
```
Antreesy, line 84: We have EventBus used in the code, which I'd used instead of custom events:

- audio component stops playing and emits event with message id
- all other audio components listen to this event.
- one which message is next from stopped one (could be a computed prop based on messagesStore) stars to play
- repeat from the start

src/components/MessagesList/MessagesGroup/Message/MessagePart/AudioPlayer.vue
Antreesy: WhatsApp (and others, I believe) plays not only grouped voice messages, but consequent ones, so from different authors, if they go one by one

src/components/MessagesList/MessagesGroup/Message/MessagePart/AudioPlayer.vue
```diff
@@ -72,6 +80,37 @@ export default {
...
+
+			const previouslyPlayedVoiceMessage = voiceMessages[e.detail.id]
+			const thisVoiceMessage = voiceMessages[previouslyPlayedVoiceMessage.nextId]
+
+			if (previouslyPlayedVoiceMessage.nextId === this.id && thisVoiceMessage && thisVoiceMessage.since === 0) {
+				this.$refs.audioPlayer.play()
+			}
+		})
+
+		this.$refs.audioPlayer.addEventListener('ended', (e) => {
```
Antreesy, line 99: That should work from template:

```vue
<audio @ended="handleEnded" />
```

src/store/messagesStore.js
```diff
@@ -134,6 +134,18 @@ const state = {
 	 * Array of temporary message id to cancel function for the "postNewMessage" action
 	 */
 	cancelPostNewMessage: {},
+	/**
+	 * Number of messages send after last voice message.
```
Antreesy, line 138: With given comments, I don't see a reason of using a store

Antreesy (commented):
Hi, thanks for the contribution!
I see the current implementation a bit overcomplicated, please see comments

src/components/MessagesList/MessagesGroup/Message/MessagePart/AudioPlayer.vue
Souptik2001: @Antreesy Yeah I meant that..

Just to confirm suppose a conversation is like this -

```
- Voice message 1
- Voice message 2
- Some text
- Voice message 3
```

Suppose we play voice message 1.
- Voice message 2 will auto-play.
- Will voice message 3 auto-play or not?

If the answer is yes, then I think we need to use a state to keep the voice messages mapping as I have done (mentioning as you said that we don't need a new state here - https://github.com/nextcloud/spreed/pull/13360#discussion_r1769582994)

```
[
    <id>: {
         id: <id>,
         nextID: <nextID>,
         prevID: <prevID>,
    },
    ....,
    ....,
]
```

And if you say that no, it shouldn't be auto-played as we have a text message in middle, then we don't need this state and we can do some computation on the `messageList` state we have to just check if this audio message is next to the previously played audio message as you said.

Please let me know your opinion!

Antreesy: Did a quick test for two messengers I am using:
- WhatsApp stops at non-voice message, if present;
- Telegram does not stop at all until the end of chat.

I don't know which way would be preferred, we'll discuss it on working day (probably the first one, to not break the flow of discussion, and don't play voice message posted after several days, before you read everything between them). But as I see, both methods don't require a state management:
- a-la WhatsApp - get id of next message in the list, and if voice - play it;
- a-la Telegram - find next voice message in the list and play it.

src/components/MessagesList/MessagesGroup/Message/MessagePart/AudioPlayer.vue
```diff
@@ -72,8 +83,44 @@ export default {
...
+	},
+
+	beforeDestroy() {
+		EventBus.off('audio-player-ended', this.autoPlay)
+	},
+
 	methods: {
 		t,
+
+		handleEnded() {
```
Antreesy, line 97: Would be better to move all the  computation logic here:
- handleEnded will be executed once
- autoPlay - for each component on page

Souptik2001: Yes makes sense! Done!

src/components/MessagesList/MessagesGroup/Message/MessagePart/AudioPlayer.vue
Souptik2001: @Antreesy Got it! I have kept it in Whatsapp's style for now. Once it is decided let me know if it needs to be changed to Telegram's style.

> But as I see, both methods don't require a state management

Ahh I got it, what you are saying.

But I have a question related to optimization here - the one I have written vs the one you are proposing -

The one I have written -

- The voice mapping is pre-computed, in an "object".
- Therefore when we want to know the next voice message of a particular voice message.
- We just get the next voice message in O(1) time-complexity.

The one you are proposing - If we use `messagesList` state and compute it on audio play end -

- The `messagesList` state is an "array" not an object.
- Therefore first we have to find the current ID in that list in O(n) time complexity.
- And after that -
  - If its by Whatsapp's style, then we don't need to iterate any further and just check the next message.
  - But if its Telegram's style, then we have to keep looping "till" we find next voice message.
- Isn't it becoming too expensive (especially, when we have a long conversation list)?

Please let me know your opinion! 👍

DorraJaouad: Hi, we appreciate your contribution :). I'd go with Whatsapp way of playing audio.

It's true that your method provides O(1) when looking for next playable audio, yet the initial computation and the state management of different elements have a few chances of being used (multiple audio messages sent in a row is not a daily case). Thus, it's preferrable to use O(n) in lookup rather than storing extra/duplicated data in store that we probably won't use at least the whole of it.

Souptik2001: Makes sense @DorraJaouad ! Updated the logic to use `messagesList` state and removed the `voiceMessages` related states I introduced. 👍

Antreesy: I'd simplify that further, given that we're checking only the next message after the current one:

```js
	nextVoiceMessageId() {
		const messages = this.$store.state.messagesStore.messages
		const currentIndex = Object.keys(messages).indexOf(this.messageId)
		
		if (currentIndex === -1) {
			return null
		}
		
		const nextMessage = messages[currentIndex + 1]
		if (nextMessage?.messageType === 'voice-message') {
			return nextMessage.id
		} else {
			return null
		}
	}
```

Souptik2001: @Antreesy I forgot to put the comment, but I made this to make it a little flexible, but do you think this is better?

Antreesy: Yes, we can always extend it later, if required. But for now, I'd keep changes as little as possible

P.S. My bad to not remember it earlier, but we already have nextMessageId in Message component, so no need to calculate it again:

https://github.com/nextcloud/spreed/blob/main/src%2Fcomponents%2FMessagesList%2FMessagesGroup%2FMessage%2FMessage.vue#L171-L174

Souptik2001: @Antreesy Okkk! Then its very simple now. Made the changes! 👍

Antreesy:
Hi, just tested the PR. Works nice!

Also looks good code-wise, left a fixup commit to reduce unneccessary changes. Waiting for @DorraJaouad approval, then we can squash and merge it.

DorraJaouad (approved):
Tested ✅ 
Code-wise ✅

PR #13397, opened 2024-09-25
fix(SelectableParticipant): fix keydown.enter event 

Antreesy:
### ☑️ Resolves

* by setting a new value to modelProxy, instead of updating native input
* Fix regression from #12850

No visual changes

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/BreakoutRoomsEditor/SelectableParticipant.vue
+++ b/src/components/BreakoutRoomsEditor/SelectableParticipant.vue
@@ -10,7 +10,7 @@
 			:aria-label="participantAriaLabel"
 			type="checkbox"
 			class="selectable-participant__checkbox"
-			@keydown.enter="handleEnter">
+			@keydown.enter.stop.prevent="handleEnter">
 		<!-- Participant's avatar -->
 		<AvatarWrapper :id="actorId"
 			token="new"
@@ -139,7 +139,11 @@ export default {
 		t,
 
 		handleEnter(event) {
-			event.target.checked = !event.target.checked
+			if (this.isBulkSelection) {
+				event.target.click()
+			} else {
+				this.$emit('click-participant', this.participant)
+			}
 		},
 	}
 }


src/components/BreakoutRoomsEditor/SelectableParticipant.vue
```diff
@@ -139,7 +139,7 @@ export default {
 		t,
 
 		handleEnter(event) {
-			event.target.checked = !event.target.checked
+			this.modelProxy = !event.target.checked
```
ShGKme, line 142: `modelProxy` is a proxy to `checked` which is an array, while `!event.target.checked` is a boolean

ShGKme: `event.target.click()` may emulate native behavior. Just in case I'd prevent the default behavior in this case (but tested that in no browser Enter works as Space)

Antreesy: Didn't check both cases, so came to a wrong assumption 🙈 

Initial goal was to emit `click-participant` event, which:
- doesn't require changing the `input.target.checked`:
- isn't reflected in UI (not a bulk selection => works almost as a button)

DorraJaouad (approved):
LGTM, I couldn't find a better solution

src/components/BreakoutRoomsEditor/SelectableParticipant.vue
```diff
@@ -139,7 +139,13 @@ export default {
 		t,
 
 		handleEnter(event) {
-			event.target.checked = !event.target.checked
+			if (this.isBulkSelection) {
+				event.preventDefault()
+				event.stopPropagation()
```
ShGKme, line 144: To simplify, we can put both to `v-on`. We don't need to prevent `preventDefault` and `stopPropagation` in the second case.

ShGKme (approved):

PR #13401, opened 2024-09-26
fix(hpb): Deprecate multiple HPBs and fake-clustering

### LINKED ISSUE #13011: Deprecate "clustered HPB" mode
### How to use GitHub

---

Once @fancycode is back from vacation, we should align on the way to scale the HPB (is it multiple configured, or one configured and that might be a load balancer, or …) and then remove the "fake clustering" which assigns a single HPB to a conversation

nickvergessen:
### ☑️ Resolves

* Fix #13011

This PR here will be backported.

Main-only follow up will be done to:
- Always use the first HPB
- Remove the conversation_clustering appconfig
- Remove UI option at add additional ones
- Throw on OCC command when trying to add additional ones
- Make the setup check an error

### 🏁 Checklist


danxuliu (commented):
Fine (and happy xD) to deprecated the _conversation_cluster_ mode.

But I am not sure about forcing a single signaling server. If an HPB cluster is used, is it expected that a single signaling server is used in Talk? Or would it be possible to set up several signaling servers, so each participant ends in a random one, and then the signaling servers take care of exchanging the messages between them as needed?

The idea with that would be to avoid having a single point of failure in the signaling server cluster. @fancycode Could you clarify that? Thanks!

nickvergessen:
> If an HPB cluster is used, is it expected that a single signaling server is used in Talk?

Yes, that is what Jojo told me. You would specify a proxy in Talk in that case and that would take care of automatically load balancing

fancycode:
> > If an HPB cluster is used, is it expected that a single signaling server is used in Talk?
> 
> Yes, that is what Jojo told me. You would specify a proxy in Talk in that case and that would take care of automatically load balancing

Right, the load-balancer in front can easily determine which signaling server(s) are available and distribute the requests across them. Having a list in Talk would require to also implement the checks there and handle all sorts of cases if the list of available servers changes (i.e. how to notify clients to now use a different server, etc.).

fancycode:
> The idea with that would be to avoid having a single point of failure in the signaling server cluster. @fancycode Could you clarify that? Thanks!

Of course the load-balancer / proxy in front should be setup so that it's redundant / highly-available. But this is out of scope for Talk or the signaling server. If your setup is so large that you need a clustered signaling server, you should have ways to run a HA frontend proxy / load balancer.

lib/SetupCheck/HighPerformanceBackend.php
```diff
@@ -36,6 +36,21 @@ public function run(): SetupResult {
...
+		if ($this->talkConfig->getSignalingMode() === Config::SIGNALING_CLUSTER_CONVERSATION) {
+			return SetupResult::warning(
+				$this->l->t('Running the High-performance backend "conversation_cluster" mode is deprecated and will no longer be supported in the upcoming version. The High-performance backend supports real clustering nowadays which should be used instead.'),
+				'https://portal.nextcloud.com/article/Partner-Products/Talk-High-Performance-Backend/Nextcloud-Talk-High-Performance-Back-End-Requirements#content-clustering-and-use-of-multiple-hpbs',
+			);
+		}
+
+		if (count($this->talkConfig->getSignalingServers()) > 1) {
+			return SetupResult::warning(
+				$this->l->t('Defining multiple High-performance backends is deprecated and will no longer be supported in the upcoming version. Instead a proxy should be set up and configured in the Talk settings.'),
```
fancycode, line 49: Small wording suggestion as there usually already is a frontend proxy doing the SSL offloading.
```suggestion
				$this->l->t('Defining multiple High-performance backends is deprecated and will no longer be supported in the upcoming version. Instead a load-balancer should be set up together with clustered signaling servers and configured in the Talk settings.'),
```

danxuliu (approved):
@fancycode Thanks for the information!

fancycode (approved):
lgtm with a wording suggestion

PR #13407, opened 2024-09-26
fix(info.xml): add discussion URL

edward-ly:
Update and explicitly specify the discussion URL as the link in the [app store page](https://apps.nextcloud.com/apps/spreed) is outdated anyway.

Related PR: https://github.com/nextcloud/appstore/pull/1496

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -45,6 +45,7 @@
 	<category>workflow</category>
 
 	<website>https://github.com/nextcloud/spreed</website>
+	<discussion>https://help.nextcloud.com/c/support/spreed/52</discussion>
 	<bugs>https://github.com/nextcloud/spreed/issues</bugs>
 	<repository>https://github.com/nextcloud/spreed.git</repository>
 


SystemKeeper (changes requested):
Migration from categories to tags is currently ongoing, let's wait until that is finished to add the URL.

nickvergessen:
https://help.nextcloud.com/tag/spreed-app looks fine, so whatever is done in the appstore, the default URL should work for talk.

SystemKeeper:
Ah got the full picture now 😅 default will now be help.nextcloud.com. So if we want to directly go to the corresponding tag, we need the discussion tag … we can also use https://help.nextcloud.com/tag/talk-app

edward-ly:
Is it possible to have a consistent tag naming scheme so that the default URL can be something like https://help.nextcloud.com/tag/{app_id}? If not, I can just update the URL too.

edward-ly:
> Is it possible to have a consistent tag naming scheme so that the default URL can be something like [https://help.nextcloud.com/tag/{app_id}](https://help.nextcloud.com/tag/%7Bapp_id%7D)? If not, I can just update the URL too.

@SystemKeeper URL is now updated, but I'd like to know if my suggestion above is possible for other apps that don't have a URL specified. If not, that's fine, we can just stick with help.nextcloud.com too.

nickvergessen:
Could we add a redirect on help so the old link forwards to the new one?

edward-ly:
Which is old and which is new?

nickvergessen:
> Which is old and which is new?

Whatever is displayed on https://apps.nextcloud.com/apps/spreed atm and whatever should be displayed there in the future?

edward-ly:
> > Which is old and which is new?
> 
> Whatever is displayed on https://apps.nextcloud.com/apps/spreed atm and whatever should be displayed there in the future?

Whenever this PR is merged and a new release is pushed, the link in https://apps.nextcloud.com/apps/spreed should be automatically updated to use the link specified here. Or are you talking about a feature where https://help.nextcloud.com/c/apps/spreed automatically redirects to https://help.nextcloud.com/tag/talk-app?

nickvergessen:
> Or are you talking about a feature where https://help.nextcloud.com/c/apps/spreed automatically redirects to https://help.nextcloud.com/tag/talk-app?

Not a feature, just a plain redirect on the web server, so that the old link doesn't lead to a wrong/404 place.

SystemKeeper:
All categories that are migrated have a permalink to the new destination.

SystemKeeper (approved):

PR #13425, opened 2024-10-01
fix(MessagesList): update date separator overnight

DorraJaouad:
### ☑️ Resolves

* Fix: when keeping the conversation open overnight, relative dates do not get updated. 

🏚️ Before | 🏡 After
-- | --
messages list showing two "today", one of them is supposed to be "yesterday" | only one correct "today"

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -25,7 +25,7 @@
 			:data-date-timestamp="dateTimestamp"
 			class="scroller__content"
 			:class="{ 'has-sticky': dateTimestamp === stickyDate }">
-			<li class="messages-group__date">
+			<li :key="dateSeparatorLabels[dateTimestamp]" class="messages-group__date">
 				<span class="messages-group__date-text" role="heading" aria-level="3">
 					{{ dateSeparatorLabels[dateTimestamp] }}
 				</span>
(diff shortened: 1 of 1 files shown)


Antreesy (commented):
After private discussion:
- move logic to `onWindowFocus()`
- use `Vue.set()` to ensure update is reactive
- use `:key="dateSeparatorLabels[dateTimestamp]"` in template if won't update without it

Tried that way, requires 1 minute for browser to grab a changed system time, then click on window trigger re-render

src/components/MessagesList/MessagesList.vue
```diff
@@ -20,7 +20,7 @@
 		</TransitionWrapper>
 
 		<ul v-for="(list, dateTimestamp) in messagesGroupedByDateByAuthor"
-			:key="`section_${dateTimestamp}`"
+			:key="`section_${dateSeparatorLabels[dateTimestamp]}`"
```
Antreesy, line 23: set key just on `<li>`

src/components/MessagesList/MessagesList.vue
```diff
@@ -1266,6 +1266,10 @@ export default {
 			// setTimeout is needed here for Safari to correctly remove the unread marker
 			setTimeout(() => {
 				this.refreshReadMarkerPosition()
+				// Regenerate relative date separators
+				Object.keys(this.dateSeparatorLabels).forEach(dateTimestamp => {
+					this.$set(this.dateSeparatorLabels, dateTimestamp, this.generateDateSeparator(dateTimestamp))
```
Antreesy, line 1271: you need to call `this.$set` on every labels change, otherwise it's not reactive from the start, when you create it in `prepareMessagesGroups`

Antreesy (approved):
Working after locking/unlocking the screen, as demanded.
Check linter and merge =)

ShGKme:
> * use `:key="dateSeparatorLabels[dateTimestamp]"` in template if won't update without it

Why?

PR #13456, opened 2024-10-02
feat(calls): Allow to enforce a maximum call length

### LINKED ISSUE #13445: Automatically end calls after defined duration
### How to use GitHub

---

E.g. 12h

nickvergessen:
### ☑️ Resolves

* Fix #13445 

## 🛠️ API Checklist

Tested with:
- Android: :ballot_box_with_check:
- iOS: :ballot_box_with_check:
- Web: :ballot_box_with_check:

### 🏁 Checklist


docs/settings.md
```diff
@@ -82,6 +82,7 @@ Legend:
...
 | `default_attachment_folder`          | string                                                           | `/Talk`    | No   |        | Specify default attachment folder location                                                                                                                                        |
 | `start_calls`                        | int                                                              | `0`        | Yes  | 🖌️    | Who can start a call, see [constants list](constants.md#start-call)                                                                                                               |
+| `max_call_duration`                  | int                                                              | `0`        | No   | ️      | Maximum duration of a call in seconds, 0 for unlimited                                                                                                                            |
```
ShGKme, line 85:
```suggestion
| `max_call_duration`                  | int                                                              | `0`        | No   | ️       | Maximum duration of a call in seconds, 0 for unlimited                                                                                                                            |
```

nickvergessen: the diff is lying due to the emoji taking up 2 spaces in bytes, it's correct like done with other rows

ShGKme: Hmm, indeed, it has 8 characters. But it is definitely different. I'm not comparing with the row with an emoji.

ShGKme: Replacing [`񥀹`](https://unicode-explorer.com/c/65039) with a space

docs/settings.md
```diff
@@ -82,6 +82,7 @@ Legend:
...
 | `default_attachment_folder`          | string                                                           | `/Talk`    | No   |        | Specify default attachment folder location                                                                                                                                        |
 | `start_calls`                        | int                                                              | `0`        | Yes  | 🖌️    | Who can start a call, see [constants list](constants.md#start-call)                                                                                                               |
+| `max_call_duration`                  | int                                                              | `0`        | No   |        | Maximum duration of a call in seconds, 0 for unlimited                                                                                                                            |
```
ShGKme, line 85: Should it be mentioned that federated rooms are not supported?

ShGKme:
This value is not available in capabilities => clients and users don't know about the limit. The call just suddenly stops. 

Shall we add it to capabilities to be able to inform users about it?

lib/BackgroundJob/MaximumCallDuration.php
```diff
@@ -0,0 +1,57 @@
...
+				continue;
+			}
+
+			$result = $this->roomService->resetActiveSinceInDatabaseOnly($room);
+			if (!$result) {
+				// Someone else won the race condition, make sure this user disconnects directly and then return
+				continue;
+			}
+
+			$this->participantService->endCallForEveryone($room, null);
```
ShGKme, line 53: To think about it in the future: this would be nice to be able to set the reason. So users know why the call has been ended.

docs/settings.md
nickvergessen: They are on the host? But yes we can add it.

nickvergessen:
> This value is not available in capabilities => clients and users don't know about the limit. > The call just suddenly stops.
> 
> Shall we add it to capabilities to be able to inform users about it?

Sounds like a good UX follow up. But out of scope for the current requirements :) 

> Tested with 300 seconds. The call was ended after 407 seconds regardless of the second participant joining time. Both 1-1 and group.
>
> Using the default cron.

Well if cron is not running there is little we can do. If you do a mobile app only call nothing is pinging the server, so we have to rely on the server cron. Planned requirement is 12h, so the ~5 min worst case overlay is "accepted risk".

ShGKme (approved):
Works, but some things to concern:
1. Users have no sign that the call time is limited
2. Clients have no possibility to inform users about the limit
3. A call ends suddenly like it was ended by a participant. User "Alice" may be confused why there is "Alice ended the call" when nobody ended the call
4. Duration is in seconds, but with the default cron it might have 300 seconds error. Might be unclear for users

ShGKme:
> Planned requirement is 12h

It's a bit confusing:
- A user can set it with accuracy of 1 sec
- The default error is 5 min (300 sec)
- The expected typical value is around 12h = 43_200 sec

Do we need 1 second precision with ~300 sec error and thousands of seconds values? Maybe then make it in minutes and if we add a UI make it 5 min step? 

Or at least note it in the docs.

nickvergessen:
> Do we need 1 second precision with ~300 sec error and thousands of seconds values?

Makes CI testing easier if we add it.

> … and if we add a UI …

We will not :)

PR #13465, opened 2024-10-04
fix: migrate pollService to TS

Antreesy:
### ☑️ Resolves

* Prerequisite to further polls PRs

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop
- [x] ⛑️ Tests are included or not possibleSigned-off-by: Maksim Sukharev <antreesy.web@gmail.com>


src/stores/polls.ts
```diff
@@ -117,8 +135,8 @@ export const usePollsStore = defineStore('polls', {
 			}
 		},
 
-		addPollToast({ token, message }) {
-			const pollId = message.messageParameters.object.id
+		addPollToast({ token, message }: { token: string, message: ChatMessage }) {
+			const pollId = +message.messageParameters.object.id
```
DorraJaouad, line 139: We are not using any numerical operations on `pollId`, so it's better to always keep as an ID string and not treat as a number.

DorraJaouad (approved):

src/stores/polls.ts
Antreesy: So far not a problem, as long as we're using `poll.id` as a number (type from OpenApi) and `pollId` as a string key to look for. we'll notice it when it comes to messages store migration =)

PR #13484, opened 2024-10-07
feat(devices): add an option to always disable streams by default befo…

### LINKED ISSUE #13446: Add option for camera and mic to be disabled by default
### How to use GitHub

---

- [ ] API: https://github.com/nextcloud/spreed/pull/13488
- [ ] Frontend: https://github.com/nextcloud/spreed/pull/13484

DorraJaouad:
…re joining the call

### ☑️ Resolves

* Fix #13446

🏚️ Before | 🏡 After
-- | --
. | image

https://github.com/user-attachments/assets/ddd1e172-67d4-4bed-a249-8d4be879c710

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/SettingsDialog/SettingsDialog.vue
+++ b/src/components/SettingsDialog/SettingsDialog.vue
@@ -204,6 +204,7 @@ import { PRIVACY } from '../../constants.js'
 import BrowserStorage from '../../services/BrowserStorage.js'
 import { getTalkConfig } from '../../services/CapabilitiesManager.ts'
 import { useCustomSettings } from '../../services/SettingsAPI.ts'
+import { setUserConfig } from '../../services/settingsService.js'
 import { useSettingsStore } from '../../stores/settings.js'
 import { useSoundsStore } from '../../stores/sounds.js'
 
@@ -297,12 +298,11 @@ export default {
 		},
 	},
 
-	created() {
+	async created() {
 		const blurred = BrowserStorage.getItem('background-blurred')
 		if (blurred === 'false' && isBackgroundBlurred === '') {
 			console.debug('Blur was disabled intentionally, propagating last choice to server')
-			axios.post(generateOcsUrl('apps/provisioning_api/api/v1/config/users/theming/force_enable_blur_filter'),
-				{ configValue: 'no' })
+			await setUserConfig('theming', 'force_enable_blur_filter', 'no')
 		}
 		BrowserStorage.removeItem('background-blurred')
 	},
--- a/src/components/SettingsDialog/SettingsDialog.vue
+++ b/src/components/SettingsDialog/SettingsDialog.vue
@@ -20,6 +20,14 @@
 			:name="t('spreed', 'Choose devices')"
 			class="app-settings-section">
 			<MediaDevicesPreview />
+			<NcCheckboxRadioSwitch id="call-media"
+				:checked="startWithoutMediaEnabled"
+				:disabled="mediaLoading"
+				type="switch"
+				class="checkbox call-media"
+				@update:checked="toggleStartWithoutMedia">
+				{{ t('spreed', 'Turn off camera and microphone by default when joining a call') }}
+			</NcCheckboxRadioSwitch>
 		</NcAppSettingsSection>
 		<NcAppSettingsSection v-if="!isGuest"
 			id="attachments"
@@ -235,6 +243,7 @@ export default {
 			attachmentFolderLoading: true,
 			privacyLoading: false,
 			playSoundsLoading: false,
+			mediaLoading: false,
 		}
 	},
 
@@ -261,6 +270,10 @@ export default {
 
 		typingStatusPrivacyIsPublic() {
 			return this.settingsStore.typingStatusPrivacy === PRIVACY.PUBLIC
+		},
+
+		startWithoutMediaEnabled() {
+			return this.settingsStore.startWithoutMedia
 		},
 
 		settingsUrl() {
@@ -362,6 +375,18 @@ export default {
 			this.playSoundsLoading = false
 		},
 
+		async toggleStartWithoutMedia(value) {
+			this.mediaLoading = true
+			try {
+				await this.settingsStore.setStartWithoutMedia(value)
+				showSuccess(t('spreed', 'Your default media state has been saved'))
+			} catch (exception) {
+				showError(t('spreed', 'Error while setting default media state'))
+			} finally {
+				this.mediaLoading = false
+			}
+		},
+
 		handleShowSettings() {
 			this.showSettings = true
 		},
@@ -418,6 +443,10 @@ export default {
 	.shortcut-description {
 		width: calc(100% - 160px);
 	}
+}
+
+.call-media {
+	margin: calc(3 * var(--default-grid-baseline)) 0;
 }
 
 </style>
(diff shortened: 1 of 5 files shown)


nickvergessen:
I think this should be on a higher level.
Either it needs to be a user setting globally for all their rooms with an admin configurable default, or even a system wide setting.

src/components/CallView/shared/LocalAudioControlButton.vue
```diff
@@ -118,6 +129,9 @@ export default {
 
 	mounted() {
 		subscribe('local-audio-control-button:toggle-audio', this.updateDeviceState)
+		if (this.mediaDefaultStateEnabled) {
+			this.model.disableAudio()
```
Antreesy, line 133: When opening MediaSettings after the first call, audio/video would be toggled off. ot sure yet, how we should handle that

Maybe force enable for device preview before call and show a hint 'Your camera and mic will be switched off when you join a call' depending on mediaDefaultStateEnabled

src/components/SettingsDialog/SettingsDialog.vue
```diff
@@ -89,6 +89,18 @@
 				{{ t('spreed', 'Sounds for chat and call notifications can be adjusted in the personal settings.') }} ↗
 			</a>
 		</NcAppSettingsSection>
+		<NcAppSettingsSection id="calls"
+			:name="t('spreed', 'Calls')"
+			class="app-settings-section">
+			<NcCheckboxRadioSwitch id="call-media"
```
Antreesy, line 95: Move it up, close to devices, or join into a single section?

src/stores/settings.js
```diff
@@ -33,7 +38,8 @@ export const useSettingsStore = defineStore('settings', {
 	state: () => ({
 		readStatusPrivacy: loadState('spreed', 'read_status_privacy', PRIVACY.PRIVATE),
 		typingStatusPrivacy: loadState('spreed', 'typing_privacy', PRIVACY.PRIVATE),
-		showMediaSettings: {}
+		showMediaSettings: {},
+		mediaDefaultState: getTalkConfig('local', 'call', 'start-without-media'),
```
Antreesy, line 42:
```suggestion
		startWithoutMedia: getTalkConfig('local', 'call', 'start-without-media'),
```

I was confused with the name `mediaDefaultStateEnabled`, when we disable everything on true.
Maybe change it to something close to API given?

Antreesy (commented):
Tested, works correctly. Left some comments to discuss:

src/components/SettingsDialog/SettingsDialog.vue
nickvergessen: > close to devices, or join into a single section?

I'd say so, that's where I'd expect to find it

src/services/settingsService.js
```diff
@@ -74,10 +74,16 @@ const setPlaySounds = async function(hasUserAccount, value) {
 	}
 }
 
+const setMediaDefaultState = async function(value) {
+	await axios.post(generateOcsUrl('apps/provisioning_api/api/v1/config/users/spreed/calls_start_without_media'),
```
nickvergessen, line 78: we are going to reuse this heavily going forward. I would recommend to make a function that receives the configKey and configValue as arguments :)

src/components/SettingsDialog/SettingsDialog.vue
DorraJaouad: Put it right under video preview in the same section or keep it separate?

src/components/CallView/shared/LocalAudioControlButton.vue
DorraJaouad: I am not sure if it's privacy-okay to always force enable it? If user turned on their cam in the meeting, it will be on (following their pref). If they didn't, it will stay off.

src/stores/settings.js
DorraJaouad: I am always open to any renaming :p

src/components/SettingsDialog/SettingsDialog.vue
nickvergessen: same section sounds good enough

src/components/CallView/shared/LocalAudioControlButton.vue
nickvergessen: Just to make sure:
1. I go to Talk settings and turn of my media by default
2. Any call I start shall be "off" with and without the device picker
3. The only place that changes the user preference is the "Talk settings", not any action during a call

DorraJaouad: Yes, it's like that. Maksim is mentioning the state in **media settings** not in call.

Antreesy: yep, I meant 'Media settings only';
So users can preview themselves right before the call, but knowing, that camera will be turned off, as soon as they joined

src/services/CapabilitiesManager.ts
```diff
@@ -169,3 +172,15 @@ function restoreRemoteCapabilities(): RemoteCapabilities {
...
 	return remoteCapabilities
 }
+
+/**
+ * Set user config using provisioning API
+ *
+ * @param configKey - key of the config to set
+ * @param configValue - value to set
+ */
+export async function setUserConfig(configKey: keyof Config[keyof Config], configValue: ConfigValueType): Promise<void> {
```
Antreesy, line 182: Config has only `call.start-without-media`, whereas configKey is `calls_start_without_media`, so you wouldn't get a match here
Both ConfigValueType and `keyof Config[keyof Config]` (in my IDE) already evaluates to never and don't make much sense

I'd suggest to not overcomplicate it here inside of Manager, but keep both functions in settingsService:
- generic setUserConfig() - I'd also pass 'spreed' separately, in case we'll want to change something in another app in future from Talk / Talk Desktop
- more specific setStartWithoutMedia() - here you know that value is boolean, and the rest is already fixed

src/components/CallView/shared/LocalVideoControlButton.vue
```diff
@@ -124,6 +135,9 @@ export default {
 
 	mounted() {
 		subscribe('local-video-control-button:toggle-video', this.updateDeviceState)
+		if (this.startWithoutMediaEnabled) {
+			this.model.disableVideo()
```
Antreesy, line 139: Let's do it in CallView `setup`?
We have localMediaModel there, and it always mounts when joining a call. Buttons can be reused elsewhere, logic doesn't really belong to them

src/components/CallView/shared/LocalAudioControlButton.vue
Antreesy: Let's leave until we see in real cases. Not a big deal to focus on now

src/stores/settings.js
```diff
@@ -96,5 +102,11 @@ export const useSettingsStore = defineStore('settings', {
 			}
 			Vue.set(this.showMediaSettings, token, value)
 		},
+
+		async setStartWithoutMedia(value) {
+			await setStartWithoutMedia(value)
+			this.startWithoutMedia = value
+
```
Antreesy, line 109:
```suggestion
```

Antreesy (approved):

src/components/CallView/shared/LocalVideoControlButton.vue
DorraJaouad: Sounds like a good approach

Antreesy (approved):

PR #13486, opened 2024-10-07
callview: Show internal clients with available video.

fancycode:
### ☑️ Resolves

* Shows video component for internal (SIP-) clients that publish a video stream.

🏚️ Before | 🏡 After
-- | --

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/CallView/CallView.vue
+++ b/src/components/CallView/CallView.vue
@@ -227,7 +227,7 @@ export default {
 		},
 
 		callParticipantModels() {
-			return callParticipantCollection.callParticipantModels.value.filter(callParticipantModel => !callParticipantModel.attributes.internal)
+			return callParticipantCollection.callParticipantModels.value.filter(callParticipantModel => !callParticipantModel.attributes.internal || callParticipantModel.attributes.videoAvailable)
 		},
 
 		callParticipantModelsWithScreen() {


fancycode:
This is ready for review.

DorraJaouad (approved):
LGTM

danxuliu (approved):
I have not tested it, but makes sense to show an internal client if it is publishing video :+1:

@fancycode One question, though. If I am not mistaken this would show the SIP bridge as a participant and the video will be shown for the SIP bridge itself, rather than for the user publishing video through the SIP bridge. Is that correct?

Is there any way to associate the current video stream sent by the SIP bridge with the specific participant sending it, so (in the future, for now the current approach is fine :-) ) it can be shown on that participant rather than on the SIP bridge participant?

fancycode:
> @fancycode One question, though. If I am not mistaken this would show the SIP bridge as a participant and the video will be shown for the SIP bridge itself, rather than for the user publishing video through the SIP bridge. Is that correct?

Right, this will show the video of the currently speaking phone user as the SIP bridge participant (which will be visible with this change if video is available).

> Is there any way to associate the current video stream sent by the SIP bridge with the specific participant sending it, so (in the future, for now the current approach is fine :-) ) it can be shown on that participant rather than on the SIP bridge participant?

Not right now, but I could for example send the session id of the published user through a datachannel in this PeerConnection so clients can show the video on the correct participant. Not sure how much client-side changes would be required for this though.

We might even support showing videos of multiple SIP participants in the future, so this would need to be considered also as all these would be coming through the same PeerConnection.

nickvergessen:
So should we still merge this in the meantime?

Issue #13439, opened 2024-10-02
Allow to prepare and import/export polls


nickvergessen:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

- [ ] Also polls need to be configurable during room creation (can redirect to conversation settings after creation)
  - [ ] 'Browse poll drafts' in conversation settings => Meeting
- [x] When importing adjusting of wording should be possible (so basically just prefill the form)

Antreesy:
Saving and retreiving form data from/to JSON should be possible.
Question is, locally or to NC instance?

nickvergessen:
import/export is from user device

Antreesy:
> Also polls need to be configurable during room creation (can redirect to conversation settings after creation)

Still needed, or navigating to drafts from shared items and poll creation is enough?

nickvergessen:
- [ ] Add option to edit a draft or delete a previous draft when editing (on API level)

nickvergessen:
@Antreesy we are missing the "drafts overview", right?

PR #13487, opened 2024-10-07
feat(polls): Import/Export pre-filled poll form to JSON 📊

Antreesy:
### ☑️ Resolves

* Ref #13439
  * [x] Handles import/export from/to JSON file (native browser functionality)
  * [x] Handles only fields defined in client code
  * [x] Skips missing and extraneous fields
  * [x] Type/value validation? (if someone would manually create/modify JSON files for import)

image
image

### 🏁 Checklist

  - [x] Firefox
  - [ ] Safari
  - [x] Talk Desktop

### DIFF
--- a/src/components/NewMessage/NewMessagePollEditor.vue
+++ b/src/components/NewMessage/NewMessagePollEditor.vue
@@ -24,12 +24,24 @@
 		</p>
 		<div class="poll-editor__wrapper">
 			<NcTextField :value.sync="pollForm.question" :label="t('spreed', 'Ask a question')" v-on="$listeners" />
+			<!--native file picker, hidden -->
+			<input id="poll-upload"
+				ref="pollImport"
+				type="file"
+				class="hidden-visually"
+				@change="importPoll">
 			<NcActions v-if="supportPollDrafts" force-menu>
 				<NcActionButton v-if="isModerator" close-after-click @click="openPollDraftHandler">
 					<template #icon>
 						<IconFileEdit :size="20" />
 					</template>
 					{{ t('spreed', 'Browse poll drafts') }}
+				</NcActionButton>
+				<NcActionButton close-after-click @click="triggerImport">
+					<template #icon>
+						<IconFileUpload :size="20" />
+					</template>
+					{{ t('spreed', 'Import draft from file') }}
 				</NcActionButton>
 			</NcActions>
 		</div>
@@ -96,8 +108,10 @@ import { computed, nextTick, reactive, ref } from 'vue'
 import IconArrowLeft from 'vue-material-design-icons/ArrowLeft.vue'
 import Close from 'vue-material-design-icons/Close.vue'
 import IconFileEdit from 'vue-material-design-icons/FileEdit.vue'
+import IconFileUpload from 'vue-material-design-icons/FileUpload.vue'
 import Plus from 'vue-material-design-icons/Plus.vue'
 
+import { showError } from '@nextcloud/dialogs'
 import { t } from '@nextcloud/l10n'
 
 import NcActionButton from '@nextcloud/vue/dist/Components/NcActionButton.js'
@@ -113,6 +127,7 @@ import { hasTalkFeature } from '../../services/CapabilitiesManager.ts'
 import { EventBus } from '../../services/EventBus.js'
 import { usePollsStore } from '../../stores/polls.ts'
 import type { createPollParams } from '../../types/index.ts'
+import { validatePollForm } from '../../utils/validatePollForm.ts'
 
 const props = defineProps<{
 	token: string,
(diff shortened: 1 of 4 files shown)


src/components/NewMessage/NewMessagePollEditor.vue
```diff
@@ -143,16 +170,41 @@ export default {
...
+			}
+
+			const reader = new FileReader()
+			reader.onload = (e) => {
+				try {
+					const jsonObject = JSON.parse(e.target.result)
+					for (const key of Object.keys(this.pollForm)) {
+						if (jsonObject[key] !== undefined) {
+							this.pollForm[key] = jsonObject[key]
+						}
```
DorraJaouad, line 199: I'd say if it is corrupted in any of its keys, we don't render it. We also need to show a hint indicating that the file is not Talk poll json file or has been modified.

src/components/NewMessage/NewMessagePollEditor.vue
```diff
@@ -12,19 +12,31 @@
...
+			ref="pollImport"
+			type="file"
+			class="hidden-visually"
+			@change="importPoll">
+		<NcButton class="poll-editor__button"
+			type="secondary"
+			wide
+			@click="triggerImport">
+			{{ t('spreed', 'Import poll from file') }}
+		</NcButton>
```
DorraJaouad, line 27: Buttons locations are a bit blatant. What if we put them both in in a 3-dot menu on top? 
Also, exporting the poll should always be available (being created, active, closed)

src/components/NewMessage/NewMessagePollEditor.vue
```diff
@@ -102,39 +121,47 @@ export default {
...
+		const isFilled = computed(() => !!pollForm.question || pollForm.options.some(option => option))
 
-	computed: {
-		isFilled() {
-			return !!this.pollQuestion || this.pollOptions.some(option => option)
-		},
+		const exportPollBlob = computed(() => {
+			if (!isFilled.value) {
+				return null
+			}
```
DorraJaouad, line 135:
```suggestion
```

Being checked in DOM

src/components/NewMessage/NewMessagePollEditor.vue
Antreesy: So we need some sort of a validation beforehand

src/components/NewMessage/NewMessagePollEditor.vue
Antreesy: Next to the question looks also good enough?
image

src/components/NewMessage/NewMessagePollEditor.vue
Antreesy: and should be lazy computed, until isFilled is true. Removing

src/components/NewMessage/NewMessagePollEditor.vue
Antreesy: Export on other pages:
voting:
image
results:
image

src/components/NewMessage/NewMessagePollEditor.vue
DorraJaouad: Yes, that are ONLY the keys we have from Poll are there (4)

src/components/NewMessage/NewMessagePollEditor.vue
nickvergessen: > Next to the question looks also good enough?

How about a "Import" Next to the "Create" button
And only have the "Export" on already Created-polls

Antreesy: I wouldn't expect 'Import' button to be close to the finish.
And personally, I can imagine cases where having 'Export' option before creation is useful:
- e.g. creating a poll beforehand in the same conversation (without posting) and save it locally, then restore from file during the call and post to people
- with same intention, but instead of locally saving keep it it Per-Conversation poll drafts (to post from there later)

Though these drafts would require some API work, especially if we want to prepare them during room creation, for example

src/components/PollViewer/PollViewer.vue
```diff
@@ -159,6 +176,16 @@ export default {
...
 
+		const exportPollBlob = computed(() => {
+			return 'data:application/json;charset=utf-8,' + encodeURIComponent(JSON.stringify({
+				question: poll.value.question,
+				options: poll.value.options,
+				resultMode: poll.value.resultMode,
+				maxVotes: poll.value.maxVotes,
+			}, null, 2))
+		})
+		const exportPollFileName = `Talk Poll ${new Date().toISOString().slice(0, 10)}`
```
DorraJaouad, line 187: Doesn't it make sense to put export logic in util?

DorraJaouad (approved):
It works as expected 👍🏻

DorraJaouad (approved):
Code 👍🏻

PR #13490, opened 2024-10-07
feat(call): add option to enable blur background by default for all c…

DorraJaouad:
…onversations

### ☑️ Resolves

It's default settings, meaning that if this config is on, the blur effect will apply automatically and it's up to user to change it before joining the call. 

Talk Settings | Media Settings
-- | --
blur-settings | blur-media-settings

### 🚧 Tasks

- [x] Clean previous browserStorage usage

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/docs/capabilities.md
+++ b/docs/capabilities.md
@@ -162,4 +162,4 @@
 * `download-call-participants` - Whether the endpoints for moderators to download the call participants is available
 * `config => call => start-without-media` (local) - Boolean, whether media should be disabled when starting or joining a conversation
 * `config => call => max-duration` - Integer, maximum call duration in seconds. Please note that this should only be used with system cron and with a reasonable high value, due to the expended duration until the background job ran.
-* `config => call => blur-background` (local) - Boolean, whether blur background is set by default when joining a conversation
+* `config => call => blur-virtual-background` (local) - Boolean, whether blur background is set by default when joining a conversation
--- a/docs/capabilities.md
+++ b/docs/capabilities.md
@@ -162,3 +162,4 @@
 * `download-call-participants` - Whether the endpoints for moderators to download the call participants is available
 * `config => call => start-without-media` (local) - Boolean, whether media should be disabled when starting or joining a conversation
 * `config => call => max-duration` - Integer, maximum call duration in seconds. Please note that this should only be used with system cron and with a reasonable high value, due to the expended duration until the background job ran.
+* `config => call => blur-background` (local) - Boolean, whether blur background is set by default when joining a conversation
--- a/lib/Controller/RoomController.php
+++ b/lib/Controller/RoomController.php
@@ -2458,8 +2458,8 @@ class RoomController extends AEnvironmentAwareController {
 			if (isset($data['config']['call']['start-without-media'])) {
 				$data['config']['call']['start-without-media'] = $this->talkConfig->getCallsStartWithoutMedia($this->userId);
 			}
-			if (isset($data['config']['call']['blur-background'])) {
-				$data['config']['call']['blur-background'] = $this->talkConfig->getBlurBackground($this->userId);
+			if (isset($data['config']['call']['blur-virtual-background'])) {
+				$data['config']['call']['blur-virtual-background'] = $this->talkConfig->getBlurBackground($this->userId);
 			}
 
 			if ($response->getHeaders()['X-Nextcloud-Talk-Hash']) {
--- a/lib/Controller/RoomController.php
+++ b/lib/Controller/RoomController.php
@@ -2458,6 +2458,9 @@ class RoomController extends AEnvironmentAwareController {
 			if (isset($data['config']['call']['start-without-media'])) {
 				$data['config']['call']['start-without-media'] = $this->talkConfig->getCallsStartWithoutMedia($this->userId);
 			}
+			if (isset($data['config']['call']['blur-background'])) {
+				$data['config']['call']['blur-background'] = $this->talkConfig->getBlurBackground($this->userId);
+			}
 
 			if ($response->getHeaders()['X-Nextcloud-Talk-Hash']) {
 				$headers['X-Nextcloud-Talk-Proxy-Hash'] = $response->getHeaders()['X-Nextcloud-Talk-Hash'];
--- a/lib/Controller/RoomController.php
+++ b/lib/Controller/RoomController.php
@@ -2459,7 +2459,7 @@ class RoomController extends AEnvironmentAwareController {
 				$data['config']['call']['start-without-media'] = $this->talkConfig->getCallsStartWithoutMedia($this->userId);
 			}
 			if (isset($data['config']['call']['blur-virtual-background'])) {
-				$data['config']['call']['blur-virtual-background'] = $this->talkConfig->getBlurBackground($this->userId);
+				$data['config']['call']['blur-virtual-background'] = $this->talkConfig->getBlurVirtualBackground($this->userId);
 			}
 
 			if ($response->getHeaders()['X-Nextcloud-Talk-Hash']) {
--- a/src/components/MediaSettings/MediaSettings.vue
+++ b/src/components/MediaSettings/MediaSettings.vue
@@ -352,6 +352,10 @@ export default {
 			return this.settingsStore.getShowMediaSettings(this.token)
 		},
 
+		blurBackgroundEnabled() {
+			return this.settingsStore.getBlurBackgroundEnabled
+		},
+
 		showVideo() {
 			return this.videoPreviewAvailable && this.videoOn
 		},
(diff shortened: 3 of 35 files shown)


nickvergessen:
I don't recall what we store in the local storage so far. its a "type" (string?) and in case type is image we also store the "name of the image" (string?) or what do we store?

DorraJaouad:
We store 3 info: enabled, type, and (url) if it is an image
image

nickvergessen:
We store all 3 in the backend in the future?

DorraJaouad:
> We store all 3 in the backend in the future?

That's a better behavior IMO (not reliant on browser). Will it stay per conversation level or we change to one setting for all conversations?

nickvergessen:
The API stored config has to be 1 for all, otherwise we save too many preferences?

lib/Controller/SettingsController.php
```diff
@@ -39,7 +39,7 @@ public function __construct(
 	/**
 	 * Update user setting
 	 *
-	 * @param 'attachment_folder'|'read_status_privacy'|'typing_privacy'|'play_sounds' $key Key to update
+	 * @param 'attachment_folder'|'read_status_privacy'|'typing_privacy'|'play_sounds'| 'blur_background' $key Key to update
```
nickvergessen, line 42: Don't add new things to old APIs :-X
```suggestion
	 * @param 'attachment_folder'|'read_status_privacy'|'typing_privacy'|'play_sounds' $key Key to update
```

lib/TInitialState.php
```diff
@@ -171,6 +171,11 @@ protected function publishInitialStateForUser(IUser $user, IRootFolder $rootFold
 			'enable_matterbridge',
 			$this->serverConfig->getAppValue('spreed', 'enable_matterbridge', '0') === '1'
 		);
+
+		$this->initialState->provideInitialState(
```
nickvergessen, line 175: should not continue this and instead use capabilities for desktop client compatibility

lib/Controller/RoomController.php
```diff
@@ -2458,6 +2458,9 @@ public function getCapabilities(): DataResponse {
 			if (isset($data['config']['call']['start-without-media'])) {
 				$data['config']['call']['start-without-media'] = $this->talkConfig->getCallsStartWithoutMedia($this->userId);
 			}
+			if (isset($data['config']['call']['blur-background'])) {
```
nickvergessen, line 2461: See note above
```
			 * IMPORTANT:
			 * When adding, changing or removing anything here, update
			 * @see ProxyRequest::overwrittenRemoteTalkHash()
			 * so clients correctly refresh their capabilities.
```

tests/php/CapabilitiesTest.php
```diff
@@ -111,6 +111,7 @@ public function testGetCapabilitiesGuest(): void {
 					],
 					'call' => [
 						'enabled' => true,
+						'blur-background' => false,
```
nickvergessen, line 114: Move down (to the same order as in the Capabilities.php)

src/stores/settings.js
```diff
@@ -71,6 +73,10 @@ export const useSettingsStore = defineStore('settings', {
 			}
 			}
 		},
+
+		getBlurBackgroundEnabled: (state) => {
```
Antreesy, line 77: getter makes no sense, If you don't use any logic here (like choose to support either via talk-config or via BrowserStorage here)

src/composables/useDevices.js
```diff
@@ -106,6 +108,15 @@ export function useDevices(video, initializeOnMounted) {
 
 		virtualBackground.value.connectTrackSink('default', videoTrackToStream.value, 'video')
 
+		const blurBackgroundEnabled = getTalkConfig('local', 'call', 'blur-background')
```
Antreesy, line 111: should check store value here

src/components/SettingsDialog/MediaDevicesPreview.vue
```diff
@@ -63,6 +63,11 @@
 				disablePictureInPicture="true"
 				tabindex="-1" />
 		</div>
+		<NcCheckboxRadioSwitch type="switch"
```
Antreesy, line 66: and should check talkConfig here in `v-if`

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -431,16 +435,21 @@ export default {
...
-				if (BrowserStorage.getItem('virtualBackgroundEnabled_' + this.token) === 'true') {
-					if (BrowserStorage.getItem('virtualBackgroundType_' + this.token) === VIRTUAL_BACKGROUND.BACKGROUND_TYPE.BLUR) {
-						this.blurVirtualBackground()
-					} else if (BrowserStorage.getItem('virtualBackgroundType_' + this.token) === VIRTUAL_BACKGROUND.BACKGROUND_TYPE.IMAGE) {
-						this.setVirtualBackgroundImage(BrowserStorage.getItem('virtualBackgroundUrl_' + this.token))
-					}
+				// Check main blur background setting
+				if (this.blurBackgroundEnabled) {
+					this.blurVirtualBackground()
+					this.blurBackground()
```
Antreesy, line 441: This uses BrowserStorage, which we don't want anymore

docs/capabilities.md
```diff
@@ -162,3 +162,4 @@
 * `download-call-participants` - Whether the endpoints for moderators to download the call participants is available
 * `config => call => start-without-media` (local) - Boolean, whether media should be disabled when starting or joining a conversation
 * `config => call => max-duration` - Integer, maximum call duration in seconds. Please note that this should only be used with system cron and with a reasonable high value, due to the expended duration until the background job ran.
+* `config => call => blur-background` (local) - Boolean, whether blur background is set by default when joining a conversation
```
Antreesy, line 165: I'm afraid we could misinterpet at some point:
- blur-background (from Talk) for video input
- background-blur (from server) for general CSS blur filter

```suggestion
* `config => call => blur-virtual-background` (local) - Boolean, whether blur background is set by default when joining a conversation
```

src/stores/settings.js
DorraJaouad: true true

src/composables/useDevices.js
DorraJaouad: I will only use talk-config without any stored value

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -431,16 +436,20 @@ export default {
...
-
-				// Set virtual background depending on BrowserStorage's settings
-				if (BrowserStorage.getItem('virtualBackgroundEnabled_' + this.token) === 'true') {
-					if (BrowserStorage.getItem('virtualBackgroundType_' + this.token) === VIRTUAL_BACKGROUND.BACKGROUND_TYPE.BLUR) {
-						this.blurVirtualBackground()
-					} else if (BrowserStorage.getItem('virtualBackgroundType_' + this.token) === VIRTUAL_BACKGROUND.BACKGROUND_TYPE.IMAGE) {
-						this.setVirtualBackgroundImage(BrowserStorage.getItem('virtualBackgroundUrl_' + this.token))
-					}
+				// Check main blur background setting
+				if (this.blurBackgroundEnabled) {
```
Antreesy, line 440: That should not override value from BrowserStorage (e.g. if I set an Image for this conversation)

src/components/MediaSettings/VideoBackgroundEditor.vue
```diff
@@ -254,6 +256,11 @@ export default {
 		},
 
 		loadBackground() {
+			// Set virtual background depending on main blur setting (in Talk settings)
+			if (this.settingsStore.blurBackgroundEnabled) {
```
Antreesy, line 260: Same here, do not override if exist

src/composables/useDevices.js
```diff
@@ -106,6 +109,15 @@ export function useDevices(video, initializeOnMounted) {
 
 		virtualBackground.value.connectTrackSink('default', videoTrackToStream.value, 'video')
 
+		const blurBackgroundEnabled = settingsStore.blurBackgroundEnabled
+		if (blurBackgroundEnabled) {
```
Antreesy, line 113:
```suggestion
		if (settingsStore.blurBackgroundEnabled) {
```

And I think, same here: MediaSettings is doing it on open, MediaDevicesPreview should also have this logic on mounted, but not the shared composable

DorraJaouad: Okay, I apply the main blur setting on component level and not in composable (open for MediaSettings, and mount for MediaDevicesPreview)

src/components/MediaSettings/MediaSettings.vue
nickvergessen: When the setting is "Off", "None" should be the default.
When the setting is "On", "Blur" should be the default.
If the user changed the selection in an individual room, the Blur, Uploaded image or Select default background should be remembered for the future.
If the user has the setting "On" and manually selects "None", it is okay to ignore this in the next call and have it "Blur" again (due to deleting the entry from the browser storage)

DorraJaouad: > When the setting is "Off", "None" should be the default.

This will affect people who never used the global feature, or should I limit it to those who turned it on and off?

> If the user changed the selection in an individual room, the Blur, Uploaded image or Select default background should be remembered for the future.

By future, do you mean only next call or for all future call?

Antreesy: 1. When the setting is 'Off', it should behave as before PR (none is default, blur/image can be remembered per conv);
2. for all future calls in this conversation

vendor-bin/php-scoper/composer.json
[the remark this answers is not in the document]
DorraJaouad: Removed

nickvergessen (approved):
PHP code is fine!

PR #13499, opened 2024-10-09
feat(email): Recognize guests invited via email

### LINKED ISSUE #6098: recognice people invited by mail when they join the conversation
I really like the feature that I can add a random email address in the participants list and the person will get a invitation by mail:

image

But what is really disturbing is that when the person joins they show up as a guest and they are not recognized as the person invited by mail:

image

Idea, why not add the email address as a parameter to the link? If we don't want to have the email address visible, we could hash or encrypt it. This way we could match the user and show them with their email address as user name instead of an additional guest.

nickvergessen:
### ☑️ Resolves

* Fix #6098 

### 🚧 Tasks

- [x] Add access token to email link
- [ ] Should show that we identified the email in the "Guest name dialog"
- [x] Using the email as actor id might cause GDPR, need to hash at least?
    - [x] Still need a reliable email address to attendee way
- [x] Don't abuse user_status line for sending the email to moderators in the participant list
- [x] Many places assume guests are `actorType = Attendee::ACTOR_GUEST` while it should be `participantType = Participant::GUEST || participantType = Participant::GUEST_MODERATOR` ?
    - [x] Email users are not counted or listed in call summary
    - [x] Auto complete search plugin and mentioning
        - [x] Requires upstream: https://github.com/nextcloud/server/pull/48760
    - [x] ParserSystemMessage::isCurrentParticipantChangedUser
    - [x] Parse SystemMessage when the guest is modified (promoted/demoted/poll/…)
    - [x] System messages generated by the email-user can not find a display name and use wrong actor id
    - [x] Reference provider
    - [x] Cleaning up sessions by last ping?
    - [x] Message::isReplyable
    - [x] Notifier
    - [x] Search result display name
    - [x] Banning should disinvite
    - [x] Check bots
    - [x] Email guests should never expire?
    - [x] ParticipantService::getGuestCount check usage whether email/display names should be expended instead of count the emails as well
    - [x] Signaling participantsModified and roomInCallChanged excludes by actorType
    - [ ] …
- [ ] ⚡ Test HPB
- [ ] ☎️ Test SIP

### 🏁 Checklist


lib/Controller/RoomController.php
```diff
@@ -1076,6 +1076,14 @@ protected function formatParticipantList(array $participants, bool $includeStatu
 				$result['displayName'] = $participant->getAttendee()->getDisplayName();
 			} elseif ($participant->getAttendee()->getActorType() === Attendee::ACTOR_CIRCLES) {
 				$result['displayName'] = $participant->getAttendee()->getDisplayName();
+			} elseif ($participant->getAttendee()->getActorType() === Attendee::ACTOR_EMAILS) {
+				$result['displayName'] = $participant->getAttendee()->getDisplayName();
+				if ($this->participant->hasModeratorPermissions()) {
+					$result['status'] = IUserStatus::OFFLINE;
+					$result['statusIcon'] = null;
+					$result['statusMessage'] = $participant->getAttendee()->getInvitedCloudId();
```
nickvergessen, line 1084: This is to bring the email to the "status line".
I would prefer to add a dedicated response data entry instead, but then clients/browser don't work out of the box, but I guess that's fine.

lib/Service/ParticipantService.php
```diff
@@ -820,7 +815,12 @@ public function inviteEmailAddress(Room $room, string $email): Participant {
 		$attendee = new Attendee();
 		$attendee->setRoomId($room->getId());
 		$attendee->setActorType(Attendee::ACTOR_EMAILS);
-		$attendee->setActorId($email);
+		$attendee->setActorId($actorId);
+		$attendee->setInvitedCloudId($email);
```
nickvergessen, line 819: Unfortunate that we didn't name it `invited id` then it would be fitting. I guess it's okay to have it switch based on the actor type instead of adding yet another column?

skalidindi53 (approved):

lib/Service/BotService.php
```diff
@@ -191,7 +191,7 @@ protected function sendAsyncRequest(BotServer $botServer, array $body, ?string $
 		];
 
 		$data = [
-			'verify' => $this->certificateManager->getAbsoluteBundlePath(),
+			'verify' => false,// $this->certificateManager->getAbsoluteBundlePath(),
```
nickvergessen, line 194: this was not intentional

skalidindi53 (approved):

Antreesy:
For quick test without receiving actual e-mail:
`https://<NC_instance>/call/<token>?email=<email>&access=<access_token_from_DB>`

lib/Controller/PageController.php
```diff
@@ -405,6 +425,63 @@ protected function guestEnterRoom(string $token, string $password): Response {
...
+			$actorId = hash('sha256', $email);
+			$this->manager->getRoomByAccessToken(
+				$token,
+				Attendee::ACTOR_EMAILS,
+				$actorId,
+				$accessToken,
+			);
+			$this->talkSession->renewSessionId();
+			$this->talkSession->setAuthedEmailActorIdForRoom($token, $actorId);
+		} catch (RoomNotFoundException) {
```
Antreesy, line 448: Banned emails are also redirected as 404 to login page, that's what we want for now?

lib/Controller/PageController.php
```diff
@@ -332,12 +341,23 @@ public function recording(string $token): Response {
...
+		string $accessToken,
+	): Response {
+		if ($email && $accessToken) {
+			return $this->invitedEmail(
+				$token,
+				$email,
+				$accessToken,
+			);
+		}
 		try {
```
Antreesy, line 361: Same link without queries works as public room, but that's for step 2, I guess

Antreesy (commented):
LGTM, tested with inviting, banning, chatting

DorraJaouad:
Tested ✅

lib/Controller/PageController.php
nickvergessen: Well yes, that's the current behaviour atm

lib/Controller/PageController.php
nickvergessen: If you got banned, yes.

PR #13503, opened 2024-10-09
feat: show conversation settings, if URL hash is specified

Antreesy:
### ☑️ Resolves

* Ref #13485
* There's not much to be done on frontend side, applied it the same way we're using for `#direct-call`

2024-10-09_17h54_21

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -474,10 +474,6 @@ export default {
 			} else if (to.name === 'notfound') {
 				this.setPageTitle('')
 			}
-
-			if (to.hash === '#direct-call') {
-				emit('talk:media-settings:show')
-			}
 		})
 
 		if (getCurrentUser()) {
@@ -500,10 +496,6 @@ export default {
 			emit('toggle-navigation', {
 				open: true,
 			})
-		}
-
-		if (this.$route.hash === '#direct-call') {
-			emit('talk:media-settings:show')
 		}
 
 		subscribe('notifications:action:execute', this.interceptNotificationActions)
--- a/src/components/ConversationSettings/ConversationSettingsDialog.vue
+++ b/src/components/ConversationSettings/ConversationSettingsDialog.vue
@@ -267,10 +267,9 @@ export default {
 		}
 	},
 
-	mounted() {
+	beforeMount() {
 		subscribe('show-conversation-settings', this.handleShowSettings)
 		subscribe('hide-conversation-settings', this.handleHideSettings)
-
 	},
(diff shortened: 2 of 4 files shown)


src/App.vue
```diff
@@ -502,6 +504,8 @@ export default {
 
 		if (this.$route.hash === '#direct-call') {
 			emit('talk:media-settings:show')
+		} else if (this.$route.hash === '#settings') {
+			emit('show-conversation-settings', { token: this.token })
```
ShGKme, line 508: This feature is a conversation-only feature, but handled here on a whole app level:
- It handles hash and emits conversation-related event on all the pages without conversation (token)
- Though it works right now, there is no guarantee conversation components are mounted and ready at this moment. It will be broken if we add anything async to the conversation page init.
- It uses `settings` fragment, leaving no tinny name for Talk settings if we add it later

ShGKme: What about moving it to the conversation view?

ShGKme: And about removing the hash from URL when the settings dialog is open

Antreesy: - ConversationSettingsDialog is detached and independent from conversations. So it could be called on empty view as well, contrary to `#direct-call`
- It requires only token+conversation object to work. Maybe some additional request is needed here to ensure we fetched a conversation prior to opening the dialog
- Removing of hash should be fine, as long as it doesn't triggering any vue-router side effects

ShGKme: > * as long as it doesn't triggering any vue-router side effects

If it does, it is a bug that should be fixed

ShGKme: > It requires only token+conversation object to work

We don't have token+conversation outside the conversation view.
And a fragment cannot specify it. If we want to add ability to open setting for any conversation, it must be a query param. (Or if we want it to be combined with other fragments, like message id).

ShGKme: /apps/spreed/#settings

image

nickvergessen:
@Antreesy this still on track?

nickvergessen (approved):
Should `/index.php/apps/spreed/#settings` actually open the talk-settings? Not sure it could be useful?

Antreesy:
> Not sure it could be useful?

Unless we have a need, I'd skip it. Otherwise we have `OCA.Talk.Settings` added recently, which can be used from outside

PR #13506, opened 2024-10-10
feat(polls): Allow moderators to draft polls

nickvergessen:
### ☑️ Resolves

* Ref #13439 

### 🚧 Tasks

- [x] Allow posting drafts
- [x] Allow deleting drafts
- [x] Allow retrieving all drafts
- [x] Integration tests

### 🏁 Checklist


lib/Controller/PollController.php
```diff
@@ -91,33 +98,66 @@ public function createPoll(string $question, array $options, int $resultMode, in
...
-			$this->logger->error($e->getMessage(), ['exception' => $e]);
+			try {
+				$this->chatManager->addSystemMessage($this->room, $attendee->getActorType(), $attendee->getActorId(), $message, $this->timeFactory->getDateTime(), true);
+			} catch (\Exception $e) {
+				$this->logger->error($e->getMessage(), ['exception' => $e]);
+			}
 		}
 
-		return new DataResponse($this->renderPoll($poll, []), Http::STATUS_CREATED);
+		return new DataResponse($this->renderPoll($poll), Http::STATUS_CREATED);
```
Antreesy, line 130: This returns `Poll` in both cases, not the `PollDraft`

Antreesy (approved):
endpoints are tested against frontend, works as expected

lib/ResponseDefinitions.php
```diff
@@ -197,18 +197,21 @@
...
  *     actorId: string,
  *     actorType: string,
- *     details?: TalkPollVote[],
  *     id: int,
  *     maxVotes: int,
- *     numVoters?: int,
  *     options: string[],
  *     question: string,
  *     resultMode: int,
  *     status: int,
```
Antreesy, line 209: Possible to do?
```suggestion
 *     resultMode: 0|1,
 *     status: 0|1|2,
```

lib/Controller/PollController.php
nickvergessen: I'll fix that

lib/ResponseDefinitions.php
nickvergessen: - I attempted it in #12783 but it was too much to finish. Maybe I can start with a single place instead, so will have another look tomorrow

lib/Controller/PollController.php
```diff
@@ -69,18 +72,22 @@ public function __construct(
...
+			return $proxy->createPoll($this->room, $this->participant, $question, $options, $resultMode, $maxVotes, $draft);
 		}
 
 		if ($this->room->getType() !== Room::TYPE_GROUP
 			&& $this->room->getType() !== Room::TYPE_PUBLIC) {
 			return new DataResponse([], Http::STATUS_BAD_REQUEST);
 		}
 
+		if ($draft === true && !$this->participant->hasModeratorPermissions()) {
+			return new DataResponse([], Http::STATUS_BAD_REQUEST);
```
Antreesy, line 88: Should be 403?

nickvergessen: Well can argue in both ways. The request is bad because you send a prop you can not. Or we send forbidden. The question is what is better/easier to handle as an api consumer.
I'd say we stay with bad request and add the `['error' => 'draft']` similarly to other places

Antreesy: Ok. I forgot to hide the option for non-moderators on frontend, when we found out, so shouldn't be a problem

tests/integration/features/chat-3/poll.feature
```diff
@@ -805,3 +805,61 @@ Feature: chat-2/poll
     Then user "participant1" sees the following system messages in room "room" with 200 (v1)
       | room | actorType     | actorId      | systemMessage        | message                          | silent | messageParameters |
       | room | users         | participant1 | history_cleared      | You cleared the history of the conversation | !ISSET | {"actor":{"type":"user","id":"participant1","name":"participant1-displayname"}} |
+
+  Scenario: Drafts
+    Given user "participant1" creates room "room" (v4)
+      | roomType | 2 |
+      | roomName | room |
+    When user "participant1" adds user "participant2" to room "room" with 200 (v4)
+    When user "participant1" creates a poll in room "room" with 201
```
Antreesy, line 814:
```suggestion
    When user "participant1" creates a poll in room "room" with 200
```

Antreesy (approved):
Tested

PR #13507, opened 2024-10-10
fix: handle background blur in Talk if server doesn't support it

Antreesy:
### ☑️ Resolves

* Ref #12633
- partially reverts ed39a6bef13f7c1312459ebc610ba669d26501e0
- returns compatibility for Talk Desktop with versions below 29.0.4

🏚️ Before | 🏡 After
-- | --
image | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [x] Talk Desktop

### DIFF
--- a/src/components/CallView/CallView.vue
+++ b/src/components/CallView/CallView.vue
@@ -4,7 +4,7 @@
 -->
 
 <template>
-	<div id="call-container">
+	<div id="call-container" :style="callContainerStyle">
 		<ViewerOverlayCallView v-if="isViewerOverlay"
 			:token="token"
 			:model="promotedParticipantModel"
@@ -137,6 +137,7 @@ import { provide, ref } from 'vue'
 
 import { showMessage } from '@nextcloud/dialogs'
 import { subscribe, unsubscribe } from '@nextcloud/event-bus'
+import { loadState } from '@nextcloud/initial-state'
 import { t } from '@nextcloud/l10n'
 
 import Grid from './Grid/Grid.vue'
@@ -151,6 +152,7 @@ import ViewerOverlayCallView from './shared/ViewerOverlayCallView.vue'
 
 import { placeholderImage, placeholderModel, placeholderName, placeholderSharedData } from './Grid/gridPlaceholders.ts'
 import { SIMULCAST } from '../../constants.js'
+import BrowserStorage from '../../services/BrowserStorage.js'
 import { fetchPeers } from '../../services/callsService.js'
 import { getTalkConfig } from '../../services/CapabilitiesManager.ts'
 import { EventBus } from '../../services/EventBus.js'
@@ -158,6 +160,9 @@ import { useCallViewStore } from '../../stores/callView.js'
 import { useSettingsStore } from '../../stores/settings.js'
 import { localMediaModel, localCallParticipantModel, callParticipantCollection } from '../../utils/webrtc/index.js'
 import RemoteVideoBlocker from '../../utils/webrtc/RemoteVideoBlocker.js'
+
+const serverVersion = loadState('core', 'config', {}).versionstring ?? '29.0.0'
+const serverSupportsBackgroundBlurred = '29.0.4'.localeCompare(serverVersion) < 1
 
 export default {
 	name: 'CallView',
@@ -204,12 +209,17 @@ export default {
 			localMediaModel.disableVideo()
 		}
 
+		const isBackgroundBlurred = ref(serverSupportsBackgroundBlurred
+			? null
+			: BrowserStorage.getItem('background-blurred') !== 'false')
+
 		return {
 			localMediaModel,
 			localCallParticipantModel,
 			callParticipantCollection,
 			devMode,
 			callViewStore: useCallViewStore(),
+			isBackgroundBlurred,
 		}
 	},
 
@@ -379,6 +389,19 @@ export default {
 		supportedReactions() {
 			return getTalkConfig(this.token, 'call', 'supported-reactions')
 		},
+
+		/**
+		 * Fallback style for versions before v29.0.4
+		 */
+		callContainerStyle() {
+			if (serverSupportsBackgroundBlurred) {
+				return
+			}
+
+			return {
+				'backdrop-filter': this.isBackgroundBlurred ? 'blur(25px)' : 'none',
+			}
+		}
 	},
 
 	watch: {
@@ -470,6 +493,7 @@ export default {
 		callParticipantCollection.on('remove', this._lowerHandWhenParticipantLeaves)
 
 		subscribe('switch-screen-to-id', this._switchScreenToId)
+		subscribe('set-background-blurred', this.setBackgroundBlurred)
 	},
 
 	beforeDestroy() {
@@ -480,6 +504,7 @@ export default {
 		callParticipantCollection.off('remove', this._lowerHandWhenParticipantLeaves)
 
 		unsubscribe('switch-screen-to-id', this._switchScreenToId)
+		unsubscribe('set-background-blurred', this.setBackgroundBlurred)
 	},
 
 	methods: {
@@ -748,6 +773,14 @@ export default {
 			}
 		},
 
+		/**
+		 * Fallback method for versions before v29.0.4
+		 * @param {boolean} value whether background should be blurred
+		 */
+		setBackgroundBlurred(value) {
+			this.isBackgroundBlurred = value
+		},
+
 		isModelWithVideo(callParticipantModel) {
 			if (!callParticipantModel) {
 				return false
@@ -776,6 +809,7 @@ export default {
 	width: 100%;
 	height: 100%;
 	background-color: $color-call-background;
+	// Default value has changed since v29.0.4: 'blur(25px)' => 'none'
 	backdrop-filter: var(--filter-background-blur);
 	--grid-gap: calc(var(--default-grid-baseline) * 2);
 }
--- a/src/components/CallView/CallView.vue
+++ b/src/components/CallView/CallView.vue
@@ -4,7 +4,7 @@
 -->
 
 <template>
-	<div id="call-container" :style="callContainerStyle">
+	<div id="call-container" :class="callContainerClass">
 		<ViewerOverlayCallView v-if="isViewerOverlay"
 			:token="token"
 			:model="promotedParticipantModel"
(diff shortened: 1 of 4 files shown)


Antreesy:
Talk Desktop is locked on v25.0.2, so it will use fallback: https://github.com/nextcloud/talk-desktop/blob/a3ce89690df2a88b2d90b4cd54186b2c06c31673/src/shared/setupWebPage.js#L172

src/components/CallView/CallView.vue
```diff
@@ -151,12 +152,16 @@ import ViewerOverlayCallView from './shared/ViewerOverlayCallView.vue'
...
 import { SIMULCAST } from '../../constants.js'
+import BrowserStorage from '../../services/BrowserStorage.js'
 import { fetchPeers } from '../../services/callsService.js'
 import { getTalkConfig } from '../../services/CapabilitiesManager.ts'
 import { EventBus } from '../../services/EventBus.js'
 import { localMediaModel, localCallParticipantModel, callParticipantCollection } from '../../utils/webrtc/index.js'
 import RemoteVideoBlocker from '../../utils/webrtc/RemoteVideoBlocker.js'
 
+const serverVersion = loadState('core', 'config', {}).versionstring ?? '29.0.0'
+const serverSupportsBackgroundBlurred = '29.0.4'.localeCompare(serverVersion) < 1
```
DorraJaouad, line 163:
```
let text1 = "29.0.4"
let text2 = "29.0.10"
let result = text1.localeCompare(text2) // 1
```

Just split into 3 and compare one by one ?

src/components/CallView/CallView.vue
```diff
@@ -151,12 +152,16 @@ import ViewerOverlayCallView from './shared/ViewerOverlayCallView.vue'
...
 import { placeholderImage, placeholderModel, placeholderName, placeholderSharedData } from './Grid/gridPlaceholders.ts'
 import { SIMULCAST } from '../../constants.js'
+import BrowserStorage from '../../services/BrowserStorage.js'
 import { fetchPeers } from '../../services/callsService.js'
 import { getTalkConfig } from '../../services/CapabilitiesManager.ts'
 import { EventBus } from '../../services/EventBus.js'
 import { localMediaModel, localCallParticipantModel, callParticipantCollection } from '../../utils/webrtc/index.js'
 import RemoteVideoBlocker from '../../utils/webrtc/RemoteVideoBlocker.js'
 
+const serverVersion = loadState('core', 'config', {}).versionstring ?? '29.0.0'
```
DorraJaouad, line 162: image
`version` seems to be a better option

src/components/SettingsDialog/SettingsDialog.vue
```diff
@@ -286,12 +305,20 @@ export default {
...
+				console.debug('Blur was disabled intentionally, propagating last choice to server')
+				axios.post(generateOcsUrl('apps/provisioning_api/api/v1/config/users/theming/force_enable_blur_filter'),
+					{ configValue: 'no' })
+			}
+			BrowserStorage.removeItem('background-blurred')
+		} else {
+			// Fallback to BrowserStorage
+			if (blurred === null) {
+				BrowserStorage.setItem('background-blurred', 'true')
+			}
```
DorraJaouad, line 320:
```suggestion
		} else if (blurred === null) {
				// Fallback to BrowserStorage
				BrowserStorage.setItem('background-blurred', 'true')
			}
```

src/components/SettingsDialog/SettingsDialog.vue
```diff
@@ -347,6 +374,15 @@ export default {
 			this.privacyLoading = false
 		},
 
+		toggleBackgroundBlurred(value) {
+			if (serverSupportsBackgroundBlurred) {
+				return
+			}
```
DorraJaouad, line 380: You don't need this, it's checked in DOM triggering button

src/components/CallView/CallView.vue
```diff
@@ -370,6 +380,19 @@ export default {
...
+		/**
+		 * Fallback style for versions before v29.0.4
+		 */
+		callContainerStyle() {
+			if (serverSupportsBackgroundBlurred) {
+				return
+			}
+
+			return {
+				'backdrop-filter': this.isBackgroundBlurred ? 'blur(25px)' : 'none',
```
DorraJaouad, line 393: `this.isBackgroundBlurred` is a ref object and you should test against `this.isBackgroundBlurred.value`

src/components/CallView/CallView.vue
```diff
@@ -739,6 +764,14 @@ export default {
 			}
 		},
 
+		/**
+		 * Fallback method for versions before v29.0.4
+		 * @param {boolean} value whether background should be blurred
+		 */
+		setBackgroundBlurred(value) {
+			this.isBackgroundBlurred = value
```
DorraJaouad, line 772:
```suggestion
			this.isBackgroundBlurred.value = value
```

src/components/CallView/CallView.vue
```diff
@@ -196,11 +201,16 @@ export default {
 		const screenshotMode = ref(false)
 		provide('CallView:screenshotModeEnabled', screenshotMode)
 
+		const isBackgroundBlurred = ref(serverSupportsBackgroundBlurred
+			? null
+			: BrowserStorage.getItem('background-blurred') !== 'false')
```
DorraJaouad, line 206:
```suggestion
		const isBackgroundBlurred = ref(BrowserStorage.getItem('background-blurred') !== 'false')
```

Because you would need to evaluate its value only when `serverSupportsBackgroundBlurred` is falsy anyway

DorraJaouad (commented):
I haven't tested yet 🚧

src/components/CallView/CallView.vue
Antreesy: .value is only valid for composition api code

Antreesy:
rebased onto webpack 6.2.0, all changes since last review are in 2nd commit (for backport to 29)

src/utils/__tests__/satisfyVersion.spec.js
```diff
@@ -0,0 +1,26 @@
...
+ */
+import { satisfyVersion } from '../satisfyVersion.ts'
+
+describe('satisfyVersion', () => {
+	const testCases = [
+		// Older
+		['29.1.4.3', '28.2.5.4', true],
+		['29.1.4.3', '29.0.15.3', true],
+		['29.1.4.3', '29.1.1.3', true],
+		['29.1.4.3', '29.1.4.0', true],
```
nickvergessen, line 13:
```suggestion
		['29.1.4.3', '29.1.4.0', true],
		['29.1.4.3', '29.1.4', true],
		['29.1.4.3', '29.1', true],
		['29.1.4.3', '29', true],
```

src/utils/__tests__/satisfyVersion.spec.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+import { satisfyVersion } from '../satisfyVersion.ts'
+
+describe('satisfyVersion', () => {
+	const testCases = [
+		// Older
```
nickvergessen, line 9:
```suggestion
		// Newer
```

src/utils/__tests__/satisfyVersion.spec.js
```diff
@@ -0,0 +1,26 @@
...
+		['29.1.4.3', '28.2.5.4', true],
+		['29.1.4.3', '29.0.15.3', true],
+		['29.1.4.3', '29.1.1.3', true],
+		['29.1.4.3', '29.1.4.0', true],
+		// Exact
+		['29.1.4.3', '29.1.4.3', true],
+		// Newer
```
nickvergessen, line 16:
```suggestion
		// Older
```

src/utils/__tests__/satisfyVersion.spec.js
```diff
@@ -0,0 +1,26 @@
...
+		['29.1.4.3', '29.1.4.4', false],
+		['29.1.4.3', '29.1.5.0', false],
+		['29.1.4.3', '29.2.0.0', false],
+		['29.1.4.3', '30.0.0.0', false],
```
nickvergessen, line 20:
```suggestion
		['29.1.4.3', '30.0.0.0', false],
		['29.1.4.3', '29.1.5', false],
		['29.1.4.3', '29.2', false],
		['29.1.4.3', '30', false],
```

src/utils/__tests__/satisfyVersion.spec.js
Antreesy: for me `older` is what was published before (like old). maybe, pick something more definitive?

src/utils/satisfyVersion.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Checks if given version satisfy requirements (newer than)
+ * Versions are expected to have format like '29.1.4.3'
+ * @param version given version
+ * @param requirement version to compare against
+ */
+function satisfyVersion(version: string, requirement: string): boolean {
+	const versionMap = version.split('.').map(Number)
+	const requirementMap = requirement.split('.').map(Number)
```
ShGKme, line 14: This won't work with pre-releases

ShGKme: We can use [semver](https://www.npmjs.com/package/semver) package

src/components/CallView/CallView.vue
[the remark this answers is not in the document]
Antreesy: We don't backport further, so i think, it would be enough for this exact feature

src/utils/satisfyVersion.ts
Antreesy: I changed it to rely on `version` and not `versionstring`. This comes from version.php, so should always be number tuple?
```php
$OC_Version = [31, 0, 0, 3];

// The human-readable string
$OC_VersionString = '31.0.0 dev';
```

src/components/CallView/CallView.vue
```diff
@@ -370,6 +380,19 @@ export default {
...
+		 */
+		callContainerStyle() {
+			if (serverSupportsBackgroundBlurred) {
+				return
+			}
+
+			return {
+				'backdrop-filter': this.isBackgroundBlurred ? 'blur(25px)' : 'none',
+			}
+		}
```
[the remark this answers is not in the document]
Antreesy: I couldn't find a nice-looking solution, given that we need to overwrite any backdrop value coming from server. See screenshot from 29.0.3:
image
Any thoughts? can also keep only 'blur none' option for DefaultTheme and not affect HighContrast ones?

src/utils/satisfyVersion.ts
nickvergessen: for the server version yes, for talk version no

Antreesy: For talk we have capabilities to count such cases.
And I don't feel like adding semver to the prod dependencies, given that it's check is relevant only for web 29 and desktop client

src/components/CallView/CallView.vue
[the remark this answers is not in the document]
ShGKme: But it is about server version, not Talk version, right?

DorraJaouad (approved):
Tested ✅ and good code-wise

ShGKme (approved):

Antreesy:
/backport 503f63902c25b60dc2865bb5ed2939c7bdf24cad to stable29

PR #13518, opened 2024-10-11
feat: draft polls management 📊

Antreesy:
### ☑️ Resolves

* Follow-up to #13516

Creation | Open poll
-- | -- 
image | image 
Closed poll | Shared Items
image | image
Filled, from Poll Editor | Empty, from Shared items
image | image
Opened from DraftHandler ('Back' arrow) | -
image | -

### 🚧 Tasks

- [x] Dialog + Placement
  - [x] Empty View
- [x] Draft deletion
- [ ] Follow-ups:
  - [ ] Poll editor to 'normal' width
    - [ ] Make fields clearable
    - [ ] add EmojiPicker
  - [ ] Unify editor + drafts into a single dialog to avoid re-rendering

### 🏁 Checklist

  - [x] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/NewMessage/NewMessage.vue
+++ b/src/components/NewMessage/NewMessage.vue
@@ -151,6 +151,10 @@
 			:token="token"
 			@close="togglePollEditor" />
 
+		<PollDraftHandler v-if="canCreatePollDrafts && showPollDraftHandler"
+			:token="token"
+			@close="togglePollDraftHandler" />
+
 		<!-- New file creation dialog -->
 		<NewMessageNewFileDialog v-if="showNewFileDialog !== -1"
 			:token="token"
@@ -194,6 +198,7 @@ import NewMessageAudioRecorder from './NewMessageAudioRecorder.vue'
 import NewMessageNewFileDialog from './NewMessageNewFileDialog.vue'
 import NewMessagePollEditor from './NewMessagePollEditor.vue'
 import NewMessageTypingIndicator from './NewMessageTypingIndicator.vue'
+import PollDraftHandler from '../PollViewer/PollDraftHandler.vue'
 import Quote from '../Quote.vue'
 
 import { useIsDarkTheme } from '../../composables/useIsDarkTheme.ts'
@@ -226,6 +231,7 @@ export default {
 		NewMessageAudioRecorder,
 		NewMessageNewFileDialog,
 		NewMessagePollEditor,
+		PollDraftHandler,
 		NewMessageTypingIndicator,
 		Quote,
 		// Icons
@@ -301,6 +307,7 @@ export default {
 			// True when the audio recorder component is recording
 			isRecordingAudio: false,
 			showPollEditor: false,
+			showPollDraftHandler: false,
 			showNewFileDialog: -1,
 			showFilePicker: false,
 			// Check empty template by default
@@ -384,6 +391,10 @@ export default {
 		canCreatePoll() {
 			return !this.isOneToOne && !this.noChatPermission
 				&& this.conversation.type !== CONVERSATION.TYPE.NOTE_TO_SELF
+		},
+
+		canCreatePollDrafts() {
+			return hasTalkFeature(this.token, 'talk-polls-drafts') && this.$store.getters.isModerator
 		},
 
 		currentConversationIsJoined() {
@@ -538,6 +549,7 @@ export default {
 		EventBus.on('upload-discard', this.handleUploadSideEffects)
 		EventBus.on('retry-message', this.handleRetryMessage)
 		EventBus.on('smart-picker-open', this.handleOpenTributeMenu)
+		EventBus.on('poll-drafts-open', this.togglePollDraftHandler)
 
 		if (!this.$store.getters.areFileTemplatesInitialised) {
 			this.$store.dispatch('getFileTemplates')
@@ -550,6 +562,7 @@ export default {
 		EventBus.off('upload-discard', this.handleUploadSideEffects)
 		EventBus.off('retry-message', this.handleRetryMessage)
 		EventBus.off('smart-picker-open', this.handleOpenTributeMenu)
+		EventBus.off('poll-drafts-open', this.togglePollDraftHandler)
 	},
 
 	methods: {
@@ -890,6 +903,10 @@ export default {
 
 		togglePollEditor() {
 			this.showPollEditor = !this.showPollEditor
+		},
+
+		togglePollDraftHandler() {
+			this.showPollDraftHandler = !this.showPollDraftHandler
 		},
 
 		async autoComplete(search, callback) {
--- a/src/components/NewMessage/NewMessage.vue
+++ b/src/components/NewMessage/NewMessage.vue
@@ -148,11 +148,13 @@
 
 		<!-- Poll creation dialog -->
 		<NewMessagePollEditor v-if="showPollEditor"
+			ref="pollEditor"
 			:token="token"
 			@close="togglePollEditor" />
 
 		<PollDraftHandler v-if="canCreatePollDrafts && showPollDraftHandler"
 			:token="token"
+			:show-create-button="!showPollEditor"
 			@close="togglePollDraftHandler" />
 
 		<!-- New file creation dialog -->
@@ -549,6 +551,7 @@ export default {
 		EventBus.on('upload-discard', this.handleUploadSideEffects)
 		EventBus.on('retry-message', this.handleRetryMessage)
 		EventBus.on('smart-picker-open', this.handleOpenTributeMenu)
+		EventBus.on('poll-editor-open', this.fillPollEditorFromDraft)
 		EventBus.on('poll-drafts-open', this.togglePollDraftHandler)
 
 		if (!this.$store.getters.areFileTemplatesInitialised) {
@@ -562,6 +565,7 @@ export default {
 		EventBus.off('upload-discard', this.handleUploadSideEffects)
 		EventBus.off('retry-message', this.handleRetryMessage)
 		EventBus.off('smart-picker-open', this.handleOpenTributeMenu)
+		EventBus.off('poll-editor-open', this.fillPollEditorFromDraft)
 		EventBus.off('poll-drafts-open', this.togglePollDraftHandler)
 	},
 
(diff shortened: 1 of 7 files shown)


src/components/PollViewer/PollDraftHandler.vue
```diff
@@ -0,0 +1,100 @@
+<!--
+  - SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+	<NcDialog class="drafts"
```
DorraJaouad, line 7: Make it bigger so it can hide the other dialog? 
image

Or close that modal once this is open. 
A good addition would be to rename the button to "Back" instead of "Create a new poll" when it's opened from poll editor :).  The flow would make sense much better.

src/components/NewMessage/NewMessagePollEditor.vue
```diff
@@ -58,6 +68,14 @@
 			<NcButton type="tertiary" @click="emit('close')">
 				{{ t('spreed', 'Dismiss') }}
 			</NcButton>
+			<NcActions v-if="supportPollDrafts" force-menu>
+				<NcActionButton v-if="isModerator" :disabled="!isFilled" @click="createPollDraft">
+					<template #icon>
+						<IconFileEdit :size="20" />
+					</template>
+					{{ t('spreed', 'Save as draft') }}
```
DorraJaouad, line 76: I might be influenced by github but I expected to find this button in a menu with "Create draft" but not like we have in CallButton (you can press directly "create draft" but for "Save as draft" needs to click on the down arrow button and then click on it).  So basically, make the 3 dot icon a down arrow and attach it to the "create draft". 

There is a discussion about this type of buttons for call button so we can change it later once we agree on using it.

src/components/NewMessage/NewMessage.vue
```diff
@@ -148,9 +148,14 @@
 
 		<!-- Poll creation dialog -->
 		<NewMessagePollEditor v-if="showPollEditor"
+			ref="pollEditor"
 			:token="token"
 			@close="togglePollEditor" />
 
+		<PollDraftHandler v-if="showPollDraftHandler"
+			:token="token"
+			@close="togglePollDraftHandler" />
```
DorraJaouad, line 157: Maybe in `SharedItemsTab` ?
It doesn't fit in both places 100% 🤔

src/components/NewMessage/NewMessagePollEditor.vue
```diff
@@ -150,10 +182,54 @@ async function createPoll() {
...
+	EventBus.emit('poll-drafts-open')
+}
 </script>
 
 <style lang="scss" scoped>
 .poll-editor {
+	&__wrapper {
+		display: flex;
+		align-items: flex-end;
+		gap: 4px;
```
DorraJaouad, line 230:
```suggestion
		gap: var(--default-grid-baseline);
```

DorraJaouad (commented):
Functionality works well

src/components/NewMessage/NewMessagePollEditor.vue
```diff
@@ -58,6 +68,14 @@
 			<NcButton type="tertiary" @click="emit('close')">
 				{{ t('spreed', 'Dismiss') }}
 			</NcButton>
+			<NcActions v-if="supportPollDrafts" force-menu>
+				<NcActionButton v-if="isModerator" :disabled="!isFilled" @click="createPollDraft">
```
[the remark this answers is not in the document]
Antreesy: I thought of 'Export draft to file', which should be allowed to non-moderators

src/components/NewMessage/NewMessagePollEditor.vue
```diff
@@ -12,7 +12,17 @@
...
+		<div class="poll-editor__wrapper">
+			<NcTextField :value.sync="pollForm.question" :label="t('spreed', 'Ask a question')" v-on="$listeners" />
+			<NcActions v-if="supportPollDrafts" force-menu>
+				<NcActionButton v-if="isModerator" close-after-click @click="openPollDraftHandler">
+					<template #icon>
+						<IconFileEdit :size="20" />
+					</template>
+					{{ t('spreed', 'Browse poll drafts') }}
+				</NcActionButton>
+			</NcActions>
```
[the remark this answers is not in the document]
Antreesy: Same here

src/components/NewMessage/NewMessagePollEditor.vue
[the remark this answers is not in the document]
DorraJaouad: I don't think it should be allowed to non-moderators, cc @nickvergessen

src/components/NewMessage/NewMessagePollEditor.vue
```diff
@@ -150,10 +182,54 @@ async function createPoll() {
...
+ */
+async function createPollDraft() {
+	const poll = await pollsStore.createPollDraft({
+		token: props.token,
+		form: pollForm,
+	})
+	if (poll) {
+		showSuccess(t('spreed', 'Poll draft has been saved'))
+	}
+}
```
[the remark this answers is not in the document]
Antreesy: moved toasts to the store

src/components/PollViewer/PollDraftHandler.vue
```diff
@@ -0,0 +1,100 @@
...
+	(event: 'close'): void,
+}>()
+
+const supportPollDrafts = hasTalkFeature(props.token, 'talk-polls-drafts')
+const store = useStore()
+const pollsStore = usePollsStore()
+/**
+ * Receive poll drafts for the current conversation as owner/moderator
+ */
+const isModerator = computed(() => (store.getters as unknown).isModerator)
```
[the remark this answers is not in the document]
Antreesy: Moved to NewMessage, where it's rendered

src/components/PollViewer/PollDraftHandler.vue
Antreesy: I'd just hide the button in case PollEditor is open (as we can see it behind), but don't close it, like ConversationSettings dialog stays behind while you're checking bans - subfeature on top of main one.

Besides, PollEditor can be larger if there are more answers filled. Though we can squezee drafts dialog to 2 drafts in a row?

src/components/NewMessage/NewMessagePollEditor.vue
Antreesy: Like this?
image

src/components/NewMessage/NewMessage.vue
Antreesy: We don't have (yet) Shared Items tab in federated rooms
And NewMessage is a natural requirement for PollEditor -> Poll drafts

src/components/NewMessage/NewMessagePollEditor.vue
Antreesy: cc @marcoambrosini

DorraJaouad: I find really nice and intuitive, maybe 3 dot is to be replaced with down arrow

src/components/PollViewer/PollDraftHandler.vue
DorraJaouad: I'd just hide the button in case PollEditor is open (as we can see it behind), but don't close it, like ConversationSettings dialog stays behind while you're checking bans - subfeature on top of main one.

Fine by me, but I'd still prefer renaming it to "back"

> Besides, PollEditor can be larger if there are more answers filled. Though we can squezee drafts dialog to 2 drafts in a row?

Yes, but we should cover all scenarios. It needs to have a default height and width that can cover the poll editor ?

src/components/NewMessage/NewMessage.vue
DorraJaouad: Ah, I see your point. Let's keep it here then

src/components/NewMessage/NewMessagePollEditor.vue
marcoambrosini: I think these two actions are not related, so I wouldn't group them. Taking into account how advanced this feature is, I would just add an unlabelled "save" button to the left of the primary action. We can also remove the dismiss button since we already have the close dialog close button

Antreesy: > to the left of the primary action

Thing is, we might want to extend the actions range (+ export to file, for example). But without a 'Dismiss' button three-dot menu should already look better

Antreesy: Dropped dismiss button for now and kept three-dot on the left (like in Media settings)

src/components/PollViewer/PollDraftHandler.vue
Antreesy: Hid the button for now. Size issue can be solved in follow-up (I thought of extending PollEditor to normal size, maybe we'll shrink drafts to 1-2 row with that change)

src/components/NewMessage/NewMessagePollEditor.vue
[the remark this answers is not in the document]
Antreesy: Empty NcActions (with no button) will not be rendered, so non-moderators can't see three-dot menu for now:
image

Let's discuss 'Export draft to file' in latter PR

src/components/PollViewer/PollDraftHandler.vue
```diff
@@ -0,0 +1,96 @@
...
+				:key="item.id"
+				:token="token"
+				:name="item.question"
+				draft
+				@click="openPollEditor" />
+		</div>
+		<template v-if="props.showCreateButton" #actions>
+			<NcButton @click="openPollEditor(null)">
+				{{ t('spreed', 'Create new poll') }}
+			</NcButton>
```
DorraJaouad, line 33: So, this is basically shown when opening from shared items tab, it will also need  a back button so the flow could be completed and revertible. 

I am not sure if it makes sense to keep the browse draft button too because you were just in the drafts handler before clicking "create new poll", it's intuitive to go back and select from there.

src/components/NewMessage/NewMessagePollEditor.vue
```diff
@@ -55,9 +65,14 @@
 			</NcCheckboxRadioSwitch>
 		</div>
 		<template #actions>
-			<NcButton type="tertiary" @click="emit('close')">
-				{{ t('spreed', 'Dismiss') }}
-			</NcButton>
```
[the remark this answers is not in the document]
Antreesy: showTrailingButton -> follow-up

DorraJaouad (approved):
Good to go 🦅

PR #13519, opened 2024-10-11
feat(calls): Allow moderators to download a call participants list

### LINKED ISSUE #13453: Moderators must be able to download the participant list as PDF
### How to use GitHub

---

- [x] CSV #13519 
- [ ] PDF
- [x] Frontend

nickvergessen:
### ☑️ Resolves

* Fix #13453 

### 🚧 Tasks

- [x] Download as CSV
- [x] Add docs
- [x] Add tests

### 🏁 Checklist

### DIFF
--- a/lib/Controller/CallController.php
+++ b/lib/Controller/CallController.php
@@ -17,6 +17,7 @@ use OCA\Talk\Middleware\Attribute\FederationSupported;
 use OCA\Talk\Middleware\Attribute\RequireCallEnabled;
 use OCA\Talk\Middleware\Attribute\RequireFederatedParticipant;
 use OCA\Talk\Middleware\Attribute\RequireModeratorOrNoLobby;
+use OCA\Talk\Middleware\Attribute\RequireModeratorParticipant;
 use OCA\Talk\Middleware\Attribute\RequireParticipant;
 use OCA\Talk\Middleware\Attribute\RequirePermission;
 use OCA\Talk\Middleware\Attribute\RequireReadWriteConversation;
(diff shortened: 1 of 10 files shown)


lib/Controller/CallController.php
```diff
@@ -111,6 +114,53 @@ public function getPeersForCall(): DataResponse {
...
+			return new Response(Http::STATUS_BAD_REQUEST);
+		}
+
+		if ($format !== 'csv' && $format !== 'pdf') {
+			// Unsupported format
+			return new Response(Http::STATUS_BAD_REQUEST);
+		}
+
+		if ($format !== 'csv') {
+			// FIXME Remove once pdf was implemented.
```
nickvergessen, line 144: *Note:* Currently only CSV is functional

lib/Controller/CallController.php
```diff
@@ -111,6 +114,53 @@ public function getPeersForCall(): DataResponse {
...
+			return new Response(Http::STATUS_BAD_REQUEST);
+		}
+
+		$output = fopen('php://memory', 'w');
+		fputcsv($output, [
+			'name',
+			'type',
+			'identifier',
+		]);
```
DorraJaouad, line 153: Should we prettify the output? (headers with capital letters) 
image

DorraJaouad (approved):
Tested 👍🏻

lib/Controller/CallController.php
nickvergessen: We can do that for the PDF version later. For the CSV I would keep it "ugly"

Antreesy: as-is should be less confusing for later use (like import)

PR #13538, opened 2024-10-14
chore: Use more gender neutral language

provokateurin:
Frontend strings are all neutral already, just code comments which are not translated.

### DIFF
--- a/tests/php/Controller/SignalingControllerTest.php
+++ b/tests/php/Controller/SignalingControllerTest.php
@@ -1178,7 +1178,7 @@ class SignalingControllerTest extends TestCase {
 		$participant = $participantService->getParticipant($room, $this->userId, $oldSessionId);
 		$this->assertEquals($oldSessionId, $participant->getSession()->getSessionId());
 
-		// The user is reloading the browser which will join him with another
+		// The user is reloading the browser which will join them with another
 		// session id.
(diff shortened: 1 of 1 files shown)


src/types/openapi/openapi-full.ts
```diff
@@ -479,7 +479,7 @@ export type paths = {
...
-         *     In any case, to create or even get the token of the room, the file must be shared and the user must be the owner of a public share of the file (like a link share, for example) or have direct access to that file; an error is returned otherwise. A user has direct access to a file if she has access to it (or to an ancestor) through a user, group, circle or room share (but not through a link share, for example), or if she is the owner of such a file.
+         *     In any case, to create or even get the token of the room, the file must be shared and the user must be the owner of a public share of the file (like a link share, for example) or have direct access to that file; an error is returned otherwise. A user has direct access to a file if she has access to it (or to an ancestor) through a user, group, circle or room share (but not through a link share, for example), or if they are the owner of such a file.
```
Antreesy, line 482:
```suggestion
         *     In any case, to create or even get the token of the room, the file must be shared and the user must be the owner of a public share of the file (like a link share, for example) or have direct access to that file; an error is returned otherwise. A user has direct access to a file if they have access to it (or to an ancestor) through a user, group, circle or room share (but not through a link share, for example), or if they are the owner of such a file.
```

Antreesy (approved):

PR #13557, opened 2024-10-15
feat(calls): Allow moderators to download a call participants list - frontend

### LINKED ISSUE #13453: Moderators must be able to download the participant list as PDF

DorraJaouad:
### ☑️ Resolves

* Ref #13453 

During call | Output
-- | --
image | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/lib/Controller/CallController.php
+++ b/lib/Controller/CallController.php
@@ -38,6 +38,7 @@ use OCP\AppFramework\Http\DataDownloadResponse;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\Http\Response;
 use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\IConfig;
 use OCP\IRequest;
 use OCP\IUserManager;
 
@@ -55,6 +56,7 @@ class CallController extends AEnvironmentAwareController {
 		private RoomService $roomService,
 		private IUserManager $userManager,
 		private ITimeFactory $timeFactory,
+		private IConfig $serverConfig,
 		private Config $talkConfig,
 		protected Authenticator $federationAuthenticator,
 		private SIPDialOutService $dialOutService,
@@ -163,9 +165,21 @@ class CallController extends AEnvironmentAwareController {
 		$cleanedRoomName = preg_replace('/[\/\\:*?"<>|\- ]+/', '-', $this->room->getName());
 		// Limit to a reasonable length
 		$cleanedRoomName = substr($cleanedRoomName, 0, 100);
-		$date = $this->timeFactory->getDateTime()->format('Y-m-d');
+
+		$timezone = 'UTC';
+		if ($this->participant->getAttendee()->getActorType() === Attendee::ACTOR_USERS) {
+			$timezone = $this->serverConfig->getUserValue($this->participant->getAttendee()->getActorId(), 'core', 'timezone', 'UTC');
+		}
+
+		try {
+			$dateTimeZone = new \DateTimeZone($timezone);
+		} catch (\DateInvalidTimeZoneException) {
+			$dateTimeZone = null;
+		}
+
+		$date = $this->timeFactory->getDateTime('now', $dateTimeZone)->format('Y-m-d');
 		$fileName = $cleanedRoomName . ' ' . $date . '.csv';
-	
+
 		return new DataDownloadResponse(stream_get_contents($output), $fileName, 'text/csv');
 	}
 
--- a/lib/Controller/CallController.php
+++ b/lib/Controller/CallController.php
@@ -127,6 +127,7 @@ class CallController extends AEnvironmentAwareController {
 	 */
 	#[PublicPage]
 	#[RequireModeratorParticipant]
+	#[Http\Attribute\NoCSRFRequired]
 	public function downloadParticipantsForCall(string $format = 'csv'): DataDownloadResponse|Response {
 		$timeout = $this->timeFactory->getTime() - Session::SESSION_TIMEOUT;
 		$participants = $this->participantService->getParticipantsInCall($this->room, $timeout);
@@ -158,7 +159,14 @@ class CallController extends AEnvironmentAwareController {
 
 		fseek($output, 0);
 
-		return new DataDownloadResponse(stream_get_contents($output), 'participants.csv', 'text/csv');
+		// Clean the room name
+		$cleanedRoomName = preg_replace('/[\/\\:*?"<>|\- ]+/', '-', $this->room->getName());
+		// Limit to a reasonable length
+		$cleanedRoomName = substr($cleanedRoomName, 0, 100);
+		$date = $this->timeFactory->getDateTime()->format('Y-m-d');
+		$fileName = $cleanedRoomName . ' ' . $date . '.csv';
+	
+		return new DataDownloadResponse(stream_get_contents($output), $fileName, 'text/csv');
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


Ivansss:
Just as an idea, would be possible to add this option somewhere in the participant list?

Somewhere here maybe?
Screenshot 2024-10-15 at 16 30 24

Antreesy:
> Somewhere here maybe?

Maybe as a second entry, if it would feel needed there.
But then this section needs some rework

DorraJaouad:
> Maybe as a second entry, if it would feel needed there.

I'd go for one entry only IMO. 

> Just as an idea, would be possible to add this option somewhere in the participant list?

We also thought it will fit better in participants tab but the current design doesn't help much in putting it there :').

src/components/TopBar/TopBarMenu.vue
```diff
@@ -492,6 +504,24 @@ export default {
 				token: this.token,
 			})
 		},
+
+		async downloadAttendanceList() {
+			try {
+				const reponse = await callDownloadParticipants(this.token)
+				// Create a link and click it to download the file
+				const url = window.URL.createObjectURL(new Blob([reponse.data]))
+				const link = document.createElement('a')
```
nickvergessen, line 513: Instead of all this magic you could simply let the browser do the request directly?
The backend is setting a beautiful file name and all :see_no_evil:

Antreesy: Sadly, it's not that simple
image

nickvergessen: Hmm, have to add `#[Http\Attribute\NoCSRFRequired]` when it's not an API call, I see.
Also we are greeten by the "Leave call" dialog. So maybe it's better to do an intercept :see_no_evil: 

I was about to do a rename of the file thou, so maybe we just have to do it in the frontend then?
Suggestion: `<cleaned room name> - <cleaned date and time>.csv`
cleaned means: replace all "breaking" characters with `-` and replace multiple `--` with a single one.

E.g. `Hello/World:://|?Tour🌎.pdf` room would generate something like: `Hello-World-Tour-pdf - 2024-10-14 18-05.csv` ( https://github.com/nextcloud/server/blob/0936af996013ab07fe845fb852b80b3ab1d536f0/apps/files/lib/Service/SettingsService.php#L27-L31 and `/` and `\` and `.`) and limit to reasonable length, so that we don't even remotely reach the 255 character limit of Windows :see_no_evil:

Antreesy: `#[Http\Attribute\NoCSRFRequired]` is enough to use NcActionLink =)
But we still need to generate a file name on a client. Don't have any cleanup util on a client side though

nickvergessen: > But we still need to generate a file name on a client.

With NcActionLink it worked for me without? The server is setting the correct Header and therefore it downloads as `participant.csv` for me atm.

Antreesy (approved):
Tested. Filename change we can do separately on backend, right?

DorraJaouad:
> Filename change we can do separately on backend

As it is received from backend, I assumed it will be modified there. I am find with adding it in this PR

lib/Controller/CallController.php
```diff
@@ -158,7 +159,14 @@ public function downloadParticipantsForCall(string $format = 'csv'): DataDownloa
 
 		fseek($output, 0);
 
-		return new DataDownloadResponse(stream_get_contents($output), 'participants.csv', 'text/csv');
+		// Clean the room name
+		$cleanedRoomName = preg_replace('/-+/', '-', preg_replace('/[\/\\:*?"<>|]/', '-', $this->room->getName()));
+		// Limit to a reasonable length
+		$cleanedRoomName = substr($cleanedRoomName, 0, 100);
+		$date = (new \DateTime())->format('Y-m-d');
+		$fileName = "{$cleanedRoomName} - {$date}.csv";
```
Antreesy, line 167: Maybe without an additional hyphen?
```suggestion
		$fileName = "{$cleanedRoomName} {$date}.csv";
```

lib/Controller/CallController.php
```diff
@@ -158,7 +159,14 @@ public function downloadParticipantsForCall(string $format = 'csv'): DataDownloa
 
 		fseek($output, 0);
 
-		return new DataDownloadResponse(stream_get_contents($output), 'participants.csv', 'text/csv');
+		// Clean the room name
+		$cleanedRoomName = preg_replace('/-+/', '-', preg_replace('/[\/\\:*?"<>|]/', '-', $this->room->getName()));
```
nickvergessen, line 163:
```suggestion
		$cleanedRoomName = preg_replace('/[\/\\:*?"<>|\-]+/', '-', $this->room->getName());
```

lib/Controller/CallController.php
```diff
@@ -158,7 +159,14 @@ public function downloadParticipantsForCall(string $format = 'csv'): DataDownloa
...
+		// Limit to a reasonable length
+		$cleanedRoomName = substr($cleanedRoomName, 0, 100);
+		$date = (new \DateTime())->format('Y-m-d');
```
nickvergessen, line 166:
```suggestion
		$date = $this->timeFactory->getDateTime()->format('Y-m-d');
```

DorraJaouad: Is it in UTC time because it is coming from server?

nickvergessen: It would be, but we can inject the user's timezone

nickvergessen:
Pushed the user timezone thing

nickvergessen (approved):
Works now and fine API wise

lib/Controller/CallController.php
```diff
@@ -158,7 +161,26 @@ public function downloadParticipantsForCall(string $format = 'csv'): DataDownloa
...
+			$timezone = $this->serverConfig->getUserValue($this->participant->getAttendee()->getActorId(), 'core', 'timezone', 'UTC');
+		}
+
+		try {
+			$dateTimeZone = new \DateTimeZone($timezone);
+		} catch (\DateInvalidTimeZoneException) {
+			$dateTimeZone = null;
+		}
+
+		$date = $this->timeFactory->getDateTime('now', $dateTimeZone)->format('Y-m-d');
```
DorraJaouad, line 180:
```suggestion
		$date = $this->timeFactory->getDateTime('now', $dateTimeZone)->format('Y-m-d H-i');
```

?

Antreesy: Unlikely someone would have a call in the same conversation with different participants twice the same day? Even so, they could rename it later themselves

DorraJaouad: but it will be nice to have, to know when you took the snapshot of participants (which helps a lot for student attendance timing).

Antreesy: Then it will come to decision-making (e.g. students claim they've attended the call), timestamp of file creation will still be more valid than a file name

Antreesy (approved):

src/components/TopBar/TopBarMenu.vue
```diff
@@ -139,6 +139,14 @@
 				</template>
 				{{ t('spreed', 'Conversation settings') }}
 			</NcActionButton>
+			<NcActionLink v-if="isInCall && canModerate"
```
Antreesy, line 142: capability check is missing

PR #13593, opened 2024-10-19
fix(activity): Fix parameter type of call activity

nickvergessen:
### 🏁 Checklist


nickvergessen:
Way to annoying that psalm doesn't complain just because we loose the type when passing the array though multiple levels...

provokateurin (approved):

PR #13607, opened 2024-10-22
fix: localURL previews

Antreesy:
### ☑️ Resolves

* Fix broken video preview, when user uploads
* Use local files for voice messages, when user uploads
* Fix store usage, when no preview is needed

⚠️ Technically, fallbackLocalUrl is not needed for upload rather than images, but if preview is available and server doesn't return it, client should fall back to mime icon anyway, so no need to duplicate it

🏚️ Before | 🏡 After
-- | --
Voice messages | --
2024-10-22_16h44_23 | 2024-10-22_16h42_54
Video previews | --
2024-10-22_16h46_25 | 2024-10-22_16h47_19

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop
- [ ] ⛑️ Tests are included or not possibleSigned-off-by: Maksim Sukharev <antreesy.web@gmail.com>

### DIFF
--- a/src/store/fileUploadStore.js
+++ b/src/store/fileUploadStore.js
@@ -127,7 +127,9 @@ const mutations = {
 			totalSize: file.size,
 			temporaryMessage,
 		 })
-		Vue.set(state.localUrls, temporaryMessage.referenceId, localUrl)
+		if (localUrl) {
+			Vue.set(state.localUrls, temporaryMessage.referenceId, localUrl)
+		}
 	},
 
 	// Marks a given file as initialized (for retry)
@@ -244,8 +246,6 @@ const actions = {
 				localUrl = URL.createObjectURL(file)
 			} else if (isVoiceMessage) {
 				localUrl = file.localUrl
-			} else {
-				localUrl = OC.MimeType.getIconUrl(file.type)
 			}
 			// Create a unique index for each file
 			const date = new Date()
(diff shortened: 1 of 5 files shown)


src/store/fileUploadStore.js
```diff
@@ -239,13 +241,11 @@ const actions = {
...
 			// Get localurl for some image previews
-			let localUrl = ''
+			let localUrl
 			if (SHARED_ITEM.MEDIA_ALLOWED_PREVIEW.includes(file.type)) {
 				localUrl = URL.createObjectURL(file)
 			} else if (isVoiceMessage) {
-				localUrl = file.localUrl
-			} else {
-				localUrl = OC.MimeType.getIconUrl(file.type)
+				localUrl = file.localURL
```
DorraJaouad, line 248:
```suggestion
				localUrl = file.localUrl
```
?
image

Antreesy: Nope, that's `localURL` from File object, not localUrl from what we add to messageStore

DorraJaouad: You are not wrong

DorraJaouad (approved):
Tested 🦅

src/store/fileUploadStore.js
nickvergessen: Time to align? :speak_no_evil:

DorraJaouad (approved):

PR #13608, opened 2024-10-22
feat: replace audioRecorderStore with composable

Antreesy:
### ☑️ Resolves

* One more Vuex store away

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


src/composables/useAudioEncoder.ts
```diff
@@ -0,0 +1,35 @@
...
+		if (encoderReady.value) {
+			return
+		}
+		// Initialize an audio encoder
+		await register(await connect())
+		encoderReady.value = true
+	})
+
+	return readonly(encoderReady)
+}
```
ShGKme, line 29:
```suggestion
let requiresInit = true

/** Initialize the audio encoder */
async function initAudioEncoder() {
	requiresInit = false
	await register(await connect())
	encoderReady.value = true
}

/**
 * Composable to use audio encoder for voice messages feature
 * @return {DeepReadonly<Ref<boolean>>} - whether the encoder is ready
 */
function useAudioEncoder() {
	if (requiresInit) {
		initAudioEncoder()
	}

	return readonly(encoderReady)
}
```

src/composables/useAudioEncoder.ts
```diff
@@ -0,0 +1,35 @@
...
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import { createSharedComposable } from '@vueuse/core'
+import { register } from 'extendable-media-recorder'
+import { connect } from 'extendable-media-recorder-wav-encoder'
+import { onBeforeMount, readonly, ref } from 'vue'
+import type { Ref, DeepReadonly } from 'vue'
+
+const encoderReady = ref<boolean>(false)
```
ShGKme, line 12: Same as
```suggestion
const encoderReady = ref(false)
```

ShGKme (approved):

PR #13609, opened 2024-10-23
feat(email): Allow email guests without public link

nickvergessen:
### 🚧 Tasks

- [x] Based on #13499 
- [x] Name is "Private conversation" in invitation email
- [x] Joining doesn't work
- [ ] Check flow with calendar app / meeting events / upcoming integration

### 🏁 Checklist


nickvergessen:
Currently reloading as an email user seems to break in the frontend as the context request is done before joining the room finished:

image

I assume this might be from the recent changes to allow chatting even when signaling does not work or something. Posting a message (selected line) afterwards fixes the problem and shows all the chat messages.

Antreesy:
> Currently reloading as an email user seems to break in the frontend as the context request is done before joining the room finished:

This line in your PR now evaluates to `true`:
https://github.com/nextcloud/spreed/blob/e12e79d215ac9850baba716f62b92042e79d525f/src/composables/useGetMessages.ts#L82

Before for public conversations it was `attendeeId: 0`, and then we waited for joining before doing other request.
If attendeeId is known in Room object, why other requests are blocked?

miaulalala (approved):
LGTM!

PR #13611, opened 2024-10-23
fix: check for false positive hidden parent when scroll

Antreesy:
### ☑️ Resolves

* Follow-up to #12205 
* if chat is mounted, but hidden (in sidebar), offsetParent is also null

## 🖌️ UI Checklist

🏚️ Before | 🏡 After
-- | --
image | Clean console

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


DorraJaouad (approved):
🙈

PR #13625, opened 2024-10-24
fix: pass token to callView store via actions

Antreesy:
### ☑️ Resolves

* Fix broken public share sidebar 
* init => import Vuex => import Pinia => import Vuex - breaking loop that should be avoided (until everything is migrated)

⚠️ server master branch (31) has now different behaviour for public shares, so testable only on stable30 (see backport)

🏚️ Before | 🏡 After
-- | --
Errors in console | Working as expected

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


DorraJaouad (approved):
Tested, circular dependency was removed

Issue #8715, opened 2023-02-07
Inconsistency of the Initial State and Capabilities 


ShGKme:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---
Currently on the Talk frontend there is two sources of configs and features.
- Initial State - Server-rendered to `<input type="hidden" />` values, used by Talk Web only
- Capabilities API - used by all clients (frontend, mobile clients, 3rd part clients) for features (depends on the version) and configs

While features are mostly relevant for clients only, config is widely used by Talk Web as well as other clients.

Some configs are available only from one source and not from another one.

| Initial State (Server Render)  | Capabilities API                         | Is Used in Talk Web            | Notes                          |
|--------------------------------|------------------------------------------|--------------------------------|--------------------------------|
| `call_enabled`                 | `config.call.enabled`                    | 💻 Talk                        | 🗃️ Server config              |
| ❌                              | `config.call['breakout-rooms']`          | ❌                              | Feature in development         |
| ❌                              | `config.call.recording`                  | ❌                              | Feature in development         |
| `signaling_mode`               | ❌                                        | ⚙️ AdminSettings, Shares       | 🗃️ Server config              |
| `sip_dialin_info`              | ❌                                        | 💻 Talk (RightSidebar)         | 🗃️ Server config              |
| `grid_videos_limit`            | ❌                                        | 💻 Talk (CallView)             | 🗃️ **Web Only** Server config |
| `grid_videos_limit_enforced`   | ❌                                        | 💻 Talk (CallView)             | 🗃️ **Web Only** Server config |
| `federation_enabled`           | ❌                                        | 💻 Talk (search)               | 🗃️ Server config              |
| `default_permissions`          | ❌                                        | 💻 Talk (ConversationSettings) | 🗃️ Server config              |
| `start_conversations`          | `config.conversations['can-create']`     | 💻 Talk                        | 🔒 User permissions            |
| `circles_enabled`              | ❌                                        | 💻 Talk                        | 🔒 User permissions            |
| `guests_accounts_enabled`      | ❌                                        | 💻 Talk                        |                                |
| ❌                              | `config.chat['max-length']`              | ❌                              | Constant                       |
[...]

PR #13659, opened 2024-10-29
feat(capabilities): add sip-dialin-info capability

ShGKme:
### ☑️ Resolves

* A part of: https://github.com/nextcloud/spreed/issues/8715

### 🚧 Tasks

- [x] Add `sip-dialin-info` available only in the initial state to capabilities for clients
- [x] Use capabilities instead of initial state on frontend
- [x] Remove unneeded initial state

### 🏁 Checklist


nickvergessen:
I'm not sure why we went with initial state back then.
~~I think we should have this on a dedicated endpoint.~~

It actually is already. Could we then simply migrate the usages over? The `sipDialinInfo` is part of the signaling settings returned on `GET /api/{apiVersion}/signaling/settings`
https://github.com/nextcloud/spreed/blob/4cb26cc724758a02ec4c355d9684942aefbcfa5c/lib/ResponseDefinitions.php#L323

nickvergessen:
Closing as per above and raised https://github.com/nextcloud/spreed/issues/13904 instead

PR #13672, opened 2024-10-31
fix: move useDocumentTitle logic to composable

Antreesy:
### ☑️ Resolves

* Refactor document.title setting
  * extract logic from App.vue component
  * minimize computations, when not needed
  * set and remove asterisk only when needed
  * rely on Vue Router to get correct title parts

Better to check commits separately

🏚️ Before | 🏡 After

* Duplicate session should be now correctly set on specified route

TODO:
- [x] watch after `conversationList` instead, make `lastMessageMap` a function to be called only when window is not visible

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/composables/useDocumentTitle.ts
+++ b/src/composables/useDocumentTitle.ts
@@ -5,6 +5,7 @@
 
 import { computed, ref, watch } from 'vue'
 import type { ComputedRef } from 'vue'
+import type { Route } from 'vue-router'
 
 import { t } from '@nextcloud/l10n'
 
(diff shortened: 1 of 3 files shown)


src/composables/useDocumentTitle.ts
```diff
@@ -0,0 +1,162 @@
...
+		const newLastMessageMap = getLastMessageMap(newValue)
+		/**
+		 * @return {boolean} Returns true, if
+		 * - a conversation is newly added to lastMessageMap
+		 * - a conversation has a different last message id then previously
+		 */
+		const shouldShowAsterisk = Object.keys(newLastMessageMap).some(token => {
+			return !savedLastMessageMap.value[token] // Conversation is new
+				|| (savedLastMessageMap.value[token] !== newLastMessageMap[token] // Last message changed
+					&& newLastMessageMap[token] !== -1) // But is not from the current user
```
DorraJaouad, line 46: Should we add if it's also visible ?

Antreesy: If by visible you mean isDocumentVisible, I'd say no, as it wasn't the case before. Asterisk was shown on background tabs, to indicate some thing's changed

DorraJaouad: No, I mean the new message will be visible to the user.

src/composables/useDocumentTitle.ts
```diff
@@ -0,0 +1,162 @@
...
+		}
+	}
+
+	/**
+	 * Set the page title to the conversation name
+	 *
+	 * @param title Prefix for the page title e.g. conversation name
+	 * @param showAsterisk Whether to show unread indicator
+	 */
+	function setPageTitle(title: string, showAsterisk: boolean) {
```
DorraJaouad, line 136: You are using this in `setPageTitleFromRoute` only which has `showAsterisk`  as `false` in both calls so it doesn't make sense to put this arg. Did you want to use it  in `conversationList` watcher maybe?

DorraJaouad (commented):
Tested btw ✅

src/composables/useDocumentTitle.ts
Antreesy: Removed second argument, as asterisk adding/removing is handled by ref

DorraJaouad (approved):
Tested

DorraJaouad:
Note: When you have two tabs open, you have unread message (so both have asterisk), and you read the message in one of them, it would make sense that the asterisk is cleared in the other tab as well. (watch over last unread message?) ---> follow-up.

PR #13699, opened 2024-11-05
feat: invite email guests when creating a conversation

### LINKED ISSUE #4937: Invite emails directly when creating a conversation
### Expected behaviour
It's all in the title.
We have a lot of 1:1 conversations with external participants. For this purpose, the conversations are mostly created exclusively. It would be noticeable improvement if the creation dialog allows this form of usage.

### Actual behaviour
Currently, I have to go to the sidebar specifically to email invite a participant after the conversation has already been created.

</details>

## Talk app

v10.0.5

Antreesy:
### ☑️ Resolves

* Fix #4937
* As for now participants added after conversation is created and token is known, that works the same way as adding participant from the right sidebar
* ~~If at least one participant is e-mail, create a public conversation instead of rewriting it later on backend~~
* Also extract API call for 'core/autocomplete/get' to not mix with conversation services

🏚️ Before | 🏡 After
-- | --
image | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.spec.js
+++ b/src/components/LeftSidebar/LeftSidebar.spec.js
@@ -15,15 +15,18 @@ import { loadState } from '@nextcloud/initial-state'
 import LeftSidebar from './LeftSidebar.vue'
 
 import router from '../../__mocks__/router.js'
-import { searchPossibleConversations, searchListedConversations } from '../../services/conversationsService.js'
+import { searchListedConversations } from '../../services/conversationsService.js'
+import { autocompleteQuery } from '../../services/coreService.ts'
 import { EventBus } from '../../services/EventBus.ts'
 import storeConfig from '../../store/storeConfig.js'
 import { findNcListItems, findNcActionButton, findNcButton } from '../../test-helpers.js'
 import { requestTabLeadership } from '../../utils/requestTabLeadership.js'
 
 jest.mock('../../services/conversationsService', () => ({
-	searchPossibleConversations: jest.fn(),
 	searchListedConversations: jest.fn(),
+}))
+jest.mock('../../services/coreService', () => ({
+	autocompleteQuery: jest.fn(),
 }))
 
 // short-circuit debounce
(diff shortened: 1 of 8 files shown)


nickvergessen:
> If at least one participant is e-mail, create a public conversation instead of rewriting it later on backend

I would not do that, so once #13609 works we don't have clients that create "wrong" conversations by default

src/components/NewConversationDialog/NewConversationDialog.vue
```diff
@@ -293,9 +294,11 @@ export default {
 			this.page = 2
 
 			try {
+				const emailParticipants = this.selectedParticipants.filter(participant => participant?.source === ATTENDEE.ACTOR_TYPE.EMAILS)
+
 				this.newConversation.token = await this.$store.dispatch('createGroupConversation', {
 					conversationName: this.conversationName,
-					isPublic: this.isPublic,
+					isPublic: this.isPublic || emailParticipants.length,
```
nickvergessen, line 301: This should not be done. Let's delay the publishing until the API did it?

Antreesy: Also fine, would be future-proof in case we allow emails in private rooms

src/components/NewConversationDialog/NewConversationDialog.vue
```diff
@@ -328,6 +329,11 @@ export default {
 				}
 
 				await Promise.all(promises)
+
+				this.selectedParticipants.filter(participant => participant?.source === ATTENDEE.ACTOR_TYPE.EMAILS)
+					.forEach(participant => {
+						showSuccess(t('spreed', 'Invitation was sent to {actorId}', { actorId: participant.id }))
+					})
```
DorraJaouad, line 336: It's overwhelming the screen, maybe better without it or just one general toast message ("invitations were sent" n())?

DorraJaouad (approved):
Tested

src/components/NewConversationDialog/NewConversationDialog.vue
DorraJaouad: I still believe it's better to keep just one for all numbers...

Antreesy: Discarded to avoid confusions

PR #13488, opened 2024-10-07
feat(calls): Add appconfig and user preference for default setting of…

nickvergessen:
… media

### ☑️ Resolves

* Fix API side of #13446 

### ⚙️ Settings API

The endpoint for actually setting it is in the provisioning API:
if you have ocs_api_viewer installed: `/index.php/apps/ocs_api_viewer/view/provisioning_api#/operations/preferences-set-preference`
grafik

Sample:
```
POST /ocs/v2.php/apps/provisioning_api/api/v1/config/users/{appId}/{configKey}
{"configValue":"yes|no"}

POST /ocs/v2.php/apps/provisioning_api/api/v1/config/users/spreed/play_sounds
{"configValue":"no"}
```

### 🏁 Checklist

- [x] ⛑️ Tests (unit and/or integration) are included or not possible
- [x] 📘 API documentation in `docs/` has been updated or is not required
- [x] 🔖 Capability is added or not needed

PR #13700, opened 2024-11-05
fix: do not compare read/typing privacy with constants casted to string

Antreesy:
### ☑️ Resolves

* Fixup to #13488 
* Read / status privacy were always expected to be passed as integers

### 🚧 Tasks

- [x] Test with frontend
- [x] Run `composer openapi `

### 🏁 Checklist

### DIFF
--- a/tests/integration/features/bootstrap/FeatureContext.php
+++ b/tests/integration/features/bootstrap/FeatureContext.php
@@ -3643,7 +3643,7 @@ class FeatureContext implements Context, SnippetAcceptingContext {
 	}
 
 	/**
-	 * @Then /^user "([^"]*)" sets setting "([^"]*)" to "([^"]*)" with (\d+)(?: \((v1)\))?$/
+	 * @Then /^user "([^"]*)" sets setting "([^"]*)" to ("[^"]*"|\d) with (\d+)(?: \((v1)\))?$/
 	 *
 	 * @param string $user
 	 * @param string $setting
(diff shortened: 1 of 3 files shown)


lib/Settings/BeforePreferenceSetEventListener.php
```diff
@@ -67,7 +67,7 @@ public function validatePreference(string $userId, string $key, string $value):
 		// "privacy" 0/1
 		if ($key === UserPreference::TYPING_PRIVACY
 			|| $key === UserPreference::READ_STATUS_PRIVACY) {
-			$valid = $value === (string)Participant::PRIVACY_PRIVATE || $value === (string)Participant::PRIVACY_PUBLIC;
+			$valid = (int)$value === Participant::PRIVACY_PRIVATE || (int)$value === Participant::PRIVACY_PUBLIC;
```
nickvergessen, line 70: grafik

:shrug: we can not cast the value on validation if we afterwards continue with the provided value.

But yeah we should add a test instead :P

Antreesy: Maybe `(string)$value === (string)Participant::PRIVACY_PRIVATE` would be better here?

Currently it's breaking only main and 30.1, but I don't want to adjust frontend for that; otherwise Talk desktop won't work with previous versions. So if anything helps to keep previous payload valid, I'm up for it

nickvergessen (approved):

PR #13711, opened 2024-11-06
fix(OpenAPI): Fix array syntax ambiguities

provokateurin:
Fixes for https://github.com/nextcloud/openapi-extractor/pull/168.
No real issues found, Talk code quality is great as always :rocket:

### 🏁 Checklist

### DIFF
--- a/lib/Controller/BanController.php
+++ b/lib/Controller/BanController.php
@@ -77,7 +77,7 @@ class BanController extends AEnvironmentAwareController {
 	 *
 	 * Required capability: `ban-v1`
 	 *
-	 * @return DataResponse<Http::STATUS_OK, TalkBan[], array{}>
+	 * @return DataResponse<Http::STATUS_OK, list<TalkBan>, array{}>
 	 *
 	 * 200: List all bans
 	 */
@@ -95,7 +95,7 @@ class BanController extends AEnvironmentAwareController {
 	 * Required capability: `ban-v1`
 	 *
 	 * @param int $banId ID of the ban to be removed
-	 * @return DataResponse<Http::STATUS_OK, array<empty>, array{}>
+	 * @return DataResponse<Http::STATUS_OK, list<empty>, array{}>
 	 *
 	 * 200: Unban successfully or not found
 	 */
--- a/lib/Controller/BotController.php
+++ b/lib/Controller/BotController.php
@@ -120,7 +120,7 @@ class BotController extends AEnvironmentAwareController {
 	 * @param string $referenceId For the message to be able to later identify it again
 	 * @param int $replyTo Parent id which this message is a reply to
 	 * @param bool $silent If sent silent the chat message will not create any notifications
-	 * @return DataResponse<Http::STATUS_CREATED|Http::STATUS_BAD_REQUEST|Http::STATUS_UNAUTHORIZED|Http::STATUS_REQUEST_ENTITY_TOO_LARGE, array<empty>, array{}>
+	 * @return DataResponse<Http::STATUS_CREATED|Http::STATUS_BAD_REQUEST|Http::STATUS_UNAUTHORIZED|Http::STATUS_REQUEST_ENTITY_TOO_LARGE, list<empty>, array{}>
 	 *
 	 * 201: Message sent successfully
 	 * 400: When the replyTo is invalid or message is empty
@@ -183,7 +183,7 @@ class BotController extends AEnvironmentAwareController {
 	 * @param string $token Conversation token
 	 * @param int $messageId ID of the message
 	 * @param string $reaction Reaction to add
-	 * @return DataResponse<Http::STATUS_OK|Http::STATUS_CREATED|Http::STATUS_BAD_REQUEST|Http::STATUS_UNAUTHORIZED|Http::STATUS_NOT_FOUND, array<empty>, array{}>
+	 * @return DataResponse<Http::STATUS_OK|Http::STATUS_CREATED|Http::STATUS_BAD_REQUEST|Http::STATUS_UNAUTHORIZED|Http::STATUS_NOT_FOUND, list<empty>, array{}>
 	 *
 	 * 200: Reaction already exists
 	 * 201: Reacted successfully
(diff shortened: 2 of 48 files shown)


lib/Controller/BanController.php
```diff
@@ -95,7 +95,7 @@ public function listBans(): DataResponse {
 	 * Required capability: `ban-v1`
 	 *
 	 * @param int $banId ID of the ban to be removed
-	 * @return DataResponse<Http::STATUS_OK, array<empty>, array{}>
+	 * @return DataResponse<Http::STATUS_OK, list<empty>, array{}>
```
nickvergessen, line 98: Really not a fan of this... it will never be a list. But we might introduce an `'error'` key like we have on many other APIs, and then the response type changes from `[]` to `{}`
I guess it's hard to know which case is preferred technically.

lib/Controller/BotController.php
```diff
@@ -120,7 +120,7 @@ protected function getBotFromHeaders(string $token, string $message): Bot {
 	 * @param string $referenceId For the message to be able to later identify it again
 	 * @param int $replyTo Parent id which this message is a reply to
 	 * @param bool $silent If sent silent the chat message will not create any notifications
-	 * @return DataResponse<Http::STATUS_CREATED|Http::STATUS_BAD_REQUEST|Http::STATUS_UNAUTHORIZED|Http::STATUS_REQUEST_ENTITY_TOO_LARGE, array<empty>, array{}>
+	 * @return DataResponse<Http::STATUS_CREATED|Http::STATUS_BAD_REQUEST|Http::STATUS_UNAUTHORIZED|Http::STATUS_REQUEST_ENTITY_TOO_LARGE, list<empty>, array{}>
```
nickvergessen, line 123: same here :see_no_evil:

lib/Controller/BanController.php
nickvergessen: > Yes, but even then this is a breaking change.

well form `array<empty>` to `array{error? string}` it should be okay?
But anyway, we will see.

provokateurin: No, the former is `[]` in JSON and the later is `{}` or `{"error": ""}` in JSON and those are not compatible.

nickvergessen: > Same goes for the default value in the constructor of DataResponse which should be null as well.

It's not, but `mixed $data = []` and exactly why we use `[]`. In the old days null was not allowed, only object and array: https://github.com/nextcloud/server/commit/0696099bad56727d96c60f6221fe02dc7c71f511#diff-d2d337748c8364e5b754080842e1d575e95dc697b0c85d95e4767296f4a893f5R46
In fact null is only allowed since 28 (added by you): https://github.com/nextcloud/server/commit/b0001c6010840362355c181b831b4517ac19b8fc#diff-c3c2a82ea83585a8354ee880ffaa6fc4bb90090e95372c47e39e82717a065ce6L43

I guess we don't care for now and simply break the API for apps that try to sniff into always empty things

nickvergessen: ah lol, nvm. the `list<empty>` on root is not added to the OpenAPI definition as a list:
```json
                                                "meta": {
                                                    "$ref": "#/components/schemas/OCSMeta"
                                                },
                                                "data": {}
                                            }
                                        }
                                    }
```

So I guess we are fine to start using it later either way

nickvergessen: I think my favorite solution for now is:
```
diff --git a/lib/Controller/AEnvironmentAwareController.php b/lib/Controller/AEnvironmentAwareController.php
index 25cfa8083..238f0af5a 100644
--- a/lib/Controller/AEnvironmentAwareController.php
+++ b/lib/Controller/AEnvironmentAwareController.php
@@ -73,4 +73,16 @@ abstract class AEnvironmentAwareController extends OCSController {
 
                return $format;
        }
+
+       /**
+        * @psalm-suppress InvalidReturnStatement,InvalidReturnType
+        * @psalm-return array{}
+        */
+       public function getEmptyObjectData(): array {
+               if ($this->getResponseFormat() === 'json') {
+                       // Cheating here to make sure the array is always a JSON object on the API.
+                       return new \stdClass();
+               }
+               return [];
+       }
 }
diff --git a/lib/Controller/BanController.php b/lib/Controller/BanController.php
index a38b9fdbe..3b5d5b690 100644
--- a/lib/Controller/BanController.php
+++ b/lib/Controller/BanController.php
@@ -95,7 +95,7 @@ class BanController extends AEnvironmentAwareController {
         * Required capability: `ban-v1`
         *
         * @param int $banId ID of the ban to be removed
-        * @return DataResponse<Http::STATUS_OK, list<empty>, array{}>
+        * @return DataResponse<Http::STATUS_OK, array{}, array{}>
         *
         * 200: Unban successfully or not found
         */
@@ -103,6 +103,6 @@ class BanController extends AEnvironmentAwareController {
        #[RequireModeratorParticipant]
        public function unbanActor(int $banId): DataResponse {
                $this->banService->findAndDeleteBanByIdForRoom($banId, $this->room->getId());
-               return new DataResponse([], Http::STATUS_OK);
+               return new DataResponse($this->getEmptyObjectData(), Http::STATUS_OK);
        }
 }
diff --git a/openapi-full.json b/openapi-full.json
index 0fd153e89..206856fca 100644
--- a/openapi-full.json
+++ b/openapi-full.json
@@ -2433,7 +2433,9 @@
                                                 "meta": {
                                                     "$ref": "#/components/schemas/OCSMeta"
                                                 },
-                                                "data": {}
+                                                "data": {
+                                                    "type": "object"
+                                                }
                                             }
                                         }
                                     }
```

provokateurin: > Any disadvantage of using array{}

You can not switch to `list<>` later without a breaking change.

> I guess the real problem is that the API does not return like this, so it actually creates invalid code-docs combination?

Yeah this should not be done, I don't think psalm or openapi-extractor can catch it.

> I think my favorite solution for now is:

Looks ok to me, even though I think null is still better as it allows you to use any type later without a breaking change.

nickvergessen: > You can not switch to `list<>` later without a breaking change.

Well those things never should return lists directly anyway.

I made a poll for the team to give feedback if toggling to null and later on to `array{error?: string}` is problematic. If it turns out to be, we can also annotate `array{error?: string}` now already to indicate the correct way going forward.

nickvergessen:
- Replacing bit by bit in https://github.com/nextcloud/spreed/pull/13830

PR #13714, opened 2024-11-06
chore: add stable30.0 as target for Transifex backports

Antreesy:
### ☑️ Resolves

* Fix missing translations


.tx/backport
```diff
@@ -0,0 +1 @@
+stable30.0
```
nickvergessen, line 1: It's a replacement:
https://github.com/nextcloud/docker-ci/blob/master/translations-app/handleAppTranslations.sh#L35-L37

So would need to be
```suggestion
stable30.0 stable30 stable29 stable28
```

at the same time this can be easily forgot on the next major, and basically after tomorrow 30.0 is stopped?

PR #13736, opened 2024-11-08
fix: overwrite EventBus.off method to clear once events

Antreesy:
### ☑️ Resolves

* Fix stray once event listeners, if they were never called

## 🖌️ UI Checklist

- test with joining call:

Before | After
-- | --
image | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/services/EventBus.ts
+++ b/src/services/EventBus.ts
@@ -11,8 +11,12 @@ type GenericEventHandler = Handler<Events[keyof Events]> | WildcardHandler<Event
 type ExtendedEmitter = Emitter<Events> & {
 	once<Key extends keyof Events>(type: Key, handler: Handler<Events[Key]>): void
 	once(type: '*', handler: WildcardHandler<Events>): void
+	_onceHandlers: Map<keyof Events | '*', Map<GenericEventHandler, GenericEventHandler>>
 }
+
 export const EventBus: ExtendedEmitter = mitt() as ExtendedEmitter
+
+EventBus._onceHandlers = new Map()
 
 /**
  * Register a one-time event handler for the given type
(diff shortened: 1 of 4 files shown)


ShGKme (commented):
I'm not a fun of adding a new function to unsubscribe from a listener. Unlisten shouldn't be different between `on` and `once`. It has the same semantics and job. Same for other Event Emitters, including native `removeEventListener`.

Also, identifying a function by name is not reliable. It can be empty, it can be not unique.

Instead, let's store `once` handlers in a `Map<type, Map<originalHandler, onceHandlers>>` and override `off` method to support new feature.

src/services/EventBus.ts
```diff
@@ -11,8 +11,13 @@ type GenericEventHandler = Handler<Events[keyof Events]> | WildcardHandler<Event
 type ExtendedEmitter = Emitter<Events> & {
 	once<Key extends keyof Events>(type: Key, handler: Handler<Events[Key]>): void
 	once(type: '*', handler: WildcardHandler<Events>): void
+	cancelOnce<Key extends keyof Events>(type: Key, handler?: Handler<Events[Key]>): void
+	cancelOnce(type: '*', handler: WildcardHandler<Events>): void
```
ShGKme, line 15: Not actual anymore

src/services/EventBus.ts
```diff
@@ -11,8 +11,13 @@ type GenericEventHandler = Handler<Events[keyof Events]> | WildcardHandler<Event
...
+	_onceHandlersMap: Map<keyof Events | '*', Map<GenericEventHandler, GenericEventHandler>>
```
ShGKme, line 16: Nitpick: We don't need it inside the `EventBus` object as a public interface. It can be just a variable in the module scope.

ShGKme: To not duplicate type in a name. Like we don't add "Array" or "String" prefixes to other variables.

```suggestion
	_onceHandlers: Map<keyof Events | '*', Map<GenericEventHandler, GenericEventHandler>>
```

src/services/EventBus.ts
```diff
@@ -30,4 +35,47 @@ EventBus.once = function<Key extends keyof Events>(type: Key, handler: GenericEv
...
+				} else if (_handler === handler) {
+					// Event was not received yet, removing the fn and reference
+					handlers.splice(handlers.indexOf(fn) >>> 0, 1)
+					onceHandlers.delete(_handler)
+				}
+			}
+		} else {
+			EventBus._onceHandlersMap.set(type, new Map())
+		}
+	}
```
ShGKme, line 80: We don't need to completely re-implement it duplicating all the `mitt` logic. We can extend it, reusing the original method.

```suggestion
const off = EventBus.off.bind(EventBus)
EventBus.off = function<Key extends keyof Events>(type: Key, handler?: GenericEventHandler) {
  off(type, handler)
  const onceHandler = handler && onceHandlers.get(type)?.get(handler)
  if (onceHandler) {
    onceHandlers.get(type).remove(handler)
    off(type, onceHandler)
  }
}
```

Antreesy: It didn't pass these tests, so adjusted in fixups step by step:

src/services/EventBus.ts
```diff
@@ -30,4 +33,38 @@ EventBus.once = function<Key extends keyof Events>(type: Key, handler: GenericEv
...
+	if (onceHandlers) {
+		for (const [originalFn, fn] of onceHandlers) {
+			if (fn === handler) {
+				onceHandlers!.delete(originalFn)
+			} else if (originalFn === handler) {
+				// @ts-expect-error: Vue: No overload matches this call
+				off(type, fn)
+				onceHandlers!.delete(originalFn)
+			}
+		}
```
ShGKme, line 66: I don't understand this part. We have a specific handler to remove, why do we need to loop?

Antreesy: [fixup](https://github.com/nextcloud/spreed/pull/13769)

ShGKme (approved):

PR #13740, opened 2024-11-08
fix: reduce amount of EventBus listeners on messages

Antreesy:
### ☑️ Resolves

* Fix regression from #12550 
  * listeners weren't removed from unmounted components
* Follow-up to 7bc38ae53f0fbf306de5e784413f5fa906185e7d
  *  Toggle CSS class with native method rather than emitting event, which should be listened to all mounted components

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


DorraJaouad (approved):
Really good reduction

Issue #13562, opened 2024-10-16
Split permissions


nickvergessen:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

- Allow restricting to see the participant list https://github.com/nextcloud/spreed/issues/13455
- Split chat into
  - Messages
  - Reactions (also affecting call reactions?)
- Sharing files
- Downloading files (?)
- Using whiteboard

PR #13762, opened 2024-11-13
🛃 feat(permissions): Contract part for conversation-permissions-v2

nickvergessen:
### ☑️ Resolves

* Ref #13562 

### 🚧 Tasks

- [ ] Setup check when [`getAppValue('spreed', 'default_permissions')`](https://github.com/nextcloud/spreed/blob/63a30ed4c4061d68ac7de5027762f06744637e7f/lib/Config.php#L275) is set and "low" (maybe we go to another config?)
- [ ] Add migration for update
    - [ ] Handling room permissions
    - [ ] Handling attendee permissions
- [ ] Actually implement the permission changes
- [ ] Develop a plan for federation

### 🏁 Checklist

### DIFF
--- a/lib/Model/Attendee.php
+++ b/lib/Model/Attendee.php
@@ -87,6 +87,12 @@ class Attendee extends Entity {
 	public const PERMISSIONS_PUBLISH_VIDEO = 32;
 	public const PERMISSIONS_PUBLISH_SCREEN = 64;
 	public const PERMISSIONS_CHAT = 128;
+	public const PERMISSIONS_CHAT_REACTION = 256;
+	public const PERMISSIONS_FILE_VIEW = 512;
+	public const PERMISSIONS_FILE_SHARE = 1024;
+	public const PERMISSIONS_WHITEBOARD = 2048;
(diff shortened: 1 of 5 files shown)


lib/Model/Attendee.php
```diff
@@ -87,6 +87,12 @@ class Attendee extends Entity {
 	public const PERMISSIONS_PUBLISH_VIDEO = 32;
 	public const PERMISSIONS_PUBLISH_SCREEN = 64;
 	public const PERMISSIONS_CHAT = 128;
+	public const PERMISSIONS_CHAT_REACTION = 256;
+	public const PERMISSIONS_FILE_VIEW = 512;
+	public const PERMISSIONS_FILE_SHARE = 1024;
```
nickvergessen, line 92: Should we do another for "create poll", or a generic one for "share (any thing but file) items"

nickvergessen: That's basically the second option. Since polls are highly more interactive I wondered if splitting makes sense, but yeah otherwise a generic object share also sounds like an option

Issue #13450, opened 2024-10-02
Add option to force passwords for public conversations


nickvergessen:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

Since we return the room object we can do this already, but the moderator has no way to see the password atm

miaulalala:
- [ ] Config option to determine if server only allows password protected talk conversations
- [ ] Change "create Room" endpoint (RoomController) to have a password option
- [ ] Password policy checking on password not empty or config option set

If config option is given
- [ ] Reject room creation if global password option is set an no password is given or it doesn't match password policy
- [ ] RoomService -> setPassword as inspiration for how to store and validate a password

PR #13767, opened 2024-11-13
feat(conversations): add option to force passwords

miaulalala:
### ☑️ Resolves
in public conversations
* Fix https://github.com/nextcloud/spreed/issues/13450

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist


docs/settings.md
```diff
@@ -111,3 +111,4 @@ Legend:
...
 | `enable_matterbridge`                | string<br>`1` or `0`                                             | `0`        | No   | 🖌️    | Whether the Matterbridge integration is enabled and can be configured                                                                                                                                                                                                            |
+| `force_public_chat_passwords`        | string<br>`1` or `0`                                             | `0`        | No   | 🖌️    | Whether public chats are forced to use a password                                                                                                                                                                                                                                |
```
nickvergessen, line 114:
```suggestion
| `force_public_chat_passwords`        | string<br>`1` or `0`                                             | `0`        | No   |      | Whether public chats are forced to use a password                                                                                                                                                                                                                                |
```
We won't add a UI option I think

lib/Controller/RoomController.php
```diff
@@ -683,7 +692,9 @@ protected function createEmptyRoom(string $roomName, bool $public = true, string
 
 		// Create the room
 		try {
-			$room = $this->roomService->createConversation($roomType, $roomName, $currentUser, $objectType, $objectId);
+			$room = $this->roomService->createConversation($roomType, $roomName, $currentUser, $objectType, $objectId, $password);
+		} catch (PasswordException $e) {
+			return new DataResponse(['reason' => $e->getReason(), 'hint' => $e->getHint()], Http::STATUS_FORBIDDEN);
```
nickvergessen, line 697: I wouldn't go with a new status code
```suggestion
			return new DataResponse(['error' => 'password', 'hint' => $e->getHint()], Http::STATUS_BAD_REQUEST);
```

or is the reason to be displayed additionally? I think it was the english string of the hint.

lib/Service/RoomService.php
```diff
@@ -167,7 +169,20 @@ public function createConversation(int $type, string $name, ?IUser $owner = null
 			throw new InvalidArgumentException('object');
 		}
 
-		$room = $this->manager->createRoom($type, $name, $objectType, $objectId);
+		if ($type !== Room::TYPE_PUBLIC || !$this->config->isPasswordEnforced()) {
+			$room = $this->manager->createRoom($type, $name, $objectType, $objectId);
+		} elseif ($password === '') {
+			throw new PasswordException(PasswordException::REASON_VALUE, 'Password needs to be set');
```
nickvergessen, line 175: hint should be translated?

lib/Controller/RoomController.php
```diff
@@ -642,10 +651,10 @@ protected function createCircleRoom(string $targetCircleId): DataResponse {
 	}
 
 	/**
-	 * @return DataResponse<Http::STATUS_CREATED, TalkRoom, array{}>|DataResponse<Http::STATUS_BAD_REQUEST, array{error?: string}, array{}>|DataResponse<Http::STATUS_NOT_FOUND, array<empty>, array{}>
+	 * @return DataResponse<Http::STATUS_CREATED, TalkRoom, array{}>|DataResponse<Http::STATUS_BAD_REQUEST, array{error?: string}, array{}>|DataResponse<Http::STATUS_FORBIDDEN, array{hint: string, reason: 'breakout-room'|'type'|'value'}, array{}>|DataResponse<Http::STATUS_NOT_FOUND, array<empty>, array{}>
```
nickvergessen, line 654: And then the response is:
```suggestion
	 * @return DataResponse<Http::STATUS_CREATED, TalkRoom, array{}>|DataResponse<Http::STATUS_BAD_REQUEST, array{error?: string, hint?: string}, array{}>|DataResponse<Http::STATUS_NOT_FOUND, array<empty>, array{}>
```

lib/Service/RoomService.php
```diff
@@ -167,7 +171,20 @@ public function createConversation(int $type, string $name, ?IUser $owner = null
...
+		} else {
+			$event = new ValidatePasswordPolicyEvent($password);
+			try {
+				$this->dispatcher->dispatchTyped($event);
+			} catch (HintException $e) {
+				throw new PasswordException(PasswordException::REASON_VALUE, $e->getHint());
+			}
+
+			$room = $this->manager->createRoom($type, $name, $objectType, $objectId, $password);
+		}
```
DorraJaouad, line 187: > This code must be created by the system for each room.

Does this mean it should be generated automatically in backend and exposed to user after room creation? So user does not need to add password when creating the room.

lib/Manager.php
```diff
@@ -1084,7 +1084,7 @@ public function getChangelogRoom(string $userId): Room {
 	 * @param string $objectId
 	 * @return Room
 	 */
-	public function createRoom(int $type, string $name = '', string $objectType = '', string $objectId = ''): Room {
+	public function createRoom(int $type, string $name = '', string $objectType = '', string $objectId = '', string $password = ''): Room {
```
nickvergessen, line 1087: password is not set during the insert.

Also it needs to be hashed before saving:
https://github.com/nextcloud/spreed/blob/c7e07968fe56e1f0bebaef3bb58db68c2eb1d42f/lib/Service/RoomService.php#L744

lib/Manager.php
```diff
@@ -1111,6 +1111,7 @@ public function createRoom(int $type, string $name = '', string $objectType = ''
 			'token' => $token,
 			'object_type' => $objectType,
 			'object_id' => $objectId,
+			'password' => $password
```
nickvergessen, line 1114: this should be the hash aswell

lib/Service/RoomService.php
nickvergessen: No, the frontend must send it. But the code is all there already (apart from checking the enforcement in the web UI), it just sends a follow up request at the moment, and in the future it needs to be part of the creation request (if the capability is there and on)

tests/php/Service/RoomServiceTest.php
```diff
@@ -46,6 +47,7 @@ class RoomServiceTest extends TestCase {
 	protected IEventDispatcher&MockObject $dispatcher;
 	protected IJobList&MockObject $jobList;
 	protected LoggerInterface&MockObject $logger;
+	private IL10N&MockObject $l10n;
```
nickvergessen, line 50: Personal preference, but I avoid private all the time to allow extending

nickvergessen:
We also need 2 capabilities:
1. A feature flag, so client can know when they can send the password in the createRoom request (and it needs to be documented in https://github.com/nextcloud/spreed/blob/f8fef7dc7933db84be1b7f032aede85747bf29e3/docs/conversation.md#L130 )
2. A config flag, if the password requirement is enforced

lib/Manager.php
miaulalala: IMHO the responsibility of hashing should be with the insert.

docs/conversation.md
```diff
@@ -115,30 +115,31 @@
...
         + `200 OK` When the "one to one" conversation already exists
         + `201 Created` When the conversation was created
         + `400 Bad Request` When an invalid conversation type was given
         + `400 Bad Request` When the conversation name is empty for `type = 3`
+        + `400 Bad Request` When a password is required for a public room or when the password is invalid according to the password policy
         + `401 Unauthorized` When the user is not logged in
         + `404 Not Found` When the target to invite does not exist
 
-    - Data: See array definition in `Get user´s conversations`
+        - Data: See array definition in `Get user´s conversations`
```
nickvergessen, line 142:
```suggestion
     - Data: See array definition in `Get user´s conversations`
```

lib/Service/RoomService.php
```diff
@@ -543,6 +560,45 @@ public function setType(Room $room, int $newType, bool $allowSwitchingOneToOne =
 		$this->dispatcher->dispatchTyped($event);
 	}
 
+	/**
+	 * @param Room $room
+	 * @param int $newType Currently it is only allowed to change between `Room::TYPE_GROUP` and `Room::TYPE_PUBLIC`
+	 * @param bool $allowSwitchingOneToOne Allows additionally to change the type from `Room::TYPE_ONE_TO_ONE` to `Room::TYPE_ONE_TO_ONE_FORMER`
+	 * @throws TypeException
```
nickvergessen, line 567:
```suggestion
	 * @throws TypeException
	 * @throws PasswordException
```

lib/Service/RoomService.php
```diff
@@ -543,6 +560,45 @@ public function setType(Room $room, int $newType, bool $allowSwitchingOneToOne =
...
+			return;
+		}
+
+		if ($room->getType() !== Room::TYPE_GROUP) {
+			throw new TypeException(TypeException::REASON_TYPE);
+		}
+
+		if ($password === '') {
+			throw new PasswordException(PasswordException::REASON_VALUE, $this->l10n->t('Password needs to be set'));
+		}
```
nickvergessen, line 580: Only when the config is enforced?

lib/Service/RoomService.php
```diff
@@ -543,6 +560,45 @@ public function setType(Room $room, int $newType, bool $allowSwitchingOneToOne =
...
+
+		$event = new ValidatePasswordPolicyEvent($password);
+		try {
+			$this->dispatcher->dispatchTyped($event);
+		} catch (HintException $e) {
+			throw new PasswordException(PasswordException::REASON_VALUE, $e->getHint());
+		}
```
nickvergessen, line 587: Only when the password is not empty.

lib/Service/RoomService.php
```diff
@@ -543,6 +560,45 @@ public function setType(Room $room, int $newType, bool $allowSwitchingOneToOne =
...
+
+		$event = new BeforeRoomModifiedEvent($room, ARoomModifiedEvent::PROPERTY_TYPE, Room::TYPE_PUBLIC, $room->getType());
+		$this->dispatcher->dispatchTyped($event);
+		$event = new BeforeRoomModifiedEvent($room, ARoomModifiedEvent::PROPERTY_PASSWORD, $password);
+		$this->dispatcher->dispatchTyped($event);
```
nickvergessen, line 592: Only when the password is not empty.

lib/Service/RoomService.php
```diff
@@ -543,6 +560,45 @@ public function setType(Room $room, int $newType, bool $allowSwitchingOneToOne =
...
+		$passwordHash = $this->hasher->hash($password);
+		$this->manager->setPublic($room->getId(), $passwordHash);
+
+		$event = new RoomModifiedEvent($room, ARoomModifiedEvent::PROPERTY_TYPE, Room::TYPE_PUBLIC, $room->getType());
+		$this->dispatcher->dispatchTyped($event);
+		$event = new RoomModifiedEvent($room, ARoomModifiedEvent::PROPERTY_PASSWORD, $password);
+		$this->dispatcher->dispatchTyped($event);
```
nickvergessen, line 599: Only when the password is not empty.

lib/Service/RoomService.php
miaulalala: see line https://github.com/nextcloud/spreed/blob/d2bd3e03ad5819b8809339e6d5c4db95ec6a1fd7/lib/Controller/RoomController.php#L1427C2-L1435C3

lib/Controller/RoomController.php
```diff
@@ -1417,21 +1427,33 @@ public function removeAttendeeFromRoom(int $attendeeId): DataResponse {
...
-	 * @return DataResponse<Http::STATUS_OK, array<empty>, array{}>|DataResponse<Http::STATUS_BAD_REQUEST, array{error: 'breakout-room'|'type'|'value'}, array{}>
+	 * @return DataResponse<Http::STATUS_OK, array<empty>, array{}>|DataResponse<Http::STATUS_BAD_REQUEST, array{error: 'breakout-room'|'type'|'value', hint ?: string}, array{}>
 	 *
 	 * 200: Allowed guests successfully
 	 * 400: Allowing guests is not possible
 	 */
 	#[NoAdminRequired]
 	#[RequireLoggedInModeratorParticipant]
-	public function makePublic(): DataResponse {
+	public function makePublic(string $password = ''): DataResponse {
```
nickvergessen, line 1437: parameter should be documented and the method should document that it has a "Require capability"
https://github.com/nextcloud/spreed/blob/de5200a83698ccdbb034fc7f3c50c1fdb5b0cd5b/lib/Controller/BanController.php#L38

lib/Service/RoomService.php
miaulalala: Password check for empty password is also in the method

lib/Service/RoomService.php
```diff
@@ -562,11 +562,10 @@ public function setType(Room $room, int $newType, bool $allowSwitchingOneToOne =
 
 	/**
 	 * @param Room $room
-	 * @param int $newType Currently it is only allowed to change between `Room::TYPE_GROUP` and `Room::TYPE_PUBLIC`
-	 * @param bool $allowSwitchingOneToOne Allows additionally to change the type from `Room::TYPE_ONE_TO_ONE` to `Room::TYPE_ONE_TO_ONE_FORMER`
-	 * @throws TypeException
+	 * @param string $password
```
nickvergessen, line 565:
```suggestion
```

same for Room

lib/Controller/RoomController.php
miaulalala: See https://github.com/nextcloud/spreed/pull/13767/files#diff-d583983624ef8769460d4aa59e64b75ee54b6955e01e620e3eacedb7ba208786R1430

nickvergessen:
Can you rebase, so CI runs?

lib/Controller/RoomController.php
```diff
@@ -1417,21 +1427,36 @@ public function removeAttendeeFromRoom(int $attendeeId): DataResponse {
 	/**
 	 * Allowed guests to join conversation
 	 *
-	 * @return DataResponse<Http::STATUS_OK, array<empty>, array{}>|DataResponse<Http::STATUS_BAD_REQUEST, array{error: 'breakout-room'|'type'|'value'}, array{}>
+	 * Required capability: `conversation-creation-password`
+	 *
+	 * @param string $password New password
```
nickvergessen, line 1432:
```suggestion
	 * @param string $password New password (only available with `conversation-creation-password` capability)
```

lib/Controller/RoomController.php
```diff
@@ -1417,21 +1427,36 @@ public function removeAttendeeFromRoom(int $attendeeId): DataResponse {
...
 	 * 200: Allowed guests successfully
 	 * 400: Allowing guests is not possible
 	 */
 	#[NoAdminRequired]
 	#[RequireLoggedInModeratorParticipant]
-	public function makePublic(): DataResponse {
+	public function makePublic(string $password = ''): DataResponse {
+		if (!$this->talkConfig->isPasswordEnforced()) {
+			try {
+				$this->roomService->setType($this->room, Room::TYPE_PUBLIC);
```
nickvergessen, line 1443: when not empty we should still set the password with a `$this->roomService->setPassword(` call?

lib/Service/RoomService.php
```diff
@@ -132,12 +134,14 @@ public function createOneToOneConversation(IUser $actor, IUser $targetUser): Roo
 	 * @param IUser|null $owner
 	 * @param string $objectType
 	 * @param string $objectId
+	 * @param string $password
 	 * @return Room
```
nickvergessen, line 138: Delete all empty parameters? Type hints are replacements

nickvergessen:
Almost there, just a small bug and some documentation fixes

lib/Controller/RoomController.php
```diff
@@ -121,6 +121,7 @@ public function __construct(
 		protected Capabilities $capabilities,
 		protected FederationManager $federationManager,
 		protected BanService $banService,
+		protected IL10n $l10n,
```
nickvergessen, line 124: missing import

docs/conversation.md
```diff
@@ -115,7 +115,6 @@
 ## Creating a new conversation
 
 *Note:* Creating a conversation as a child breakout room, will automatically set the lobby when breakout rooms are not started and will always overwrite the room type with the parent room type. Also, moderators of the parent conversation will be automatically added as moderators.
-
```
nickvergessen, line 118: this needs to be readded (otherwise the docs list breaks), try to run `mkdocs serve` locally
grafik

docs/conversation.md
nickvergessen, line 1: $password parameter should also be added to `Allow guests in a conversation (public conversation)`

lib/Controller/RoomController.php
```diff
@@ -645,10 +654,10 @@ protected function createCircleRoom(string $targetCircleId): DataResponse {
 	}
 
 	/**
-	 * @return DataResponse<Http::STATUS_CREATED, TalkRoom, array{}>|DataResponse<Http::STATUS_BAD_REQUEST, array{error?: string}, array{}>|DataResponse<Http::STATUS_NOT_FOUND, array<empty>, array{}>
+	 * @return DataResponse<Http::STATUS_CREATED, TalkRoom, array{}>|DataResponse<Http::STATUS_BAD_REQUEST, array{error?: string, hint?: string}, array{}>|DataResponse<Http::STATUS_NOT_FOUND, array<empty>, array{}>
```
nickvergessen, line 657: Should we use `message` to align with the other cases where we add the human readable message?
```suggestion
	 * @return DataResponse<Http::STATUS_CREATED, TalkRoom, array{}>|DataResponse<Http::STATUS_BAD_REQUEST, array{error?: string, message?: string}, array{}>|DataResponse<Http::STATUS_NOT_FOUND, array<empty>, array{}>
```
Same then in the actual code I guess

lib/Controller/RoomController.php
```diff
@@ -1420,21 +1431,35 @@ public function removeAttendeeFromRoom(int $attendeeId): DataResponse {
...
+				$this->roomService->setType($this->room, Room::TYPE_PUBLIC);
+			}
+		} catch (PasswordException $e) {
+			return new DataResponse(['error' => 'password', 'hint' => $e->getHint()], Http::STATUS_BAD_REQUEST);
 		} catch (TypeException $e) {
 			return new DataResponse(['error' => $e->getReason()], Http::STATUS_BAD_REQUEST);
 		}
 
 		return new DataResponse();
+
```
nickvergessen, line 1462:
```suggestion
```
*nitpick*

nickvergessen:
Need a command to recompile openapi :P

nickvergessen (approved):

PR #13769, opened 2024-11-14
fixup! fix: overwrite EventBus.off method to clear once events

ShGKme:


Antreesy:
Implemented in original PR, thanks!

PR #13784, opened 2024-11-15
fix: extract createTemporaryMessage to utils

Antreesy:
### ☑️ Resolves

* pure function, not connected to messagesStore
* default messageType is now set to 'comment'
* fix: pass mimetype of voice message file

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


DorraJaouad:
This is better if it is a composable actually. That's because `actorId`, `actorType`, `actorDisplayName`, `parent` are always available from store, so the method would need only these as arguments (variables):
```
	text,
	token,
	uploadId,
	index,
	file,
	localUrl,
	isVoiceMessage,
```

DorraJaouad (approved):
Tested

PR #13785, opened 2024-11-15
fix(calls): Add email to exported participant list and list everyone …

nickvergessen:
…that had joined the call

### ☑️ Resolves

* Follow-up to #6098 and #13453 

### 🚧 Tasks

- [x] Export the email addresses of email guests now that they are no longer the user id
- [x] Export all users that had joined the call

### 🏁 Checklist


miaulalala (approved):
LGTM

PR #13795, opened 2024-11-16
fix: use useDarkTheme from nextcloud/vue library

Antreesy:
### ☑️ Resolves

* functionality was moved to lib in https://github.com/nextcloud-libraries/nextcloud-vue/releases/tag/v8.20.0 , we don't need a copy in Talk anymore

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


DorraJaouad (approved):
All good, thank you for taking care of it.

PR #13807, opened 2024-11-18
feat(AI-call-summary): Automatically summarize call transcript

### LINKED ISSUE #13429: AI generated call summaries
Add ability to generate call summaries

nickvergessen:
### ☑️ Resolves

* Fix #13429

### 🚧 Tasks

- [ ] ~Add capability or not needed for admin interface~ @Antreesy 
- [x] Integration tests depend on https://github.com/nextcloud/server/pull/49357

### 👣 Steps

1. Set appconfig:
```
occ config:app:set spreed call_recording_summary --value='yes'
```

2. Have a recording backend
3. Follow "Testing locally" steps from https://github.com/nextcloud/spreed/pull/13661
4. Start a call with recording
5. End the call
6. Run cron jobs multiple times
7. Check notifications

### 🏁 Checklist


lib/Recording/Listener.php
```diff
@@ -54,40 +59,50 @@ public function handle(Event $event): void {
...
 			return;
 		}
 
-		if ($owner === null) {
+		if ($task->getUserId() === null) {
 			return;
 		}
 
-		$this->recordingService->notifyAboutFailedTranscript($owner, $fileNode);
+		if ($isTranscription) {
```
miaulalala, line 93: return early on ` !$isTranscription && !$isSummary`

lib/Recording/Listener.php
```diff
@@ -54,40 +59,50 @@ public function handle(Event $event): void {
...
+			} elseif ($event instanceof TaskFailedEvent) {
+				$this->recordingService->notifyAboutFailedTranscript($task->getUserId(), $roomToken, $fileId, 'transcript');
+			}
+		} elseif ($isSummary) {
+			if ($event instanceof TaskSuccessfulEvent) {
+				$this->recordingService->storeTranscript($task->getUserId(), $roomToken, $fileId, $task->getOutput()['output'] ?? '', 'summary');
+			} elseif ($event instanceof TaskFailedEvent) {
+				$this->recordingService->notifyAboutFailedTranscript($task->getUserId(), $roomToken, $fileId, 'summary');
+			}
+		}
```
miaulalala, line 105: deduplicate code here and check for the type `'transcript'` or `'summary'` only

lib/Recording/Listener.php
```diff
@@ -34,12 +34,17 @@ public function __construct(
...
 	) {
 	}
 
 	public function handle(Event $event): void {
-		if ($event instanceof AbstractTranscriptionEvent) {
-			$this->handleTranscriptionEvents($event);
+		if ($event instanceof AbstractTaskProcessingEvent) {
+			try {
+				$this->handleTranscriptionEvents($event);
+			} catch (\Throwable $e) {
```
miaulalala, line 45: Would it make sense to give this process it's own exception?

lib/Service/RecordingService.php
```diff
@@ -139,42 +142,120 @@ public function store(Room $room, string $owner, array $file): void {
...
 			$participant = $this->participantService->getParticipant($room, $owner);
 		} catch (ParticipantNotFoundException) {
-			$this->logger->warning('Could not determinate conversation when trying to store transcription of call recording');
+			$this->logger->warning("Could not determinate conversation when trying to store $aiTask of call recording");
 			throw new InvalidArgumentException('owner_participant');
 		}
 
-		$transcriptFileName = pathinfo($recording->getName(), PATHINFO_FILENAME) . '.md';
+		$shouldTranscribe = $this->serverConfig->getAppValue('spreed', 'call_recording_transcription', 'no') === 'yes';
+		$shouldSummarize = $this->serverConfig->getAppValue('spreed', 'call_recording_summary', 'no') === 'yes';
```
miaulalala, line 206: check this first so you can return early if both features are turned off

miaulalala (changes requested):

lib/Recording/Listener.php
nickvergessen: It's not intentional exceptions, I just noticed that a throws exception (due to type error) here killed/removed the re-scheduled TaskProcessing background job.

lib/Recording/Listener.php
nickvergessen: it's there?

```php
		} else {
			return;
		}
```
In line 80?

lib/Recording/Listener.php
nickvergessen: :see_no_evil:

lib/Service/RecordingService.php
nickvergessen: Well either is true, otherwise we wouldn't be in storeTranscript (which could be renamed to storeAITextOutput now), because we only end up here, if a task got resolved after `store()` checked that either is true already.

miaulalala (approved):
:shipit:

PR #13813, opened 2024-11-19
fix: delegate message posting from NewMessage to parent components

Antreesy:
### ☑️ Resolves

  - remove breakoutRooms logic from reused SendMessageDialog

No visual changes

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/NewMessage/NewMessage.vue
+++ b/src/components/NewMessage/NewMessage.vue
@@ -217,7 +217,6 @@ import BrowserStorage from '../../services/BrowserStorage.js'
 import { getTalkConfig, hasTalkFeature } from '../../services/CapabilitiesManager.ts'
 import { EventBus } from '../../services/EventBus.ts'
 import { shareFile } from '../../services/filesSharingServices.js'
-import { useBreakoutRoomsStore } from '../../stores/breakoutRooms.ts'
 import { useChatExtrasStore } from '../../stores/chatExtras.js'
 import { useSettingsStore } from '../../stores/settings.js'
 import { fetchClipboardContent } from '../../utils/clipboard.js'
@@ -269,6 +268,15 @@ export default {
 		},
 
 		/**
+		 * If component is used in a dialog, submit should emit an event to parent
+		 * instead of posting the message
+		 */
+		dialog: {
+			type: Boolean,
+			default: false,
+		},
+
+		/**
 		 * Broadcast messages to all breakout rooms of a given conversation.
 		 */
 		broadcast: {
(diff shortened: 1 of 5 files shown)


src/components/NewMessage/NewMessageUploadEditor.vue
```diff
@@ -70,8 +70,7 @@
 				:container="modalContainerId"
 				:aria-label="t('spreed', 'Post message')"
 				@upload="handleUpload"
-				@sent="handleDismiss"
-				@failure="handleDismiss" />
+				@dismiss="handleDismiss" />
```
DorraJaouad, line 73: Following the idea "delegate posting a message to its parent when component is in dialog", this should have the dialog prop truthy and send the message from here (receive submit event) :p

Antreesy: Added in fixup

src/components/NewMessage/NewMessage.vue
```diff
@@ -699,9 +706,13 @@ export default {
...
-				this.broadcast
-					? await this.broadcastMessage(this.token, temporaryMessage.message)
-					: await this.postMessage(this.token, temporaryMessage, options)
+				// Delegate sending to HOC, unless it's upload dialog
+				// If upload reaches there, then it is a normal message without any shares
+				if (this.dialog && !this.upload) {
+					await this.submitMessage(this.token, temporaryMessage, options)
+				} else {
+					await this.postMessage(this.token, temporaryMessage, options)
+				}
```
DorraJaouad, line 715: It should still delegate sending the message to the parent when it's upload (also dialog) 
[The case where you have text in upload editor but removed all files]

src/components/NewMessage/NewMessage.vue
```diff
@@ -682,21 +689,7 @@ export default {
...
-					this.$emit('upload', { caption: this.text, options })
-					return
-				} else {
-					// Dismiss dialog, process as normal message sending otherwise
-					this.$emit('sent')
-				}
-			}
-
-			if (this.hasText) {
+			if (this.hasText || (this.dialog && this.upload)) {
```
DorraJaouad, line 692: `dialog` is sufficient because each `upload` is `dialog` in our approach (and this assumption is already used in removing non-dialog components of NewMessage like out of office)

DorraJaouad (approved):
Tested, just a semantic related comment. Otherwise, all good 🦅

src/components/NewMessage/NewMessage.vue
Antreesy: I'd keep it as-is, since we have dialogs - non-uploads. That change would allow broadcast empty strings or post empty messages to BR, which we don't want to have

PR #13830, opened 2024-11-20
fix(openapi): Improve empty array cases

nickvergessen:
### ☑️ Resolves

* Ref #13792

### 🏁 Checklist

### DIFF
--- a/lib/Capabilities.php
+++ b/lib/Capabilities.php
@@ -272,18 +272,18 @@ class Capabilities implements IPublicCapability {
 			$capabilities['config']['signaling']['hello-v2-token-key'] = $pubKey;
 		}
 
-		/** @var ?string[] $predefinedBackgrounds */
+		/** @var ?list<string> $predefinedBackgrounds */
 		$predefinedBackgrounds = null;
 		$cachedPredefinedBackgrounds = $this->talkCache->get('predefined_backgrounds');
 		if ($cachedPredefinedBackgrounds !== null) {
 			// Try using cached value
-			/** @var string[]|null $predefinedBackgrounds */
+			/** @var list<string>|null $predefinedBackgrounds */
 			$predefinedBackgrounds = json_decode($cachedPredefinedBackgrounds, true);
 		}
 
 		if (!is_array($predefinedBackgrounds)) {
 			// Cache was empty or invalid, regenerate
-			/** @var string[] $predefinedBackgrounds */
+			/** @var list<string> $predefinedBackgrounds */
 			$predefinedBackgrounds = [];
 			if (file_exists(__DIR__ . '/../img/backgrounds')) {
 				$directoryIterator = new \DirectoryIterator(__DIR__ . '/../img/backgrounds');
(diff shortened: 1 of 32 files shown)


lib/Controller/TempAvatarController.php
```diff
@@ -111,7 +111,7 @@ public function postAvatar(): DataResponse {
 	/**
 	 * Delete your avatar as a user
 	 *
-	 * @return DataResponse<Http::STATUS_OK|Http::STATUS_BAD_REQUEST, array<empty>, array{}>
+	 * @return DataResponse<Http::STATUS_OK, null, array{}>|DataResponse<Http::STATUS_BAD_REQUEST, array{error: 'avatar'}, array{}>
```
provokateurin, line 114: I think for a response you should avoid literal values.
If a another value can be returned later the current spec is not compatible later.

lib/Capabilities.php
```diff
@@ -184,7 +184,7 @@ public function __construct(
 	/**
 	 * @return array{
 	 *      spreed: TalkCapabilities,
-	 * }|array<empty>
+	 * }|array<string, empty>
```
provokateurin, line 187: This is not parsed well by openapi-extractor, I think the better approach is to use `spreed?: TalkCapabilities`.

provokateurin (commented):
I wonder if openapi-extractor should skip the response type when the only type is `null` just like it already does with `array<empty>`. Both essentially mean "there is nothing to look at" and it would make the spec a bit cleaner (and remove almost all spec changes in this PR).

miaulalala (approved):

lib/Capabilities.php
nickvergessen: but `array{spreed?: …}` can be an empty array, which is then a list and later on fails, but I can't do the stdClass cheat there as the capabilities interface only allows arrays to be returned.
But yeah I noticed this mistake already as well and am thinking of a solution.

lib/Controller/TempAvatarController.php
nickvergessen: well if we make it string instead of literals, it can not be the documentation anymore, as we need a second list of known error codes, and if we put that into the status code line it means psalm can not help us anymore to ensure the list is up to date.

lib/Capabilities.php
provokateurin: It doesn't matter though, as the empty case is handled by the server. The annotation is also valid for `return [];` so it is fine.

lib/Controller/TempAvatarController.php
provokateurin: Ok makes sense, I didn't know you wanted to use it this way 👍

lib/Capabilities.php
nickvergessen: It's handled by the server but not by the room/v4/.../capabilities endpoint. But yeah I will cheat then

provokateurin (approved):

nickvergessen:
/backport ad5f9a56cb29a6b1533f6d324b2a6692102f9f45,15410d5e447b969be926e50ccefdbc5f786c5c7e,ff438bc23bb75156fc6c3a7101ea35e392c16a24 to stable30

nickvergessen:
/backport ad5f9a56cb29a6b1533f6d324b2a6692102f9f45 15410d5e447b969be926e50ccefdbc5f786c5c7e ff438bc23bb75156fc6c3a7101ea35e392c16a24 to stable30

Issue #13451, opened 2024-10-02
Allow to import list of participants


nickvergessen:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

E.g. CSV list of `email,displayname` to add many (at another place exported) participants easily

Antreesy:
1) We do not currently have an API option for bulk adding of participants to reuse here (and during conversation creation).
2) Javascript should be able to generate CSV file for import/export, but if it's preferred to done on backend, also fine by me

### Frontend:
#### Design note:

Participants setup dialog from conversation creation can be reused (to be called from Right Sidebar):

![Image](https://github.com/user-attachments/assets/6e7cc600-4a08-409f-ac86-289d087beaf1)

- Import option in three-dot-menu next to the searchbar (combined with dialpad button or separately?)
- Export option in the dialog footer (or also in three-dot menu?)

nickvergessen:
The export is most likely from another product, not from another chat/conversation in first place.
So not sure this makes sense to combine UI wise. Export needs to happen during/after a call/meeting.

nickvergessen:
Done for now

PR #13871, opened 2024-11-26
feat(invitations): Allow importing CSV email lists

nickvergessen:
### ☑️ Resolves

* API for #13451 

### 🚧 Tasks

- [x] Fix psalm
- [x] Fix OpenAPI
- [x] Document capability
- [ ] Write integration test

### 🏁 Checklist

### DIFF
--- a/lib/Controller/RoomController.php
+++ b/lib/Controller/RoomController.php
@@ -14,6 +14,7 @@ use OCA\Talk\Events\AAttendeeRemovedEvent;
 use OCA\Talk\Events\BeforeRoomsFetchEvent;
 use OCA\Talk\Exceptions\CannotReachRemoteException;
 use OCA\Talk\Exceptions\ForbiddenException;
+use OCA\Talk\Exceptions\GuestImportException;
 use OCA\Talk\Exceptions\InvalidPasswordException;
 use OCA\Talk\Exceptions\ParticipantNotFoundException;
 use OCA\Talk\Exceptions\RoomNotFoundException;
@@ -73,6 +74,7 @@ use OCP\Federation\ICloudIdManager;
 use OCP\IConfig;
 use OCP\IGroup;
 use OCP\IGroupManager;
+use OCP\IL10N;
 use OCP\IPhoneNumberUtil;
 use OCP\IRequest;
 use OCP\IUser;
@@ -121,6 +123,7 @@ class RoomController extends AEnvironmentAwareController {
 		protected Capabilities $capabilities,
 		protected FederationManager $federationManager,
 		protected BanService $banService,
+		protected IL10N $l,
 	) {
 		parent::__construct($appName, $request);
 	}
@@ -1204,7 +1207,7 @@ class RoomController extends AEnvironmentAwareController {
 			} catch (TypeException) {
 			}
 
-			$email = $newParticipant;
+			$email = strtolower($newParticipant);
 			$actorId = hash('sha256', $email);
 			try {
 				$this->participantService->getParticipantByActor($this->room, Attendee::ACTOR_EMAILS, $actorId);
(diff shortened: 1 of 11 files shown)


lib/Controller/RoomController.php
```diff
@@ -2424,6 +2427,59 @@ public function setMessageExpiration(int $seconds): DataResponse {
...
+	 * Import a list of email attendees (comma separated values, one per line, either `"email","name"` or only `"email"`)
+	 *
+	 * @return DataResponse<Http::STATUS_OK, array{invites: non-negative-int, duplicates: non-negative-int, type?: int<-1, 6>}, array{}>|DataResponse<Http::STATUS_BAD_REQUEST, array{error: 'room'|'file'|'header-email'|'header-name'|'rows', message?: string, invites?: non-negative-int, duplicates?: non-negative-int, type: int<-1, 6>}, array{}>
+	 *
+	 * 200: All entries imported successfully
+	 * 400: Import was not successful
+	 */
+	#[NoAdminRequired]
+	#[RequireModeratorParticipant]
+	public function importEmails(bool $testRun = true): DataResponse {
```
miaulalala, line 2440:
```suggestion
	public function importParticipantsByEmail(bool $testRun = true): DataResponse {
```

lib/Exceptions/GuestImportException.php
```diff
@@ -0,0 +1,68 @@
...
+	 */
+	public function getDuplicates(): ?int {
+		return $this->duplicates;
+	}
+
+	/**
+	 * @return non-negative-int|null
+	 */
+	public function getInvalid(): ?int {
+		return $this->invalidLines === null ? null : count($this->invalidLines);
```
miaulalala, line 62:
```suggestion
		return count($this->invalidLines) ?? null;
```

should work

lib/GuestManager.php
```diff
@@ -69,6 +76,125 @@ public function updateName(Room $room, Participant $participant, string $display
...
+			if (count($details) === 2) {
+				[$email, $name] = $details;
+			} else {
+				$email = $details[0];
+				$name = null;
+			}
+
+			$email = strtolower($email);
+			if (!$this->validateMailAddress($email)) {
+				$this->logger->debug('Invalid email address on line: ' . $line);
```
miaulalala, line 139:
```suggestion
				$this->logger->debug('Invalid email ' . $email .  ' on line: ' . $line);
```

lib/GuestManager.php
```diff
@@ -69,6 +76,125 @@ public function updateName(Room $room, Participant $participant, string $display
...
+				$invalidLines[] = $line;
+				continue;
+			}
+
+			if ($name !== null && strtolower($name) === $email) {
```
miaulalala, line 144:
```suggestion
			if ($name !== null && strcasecmp($name, $email) === 0) {
```

https://stackoverflow.com/questions/9905580/which-one-is-faster-better-strcasecmp-or-equal-to-operator

miaulalala (commented):
some nitpicks

lib/GuestManager.php
nickvergessen: TIL

lib/GuestManager.php
```diff
@@ -69,6 +76,125 @@ public function updateName(Room $room, Participant $participant, string $display
...
+			if (empty($invalidLines)) {
+				return [
+					'invites' => count($emailsToAdd),
+					'duplicates' => $duplicates,
+				];
+			}
+
+			throw new GuestImportException(
+				GuestImportException::REASON_ROWS,
+				$this->l->t('Following lines where invalid: %s', implode(', ', $invalidLines)),
```
Antreesy, line 167:
```suggestion
				$this->l->t('Following lines were invalid: %s', implode(', ', $invalidLines)),
```
or `are invalid`

lib/GuestManager.php
```diff
@@ -69,6 +76,125 @@ public function updateName(Room $room, Participant $participant, string $display
...
+				GuestImportException::REASON_ROWS,
+				$this->l->t('Following lines are invalid: %s', implode(', ', $invalidLines)),
+				$invalidLines,
+				count($emailsToAdd),
+				$duplicates,
+			);
+		}
+
+		foreach ($emailsToAdd as $add) {
+			$participant = $this->participantService->inviteEmailAddress($room, $add['actorId'], $add['email'], $add['name']);
```
Antreesy, line 175: I'm not receiving either e-mail invites in mailhog / new entries in DB / participants in the list, but should happened here?

nickvergessen: Default of the testRun parameter was true, but you don't send the parameter when it's false it seems. I can change the default.

nickvergessen: - It's actually a bug in server because you DO send the value as false: https://github.com/nextcloud/server/pull/49515
We can work around it by changing the default to false and you not sending the value in that case.

nickvergessen: Changed here in the API PR and in the frontend PR and it's working now.

lib/GuestManager.php
```diff
@@ -69,6 +76,125 @@ public function updateName(Room $room, Participant $participant, string $display
...
+			}
+
+			if (count($details) > 2) {
+				$this->logger->debug('Invalid entry with too many fields on line: ' . $line);
+				$invalidLines[] = $line;
+				continue;
+			}
+
+			if (count($details) === 2) {
+				[$email, $name] = $details;
```
miaulalala, line 131:
```suggestion
				[$email, $name] = $details;
```

Would it be wise to add a `trim()` to these values?

miaulalala (approved):
LGTM! i'd suggest trimming the incoming names / emails

Issue #9016, opened 2023-03-10
Message jumps when MessageButtonBar is shown at the bottom


SystemKeeper:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Steps to reproduce
1. Open a conversation
2. Hover over the last message

### Expected behaviour
MessageButtonsBar should be shown, without moving the message

### Actual behaviour
The message is moved up a bit and the MessageButtonsBar is not fully visible

https://user-images.githubusercontent.com/1580193/224322708-2fc78678-551a-45d6-ba91-19551c1a3f8b.mov

Probably related to #9003 ?

## Talk app

**Talk app version:** On c.nc.c

## Browser

**Operating system:** MacOS Ventura 13.2

**Browser name:** Chrome

**Browser version:** 110

SystemKeeper:
> at least we won't revert the other PR.

Please no. 😅

Yes, not sure if the current one can be tweaked a bit, otherwise this is way better than the last one.

ShGKme:
With two one-line messages sometimes it makes an infinity loop:

https://user-images.githubusercontent.com/25978914/224427177-0b9c1055-9ccf-43c4-a3bc-21179be8b746.mp4

P.S. the video has lower FPS that the message actually jumps

nickvergessen:
Btw this is Chrome only.
Firefox "underflows" the Input:
![grafik](https://user-images.githubusercontent.com/213943/224428729-c7d499cd-48e3-4d6b-b1e6-e0bde9d7623a.png)

PR #13888, opened 2024-11-27
Move the message hover position

Simounet:

🏚️ Before | 🏡 After
-- | --
message-hover-position-before | message-hover-position-after

### 🏁 Checklist

- [X] 🌏 Tested with different browsers / clients:
  - [X] Chromium (Chrome / Edge / Opera / Brave)
  - [X] Firefox
  - [ ] Safari
  - [ ] Talk Desktop
  - [X] Not risky to browser differences / client

### Purpose

Being able to get the info from the check icon and send date even when hovering the message. Until now, if it's my first message, I can't be sure if it's the sent or read status.

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/Message.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/Message.vue
@@ -475,7 +475,7 @@ export default {
 
 	&__scroll {
 		position: absolute;
-		top: 0;
+		top: 20px;
 		right: 0;
 		width: fit-content;
 		height: 100%;


nickvergessen:
As easy as it sounds, we had it in the past and reverted it, as it creates flickering and issues when you hover the last message and it's single line. Buttons are cut of, or scrolling misbehaves, etc. :see_no_evil: 
grafik

Ref https://github.com/nextcloud/spreed/issues/9016 and many other things.
So feels quite dangerous.

Simounet:
I understand you being extra cautious about this sort of "little" things. I saw the video with the flickering but I can't see why it would produce this kind of issue with an absolute position. Did you have other options in mind? To me, actions should not be displayed on hover, hidden from the user.
Should we use the extra space on the right side of the message on desktop view and absolute position be always display on mobile?

DorraJaouad:
> I can't see why it would produce this kind of issue with an absolute position.

Probably because the absolute positioning is relative to its ancestor only (message item). 

> To me, actions should not be displayed on hover, hidden from the user.

Indeed, it makes it more accessible to keep it shown for user regardless of hover especially for mobile view yet we have the space constraint. As shown below, this is a view from desktop browser and there is no space that fits the message actions :/
<image src="https://github.com/user-attachments/assets/ec17b64f-0fe8-4eb4-b29b-e1f86ba2da31" width="500" />

> Should we use the extra space on the right side of the message on desktop view

The side spaces are not fixed and are adjustable based on window width.

DorraJaouad:
@Simounet Thank you for taking the time to contribute to our project. I am afraid we're unable to merge this pull request as it would remove a critical workaround we implemented to address the existing flickering issue.

We appreciate your effort and want to provide more context. The current workaround is essential for maintaining the stability of our user experience. If you'd like to understand the specifics of the flickering problem or any other issue, we'd be happy to discuss it in more detail in [our public chat](https://cloud.nextcloud.com/call/c7fz9qpr).

We would like to invite you to explore our open issues on Talk. Your insights and code contribution are always welcome.

PR #13891, opened 2024-11-28
fix(openapi): Remove lastMessage from TalkRoom object when empty

nickvergessen:
### ☑️ Resolves

* Last step after #13837 before #13792

### 🏁 Checklist

### DIFF
--- a/src/composables/useConversationInfo.js
+++ b/src/composables/useConversationInfo.js
@@ -43,7 +43,7 @@ export function useConversationInfo({
 	})
 
 	const hasLastMessage = computed(() => {
-		return !!Object.keys(Object(item.value?.lastMessage)).length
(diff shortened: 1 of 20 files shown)


src/composables/useConversationInfo.js
```diff
@@ -43,7 +43,7 @@ export function useConversationInfo({
 	})
 
 	const hasLastMessage = computed(() => {
-		return !!Object.keys(Object(item.value?.lastMessage)).length
```
nickvergessen, line 46: @DorraJaouad don't we need the old check for desktop client compatibility with old servers?

DorraJaouad: True

DorraJaouad (approved):

PR #13892, opened 2024-11-28
fix: respect user changes of devices state when toggling it

DorraJaouad:

🏚️ Before 
If you have default device state to be off and you toggle either audio or video on before joining the call, your changes won't be considered

 🏡 After
If you have default device state to be off and you toggle either audio or video on before joining the call, your changes will be considered

### 🏁 Checklist

  - [x] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/CallView/CallView.vue
+++ b/src/components/CallView/CallView.vue
@@ -197,15 +197,15 @@ export default {
 		},
 	},
 
-	setup() {
+	setup(props) {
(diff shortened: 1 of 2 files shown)


nickvergessen (approved):
In the future we should store enabled and disabled, so that we could know if the user enabled it already in a conversation at some point and not always disable it. So that it is a real "default", not a "force"

PR #13930, opened 2024-12-02
feat(call): Direct endpoint to check if call notification should be dismissed

nickvergessen:
## 🛠️ API Checklist

Just as I was about to send this PR, I remembered that we built https://github.com/nextcloud/notifications/blob/master/openapi.json#L742-L748
That should be sufficient here as well.

### 🏁 Checklist

### DIFF
--- a/docs/capabilities.md
+++ b/docs/capabilities.md
@@ -170,3 +170,4 @@
 ## 21
 * `config => conversations => force-passwords` - Whether passwords are enforced for public rooms
 * `conversation-creation-password` - Whether the endpoints for creating public conversations or making a conversation public support setting a password
+* `call-notification-state-api` (local) - Whether the endpoints exists for checking if a call notification should be dismissed
--- a/lib/Controller/CallNotificationController.php
+++ b/lib/Controller/CallNotificationController.php
@@ -1,0 +1,65 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Talk\Controller;
+
+use OCA\Talk\Service\ParticipantService;
+use OCP\AppFramework\Http;
+use OCP\AppFramework\Http\Attribute\NoAdminRequired;
+use OCP\AppFramework\Http\Attribute\OpenAPI;
+use OCP\AppFramework\Http\DataResponse;
+use OCP\AppFramework\OCSController;
+use OCP\IRequest;
+
+class CallNotificationController extends OCSController {
+	public const CASE_STILL_CURRENT = 0;
+	public const CASE_ROOM_NOT_FOUND = 1;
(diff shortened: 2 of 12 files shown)


docs/capabilities.md
```diff
@@ -170,3 +170,4 @@
 ## 21
 * `config => conversations => force-passwords` - Whether passwords are enforced for public rooms
 * `conversation-creation-password` - Whether the endpoints for creating public conversations or making a conversation public support setting a password
+* `call-notification-state-api` - Whether the endpoints exists to checking if a call notification should be dismissed
```
miaulalala, line 173:
```suggestion
* `call-notification-state-api` - Whether the endpoints exists for checking if a call notification should be dismissed
```

miaulalala (changes requested):

lib/Service/ParticipantService.php
```diff
@@ -1601,6 +1602,60 @@ public function getParticipantsInCall(Room $room, int $maxAge = 0): array {
...
+					$query->expr()->eq('s.attendee_id', 'a.id'),
+					$query->expr()->neq('s.in_call', $query->createNamedParameter(Participant::FLAG_DISCONNECTED, IQueryBuilder::PARAM_INT), IQueryBuilder::PARAM_INT),
+				)
+			)
+			->where($query->expr()->eq('r.token', $query->createNamedParameter($token)))
+			->andWhere($query->expr()->eq('a.actor_type', $query->createNamedParameter(Attendee::ACTOR_USERS)))
+			->andWhere($query->expr()->eq('a.actor_id', $query->createNamedParameter($userId)));
+
+		$result = $query->executeQuery();
+		$row = $result->fetch();
```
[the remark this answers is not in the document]
miaulalala: :+1: just wondering why tho - since you're not looping them with a foreach?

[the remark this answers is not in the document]
nickvergessen: fetchOne() selects 1 column, the loop is a row :P

miaulalala (approved):

lib/Controller/CallNotificationController.php
```diff
@@ -0,0 +1,65 @@
...
+	 */
+	#[NoAdminRequired]
+	#[OpenAPI(tags: ['call'])]
+	public function state(string $token): DataResponse {
+		if ($this->userId === null) {
+			return new DataResponse(null, Http::STATUS_FORBIDDEN);
+		}
+
+		$status = match($this->participantService->checkIfUserIsMissingCall($token, $this->userId)) {
+			self::CASE_PARTICIPANT_JOINED,
```
Antreesy, line 57: I'd like to not handle this as an error (case if participant joined same call from this or another client). Can we return `202 Accepted` or `204 No Content`?

nickvergessen: As mentioned in the office, I went for an error, as basically 2xx means: keep a notification (of either kind, call or missed), and 4xx means: remove the notification

nickvergessen:
Rebasing to run federation integration tests

PR #13933, opened 2024-12-03
fix(dependencies): Always install newest roave/security-advisories on CI

nickvergessen:
Remove the package from locked composer files,
which defeated the purpose of the whole package:
checking live for the newest version and if something is a problem.


blizzz (approved):
:four_leaf_clover:

PR #13943, opened 2024-12-05
fix(federation) Implement signing of federation requests in Talk

nickvergessen:
### ☑️ Resolves

* Ref https://github.com/nextcloud/server/pull/49646

### 🏁 Checklist

### DIFF
--- a/lib/Model/InvitationMapper.php
+++ b/lib/Model/InvitationMapper.php
@@ -46,6 +46,25 @@ class InvitationMapper extends QBMapper {
 
 	/**
 	 * @throws DoesNotExistException
+	 * @internal Does not check user relation
+	 */
+	public function getByRemoteServerOnlyWithAccessToken(
+		string $remoteServerUrl,
+		#[SensitiveParameter]
+		string $accessToken,
+	): Invitation {
+		$qb = $this->db->getQueryBuilder();
+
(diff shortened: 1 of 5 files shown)


tests/psalm-baseline.xml
```diff
@@ -26,6 +29,17 @@
       <code><![CDATA[$this->request->server]]></code>
     </NoInterfaceProperties>
   </file>
+  <file src="lib/Federation/CloudFederationProviderTalk.php">
+    <UndefinedClass>
+      <code><![CDATA[ISignedCloudFederationProvider]]></code>
```
nickvergessen, line 34: Should be not needed anymore after https://github.com/nextcloud-deps/ocp/pull/28 is there

lib/Model/InvitationMapper.php
```diff
@@ -44,6 +44,25 @@ public function getInvitationById(int $id): Invitation {
 		return $this->findEntity($qb);
 	}
 
+	/**
+	 * @throws DoesNotExistException
+	 * @internal Does not check user relation
+	 */
+	public function getByRemoteServerOnlyWithAccessToken(
```
miaulalala, line 51:
```suggestion
	public function getByRemoteServerByAccessToken(
```

PR #13944, opened 2024-12-05
fix(conversations): fix password enforcement logic error

miaulalala:
### ☑️ Resolves

Passwords should be settable on public rooms no matter if the `force_passwords` options is set. Only check if a password is set if the option is set to `true` and the room is public, but still let public rooms set a password independent from the option.

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist


lib/Service/RoomService.php
```diff
@@ -165,7 +165,11 @@ public function createConversation(int $type, string $name, ?IUser $owner = null
 			throw new InvalidArgumentException('object');
 		}
 
-		if ($type !== Room::TYPE_PUBLIC || !$this->config->isPasswordEnforced()) {
+		if ($type === Room::TYPE_PUBLIC && $password === '' && $this->config->isPasswordEnforced() === true) {
+			throw new InvalidArgumentException('password');
```
nickvergessen, line 169:
```suggestion
			throw new PasswordException(PasswordException::REASON_VALUE, $this->l10n->t('Password needs to be set'));
```

DorraJaouad (approved):

lib/Service/RoomService.php
```diff
@@ -165,10 +165,12 @@ public function createConversation(int $type, string $name, ?IUser $owner = null
 			throw new InvalidArgumentException('object');
 		}
 
-		if ($type !== Room::TYPE_PUBLIC || !$this->config->isPasswordEnforced()) {
-			$room = $this->manager->createRoom($type, $name, $objectType, $objectId);
-		} elseif ($password === '') {
+		if ($type === Room::TYPE_PUBLIC && $password === '' && $this->config->isPasswordEnforced() === true) {
```
nickvergessen, line 168:
```suggestion
		if ($type === Room::TYPE_PUBLIC && $password === '' && $this->config->isPasswordEnforced()) {
```

nickvergessen (approved):
Could add an integration test?

tests/integration/features/chat-3/public.feature
```diff
@@ -111,3 +111,9 @@ Feature: chat-2/public
     And user "participant1" renames room "public room" to "A name with 260 chars 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 12345678" with 400 (v4)
     # 255 chars
     And user "participant1" renames room "public room" to "Another name with 255 chars 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 123456789 1234567" with 200 (v4)
+
+  Scenario: Create public room with a password
+    When user "participant1" creates room "public room" with 201 (v4)
```
nickvergessen, line 116: also add one where you set the config to "required" and try to create one without sending a password and expect 400?

PR #13951, opened 2024-12-06
fix(chat): Add a repair-step to handle the last-read-message=0 case

nickvergessen:
### ☑️ Resolves

**Repair step doc block**
- In previous versions of Talk, the value 0 was used to mark a chat as
completely unread. This meant that the next time a client/browser loaded the
chat it would start all from the beginning. However, there were various
issues over the time that made the frontend getting `null` or `undefined`
into the wrong place and then accidentally loading 8 years of chat from the
beginning. So it was agreed that the frontend manually blocks loading with
last-read-message=0 and the special cases use -2 for this situation.

### 🏁 Checklist

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -81,6 +81,7 @@
 		<post-migration>
 			<step>OCA\Talk\Migration\ClearResourceAccessCache</step>
 			<step>OCA\Talk\Migration\CacheUserDisplayNames</step>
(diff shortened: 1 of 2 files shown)


Antreesy (approved):
LGTM

DorraJaouad (approved):

miaulalala (approved):

Issue #6292, opened 2021-09-27
Ability to quickly schedule a meeting


jancborchardt:
**Is your feature request related to a problem? Please describe.**
Sometimes you just want to schedule a meeting instead of starting it immediately.

**Describe the solution you'd like**
Microsoft Teams does it like this:
![MS Teams - Schedule a meeting](https://user-images.githubusercontent.com/925062/134863048-3fbdb1b9-95bd-424b-8ce6-cbb69aa07f16.png)

This could open a modal which lets you fill in Calendar details, with relevant things prefilled like location with the link to the Talk room, and participants probably picked from the conversation.

nickvergessen:
The "Meet" button is basically our call button. So this is about existing conversations and it eases the user as they don't have to open two tabs and go back and forth until they invited all participants in the room and copied the talk room url, etc.

ChristophWurst:
Let me share my thoughts on this.

What we need on a technical level is some code that pushes a new event into the user's calendar. The scheduling mechanism will take care of sending an invite to all participants. The event should likely also link back to this chat room, as we do in the existing took room integration for calendar events.

Now, the questions is: who creates this event? In Mail we recently had a somewhat similar feature done where a user can create a new event from an incoming message. We thought about re-using some components from Calendar to show the actual *new event* modal from Calendar, but this turned out to be a lot harder than possible. So now Mail just has its own event modal, constructs the event with the help of some shared calendar libs and directly pushes the event to the Nextcloud user's calendar of choice.

We can possibly do the same here. Add a small modal with the input fields we need, like title of the event and date pickers. The participants (email addresses¹) can be acquired from Talk. Then we send the event to the user's selected calendar. Done. The Calendar app doesn't have to be loaded on the Talk page, nor are there any dependencies from Calendar->Talk nor the other way around.

¹dav scheduling only works for *users with an email address*. Like right now you can't create meetings in Calendar before you set your email address in the Nextcloud account settings. That's because sabre needs to map users to dav resources and back. We might have to limit this feature so it only shows **if all** participants have their email set. Otherwise you will schedule a meeting to only parts of the group. And from a user's perspective this could be really confusing when part of the expected attendees won't show up just because they never got an invite.

nickvergessen:
[...]

PR #13964, opened 2024-12-09
feat(meetings): Add endpoint to get writable calendars and the defaul…

nickvergessen:
…t calendar

### ☑️ Resolves

* Part 1 of #6292 

### 🚧 Tasks

- [ ] Requires https://github.com/nextcloud/server/pull/49731/
- [ ] Capability `schedule-meeting` will be added once all APIs are implemented

### 🏁 Checklist

### DIFF
--- a/lib/Controller/SettingsController.php
+++ b/lib/Controller/SettingsController.php
@@ -125,7 +125,7 @@ class SettingsController extends OCSController {
 				continue;
 			}
 
-			if (!($calendar->getPermissions('principals/users/' . $this->userId) & Constants::PERMISSION_CREATE)) {
+			if (!($calendar->getPermissions() & Constants::PERMISSION_CREATE)) {
 				continue;
 			}
 
--- a/lib/Controller/SettingsController.php
+++ b/lib/Controller/SettingsController.php
@@ -8,29 +8,38 @@ declare(strict_types=1);
 
 namespace OCA\Talk\Controller;
 
+use OCA\Talk\ResponseDefinitions;
 use OCA\Talk\Settings\BeforePreferenceSetEventListener;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\Attribute\NoAdminRequired;
 use OCP\AppFramework\Http\Attribute\OpenAPI;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\OCSController;
+use OCP\Calendar\IManager;
+use OCP\Constants;
 use OCP\Files\IRootFolder;
 use OCP\IConfig;
+use OCP\IDBConnection;
 use OCP\IGroup;
 use OCP\IGroupManager;
 use OCP\IRequest;
 use Psr\Log\LoggerInterface;
 
+/**
+ * @psalm-import-type TalkCalendar from ResponseDefinitions
+ */
 class SettingsController extends OCSController {
 
 	public function __construct(
 		string $appName,
 		IRequest $request,
 		protected IRootFolder $rootFolder,
+		protected IDBConnection $db,
 		protected IConfig $config,
 		protected IGroupManager $groupManager,
 		protected LoggerInterface $logger,
 		protected BeforePreferenceSetEventListener $preferenceListener,
+		protected IManager $calendarManager,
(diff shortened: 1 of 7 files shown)


Antreesy:
Noticed that default calendar has a color 'null'. Is it also the case for you, and is it on our side?

lib/Controller/SettingsController.php
```diff
@@ -86,4 +95,81 @@ public function setSIPSettings(
...
+			->where($query->expr()->eq('userid', $query->createNamedParameter($userId)))
+			->andWhere($query->expr()->eq('propertypath', $query->createNamedParameter($propertyPath)))
+			->andWhere($query->expr()->eq('propertyname', $query->createNamedParameter($propertyName)))
+			->setMaxResults(1);
+
+		$result = $query->executeQuery();
+		$property = $result->fetchOne();
+		$result->closeCursor();
+
+		return $property;
```
miaulalala, line 173: You should go through the DAV app to get the default calendar URI, or you can use the user config:


`$uri = $this->appConfig->getUserValue($userId, 'dav', 'defaultCalendar', CalDavBackend::PERSONAL_CALENDAR_URI);`

miaulalala (changes requested):

lib/Controller/SettingsController.php
nickvergessen: right, should use that preference,
but going through the dav app is to risky as it's all private api

nickvergessen: So I toggled this:
grafik

But there is no value in oc_preferences

nickvergessen: I also only find delete and get for that preference, no set anywhere:
grafik

miaulalala: @st3iny any ideas?

st3iny: Yeah, the setting can only be set by admins and is currently not documented. It acts as an override but will not be configured at all in most cases (see https://github.com/nextcloud/server/pull/19852).

The official way would be to go through the CalDAV API and extract the default calendar from the principal's node.

st3iny: Exemplary code:

```php
        $result = $this->server->getProperties(
            $principalUri,
            [
                '{DAV:}principal-URL',
                 $caldavNS.'calendar-home-set',
                 $caldavNS.'schedule-inbox-URL',
                 $caldavNS.'schedule-default-calendar-URL',
                '{http://sabredav.org/ns}email-address',
            ]
        );
        $calendarPath = $result['{urn:ietf:params:xml:ns:caldav}schedule-default-calendar-URL']->getHref();
```

This requires access to an instance of `\Sabre\DAV\Server`.

st3iny: Alternative: Do a `PROPFIND` request to the CalDAV backend.

```
curl -u admin:admin --data @propfind.xml -X PROPFIND -H 'Content-Type: application/xml' -H 'Depth: 0' https://localhost/remote.php/dav/principals/users/admin
```

### Request body (`propfind.xml`)

```xml
<d:propfind xmlns:d="DAV:" xmlns:cd="urn:ietf:params:xml:ns:caldav">
  <d:prop>
     <cd:schedule-default-calendar-URL />
  </d:prop>
</d:propfind>
```

nickvergessen: ### Options

1. Settings up a DAV server in OCS
2. Doing a DAV request from client side
3. Stick with the hacky query and add the preference check before that

---

### My POV

1. In my opinion 1. is not possible. We would have to make sure that so many things are chained together in the right order, all apps are loaded, etc. That's not suitable.
2. Doable/best option from API side, but requires clients to send a DAV request, but I think they all have something already for file uploads, but not sure those libs also allow caldav
3. Basically there already and could be finished in no time

Since we are on feature freeze today, I will check with the assigned frontender if they can do it quickly with 2, otherwise we go with 3 for now.

nickvergessen:
Frontend is now talking CalDAV to get the calendar list and default calendar, so we will not add this.

nickvergessen:
Still need the POST endpoint

PR #13973, opened 2024-12-10
fix(AdminSettings): test Websocket connection

Antreesy:
### ☑️ Resolves

Allows to test websocket connection with HPB in Admin settings (before users will face it in Talk interface): client attempts to establish new websocket, then:
  - clean close on success
  - throw error if any

Small screen | Wide screen
-- | --
image | image
image | image

### 🚧 Tasks

- [x] Handle NC <-> HPB servers time sync issue
- [ ] Handle other cases that could be checked?

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/AdminSettings/SignalingServer.vue
+++ b/src/components/AdminSettings/SignalingServer.vue
@@ -36,17 +36,30 @@
 			<IconAlertCircleOutline v-else-if="warningMessage" :size="20" fill-color="var(--color-warning)" />
 			<IconCheck v-else :size="20" fill-color="var(--color-success)" />
 			{{ connectionState }}
+
+			<NcButton v-if="server && checked"
+				type="tertiary"
+				:title="t('spreed', 'Test this server')"
+				:aria-label="t('spreed', 'Test this server')"
+				@click="checkServerVersion">
+				<template #icon>
+					<IconReload :size="20" />
+				</template>
+			</NcButton>
 		</span>
 
-		<NcButton v-if="server && checked"
-			type="tertiary"
-			:title="t('spreed', 'Test this server')"
-			:aria-label="t('spreed', 'Test this server')"
-			@click="checkServerVersion">
-			<template #icon>
-				<IconReload :size="20" />
-			</template>
-		</NcButton>
+		<ul v-if="signalingTestInfo.length" class="test-connection-data">
+			<li v-for="(row, idx) in signalingTestInfo"
+				:key="idx"
+				class="test-connection-data__item">
+				<span class="test-connection-data__caption">
+					{{ row.caption }}
+				</span>
+				<span>
+					{{ row.description }}
+				</span>
+			</li>
+		</ul>
 	</li>
 </template>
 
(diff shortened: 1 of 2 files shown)


Antreesy:
I'm lacking knowledge on how else this connection could fail (and how to reproduce it), so wouldn't mind a second opinion here
cc @danxuliu @fancycode @SystemKeeper

fancycode:
Another thing you could test here is sending a [`hello` request](https://github.com/nextcloud/spreed/blob/ac7f14f59d25e9ab1cef9074860cc098c15b3b76/src/utils/signaling.js#L976) after the connection was established. This will verify that the backend url and shared secret are correctly configured in the signaling server.

fancycode:
> Another thing you could test here is sending a [`hello` request](https://github.com/nextcloud/spreed/blob/ac7f14f59d25e9ab1cef9074860cc098c15b3b76/src/utils/signaling.js#L976) after the connection was established. This will verify that the backend url and shared secret are correctly configured in the signaling server.

As I had a HPB setup support request from a customer yesterday where the backend url was a problem, this would be really great to test from the admin settings. Maybe you could also show the backend url ithat Talk will send to the signaling server in the UI so the user can compare with the configuration of the signaling server (see https://github.com/strukturag/nextcloud-spreed-signaling/blob/0a04ea290969cfe1ebc3ef3aaeeae1bdc49ea552/server.conf.in#L125-L126).

- URL used for backend requests: https://github.com/nextcloud/spreed/blob/bab024191b4b510a3f49bc25e42d771a88f83d01/lib/Signaling/BackendNotifier.php#L122
- URL used for hello requests: https://github.com/nextcloud/spreed/blob/bab024191b4b510a3f49bc25e42d771a88f83d01/lib/Config.php#L611

Antreesy:
Wil take a look

danxuliu (commented):
In my development environment if the secret is wrong the WebSocket is still successfully connected and no error message is shown.

I guess it is caused by using the `hello` protocol version `2.0` if available; if I am not mistaken in that case the signaling server does not send a request to the Nextcloud server to verify the authentication parameters, so the Websocket connection can be established even if the secret is wrong (although then an error will be received as soon as the signaling server sends a request to the Nextcloud server, for example, when the user tries to join a conversation).

If my guess is right then it would be better to use the `hello` protocol version `1.0` for the test, independently of what is supported by the signaling server. @fancycode Would there be any drawbacks with that? And even more important, is my guess right, or am I missing something? :-)

nickvergessen (approved):
Works better then before and is blocking an upcoming feature, so fine by me to merge like this and look into secret verification as a follow-up?

fancycode:
> In my development environment if the secret is wrong the WebSocket is still successfully connected and no error message is shown.
> 
> I guess it is caused by using the `hello` protocol version `2.0` if available; if I am not mistaken in that case the signaling server does not send a request to the Nextcloud server to verify the authentication parameters, so the Websocket connection can be established even if the secret is wrong (although then an error will be received as soon as the signaling server sends a request to the Nextcloud server, for example, when the user tries to join a conversation).

Right, the hello v2 was introduced to get rid of the request to verify the tickets in the hello request. With v2, the public key is fetched from the capabilities (and cached), so the provided ticket can be checked without additional requests.

> If my guess is right then it would be better to use the `hello` protocol version `1.0` for the test, independently of what is supported by the signaling server. @fancycode Would there be any drawbacks with that? And even more important, is my guess right, or am I missing something? :-)

That would be a way to check the secret, however you will not test that v2 also works ;) So the best would be to check a v1 hello to verify the secret and v2 to verify the public key from the capabilities. Note that you will have to establish a new WebSocket connection for each test (sending a second hello through an already established session will not work).

Antreesy:
Thanks, @danxuliu @fancycode !

Attempt is made at https://github.com/nextcloud/spreed/pull/14121.
Although I'm unable to catch failed secret: 
image
Maybe, there is no actual request made (as in BackendNotifier.php)

PR #13977, opened 2024-12-10
fix: leave call from system message 'call_ended_everyone'

Antreesy:
### ☑️ Resolves

* Fix hanging interface when call with a lot of participants ended for everyone
* Client wait for signaling update (`usersInCallChanged`) to leave by itself
* Additionaly, system message in chat is a good equivalent of the fact call has ended

How to test: 
1) apply patch to imitate delay in receiving a signal
```diff
diff --git a/src/utils/webrtc/webrtc.js b/src/utils/webrtc/webrtc.js
--- a/src/utils/webrtc/webrtc.js	(revision c4a63ff18ae6aa83acb0f54e14c7d91832fabee5)
+++ b/src/utils/webrtc/webrtc.js	(date 1733849256161)
@@ -473,10 +473,12 @@
 		&& localUserInCall) {
 		console.info('Force leaving the call for current participant')
 
+		setTimeout(() => {
 		store.dispatch('leaveCall', {
 			token: store.getters.getToken(),
 			participantIdentifier: store.getters.getParticipantIdentifier(),
 		})
+		}, 3000)
 
 		// Do not return to disconnect already from the other participants
 		// without waiting for another signaling event about changed users.
```
2) Join call as user1 + user2
3) As user1, end call for everyone

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/store/messagesStore.js
+++ b/src/store/messagesStore.js
@@ -1132,6 +1132,11 @@ const actions = {
 						&& message.actorType === context.getters.getActorType())) {
 					const callViewStore = useCallViewStore()
 					callViewStore.setCallHasJustEnded(message.timestamp)
+
+					context.dispatch('leaveCall', {
+						token,
+						participantIdentifier: context.getters.getParticipantIdentifier(),
+					})
 				}
 			}
 


src/store/messagesStore.js
```diff
@@ -1130,6 +1130,10 @@ const actions = {
 					&& conversation.type !== CONVERSATION.TYPE.ONE_TO_ONE
 					&& !(message.actorId === context.getters.getActorId()
 						&& message.actorType === context.getters.getActorType())) {
+					context.dispatch('leaveCall', {
+						token,
+						participantIdentifier: context.getters.getParticipantIdentifier(),
+					})
 					const callViewStore = useCallViewStore()
 					callViewStore.setCallHasJustEnded(message.timestamp)
```
nickvergessen, line 1138: I think we should do this first, to disable the button already

nickvergessen (changes requested):
In a quick test, this only seems to work/make a difference with HPB and not for the moderator.

but I tested with 2 sleeps in the leave call controller method:
```diff
diff --git a/lib/Controller/CallController.php b/lib/Controller/CallController.php
index 8128e4314..0c0f703c0 100644
--- a/lib/Controller/CallController.php
+++ b/lib/Controller/CallController.php
@@ -510,6 +510,7 @@ class CallController extends AEnvironmentAwareOCSController {
                        return $response;
                }
 
+               sleep(5);
                if ($all && $this->participant->hasModeratorPermissions()) {
                        $result = $this->roomService->resetActiveSinceInDatabaseOnly($this->room);
                        if (!$result) {
@@ -526,6 +527,7 @@ class CallController extends AEnvironmentAwareOCSController {
                        }
                }
 
+               sleep(15);
                return new DataResponse(null);
        }
 
```

[Bildschirmaufzeichnung vom 2024-12-11 07-34-08.webm](https://github.com/user-attachments/assets/538e7dff-62db-4c0f-b3a1-31519caac52c)

nickvergessen (approved):

PR #13979, opened 2024-12-11
fix(call): Fix call notification when being pinged again after leavin…

### LINKED ISSUE #13966: Call notification state api does not handle leaving in between
### How to use GitHub

---

## Steps to reproduce
1. User A starts call, push notification is shown
2. User B joins call, push notification is dismissed
3. User B leaves call
4. User A sends a call notification to user B again (without leaving the call)
5. API now returns participant joined immediately, although not joined anymore.

### Expected behaviour
API should return "call in progress", not "participant joined"

## Talk app

**Talk app version:** 20.1.0

nickvergessen:
…g a call

### ☑️ Resolves

Fix #13966

The participant had joined the call, but left again.
In this case we should not ring any more, but clients stop
pinging the endpoint 45s after receiving the push anyway.
However, it is also possible that the participant was ringed
again by a moderator after they had joined the call before.
So if a client pings the endpoint after 45s initial ringing
\+ 15 seconds for worst case push notification delay, we will
again tell them to show the call notification.

### 🚧 Tasks

- I don't feel like making an integration test that waits 60+ seconds, so can't add a test for this.

### 🏁 Checklist

### DIFF
--- a/lib/Service/ParticipantService.php
+++ b/lib/Service/ParticipantService.php
@@ -1653,6 +1653,18 @@ class ParticipantService {
 		if ($activeSince->getTimestamp() >= $row['last_joined_call']) {
 			return CallNotificationController::CASE_STILL_CURRENT;
 		}
+
+		// The participant had joined the call, but left again.
+		// In this case we should not ring any more, but clients stop
+		// pinging the endpoint 45s after receiving the push anyway.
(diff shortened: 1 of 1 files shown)


SystemKeeper:
Test

Antreesy (approved):
Tested with Desktop client

PR #13980, opened 2024-12-11
fix(guests): Don't force remove guests but leave it to the "purge logic"

nickvergessen:
Currently guests with a display name are not cleaned up, to keep their name for chat messages and other activity.

## 👣 Steps

1. Set up HPB
2. Create a conversation with public access
3. Join as a guest and set a name
4. Write a message
5. Reload the page
6. Check if the message from 4 still has the display name

x | Firefox | Chrome
---|---|---
Before | :white_check_mark: | :x:
After | :white_check_mark: |:white_check_mark: 

The main difference is that Firefox seems to keep the session and is still recognized as the previous guest, chrome creates a new attendees entry on each reload

## 🛠️ API Checklist

- As this is the reaction to a request from the HPB no suitable test can be written.

### 🏁 Checklist

### DIFF
--- a/lib/Controller/SignalingController.php
+++ b/lib/Controller/SignalingController.php
@@ -10,7 +10,6 @@ namespace OCA\Talk\Controller;
 
 use GuzzleHttp\Exception\ConnectException;
 use OCA\Talk\Config;
-use OCA\Talk\Events\AAttendeeRemovedEvent;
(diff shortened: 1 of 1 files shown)


miaulalala (approved):

Antreesy (approved):
Tested

PR #14008, opened 2024-12-16
feat(meetings): Preview upcoming events

Antreesy:
### ☑️ Resolves

* Ref #6292

Light | Dark
-- | --
image | image
image | image
image | image

2024-12-17_17h24_33

### 🚧 Tasks

- [x] Get calendars info and map to Select option
- [x] Preview upcoming events (in fancy way)
- [x] 401 - public conversations
- [ ] Add "shedule meeting" form to submit - https://github.com/nextcloud/spreed/pull/14074

Trigger | Form
-- | --
2024-12-16_19h34_42 | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/CalendarEventsDialog.vue
+++ b/src/components/CalendarEventsDialog.vue
@@ -1,0 +1,183 @@
+<!--
+  - SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<script setup lang="ts">
+import { computed, onBeforeMount, ref } from 'vue'
+
+import IconCalendarBlank from 'vue-material-design-icons/CalendarBlank.vue'
+import IconCalendarRefresh from 'vue-material-design-icons/CalendarRefresh.vue'
+
+import { t } from '@nextcloud/l10n'
+import moment from '@nextcloud/moment'
+
+import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
+import NcDialog from '@nextcloud/vue/dist/Components/NcDialog.js'
+import NcEmptyContent from '@nextcloud/vue/dist/Components/NcEmptyContent.js'
+import NcLoadingIcon from '@nextcloud/vue/dist/Components/NcLoadingIcon.js'
+import usernameToColor from '@nextcloud/vue/dist/Functions/usernameToColor.js'
+
+import { useGroupwareStore } from '../stores/groupware.ts'
+
+const props = defineProps<{
+	token: string,
+	container?: string,
+}>()
+const emit = defineEmits<{
+	(event: 'close'): void,
+}>()
+
+const groupwareStore = useGroupwareStore()
+
+const open = ref(false)
+const loading = ref(Object.keys(groupwareStore.calendars).length === 0)
+
+const calendars = computed(() => groupwareStore.calendars)
+const upcomingEvents = computed(() => {
+	const now = moment().unix()
+	return groupwareStore.getAllEvents(props.token)
+		.sort((a, b) => (a.start && b.start) ? (a.start - b.start) : 0)
+		.map(event => {
+			const start = event.start
+				? (event.start <= now) ? t('spreed', 'Now') : moment(event.start * 1000).calendar()
+				: ''
+			const color = calendars.value[event.calendarUri]?.color ?? usernameToColor(event.calendarUri).color
+
+			return { ...event, start, color, href: event.calendarAppUrl ?? undefined }
+		})
+})
+
+onBeforeMount(() => {
+	getCalendars()
+})
+
+/**
+ * Show upcoming events dialog
+ */
+function openDialog() {
+	open.value = true
+}
+
+/**
+ * Get user's calendars to identify belonging of known and future events
+ */
+async function getCalendars() {
+	await groupwareStore.getPersonalCalendars()
+	loading.value = false
+}
+</script>
+
+<template>
+	<div>
+		<slot name="trigger" :on-click="openDialog">
+			<NcButton :title="t('spreed', 'Upcoming events')"
+				:aria-label="t('spreed', 'Upcoming events')"
+				@click="openDialog">
+				<template #icon>
+					<IconCalendarBlank :size="20" />
+				</template>
+			</NcButton>
+		</slot>
+
+		<NcDialog :open.sync="open"
+			class="calendar-events"
+			:name="t('spreed', 'Upcoming events')"
+			size="normal"
+			close-on-click-outside
+			:container="container">
+			<template v-if="!loading && upcomingEvents.length">
+				<ul class="calendar-events__list">
+					<!-- Upcoming event -->
+					<li v-for="event in upcomingEvents" :key="event.uri">
+						<a class="calendar-events__item"
+							:class="{ 'calendar-events__item--thumb': !event.href }"
+							:href="event.href"
+							:title="t('spreed', 'Open Calendar')"
+							target="_blank">
+							<IconCalendarRefresh v-if="event.recurrenceId" :size="20" />
+							<IconCalendarBlank v-else :size="20" />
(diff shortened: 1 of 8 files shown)


Antreesy:
cc @nimishavijay for initial design feedback

nimishavijay:
Nice! Looks great already! :) Some points for feedback (sorted by priority from a design perspective, so first item is highest prio, last one is low prio)

- [ ] Can we show the "Schedule a new meeting" inside the modal/popover as the last item with secondary styling and a plus icon to avoid putting more and more things in the 3 dot menu?

- [ ] Would it also be possible to move this whole thing into a popover that appears below the "Next call" button? Then we can remove the "Upcoming events" heading too?

- [ ] Would it be possible to replicate the UI of the calendar list view? That way it's not a deviation from the existing design already

Altogether ideally it could look like this, but the 3 points can be worked on separately as well :) 

image

Great work so far! :)

Antreesy:
> * Can we show the "Schedule a new meeting" inside the modal/popover as the last item with secondary styling and a plus icon to avoid putting more and more things in the 3 dot menu?

Doable, but atm we do not show 'Next call' popup if there are no events assigned to this conversation (also hidden on mobile screensize). Shall we show it as always-on?
image


> * Would it be possible to replicate the UI of the calendar list view?

We're missing some attributes from API (like endtime of meeting, or whether it's an "All day" event. But I think, reccurence indicator and start time would be enough for now?

image

nimishavijay:
> Doable, but atm we do not show 'Next call' popup if there are no events assigned to this conversation (also hidden on mobile screensize). Shall we show it as always-on?

Ah I see, no worries, the screenshot you posted seems great. We should just reduce the width and the "Schedule meeting" button at the bottom of the empty content :) 

> We're missing some attributes from API (like endtime of meeting, or whether it's an "All day" event. But I think, reccurence indicator and start time would be enough for now?

Checked with the rest of the design team and the verdict was that your original design looks way better visually 😅 So we could go ahead with that instead of the list view in calendar, with only the wording change to use the same wording as it is shown in the "Next call" button. So instead of 
```
Talk team call
17 December 2024 19:00 
```

It can be 
```
Talk team call
Tuesday 19:00
```

Sorry for the confusion!

src/components/TopBar/TopBar.vue
```diff
@@ -300,9 +304,12 @@ export default {
...
 		},
 
+		showCalendarEvents() {
+			return !this.isInCall && !this.isSidebar && this.conversation.type !== CONVERSATION.TYPE.NOTE_TO_SELF
+		},
+
 		showUpcomingEvent() {
-			return this.nextEvent && !this.isInCall && !this.isSidebar && !this.isMobile
-				&& this.conversation.type !== CONVERSATION.TYPE.NOTE_TO_SELF
+			return this.showCalendarEvents && this.nextEvent && !this.isMobile
```
DorraJaouad, line 312:
```suggestion
			return this.nextEvent && !this.isMobile
```

src/components/CalendarEventsDialog.vue
```diff
@@ -0,0 +1,156 @@
...
+			<ul v-if="!loading && Object.keys(upcomingEvents).length" class="calendar-events__wrapper">
+				<template v-for="(day, key) in upcomingEvents">
+					<li :key="key" class="calendar-events__date">
+						{{ key }}
+					</li>
+					<li v-for="event in day" :key="event.uri">
+						<a class="calendar-events__item"
+							:href="event.calendarAppUrl"
+							:title="t('spreed', 'Open Calendar')"
+							target="_blank">
```
DorraJaouad, line 91: it's the same link, just add a button at the bottom?

src/components/CalendarEventsDialog.vue
```diff
@@ -0,0 +1,156 @@
...
+		</template>
+	</NcPopover>
+</template>
+
+<style lang="scss" scoped>
+.calendar-events {
+	&__wrapper {
+		width: 400px;
+		padding: calc(var(--default-grid-baseline) * 2);
+	}
```
DorraJaouad, line 120: `max-height` ?

src/components/TopBar/TopBar.vue
```diff
@@ -56,22 +56,24 @@
...
-				<div class="event-info">
-					<p class="event-info__header">
-						{{ t('spreed', 'Next call') }}
-					</p>
-					<p> {{ eventInfo }} </p>
-				</div>
-			</a>
+			<!-- Upcoming events dialog -->
+			<CalendarEventsDialog v-show="showCalendarEvents" :token="token">
+				<template #trigger>
```
DorraJaouad, line 61: image

src/components/CalendarEventsDialog.vue
```diff
@@ -0,0 +1,174 @@
...
+	:deep(.dialog__content) {
+		padding-block-end: calc(var(--default-grid-baseline) * 3);
+	}
+
+	&__list {
+		display: flex;
+		flex-direction: column;
+		margin: var(--default-grid-baseline);
+		gap: calc(var(--default-grid-baseline) * 2);
+	}
```
DorraJaouad, line 137: Make it scrollable? so it doesn't exceed a certain height

src/components/TopBar/TopBar.vue
```diff
@@ -300,11 +305,15 @@ export default {
 			return moment(this.nextEvent.start * 1000).calendar()
 		},
 
-		showUpcomingEvent() {
-			return this.nextEvent && !this.isInCall && !this.isSidebar && !this.isMobile
+		showCalendarEvents() {
+			return this.getUserId && !this.isInCall && !this.isSidebar
```
DorraJaouad, line 309: What about public calendars?

src/stores/groupware.ts
```diff
@@ -9,33 +9,49 @@ import Vue from 'vue'
...
 		getAllEvents: (state) => (token: string) => {
-			return state.upcomingEvents[token]
+			return state.upcomingEvents[token] ?? []
 		},
 		getNextEvent: (state) => (token: string) => {
 			return state.upcomingEvents[token]?.[0]
 		},
+		writeableCalendars: (state) => {
+			return Object.values(state.calendars).filter(calendar => {
+				return calendar.isWriteable() && calendar.components.includes('VEVENT')
```
DorraJaouad, line 52: Can you please share the source of information?

src/components/CalendarEventsDialog.vue
```diff
@@ -0,0 +1,174 @@
...
+	&__item {
+		display: flex;
+		flex-direction: row;
+		align-items: center;
+		gap: calc(var(--default-grid-baseline) * 2);
+		padding: 0 calc(var(--default-grid-baseline) * 2);
+		color: var(--color-primary-element-light-text);
+		background-color: var(--color-primary-element-light);
+		height: 100%;
+		border-radius: var(--border-radius);
```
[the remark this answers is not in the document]
Antreesy: Looks better with 4px
B | A
-- | --
image | image

src/components/TopBar/TopBar.vue
Antreesy: Still the issue on server side. But this logic is the same as in watcher:
```js
if (!value || this.isSidebar || !this.getUserId) {
	// Do not fetch upcoming events for guests (401 unauthorzied) or in sidebar
	return
}
```

src/stores/groupware.ts
Antreesy: From calendar app:
- can write: https://github.com/nextcloud/calendar/blob/c1a408e8e0bc9df791c5a066bd1e44f8408417e6/src/models/calendar.js#L84
- can create new events: https://github.com/nextcloud/calendar/blob/c1a408e8e0bc9df791c5a066bd1e44f8408417e6/src/models/calendar.js#L80

DorraJaouad (approved):
Tested 👍🏻 ✅

PR #14010, opened 2024-12-17
chore(deps): Update openapi-extractor to v1.3.0

provokateurin:
### 🏁 Checklist

### DIFF
--- a/lib/Model/BotServer.php
+++ b/lib/Model/BotServer.php
@@ -8,7 +8,6 @@ declare(strict_types=1);
 
 namespace OCA\Talk\Model;
 
-use OCA\Talk\ResponseDefinitions;
 use OCP\AppFramework\Db\Entity;
 use OCP\DB\Types;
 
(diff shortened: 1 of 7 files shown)


provokateurin:
Fun, psalm locally doesn't produce the error

nickvergessen:
> Fun, psalm locally doesn't produce the error

It does for me, but dropping 073096c570e63414aa587f80b794f775febe5324 should work? It claims to drop an unused schema and its used but you replace it with something?

nickvergessen:
Ah, it's not used in the OpenAPI, but only when doing commands.

lib/Model/BotServer.php
```diff
@@ -62,7 +59,19 @@ public function __construct() {
 	}
 
 	/**
-	 * @return TalkBotWithDetailsAndSecret
+	 * @return array{
```
nickvergessen, line 62: booo :mega:

nickvergessen (approved):

lib/Model/BotServer.php
provokateurin: I knowwww :pensive:

PR #14012, opened 2024-12-18
fix(backends): Check times of the backend servers

nickvergessen:
## 🛠️ API Checklist

Recording | HPB
---|---
Bildschirmfoto vom 2024-12-18 06-58-27 | Bildschirmfoto vom 2024-12-18 06-45-55

### 🏁 Checklist


SystemKeeper (approved):
Tested and works, nice!
It’s a very narrow window we enforce, should there be some kind of threshold of a few seconds that we still accept as being valid?

fancycode (commented):
+1 for some leeway.

The signaling server allows up to one minute time difference when validating tokens:
https://github.com/strukturag/nextcloud-spreed-signaling/blob/0a04ea290969cfe1ebc3ef3aaeeae1bdc49ea552/hub.go#L114-L115

nickvergessen:
then we use 45 to be on the safe side, so there are still ~15s in worst case between getting the token and sending the request?

fancycode:
> then we use 45 to be on the safe side, so there are still ~15s in worst case between getting the token and sending the request?

:+1: sounds good

fancycode (approved):
:rocket: lgtm

Issue #14018, opened 2024-12-18
Improve HPB warning scenarios


nickvergessen:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

- Warn in all calls if HPB is not present (the 4 person warning atm seems to have disappeared?)
- Warn in admin settings and setupcheck about incomplete setup if HPB is not present

>[!Warning]
> ## Nextcloud Talk setup not complete
> Install the High Performance Backend to ensure calls with multiple people work seamlessly.

Antreesy:
> * Warn in admin settings and setupcheck about incomplete setup if HPB is not present

In which place?
<img width="247" alt="Image" src="https://github.com/user-attachments/assets/dead3dd8-1a70-42e2-9aee-4768e5e8312d" />
<img width="355" alt="Image" src="https://github.com/user-attachments/assets/a2dda62d-84da-45df-9b08-f4ef90d691c7" />

nimishavijay:
Great idea! We should also add a call to action button in the note card as an obvious next step, possibly with a link to the relevant documentation or something else that would help the set up process

nickvergessen:
**Warning history**

- Added https://github.com/nextcloud/spreed/pull/1649
- Simplified https://github.com/nextcloud/spreed/pull/1653
- Moved from signaling to application in https://github.com/nextcloud/spreed/pull/1821
- `isNoMcuWarningEnabled` is defined but unused after the vue rewrite in stable18

Antreesy:
From PR discussion (QualityWarningTooltip):

- [ ] we can use an icon in the call view, I'd even say use the same connection icon entry, just add another block of text in the popover?

nickvergessen:
- Setup warning in https://github.com/nextcloud/spreed/pull/14063
- Sidebar warning during a call in https://github.com/nextcloud/spreed/pull/14022

PR #14022, opened 2024-12-18
fix(signaling): add UI warning in call with 2+ participants

Antreesy:
### ☑️ Resolves

* Ref #14018
* Show warning during the call in group / public conversations, when only internal signaling is available

🏚️ Before | 🏡 After
-- | --
Talk app | image
Files app and sidebars | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/PublicShareAuthSidebar.vue
+++ b/src/PublicShareAuthSidebar.vue
@@ -13,6 +13,7 @@
 			<template v-else>
 				<TopBar is-in-call is-sidebar />
 				<CallView :token="token" is-sidebar />
+				<InternalSignalingHint />
 				<ChatView is-sidebar />
 				<PollViewer />
 				<MediaSettings :recording-consent-given.sync="recordingConsentGiven" />
@@ -31,6 +32,7 @@ import CallView from './components/CallView/CallView.vue'
 import ChatView from './components/ChatView.vue'
 import MediaSettings from './components/MediaSettings/MediaSettings.vue'
 import PollViewer from './components/PollViewer/PollViewer.vue'
+import InternalSignalingHint from './components/RightSidebar/InternalSignalingHint.vue'
 import TopBar from './components/TopBar/TopBar.vue'
 import TransitionWrapper from './components/UIShared/TransitionWrapper.vue'
 
@@ -48,6 +50,7 @@ export default {
 	name: 'PublicShareAuthSidebar',
 
 	components: {
+		InternalSignalingHint,
 		CallView,
 		ChatView,
 		MediaSettings,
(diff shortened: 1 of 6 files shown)


src/App.vue
```diff
@@ -615,7 +622,13 @@ export default {
 
 			// Breakout to breakout
 			return oldConversation.objectType === CONVERSATION.OBJECT_TYPE.BREAKOUT_ROOM && newConversation.objectType === CONVERSATION.OBJECT_TYPE.BREAKOUT_ROOM
-		}
+		},
+
+		showSignalingInternalWarning(token) {
+			if (this.$store.getters.isModerator || this.$store.getters.participantsList(token)?.length > 4) {
```
nickvergessen, line 628:
```suggestion
			if (this.$store.getters.participantsList(token)?.length > 2) {
```

nickvergessen (commented):
Just a quick comment, two things:
- the warning basically will be shown in any non 2 person call in the future
- we had this warning or a similar one in the past, can you check and find it or find where it got lost?

Antreesy:
> we had this warning or a similar one in the past, can you check and find it or find where it got lost?

Never appeared after migration to Vue in 2019: https://github.com/nextcloud/spreed/commit/88d7d7820e5ae8f4a3d897605124ad7755926494

jancborchardt (changes requested):
The issue with the toast is that it gets clicked away and is not permanent enough. I would suggest to put a yellow NcNoteCard on the top of the right sidebar (or left navigation, although that seems to be less relevant if this only shows in a call) that shows throughout the duration of the call.

And it also needs to provide how to fix it – if shown to users, needs to say something actionable like "Contact your admin for further assistance"

nickvergessen:
If a sidebar it should be right, as left is folded away when starting a call.
Could also see an icon on the call view similarly to the connection warning?

src/App.vue
```diff
@@ -615,7 +622,13 @@ export default {
...
+				showWarning(t('spreed', 'Having a call with more than 4 participants without external signaling server can cause connectivity issues and cause high load on participating devices'))
```
nickvergessen, line 629: Instead of showWarning we should use a NcNoteCard in the header of the right sidebar below the Conversation name (above the lobby and other quick settings). Together with an actionable "Learn more" link which links to https://portal.nextcloud.com/article/Nextcloud-Talk/High-Performance-Backend/Installation-of-Nextcloud-Talk-High-Performance-Backend for now.

nickvergessen (approved):
Good for me.

We can also add it to the call view additionally as described by Jan, or we wait to see how the current version resonates.

PR #14024, opened 2024-12-19
fix(search): Paginate unified search results

### LINKED ISSUE #14011: Support pagination for search provider
Currently (as of 20.1.0) Talk does not support pagination for search results from unified search. That can lead to performance issues with a large number of conversations being returned for a search query and rendering the conversations.

We should think about supporting pagination for the search provider, or make sure that virtual scrolling is used for the results (different issue on the server/vue lib).

Reference for the search provider:

https://github.com/nextcloud/server/blob/fe90aa3ffe0a8127a92cf0226333c44ed5b793dc/lib/public/Search/IProvider.php#L63-L89

nickvergessen:
### ☑️ Resolves

* Fix #14011 

### 🚧 Tasks

- [ ] Notify clients

### 🏁 Checklist

- [?] 🔖 Capability is added or not needed


SystemKeeper (approved):
Nice!
Tested with Unified Search and Smart Picker, both seem to work fine.

Thanks for adding the comments.

miaulalala (approved):
looking good

PR #13967, opened 2024-12-10
test(federation): Run cross major federation on CI

nickvergessen:
### ☑️ Resolves

* Ref #12805
* Fix #12790

PR #14066, opened 2025-01-02
test: Fix local integration tests with custom skeleton directory

nickvergessen:
### ☑️ Resolves

```
System config value overwrite.cli.url set to string http://localhost:8080/
Resetting custom skeletondirectory so that tests pass
./run.sh: Zeile 120: config:system:delete: Befehl nicht gefunden
System config value overwrite.cli.url set to string http://localhost:8280/
```

And following that any integration test that checks files is failing.

- Follow up to the restructuring of https://github.com/nextcloud/spreed/pull/13967

### 🏁 Checklist


tests/integration/run.sh
```diff
@@ -117,7 +117,7 @@ MAIN_SKELETON_DIR=$(occ_host config:system:get skeletondirectory)
 occ_host config:system:set overwrite.cli.url --value "http://localhost:8080/"
```
nickvergessen, line 117: see here :see_no_evil:

SystemKeeper (approved):
🙈

PR #14074, opened 2025-01-07
📆 feat(meetings): Allow moderators to schedule a meeting

Antreesy:
### ☑️ Resolves

* Follow-up to #14008 
* API  #14073

image
image

### 🚧 Tasks

- [x] Add API handling
- [x] Split into Popover and Dialog
- [ ] Participants selector - follow-up

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/CalendarEventsDialog.vue
+++ b/src/components/CalendarEventsDialog.vue
@@ -4,20 +4,27 @@
 -->
 
 <script setup lang="ts">
-import { computed, onBeforeMount, ref } from 'vue'
+import { computed, onBeforeMount, ref, watch } from 'vue'
 
 import IconCalendarBlank from 'vue-material-design-icons/CalendarBlank.vue'
 import IconCalendarRefresh from 'vue-material-design-icons/CalendarRefresh.vue'
+import IconCheck from 'vue-material-design-icons/Check.vue'
 
 import { t } from '@nextcloud/l10n'
 import moment from '@nextcloud/moment'
 
 import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
+import NcDateTimePickerNative from '@nextcloud/vue/dist/Components/NcDateTimePickerNative.js'
 import NcDialog from '@nextcloud/vue/dist/Components/NcDialog.js'
 import NcEmptyContent from '@nextcloud/vue/dist/Components/NcEmptyContent.js'
 import NcLoadingIcon from '@nextcloud/vue/dist/Components/NcLoadingIcon.js'
+import NcSelect from '@nextcloud/vue/dist/Components/NcSelect.js'
+import NcTextArea from '@nextcloud/vue/dist/Components/NcTextArea.js'
+import NcTextField from '@nextcloud/vue/dist/Components/NcTextField.js'
 import usernameToColor from '@nextcloud/vue/dist/Functions/usernameToColor.js'
 
+import { useStore } from '../composables/useStore.js'
+import { hasTalkFeature } from '../services/CapabilitiesManager.ts'
 import { useGroupwareStore } from '../stores/groupware.ts'
 
 const props = defineProps<{
@@ -28,10 +35,12 @@ const emit = defineEmits<{
 	(event: 'close'): void,
 }>()
 
+const store = useStore()
 const groupwareStore = useGroupwareStore()
 
 const open = ref(false)
 const loading = ref(Object.keys(groupwareStore.calendars).length === 0)
+const submitting = ref(false)
 
 const calendars = computed(() => groupwareStore.calendars)
 const upcomingEvents = computed(() => {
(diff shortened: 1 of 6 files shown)


src/components/CalendarEventsDialog.vue
```diff
@@ -117,6 +156,45 @@ async function getCalendars() {
...
+						</template>
+					</NcSelect>
+					<NcDateTimePickerNative id="schedule_meeting_input"
+						v-model="selectedDateTime"
+						:label="t('spreed', 'Meeting start time')"
+						type="datetime-local" />
+				</div>
+				<NcTextField v-model="newMeetingTitle"
+					:label="t('spreed', 'Name')"
+					:placeholder="t('spreed', 'Name')"
```
nickvergessen, line 182: In the calendar app it's called "Event title" maybe we go with title then instead?

nimishavijay (commented):
Awesome!! Looks very cool! 
Some questions from my end: 
- [x] is it not possible to set the end time too? Seems a bit odd that we can set only the start time, and it's also not so clear how long the event is 
- [ ] will this be shown when you click on the top right "Next event" button in Talk? It would be nice to show the upcoming events in a popover along with a "schedule event" button, and show the schedule event section as a separate modal when you click on the button.

nickvergessen:
>  is it not possible to set the end time too? Seems a bit odd that we can set only the start time, and it's also not so clear how long the event is


Fine by me to add a second date-time picker

Antreesy:
> will this be shown when you click on the top right "Next event" button in Talk? It would be nice to show the upcoming events in a popover along with a "schedule event" button, and show the schedule event section as a separate modal when you click on the button.

Yes, upon the click. Wasn't it decided here to return back to 'previous design' (dialog instead of popover)? https://github.com/nextcloud/spreed/pull/14008#issuecomment-2554145365

nimishavijay:
@Antreesy your original list item design, but in a popover :)

nickvergessen:
Maybe we do a quick call to align on it? I'm also quite confused at the moment :D

src/components/CalendarEventsDialog.vue
Antreesy: aligned with Calendar app

src/components/CalendarEventsDialog.vue
```diff
@@ -118,6 +163,49 @@ async function getCalendars() {
...
+					v-model="selectedCalendar"
+					:options="calendarOptions"
+					:input-label="t('spreed', 'Select calendar')">
+					<template #option="option">
+						<span class="calendar-badge" :style="{ backgroundColor: option.color }" />
+						{{ option.label }}
+					</template>
+				</NcSelect>
+				<div class="calendar-meeting__flex-wrapper">
+					<NcDateTimePickerNative id="schedule_meeting_input"
```
nickvergessen, line 181: there is an option to enforce that start is > now

src/components/TopBar/TopBar.vue
```diff
@@ -56,25 +56,8 @@
...
-							<p class="event-info__header">
-								{{ t('spreed', 'Next call') }}
-							</p>
-							<p> {{ eventInfo }} </p>
-						</div>
-					</a>
-				</template>
-			</CalendarEventsDialog>
+			<!-- Upcoming meetings -->
+			<CalendarEventsDialog v-if="showCalendarEvents" :token="token" />
```
DorraJaouad, line 60: `container` prop missing or put it in the component directly

src/components/CalendarEventsDialog.vue
```diff
@@ -48,76 +63,217 @@ const upcomingEvents = computed(() => {
...
+		return
+	}
+
+	// Reset the default form values
+	selectedCalendar.value = calendarOptions.value.find(o => o.value === groupwareStore.defaultCalendarUri) ?? null
+	selectedDateTimeStart.value = new Date(moment().add(1, 'hours').startOf('hour'))
+	selectedDateTimeEnd.value = new Date(moment().add(2, 'hours').startOf('hour'))
+	newMeetingTitle.value = ''
+	newMeetingDescription.value = ''
+})
```
DorraJaouad, line 110: Better to use event `@closing` instead of a watcher ?

src/components/CalendarEventsDialog.vue
```diff
@@ -48,76 +63,217 @@ const upcomingEvents = computed(() => {
...
+				</p>
+			</div>
 
-				<template #description>
-					<p>{{ loading ? t('spreed', 'Loading …') : t('spreed', 'No upcoming events') }}</p>
-				</template>
-			</NcEmptyContent>
+			<template #actions>
+				<NcButton type="primary"
+					:disabled="!selectedCalendar || submitting || invalid"
```
DorraJaouad, line 268: Also when title is empty

src/components/CalendarEventsDialog.vue
```diff
@@ -48,76 +63,217 @@ const upcomingEvents = computed(() => {
...
+			</template>
+		</NcPopover>
 
-		<NcDialog :open.sync="open"
+		<NcDialog v-if="canScheduleMeeting"
+			:open.sync="open"
 			class="calendar-events"
-			:name="t('spreed', 'Upcoming events')"
-			size="normal"
+			:name="t('spreed', 'Schedule new meeting')"
```
DorraJaouad, line 221: `Schedule a meeting`?

src/components/CalendarEventsDialog.vue
```diff
@@ -48,76 +63,217 @@ const upcomingEvents = computed(() => {
...
+							<IconPlus :size="20" />
+						</template>
+						{{ t('spreed', 'Schedule new meeting') }}
+					</NcButton>
+				</div>
+			</template>
+		</NcPopover>
 
-		<NcDialog :open.sync="open"
+		<NcDialog v-if="canScheduleMeeting"
```
DorraJaouad, line 218: *Optional* change the size to `normal` so probability that "AM" "PM" are hidden is low
image

src/components/CalendarEventsDialog.vue
```diff
@@ -48,76 +63,217 @@ const upcomingEvents = computed(() => {
...
+	}
+
+	submitting.value = true
+	await groupwareStore.scheduleMeeting(props.token, {
+		calendarUri: selectedCalendar.value.value,
+		start: selectedDateTimeStart.value.getTime() / 1000,
+		end: selectedDateTimeEnd.value.getTime() / 1000,
+		title: newMeetingTitle.value || null,
+		description: newMeetingDescription.value || null,
+	})
```
DorraJaouad, line 149: It should not close the modal when there is an error, thus it's better to catch the error here and not in the store

src/components/CalendarEventsDialog.vue
Antreesy: title is nullable, will be conversation name by default. So we don't force it

src/components/CalendarEventsDialog.vue
Antreesy: we need it when opening (so selected time is up to date), so `update:open`

src/components/TopBar/TopBar.vue
Antreesy: it's optional, if we want to reuse component somewhere. Now it's attached directly to body

DorraJaouad (approved):
Tested 🦅 🥳

PR #14085, opened 2025-01-09
fix(hpb): Add a small heading to the call note

nickvergessen:

grafik


DorraJaouad (approved):
I also a warning in settings added recently for the same thing, shouldn't they be the same color (warning not error)?
image

nickvergessen:
Well users might not be able to do anything about it, so we only warn there and we want to highlight it important in the admin settings so we make it an error there

jancborchardt (approved):

PR #14098, opened 2025-01-10
feat: add admin toggle for e2ee calls

Antreesy:
### ☑️ Resolves

* Follow-up to #14005

🏚️ Before | 🏡 After
-- | --
image | image

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

---

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist


src/components/AdminSettings/GeneralSettings.vue
```diff
@@ -86,10 +114,28 @@ export default {
...
 	methods: {
 		t,
+
+		updateSignalingServers(servers) {
+			this.canEnableE2EE = servers.length > 0
+		},
+
+		updateE2EECallsEnabled(value) {
+			// TODO: add API handling
+			console.log(value)
```
nickvergessen, line 136: Set `call_end_to_end_encryption`

src/components/AdminSettings/GeneralSettings.vue
```diff
@@ -86,10 +114,28 @@ export default {
...
+
+	beforeDestroy() {
+		EventBus.off('signaling-servers-updated', this.updateSignalingServers)
 	},
 
 	methods: {
 		t,
+
+		updateSignalingServers(servers) {
+			this.canEnableE2EE = servers.length > 0
```
nickvergessen, line 131: gets kind of tricky, but basically we need to additionally check the HPBs welcome page result and it must not be missing the feature `join-features`

nickvergessen (changes requested):

src/components/AdminSettings/GeneralSettings.vue
fancycode: As discussed, the feature `join-features` is only necessary with the current implementation in #14005. I will change this to evaluate the global feature / appconfig from here.

src/components/AdminSettings/GeneralSettings.vue
```diff
@@ -27,15 +27,36 @@
...
+		<NcCheckboxRadioSwitch v-model="isE2EECallsEnabled"
+			type="switch"
+			:disabled="loading || !canEnableE2EE"
+			@update:model-value="updateE2EECallsEnabled">
+			{{ t('spreed', 'Enable encryption') }}
+		</NcCheckboxRadioSwitch>
+
+		<NcNoteCard v-if="!canEnableE2EE"
+			type="warning"
+			:text="t('spreed', 'End-to-end encryption is only possible with a High-performance backend.')" />
```
fancycode, line 56: This is a bit misleading. Without the HPB, the clients will create PeerConnections between themselves, which by definition then are also end-to-end encrypted.

nickvergessen: > Without the HPB, the clients will create PeerConnections between themselves, which by definition then are also end-to-end encrypted.

Well they shouldn't as we don't know if the other client can do it?
Maybe we need another call later today

nickvergessen: Summary from the call:
- Without HPB we shouldn't show the setting as it's "default"
- Without SIP we don't need to check if the HPB was updated
- If the setting is enabled, older mobile clients need to be blocked

fancycode:
@nickvergessen I have updated #14005 to check the feature / app config to enable e2ee.

nickvergessen (approved):

PR #14120, opened 2025-01-13
fix(meeting): adjust styles and wording

Antreesy:
### ☑️ Resolves

* Follow-up to #14097

- [x] Distinguish the entire section by adding a heading "Attendees"
- [x] Instead of a toggle, use a checkmark
- [x] "Select attendees" button can be tertiary with icon
- [x] "Select attendees" button and summary can be in different rows to prevent the button position shifting slightly when the number changes 
- [ ] Wording adjustments:
  - [ ] It took a second for me to understand `Invite all users and email guests` --> `Invite all participants in the conversation and email guests`
  - [x] `Select attendees` --> `Add attendees`
  - [ ] `Sending 14 invitations` --> `[👩 Alice Doe], [👨 Sam Smith] and 12 more guests will receive invites` (ideal) or `14 guests will receive invites`
    - [x] ⚠️  Using of rich components is troublesome, approached with what is used for reactions
  - [x] For errors, don't prefix with "Error: "
- [x] If possible, show the selected attendees up top like when you create a new group conversation in Talk (in fact, the whole modal can just be replicated)
- [x] Add a `✔️ Done` or `✔️ Save` button at the end to prevent confusion about whether or not the attendees are added  
 
🏚️ Before | 🏡 After
-- | --
image | image
image | image
loader | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/CalendarEventsDialog.vue
+++ b/src/components/CalendarEventsDialog.vue
@@ -6,6 +6,7 @@
 <script setup lang="ts">
 import { computed, onBeforeMount, provide, ref, watch } from 'vue'
 
+import IconAccountPlus from 'vue-material-design-icons/AccountPlus.vue'
 import IconAccountSearch from 'vue-material-design-icons/AccountSearch.vue'
 import IconCalendarBlank from 'vue-material-design-icons/CalendarBlank.vue'
 import IconCheck from 'vue-material-design-icons/Check.vue'
@@ -96,14 +97,14 @@ const invalidHint = computed(() => {
 	case null:
 		return ''
 	case 'calendar':
-		return t('spreed', 'Error: Invalid calendar selected')
+		return t('spreed', 'Invalid calendar selected')
 	case 'start':
-		return t('spreed', 'Error: Invalid start time selected')
+		return t('spreed', 'Invalid start time selected')
 	case 'end':
-		return t('spreed', 'Error: Invalid end time selected')
+		return t('spreed', 'Invalid end time selected')
 	case 'unknown':
 	default:
-		return t('spreed', 'Error: Unknown error occurred')
+		return t('spreed', 'Unknown error occurred')
 	}
 })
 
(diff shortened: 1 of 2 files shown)


nimishavijay (changes requested):
Looks super nice! 🎉 Great workaround considering we are limited to not using rich components! Only 2 small comments regarding wording:


- @nickvergessen in #14097 said

> Well but it does not invite all participants... federated users don't work, normal guests don't work, ...

I understand, I would suggest then that we say `Invite participants in the conversation and email guests` to avoid using the word "users", what do you think? 

- `Oleg, Alice Ricketts and 1 other will receive invitations` --> `Oleg, Alice Ricketts and 1 more guest will receive invitations` 

Other than that it's good to go 🚀

Antreesy:
> to avoid using the word "users", what do you think?

I think there is still need to specify who is invited. Term 'Participants' in conversation includes a lot of types, but not everyone can receive an invite:
- [x] Users/Accounts
  - [ ] Guest accounts - need to check
- [ ] Public link guests - have no email to sent invites to
  - [x] Email guests
- [ ] Federated users
- [ ] Groups
- [ ] Teams/Circles
- [ ] Bots
- [ ] SIP users (phones)
- [ ] Bridged users (from other services)

> * `Oleg, Alice Ricketts and 1 other will receive invitations` --> `Oleg, Alice Ricketts and 1 more guest will receive invitations`

Same here, guest is not what meant to be here (could be other user/account). If we properly define the text above (in checkbox, or in a hint, and then 'Invite everyone' in checkbox), then '...others...' should be fine here?

jancborchardt:
@nickvergessen agree wording should be consistent. We are doing a slow but steady push to replace "users" with "accounts" or "people" depending on the case. In the cases you show, "accounts" would be a good fit.

nickvergessen:
> We are doing a slow but steady push to replace "users" with "accounts" or "people" depending on the case. In the cases you show, "accounts" would be a good fit.

We can cut through the whole code base after feature freeze, not during it.
We have 98 mentions of "user" in our language file and it will not be possible to transition and afterwards translate this in time for the release.
So for the Hub 10 / Nextcloud 31 release Talk will have to continue with the word "user" when "accounts" are meant.

DorraJaouad (approved):
Tested, Talk-side works
Logic and UI are good

PR #14143, opened 2025-01-16
fix: user contacts menu instead of conversation participants (there c…

DorraJaouad:
### ☑️ Resolves

* Follow up to #14125 (review comments)

- [x] Missing items navigation
- [x] Correct list of contacts rather conversation participants (some users might have left and they have messages)
- [x] Trigger re-fetch after removing filter chips
- [x] Types syntax

https://github.com/user-attachments/assets/4da51642-6b6b-44be-b340-d29ce0fc79b5

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop
…ould be users who left)

### DIFF
--- a/src/components/RightSidebar/SearchMessages/SearchMessagesTab.vue
+++ b/src/components/RightSidebar/SearchMessages/SearchMessagesTab.vue
@@ -54,6 +54,7 @@ const emit = defineEmits<{
 const searchMessagesTab = ref<HTMLElement | null>(null)
 const searchBox = ref<InstanceType<typeof SearchBox> | null>(null)
 const { initializeNavigation, resetNavigation } = useArrowNavigation(searchMessagesTab, searchBox)
+
 const isFocused = ref(false)
 const searchResults = ref<(CoreUnifiedSearchResultEntry &
 {
@@ -247,6 +248,8 @@ async function fetchSearchResults(isNew = true): Promise<void> {
 }
 
 const debounceFetchSearchResults = debounce(fetchNewSearchResult, 250)
+
+watch([searchText, fromUser, sinceDate, untilDate], debounceFetchSearchResults)
 </script>
 
 <template>
@@ -257,8 +260,7 @@ const debounceFetchSearchResults = debounce(fetchNewSearchResult, 250)
 					<SearchBox ref="searchBox"
 						:value.sync="searchText"
 						:placeholder-text="t('spreed', 'Search messages …')"
-						:is-focused.sync="isFocused"
-						@input="debounceFetchSearchResults" />
+						:is-focused.sync="isFocused" />
 					<NcButton :pressed.sync="searchDetailsOpened"
 						:aria-label="t('spreed', 'Search options')"
 						:title="t('spreed', 'Search options')"
@@ -276,8 +278,7 @@ const debounceFetchSearchResults = debounce(fetchNewSearchResult, 250)
 							:placeholder="t('spreed', 'From User')"
 							user-select
 							:loading="!participantsInitialised"
-							:options="participants"
-							@update:modelValue="debounceFetchSearchResults" />
+							:options="participants" />
 						<div class="search-form__search-detail__date-picker-wrapper">
 							<NcDateTimePickerNative id="search-form__search-detail__date-picker--since"
 								v-model="sinceDate"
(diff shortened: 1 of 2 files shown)


src/components/RightSidebar/SearchMessages/SearchMessagesTab.vue
```diff
@@ -105,13 +108,60 @@ const onRouteChange = ({ from, to }: { from: Route, to: Route }): void => {
...
-		searchResults.value = []
-		searchCursor.value = 0
-		isSearchExhausted.value = false
+/**
+ * Fetch contacts list
+ * @param searchTerm - Search term to filter contacts
+ */
+async function fetchContacts(searchTerm: string = '') {
+	try {
+		const { data: { contacts } } = await getContacts(searchTerm)
```
nickvergessen, line 117: Ah, this is what Maksim had asked me. Hmm using contacts is quite bad for multiple reasons:
1. It's limited by share restrictions, so you might not even see everyone that is in the room
2. Guests, federated users, emails, … are not supported

I would stay with the current participants for now. As a follow up I would then as a replacement create a new API endpoint which provides the information that you need for all message authors instead?

Antreesy: Better to do it together with messages/search endpoint then, since UnifiedSearch provider is also blocking guests, federated users, e.t.c

DorraJaouad: I still think if we temporarily use contacts would be better than participants:
* It includes majority of participants and aligned of what's shown in Unified Search, thus covering the case where users leave the conversation (which is frequent)
* Although participants approach includes guests and federated users, the search endpoint doesn't accept them so not useful.

src/components/RightSidebar/SearchMessages/SearchMessagesTab.vue
```diff
@@ -49,6 +49,20 @@ const emit = defineEmits<{
 	(event: 'close'): void
 }>()
 
+const participantsInitialised = computed(() => store.getters.participantsInitialised(token.value))
```
Antreesy, line 53: Why move it? Or it's leftovers from old commit?

src/types/index.ts
```diff
@@ -372,3 +372,28 @@ export type SearchMessagePayload = {
 	limit?: number,
 	from?: string
 }
+
+// Contacts menu
+export interface ContactsMenuEntry {
```
Antreesy, line 377: Not needed anymore

src/components/RightSidebar/SearchMessages/SearchMessagesTab.vue
```diff
@@ -5,7 +5,7 @@
 
 <script setup lang="ts">
 import debounce from 'debounce'
-import { computed, nextTick, onBeforeUnmount, onMounted, ref, watch } from 'vue'
+import { computed, onBeforeUnmount, onMounted, ref, watch, nextTick } from 'vue'
```
Antreesy, line 8: ?

src/types/index.ts
DorraJaouad: I thought of keeping it for future use?

src/components/RightSidebar/SearchMessages/SearchMessagesTab.vue
DorraJaouad: Yes, and I thought of keeping it for better order, but I can revert if you want

src/components/RightSidebar/SearchMessages/SearchMessagesTab.vue
DorraJaouad: Maybe moved during rebase

src/types/index.ts
Antreesy: If we'll get a promised endpoint for user searching, it will likely be `TalkParticipant`, not server `ContactsMenuEntry`

src/components/RightSidebar/SearchMessages/SearchMessagesTab.vue
Antreesy: It was close to usual Options API order (props, setup, data, computed, watch, e.t.c.).
It is also now above `store` initialization, not sure it won't break it

We can move it later, when it's not dependant on stores, but internal logic with dedicated API

src/components/RightSidebar/SearchMessages/SearchMessagesTab.vue
```diff
@@ -242,18 +248,27 @@ async function fetchSearchResults(isNew = true): Promise<void> {
...
+
+watch([searchText, fromUser, sinceDate, untilDate], debounceFetchSearchResults)
+
+watch(searchText, (value) => {
+	if (value.trim().length === 0) {
+		searchResults.value = []
+		searchCursor.value = 0
+		isSearchExhausted.value = false
+	}
+})
```
Antreesy, line 260: https://github.com/nextcloud/spreed/pull/14143/files#diff-fb038377a7ad3d226f9f258218f106ce9f98fdcb58248ed084bab77c48ca2d6eR123-R129

image

Antreesy (approved):
The rest is good

Issue #14132, opened 2025-01-15
Migrate AdminSettings to Pinia


Antreesy:
Currently components in AdminSettings do not have a shared state, and data exchange could only be done with EventBus:
https://github.com/nextcloud/spreed/blob/fe2dd663b08fc296df882430b2ab3af7aae3ca2b/src/components/AdminSettings/SignalingServers.vue#L168-L173

### Expected behaviour
🍍 Shared Pinia store and gradual migrate of useful data to it (currently HPB settings, SIP settings)

nickvergessen:
Also sipbridge is referenced by another setting (soon, PR is there and awaiting merge)

PR #14149, opened 2025-01-17
fix(AdminSettings): move signaling state to the root

Antreesy:
### ☑️ Resolves

* Ref #14132

### 🚧 Tasks

- [x] Move current components to TS (2/17)
- [x] replace `EventBus.emit('signaling-servers-updated')`, with reactive state

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/AdminSettings/SignalingServers.vue
+++ b/src/components/AdminSettings/SignalingServers.vue
@@ -87,7 +87,6 @@ import NcPasswordField from '@nextcloud/vue/dist/Components/NcPasswordField.js'
 import SignalingServer from '../../components/AdminSettings/SignalingServer.vue'
 
 import { SIGNALING } from '../../constants.js'
-import { EventBus } from '../../services/EventBus.ts'
 import type { InitialState } from '../../types/index.ts'
 
 const isCacheConfigured = loadState('spreed', 'has_cache_configured')
@@ -180,7 +179,6 @@ function updateServers() {
 	}), {
 		success: () => {
 			showSuccess(t('spreed', 'High-performance backend settings saved'))
-			EventBus.emit('signaling-servers-updated', serversProxy.value)
 			loading.value = false
 		},
 	})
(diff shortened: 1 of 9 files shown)


src/env.d.ts
```diff
@@ -15,6 +15,12 @@ declare global {
 		webkitRequestFullscreen: RequestFullscreen;
 	}
 
+	const OCP: {
```
DorraJaouad, line 18: My IDE cannot find it :/

src/components/AdminSettings/SignalingServers.vue
```diff
@@ -87,100 +80,105 @@ import NcNoteCard from '@nextcloud/vue/dist/Components/NcNoteCard.js'
...
-		this.debounceUpdateServers = debounce(this.updateServers, 1000)
-		const state = loadState('spreed', 'signaling_servers')
-		this.servers = state.servers
-		this.secret = state.secret
-		this.hideWarning = state.hideWarning
-	},
+const props = defineProps<{
+	hideWarning: boolean,
+	secret: string,
+	servers: { server: string, verify: boolean }[],
```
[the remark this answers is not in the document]
Antreesy: I was trying to use `const props = defineProps<InitialState['spreed']['signaling_servers']>()`, but compile failed:
`[@vue/compiler-sfc] type argument passed to defineProps() must be a literal type, or a reference to an interface or literal type.`

Your suggestion is working though

src/env.d.ts
Antreesy: Working for me after modifying tsconfig.json:
`	"include": ["src/**/*.vue", "src/**/*.ts", "src/env.d.ts"],`

But we need to install back 'vue-tsc' package and check support of Vue files again (as we disabled it in 2a3c1b75565e54ae98e053dd974cc2030cdb8e4e) -> follow-up

PR #14168, opened 2025-01-21
fix(federation): Always return pending invitation header

SystemKeeper:
## 🛠️ API Checklist

We should also return the invitation header, when there are no invitations, to allow for quick removal of "Pending Invitations" message.

### 🏁 Checklist


SystemKeeper:
We decided that we don't want to do that now. If the header is not there on the endpoint, we assume 0 pending invites.

PR #14181, opened 2025-01-23
fix(desktop): change "reload page" messages to "restart app"

ShGKme:
### ☑️ Resolves

* On Talk Desktop it shouldn't say, "Reload the page". Instead, it must be "Restart the app".
* Module name might not be the best, but rushing before the string freeze to have the strings. Module/structure can be refactored afterwards.

image

image

image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/services/CapabilitiesManager.ts
+++ b/src/services/CapabilitiesManager.ts
@@ -11,6 +11,7 @@ import { getRemoteCapabilities } from './federationService.ts'
 import BrowserStorage from '../services/BrowserStorage.js'
 import { useTalkHashStore } from '../stores/talkHash.js'
 import type { Capabilities, Conversation, JoinRoomFullResponse } from '../types/index.ts'
+import { messagePleaseReload } from '../utils/talkDesktopUtils.ts'
 
 type Config = Capabilities['spreed']['config']
 type RemoteCapability = Capabilities & Partial<{ hash: string }>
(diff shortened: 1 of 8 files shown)


src/services/CapabilitiesManager.ts
```diff
@@ -136,7 +137,7 @@ export async function setRemoteCapabilities(joinRoomResponse: JoinRoomFullRespon
 	patchTokenMap(joinRoomResponse.data.ocs.data)
 
 	// As normal capabilities update, requires a reload to take effect
-	showError(t('spreed', 'Nextcloud Talk Federation was updated, please reload the page'), {
+	showError(t('spreed', 'Nextcloud Talk Federation was updated.') + '\n' + messagePleaseReload, {
```
nickvergessen, line 140: Something for later, but this should say something like:
> "Nextcloud Talk was updated on the federated server."

And we should have a look if we can make sure it does not yield when you open such a conversation the first time

nickvergessen (approved):

src/services/CapabilitiesManager.ts
nickvergessen: Extracted to https://github.com/nextcloud/spreed/issues/14201

Antreesy: Another point regarding `\n`: we discussed with Dorra, if toasts width should be limited upstream to ~750px (80 characters per line is recommended by W3C standard)?

PR #14204, opened 2025-01-23
feat(mentions): allow teams to be mentioned

### LINKED ISSUE #12036: Add ability to mention a team
Teams can be added and removed from conversations. It would be nice if we could also mention them.

miaulalala:
### ☑️ Resolves

* Ref #12036 

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/lib/Notification/Notifier.php
+++ b/lib/Notification/Notifier.php
@@ -8,6 +8,7 @@ declare(strict_types=1);
 
 namespace OCA\Talk\Notification;
 
+use OCA\Circles\CirclesManager;
 use OCA\FederatedFileSharing\AddressHandler;
 use OCA\Talk\AppInfo\Application;
 use OCA\Talk\Chat\ChatManager;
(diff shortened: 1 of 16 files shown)


lib/Notification/Notifier.php
```diff
@@ -774,6 +774,16 @@ protected function parseChatMessage(INotification $notification, Room $room, Par
 				];
 
 				$subject = $l->t('{user} mentioned group {group} in conversation {call}');
+			} elseif ($notification->getSubject() === 'mention_team') {
+				//				$circlesManager = OC::$server->getCircleManager();
+				//				$groupName = $this->groupManager->getDisplayName($subjectParameters['sourceId']) ?? $subjectParameters['sourceId'];
```
nickvergessen, line 779: Agreed on a follow up, to get translation strings in earlier

nickvergessen (approved):

lib/Notification/Notifier.php
nickvergessen: Actually done now

PR #14205, opened 2025-01-23
fix(MessagesList): ease list around lastReadMessage

Antreesy:
### ☑️ Resolves

* Fix following case:
  * Open chat with 300+ unread messages, poll them all, switch to another chat, switch back - last read is no longer found, user is at the bottom of the chat, need to load old messages again
* Drawbacks:
  * Need to poll all new messages again
    * [ ] should be fixed with message history (follow-up)
* Known issues:
  * after page reload first rich text will be re-rendered again (async lib load)
    * [ ] drop async load or make workaround (follow-up)

🏚️ Before 

https://github.com/user-attachments/assets/e13caf39-a8d6-4835-8ced-bbd967dea5c5

🏡 After

https://github.com/user-attachments/assets/b70f1722-9c4f-4b8b-8a25-026d9e0d7238

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -621,7 +621,7 @@ export default {
 			let isFocused = null
 			if (focusMessageId) {
 				// scroll to message in URL anchor
-				this.focusMessage(focusMessageId)
+				this.focusMessage(focusMessageId, false)
 				return
 			}
 
--- a/src/store/messagesStore.spec.js
+++ b/src/store/messagesStore.spec.js
@@ -258,35 +258,62 @@ describe('messagesStore', () => {
 		})
 	})
 
-	test('message list', () => {
-		const message1 = {
-			id: 1,
-			token: TOKEN,
-		}
-		const message2 = {
-			id: 2,
-			token: 'token-2',
-		}
-		const message3 = {
-			id: 3,
-			token: TOKEN,
-		}
+	describe('messages list', () => {
+		test('returns messages list', () => {
+			const message1 = {
+				id: 1,
+				token: TOKEN,
+			}
+			const message2 = {
+				id: 2,
+				token: 'token-2',
+			}
+			const message3 = {
+				id: 3,
+				token: TOKEN,
+			}
 
-		store.dispatch('processMessage', { token: message1.token, message: message1 })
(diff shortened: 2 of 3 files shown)


src/store/messagesStore.spec.js
```diff
@@ -258,35 +258,77 @@ describe('messagesStore', () => {
...
+					store.dispatch('processMessage', { token: TOKEN, message: { token: TOKEN, id } })
+				}
+				store.dispatch('setFirstKnownMessageId', { token: TOKEN, id: oldFirst })
+				store.dispatch('setLastKnownMessageId', { token: TOKEN, id: oldLast })
+
+				// Act
+				store.dispatch('easeMessageList', { token: TOKEN })
+
+				// Assert
+				expect(store.getters.messagesList(TOKEN)).toHaveLength(length)
```
DorraJaouad, line 326: Also test that lastReadMessage Id (if any) is still in the list

src/components/MessagesList/MessagesList.vue
```diff
@@ -621,7 +621,7 @@ export default {
 			let isFocused = null
 			if (focusMessageId) {
 				// scroll to message in URL anchor
-				this.focusMessage(focusMessageId)
+				this.focusMessage(focusMessageId, false)
```
DorraJaouad, line 624: Why? and I must say it's not related to the scope of PR :p

DorraJaouad (commented):
Logic seems good

src/components/MessagesList/MessagesList.vue
Antreesy: It only applies from `handleStartGettingMessagesPreconditions`, when we see a chat first time (or after conversation switch. Please try this:
- make a long chat, with 300+ unread messages (let's say lastRead is id_300)
- copy link of message id_330, paste into another chat and click the link

For me it looks buggy, chat opened, then scrolled, rather then open directly on message from URL

DorraJaouad: So you want to keep the smooth scroll only when you focus on a message inside the same conversation (like clicking on quote reply)? That's fine by me.

DorraJaouad (approved):
Tested 🦅

PR #14209, opened 2025-01-24
fix: allow to create public room without password if password_policy enabled

Antreesy:
### ☑️ Resolves

* Follow-up to #13944

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist


provokateurin (commented):
Can confirm it's broken and this change fixes it :+1:

Antreesy:
We discussed internally, that password_policy might not be considered in CI tests (or it's only this specific case), so any help with covering it is appreciated

nickvergessen (approved):
Makes sense, but would be good to add a CI test for it.
We can do that in a follow up as well

PR #14232, opened 2025-01-27
fix(CalendarEventsDialog): keep components in one node

Antreesy:
### ☑️ Resolves

* Fix insertBefore method:
  * Each inner dialog is replaced with comments in DOM, if v-if=false
  * Now they're under a single template

To test:
- Open calendarevents dialog, then close
- Switch to conversation, with no moderator permissions (e.g. Changelog)

🏚️ Before | 🏡 After
-- | --
image | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/CalendarEventsDialog.vue
+++ b/src/components/CalendarEventsDialog.vue
@@ -350,126 +350,127 @@ async function submitNewMeeting() {
 			</template>
 		</NcPopover>
 
-		<NcDialog v-if="canScheduleMeeting"
-			id="calendar-meeting"
-			:open.sync="isFormOpen"
-			class="calendar-meeting"
-			:name="t('spreed', 'Schedule a meeting')"
-			size="normal"
-			close-on-click-outside
-			:container="container">
-			<NcTextField v-model="newMeetingTitle"
-				:label="t('spreed', 'Meeting title')"
-				label-visible />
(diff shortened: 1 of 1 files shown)


src/components/CalendarEventsDialog.vue
```diff
@@ -350,126 +350,127 @@ async function submitNewMeeting() {
...
+						</template>
+						{{ t('spreed', 'Save') }}
+					</NcButton>
 				</template>
-			</NcEmptyContent>
-			<template #actions>
-				<NcButton type="primary" @click="isSelectorOpen = false">
+			</NcDialog>
+
+			<NcDialog v-if="isSelectorOpen"
```
DorraJaouad, line 427: I tried without this condition and it works

DorraJaouad (approved):
There is also a focus-trap issue when there is an empty view (no tabbable element) and maybe in loading?
`:focus-trap="loading || upcomingEvents.length !== 0"` This solves when there is an empty view

src/components/CalendarEventsDialog.vue
DorraJaouad: I didn't mean to approve :')

Antreesy: Good catch anyway, thanks!

PR #14255, opened 2025-01-30
fix(Screen): change double-click behaviour for screenshare panzoom

### LINKED ISSUE #14243: Zoom on screenshare with double click is non intuitive
### How to use GitHub

---

## Feature request

* Receive a screenshare
* See that you can zoom in, because the cursor changed to a magnifying glass, via double click
* Double click again -> zooms again

There's no indication that scrolling can be used to scroll out again.

Questions:
* Should zooming via double click work only once and the next double click zooms out again? (Idea by nickvergessen)
* Should there be a button "Reset zoom" or the like?

Antreesy:
### ☑️ Resolves

* Fix #14243
  * 👀 Zoom factor is 4, which is the middle between min/max boundaries
  * 🧪 First double-click zooms in to 4x scale, second (or first one, if you already scaled) - resets to default

https://github.com/user-attachments/assets/e15f32c3-5b11-463a-be28-f7a2be8252d2

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/CallView/shared/Screen.vue
+++ b/src/components/CallView/shared/Screen.vue
@@ -4,7 +4,9 @@
 -->
 
 <template>
-	<div :id="screenContainerId" class="screenContainer">
+	<div :id="screenContainerId"
+		class="screenContainer"
+		@dblclick.capture="onDoubleClick">
 		<video v-show="(localMediaModel && localMediaModel.attributes.localScreen) || (callParticipantModel && callParticipantModel.attributes.screen)"
 			ref="screen"
 			:disablePictureInPicture="!isBig ? 'true' : 'false'"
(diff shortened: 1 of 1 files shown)


DorraJaouad:
> ⚠️ There is no option to rewrite upstream doubleClick beaviour, so it need to be disabled (with zoomDoubleClickSpeed: 1)

I think you can do so, by making a custom trigger (listen to click event)
https://github.com/anvaka/panzoom#custom-ui-to-trigger-zoom

> Smooth zoom doesn't work for some reason, if you call it manually, so normal zoom is used

Smooth is not available for absZoom but it is available in `smoothZoom`
https://github.com/anvaka/panzoom#triggering-zoom

DorraJaouad (approved):
Acceptable

PR #14270, opened 2025-01-31
feat(chat): Add the mention id to parameters for easier editing

nickvergessen:
### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/lib/Chat/Parser/SystemMessage.php
+++ b/lib/Chat/Parser/SystemMessage.php
@@ -924,6 +924,7 @@ class SystemMessage implements IEventListener {
 			'type' => 'user',
 			'id' => $uid,
 			'name' => $this->displayNames[$uid],
+			'mention-id' => $uid,
 		];
 	}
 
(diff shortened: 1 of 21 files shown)


SystemKeeper:
Is there any reason to _not_ always wrap a mention in quotes? Currently we only do it if there's a "/" or whitespace, but does it really matter?

nickvergessen:
> Is there any reason to not always wrap a mention in quotes? Currently we only do it if there's a "/" or whitespace, but does it really matter?

IIRC it works always so its save to do it always, but e.g. on our instance since we use "first name", I always type it directly manually without waiting for the autocomplete

SystemKeeper:
> when proxying such a message I guess the server should extend the attribute if it was missing?

Would be nice, if it’s not too much work. Then it consistent across all clients.

nickvergessen:
Can't add it to OpenAPI docs as the type then conflicts with the RichObject Definitions and Psalm complains

SystemKeeper (approved):

SystemKeeper:
Can we backport to stable30 and stable29 as well, since we introduced message editing with 19/29? This would allow for more consistent editing and testing.

PR #14327, opened 2025-02-10
fix(MessagesList): fix resizeObserver and scrolledToBottom state update

### LINKED ISSUE #14392: Typing a message invokes the scroll indicator, and prevents the other party's next message from auto scrolling into view
In a Talk chat, scrolling up invokes the double-down-chevron indicator, to show that we are not at the bottom of the chat. Any new messages from the other party will not be auto scrolled into view until I scroll down or click the indicator to go back to the bottom. This is as expected.

However, often the indicator appears as a result of me typing my own message. Everything (so far) is still in view, but the chat now thinks it's not fully scrolled - so won't auto scroll a reply from the other party into view. When this happens, I need to either scroll up+down, or click the indicator, to ensure the next reply appears. It's easy to miss this and thus miss a reply.

This seems random, I can't pin it down to anything. Reproducible on 2 different NC installs, and in Firefox and Chromium.

## Steps to reproduce

See example video - **skip to 3:00**:

https://github.com/user-attachments/assets/e8a3e201-e0d0-4247-b099-e3a2a42c0c25

### Expected behaviour
New messages from the other party should automatically scroll into view

### Actual behaviour
But they don't because my own message triggered the scroll indicator

## Talk app

**Talk app version:** 20.1.4
**Custom Signaling server configured:** no

Antreesy:
### ☑️ Resolves

* Follow-up to #14266
  * Keeps node reference for ResizeObserver after switching conversations
  * Ensure non-scrollable chat doesn't have 'scroll to bottom' button
* Fix #14392 🚧
  * `setChatScrolledToBottom(false)` should be only set on scroll up (if we're scrolling down. it should be already set)

🏡 After

https://github.com/user-attachments/assets/0cdcac86-e967-43e4-a894-94f5b5861c86

### 🔬  Test
- [x] removing `:key`:
  - [x] ResizeObserver works after conversation switch
  - [x] nothing is broken, if we get rid of 'force-refresh' of whole component
- [x] non-scrollable chats do not have 🔽 button:
  - [x] on open (on open with focusing message)
  - [x] on resize
  - [x] on messages list (height) change
- [x] scrollable chats do not have 🔽 button:
  - [x] when scrolling down
  - [x] when nothing to scroll
  - [x] when new message arrives, and there's a mismatch of last read (follow-up from above)

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -300,11 +300,14 @@ export default {
 				// scroll to bottom if needed
 				this.scrollToBottom({ smooth: false })
 
-				if (this.conversation?.type === CONVERSATION.TYPE.NOTE_TO_SELF) {
-					this.$nextTick(() => {
+				this.$nextTick(() => {
+					this.checkChatNotScrollable()
+
+					if (this.conversation?.type === CONVERSATION.TYPE.NOTE_TO_SELF) {
 						this.updateTasksCount()
-					})
-				}
+					}
+				})
+
 			},
 		},
 
@@ -320,8 +323,7 @@ export default {
 				this.$nextTick(() => {
 					this.checkSticky()
 					// setting wheel event for non-scrollable chat
-					const isScrollable = this.$refs.scroller.scrollHeight > this.$refs.scroller.clientHeight
-					if (!this.isChatBeginningReached && !isScrollable) {
+					if (!this.isChatBeginningReached && this.checkChatNotScrollable()) {
 						this.$refs.scroller.addEventListener('wheel', this.handleWheelEvent, { passive: true })
 					}
 				})
@@ -385,6 +387,8 @@ export default {
 				this.$refs.scroller.scrollTo({
 					top: this.$refs.scroller.scrollHeight,
 				})
+			} else {
+				this.checkChatNotScrollable()
 			}
 		},
 
@@ -1161,10 +1165,7 @@ export default {
 				this.$refs.scroller.scrollTop += this.$refs.scroller.offsetHeight / 4
 			}
 
-			if (this.$refs.scroller && this.$refs.scroller.clientHeight === this.$refs.scroller.scrollHeight) {
-				// chat is not scrollable
-				this.setChatScrolledToBottom(true)
-			}
+			this.checkChatNotScrollable()
 
 			if (highlightAnimation && scrollElement === element) {
 				// element is visible, highlight it
@@ -1284,6 +1285,18 @@ export default {
 			const tasksDoneCount = this.$refs.scroller.querySelectorAll('.checkbox-content__icon--checked')?.length
 			const tasksCount = this.$refs.scroller.querySelectorAll('.task-list-item')?.length
 			this.chatExtrasStore.setTasksCounters({ tasksCount, tasksDoneCount })
+		},
+
+		checkChatNotScrollable() {
+			const isNotScrollable = this.$refs.scroller
+				? this.$refs.scroller.clientHeight === this.$refs.scroller.scrollHeight
+				: false
+
+			if (isNotScrollable && !this.isChatScrolledToBottom) {
+				this.setChatScrolledToBottom(true)
+			}
+
+			return isNotScrollable
 		},
 
 		handleWheelEvent(event) {
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -7,7 +7,6 @@
 	<!-- size and remain refer to the amount and initial height of the items that
 	are outside of the viewport -->
 	<div ref="scroller"
-		:key="token"
 		class="scroller messages-list__scroller"
 		:class="{'scroller--chatScrolledToBottom': isChatScrolledToBottom,
 			'scroller--isScrolling': isScrolling}"
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -898,17 +898,20 @@ export default {
 			}
 
 			const { scrollHeight, scrollTop, clientHeight } = this.$refs.scroller
-			const scrollOffset = scrollHeight - scrollTop
+			const scrollOffsetFromTop = scrollHeight - scrollTop
+			const scrollOffsetFromBottom = Math.abs(scrollOffsetFromTop - clientHeight)
 
 			// For chats that are scrolled to bottom and not fitted in one screen
-			if (Math.abs(scrollOffset - clientHeight) < SCROLL_TOLERANCE && !this.hasMoreMessagesToLoad && scrollTop > 0) {
+			if (scrollOffsetFromBottom < SCROLL_TOLERANCE && !this.hasMoreMessagesToLoad && scrollTop > 0) {
 				this.setChatScrolledToBottom(true)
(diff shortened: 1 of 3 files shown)


src/components/MessagesList/MessagesList.vue
```diff
@@ -7,7 +7,6 @@
 	<!-- size and remain refer to the amount and initial height of the items that
 	are outside of the viewport -->
 	<div ref="scroller"
-		:key="token"
```
Antreesy, line 10: Was added during messages list soft update refactoring, but now it seems to be covered with `messagesList` watcher
With key in place, ResizeObserver loses reference to the scroller

src/components/MessagesList/MessagesList.vue
```diff
@@ -387,6 +386,9 @@ export default {
 				this.$refs.scroller.scrollTo({
 					top: this.$refs.scroller.scrollHeight,
 				})
+			} else if (this.$refs.scroller.clientHeight === this.$refs.scroller.scrollHeight) {
+				// chat is not scrollable
+				this.setChatScrolledToBottom(true)
```
Antreesy, line 391: For non-scrollable chats, removes the scroll to bottom on resize

src/components/MessagesList/MessagesList.vue
```diff
@@ -931,10 +933,10 @@ export default {
 				this.displayMessagesLoader = false
 				this.debounceUpdateReadMarkerPosition()
 				return
+			} else if (scrollHeight > clientHeight) {
+				this.setChatScrolledToBottom(false)
```
Antreesy, line 937: After swtiching to conversation, handleScroll is triggered. Regardless of container height, it sets value to false (whereas it shouldn't for non-scrollable chats)

src/components/MessagesList/MessagesList.vue
```diff
@@ -932,24 +934,25 @@ export default {
 				return
 			}
 
-			this.setChatScrolledToBottom(false)
```
Antreesy, line 935: After swtiching to conversation, handleScroll is triggered. Regardless of container height, it sets value to false (whereas it shouldn't for non-scrollable chats)
For scrollable chats it also shouldn't set to false, when receiiving message (handleScroll is triggered with 'down' direction)

src/components/MessagesList/MessagesList.vue
DorraJaouad: fine

src/components/MessagesList/MessagesList.vue
DorraJaouad: Maybe we need `nextTick` here? edge case:
* Have one long message in a conversation (the scroller is scrollable)
* Delete that message (no Longer scrollable)
* You will have `chatScrolledToBottom` truthy

DorraJaouad (commented):
Sthg is wrong, chat is sticky when document is not focused (hidden) and it scrolls to bottom ⚠️

src/components/MessagesList/MessagesList.vue
Antreesy: added to messageList watcher

src/components/MessagesList/MessagesList.vue
```diff
@@ -301,6 +300,10 @@ export default {
 				// scroll to bottom if needed
 				this.scrollToBottom({ smooth: false })
 
+				this.$nextTick(() => {
+					this.checkChatNotScrollable()
+				})
+
 				if (this.conversation?.type === CONVERSATION.TYPE.NOTE_TO_SELF) {
 					this.$nextTick(() => {
 						this.updateTasksCount()
```
DorraJaouad, line 309: Wrap them together?

src/components/MessagesList/MessagesList.vue
```diff
@@ -932,7 +936,9 @@ export default {
 				return
 			}
 
-			this.setChatScrolledToBottom(false)
+			if (scrollHeight > clientHeight && this.isScrolling === 'up') {
+				this.setChatScrolledToBottom(false)
+			}
```
DorraJaouad, line 941: Now, when there are unread messages, it scrolls correctly but `isChatScrolledToBottom` is truthy (hence isSticky) and clears all unread messages.

DorraJaouad: The scroll to bottom button is missing and it reads all of the chat at once (without even scrolling).
For the second part, it might sound okay but it's a new behaviour that we should agree on, no?

https://github.com/user-attachments/assets/54b39fce-4496-43c4-b3ff-942658bc7539

Antreesy: This is incorrect, let me reproduce and fix it

Antreesy: Was a false-positive check for `isScrolling`

src/store/messagesStore.js
```diff
@@ -1205,6 +1210,14 @@ const actions = {
 			}
 		}
 
+		if (conversation && lastMessageFromSelf?.id > conversation.lastReadMessage) {
+			context.dispatch('updateLastReadMessage', {
+				token,
+				id: lastMessageFromSelf.id,
+				updateVisually: true,
+			})
+		}
```
DorraJaouad, line 1219: I am afraid it should not be on store level but rather on message posting level as it kills mark a particular message X as unread functionality when you have already posted a message Y such that X.id < Y.id.

Antreesy: Ah, right
Hm, maybe it could be done then we're replacing temp message

src/store/messagesStore.js
```diff
@@ -586,9 +586,21 @@ const actions = {
 
 		if (message.referenceId) {
 			const tempMessages = context.getters.getTemporaryReferences(token, message.referenceId)
-			tempMessages.forEach(tempMessage => {
-				context.commit('deleteMessage', { token, id: tempMessage.id })
-			})
+			if (tempMessages.length > 0) {
```
DorraJaouad, line 589: `// If successful, deletes the temporary message from the store
			context.dispatch('removeTemporaryMessageFromStore', { token, id: temporaryMessage.id })`

temp message is removed before processing the message is `postNewMessage`

Also, `processMessage` in multiple points, I wouldn't risk updating the last read message there. As suggested before, it should rather focus on post new message only. What's odd is that the current addition already exists there but it wasn't working efficiently

https://github.com/nextcloud/spreed/blob/main/src/store/messagesStore.js#L1276-L1291

Antreesy: it exists and works for text messages only. File share (with /without caption) is a one-way, a we will receive a feedback only from message polling

postNewMessage can do it, as we receive a message object as a response and can replace it safely

Antreesy: As discussed, moved logic from postNewMessage to processMessage, to not duplicate it and also cover file shares

src/components/MessagesList/MessagesList.vue
Antreesy: Modify to set 'false' only if we're above 10px from bottom. So if we're sticky, that shouldn't cause button appearance

DorraJaouad:
Time has come 🦅 
Tested ✅

DorraJaouad (approved):

PR #14344, opened 2025-02-11
fix(chat): Allow to edit messages of bots in one-to-one conversations

nickvergessen:
### 🏁 Checklist


nickvergessen:
Do we need a capability and explicit documentation, or should we just assume it is a bug fix and if editing fails because the client was already updated but the server not, we don't care?

SystemKeeper:
> Do we need a capability and explicit documentation, or should we just assume it is a bug fix and if editing fails because the client was already updated but the server not, we don't care?

We need to change it on client side as well:

https://github.com/nextcloud/talk-ios/blob/fc5a9ebda821f257dddc73ec7169fcbc9f544571/NextcloudTalk/NCChatMessage.swift#L80-L83

But I would say we just enable it server-side and don't care if it wasn't updated.

Antreesy (approved):
> Do we need a capability and explicit documentation

'edit-messages' should be enough, if we backport it to 29?

PR #14354, opened 2025-02-11
fix(teams): Track team name updates

nickvergessen:
### ☑️ Resolves

* Add a team to a conversation
* Rename the team
* Check name in chat where it was added or mentioned :checkered_flag: 
* Check name in participant list, before :boom: after :checkered_flag: 

### 🏁 Checklist

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -10,6 +10,8 @@ namespace OCA\Talk\AppInfo;
 
 use OCA\Circles\Events\AddingCircleMemberEvent;
 use OCA\Circles\Events\CircleDestroyedEvent;
+use OCA\Circles\Events\CircleEditedEvent;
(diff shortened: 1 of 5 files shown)


lib/AppInfo/Application.php
```diff
@@ -10,6 +10,8 @@
 
 use OCA\Circles\Events\AddingCircleMemberEvent;
 use OCA\Circles\Events\CircleDestroyedEvent;
+use OCA\Circles\Events\CircleEditedEvent;
```
Antreesy, line 13: teams/circles edited before that patch don't seem to be updated afterwards? Is it critical, or shoudl we somehow trigger migration step to re-check existing mentions and participants?

nickvergessen: I don't plan to fix that. Renames are not that often. It can be fixed by renaming twice

Antreesy (approved):

PR #14361, opened 2025-02-12
fix(chat): Correctly limit editing to bots IN 1-1

nickvergessen:
- Regression from #14344 

Noticed on the backport to 29 that I forgot the "IN one-to-one" part.

### DIFF
--- a/src/composables/useMessageInfo.js
+++ b/src/composables/useMessageInfo.js
@@ -61,6 +61,8 @@ export function useMessageInfo(message = ref({})) {
 	const isBotInOneToOne = computed(() =>
 		message.value.actorId.startsWith(ATTENDEE.BOT_PREFIX)
 		&& message.value.actorType === ATTENDEE.ACTOR_TYPE.BOTS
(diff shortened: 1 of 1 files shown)


src/composables/useMessageInfo.js
```diff
@@ -61,6 +61,8 @@ export function useMessageInfo(message = ref({})) {
 	const isBotInOneToOne = computed(() =>
 		message.value.actorId.startsWith(ATTENDEE.BOT_PREFIX)
 		&& message.value.actorType === ATTENDEE.ACTOR_TYPE.BOTS
+		&& (conversation.value.type === CONVERSATION.TYPE.ONE_TO_ONE
+			|| conversation.value.type === CONVERSATION.TYPE.ONE_TO_ONE_FORMER)
```
Antreesy, line 65: Would be better to maintain
```suggestion
		&& isOneToOneConversation.value
```

Antreesy (approved):

src/composables/useMessageInfo.js
nickvergessen: I didn't find it

PR #14369, opened 2025-02-13
fix(PresenterOverlay): do not move component to negative values

Antreesy:
### ☑️ Resolves

* Fix placing presenter video out of visible bounds while screensharing
* Parent size is recalculated after mounting, which results in smaller parentHeight than actual -> moveVertically pushes it behind the CallView

🏚️ Before 

https://github.com/user-attachments/assets/4af0ca81-f779-424e-8d06-2aaf7cfa16df

🏡 After

https://github.com/user-attachments/assets/d1e54469-94d5-4f63-8a89-1db9800c59eb

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/CallView/shared/PresenterOverlay.vue
+++ b/src/components/CallView/shared/PresenterOverlay.vue
@@ -146,10 +146,10 @@ export default {
 			this.$refs.presenterOverlay.checkParentSize()
 			// FIXME: if it stays out of bounds (right and bottom), bring it back
 			// FIXME: should consider RTL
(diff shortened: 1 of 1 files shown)


Antreesy:
> Is it guaranteed that the value is initially (before `update` trigger) are in the valid range?

I don't think so. Caught it in the moment of transition (Grid component), believe, that issue might originate from here:
image

nickvergessen (approved):

PR #14393, opened 2025-02-14
fix(MessagesList): rename method to pollNewMessages

Antreesy:
### ☑️ Resolves

- Non-breaking pre-requisite to #10084
  - I'm giving up and will implement it by parts 👀 
- rename `getNewMessages` and `lookForNewMessages` to `pollNewMessages` (to avoid confusion - get*Messages should not poll)
- drop `lookForNewMessages` component method as redundant
  - `_isBeingDestroyed`, `_isDestroyed` are internal Vue flags, different in Vue3, and not needed here
- refactor `fetchMessages` to support both direction of `lookIntoFuture` (to make fetching the chunk of **new** messages possible)
- refactor tests

No visual changes

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -273,7 +273,7 @@ export default {
 			immediate: true,
 			handler(newValue, oldValue) {
 				if (oldValue) {
-					this.$store.dispatch('cancelLookForNewMessages', { requestId: oldValue })
+					this.$store.dispatch('cancelPollNewMessages', { requestId: oldValue })
 				}
 				this.handleStartGettingMessagesPreconditions(this.token)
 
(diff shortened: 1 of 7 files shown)


src/components/MessagesList/MessagesList.vue
```diff
@@ -707,29 +707,14 @@ export default {
...
 				this.$store.dispatch('cancelLookForNewMessages', { requestId: this.chatIdentifier })
 			}
 		},
 
-		/**
-		 * Fetches the messages of a conversation given the conversation token. Triggers
-		 * a long-polling request for new messages.
-		 * @param token token of conversation where a method was called
-		 */
-		async lookForNewMessages(token) {
```
DorraJaouad, line 723: Better to fully replace `lookForNewMessages` in this component  (log text and the boolean), because what you replaced is just a wrapper for the poll request (processing its responses)
and thus change in store?

src/store/messagesStore.js
```diff
@@ -1021,6 +1036,7 @@ const actions = {
 				token,
 				lastKnownMessageId: context.getters.getFirstKnownMessageId(token),
 				includeLastKnown: false,
+				lookIntoFuture: CHAT.FETCH_OLD,
```
DorraJaouad, line 1039: It's already default to old or is it for readability purposes?

Antreesy: Better to explicitly set it here, so we know - if context doesn't have enough messages, get some more old messages

src/components/MessagesList/MessagesList.vue
Antreesy: Renamed all occurences.
Although it's still splitted between files:
- message fetch and processing are in store
- but actual polling intervals/timeouts are in component

DorraJaouad (approved):
Tested

Issue #12803, opened 2024-07-24
Typing indicator privacy is always set to private for federated rooms


SystemKeeper:
See https://github.com/nextcloud/spreed/blob/8d68d25ba05c8d34532679269696eca1f28fd2b2/lib/Controller/RoomController.php#L2343-L2345

With federated external signaling, it is now possible to federate these as well.

PR #14443, opened 2025-02-17
feat(conversation): add federated typing indicator

miaulalala:
### ☑️ Resolves

* Fix https://github.com/nextcloud/spreed/issues/12803

### 🚧 Tasks

- [x] ...

### 🏁 Checklist

### DIFF
--- a/lib/Federation/Proxy/TalkV1/ProxyRequest.php
+++ b/lib/Federation/Proxy/TalkV1/ProxyRequest.php
@@ -12,8 +12,10 @@ namespace OCA\Talk\Federation\Proxy\TalkV1;
 use GuzzleHttp\Exception\ClientException;
 use GuzzleHttp\Exception\ServerException;
 use OC\Http\Client\Response;
+use OCA\Talk\AppInfo\Application;
 use OCA\Talk\Exceptions\CannotReachRemoteException;
 use OCA\Talk\Exceptions\RemoteClientException;
+use OCA\Talk\Participant;
 use OCP\AppFramework\Http;
 use OCP\Http\Client\IClientService;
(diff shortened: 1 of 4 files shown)


lib/Federation/Proxy/TalkV1/ProxyRequest.php
```diff
@@ -40,7 +40,6 @@ public function overwrittenRemoteTalkHash(string $hash): string {
 				'config' => [
 					'chat' => [
 						'read-privacy',
-						'typing-privacy',
```
nickvergessen, line 43: it's still overwritten, just not with a hardcoded value anymore, but the value might not be what the host send, so we should keep that.
In fact we might want to add the value in here, so when you change your setting the ones of the room also change?

nickvergessen: > In fact we might want to add the value in here, so when you change your setting the ones of the room also change?

I think we should do this? Otherwise remote capabilities are not refreshed, when you change your typing privacy?

nickvergessen: yeah, now we could extend the integration test:
Make a second copy of https://github.com/nextcloud/spreed/blob/86e13f9233e968fd7770a21b3057b0564aae0644/tests/integration/features/bootstrap/FeatureContext.php#L3803-L3834 which calls https://github.com/nextcloud/spreed/blob/9c3a8012c2864c5fb1710a039b92e631a98e9cc8/appinfo/routes/routesRoomController.php#L112-L113 to check that the capability also changed in the federated room and your overwrite worked

nickvergessen (approved):

PR #14444, opened 2025-02-17
tests: Make federation tests do real federation

nickvergessen:
Progress almost completed, only invite.feature is missing after line 107


miaulalala (approved):
:elephant:

PR #14445, opened 2025-02-18
fix(conversations): remove Delete conversation from quick actions conversation menu

ShGKme:
### ☑️ Resolves

* There is a "Delete conversation" in the conversation list quick actions menu
* Unlike leaving, it is a very destructive operation
* IMO, it shouldn't be available in a conversation menu
* Even with the confirmation dialog, a user may just quickly click "Confirm" without reading
  * There is a similar dialog on leaving which is not so dangerous 
* To reduce the risk of accidental deletion, I'd propose removing it from the close menu
* A user can still remove conversation from the settings, where it is more obviously a destructive operation

Alternative - having double confirmation.

🏚️ Before | 🏡 After
-- | --
image | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/LeftSidebar/ConversationsList/Conversation.vue
+++ b/src/components/LeftSidebar/ConversationsList/Conversation.vue
@@ -105,17 +105,6 @@
 					</template>
 					{{ t('spreed', 'Leave conversation') }}
 				</NcActionButton>
-
-				<NcActionButton v-if="item.canDeleteConversation"
-					key="delete-conversation"
-					close-after-click
-					class="critical"
-					@click="isDeleteDialogOpen = true">
-					<template #icon>
-						<IconDelete :size="16" />
-					</template>
-					{{ t('spreed', 'Delete conversation') }}
-				</NcActionButton>
 			</template>
 			<template v-else-if="submenu === 'notifications'">
 				<NcActionButton :aria-label="t('spreed', 'Back')"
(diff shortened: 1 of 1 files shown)


nickvergessen:
> Alternative - having double confirmation.

We had some thoughts about that in another call. It e.g. would make sense to ask for further confirmation if the conversation is:
- Having many chats (ours had 25k)
- Having many shares (ours had 1.3k)
- Is Old but still Current (first message from 2019, latest from 2025)

nimishavijay:
> Instead of removing it, we could open the "Danger zone" area when click on Delete?

Works for me, also makes it different from the "leave conversation" confirmation.

Antreesy:
Something like this should work then:

Conversation.vue:
```
emit('show-conversation-settings', { token: this.item.token, sectionId: 'dangerzone' })
```

ConversationSettingsDialog.vue:
```
handleShowSettings({ token, sectionId }) {
	...
	if (sectionId) {
		this.$nextTick(() => {
			document.getElementById(`settings-section_${sectionId}`).scrollIntoView()
		})
	}
}
```

ShGKme:
> Something like this should work then

It should move the focus first, otherwise the user is moved back on keyboard navigation

> document.getElementById(`settings-section_${sectionId}`)

It is not a public API, can be broken on minor update

Antreesy:
> can be broken on minor update

I just meant the approach. We can scroll to the Talk-defined header we know will exist in this place: `DangerZone` -> `<h4>Delete conversation</h4>`

nickvergessen:
Something for 21.1?

PR #14488, opened 2025-02-23
fix(update/openapi): do not skip PR creating on workflow

Antreesy:
### ☑️ Resolves

image
image

### DIFF
--- a/.github/workflows/update-nextcloud-openapi.yml
+++ b/.github/workflows/update-nextcloud-openapi.yml
@@ -32,13 +32,16 @@ jobs:
           repository: nextcloud/server
           ref: ${{ matrix.server-versions }}
 
-      - name: Checkout app
+      - id: checkout
         uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
         with:
+          persist-credentials: false
           path: apps/${{ env.APP_NAME }}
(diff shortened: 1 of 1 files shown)


.github/workflows/update-nextcloud-openapi.yml
```diff
@@ -32,28 +32,34 @@ jobs:
...
           ref: ${{ matrix.server-versions }}
 
-      - name: Checkout app
+      - id: checkout
         uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
         with:
+          persist-credentials: false
           path: apps/${{ env.APP_NAME }}
           ref: ${{ matrix.branches }}
+        continue-on-error: true
```
nickvergessen, line 41: Instead remove branch master from initial list?

Antreesy: Also should work

nickvergessen: Ah, now I see. We do the same on update-nextcloud-ocp, so it can be upstreamed and used by either apps.

nickvergessen (approved):

nickvergessen:
Cron jobs only run from the default branch. That's why it mentions stable31 in the action.

Only the scripts would need backporting

PR #14489, opened 2025-02-23
[l10n]: Change apostrophe

Valdnet:
Corrected the apostrophe.

### DIFF
--- a/lib/Service/SampleConversationsService.php
+++ b/lib/Service/SampleConversationsService.php
@@ -78,7 +78,7 @@ class SampleConversationsService {
 	protected function defaultSampleConversation(IUser $user): void {
 		$room = $this->roomService->createConversation(
 			Room::TYPE_GROUP,
-			$this->l->t('Let´s get started!'),
+			$this->l->t('Let\'s get started!'),
 			$user,
 			Room::OBJECT_TYPE_SAMPLE,
 			$user->getUID()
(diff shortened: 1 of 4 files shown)


lib/Service/SampleConversationsService.php
```diff
@@ -78,7 +78,7 @@ public function initialCreateSamples(string $userId): void {
 	protected function defaultSampleConversation(IUser $user): void {
 		$room = $this->roomService->createConversation(
 			Room::TYPE_GROUP,
-			$this->l->t('Let´s get started!'),
+			$this->l->t('Let's get started!'),
```
Antreesy, line 81: This breaks code parsing, as you're closing the line. Try to escape a character instead:
```suggestion
			$this->l->t('Let\'s get started!'),
```

Valdnet: @Antreesy Thanks. I forgot about that.

nickvergessen:
Scheduling backport to 31, but should only merge after the final, so translations can be updated in time.

nickvergessen:
Yeah migrated all usages, so it broke when testing federation with another version.
Filtering both fixes it.

nickvergessen:
Counter direction has to be fixed in stable31, so let's wait with merging until tomorrow and then also merge the backport right away

lib/Service/SampleConversationsService.php
```diff
@@ -114,7 +114,7 @@ protected function defaultSampleConversation(IUser $user): void {
 			. "\n\n" . '{FILE:Readme.md}',
 			$this->l->t('## 💭 Let the conversations flow: mention users, react to messages and more
 
-You can mention everybody in a the conversation @all or mention specific participants by typing "@" and picking their name from the list.')
+You can mention everybody in a the conversation "@all" or mention specific participants by typing "@" and picking their name from the list.')
```
nickvergessen, line 117: This was intentional to should the mention chip. If it's a problem to translations I will make it a parameter
```suggestion
You can mention everybody in a the conversation @all or mention specific participants by typing "@" and picking their name from the list.')
```

PR #14494, opened 2025-02-25
fix(Participant): unexpected change of participant to be removed and users can ban themself

Antreesy:
### ☑️ Resolves

fix multiple issues with participants virual list
- detect self with immediate data (do not rely on sessionId, as its update could be delayed, and client does not know if you is you)
- rerender virtual list when switch conversations (Recycle scroller is keeping items in the poll and reuse them, which may differ and interfere with actual list of participants from another conv)
- reset initial state of component if reused (same as above - Removed user from the list may be replaced by any other user from poll (usually first in the list) with open dialog and possibility to remove self)

To reproduce:
1. Open conversation 1
2. Switch to conversation 2
3. Open a dialog for deleting participant from conv2
4. Click 'Back' in browser, to return to previous route
-> dialog should dissapear if re-rendering the virtual list

To reproduce:
1. Open conversation 1
2. Open conversation 1 as a guest
3. Open a dialog for deleting guest from conv1
4. Reload (or leave) a page as a guest, to remove attendee from the list
-> dialog should dissapear if re-rendering the virtual item

🏚️ Before | 🏡 After
-- | --
fix(Participant): detect self with immediate data | --
2025-02-25_11h07_34 | 2025-02-25_11h09_41
fix(ParticipantsTab): rerender virtual list when switch conversations | --
2025-02-25_11h13_53 | 2025-02-25_11h15_19
fix(Participant): reset initial state of component if reused | --
2025-02-25_11h12_35 | 2025-02-25_11h11_04

### 🚧 Tasks

- [ ] Possible drawbacks?
  - [ ] RecycleScroller doesn't work well with constantly changing lists. Maybe, go with own solution?

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/RightSidebar/Participants/Participant.vue
+++ b/src/components/RightSidebar/Participants/Participant.vue
@@ -4,8 +4,7 @@
 -->
 
 <template>
-	<NcListItem :key="participantNavigationId"
-		:name="computedName"
+	<NcListItem :name="computedName"
 		:data-nav-id="participantNavigationId"
 		class="participant"
 		:class="{ 'participant--offline': isOffline }"
@@ -267,7 +266,7 @@
 		</template>
 
 		<template #extra>
-			<ParticipantPermissionsEditor v-if="permissionsEditor"
+			<ParticipantPermissionsEditor v-if="showPermissionsOptions && permissionsEditor"
 				:actor-id="participant.actorId"
 				close-after-click
 				:participant="participant"
@@ -275,7 +274,7 @@
 				@close="permissionsEditor = false" />
 
 			<!-- Confirmation required to remove participant -->
-			<NcDialog v-if="isRemoveDialogOpen"
+			<NcDialog v-if="canBeModerated && isRemoveDialogOpen"
 				:open.sync="isRemoveDialogOpen"
 				:name="removeParticipantLabel">
 				<p> {{ removeDialogMessage }} </p>
--- a/src/components/RightSidebar/Participants/Participant.vue
+++ b/src/components/RightSidebar/Participants/Participant.vue
@@ -628,7 +628,7 @@ export default {
 		},
 
 		isSelf() {
-			return this.sessionIds.length && this.sessionIds.includes(this.currentParticipant.sessionId)
+			return this.participant.actorType === this.$store.getters.getActorType() && this.participant.actorId === this.$store.getters.getActorId()
 		},
 
 		selfIsModerator() {
--- a/src/components/RightSidebar/Participants/Participant.spec.js
+++ b/src/components/RightSidebar/Participants/Participant.spec.js
@@ -66,8 +66,12 @@ describe('Participant.vue', () => {
 		}
 
 		const conversationGetterMock = jest.fn().mockReturnValue(conversation)
+		const actorIdMock = jest.fn().mockReturnValue('user-actor-id')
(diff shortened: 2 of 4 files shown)


src/components/RightSidebar/Participants/Participant.vue
```diff
@@ -628,7 +628,8 @@ export default {
 		},
 
 		isSelf() {
-			return this.sessionIds.length && this.sessionIds.includes(this.currentParticipant.sessionId)
+			return (this.participant.actorType === this.$store.getters.getActorType() && this.participant.actorId === this.$store.getters.getActorId())
+				|| (this.sessionIds.length && this.sessionIds.includes(this.currentParticipant.sessionId))
```
DorraJaouad, line 632: also drop the session comparison? actor comparison looks enough

Antreesy: Yes, was discussed with @nickvergessen that it comes from an old code and seems to be redundant now
- we don't show yourself from another device as a different participant
- you can't moderate yourself anyway

src/components/RightSidebar/Participants/Participant.vue
```diff
@@ -787,7 +788,16 @@ export default {
...
+		},
+
+		participantNavigationId() {
+			// FIXME: as RecycleScroller reuses item components, we need to reset the state,
+			// if participant object has been replaced by another one
+			this.permissionsEditor = false
+			this.isRemoveDialogOpen = false
+			this.isBanParticipant = false
+			this.internalNote = ''
+		},
```
ShGKme, line 800: This is expected behavior of `RecycleScroller` and the same may happen in other cases where the component is reused, but it doesn't fully support props update.

This `watch` fixes props update support.

But in general case I'd:
- Add `:key` where we render this component in `ParticipantsListVirtual`
- Remove `:key` from `<NcListItem>` in this component (not needed)

It would be the same as setting `:key` with `v-for`. With `key` we don't need to worry about missing one property in this watch or similar issues N components deeper in the tree.

ShGKme:
> * RecycleScroller doesn't work well with constantly changing lists

What do you mean?

src/components/RightSidebar/Participants/Participant.vue
Antreesy: > Add :key where we render this component in ParticipantsListVirtual

Isn't that killing the meaning of RecycleScroller? But I'll try with large lists

Antreesy:
> What do you mean?

If you're switching to a different conversation, components with attendees from previous one may still be mounted and stay in RecycleScroller pool. By passing key to ParticipantVirtualList and Participant they will re-render, so I don't see an issue there anymore.

See reproduction steps in description

src/components/RightSidebar/Participants/Participant.vue
ShGKme: The meaning of the virtual list is to render only visible items and not all items. So if there are 500 participants in the call, only 20 are rendered in the document.

Adding `:key` doesn't lead to rendering all items.

It makes scrolling slower by re-creating a node inside the element. But the same does existing `:key` on `NcListItem` on the next child.

The best (most performant) case is when all the components in the virtual list are stateless, having only props and data from outside the components, but not local data. Making it may require more changes.

ShGKme (approved):

PR #14499, opened 2025-02-25
feat(conversations): handle event conversation rooms

### LINKED ISSUE #14402: 📅 Handling of event conversations - ⚙️ API
- Ref https://github.com/nextcloud/spreed/issues/14401

miaulalala:
### ☑️ Resolves

* Needs https://github.com/nextcloud/calendar/pull/6768

* Fix #14402 

### 🚧 Tasks

- [x] Listen to creation of VEVENTs on the DAV events and write start and end date to the DB (be careful of TZ)
- [x] ...

### 🏁 Checklist

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -83,6 +83,7 @@ use OCA\Talk\Files\TemplateLoader as FilesTemplateLoader;
 use OCA\Talk\Flow\RegisterOperationsListener;
 use OCA\Talk\Listener\BeforeUserLoggedOutListener;
 use OCA\Talk\Listener\BotListener;
+use OCA\Talk\Listener\CalDavEventListener;
 use OCA\Talk\Listener\CircleDeletedListener;
 use OCA\Talk\Listener\CircleEditedListener;
 use OCA\Talk\Listener\CircleMembershipListener;
@@ -130,6 +131,8 @@ use OCP\AppFramework\App;
 use OCP\AppFramework\Bootstrap\IBootContext;
 use OCP\AppFramework\Bootstrap\IBootstrap;
 use OCP\AppFramework\Bootstrap\IRegistrationContext;
+use OCP\Calendar\Events\CalendarObjectCreatedEvent;
+use OCP\Calendar\Events\CalendarObjectUpdatedEvent;
 use OCP\Collaboration\AutoComplete\AutoCompleteFilterEvent;
 use OCP\Collaboration\Resources\IProviderManager;
 use OCP\Collaboration\Resources\LoadAdditionalScriptsEvent;
@@ -227,6 +230,10 @@ class Application extends App implements IBootstrap {
 		$context->registerEventListener(MessageParseEvent::class, SystemMessage::class);
 		$context->registerEventListener(MessageParseEvent::class, SystemMessage::class, 9999);
 		$context->registerEventListener(MessageParseEvent::class, UserMention::class, -100);
+
+		// Calendar listeners
+		$context->registerEventListener(CalendarObjectCreatedEvent::class, CalDavEventListener::class);
+		$context->registerEventListener(CalendarObjectUpdatedEvent::class, CalDavEventListener::class);
 
 		// Files integration listeners
 		$context->registerEventListener(BeforeGuestJoinedRoomEvent::class, FilesListener::class);
--- a/lib/Listener/CalDavEventListener.php
+++ b/lib/Listener/CalDavEventListener.php
@@ -1,0 +1,136 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+namespace OCA\Talk\Listener;
+
+use OCA\DAV\CalDAV\TimezoneService;
+use OCA\Talk\Exceptions\ParticipantNotFoundException;
+use OCA\Talk\Exceptions\RoomNotFoundException;
+use OCA\Talk\Manager;
+use OCA\Talk\Room;
+use OCA\Talk\Service\ParticipantService;
+use OCA\Talk\Service\RoomService;
+use OCP\Calendar\Events\CalendarObjectCreatedEvent;
+use OCP\Calendar\Events\CalendarObjectUpdatedEvent;
+use OCP\EventDispatcher\Event;
+use OCP\EventDispatcher\IEventListener;
+use Psr\Log\LoggerInterface;
+use Sabre\VObject\ParseException;
+use Sabre\VObject\Property\ICalendar\Date;
+use Sabre\VObject\Property\ICalendar\DateTime;
+use Sabre\VObject\Reader;
+
+/** @template-implements IEventListener<CalendarObjectCreatedEvent|CalendarObjectUpdatedEvent> */
+class CalDavEventListener implements IEventListener {
+
+	public function __construct(
+		private Manager $manager,
+		private RoomService $roomService,
+		private LoggerInterface $logger,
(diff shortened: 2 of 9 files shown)


lib/Controller/RoomController.php
```diff
@@ -705,6 +705,9 @@ protected function createEmptyRoom(string $roomName, bool $public = true, string
 		} elseif ($objectType === Room::OBJECT_TYPE_PHONE) {
 			// Ignoring any user input on this one
 			$objectId = $objectType;
+		} elseif($objectType === Room::OBJECT_TYPE_EVENT) {
+			// We don't have a UUID for this (yet)
+			$objectId = sha1($this->userId . $this->timeFactory->getTime());
```
nickvergessen, line 710: No, the sending client has to do this

lib/Migration/Version22000Date20250224113228.php
```diff
@@ -0,0 +1,61 @@
...
+
+		if (!$schema->hasTable('talk_room_events')) {
+			$table = $schema->createTable('talk_room_events');
+
+			$table->addColumn('id', Types::INTEGER, [
+				'autoincrement' => true,
+				'notnull' => true,
+				'length' => 11,
+			]);
+			$table->addColumn('room_id', Types::INTEGER, [
```
nickvergessen, line 39: we do ourselves a favor query/recognizing wise if we use the token :see_no_evil:

lib/Migration/Version22000Date20250224113228.php
```diff
@@ -0,0 +1,61 @@
...
+				'notnull' => true,
+				'length' => 11,
+			]);
+			$table->addColumn('start', Types::INTEGER, [
```
nickvergessen, line 43: Nowadays we could use datetime?

lib/Model/RoomEvent.php
```diff
@@ -0,0 +1,50 @@
...
+
+	public function __construct() {
+		$this->addType('roomId', Types::BIGINT);
+		$this->addType('start', Types::INTEGER);
+		$this->addType('end', Types::INTEGER);
+		$this->addType('description', Types::STRING);
+	}
+
+	/**
+	 * @return TalkRoomEvent
```
nickvergessen, line 40: Missing
```
@psalm-import-type ...
```

In the class comment

lib/Model/RoomEventMapper.php
```diff
@@ -0,0 +1,45 @@
...
+ * @method RoomEvent mapRowToEntity(array $row)
+ * @method RoomEvent findEntity(IQueryBuilder $query)
+ * @method list<RoomEvent> findEntities(IQueryBuilder $query)
+ * @template-extends QBMapper<RoomEvent>
+ */
+class RoomEventMapper extends QBMapper {
+	public function __construct(
+		IDBConnection $db,
+	) {
+		parent::__construct($db, 'talk_room_events', Consent::class);
```
nickvergessen, line 25:
```suggestion
		parent::__construct($db, 'talk_room_events', RoomEvent::class);
```

lib/Model/RoomEventMapper.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * SPDX-FileCopyrightText: 2023 Nextcloud GmbH and Nextcloud contributors
```
nickvergessen, line 5:
```suggestion
 * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
```

lib/Model/RoomEvent.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2022 Nextcloud GmbH and Nextcloud contributors
```
nickvergessen, line 6:
```suggestion
 * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
```

lib/Controller/RoomController.php
miaulalala: you mean in https://github.com/nextcloud/calendar/pull/6768 ?

lib/Migration/Version22000Date20250224113228.php
miaulalala: does that work across all platforms? I also prefer timestamps to datetimes (or is that the same thing in sql)?

lib/Model/RoomEvent.php
miaulalala: Every time :see_no_evil:

lib/Controller/RoomController.php
miaulalala: Done.

miaulalala:
Questions / tasks for monday:

- Where do we get the event data from? Will the code that's currently getting the latest event from CalDAV insert his into the DB
- Speaking of that, do we need a new endpoint to allow adding event details?
- Does the room need more info on a DB level or is the object type enough?

lib/Migration/Version22000Date20250224113228.php
nickvergessen: > does that work across all platforms?

It does and we actually use it in the Mapper/Model that you copied :D Consent
Also Bots, Bans, Federation and Notifications :P 

> I also prefer timestamps to datetimes

As a sysadmin I prefer readable things in results :P

miaulalala: Hm I'm more partial to timestamps are they're easily worked with whereas DateTime values always make me nervous with Timezones.

lib/Manager.php
nickvergessen, line 1: 🙈

lib/AppInfo/Application.php
```diff
@@ -227,6 +230,10 @@ public function register(IRegistrationContext $context): void {
 		$context->registerEventListener(MessageParseEvent::class, SystemMessage::class, 9999);
 		$context->registerEventListener(MessageParseEvent::class, UserMention::class, -100);
 
+		// Calendar listeners
+		$context->registerEventListener(CalendarObjectCreatedEvent::class, CalDavEventListener::class);
```
nickvergessen, line 234: psalm is not happy, we should add a stub in tests/stubs similar to https://github.com/nextcloud/spreed/blob/main/tests/stubs/oca_files_events.php

lib/Listener/CalDavEventListener.php
```diff
@@ -0,0 +1,108 @@
...
+		$vevent = $vobject->VEVENT;
+		// Check if the location is set and if the location string contains a call url
+		$location = $vevent->LOCATION->getValue();
+		if ($location === null || !str_contains($location, '/call/')) {
+			$this->logger->debug('No location for the event or event is not call link, skipping.');
+			return;
+		}
+
+		// Check if room exists and check if user is part of room
+		$roomToken = array_reverse(explode('/', $location))[0];
```
nickvergessen, line 65:
```suggestion
		$roomToken = end(explode('/', $location));
```

lib/Listener/CalDavEventListener.php
```diff
@@ -0,0 +1,108 @@
...
+		try {
+			$room = $this->manager->getRoomByToken($roomToken, $this->userId);
```
nickvergessen, line 67: `getRoomByToken` does not force a membership. It just loads it when it exists. `getRoomForUserByToken` throws `RoomNotFoundException` when the user is not a member, so we should use that:
```suggestion
			$room = $this->manager->getRoomForUserByToken($roomToken, $this->userId);
```

but it still returns open and listable rooms, so afterwards we should get the membership (will not run a query when the user was found before):
```php
$participant = $this->participantService->getParticipant($room, $userId, false);
```
that throws ParticipantNotFoundException => return

Should we limit this to moderators, same as with the "schedule a meeting"?

nickvergessen (approved):

Issue #14401, opened 2025-02-14
📅 Handling of event conversations


nickvergessen:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

### Related issues
- https://github.com/nextcloud/spreed/issues/2448
- https://github.com/nextcloud/spreed/issues/13975

### Tracking
- API https://github.com/nextcloud/spreed/issues/14402
- Frontend https://github.com/nextcloud/spreed/issues/14403
- Android https://github.com/nextcloud/talk-android/issues/4713
- iOS https://github.com/nextcloud/talk-ios/issues/1986

nickvergessen:
Comments:
- Optional to be combined with "Tags", "Labels", "Categories", …
- Event conversations created via calendar integrations flood the left sidebar and should be "hidden" until "relevant" (1h before the meeting, 1h afterwards?, exit poll to retain or purge the conversation "soon")

nickvergessen:
- Give them a dedicated icon
- Replace subline with relative meeting time if "in the future"
- API needs to be extended to store an object_type and the meeting time
- Offer an option to archive now / delete in one day (moderator) in rooms of these object_type
- Hide event conversations until "timely close"
- Filter option to show all calendar events even when not time close

nickvergessen:
Object part in server's OCP\Talk connector: https://github.com/nextcloud/server/pull/50921

miaulalala:
- Hide room type EVENT if object id is smaller (timestamp - 3600 * 24)
- Archive room type EVENT if the last activty timestamp is (current - 3600*24*7) - i. e. if they haven't been used for a week and no new messages were sent
- When a call ends in the room type EVENT, show buttons to archive / delete the room immediately - similar to green "join call" button we have now in the chat area

PR #14510, opened 2025-02-27
fix(conversations): whitelist event object type

miaulalala:
Needed for https://github.com/nextcloud/spreed/issues/14401

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist


lib/Controller/RoomController.php
```diff
@@ -705,6 +705,10 @@ protected function createEmptyRoom(string $roomName, bool $public = true, string
 		} elseif ($objectType === Room::OBJECT_TYPE_PHONE) {
 			// Ignoring any user input on this one
 			$objectId = $objectType;
+		} elseif ($objectType === Room::OBJECT_TYPE_EVENT) {
+			// Allow event rooms in future versions without breaking in older talk rooms that the same calendar version supports
```
nickvergessen, line 709:
```suggestion
			// Allow event rooms in future versions without breaking in older talk versions that the same calendar version supports
```

Sorry I failed

Issue #7125, opened 2022-04-08
Allow selecting Speaker (Audio Out)


danir-de:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

**Is your feature request related to a problem? Please describe.**
Currently and thanks to #358 it is possible to select the desired devices like microphone and camera.
What therefor should be possible and is expected by most users, would be the ability to change the audio output device (speaker, headphones,...) on the fly - but it isn't.

We encountered this on multiple occasions:
- When using a Bluetooth headset, there are two audio device profiles: Default (just audio out) and Handsfree (Mic+Out). When starting a call, using the Bluetooth devices microphone, the audio profile switches from Default to Handsfree, in order to be able to use the devices microphone. Since the default output lays on the Default profile of the device, the user can't hear anything anymore, unless he switches the speaker output from the "Default Profile" device to the "Handsfree Profile" device - which can't be done in Talk.
- The user has two sound outputs: Headphones (for music) and a speaker (for Talk). The default output device on the system level is set to the headphone device, since that's what he uses the most. When engaging in a Talk call, the user might want to use the speaker device to be able to speak more freely. This isn't possible, without changing the system settings for the entire browser (or system) to use the alternate device.

**Describe the solution you'd like**
We currently already have settings / selectors for the camera device, as well as the microphone device used. 
"Just add"™ another option for selecting the speaker device, like any other app (TS3, Discord, Skype, Zoom, Teams,...)

**Describe alternatives you've considered**
There are no real alternatives. Changing the output device on system level is only possible for the entire browser, not just for the single tab and isn't something regular users would want to do.

**Additional context**
Here is a screenshot of this being a thing on the browser version of zoom:
![image](https://user-images.githubusercontent.com/4672840/162400744-0bb94817-0c1d-483e-b7d0-245436264c0b.png)

meonkeys:
Can we rename this to "allow changing audio input and output during a call"?

I use a couple of workarounds that do require leaving and re-joining the call, but perhaps they are worth a mention:

1. leave the call, click join, then change the audio input/output just before re-joining
2. change the default mic & speakers in the system settings

My server is v28 and my Talk app is v18.0.5.

[...]

PR #14511, opened 2025-02-27
fix(MediaDevicesSelector): refactor, allow audio output

Antreesy:
### ☑️ Resolves

* Ref #7125 
  * refactor commits are to be squashed
  * [wip] commits are to be removed
  * check by commits for easier review
* Design icon was discussed => test section will be reworked instead separately

image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


ShGKme:
> * Not risky to browser differences / client

I'm unsure it is not risky

src/components/MediaSettings/MediaDevicesSelector.vue
```diff
@@ -60,47 +58,37 @@ const deviceSelectorPlaceholder = computed(() => {
...
-
-// The watcher should not be set as "immediate" to prevent
-// "update:deviceId" from being emitted when mounted with the same value
-// initially passed to the component.
-watch(deviceSelectedOption, (deviceSelectedOption, previousSelectedOption) => {
+/**
+ * Update deviceId if passes required checks
+ * @param newValue selected NcSelect option to update with
+ */
+function updateDeviceId(newValue: NcSelectOption | null) {
```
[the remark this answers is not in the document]
ShGKme: **nitpick:** I think it's simpler to understand the function if it has `deviceId` as the value, and not a general `newValue`

src/components/MediaSettings/MediaDevicesSelector.vue
```diff
@@ -3,21 +3,110 @@
...
+		<NcSelect :model-value="deviceSelectedOption"
+			:input-id="`device-selector-${props.kind}`"
 			:options="deviceOptions"
 			label="label"
 			:aria-label-combobox="t('spreed', 'Select a device')"
 			:clearable="false"
 			:placeholder="deviceSelectorPlaceholder"
-			:disabled="!enabled || !deviceOptionsAvailable" />
+			:disabled="!enabled || !deviceOptionsAvailable"
+			@option:selected="updateDeviceId" />
```
ShGKme, line 109: Before there was `v-model` which handles `@update:modelValue`.

Unlike `update:modelValue`, the new `option:selected` event is only triggered when some new option is selected, but not then it is unselected (`modelValue` is updated to `null`.

ShGKme: Haven't tested, but feels like an error to me

DorraJaouad: We can't unselect there

src/stores/sounds.js
```diff
@@ -52,14 +53,23 @@ export const useSoundsStore = defineStore('sounds', {
...
+					this.audioObjectsPromises[key].catch(error => {
+						console.error(error)
+					})
+				})
+			} else {
+				this.audioObjectsPromises[key] = this.audioObjects[key].play()
+				this.audioObjectsPromises[key].catch(error => {
+					console.error(error)
+				})
+			}
```
DorraJaouad, line 72:
```js
if (this.audioOutputSinkId) {
		 try {
			await audioObject.setSinkId(this.audioOutputSinkId)
		 } catch (error) {
			console.error(error)
		 }
	 }
this.audioObjectsPromises[key] = this.audioObjects[key].play()
				this.audioObjectsPromises[key].catch(error => {
					console.error(error)
				})
```

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -85,6 +85,11 @@
 						:device-id="videoInputId"
 						@refresh="updateDevices"
 						@update:deviceId="handleVideoInputIdChange" />
+					<MediaDevicesSelector kind="audiooutput"
+						:devices="devices"
+						:device-id="soundsStore.audioOutputSinkId"
+						@refresh="updateDevices"
+						@update:deviceId="handleAudioOutputIdChange" />
```
DorraJaouad, line 92: Devices contain audio input too, we need another key to filter with
Also, when testing the speaker and you change the device, the test needs to stop.

src/stores/sounds.js
Antreesy: I'll take a look. But it's wip for testing, will drop it from this PR anyway

src/components/MediaSettings/MediaSettings.vue
Antreesy: yes, and video input too. They are all filtered inside the MediaDevicesSelector.vue

src/components/MediaSettings/MediaDevicesSelector.vue
```diff
@@ -60,47 +58,37 @@ const deviceSelectorPlaceholder = computed(() => {
 	}
 })
 
-
-const deviceSelectedOptionFromDeviceId = computed(() => {
-	return deviceOptions.value.find(option => option.id === props.deviceId)
+const deviceSelectedOption = computed<NcSelectOption | null>(() => {
+	return deviceOptions.value.find(option => option.id === props.deviceId) ?? null
```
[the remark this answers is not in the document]
Antreesy: I'm pretty sure `Array.find()` returns undefined 👀

src/components/MediaSettings/MediaDevicesSelector.vue
ShGKme: Yeah, missed `clearable="false"`.

Then `updateDeviceId` function cannot receive null as an option?

Antreesy: We can't unselect there, but I now found a way to compare with previous selection, like watcher did before

ShGKme (approved):

src/components/MediaSettings/MediaDevicesSelector.vue
Antreesy: Same timing =)
I think it's safer to handle `NcSelectOption|null`, as component supports it? And `updateDeviceId` not receives id only, not the whole option object, so it's `string|null`

src/components/MediaSettings/MediaSettings.vue
DorraJaouad: Indeed, it's not filtered correctly in `MediaDevicesSelector`, that's what I meant

DorraJaouad (approved):

Issue #14406, opened 2025-02-14
🍱 Talk Dashboard


nickvergessen:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

### Ref
- https://github.com/nextcloud/spreed/issues/13975

### Tracking issues
- API https://github.com/nextcloud/spreed/issues/14407
- Frontend https://github.com/nextcloud/spreed/issues/14408

nickvergessen:
- "Quick RSPV"

nickvergessen:
MVP (backend)

- Upcoming events
	- Go via CalDAV
	- filter by location field to get rooms that are relevant
- New table for recent (i. e. past) calls:
	- Start / end
	- has_recording
	- has_shared_recording
	- recording file_id
	- has_transcription
	- etc

miaulalala:
Nimisha's draft from 2025/04/10

![Image](https://github.com/user-attachments/assets/c3af0d47-1c23-46a0-8536-94cef6284c36)

## Needed data
- [ ] Calendar Name
	- [ ] If multiple, use "Personal" or highest permission first
- [ ] Calendar Colour
	- [ ] If multiple, use "Personal" or highest permission first
- [ ] Name of calendar event
- [ ] Times:
	- [ ] Start and end time 
	- [ ] Elapsed call time
- [ ] Room Token
- [ ] Room Avatar Version
- [ ] Room Name
- [ ] Room Display Name
- [ ] Room Type
- [ ] Room Active Since (for Ongoing + elapsed time)
	- If clients have room in room list, use the values from that (more accurate) 
- [ ] Description of calendar event
- [ ] Calendar attachment data
	- Ex: `ATTACH;FMTTYPE=image/jpeg;FILENAME=/Alice.jpg;X-NC-FILE-ID=123;X-NC-HAS-PR
EVIEW=true:/index.php/f/123`
- [ ] Event attendees
	- [ ] Count attendees by partstat, return count for accepted, tentative, declined, needs action
	- [ ] For 1:1 ignore self
- [ ] Calendar event link (Calendar app link)

Participant check should take into account if room is open / public and should also return rooms that participant hasn't joined yet but are public

## Change requests for Design
[...]

PR #14522, opened 2025-02-28
fix(Views): migrate to ts

DorraJaouad:
### ☑️ Resolves

* Prerequisite to #14406 

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


src/env.d.ts
```diff
@@ -20,6 +20,10 @@ declare global {
 			setValue: (app: string, key: string, value: string | number | boolean, options?: { success?: () => void, error?: () => void }) => void,
 		},
 	}
+	module '*.svg?raw' {
+		const content: string
+		export default content
+	  }
```
ShGKme, line 26:
```suggestion
		const content: string
		export default content
}
```

src/router/router.ts
```diff
@@ -76,5 +76,5 @@ export default new Router({
 			component: CallView,
 			props: true,
 		},
-	],
+	]
```
ShGKme, line 79:
```suggestion
	],
```

src/views/WelcomeView.vue
```diff
@@ -23,93 +15,76 @@ import NcLoadingIcon from '@nextcloud/vue/dist/Components/NcLoadingIcon.js'
 import EmptyView from '../components/EmptyView.vue'
 
 import IconTalk from '../../img/app-dark.svg?raw'
+import { useStore } from '../composables/useStore.js'
+import Router from '../router/router.ts'
```
ShGKme, line 19: Instead of direct import, `useRouter` should be user in Composition API, same as `this.$router` in Options API

src/views/WelcomeView.vue
```diff
@@ -23,93 +15,76 @@ import NcLoadingIcon from '@nextcloud/vue/dist/Components/NcLoadingIcon.js'
...
+			params: { token: newConversation.token },
+		})
+	} catch (error) {
+		showError(t('spreed', 'Error while joining the conversation'))
+		console.error(error)
+		Router.push({ name: 'notfound' })
 	}
 }
+
+watch(callUser, (newCallUser) => {
```
Antreesy, line 75: RFC: instead, Router navigation guard can be used: https://router.vuejs.org/guide/advanced/composition-api.html#Navigation-Guards

This could  provide us more opportunities, e.g. prevent navigation and show input dialog like "You don't have a conversation with this person yet. Do you like to write your first message?" before creating

src/views/MainView.vue
```diff
@@ -27,74 +16,57 @@ import PollViewer from '../components/PollViewer/PollViewer.vue'
...
-	mounted() {
-		const handleRouteHashChange = (token, route) => {
-			if (route?.hash === '#direct-call') {
-				emit('talk:media-settings:show')
-				Router.replace({ ...route, hash: '' })
-			} else if (route?.hash === '#settings') {
-				emit('show-conversation-settings', { token })
-				Router.replace({ ...route, hash: '' })
-			}
+onMounted(() => {
```
Antreesy, line 42: Works without hook as well
```suggestion
```

src/views/MainView.vue
```diff
@@ -2,21 +2,10 @@
   - SPDX-FileCopyrightText: 2019 Nextcloud GmbH and Nextcloud contributors
   - SPDX-License-Identifier: AGPL-3.0-or-later
 -->
+<script lang="ts" setup>
+import { computed, defineProps, onMounted, watch } from 'vue'
```
Antreesy, line 6:
```suggestion
import { computed, watch } from 'vue'
```

src/views/WelcomeView.vue
DorraJaouad: We are using Vue Router 3 still, and `useRouter` is only available from Vue Router 4.

src/env.d.ts
```diff
@@ -21,6 +21,11 @@ declare global {
 		},
 	}
 
+	module '*.svg?raw' {
+		const content: string
+		export default content
+	}
```
ShGKme, line 27:
```suggestion
	declare module '*.svg?raw' {
		const content: string
		export default content
	}
```

src/views/WelcomeView.vue
ShGKme:
```suggestion
import { useRouter } from 'vue-router/composables'
```

- Source: https://github.com/vuejs/vue-router/blob/v3.6.5/src/composables/globals.js
- Export: https://github.com/vuejs/vue-router/blob/v3.6.5/package.json#L37

src/env.d.ts
ShGKme: Without `declare` it's implementation and suppsed to be compiled to
```ts
"use strict";
var '*.svg?raw';
(function ('*.svg?raw') {
    const content;
    export default content;
})('*.svg?raw' || ('*.svg?raw' = {}));

```

src/views/MainView.vue
```diff
@@ -27,74 +13,43 @@ import PollViewer from '../components/PollViewer/PollViewer.vue'
...
-				Router.replace({ ...route, hash: '' })
-			} else if (route?.hash === '#settings') {
-				emit('show-conversation-settings', { token })
-				Router.replace({ ...route, hash: '' })
-			}
-		}
-
-		handleRouteHashChange(this.token, Router.currentRoute)
-		Router.afterEach((to) => handleRouteHashChange(this.token, to))
-	},
```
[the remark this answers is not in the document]
DorraJaouad: I misunderstood Maksim, it's still there

src/views/WelcomeView.vue
```diff
@@ -23,93 +16,57 @@ import NcLoadingIcon from '@nextcloud/vue/dist/Components/NcLoadingIcon.js'
 import EmptyView from '../components/EmptyView.vue'
 
 import IconTalk from '../../img/app-dark.svg?raw'
+import { useStore } from '../composables/useStore.js'
 
-export default {
-	name: 'WelcomeView',
+const store = useStore()
+const Router = useRouter()
```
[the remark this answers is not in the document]
DorraJaouad: I knew you will see it :D

src/views/MainView.vue
```diff
@@ -27,74 +17,60 @@ import PollViewer from '../components/PollViewer/PollViewer.vue'
...
+			next({ ...route, hash: '', name: route.name ?? undefined })
+		} else {
+			next()
 		}
+	}
+
+	handleRouteHashChange(store.state.token, to)
+})
 
-		handleRouteHashChange(this.token, Router.currentRoute)
```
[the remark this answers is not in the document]
ShGKme: Alternative - move it from this component to a more relevant place, so we don't need to wonder if `emit` works here or not.

ShGKme (approved):

src/views/MainView.vue
```diff
@@ -27,74 +16,56 @@ import PollViewer from '../components/PollViewer/PollViewer.vue'
...
-				emit('show-conversation-settings', { token })
-				Router.replace({ ...route, hash: '' })
-			}
+onMounted(() => {
+	watchEffect(() => {
+		if (route.hash === '#direct-call') {
+			emit('talk:media-settings:show', '')
+			router.replace({ hash: '' })
+		} else if (route.hash === '#settings') {
+			emit('show-conversation-settings', { token: store.state.token })
```
Antreesy, line 49: I'm afraid to forget about it during some store migration:

```suggestion
			emit('show-conversation-settings', { token: store.getters.getToken() })
```

Antreesy (approved):

PR #14566, opened 2025-03-05
fix(AvatarWrapper): fix img display property, provide fallback icon on failed src load

Antreesy:
### ☑️ Resolves

* Fix 14556

🏚️ Before | 🏡 After | 🏡 After with fallback
-- | -- | --
image | image | image

### 🏁 Checklist

  - [x] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


src/components/AvatarWrapper/AvatarWrapper.vue
```diff
@@ -18,7 +18,8 @@
 			:width="size"
 			:height="size"
 			:alt="name"
-			class="avatar icon">
+			class="avatar icon"
+			@error="onError">
```
ShGKme, line 22: Or

```suggestion
			@error="failed = true">
```

ShGKme (approved):

src/components/AvatarWrapper/AvatarWrapper.vue
```diff
@@ -136,8 +139,12 @@ export default {
 
 	setup() {
 		const isDarkTheme = useIsDarkTheme()
+
+		const failed = ref(false)
```
ShGKme, line 143: Probably we should also reset it on token/source update. Otherwise the component doesn't support re-rendering with new props after it failed loading.

ShGKme: Or the component must be used with `key`

Antreesy: watch after avatarUrl would be enough, I think

ShGKme: Yes, if you mean reset failed in watch for avatarUrl

ShGKme (approved):

PR #14621, opened 2025-03-11
fix(l10n): Fix translation source strings

nickvergessen:
### ☑️ Resolves

* Found with https://github.com/nextcloud/docker-ci/pull/740


nickvergessen:
Backporting only after tomorrows releases

Antreesy (approved):

PR #14629, opened 2025-03-12
fix(calls): Fix displayname when exporting call participants

nickvergessen:
### 🏁 Checklist


Antreesy (approved):
Lgtm

lib/Controller/CallController.php
```diff
@@ -166,12 +161,12 @@ public function downloadParticipantsForCall(string $format = 'csv'): DataDownloa
 			} elseif ($participant->getAttendee()->getActorType() === Attendee::ACTOR_USERS) {
 				$email = $this->userManager->get($participant->getAttendee()->getActorId())?->getEMailAddress() ?? '';
 			}
-			fputcsv($output, [
+			fputcsv($output, array_map([$this, 'escapeFormulae'], [
```
DorraJaouad, line 164: Is that a typo? `escapeFormula`

Antreesy: It's plural

DorraJaouad (approved):

PR #14637, opened 2025-03-12
fix(Grid): fix cropped videos in the last row

Antreesy:
### ☑️ Resolves

* replace `window.addEventListener('resize')` with `ResizeObserver`
  * `$refs.grid` is rendered conditionally, so we might lose a reference to it once mounted
  * 1 listener instead of 2, looks only for exact DOM element
* fix video cropping at the last row
* drop redundant method and store getter
* cleanup comments
* improve DX by exposing (in debug mode only):
  * `OCA.Talk.gridDebugInformation`
  * `OCA.Talk.gridDevModeEnable`

🏚️ Before | 🏡 After
-- | --
image | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/CallView/Grid/Grid.vue
+++ b/src/components/CallView/Grid/Grid.vue
@@ -4,7 +4,7 @@
 -->
 
 <template>
-	<div class="grid-main-wrapper" :class="{'is-grid': !isStripe, 'transparent': isLessThanTwoVideos}">
+	<div ref="gridWrapper" class="grid-main-wrapper" :class="{'is-grid': !isStripe, 'transparent': isLessThanTwoVideos}">
 		<NcButton v-if="isStripe && !isRecording"
 			class="stripe--collapse"
 			type="tertiary-no-background"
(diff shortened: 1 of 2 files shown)


src/components/CallView/Grid/Grid.vue
```diff
@@ -398,10 +403,10 @@ export default {
 
 		// Max number of columns possible
 		columnsMax() {
-			// Max amount of columns that fits on screen, including gaps and paddings (8px)
-			const calculatedApproxColumnsMax = Math.floor((this.gridWidth - 8 * this.columns) / this.dpiAwareMinWidth)
+			// Max amount of columns that fits on screen, including gaps and paddings (--grid-gap, 8px)
+			const calculatedApproxColumnsMax = Math.floor((this.gridWidth - GRID_GAP * this.columns) / this.dpiAwareMinWidth)
```
DorraJaouad, line 407: Hmm it seems like we are counting extra gap in `max*` 
We always have n-1 gaps for each n columns

src/components/CallView/Grid/Grid.vue
```diff
@@ -398,10 +403,10 @@ export default {
...
 			// Max amount of columns that fits on screen (with one more gap, as if we try to fit one more column)
-			const calculatedHypotheticalColumnsMax = Math.floor((this.gridWidth - 8 * (this.columns + 1)) / this.dpiAwareMinWidth)
+			const calculatedHypotheticalColumnsMax = Math.floor((this.gridWidth - GRID_GAP * (this.columns + 1)) / this.dpiAwareMinWidth)
```
DorraJaouad, line 409: Same here, if we were to count extra gap as if we have another column, it will be n not n+1

DorraJaouad (approved):
tested

Issue #14410, opened 2025-02-14
🔔 Setup check to ensure notify_push is configured - ⚙️ API


nickvergessen:
- Ref https://github.com/nextcloud/spreed/issues/14409

PR #14638, opened 2025-03-12
feat: add notify_push setup check

miaulalala:
### ☑️ Resolves

* Fix https://github.com/nextcloud/spreed/issues/14410

### 🏁 Checklist

### DIFF
--- a/lib/SetupCheck/NotifyPush.php
+++ b/lib/SetupCheck/NotifyPush.php
@@ -1,0 +1,52 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Talk\SetupCheck;
+
+use OCP\App\IAppManager;
+use OCP\IL10N;
+use OCP\SetupCheck\ISetupCheck;
+use OCP\SetupCheck\SetupResult;
+
+class NotifyPush implements ISetupCheck {
+
+	public function __construct(
+		protected IL10N $l10n,
+		protected IAppManager $appManager,
+	) {
+	}
+
+	public function getName(): string {
+		return $this->l10n->t('Client Push'); // TRANSLATORS: this is the app name of the notify_push app.
+	}
+
+	public function getCategory(): string {
+		return 'talk';
+	}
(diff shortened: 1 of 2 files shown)


lib/SetupCheck/NotifyPush.php
```diff
@@ -0,0 +1,52 @@
...
+		protected IAppManager $appManager,
+	) {
+	}
+
+	public function getName(): string {
+		return $this->l10n->t('Client Push'); // TRANSLATORS: this is the app name of the notify_push app.
+	}
+
+	public function getCategory(): string {
+		return 'accounts';
```
nickvergessen, line 30: talk

nickvergessen (approved):

lib/SetupCheck/NotifyPush.php
```diff
@@ -0,0 +1,52 @@
...
+
+	public function run(): SetupResult {
+		if ($this->appManager->isEnabledForAnyone('notify_push')) {
+			return SetupResult::success(
+				$this->l10n->t('Client Push is installed, this improves the performance of desktop clients.')
+			);
+		}
+
+		return SetupResult::warning(
+			$this->l10n->t('{notify_push} is not installed, this might lead to performance issues when using desktop clients.'),
```
nickvergessen, line 41: While this is true, I wonder if we should outline that it's not about performance in the case of Talk desktop clients, but about punctuality aka delay of notifications

joshtrichards: Just ran across someone on the forum that encountered this warning (a new user) and Talk didn't even come up in the interaction. I wonder if we should do two things here:

a) Change the warning to something like:

> The `{notify_push}` service is not installed. This may result in: **Increased Server Load** (because desktop sync clients must regularly poll the server for file changes, using more resources) and/or **Delayed Updates/Notifications** (file sync changes and Talk notifications are only delivered during these periodic checks, rather than instantly).

b) Move this check to core since it's not specific *only* Talk

And, technically, there's a (c) todo item: Document `notify_push` in the *Admin Manual* (it's not even mentioned there)

nickvergessen: > Move this check to core since it's not specific only Talk

It can be copied there, but we in Talk want to have it specifically listed in the Talk section of the setupcheck.

PR #14664, opened 2025-03-15
fix: do not add organizer as attendee and set status

SebastianKrupinski:
### ☑️ Resolves

- Requires: https://github.com/nextcloud/server/pull/51501
- Resolves: https://github.com/nextcloud/calendar/issues/6819

Event status should be set on new events, the organizer should also not be added as a participant.

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### 🏁 Checklist

### DIFF
--- a/lib/Controller/RoomController.php
+++ b/lib/Controller/RoomController.php
@@ -76,6 +76,7 @@ use OCP\AppFramework\Http\Attribute\PublicPage;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\Services\IAppConfig;
 use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\Calendar\CalendarEventStatus;
 use OCP\Calendar\Exceptions\CalendarException;
 use OCP\Calendar\ICreateFromString;
 use OCP\Calendar\IManager as ICalendarManager;
@@ -2650,6 +2651,7 @@ class RoomController extends AEnvironmentAwareOCSController {
 		$eventBuilder->setOrganizer($user->getEMailAddress(), $user->getDisplayName() ?: $this->userId);
 		$eventBuilder->setStartDate($startDate);
 		$eventBuilder->setEndDate($endDate);
+		$eventBuilder->setStatus(CalendarEventStatus::CONFIRMED);
 
 		$userAttendees = $this->participantService->getParticipantsByActorType($this->room, Attendee::ACTOR_USERS);
 		foreach ($userAttendees as $userAttendee) {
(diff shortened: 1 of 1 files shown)


miaulalala:
Can I ask why `CONFIRMED`? Theoretically, it should be a PARTSTAT `NEEDS-ACTION`. The STATUS should not be set for the user without their input.

miaulalala (changes requested):
Can I ask why `CONFIRMED`? Theoretically, it should be a PARTSTAT `NEEDS-ACTION`. The STATUS should not be set for the user without their input.

SebastianKrupinski:
> Can I ask why `CONFIRMED`? Theoretically, it should be a PARTSTAT `NEEDS-ACTION`. The STATUS should not be set for the user without their input.

Morning, the status is for the event not for the attendees, "PARTSTAT `NEEDS-ACTION`" this would be for the attendees. The event STATUS and Organizer status should be set to confimed if the organizer is creating the ticket.

miaulalala:
> > Can I ask why `CONFIRMED`? Theoretically, it should be a PARTSTAT `NEEDS-ACTION`. The STATUS should not be set for the user without their input.
> 
> Morning, the status is for the event not for the attendees, "PARTSTAT `NEEDS-ACTION`" this would be for the attendees. The event STATUS and Organizer status should be set to confimed if the organizer is creating the event.

damn, right. Looks good!

miaulalala (approved):
Static analysis is complaining since the upstream PR is not merged. Please rebase when it's done, then you can merge.

lib/Controller/RoomController.php
```diff
@@ -2634,6 +2634,7 @@ public function scheduleMeeting(string $calendarUri, int $start, ?array $attende
 		$eventBuilder->setOrganizer($user->getEMailAddress(), $user->getDisplayName() ?: $this->userId);
 		$eventBuilder->setStartDate($startDate);
 		$eventBuilder->setEndDate($endDate);
+		$eventBuilder->setStatus('CONFIRMED');
```
nickvergessen, line 2637:
```suggestion
		if (method_exists($eventBuilder, 'setStatus')) {
			$eventBuilder->setStatus('CONFIRMED');
		}
```
if we want to backport

lib/Controller/RoomController.php
```diff
@@ -2650,6 +2651,7 @@ public function scheduleMeeting(string $calendarUri, int $start, ?array $attende
 		$eventBuilder->setOrganizer($user->getEMailAddress(), $user->getDisplayName() ?: $this->userId);
 		$eventBuilder->setStartDate($startDate);
 		$eventBuilder->setEndDate($endDate);
+		$eventBuilder->setStatus(CalendarEventStatus::CONFIRMED);
```
nickvergessen, line 2654: Needs to be wrapped with method exists at least on the backport

SebastianKrupinski: I saw your previous comment and was going to ask you why this would be needed since the talk app tracks NC version for version, but then I realized that the apps can be upgraded before the server. 💡 

I think wrapping it in the back port would be cleaner, that way the code from this version on will be cleaner. What would you prefer?

Issue #14667, opened 2025-03-16
Implement Centralized Unread Count Management


markusstoll:
This issue relates to talk-desktop issue https://github.com/nextcloud/talk-desktop/issues/1171

## Current Situation & Problem
The current implementation of unread message counting in Talk lacks a centralized approach, leading to:
- Inconsistent unread count management and presentation interfaces (web, desktop, mobile)

This particularly affects the desktop application, where unreliable badge counts have been a major usability issue.

## Architectural Changes
The new implementation introduces:
1. A dedicated `unreadCountStore` that:
   - Maintains the source of truth for all unread counts
   - Uses an efficient map-based tracking system
   - Emits standardized events for count updates
2. Integration with the existing conversations store
3. A new standardized event (`UNREAD_COUNT_UPDATED`) for all interfaces

## Technical Implementation
The changes include:
1. New `unreadCountStore.js`:
   - Efficient tracking of per-conversation counts using maps
   - Calculation of differences instead of full recounts
   - Standardized event emission
2. Updates to `conversationsStore.js`:
   - Integration with the new unread count store
   - Removal of redundant calculations
3. Clear event interface for all clients

These changes are required for implementing an correct management of total unread message count and presenting this in the Application Badge

I did all required implementations and will now create pull requests here and with talk-desktop

nickvergessen:
Closing as the desktop client has the necessary stuff now.

PR #14668, opened 2025-03-17
Introduce unreadCount store

markusstoll:
### ☑️ Implements issue https://github.com/nextcloud/spreed/issues/14667

* talk-desktop ref: https://github.com/nextcloud/talk-desktop/pull/1174

I had to introduce some changes to spreed that are described below to create a central unreadCountStore to have an even hook in talk-desktop to get updates of these unreadCounts and enable talk-desktop to set the correct badge count on the app icon.

### 1. New unreadCountStore
A new Vuex store module that serves as the single source of truth for all unread counters:

```javascript
state: {
    // Total counters
    totalUnreadMessages: 0,
    totalUnreadMentions: 0,
    totalUnreadMentionDirect: 0,

    // Per-conversation tracking
    unreadMessagesMap: {},
    unreadMentionsMap: {},
    unreadMentionDirectMap: {},
}
```

#### Conversations Store
- Enhanced `addConversation` mutation:
  - Automatically updates unread counters when adding new conversations
  - Handles initial state synchronization

- Improved `updateConversation` mutation:
  - Detects changes in unread counts
  - Dispatches updates to unreadCountStore
  - Maintains consistency across state changes

- Updated `deleteConversation` mutation:
  - Properly removes conversation counts from tracking
  - Updates total counters accordingly

#### Messages Store
- Enhanced `updateLastReadMessage` action:
  - Improved handling of "all messages read" scenarios
  - Proper counter reset when reaching conversation end
  - Better integration with chat-read-last feature

### 3. Event System
New event for badge updates:
```javascript
EVENTS.UNREAD_COUNT_UPDATED: 'talk:unread:updated'
```

### 🏁 Checklist

This issue is not browser depending.
But I added a matching change in talk-desktop and have a working application icon badge implementation.
This is already tested in Windows and MacOs builds of talk-desktop.

Pull request for talk-desktop is about to follow

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -95,6 +95,33 @@ export default {
 	},
 
 	computed: {
+		unreadCountsMap() {
+			return this.$store.getters.conversationsList.reduce((acc, conversation) => {
+				if (conversation.isArchived) {
+					// Do not consider archived conversations in counting
+					return acc
+				}
+
+				if (conversation.unreadMessages > 0) {
+					acc.conversations++
+					acc.messages += conversation.unreadMessages
+				}
+
+				if (conversation.unreadMention) {
+					acc.mentions++
+				}
+
+				if (conversation.unreadMentionDirect) {
+					acc.mentionsDirect++
+				}
+				return acc
+			}, {
+				conversations: 0,
+				messages: 0,
+				mentions: 0,
+				mentionsDirect: 0,
+			})
+		},
 		getUserId() {
 			return this.$store.getters.getUserId()
 		},
@@ -163,6 +190,14 @@ export default {
 				}
 			}
 		},
+
+		unreadCountsMap: {
+			deep: true,
+			immediate: true,
+			handler(value) {
+				emit('talk:unread:updated', value)
+			},
+		}
 	},
 
 	beforeCreate() {


nickvergessen:
The main colleagues are currently out a bit. Hopefully they are able to give feedback towards the end of the week

markusstoll:
@Antreesy I understand your point, but unaware of your replacement solution for Vuex, I hardly could have taken that into account.
I pull request for talk-desktop is almost finished, you will have it shortly

Antreesy:
Don't worry about it, we appreciate efforts you put into this and will help with fixing that 😉
Thanks again for your contribution! I'll be back at Wednesday and will try to check it ASAP

src/store/unreadCountStore.js
```diff
@@ -0,0 +1,182 @@
...
+
+		// Update maps
+		Vue.set(state.unreadMessagesMap, token, unreadMessages || 0)
+		Vue.set(state.unreadMentionsMap, token, unreadMentions || 0)
+		Vue.set(state.unreadMentionDirectMap, token, unreadMentionDirect || 0)
+
+		// Update totals
+		state.totalUnreadMessages += messageDiff
+		state.totalUnreadMentions += mentionsDiff
+		state.totalUnreadMentionDirect += mentionDirectDiff
```
Antreesy, line 62: getters should do this math, otherwise we're risking to miscalculate the total number at some point

src/store/conversationsStore.js
```diff
@@ -232,6 +264,14 @@ const mutations = {
 		if (unreadMentionDirect !== undefined) {
 			Vue.set(state.conversations[token], 'unreadMentionDirect', unreadMentionDirect)
 		}
+
+		// Update unread counters in the unreadCount store
+		this.dispatch('unreadCount/updateConversationCounts', {
+			token,
+			unreadMessages: state.conversations[token].unreadMessages || 0,
+			unreadMentions: state.conversations[token].unreadMention || 0,
+			unreadMentionDirect: state.conversations[token].unreadMentionDirect || 0,
```
Antreesy, line 273: these two are either `true` or `0`. I'd treat them as boolean only. I wouldn't want to rely on `true - 0 === 1` and `0 - true === -1`

src/store/unreadCountStore.js
Antreesy: while ommiting these mutations here, PR might be reduced to:
- having a getter in conversationsStore (e.g. returning unreadMessagesMap, which includes:
`[token]: { unreadMessages: number, unreadMentions: boolean, unreadMentionDirect: boolean }`
  - with a logic similar to `getLastMessageMap()` function from composable (archived conversations are considered correctly, +1 in one conversation and -1 in another conversation are considered correctly);
  - less performant, but also less error-prone (and we're already iterating the list for similar purposes in `useDocumentTitle`, maybe result could also be reused in both places?)
- having an initialization logic (e.g. in App.vue or LeftSidebar.vue), which watch after this getter changes and emit events (with total payload) to Desktop client (or another recipients, where needed)

Antreesy (commented):
Gave it a smoke test, seems to be working as expected 🎉

Leaving my thoughts for this PR. I suggest you to wait for another devs opinions, before rewriting anything; then we can find a common ground before putting more efforts into this feature.

markusstoll:
Here as with talk-desktop:
I do not want anyone to rush things, but I understood correctly that we shall wait for more input from your devs?

ShGKme:
Thank you for your contribution and moving forward to adding the counter-badge in the desktop client!

It seems to work, but I agree with @Antreesy's points.

We should not create a new store state for the unread counters, as this is not an independent state, but a computation on top of the existing state (conversation list). Having it as a new state will likely lead to inconsistent application state in the future and requires manual updates.

This makes sense for performance. However, even with ~1000 conversations counting, there should be no performance issue.

Let's simplify it to one getter in the existing `conversationsStore` and then watch for its changes to emit the global event in the `App.vue`.

ShGKme:
Hey, @markusstoll!

Do you plan to adjust the PRs? Maybe we can assist with anything?

This is an important feature that we'd like to add. If you don't have time now - no problem, but then we'd need to take it over.

markusstoll:
Sorry, I was drawn into some urgent issues in my paid work. I will start the rework soon, hopefully this week

markusstoll:
Hi,

I hope I understood your comments correctly. Pls review the reworked PR - will submit a matching PR for talk-desktop asap

src/App.vue
```diff
@@ -95,6 +96,30 @@ export default {
 	},
 
 	computed: {
+		// Computed properties for unread counts
```
ShGKme, line 99: This comment doesn't have additional value. The code below is `computed`-s because it's literally the next line after `computed: {` and they are for `unread`-s as their names say.
```suggestion
```

src/App.vue
```diff
@@ -95,6 +96,30 @@ export default {
...
+			return Object.values(this.$store.state.conversationsStore.conversations).filter(conv => conv.unreadMention).length;
+		},
+		
+		getTotalUnreadMentionsDirect() {
+			if (!this.$store.state.conversationsStore?.conversations) {
+				return 0;
+			}
+			return Object.values(this.$store.state.conversationsStore.conversations).filter(conv => conv.unreadMentionDirect).length;
+		},
+
```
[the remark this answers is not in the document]
ShGKme: Could you also use `conversation` instead of `conv`? Except for very common cases, we always use full names for variables.

src/App.vue
```diff
@@ -619,6 +661,19 @@ export default {
...
+		 * Emits the UNREAD_COUNT_UPDATED event with the current counter values
+		 */
+		emitUnreadCountUpdated() {
+			const eventData = {
+				totalUnreadMessages: this.getTotalUnreadMessages,
+				totalUnreadMentions: this.getTotalUnreadMentions,
+				totalUnreadMentionsDirect: this.getTotalUnreadMentionsDirect
+			}
+
+			emit(EVENTS.UNREAD_COUNT_UPDATED, eventData)
```
ShGKme, line 675: To simplify naming a bit and remove the constant (see comment in the `talk-desktop`)
- The entire event payload is about counters, the total number of unread _something_, so we can omit it from every property name
- Let's add `conversations` in case a user would like to see the number of conversations, but not total messages in all the conversations
- Event is renamed to `unreads` to be plural (it's "unreads" that were updated)

```suggestion
			emit('talk:unreads:updated', {
				conversations: /* Number of conversations with unread messages */,
				messages: this.getTotalUnreadMessages,
				mentions: this.getTotalUnreadMentions,
				mentionsDirect: this.getTotalUnreadMentionsDirect
			})
```

src/App.vue
```diff
@@ -163,6 +188,17 @@ export default {
 				}
 			}
 		},
+
+		// Watch for changes in unread counters and emit events
```
ShGKme, line 192: Same as with computed-s above
```suggestion
```

markusstoll:
Hi, I did take into account all your remarks and fixed them where needed.

On 
```
It didn't work without emitting here? On updates it's supposed to change conversations data and trigger computing
```
Indeed without initial counter is not shown on badge

On
```
May I ask if this was AI-generated? :)
```

You bet, as being a Java Developer for 25 years, I am mostly illiterate on such fancy JavaScript stuff and need AI support for all this syntax. But here I forgot to remove these additions in conversationsStore.js after deciding for smaller approach - I apologise for that.

After waiting for this feature for 1.5 years, I finally had to do something about it, using the means I have.

ShGKme:
> Indeed without initial counter is not shown on badge

In initial load (`mounted`) it could be necessary, but you also have it in `conversations-received` event handler.

src/App.vue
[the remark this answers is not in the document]
ShGKme: > However, getters in the store took into account archived conversations.
> 
> Here we also need to filter out archived conversations. You can do it by checking `!conversation.isArchived`.

@markusstoll This is still actual

src/App.vue
```diff
@@ -619,6 +651,18 @@ export default {
 				this.$router.push({ name: 'root' })
 			}
 		},
+
+		/**
+		 * Emits the UNREAD_COUNT_UPDATED event with the current counter values
+		 */
+		emitUnreadCountUpdated() {
+			emit('talk:unread:updated', {
```
ShGKme, line 659: As per the comment above
```suggestion
		emitUnreadCountUpdated() {
			emit('talk:unreads:updated', {
```

ShGKme (commented):
Some comments above are still unresolved

markusstoll:
you were right, but pls note the spelling difference, which I changed back to 

```
emit('talk:unreads:updated', {
```

so it matches the doc in constants and the usage in talk-desktop

archived conversations are now ignored

Antreesy (commented):
Nice enhancements! Tested with following suggestions, to reduce amount of computations.

Feel free to squash all commits into one, and name it according to convention spec. something like `feat: <what's it doing>` should be fine. Then I'll trigger the CI to get it merged

src/App.vue
```diff
@@ -95,6 +95,32 @@ export default {
...
+			return this.$store.getters.conversationsList
+				.filter(conversation => !conversation.isArchived && conversation.unreadMentionDirect)
+				.length;
+		},
+
+		getUnreadConversationCount() {
+			return this.$store.getters.conversationsList
+				.filter(conversation => !conversation.isArchived && (conversation.unreadMessages || 0) > 0)
+				.length;
+		},
```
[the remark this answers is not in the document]
ShGKme: In addition, with 4 watches it emitted event 4 times with the same on update.

markusstoll:
Thanks for your comments which made this change even more lean and clean.
I followed your advice, but changed to my proposed event name "talk:unread:updated" which is used in my PR on talk-desktop. As the details in the event data bear the plural I think we it makes sense to stick with singular in the event name.

ShGKme (approved):

Antreesy (approved):
Thanks, Markus 🚀

PR #14714, opened 2025-03-25
feat(conversations): handle event conversation

DorraJaouad:
### ☑️ Resolves

* Fix #14401

### 🖼️ Screenshots / Screencasts

 LS | Filters
-- | --
image | image

image

### 🚧 Tasks

- [x] On frontend/clients, event rooms should not be able to change the title and description (as this should be updated by the CalDAV listener as well). You can change avatar in talk.
- [x] We replace the calendar icon inside of the chat for event rooms (in the topbar) with a symbol that indicates that this room will be archived after 7 days? and actions. The actions should be in the menu, where we currently have "Schedule meeting".
- [x] On frontend: Hide room type EVENT if ~object id is smaller (timestamp - 3600 * 24)~ event is not happening today
- [] ~When a call ends in the room type EVENT, show buttons to archive / delete the room immediately - similar to green "join call" button we have now in the chat area~ It was redundant for the UI

### 🏁 Checklist

- [ ] 🌏 Tested with different browsers / clients:
  - [x] Chromium (Chrome / Edge / Opera / Brave)
  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop
  - [ ] Not risky to browser differences / client
- [x] 🖌️ Design was reviewed, approved or inspired by the design team
- [ ] ⛑️ Tests are included or not possible
- [ ] 📗 User documentation in https://github.com/nextcloud/documentation/tree/master/user_manual/talk has been updated or is not required

PR #14716, opened 2025-03-25
feat(LeftSidebar): transform filters to chips

DorraJaouad:
### ☑️ Resolves

* Pre-requisite to  #14714
* With this approach, we add more filters and they can be used together (like filter group/individual)

🏚️ Before | 🏡 After
-- | --
image | image
image | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -20,7 +20,7 @@
 				<TransitionWrapper name="radial-reveal">
 					<!-- Filters -->
 					<NcActions v-show="searchText === ''"
-						:primary="filters.length !== 0"
+						:type="isFiltered ? 'secondary' : 'tertiary'"
 						class="filters"
 						:class="{'hidden-visually': isSearching}">
 						<template #icon>
@@ -29,7 +29,6 @@
 						<NcActionButton close-after-click
 							type="checkbox"
 							:model-value="filters.includes('mentions')"
-							value="mentions"
 							@click="handleFilter('mentions')">
 							<template #icon>
 								<AtIcon :size="20" />
@@ -39,7 +38,6 @@
 
 						<NcActionButton close-after-click
 							type="checkbox"
-							value="unread"
 							:model-value="filters.includes('unread')"
 							@click="handleFilter('unread')">
 							<template #icon>
@@ -48,7 +46,7 @@
 							{{ t('spreed', 'Filter unread messages') }}
 						</NcActionButton>
 
-						<NcActionButton v-if="filters.length !== 0"
+						<NcActionButton v-if="isFiltered"
 							close-after-click
 							class="filter-actions__clearbutton"
 							@click="handleFilter(null)">
@@ -117,6 +115,15 @@
 				<!-- New Pending Invitations dialog -->
 				<InvitationHandler v-if="pendingInvitationsCount" ref="invitationHandler" />
 			</div>
+			<TransitionWrapper class="conversations__filters"
+				name="zoom"
+				tag="div"
+				group>
+				<NcChip v-for="filter in filters"
+					:key="filter"
+					:text="FILTER_LABELS[filter]"
+					@close="handleFilter(filter)" />
+			</TransitionWrapper>
 			<NcAppNavigationItem v-if="pendingInvitationsCount"
 				class="invitation-button"
 				:name="t('spreed', 'Pending invitations')"
@@ -137,13 +144,13 @@
 					:name="emptyContentLabel"
 					:description="emptyContentDescription">
 					<template #icon>
-						<AtIcon v-if="filters.includes('mentions')" :size="64" />
-						<MessageBadge v-else-if="filters.includes('unread')" :size="64" />
+						<AtIcon v-if="filters.length === 1 && filters[0] === 'mentions'" :size="64" />
+						<MessageBadge v-else-if="filters.length === 1 && filters[0] === 'unread'" :size="64" />
 						<IconArchive v-else-if="showArchived" :size="64" />
 						<MessageOutline v-else :size="64" />
 					</template>
 					<template #action>
-						<NcButton v-if="filters.length !== 0" @click="handleFilter(null)">
+						<NcButton v-if="isFiltered" @click="handleFilter(null)">
 							<template #icon>
 								<FilterRemoveIcon :size="20" />
 							</template>
@@ -244,6 +251,7 @@ import NcActions from '@nextcloud/vue/components/NcActions'
 import NcAppNavigation from '@nextcloud/vue/components/NcAppNavigation'
 import NcAppNavigationItem from '@nextcloud/vue/components/NcAppNavigationItem'
 import NcButton from '@nextcloud/vue/components/NcButton'
+import NcChip from '@nextcloud/vue/components/NcChip'
 import NcCounterBubble from '@nextcloud/vue/components/NcCounterBubble'
 import NcEmptyContent from '@nextcloud/vue/components/NcEmptyContent'
 import { useIsMobile } from '@nextcloud/vue/composables/useIsMobile'
@@ -283,6 +291,11 @@ const canModerateSipDialOut = hasTalkFeature('local', 'sip-support-dialout')
 	&& getTalkConfig('local', 'call', 'can-enable-sip')
 const canNoteToSelf = hasTalkFeature('local', 'note-to-self')
 const supportsArchive = hasTalkFeature('local', 'archived-conversations-v2')
+const FILTER_LABELS = {
+	unread: t('spreed', 'Unread'),
+	mentions: t('spreed', 'Mentions'),
+	default: '',
+}
 
 export default {
 	name: 'LeftSidebar',
@@ -294,6 +307,7 @@ export default {
 		NcAppNavigationItem,
 		NcButton,
 		NcCounterBubble,
+		NcChip,
 		SearchBox,
 		NewConversationDialog,
 		OpenConversationsList,
@@ -326,7 +340,7 @@ export default {
 		const scroller = ref(null)
 
 		const showArchived = ref(false)
-		const filters = ref([])
+		const filters = ref(BrowserStorage.getItem('filterEnabled')?.split(',') ?? [])
 
 		const federationStore = useFederationStore()
 		const talkHashStore = useTalkHashStore()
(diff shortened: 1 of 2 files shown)


src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -113,6 +117,15 @@
 				<!-- New Pending Invitations dialog -->
 				<InvitationHandler v-if="pendingInvitationsCount" ref="invitationHandler" />
 			</div>
+			<TransitionWrapper class="conversations__filters"
+				name="zoom"
+				tag="div"
+				group>
+				<NcChip v-for="filter in filters"
+					:key="filter"
+					:text="t('spreed', filter)"
```
nickvergessen, line 126: this does not work with the translation extraction. you have to have `t('spreed', 'Unread')` and `t('spreed', 'Mentions')` in the source code, otherwise the strings are not listed on transifex

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -20,31 +20,35 @@
...
 							</template>
 							{{ t('spreed', 'Filter unread mentions') }}
 						</NcActionButton>
 
 						<NcActionButton close-after-click
-							:model-value="isFiltered === 'unread'"
-							@click="handleFilter('unread')">
+							type="checkbox"
+							value="unread"
+							:model-value="filters.includes('Unread')"
```
nickvergessen, line 43: this is uppercase

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -133,13 +146,13 @@
 					:name="emptyContentLabel"
 					:description="emptyContentDescription">
 					<template #icon>
-						<AtIcon v-if="isFiltered === 'mentions'" :size="64" />
-						<MessageBadge v-else-if="isFiltered === 'unread'" :size="64" />
+						<AtIcon v-if="filters.includes('mentions')" :size="64" />
+						<MessageBadge v-else-if="filters.includes('unread')" :size="64" />
```
nickvergessen, line 150: this is lowercase?

nickvergessen (dismissed):

src/components/LeftSidebar/LeftSidebar.vue
Antreesy: better to have some `getFilterLabel` function, where we can also put some comments for translators

ShGKme: > this does not work with the translation extraction. you have to have `t('spreed', 'Unread')` and `t('spreed', 'Mentions')` in the source code, otherwise the strings are not listed on transifex

Are you sure?

We use it in many places in many apps.

ShGKme: In 100+ places in Talk

ShGKme: image

ShGKme: Oops

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -534,12 +548,11 @@ export default {
 			if (this.showArchived) {
 				return t('spreed', 'You have no archived conversations.')
 			}
-			switch (this.isFiltered) {
-			case 'mentions':
+			if (this.filters.includes('mentions')) {
```
Antreesy, line 551:
```suggestion
			if (this.filters.length === 1 && this.filters.includes('mentions')) {
```
Otherwise several filters are applied, and you can not say, what user does not have only "...mentions", but not "...messages"

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -726,13 +744,18 @@ export default {
 		},
 
 		handleFilter(filter) {
-			this.isFiltered = filter
 			// Store the active filter
-			if (filter) {
-				BrowserStorage.setItem('filterEnabled', filter)
+			if (filter !== null && this.filters.indexOf(filter) === -1) {
```
Antreesy, line 748:
```suggestion
			if (filter !== null && !this.filters.includes(filter)) {
```

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -1053,6 +1076,15 @@ export default {
...
 		},
+
+		getFilterLabel(filter) {
+			if (filter === 'unread') {
+				return t('spreed', 'Unread')
+			} else if (filter === 'mentions') {
+				return t('spreed', 'Mentions')
+			}
+			return ''
+		},
```
Antreesy, line 1087:
```suggestion
const FILTER_LABELS = {
	unread: t('spreed', 'Unread'),
	mentions: t('spreed', 'Mentions'),
	defult: '',
} as const
```

and then `FILTER_LABELS[filter]`

src/utils/conversation.js
```diff
@@ -55,10 +55,10 @@ export function shouldIncludeArchived(conversation, showArchived) {
...
+ * @param {Array} filters the filter option
  */
-export function filterConversation(conversation, filter) {
-	return filter === null
-		|| (filter === 'unread' && hasUnreadMessages(conversation))
-		|| (filter === 'mentions' && hasUnreadMentions(conversation))
+export function filterConversation(conversation, filters) {
+	return filters.length === 0
+		|| ((!filters.includes('unread') || (filters.includes('unread') && hasUnreadMessages(conversation)))
+		&& (!filters.includes('mentions') || (filters.includes('mentions') && hasUnreadMentions(conversation))))
```
Antreesy, line 63:
```suggestion
		|| ((!filters.includes('unread') || hasUnreadMessages(conversation)))
		&& (!filters.includes('mentions') || hasUnreadMentions(conversation))))
```

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -405,6 +405,11 @@ const canModerateSipDialOut = hasTalkFeature('local', 'sip-support-dialout')
 	&& getTalkConfig('local', 'call', 'can-enable-sip')
 const canNoteToSelf = hasTalkFeature('local', 'note-to-self')
 const supportsArchive = hasTalkFeature('local', 'archived-conversations-v2')
+const FILTER_LABELS = {
+	unread: t('spreed', 'Unread'),
+	mentions: t('spreed', 'Mentions'),
+	defult: '',
```
Antreesy, line 411: no

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -548,9 +554,9 @@ export default {
 			if (this.showArchived) {
 				return t('spreed', 'You have no archived conversations.')
 			}
-			if (this.filters.includes('mentions')) {
+			if (this.filters.length === 1 && this.filters[0] === 'mentions') {
```
Antreesy, line 557: nice, could be used in template as well

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -20,14 +20,16 @@
...
 						class="filters"
 						:class="{'hidden-visually': isSearching}">
 						<template #icon>
 							<FilterIcon :size="15" />
 						</template>
 						<NcActionButton close-after-click
-							:model-value="isFiltered === 'mentions'"
+							type="checkbox"
+							:model-value="filters.includes('mentions')"
+							value="mentions"
```
Antreesy, line 32: value is not needed in both buttons, unless it's `type="radio"`

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -20,14 +20,16 @@
 				<TransitionWrapper name="radial-reveal">
 					<!-- Filters -->
 					<NcActions v-show="searchText === ''"
-						:primary="isFiltered !== null"
+						:type="filters.length !== 0 ? 'secondary' : 'tertiary'"
```
Antreesy, line 23: Maybe we keep `filters.length !== 0` as computed? It's repeated in 5 places
```suggestion
						:type="isFiltered ? 'secondary' : 'tertiary'"
```

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -439,6 +459,7 @@ export default {
 		const scroller = ref(null)
 
 		const showArchived = ref(false)
+		const filters = ref([])
```
Antreesy, line 462: Instead of doing this in `mounted`? I don't see why we should call `handleFilter()`
```suggestion
		const filters = ref(BrowserStorage.getItem('filterEnabled')?.split(',') ?? [])
```

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -726,13 +750,18 @@ export default {
...
+			if (filter !== null && !this.filters.includes(filter)) {
+				this.filters = [...this.filters, filter]
+				BrowserStorage.setItem('filterEnabled', this.filters)
+			} else if (filter !== null) {
+				this.filters = this.filters.filter(f => f !== filter)
+				BrowserStorage.setItem('filterEnabled', this.filters)
 			} else {
+				this.filters = []
 				BrowserStorage.removeItem('filterEnabled')
 			}
```
Antreesy, line 763: wdyt?
```js
if (filter === null) {
	this.filters = []
	BrowserStorage.removeItem('filterEnabled')
} else {
	this.filters = this.filters.includes(filter)
		? this.filters.filter(f => f !== filter)
		: [...this.filters, filter]
	BrowserStorage.setItem('filterEnabled', this.filters)
}
```

nickvergessen:
The second screenshot
[image](https://github.com/user-attachments/assets/3c22432a-a0ac-47d6-84d9-2ab2d7961138)

Is a bit confusing, as you can always only do one. It's unread message or unread mentions.

DorraJaouad:
> Is a bit confusing, as you can always only do one. It's unread message or unread mentions.

You can do both at the same time, but as one (mentions) is a subset of the other (unread), it might look confusing indeed, but it's not the usual use case (users will usually just pick one).

Antreesy:
> but as one (mentions) is a subset of the other (unread)

Then we can hardcode these two values, so one is replacing another, and they work as radio buttons (without changing current behaviour)

DorraJaouad:
so @nickvergessen the idea now is:
- "mentions" and "unread" cannot exist together
- Any other filter (in the future) is combine-able with other filter(s)

Antreesy (approved):
Tested, works as expected. Good to go!

PR #14717, opened 2025-03-25
feat(conversation): Send switch-to signaling message when creating fr…

### LINKED ISSUE #14400: 👥 Adding participants to one-to-one calls - ⚙️ API
- Ref https://github.com/nextcloud/spreed/issues/14398

nickvergessen:
…om one-to-one

### ☑️ Resolves

* Fix #14400 

### 🏁 Checklist

### DIFF
--- a/docs/constants.md
+++ b/docs/constants.md
@@ -12,13 +12,15 @@
 
 ### Object types
 
-| Constant         | Can be created | Description                                                      | Object ID                                               |
-|------------------|----------------|------------------------------------------------------------------|---------------------------------------------------------|
-| `file`           | No             | Conversations about a file in the right sidebar of the files app | File ID                                                 |
-| `share:password` | No             | Video verification to verify the identity of the share recipient | Share token                                             |
-| `room`           | Yes            | Room is a breakout room                                          | Token of the main/parent conversation                   |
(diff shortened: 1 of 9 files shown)


SystemKeeper:
From the code it seems like we can also use it in groups conversations, there’s no check for 1:1?

nickvergessen:
Which I generally would be fine with?

SystemKeeper:
Then it feels weird to have to use an object type that’s called „one-to-one“. But yea, fine in general then.

SystemKeeper:
> Then it feels weird to have to use an object type that’s called „one-to-one“. But yea, fine in general then.

Maybe we should then have something like `objectType = "extend-room"`? Also, we only chat that I am able to access the room, not that I am in the call of that room before switching. Do we need to do that?

nickvergessen:
> Also, we only chat that I am able to access the room, not that I am in the call of that room before switching. Do we need to do that?

That's done in the listener. You can create extended rooms even when not in the call, we just don't switch atm?

SystemKeeper (approved):
Tested between iOS and Web, initiated by iOS -> Worked fine without changes in the clients 👍

miaulalala (approved):

Issue #14418, opened 2025-02-14
☎️ SIP Dial-In - ⚙️ API


nickvergessen:
- Ref https://github.com/nextcloud/spreed/issues/14416

PR #14729, opened 2025-03-27
feat(sip): Allow adding, removing and importing phone numbers

nickvergessen:
### ☑️ Resolves

* Ref #14418 

### 🚧 Tasks

- [x] Validate phone number on add and import
- [x] Validate user id on add and import

### 🏁 Checklist

### DIFF
--- a/lib/Command/PhoneNumber/AddPhoneNumber.php
+++ b/lib/Command/PhoneNumber/AddPhoneNumber.php
@@ -1,0 +1,105 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Talk\Command\PhoneNumber;
+
+use OC\Core\Command\Base;
+use OCA\Talk\Model\PhoneNumber;
+use OCA\Talk\Model\PhoneNumberMapper;
+use OCP\AppFramework\Db\DoesNotExistException;
+use OCP\IPhoneNumberUtil;
+use OCP\IUser;
+use OCP\IUserManager;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class AddPhoneNumber extends Base {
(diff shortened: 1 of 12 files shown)


lib/Command/PhoneNumber/AddPhoneNumber.php
```diff
@@ -0,0 +1,105 @@
...
+		$this
+			->setName('talk:phone-number:add')
+			->setDescription('Add a mapping entry to map a phone number to an account')
+			->addArgument(
+				'phone',
+				InputArgument::REQUIRED,
+				'Phone number that will be called',
+			)
+			->addArgument(
+				'account',
```
miaulalala, line 44:
```suggestion
				'user',
```

Other occ commands use `user`. For occ consistency I would keep with that.

lib/Command/PhoneNumber/ImportPhoneNumbers.php
```diff
@@ -0,0 +1,122 @@
...
+
+		$map = [];
+		while ($row = fgetcsv($handle, escape: '')) {
+			if (count($row) !== 2 || $row[0] === '' || $row[1] === '') {
+				continue;
+			}
+
+			$phoneNumberStandard = $this->phoneNumberUtil->convertToStandardFormat($row[0]);
+			if ($phoneNumberStandard === null) {
+				$output->writeln('<error>Invalid phone number ' . $row[0] . ' provided</error>');
```
miaulalala, line 77:
```suggestion
				$output->writeln('<error>Not a valid phone number ' . $row[0] . '. The format is invalid.</error>');
```

miaulalala (approved):

lib/Migration/Version21001Date20250327161337.php
```diff
@@ -0,0 +1,47 @@
...
+	 * @param Closure(): ISchemaWrapper $schemaClosure
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	#[\Override]
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if (!$schema->hasTable('talk_phone_numbers')) {
```
fancycode, line 29: This will cause the migration to only try to create the table if it already exists.

nickvergessen: arg

PR #14731, opened 2025-03-27
feat(sip): Allow the sipbridge to create a conversation with a user for  a phone number

nickvergessen:
### ☑️ Resolves

- [x] Based on #14729

### 🏁 Checklist

### DIFF
--- a/appinfo/routes/routesRoomController.php
+++ b/appinfo/routes/routesRoomController.php
@@ -95,6 +95,8 @@ return [
 		['name' => 'Room#verifyDialInPin', 'url' => '/api/{apiVersion}/room/{token}/verify-dialin', 'verb' => 'POST', 'requirements' => $requirementsWithToken],
 		/** @see \OCA\Talk\Controller\RoomController::verifyDialOutNumber() */
 		['name' => 'Room#verifyDialOutNumber', 'url' => '/api/{apiVersion}/room/{token}/verify-dialout', 'verb' => 'POST', 'requirements' => $requirementsWithToken],
+		/** @see \OCA\Talk\Controller\RoomController::directDialIn() */
+		['name' => 'Room#directDialIn', 'url' => '/api/{apiVersion}/room/direct-dial-in', 'verb' => 'POST', 'requirements' => $requirements],
 		/** @see \OCA\Talk\Controller\RoomController::createGuestByDialIn() */
 		['name' => 'Room#createGuestByDialIn', 'url' => '/api/{apiVersion}/room/{token}/open-dial-in', 'verb' => 'POST', 'requirements' => $requirementsWithToken],
 		/** @see \OCA\Talk\Controller\RoomController::rejectedDialOutRequest() */
--- a/lib/Controller/RoomController.php
+++ b/lib/Controller/RoomController.php
@@ -1945,6 +1945,10 @@ class RoomController extends AEnvironmentAwareOCSController {
 	#[OpenAPI(scope: 'backend-sipbridge')]
 	#[RequireRoom]
 	public function verifyDialInPin(string $pin): DataResponse {
+		if (!$this->talkConfig->isSIPConfigured()) {
+			return new DataResponse(null, Http::STATUS_NOT_IMPLEMENTED);
+		}
+
 		try {
 			if (!$this->validateSIPBridgeRequest($this->room->getToken())) {
 				$response = new DataResponse(null, Http::STATUS_UNAUTHORIZED);
@@ -1955,10 +1959,6 @@ class RoomController extends AEnvironmentAwareOCSController {
 			$response = new DataResponse(null, Http::STATUS_UNAUTHORIZED);
 			$response->throttle(['action' => 'talkSipBridgeSecret']);
 			return $response;
-		}
-
-		if (!$this->talkConfig->isSIPConfigured()) {
-			return new DataResponse(null, Http::STATUS_NOT_IMPLEMENTED);
 		}
 
 		try {
@@ -1989,6 +1989,10 @@ class RoomController extends AEnvironmentAwareOCSController {
 	#[BruteForceProtection(action: 'talkSipBridgeSecret')]
 	#[OpenAPI(scope: 'backend-sipbridge')]
 	public function directDialIn(string $phoneNumber, string $caller): DataResponse {
+		if (!$this->talkConfig->isSIPConfigured()) {
+			return new DataResponse(null, Http::STATUS_NOT_IMPLEMENTED);
+		}
+
 		try {
 			if (!$this->validateSIPBridgeRequest($phoneNumber)) {
 				$response = new DataResponse(null, Http::STATUS_UNAUTHORIZED);
@@ -1999,10 +2003,6 @@ class RoomController extends AEnvironmentAwareOCSController {
 			$response = new DataResponse(null, Http::STATUS_UNAUTHORIZED);
 			$response->throttle(['action' => 'talkSipBridgeSecret']);
 			return $response;
-		}
-
-		if (!$this->talkConfig->isSIPConfigured()) {
-			return new DataResponse(null, Http::STATUS_NOT_IMPLEMENTED);
 		}
 
 		try {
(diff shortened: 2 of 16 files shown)


lib/Controller/RoomController.php
```diff
@@ -1907,6 +1910,50 @@ public function verifyDialInPin(string $pin): DataResponse {
...
 
+	/**
+	 * Direct dial-in (SIP bridge)
+	 *
+	 * @param string $phoneNumber Phone number that is called
+	 * @return DataResponse<Http::STATUS_OK, TalkRoom, array{}>|DataResponse<Http::STATUS_UNAUTHORIZED|Http::STATUS_NOT_FOUND|Http::STATUS_NOT_IMPLEMENTED, null, array{}>
+	 *
+	 * 200: Participant returned
+	 * 401: SIP request invalid
+	 * 404: Participant not found
```
fancycode, line 1921:
```suggestion
	 * 404: Participant not found
     * 486: Participant is currently busy
```

lib/Controller/RoomController.php
```diff
@@ -1907,6 +1910,50 @@ public function verifyDialInPin(string $pin): DataResponse {
...
+	 * 501: SIP dial-in is not configured
+	 */
+	#[PublicPage]
+	#[BruteForceProtection(action: 'talkSipBridgeSecret')]
+	#[OpenAPI(scope: 'backend-sipbridge')]
+	#[RequireRoom]
+	public function directDialIn(string $phoneNumber): DataResponse {
```
fancycode, line 1928: This should also receive the number of the caller so the rooms could be returned depending on who is calling.

lib/Controller/RoomController.php
nickvergessen: I'm finding it difficult to see this behaviour without a user option?
It's like "should there be knocking" or not. And I think the answer is "it depends".
Depends on the personal preference, on the person that is calling, and the call I'm currently in.
Should it be a user option?

lib/Controller/RoomController.php
nickvergessen: @fancycode when someone calls with masked number, you send us empty string? or null, or …?
Since the current take is "Reuse room for same number" I kind of need to know when not to look but create a new conversation instead.

fancycode: Well, problem is there is no clear definition for that case in SIP. Some providers send "Anonymous" / "anonymous", some send an empty number, etc.

Maybe in a first step you can handle all non-numeric numbers as "anonymous"? In the future there might be cases where this should be supported (e.g. direct calling from "known" devices where the SIP provider might send the name of the user that is calling as display name in the call, e.g. `"John Doe" <sip:1234@asterisk.sipprovider.invalid>`).

nickvergessen: Okay, so you don't get to parse that info before, or you don't want to?
We can also `lowercase and in array['anonymous']` with the array being a config administrators can extend, based on user feedback?

lib/Controller/RoomController.php
nickvergessen: 486 is currently not supported :see_no_evil: 
We can add it to the openapi extractor and the Server APIs if we want, but I'm a bit unsure if it's needed as per above

lib/Controller/RoomController.php
```diff
@@ -1923,6 +1926,77 @@ public function verifyDialInPin(string $pin): DataResponse {
...
+			}
+		} else {
+			$objectId = Room::OBJECT_TYPE_PHONE;
+		}
+
+		if (!$room instanceof Room) {
+			$user = $this->userManager->get($entity->getActorId());
+			try {
+				$room = $this->roomService->createConversation(
+					Room::TYPE_GROUP,
```
nickvergessen, line 1985: @fancycode can you give it a try if group is okay, or whether we need:
```suggestion
					Room::TYPE_PUBLIC,
```

appinfo/routes/routesRoomController.php
```diff
@@ -91,6 +91,8 @@
 		['name' => 'Room#verifyDialInPin', 'url' => '/api/{apiVersion}/room/{token}/verify-dialin', 'verb' => 'POST', 'requirements' => $requirementsWithToken],
 		/** @see \OCA\Talk\Controller\RoomController::verifyDialOutNumber() */
 		['name' => 'Room#verifyDialOutNumber', 'url' => '/api/{apiVersion}/room/{token}/verify-dialout', 'verb' => 'POST', 'requirements' => $requirementsWithToken],
+		/** @see \OCA\Talk\Controller\RoomController::directDialIn() */
+		['name' => 'Room#directDialIn', 'url' => '/api/{apiVersion}/room/{token}/direct-dial-in', 'verb' => 'POST', 'requirements' => $requirementsWithToken],
```
fancycode, line 95: Which `token` should I send here? The room does not exist at the time of the request.

lib/Controller/RoomController.php
fancycode: `TYPE_GROUP` works (I can hear the phone in the web and vice versa). Didn't test `TYPE_PUBLIC`.

lib/Controller/RoomController.php
```diff
@@ -1923,6 +1926,78 @@ public function verifyDialInPin(string $pin): DataResponse {
...
+					Room::OBJECT_TYPE_PHONE,
+					$objectId,
+					sipEnabled: Webinary::SIP_ENABLED_NO_PIN,
+				);
+			} catch (CreationException|PasswordException $e) {
+				$this->logger->error('Error occurred while creating SIP direct dial-in conversation', ['exception' => $e]);
+			}
+		}
+
+		$participant = $this->participantService->joinRoomAsNewGuest($this->roomService, $this->room, '', true);
```
fancycode, line 1997:
```suggestion
		$participant = $this->participantService->joinRoomAsNewGuest($this->roomService, $room, '', true);
```

fancycode:
We probably need a dedicated response code for when the regular call workflow should be used, or should that always happen on `404`?

appinfo/routes/routesRoomController.php
nickvergessen: that's a correct oversight. Should be a URL without a token

nickvergessen:
> or should that always happen on 404?

I'm not really sure I can distinguish a random number from a workflow one. So yeah I'd say 404 means there is no direct-dialin => normal workflow?

nickvergessen:
- Conversation is not marked as "In call" yet
- Neither does the called user receive a notification

nickvergessen:
/backport a5eee091917d2544939bd99cfa7497e3a3d18f5c to stable31.0

fancycode:
Web-UI tries to call the phone user when clicking the button to join the call.

POST-URL is `https://myserver/ocs/v2.php/apps/spreed/api/v4/call/<token>/dialout/%7BattendeeId%7D`.

lib/Controller/RoomController.php
```diff
@@ -1963,6 +1970,67 @@ public function verifyDialInPin(string $pin): DataResponse {
...
+				Room::OBJECT_TYPE_PHONE,
+				Room::OBJECT_ID_PHONE_INCOMING,
+				sipEnabled: Webinary::SIP_ENABLED_NO_PIN,
+			);
+		} catch (CreationException|PasswordException $e) {
+			$this->logger->error('Error occurred while creating SIP direct dial-in conversation', ['exception' => $e]);
+			return new DataResponse(null, Http::STATUS_INTERNAL_SERVER_ERROR);
+		}
+
+		$participant = $this->participantService->joinRoomAsNewGuest($this->roomService, $room, '', true);
```
fancycode, line 2030: Maybe also set the display name of the participant to the (anonymized) number of the caller?

nickvergessen: Fixed

nickvergessen:
> The triggered dial-out when joining from the web should be fixed imho, the display name is nice to have.

It was done from frontend, but fixed it myself

lib/Controller/RoomController.php
```diff
@@ -1950,17 +1961,75 @@ public function verifyDialInPin(string $pin): DataResponse {
...
+	 * 404: Number is not assigned to any user
+	 * 500: Error occurred while creating conversation
+	 * 501: SIP dial-in is not configured
+	 */
+	#[PublicPage]
+	#[BruteForceProtection(action: 'talkSipBridgeSecret')]
+	#[OpenAPI(scope: 'backend-sipbridge')]
+	public function directDialIn(string $phoneNumber, string $caller): DataResponse {
 		if (!$this->talkConfig->isSIPConfigured()) {
 			return new DataResponse(null, Http::STATUS_NOT_IMPLEMENTED);
```
miaulalala, line 1993: should we also throttle the "not implemented" response?

miaulalala (approved):

PR #14738, opened 2025-03-28
feat(conversations): Add "Important conversations" which still notify…

nickvergessen:
… during DND

### ☑️ Resolves

* Ref https://github.com/nextcloud/notifications/pull/2059
* Ref https://github.com/nextcloud/server/pull/48274

### 🏁 Checklist

### DIFF
--- a/lib/Controller/RoomController.php
+++ b/lib/Controller/RoomController.php
@@ -1682,6 +1682,40 @@ class RoomController extends AEnvironmentAwareOCSController {
 	}
 
 	/**
+	 * Mark a conversation as important (still sending notifications while on DND)
+	 *
+	 * Required capability: `important-conversations`
+	 *
+	 * @return DataResponse<Http::STATUS_OK, TalkRoom, array{}>
+	 *
+	 * 200: Conversation was marked as important
+	 */
+	#[NoAdminRequired]
+	#[FederationSupported]
+	#[RequireLoggedInParticipant]
+	public function markConversationAsImportant(): DataResponse {
+		$this->participantService->markConversationAsImportant($this->participant);
+		return new DataResponse($this->formatRoom($this->room, $this->participant));
+	}
+
+	/**
+	 * Mark a conversation as unimportant (no longer sending notifications while on DND)
+	 *
+	 * Required capability: `important-conversations`
+	 *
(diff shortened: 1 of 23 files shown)


lib/Notification/Listener.php
```diff
@@ -275,11 +275,11 @@ protected function sendCallNotifications(Room $room): void {
 		}
 
 		$this->preparedCallNotifications = [];
-		$userIds = $this->participantsService->getParticipantUserIdsForCallNotifications($room);
+		$users = $this->participantsService->getParticipantUsersForCallNotifications($room);
```
miaulalala, line 278:
```suggestion
        $userIds = $this->participantsService->getParticipantUsersForCallNotifications($room);
```

or rename the variable in L291

lib/Service/ParticipantService.php
```diff
@@ -1956,14 +1976,14 @@ public function getParticipantUserIdsForCallNotifications(Room $room): array {
 			);
 		}
 
-		$userIds = [];
+		$users = [];
 		$result = $query->executeQuery();
 		while ($row = $result->fetch()) {
-			$userIds[] = $row['actor_id'];
+			$users[$row['actor_id']] = (bool)$row['important'];
```
miaulalala, line 1982: What happens if the user is found twice, once with `important===true` and once not? Is that possible?

nickvergessen: Unique constraint on DB says no

lib/Notification/Listener.php
nickvergessen: Have to take array keys in 291.
Fixed

lib/Controller/RoomController.php
```diff
@@ -1681,6 +1681,40 @@ public function unarchiveConversation(): DataResponse {
 		return new DataResponse($this->formatRoom($this->room, $this->participant));
 	}
 
+	/**
+	 * Mark a conversation as important (still sending notifications while on DND)
+	 *
+	 *  Required capability: `important-conversations`
```
miaulalala, line 1687: *nit*
```suggestion
	 * Required capability: `important-conversations`
```

miaulalala (approved):

PR #14763, opened 2025-04-01
fix(CalendarEventsDialog): adjust UI based on conversation type

DorraJaouad:
### ☑️ Resolves

- Ref: https://github.com/nextcloud/spreed/issues/14802
  - [x] Considered closed? Button is not shown by default in group and 1-1 conversation; only shown when unselect all, then you need to "add attendees" to the meeting, so they receive invites

🏚️ Before | 🏡 After
-- | --
calendar-dialog-before | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/CalendarEventsDialog.vue
+++ b/src/components/CalendarEventsDialog.vue
@@ -36,7 +36,7 @@ import SearchBox from './UIShared/SearchBox.vue'
 import TransitionWrapper from './UIShared/TransitionWrapper.vue'
 
 import { useStore } from '../composables/useStore.js'
-import { ATTENDEE } from '../constants.ts'
+import { ATTENDEE, CONVERSATION } from '../constants.ts'
 import { hasTalkFeature } from '../services/CapabilitiesManager.ts'
 import { useGroupwareStore } from '../stores/groupware.ts'
 import type { Conversation, Participant } from '../types/index.ts'
@@ -158,11 +158,11 @@ const attendeeHint = computed(() => {
 const searchText = ref('')
 const isMatch = (string: string = '') => string.toLowerCase().includes(searchText.value.toLowerCase())
 
+const conversation = computed<Conversation>(() => store.getters.conversation(props.token))
 const participants = computed(() => {
-	const conversation: Conversation = store.getters.conversation(props.token)
 	return store.getters.participantsList(props.token).filter((participant: Participant) => {
 		return [ATTENDEE.ACTOR_TYPE.USERS, ATTENDEE.ACTOR_TYPE.EMAILS].includes(participant.actorType)
-			&& participant.attendeeId !== conversation.attendeeId
+			&& participant.attendeeId !== conversation.value.attendeeId
 	})
 })
 const participantsInitialised = computed(() => store.getters.participantsInitialised(props.token))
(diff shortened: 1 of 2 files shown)


src/components/CalendarEventsDialog.vue
```diff
@@ -407,13 +412,13 @@ async function submitNewMeeting() {
 						{{ option.label }}
 					</template>
 				</NcSelect>
-				<h5 class="calendar-meeting__header">
+				<h5 v-if="!isOneToOneConversation" class="calendar-meeting__header">
 					{{ t('spreed', 'Attendees') }}
 				</h5>
 				<NcCheckboxRadioSwitch v-model="selectAll" @update:modelValue="toggleAll">
-					{{ t('spreed', 'Invite all users and emails') }}
+					{{ !isOneToOneConversation? t('spreed', 'Invite all users and emails') : t('spreed', 'Invite user') }}
```
Antreesy, line 419: We just recently discussed with design team, that 'user' should be avoided. How about:
```suggestion
					{{ !isOneToOneConversation ? t('spreed', 'Invite all users and emails') : t('spreed', 'Invite {user}', user: <secondParticipant.displayName>) }}
```
And move it to script?

src/components/CalendarEventsDialog.vue
```diff
@@ -421,17 +421,24 @@ async function submitNewMeeting() {
...
-					</template>
-					{{ t('spreed', 'Add attendees') }}
-				</NcButton>
-				<p>{{ attendeeHint }}</p>
-
+				<NcLoadingIcon v-if="!participantsInitialised"
+					class="calendar-meeting--loading"
+					:name="t('spreed', 'Loading participants')" />
+				<p v-else-if="participantsInitialised && participants.length === 0">
+					{{ t('spreed', 'You are the only participant in the conversation') }}
```
Antreesy, line 428: Could be guests / federated users. Maybe:
```suggestion
					{{ t('spreed', 'No other participants to send invitations to') }}
```

src/components/CalendarEventsDialog.vue
```diff
@@ -629,6 +636,10 @@ async function submitNewMeeting() {
 		margin-block: auto !important;
 	}
 
+	&--loading {
+		height: 32px;
```
Antreesy, line 640: align to the left side, there all other elements are?
also `<NcLoadingIcon :size="32"`

src/components/CalendarEventsDialog.vue
```diff
@@ -421,17 +421,24 @@ async function submitNewMeeting() {
...
-						<IconAccountPlus :size="20" />
-					</template>
-					{{ t('spreed', 'Add attendees') }}
-				</NcButton>
-				<p>{{ attendeeHint }}</p>
-
+				<NcLoadingIcon v-if="!participantsInitialised"
+					class="calendar-meeting--loading"
+					:name="t('spreed', 'Loading participants')" />
+				<p v-else-if="participantsInitialised && participants.length === 0">
```
Antreesy, line 427:
```suggestion
				<p v-else-if="participants.length === 0">
``

Antreesy (approved):
Tested, nothing blocking

src/components/CalendarEventsDialog.vue
DorraJaouad: the height is flickering that's why I fix it in css

Antreesy: Could be for this component 🙈
Then at least align height with the `1lh`. Otherwise it's jumpy, when comes from loading (32px) to no participants (22.5px)

Antreesy (approved):
Rebased

PR #14771, opened 2025-04-02
fix(performance): Parse file shares inaccurately to prevent setupFS

nickvergessen:
In some cases we only use the "plain text" version of a chat message. We can render the conversation list subline without knowing the exact path and permissions of a file. All other information is still available on the cache entry. The path is only needed in some special cases when afterwards the viewer is proposed to the user.

## :footprints: Steps
- As user1 create a conversation with only yourself and upload test.txt
- As user2 create a conversation with user1 and upload test.txt
- As user2 go to your Files app and change the share permissions to view-only
- As user1 check the message parameters from the last message of the conversation and compare them with the message parameters from the chat request: path should be "test.txt instead of Talk/test.txt" and permissions are "27 instead of 1"

image

### 🏁 Checklist

### DIFF
--- a/lib/Chat/MessageParser.php
+++ b/lib/Chat/MessageParser.php
@@ -66,7 +66,12 @@ class MessageParser {
 		return $message;
 	}
 
-	public function parseMessage(Message $message): void {
+	/**
+	 * @param bool $allowInaccurate File share messages will not have fully correct data for the file object
+	 *                              E.g. path is only the file name and preview generation is estimated by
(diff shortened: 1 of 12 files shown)


Antreesy (approved):
Tested from the frontend side, IMO acceptable sacrifice. Put a screenshot in PR description

miaulalala (approved):
Agree with Maksim, the inaccuracy is well worth it for the performance improvement of not setting up the FS

PR #14772, opened 2025-04-02
feat(conversations): Allow to not request the last message when getting rooms

nickvergessen:
This is useful e.g. when only showing a compact list

### 🏁 Checklist


Antreesy (approved):
Tested with frontend, see follow-up PR

DorraJaouad (approved):
Tested, nice addition 👍🏻

Antreesy:
> does this need a capability?

If backported to stable31 only: we can rely on: `config => conversations => list-style`?

nickvergessen:
From my POV it doesn't need one. You can always send the parameter. It will just have no affect on the performance if the server doesn't consume it.

> we can rely on: config => conversations => list-style?

That wouldn't work as there are releases out that have the capability but not the feature.
But as said, I don't think we need it.

PR #14773, opened 2025-04-02
feat(conversations): do not request last message when in compact mode

Antreesy:
### ☑️ Resolves

* Follow-up to #14772

https://github.com/user-attachments/assets/261dfd22-a199-4036-a909-6d492fa7d616

### 🚧 TODO
* Has no effect without backend, so should be safe for Desktop client?
* Shows a fallback for a moment when switching from 'compact-mode' - acceptable?

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -461,6 +461,14 @@ export default {
 				this.isNavigating = true
 			}
 		},
+		isCompact(value) {
+			if (!value) {
+				// Last messages are likely missing from the store, need to fetch with modifiedSince=0
+				this.roomListModifiedBefore = 0
+				this.forceFullRoomListRefreshAfterXLoops = 10
+				this.fetchConversations()
+			}
+		}
 	},
 
 	beforeMount() {
@@ -759,6 +767,7 @@ export default {
 			try {
 				const response = await this.$store.dispatch('fetchConversations', {
 					modifiedSince: this.roomListModifiedBefore,
+					includeLastMessage: this.isCompact ? 0 : 1,
 				})
 
 				// We can only support this with the HPB as otherwise rooms,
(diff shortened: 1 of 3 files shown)


ShGKme (approved):

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -461,6 +461,11 @@ export default {
 				this.isNavigating = true
 			}
 		},
+		isCompact() {
+			this.roomListModifiedBefore = 0
+			this.forceFullRoomListRefreshAfterXLoops = 10
+			this.fetchConversations()
+		}
```
DorraJaouad, line 468: We don't need to fetch when list becomes compact but only when it comes back to normal mode

DorraJaouad (commented):
Tested, no errors and also we have established that `lastMessage` can be absent since the introduction of federation so no other changes needed.

src/components/LeftSidebar/LeftSidebar.vue
Antreesy: Good point. But then need to check, if last messages retain in the store, and if it might interfere with some other logic

DorraJaouad: If you are using the fetch request to wipe last message, then make a method to clean it instead

nickvergessen (approved):
But agree with Dorra, we don't have to rebuild when you disable it.

src/components/LeftSidebar/LeftSidebar.vue
Antreesy: Nah. Last messages will be cleared after full fetch of the room:
- for 1-1 rooms in the next 30 seconds,
- for group (no activity) rooms in the next 5 minutes,
- after page reload, they do not survive from browser storage

PR #14795, opened 2025-04-04
fix(outOfOffice): show expand button during call

### LINKED ISSUE #14457: can not expand out-of-office notice
Hi all

I just noticed that I can't expand the out-of-office notice on the desktop client, the arrow at the top right is missing:

Desktop Client

Image

Web:

Image

DorraJaouad:
### ☑️ Resolves

* Fix #14457

🏚️ Before | 🏡 After
-- | --
image | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/NewMessage/NewMessageAbsenceInfo.vue
+++ b/src/components/NewMessage/NewMessageAbsenceInfo.vue
@@ -92,6 +92,7 @@ export default {
 		return {
 			collapsed: true,
 			isTextMoreThanOneLine: false,
+			resizeObserver: null,
 		}
 	},
 
@@ -122,16 +123,16 @@ export default {
 		},
 	},
 
-	watch: {
-		userAbsenceMessage() {
-			this.$nextTick(() => {
-				this.setIsTextMoreThanOneLine()
-			})
-		},
-	},
-
 	mounted() {
 		this.setIsTextMoreThanOneLine()
+		this.resizeObserver = new ResizeObserver(this.setIsTextMoreThanOneLine)
+		this.resizeObserver.observe(this.$refs.absenceMessage)
+	},
+
+	beforeDestroy() {
+		if (this.resizeObserver) {
+			this.resizeObserver.disconnect()
+		}
 	},
 
 	methods: {
@@ -141,6 +142,9 @@ export default {
 		},
 
 		setIsTextMoreThanOneLine() {
+			if (!this.collapsed) {
+				return
+			}
 			this.isTextMoreThanOneLine = this.$refs.absenceMessage?.scrollHeight > this.$refs.absenceMessage?.clientHeight
 		},
 


src/components/NewMessage/NewMessageAbsenceInfo.vue
```diff
@@ -128,6 +132,14 @@ export default {
...
 			})
 		},
+
+		isVisible(value) {
+			if (value) {
+				this.$nextTick(() => {
+					this.setIsTextMoreThanOneLine()
+				})
+			}
+		},
```
ShGKme, line 142: If I understand it correctly, the same problem happens if a user resizes the window...
So that the source of the problem is changed width, not visibility.

DorraJaouad: hmmm you are not wrong

src/components/ChatView.vue
```diff
@@ -107,7 +109,8 @@ export default {
 		},
 	},
 
-	setup() {
+	setup(props) {
+		provide('isVisible', computed(() => props.isVisible))
```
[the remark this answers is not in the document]
DorraJaouad: Noted for future use ;)

ShGKme (approved):

ShGKme:
A complex solution - render it twise - once hidden for scrollHeight check and once for the user

src/components/NewMessage/NewMessageAbsenceInfo.vue
```diff
@@ -122,16 +123,16 @@ export default {
...
 	mounted() {
 		this.setIsTextMoreThanOneLine()
+		this.resizeObserver = new ResizeObserver(this.setIsTextMoreThanOneLine)
+		this.resizeObserver.observe(this.$refs.absenceMessage)
+	},
+
+	beforeDestroy() {
+		if (this.resizeObserver) {
+			this.resizeObserver.disconnect()
+		}
```
Antreesy, line 135: you don't have conditions for starting observer, so no need to check when destroying?
```suggestion
		this.resizeObserver.disconnect()
```

src/components/NewMessage/NewMessageAbsenceInfo.vue
```diff
@@ -122,16 +123,16 @@ export default {
 		},
 	},
 
-	watch: {
-		userAbsenceMessage() {
-			this.$nextTick(() => {
-				this.setIsTextMoreThanOneLine()
```
Antreesy, line 128: When switching between two one-to-ones conversations, component is not unmounted, only updated:
- `resize` event is triggered, because `userAbsenceMessage` is changed and therefore a width of DOM element might have changed. 
- if same person has changed absense message in the meantime fo something longer, value of this computed will change, but not the element. Then you will miss the button again.

Let's keep the watcher?
Should width be always 100%, maybe, to cause less resize emits?

src/components/NewMessage/NewMessageAbsenceInfo.vue
ShGKme: Or use `useResizeObserver` from vueuse :D

src/components/NewMessage/NewMessageAbsenceInfo.vue
DorraJaouad: > if same person has changed absense message in the meantime fo something longer, value of this computed will change, but not the element. Then you will miss the button again.

this will trigger the resize event

src/components/NewMessage/NewMessageAbsenceInfo.vue
Antreesy: It's not used anywhere in project yet.. but we can keep it in mind for ts migration in each case

src/components/NewMessage/NewMessageAbsenceInfo.vue
Antreesy: Ok, can't reproduce it anymore

Antreesy (approved):

src/components/NewMessage/NewMessageAbsenceInfo.vue
ShGKme: TS?

Antreesy: I meant `script setup` 👀. But we do it together usually

ShGKme: We can use composables in Options API, even mixing with properties from Options API.

PR #14813, opened 2025-04-07
fix(LeftSidebar): conversation subline hardening

Antreesy:
### ☑️ Resolves

* escape possible HTML syntax in display names
* migrate to TS:
  * `hasLastMessage` ensure that `item.value.lastMessage!` always exists
* check that repeated mentions are parsed by `subtitle.replaceAll`

🏚️ Before | 🏡 After
-- | --
image | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


src/composables/useConversationInfo.ts
```diff
@@ -56,32 +64,32 @@ export function useConversationInfo({
...
+	 * @return Part of the name until the first space
 	 */
 	const shortLastChatMessageAuthor = computed(() => {
-		if (!exposeMessages || !hasLastMessage.value || item.value.lastMessage.systemMessage.length) {
+		if (!exposeMessages || !hasLastMessage.value || item.value.lastMessage!.systemMessage.length) {
 			return ''
 		}
 
-		const author = item.value.lastMessage.actorDisplayName.trim().split(' ').shift()
+		const author = item.value.lastMessage!.actorDisplayName.trim().split(' ').at(0)
```
ShGKme, line 86: ?
```suggestion
		const author = item.value.lastMessage!.actorDisplayName.trim().split(' ')[0]
```

src/composables/useConversationInfo.ts
```diff
@@ -56,32 +64,32 @@ export function useConversationInfo({
 			return ''
 		}
 
-		const params = item.value?.lastMessage.messageParameters
-		let subtitle = (getMessageIcon(item.value?.lastMessage) + ' ' + escapeHtml(item.value?.lastMessage.message)).trim()
+		const params = item.value.lastMessage!.messageParameters
```
ShGKme, line 67: General TS note:

We can define a new variable for the last message. Then we don't need to add `!` in all the places where we use it, and there is a place for a comment to explain, why `!` is added here.

```ts
/* Without reactivity */

// lastMessage is guaranteed to be defined becase <proof>
const lastMessage = item.value.lastMessage!

/* With reactivity */

// lastMessage is guaranteed to be defined becase <proof>
const lastMessage = toRef(() => item.value.lastMessage!)
```

ShGKme: `vueuse` provide `toRef` for compatibility

DorraJaouad (approved):
Tested ✅

src/composables/useConversationInfo.ts
ShGKme: `computed` is also ok, it's just slightly slower (on nano-optimization level) and has a different purpose semantically (lazy derived computing).

Antreesy: I haven't tried, but wouldn't that yield an error on composable call?

As computed is lazy, it won't be called until we asked for it (which we only do after ensuring lastMessage exists)

Issue #14807, opened 2025-04-06
Performance of `getRooms`


SystemKeeper:
Just some things I noticed while logging all DB queries on `getRooms` call, feel free to close.

### Last message is queried individually

Every last message is queried individually, like this:

```
DB QueryBuilder: 'SELECT * FROM `*PREFIX*comments` WHERE `id` = :id' with parameters: id => '990'
DB QueryBuilder: 'SELECT * FROM `*PREFIX*comments` WHERE `id` = :id' with parameters: id => '997'
...
```

Can we load these messages in a single query?

### Preload shares and files metadata

While the latest changes significantly improved performance with not setting up the FS, for every last message that is a share, we query the share and the files metadata:

```
DB QueryBuilder: 'SELECT `s`.*, `f`.`fileid`, `f`.`path`, `f`.`permissions` as `f_permissions`, `f`.`storage`, `f`.`path_hash`, `f`.`parent` as `f_parent`, `f`.`name`, `f`.`mimetype`, `f`.`mimepart`, `f`.`size`, `f`.`mtime`, `f`.`storage_mtime`, `f`.`encrypted`, `f`.`unencrypted_size`, `f`.`etag`, `f`.`checksum`, `st`.`id` AS `storage_string_id` FROM `*PREFIX*share` `s` LEFT JOIN `*PREFIX*filecache` `f` ON `s`.`file_source` = `f`.`fileid` LEFT JOIN `*PREFIX*storages` `st` ON `f`.`storage` = `st`.`numeric_id` WHERE (`s`.`id` IN (:dcValue1)) AND (`s`.`share_type` = :dcValue2)' with parameters: dcValue1 => ('18'), dcValue2 => '10'
DB QueryBuilder: 'SELECT `json`, `sync_token` FROM `*PREFIX*files_metadata` WHERE `file_id` = :dcValue1' with parameters: dcValue1 => '847'
```

We already have a `perloadShares` method for messages, which loads the shares and the files metdata. Can we do the same here ?
https://github.com/nextcloud/spreed/blob/3afc9c85526916bf04be579f5d49bd3cc41c5967/lib/Controller/ChatController.php#L348

### Addressbook queried

```
DB QueryBuilder: 'SELECT * FROM `*PREFIX*dav_cal_proxy` WHERE `proxy_id` = :dcValue1' with parameters: dcValue1 => 'principals/users/admin'
[...]

PR #14829, opened 2025-04-09
fix(performance): Improve performance with many file shares as last m…

nickvergessen:
…essage

### ☑️ Resolves

* Second item "Preload shares and files metadata" from #14807 
* Performance integration test added in https://github.com/nextcloud/spreed/pull/14812

Before | After
---|---
17 + NumShares queries + NumPhotos queries | 17 + 1 (per 1k shares) + 1 (per 1k photos) queries

### 🏁 Checklist


miaulalala (approved):
very nice

PR #14847, opened 2025-04-10
Small feature added: Edit last message on cursor up

markusstoll:
This is a minor change, adding a most common feature as seen in many other chat systems:

If input field for "new message" is empty, "cursor up" will trigger the "edit last message function"

### DIFF
--- a/src/components/NewMessage/NewMessage.vue
+++ b/src/components/NewMessage/NewMessage.vue
@@ -85,6 +85,7 @@
 					@keydown.esc="handleInputEsc"
 					@keydown.ctrl.up="handleEditLastMessage"
 					@keydown.meta.up="handleEditLastMessage"
+					@keydown.up="handleUpKeyInEmptyInput"
 					@input="handleTyping"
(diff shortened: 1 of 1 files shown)


nickvergessen:
Hi @markusstoll 

we already have Ctrl+ArrowUp to achieve exactly that.
In the future we plan to use arrow keys to navigate the message list for accessibility reasons. Arrow up from the chat input would then be the "navigate to the last message in the chat".

In order to not break this second way of editing the last message, I would therefore like to not add it now. I will check with the frontend developers and let them have the final say, but that's my current suggestion.

markusstoll:
Not adding for now and letting you deal with this otherwise is totally fine for me.
Let me just add, that I think this CMD-Cursor-Up feature is not really obvious.
I maintain 2 NextcloudTalk instance with about 30 users (mostly developers) and got from multiple users this feature request (including from our UX expert and they complained about how clumsy it is when you mistyped something to open up the editor and obviously they did not find the CMD-Up feature)

PR #14730, opened 2025-03-27
feat: allow to create new conversation from one-to-one

Antreesy:
### ☑️ Resolves

* Fix #14399
	* New util "getDisplayNamesList" to format display name lists with length constraints.
	* Introduce the "extendOneToOneConversation" store action
	* Show the "Participants" button in one-to-one
	* Allow to search for user participant only in the one-to-one call
	* Allow to lock pre-selected participant when creating a new group conversation
	* New popover to extend one-to-one conversations in the TopBar.

### 🖼️ Screenshots / Screencasts

From LeftSidebar (locking group/team)
image

From TopBar (in chat)
image

From RightSidebar (in call)
image
image
image

### 🏁 Checklist

- [ ] 🌏 Tested with different browsers / clients:
  - [ ] Chromium (Chrome / Edge / Opera / Brave)
  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop
  - [ ] Not risky to browser differences / client
- [ ] 🖌️ Design was reviewed, approved or inspired by the design team
- [ ] ⛑️ Tests are included or not possible
- [ ] 📗 User documentation in https://github.com/nextcloud/documentation/tree/master/user_manual/talk has been updated or is not required

PR #14897, opened 2025-04-17
fix(conversations): do not add second participant to extended 1-1 twice

Antreesy:
### ☑️ Resolves

* Fix regression from #14730
* In store:
  * second participant was duplicated from event dialog -> removed
* In conversation:
  * second participant of 1-1 was passed to vuex store with other newParticipants -> kept
* In call:
  * only one new participant always passed -> add original

🏚️ Before | 🏡 After
-- | --
image | image
a | 2025-04-24_14h47_092025-04-24_14h47_22

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/RightSidebar/Participants/ParticipantsTab.vue
+++ b/src/components/RightSidebar/Participants/ParticipantsTab.vue
@@ -332,7 +332,10 @@ export default {
 				if (this.isOneToOneConversation) {
 					await this.$store.dispatch('extendOneToOneConversation', {
 						token: this.token,
-						newParticipants: [participant],
+						newParticipants: [
+							{ id: this.conversation.name, source: ATTENDEE.ACTOR_TYPE.USERS, label: this.conversation.displayName },
+							participant,
+						],
 					})
 				} else {
 					await addParticipant(this.token, participant.id, participant.source)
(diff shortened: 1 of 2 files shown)


src/components/ExtendOneToOneDialog.vue
```diff
@@ -59,9 +59,12 @@ function getArrayWithSecondAttendee(token: string) {
  */
 async function extendOneToOneConversation() {
 	try {
+		const secondParticipant = getArrayWithSecondAttendee(props.token)[0]
 		const newConversation = await store.dispatch('extendOneToOneConversation', {
 			token: props.token,
-			newParticipants: selectedParticipants.value,
+			newParticipants: selectedParticipants.value.filter(participant => {
+				return !(participant.id === secondParticipant.id && participant.source === secondParticipant.source)
+			}),
```
DorraJaouad, line 67: or remove `{ id: conversation.name, source: ATTENDEE.ACTOR_TYPE.USERS, label: conversation.displayName },` from the function in store as that's the root of duplication?

Antreesy: Then we need to add the same participant to the right sidebar, when calling in 1-1 🙄

DorraJaouad (approved):

Issue #14413, opened 2025-02-14
📇 Improve right sidebar in one-to-one conversations - 🖌️ Frontend


nickvergessen:
- Ref https://github.com/nextcloud/spreed/issues/14411

nickvergessen:
Mostly done for now? So let's close it

PR #14926, opened 2025-04-23
feat(RightSidebar): improve appearance, show profile details

Antreesy:
### ☑️ Resolves

* Ref #14413
* Requires https://github.com/nextcloud-libraries/nextcloud-vue/pull/6666
* Requires nextcloud/server#52366

No info | Minimum | Maximum | Actions
-- | -- | -- | --
image | image | image | image

Preview | Narrow screen 
-- | --
image | image
Full pic | Narrow screen 
image | image

### 🚧 Tasks

- [x] work on small screens (320*256)

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -153,23 +153,12 @@ export default {
 		currentConversation() {
 			return this.$store.getters.conversation(this.token)
 		},
-
-		/**
-		 * Computes whether the current conversation is one to one
-		 *
-		 * @return {boolean} The result
-		 */
-		isOneToOne() {
-			return this.currentConversation?.type === CONVERSATION.TYPE.ONE_TO_ONE
-				|| this.currentConversation?.type === CONVERSATION.TYPE.ONE_TO_ONE_FORMER
-		},
 	},
 
 	watch: {
 		token(newValue, oldValue) {
 			const shouldShowSidebar = BrowserStorage.getItem('sidebarOpen') !== 'false'
-			// Collapse the sidebar if it's a one to one conversation
-			if (this.isOneToOne || !shouldShowSidebar || this.isMobile) {
+			if (!shouldShowSidebar || this.isMobile) {
 				this.sidebarStore.hideSidebar({ cache: false })
 			} else if (shouldShowSidebar) {
 				this.sidebarStore.showSidebar({ cache: false })
--- a/src/components/RightSidebar/RightSidebar.vue
+++ b/src/components/RightSidebar/RightSidebar.vue
@@ -6,8 +6,8 @@
 <template>
 	<NcAppSidebar v-if="isSidebarAvailable"
 		:open="opened"
-		:name="sidebarTitle"
-		:title="sidebarTitle"
+		:name="conversation.displayName"
+		:title="conversation.displayName"
 		:active.sync="activeTab"
 		:class="'active-tab-' + activeTab"
 		:toggle-classes="{ 'chat-button-sidebar-toggle': isInCall }"
@@ -20,26 +20,12 @@
 			<IconMessageText :size="20" />
 			<span v-if="unreadMessagesCounter > 0" class="chat-button-unread-marker" />
 		</template>
-		<!-- search in messages button-->
-		<template v-if="!showSearchMessagesTab && getUserId" #secondary-actions>
-			<NcActionButton type="tertiary"
-				:title="t('spreed', 'Search messages')"
-				:aria-label="t('spreed', 'Search messages')"
-				@click="handleShowSearch(true)">
-				<template #icon>
-					<IconMagnify :size="20" />
-				</template>
-			</NcActionButton>
-		</template>
-		<template v-else-if="getUserId" #tertiary-actions>
-			<NcButton type="tertiary"
-				:title="t('spreed', 'Back')"
-				:aria-label="t('spreed', 'Back')"
-				@click="handleShowSearch(false)">
-				<template #icon>
-					<IconArrowLeft class="bidirectional-icon" :size="20" />
-				</template>
-			</NcButton>
+		<template #info>
+			<RightSidebarContent :is-user="!!getUserId"
+				:mode="CONTENT_MODES[contentModeIndex]"
+				:state="showSearchMessagesTab ? 'search' : 'default'"
+				@update:mode="handleUpdateMode"
+				@update:search="handleShowSearch" />
 		</template>
 		<template #description>
 			<InternalSignalingHint />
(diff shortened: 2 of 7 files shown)


img/icon-conversation-group-bright.svg
nickvergessen, line 1: We have to check if this breaks dashboard icons or mobile apps

DorraJaouad: Reminder: event icon is also coming 👀

src/components/RightSidebar/RightSidebarContent.vue
```diff
@@ -0,0 +1,378 @@
...
+	}
+}
+
+/**
+ * Concatenates profile strings
+ * @param firstSubstring first part of string
+ * @param secondSubstring second part of string
+ */
+function joinFields(firstSubstring?: string | null, secondSubstring?: string | null): string {
+	return [firstSubstring, secondSubstring].filter(Boolean).join(' · ')
```
DorraJaouad, line 145:
```suggestion
	return [firstSubstring, secondSubstring].join(' · ')
```

src/components/RightSidebar/RightSidebar.vue
```diff
@@ -198,7 +187,45 @@ export default {
 	},
 
 	setup() {
+		const sidebar = ref(null)
+		const contentModeIndex = ref(parseInt(BrowserStorage.getItem('sidebarContentMode') ?? 1, 10))
+
+		useEventListener(sidebar, 'wheel', handleWheelEvent, { capture: true })
```
DorraJaouad, line 193: Shared items are scrollable too, it will lose its scrolling when the profile info are not at the edge of directions

src/components/RightSidebar/RightSidebarContent.vue
DorraJaouad: Check their existence

DorraJaouad (commented):
Nice styling 🧑‍🎨

src/components/RightSidebar/RightSidebarContent.vue
Antreesy: I want either one of these in the output:
'a · b'
'a'
'b'
Without filtering nulls e.t.c it will be:
' · b'

src/components/RightSidebar/RightSidebar.vue
Antreesy: All tabs should be locked from scrolling, while avatar+profile are expanded

nickvergessen:
Last minor things:
- I would reset the scrolling when changing conversation
- I would start with all scrolled up (old behaviour) in group conversations?

src/components/RightSidebar/RightSidebarContent.vue
```diff
@@ -0,0 +1,413 @@
...
+	})
+
+	return fields
+})
+
+watch(token, async () => {
+	profileLoading.value = true
+	await groupwareStore.getUserProfileInformation(conversation.value)
+	profileLoading.value = false
+}, { immediate: true })
```
DorraJaouad, line 124: We can skip fetch when there is data already in the store, profile information are not often changing so it's safe to do it only once per reload.

src/components/RightSidebar/RightSidebar.vue
```diff
@@ -150,42 +137,44 @@ import BreakoutRoomsTab from './BreakoutRooms/BreakoutRoomsTab.vue'
...
 import SetGuestUsername from '../SetGuestUsername.vue'
 
 import { CONVERSATION, WEBINAR, PARTICIPANT } from '../../constants.ts'
+import BrowserStorage from '../../services/BrowserStorage.js'
 import { hasTalkFeature } from '../../services/CapabilitiesManager.ts'
 import { useSidebarStore } from '../../stores/sidebar.ts'
 
 const supportConversationCreationAll = hasTalkFeature('local', 'conversation-creation-all')
 
+const CONTENT_MODES = ['compact', 'preview', 'full']
```
DorraJaouad, line 154: Maybe we should enable it only for conversation with images and not with generated avatar name initials or emojis?

src/components/RightSidebar/RightSidebar.vue
```diff
@@ -198,7 +187,57 @@ export default {
 	},
 
 	setup() {
+		const sidebar = ref(null)
+		const contentModeIndex = ref(parseInt(BrowserStorage.getItem('sidebarContentMode') ?? 1, 10))
```
DorraJaouad, line 191: I don't think there is a reason to keep using previous state, it's just scrolling for user and he can end up in any state that he did think about it too much. I'd suggest if we always initiate it preview?

src/App.vue
```diff
@@ -153,23 +153,12 @@ export default {
...
-			return this.currentConversation?.type === CONVERSATION.TYPE.ONE_TO_ONE
-				|| this.currentConversation?.type === CONVERSATION.TYPE.ONE_TO_ONE_FORMER
-		},
 	},
 
 	watch: {
 		token(newValue, oldValue) {
 			const shouldShowSidebar = BrowserStorage.getItem('sidebarOpen') !== 'false'
-			// Collapse the sidebar if it's a one to one conversation
-			if (this.isOneToOne || !shouldShowSidebar || this.isMobile) {
```
DorraJaouad, line 172: Why are we not closing it again?

src/components/RightSidebar/RightSidebarContent.vue
Antreesy: Need to implement some sort of caching mechanism for that (and user absence, e.g.) Let's do it in follow-up

src/components/RightSidebar/RightSidebar.vue
Antreesy: yeah, can drop it. only need to check which conversation is open first

src/App.vue
Antreesy: to show a new feature to end-users

src/components/RightSidebar/RightSidebar.vue
nickvergessen: we sadly don't have the info on API side at the moment

src/components/RightSidebar/RightSidebar.vue
nickvergessen: As mentioned in the office, I would init with `preview` in 1-1 and with `compact` in group conversations

src/App.vue
DorraJaouad: but we don't have to do that, it's packing the opening view for 1-1 just to show a feature :')

nickvergessen: but we want to show the sidebar by default :) 
We do it for group conversations already because it adds useful information and it does so now in 1-1
If you don't want to learn/see stuff, you can close it and `shouldShowSidebar` will take care?

src/components/RightSidebar/RightSidebar.vue
Antreesy: Addressed

src/App.vue
Antreesy: yup, it just doesn't hide the sidebar anymore, if it's already open

DorraJaouad (approved):
Looks good 
Tested ✅

PR #14954, opened 2025-04-26
refactor: hand raised notification logic

aynugek:
The previous implementation used an ungodly if else statement.
This switches to a concise ternary operator statement


Antreesy (commented):
Thanks for the contribution! Would you be up for finding more places like this?

Current micro-optimisation in a single place introduces more issues for us as developers in terms of a difference with stable versions and backport conflicts, than we would benefit from it.

PR #15008, opened 2025-05-03
fix(useDocumentTitle): exclude system messages from triggering the asterik in tab name (document title)

### LINKED ISSUE #14998: Asterix marker * appears for system messages that don't trigger the unread counter which can be confusing
### How to use GitHub

---

## Steps to reproduce
1. Open your conversation list
2. Create an open conversation
3. Exit to empty main screen of Talk
4. Navigate to a different tab
5. As a second user join the open conversation

### Expected behaviour
User 1 has "nothing"

### Actual behaviour
User 1 has the * navigates to Talk tab and doesn't see any change, as it's only a system message and no unread marker is shown.

DorraJaouad:
…terik in tab name (document title)

### ☑️ Resolves

* Fix #14998

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


src/composables/useDocumentTitle.ts
```diff
@@ -99,10 +99,12 @@ export function useDocumentTitle() {
 			}
 
 			if ((lastMessage.actorId === actorId.value && lastMessage.actorType === actorType.value)
-				|| (isArchived && !hasUnreadMentions(conversation) && !hasCall(conversation))) {
+				|| (isArchived && !hasUnreadMentions(conversation) && !hasCall(conversation))
+				|| lastMessage.systemMessage) {
```
nickvergessen, line 103: This is incorrect as per chat last week. Just because the last message is a system message, doesn't mean all messages we loaded are.

Antreesy (approved):
Tested, looks good

I would still follow-up with unifying this and `App.vue#unreadCountsMap` logic, to iterate conversations once

PR #15029, opened 2025-05-07
feat: Talk Dashboard

### LINKED ISSUE #14408: 🍱 Talk Dashboard - 🖌️ Frontend
- Ref https://github.com/nextcloud/spreed/issues/14406

DorraJaouad:
- ### ☑️ Resolves

* Fix #14408

Initial views | During call
-- | --
image | image
image|

image

### 🚧 Tasks

- [ ]  Move to another component
- [ ] Calendar badge
-  [ ]  Call event room
-  [ ]  Hover effect on room
-  [ ]  Handle duplicates of events (when they are in multiple calendars)
-  [ ]  Add instant meeting button
-  [ ]  specify max width for the whole of it (like chat)
-  [ ]  upcoming reminders
-  [ ]  Device checker
-  [ ]  Unread mentions
-  [ ]  Periodic fetch  + add button
-  [ ]  Loading placeholder
-  [ ]  Fix calendar badge and In {room}

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/Dashboard/EventCard.vue
+++ b/src/components/Dashboard/EventCard.vue
@@ -85,6 +85,7 @@ const invitesLabel = computed(() => {
 
 const hasCall = computed(() => {
 	return props.eventRoom.roomActiveSince !== null
+		&& props.eventRoom.start * 1000 <= (Date.now() - 600_000) // 10 minutes buffer
 })
 
 /**
--- a/src/components/Dashboard/EventCard.vue
+++ b/src/components/Dashboard/EventCard.vue
@@ -1,6 +1,6 @@
 <!--
-  - SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
-  - SPDX-License-Identifier: AGPL-3.0-or-later
+	- SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+	- SPDX-License-Identifier: AGPL-3.0-or-later
 -->
 
 <script lang="ts" setup>
@@ -13,7 +13,7 @@ import IconCalendarBlank from 'vue-material-design-icons/CalendarBlank.vue'
 import IconTextBox from 'vue-material-design-icons/TextBox.vue'
 import IconVideo from 'vue-material-design-icons/Video.vue'
 
-import { t, n, getLocale } from '@nextcloud/l10n'
+import { t, n, getCanonicalLocale } from '@nextcloud/l10n'
 import moment from '@nextcloud/moment'
 
 import NcButton from '@nextcloud/vue/components/NcButton'
@@ -30,24 +30,21 @@ import { CONVERSATION } from '../../constants.ts'
 import type { DashboardEventRoom } from '../../types/index.ts'
 import { formattedTime } from '../../utils/formattedTime.ts'
 
-// props
 const props = defineProps<{
-	eventRoom: DashboardEventRoom
+	eventRoom: DashboardEventRoom,
 }>()
 const store = useStore()
 const router = useRouter()
 const isInCall = useIsInCall()
 const elapsedTime = computed(() => {
-	if (!hasCall.value) {
+	if (!hasCall.value || !(props.eventRoom.roomActiveSince ?? conversation.value.callStartTime)) {
 		return ''
 	}
-	return formattedTime(+useNow({ interval: 1_000 }).value - (props.eventRoom.roomActiveSince ?? 0) * 1000)
+	return formattedTime(+useNow({ interval: 1_000 }).value - (props.eventRoom.roomActiveSince ?? conversation.value.callStartTime) * 1000)
 })
 
 const isToday = computed(() => {
-	const startDate = new Date(props.eventRoom.start * 1000)
-	const today = new Date()
-	return startDate.toDateString() === today.toDateString()
+	return new Date(props.eventRoom.start * 1000).toDateString() === new Date().toDateString()
 })
 
 const conversation = computed(() => {
@@ -60,8 +57,8 @@ const eventDateLabel = computed(() => {
 	}
 	const startDate = new Date(props.eventRoom.start * 1000)
 	const endDate = new Date(props.eventRoom.end * 1000)
-	const startDateString = startDate.toLocaleString(getLocale(), { hour: '2-digit', minute: '2-digit' })
-	const endDateString = endDate.toLocaleString(getLocale(), { hour: '2-digit', minute: '2-digit' })
+	const startDateString = startDate.toLocaleString(getCanonicalLocale(), { hour: '2-digit', minute: '2-digit' })
+	const endDateString = endDate.toLocaleString(getCanonicalLocale(), { hour: '2-digit', minute: '2-digit' })
 	const dateString = isToday.value
 		? t('spreed', 'Today')
 		: moment(startDate).calendar(null, {
@@ -69,6 +66,7 @@ const eventDateLabel = computed(() => {
 			nextWeek: 'dddd',
 			sameElse: 'dddd'
 		})
+	// FIXME should be a translated string
 	return `${dateString} ${startDateString} - ${endDateString}`
 })
 
@@ -79,12 +77,12 @@ const hasAttachments = computed(() => {
 const invitesLabel = computed(() => {
 	const acceptedInvites = props.eventRoom.accepted ? n('spreed', '%n person accepted', '%n people accepted', props.eventRoom.accepted) : ''
 	const declinedInvites = props.eventRoom.declined ? n('spreed', '%n person declined', '%n people declined', props.eventRoom.declined) : ''
-	const separator = acceptedInvites && declinedInvites ? ', ' : ''
-	return `${acceptedInvites}${separator}${declinedInvites}`
+	// FIXME should be a translated string ??
+	return [acceptedInvites, declinedInvites].filter(Boolean).join(', ')
 })
 
 const hasCall = computed(() => {
-	return props.eventRoom.roomActiveSince !== null
+	return (conversation.value.hasCall || props.eventRoom.roomActiveSince !== null)
 		&& props.eventRoom.start * 1000 <= (Date.now() - 600_000) // 10 minutes buffer
 })
 
@@ -184,13 +182,15 @@ function handleJoin({ call = false } = {}) {
 	height: 225px;
 	display: flex;
 	flex-direction: column;
-	flex: 0 0 300px;
+	flex: 0 0 100%;
+	max-width: 300px;
 	border: 3px solid var(--color-border);
 	padding: calc(var(--default-grid-baseline) * 2);
 	border-radius: var(--border-radius-large);
 
 	&--highlighted {
 		background-color: var(--color-primary-light);
+
 		&:not(.event-card--in-call) {
 			border-color: var(--color-primary-light) !important;
 		}
@@ -203,6 +203,7 @@ function handleJoin({ call = false } = {}) {
 	&:not(.event-card--in-call):hover > .event-card__invitation-info.initial {
 		display: none;
 	}
+
 	&:not(.event-card--in-call):hover > .event-card__invitation-info.hovered {
 		display: flex;
 	}
--- a/src/components/Dashboard/EventCard.vue
+++ b/src/components/Dashboard/EventCard.vue
@@ -180,11 +180,10 @@ function handleJoin({ call = false } = {}) {
 <style scoped lang="scss">
 .event-card {
 	position: relative;
-	min-height: 225px;
+	height: 225px;
 	display: flex;
 	flex-direction: column;
-	max-width: 300px;
-	min-width: 200px;
+	flex: 0 0 300px;
 	border: 3px solid var(--color-border);
 	padding: calc(var(--default-grid-baseline) * 2);
 	border-radius: var(--border-radius-large);
@@ -200,11 +199,11 @@ function handleJoin({ call = false } = {}) {
 		border-color: var(--color-primary) !important;
 	}
 
-	&:hover > .event-card__invitation-info.initial {
-		opacity: 0;
+	&:not(.event-card--in-call):hover > .event-card__invitation-info.initial {
+		display: none;
 	}
-	&:hover > .event-card__invitation-info.hovered {
-		opacity: 1;
+	&:not(.event-card--in-call):hover > .event-card__invitation-info.hovered {
+		display: flex;
 	}
 
 	&__date {
@@ -264,12 +263,10 @@ function handleJoin({ call = false } = {}) {
 		display: flex;
 		align-items: center;
 		justify-content: space-between;
-		padding: var(--default-grid-baseline);
-		opacity: 1;
-		transition: opacity 0.05s ease-in-out;
+		padding: calc(var(--default-grid-baseline) * 2);
 
 		&.hovered {
-			opacity: 0;
+			display: none;
 		}
 	}
 }
--- a/src/components/Dashboard/EventCard.vue
+++ b/src/components/Dashboard/EventCard.vue
@@ -5,27 +5,41 @@
 
 <script lang="ts" setup>
 
-import { computed } from 'vue'
+import { useNow } from '@vueuse/core'
+import { computed, onMounted, onUnmounted, ref } from 'vue'
+import { useRouter } from 'vue-router/composables'
 
 import IconTextBox from 'vue-material-design-icons/TextBox.vue'
+import IconVideo from 'vue-material-design-icons/Video.vue'
 
 import { t, n, getLocale } from '@nextcloud/l10n'
 import moment from '@nextcloud/moment'
 
+import NcButton from '@nextcloud/vue/components/NcButton'
 import NcChip from '@nextcloud/vue/components/NcChip'
 import usernameToColor from '@nextcloud/vue/functions/usernameToColor'
 
 import ConversationIcon from '../ConversationIcon.vue'
 
+import { useIsInCall } from '../../composables/useIsInCall.js'
 import { useStore } from '../../composables/useStore.js'
 import { CONVERSATION } from '../../constants.ts'
 import type { DashboardEventRoom } from '../../types/index.ts'
+import { formattedTime } from '../../utils/formattedTime.ts'
 
 // props
 const props = defineProps<{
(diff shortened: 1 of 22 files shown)


src/stores/talkdashboard.ts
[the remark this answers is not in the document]
Antreesy: Or with, but you can write it in Composition API:
- Docs ref - https://pinia.vuejs.org/core-concepts/#Setup-Stores
- mounted, unmounted, interval fetch logic can be moved here
- useRoute to check if we're on home route

and just `dashboard.ts` maybe?

src/views/WelcomeView.vue
```diff
@@ -65,11 +69,15 @@ watchEffect(async () => {
...
 
 <template>
-	<EmptyView :name="text.name"
-		:description="text.description">
-		<template #icon>
-			<NcLoadingIcon v-if="isCreatingConversationForCallUser" />
-			<NcIconSvgWrapper v-else :svg="IconTalk" />
-		</template>
-	</EmptyView>
+	<div>
```
Antreesy, line 72: you don't need extra div with v-if/v-else, it's one child
```suggestion
```

src/components/Dashboard/TalkDashboard.vue
```diff
@@ -0,0 +1,226 @@
...
+	} catch (error) {
+		console.error('Error creating conversation:', error)
+		showError(t('spreed', 'Error creating a meeting'))
+	}
+}
+</script>
+<template>
+	<div class="talk-dashboard-wrapper">
+		<div class="talk-dashboard__header">
+			{{ t('spreed', 'Talk home') }}
```
Antreesy, line 94: I don't feel the need in this header, it's just taking space 🤷🏽 

But I would incorporate a TopBar styled header with 'Start meeting now' button, so the place is always reserved (and non-scrollable like the rest dashboard content):
image

src/components/Dashboard/TalkDashboard.vue
```diff
@@ -0,0 +1,226 @@
...
+		<div v-if="!loading && eventRooms.length > 0" class="talk-dashboard__event-cards">
+			<EventCard v-for="eventRoom in eventRooms"
+				:key="eventRoom.eventLink"
+				:event-room="eventRoom"
+				class="talk-dashboard__event-card" />
+		</div>
+		<div v-else-if="loading"
+			class="event-cards__loading-placeholder">
+			<LoadingPlaceholder type="event-cards" />
+		</div>
```
Antreesy, line 115:
```suggestion
		<LoadingPlaceholder v-else-if="loading"
			class="event-cards__loading-placeholder"
			type="event-cards" />
```

Make sure placeholder is horizontally scrollable on small screen

src/components/Dashboard/TalkDashboard.vue
```diff
@@ -0,0 +1,226 @@
...
+		height: var(--default-clickable-area);
+		padding : var(--default-grid-baseline);
+	}
+
+	&.loading {
+		overflow: hidden;
+	}
+}
+
+.talk-dashboard__empty-content {
```
Antreesy, line 210: I don't like the applying of color here, it's standing out too much (in comparison to a single meeting):

0 | 1
-- | --
image | image

Also pls set padding-inline

src/components/Dashboard/TalkDashboard.vue
```diff
@@ -0,0 +1,226 @@
...
+.talk-dashboard__devices-button {
+	margin-block: calc(var(--default-grid-baseline) * 4);
+}
+
+.talk-dashboard__unread-mentions {
+	max-height: 300px;
+	width: 300px;
+	overflow-y: auto;
+
+	& > .title {
```
Antreesy, line 199: Align title with 'Check devices' button?

image

src/components/Dashboard/TalkDashboard.vue
```diff
@@ -0,0 +1,226 @@
...
+}
+
+/**
+ * Creates a new group conversation and navigates to the conversation page.
+ */
+async function startMeeting() {
+	try {
+		const conversation = await store.dispatch('createGroupConversation', {
+			roomName: t('spreed', 'Meeting'), // NOTE: it uses user language statically
+			roomType: CONVERSATION.TYPE.PUBLIC,
```
Antreesy, line 78:
```suggestion
			roomType: CONVERSATION.TYPE.PUBLIC,
			objectType: 'instant_meeting',
```

Align with https://github.com/nextcloud/spreed/pull/15051 and use `'conversation-creation-all'` capability (just in case)

src/components/Dashboard/TalkDashboard.vue
```diff
@@ -0,0 +1,226 @@
...
+			class="talk-dashboard__empty-content"
+			:name="t('spreed', 'No upcoming meetings')"
+			:description="t('spreed', 'You have no upcoming meetings scheduled in the next 7 days')">
+			<template #icon>
+				<IconCalendarBlank :size="40" />
+			</template>
+		</NcEmptyContent>
+		<NcButton class="talk-dashboard__devices-button"
+			type="tertiary"
+			@click="emit('talk:media-settings:show', '')">
```
Antreesy, line 126: Maybe pass a param here instead of having computed `isInTalkDashboard`?
```suggestion
			@click="emit('talk:media-settings:show', 'device-check')">
```

As a follow-up, I would replace 'Choose device' section in Talk Settings with that button

src/stores/talkdashboard.ts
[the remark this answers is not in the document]
DorraJaouad: > and just dashboard.ts maybe?

I don't want to confuse it with dashboard up, especially that we also have Talk there.

src/components/Dashboard/TalkDashboard.vue
DorraJaouad: cc @nimishavijay

src/components/Dashboard/TalkDashboard.vue
DorraJaouad: Without background, the text seems to be lost in the middle

src/components/Dashboard/TalkDashboard.vue
DorraJaouad: or maybe like this ?
image

src/components/Dashboard/TalkDashboard.vue
DorraJaouad: >  use 'conversation-creation-all' capability

it's implemented inside the store action

src/store/conversationsStore.js
```diff
@@ -115,6 +115,7 @@ function emitUserStatusUpdated(conversation) {
 const state = {
 	conversations: {
 	},
+	conversationsInitialised: false,
```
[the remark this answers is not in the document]
DorraJaouad: > Object.keys(state.conversations[token]).length !== 0

We can't use this as user might not have conversations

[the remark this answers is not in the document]
Antreesy: Right. Then let's proceed with current solution and keep a follow-up in mind

src/components/Dashboard/TalkDashboard.vue
Antreesy: E.g. Teams is using Eventcard-styled component as a placeholder:

image

Maybe we can make it a similar thumbnail with link to Calendar (or Schedule a meeting)? cc @nimishavijay

nimishavijay:
Looks great Dorra! 

One main point would be to have a carousel instead of a side-scroll with a scroll bar. I imagine it is very similar to the Collectives recent pages element, with an arrow on the right.

image

> I don't feel the need in this header, it's just taking space 🤷🏽
But I would incorporate a TopBar styled header with 'Start meeting now' button, so the place is always reserved (and non-scrollable like the rest dashboard content):

Well I'd say it's nice to have a header, similar to the Dashboard app, but we can style it differently: left align, and change wording to "Welcome to Nextcloud Talk" (or whatever branded name)
I'd say it's okay to scroll away as it's not extremely important info (unlike the name when you have a conversation open for example, which should stay)
image

> I don't like the applying of color here, it's standing out too much (in comparison to a single meeting)
...
Without background, the text seems to be lost in the middle

Agree with both. 
- [ ] We can remove the background
- [ ] Instead of a full empty content component, we can just use the text "You have no meetings ..." below "Upcoming meetings"
- [ ] And then below that show the start a meeting button and the actions in the 💬➕ menu: `Start a meeting` `Create a new conversation` `Call a phone number`, all secondary buttons with 44px height (not showing open conversations as they will be shown below anyway (?))
- [ ] Don't show unread mentions if empty, we can directly show the open conversations

src/components/Dashboard/TalkDashboard.vue
```diff
@@ -0,0 +1,435 @@
...
+const store = useStore()
+const router = useRouter()
+const talkDashboardStore = useTalkDashboardStore()
+const forwardScrollable = ref(false)
+const backwardScrollable = ref(false)
+const eventCardsWrapper = ref<HTMLInputElement | null>(null)
+const eventRooms = ref<DashboardEventRoom[]>(talkDashboardStore.getEventRooms)
+const upcomingReminders = ref<UpcomingReminder[]>(talkDashboardStore.getUpcomingReminders)
+const eventsInitialised = ref(eventRooms.value.length > 0)
+const remindersInitialised = ref(upcomingReminders.value.length > 0)
```
Antreesy, line 48: What's the point in the store then, if you request information on mount?
```suggestion
const eventRooms = computed(() => talkDashboardStore.eventRooms)
const upcomingReminders = computed(() => talkDashboardStore.upcomingReminders)
const eventsInitialised = computed(() => talkDashboardStore.eventRoomsInitialised)
const remindersInitialised = computed(() => talkDashboardStore.upcomingRemindersInitialised)
```

src/stores/talkdashboard.ts
Antreesy, line 1: please, just `dashboard.ts` 🙏🏽

src/stores/talkdashboard.ts
```diff
@@ -0,0 +1,56 @@
...
+
+import { hasTalkFeature } from '../services/CapabilitiesManager.ts'
+import { getDashboardEventRooms } from '../services/dashboardService.ts'
+import { getUpcomingReminders } from '../services/remindersService.js'
+import type { DashboardEventRoom, UpcomingReminder } from '../types/index.ts'
+
+const supportsUpcomingReminders = hasTalkFeature('local', 'upcoming-reminders')
+
+type State = {
+	eventrooms: DashboardEventRoom[],
```
Antreesy, line 16:
```suggestion
	eventRooms: DashboardEventRoom[],
```

src/stores/talkdashboard.ts
```diff
@@ -0,0 +1,56 @@
...
+export const useTalkDashboardStore = defineStore('talkdashboard', {
+	state: (): State => ({
+		eventrooms: [],
+		upcomingReminders: [],
+	}),
+
+	getters: {
+		getEventRooms: (state) => state.eventrooms,
+		getUpcomingReminders: (state) => state.upcomingReminders,
+	},
```
Antreesy, line 28: For pinia store there is no point to return from, you can simply use `dashboardStore.eventRooms`
```suggestion
```

See this and following comments for `upcomingReminders` too

src/stores/talkdashboard.ts
```diff
@@ -0,0 +1,56 @@
...
+
+	actions: {
+		async fetchDashboardEventRooms() {
+			try {
+				const response = await getDashboardEventRooms()
+				this.eventrooms = response.data.ocs.data
+				return this.eventrooms
```
Antreesy, line 35: We can return from the store action, but in that case I would rely on getter and do interval requests to update the state
```suggestion
				Vue.set(this, 'eventRooms', response.data.ocs.data)
```

src/stores/talkdashboard.ts
```diff
@@ -0,0 +1,56 @@
...
+			} catch (error) {
+				console.error('Error fetching dashboard event rooms:', error)
+				throw error
```
Antreesy, line 38: Maybe show a toast here and do not throw an error? then we don't need excessive logic in component

src/components/Dashboard/TalkDashboard.vue
```diff
@@ -0,0 +1,435 @@
...
+const actualiseDataInterval = ref<ReturnType<typeof setInterval> | null>(null)
```
Antreesy, line 49: No need for reactive here
```suggestion
let actualiseDataInterval = null
```

src/components/UIShared/LoadingPlaceholder.vue
```diff
@@ -156,6 +156,33 @@ export default {
 	}
 }
 
+// Event cards placeholder ruleset
+.placeholder-list--event-cards {
+	display: flex;
+	flex-wrap: nowrap;
```
Antreesy, line 162: This still creates horizontal scroll, if items are still loading
```suggestion
	flex-wrap: nowrap;
	overflow: hidden;
```

src/components/UIShared/LoadingPlaceholder.vue
```diff
@@ -156,6 +156,33 @@ export default {
...
+			width: 100%;
+			padding: var(--default-grid-baseline);
+			margin: 0;
+			gap: var(--default-grid-baseline);
+
+			&-line {
+				margin: 0;
+				width: 100%;
+				min-width: 200px;
+				max-width: 300px;
```
Antreesy, line 177: Since we have fancy buttons to scroll, maybe it's better to have a fixed width of 300px for cards? (and placeholders)?

Antreesy (commented):
First batch

Antreesy (commented):
batch two

src/components/Dashboard/EventCard.vue
```diff
@@ -0,0 +1,330 @@
...
+	const startDateString = startDate.toLocaleString(undefined, { hour: '2-digit', minute: '2-digit' })
+	const endDateString = endDate.toLocaleString(undefined, { hour: '2-digit', minute: '2-digit' })
+	const dateString = isToday.value
+		? t('spreed', 'Today')
+		: moment(startDate).calendar(null, {
+			nextDay: '[Tomorrow]',
+			nextWeek: 'dddd',
+			sameElse: 'dddd'
+		})
+	return `${dateString} ${startDateString} - ${endDateString}`
```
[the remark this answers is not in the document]
DorraJaouad: Will translation work like that?

nimishavijay:
Discussed during the team call:
- [ ] Add action buttons on top (wrap to the next row if width is not enough, 44px height and around 4px of extra padding on the right and left)
- [ ] "Hello, [display name]" on the left, center aligned with the navigation button
- [ ] Add a full height `main-background` bg to the right/left buttons in the upcoming events section
- [ ] Add more padding for the entire screen (around 12-16px)
- [ ] If there are no upcoming meetings, show a "sample card" with a link to the calendar 
- [ ] When you click on "Start meeting now" have a little popover where you can fill the name, prefilled with "Meeting on March 14th" and a copy link button (ref MS Teams screenshot)


image

image

image

src/components/Dashboard/TalkDashboard.vue
```diff
@@ -0,0 +1,435 @@
...
+			class="talk-dashboard__event-cards-wrapper"
+			:class="{'forward-scrollable': forwardScrollable, 'backward-scrollable': backwardScrollable}">
+			<div ref="eventCardsWrapper"
+				class="talk-dashboard__event-cards">
+				<EventCard v-for="eventRoom in eventRooms"
+					:key="eventRoom.eventLink"
+					:event-room="eventRoom"
+					class="talk-dashboard__event-card" />
+			</div>
+			<div class="talk-dashboard__event-cards__scroll-indicator">
```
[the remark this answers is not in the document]
DorraJaouad: following collectives style

src/components/NewConversationDialog/NewConversationDialog.vue
```diff
@@ -241,6 +242,14 @@ export default {
 		},
 	},
 
+	mounted() {
+		subscribe('new-conversation-dialog:show', this.showModal)
```
DorraJaouad, line 246: We need to add the new bus events into the list!

DorraJaouad: and collect follow-ups in an issue

src/components/Dashboard/EventCard.vue
```diff
@@ -0,0 +1,330 @@
...
+})
+
+const hasAttachments = computed(() => {
+	return Object.keys(props.eventRoom.eventAttachments).length > 0
+})
+
+const invitesLabel = computed(() => {
```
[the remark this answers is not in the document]
DorraJaouad: We don't have data for it

src/components/Dashboard/TalkDashboard.vue
```diff
@@ -0,0 +1,435 @@
...
+async function actualiseData() {
+	await Promise.all([
+		fetchEventRooms(),
+		fetchUpcomingReminders(),
+	])
+}
+
+onMounted(() => {
+	actualiseData()
+	actualiseDataInterval.value = setInterval(actualiseData, FIVE_MINUTES)
```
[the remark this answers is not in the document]
DorraJaouad: It doesn't seem a better solution than the current one, we are in no need for a counter in our case

nimishavijay (approved):
Addressed design comments in chat 
Looks great! 🚀

src/components/Dashboard/EventCard.vue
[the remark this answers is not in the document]
Antreesy: I meant, there's sometime 'person' or 'people' omitted, and second participant used in 1-1. But might be OK for now

src/components/Dashboard/EventCard.vue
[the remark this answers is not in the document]
Antreesy: it should, let this be a follow-up

Antreesy (approved):
Should be good to go 😅

CorentinS6:
Hi,
Will this nice feature backported to spreed 20.x (NC 30) ?
Regards

Antreesy:
> Will this nice feature backported to spreed 20.x (NC 30) ?

Hi, thanks for the interest! No, it will be a part of 21.1 and heavy relies on server core and calendar app for NC31

PR #15039, opened 2025-05-08
feat(RightSidebar): show mutualEvents in 1-1 conversations

Antreesy:
### ☑️ Resolves

* Follow-up to #15016
* On top of #14926

🏡 After

2025-05-08_16h53_19

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


miaulalala (approved):
nice, works

src/components/CalendarEventsDialog.vue
```diff
@@ -339,24 +340,14 @@ async function submitNewMeeting() {
 			<template #default>
 				<template v-if="!loading && upcomingEvents.length">
 					<ul class="calendar-events__list">
+						<CalendarEventSmall v-for="event in upcomingEvents"
+							:key="event.uri"
+							:name="event.summary"
+							:start="event.start"
+							:href="event.href"
+							:color="event.color"
+							:is-recurring="!!event.recurrenceId" />
```
DorraJaouad, line 349: Why not passing the whole object and then in prop definition you make a default with those attributes showing that they are required?

Antreesy: Something is telling me, the current way is more performant for rendering
Thoughts, @shgkme?

src/components/RightSidebar/RightSidebarContent.vue
```diff
@@ -184,6 +219,18 @@ function joinFields(firstSubstring?: string | null, secondSubstring?: string | n
 					</div>
 				</div>
 			</div>
+			<div v-if="mode !== 'compact' && mutualEventsInformation.length"
+				class="content__events">
+				<NcAppNavigationCaption :name="t('spreed', 'Upcoming meetings')" />
```
DorraJaouad, line 224: I wouldn't name it "upcoming meetings" as it will be confusing to infer with meetings list in Topbar? should it be "meetings in common" or sthg similar ? @nimishavijay  @nickvergessen 

Follow-up, we need to find a better wording or explanation of the roles of both events lists.

DorraJaouad (approved):
Good otherwise

src/components/RightSidebar/RightSidebarContent.vue
nickvergessen: Extracted to https://github.com/nextcloud/spreed/issues/15096 so we can test the functionality with RC2

PR #15052, opened 2025-05-09
fix(RightSidebar): better error handling for user profile

Antreesy:
* fix: init audio objects before setting sinkId (Not the RightSidebar, but already pushed)
  * before: join/leave sounds will be played on default system devices
* fix: update local time in RightSidebar with periodic interval
  * before: time stuck once it's computed
* fix: extract icon fallback class to utils
  * reuse function from ConversationIcon
* fix: better throttle wheel events from touchpad
  * before: this can skip preview modes, as touchpad firing a ton of events
* Fix: #15072
  * no avatar for deleted users, show conversation avatar instead

🏚️ Before | 🏡 After
-- | --
image | image
image | No error

### 🚧 Tasks

- [ ] wheel event from touchpad / magic mouse

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/RightSidebar/RightSidebar.vue
+++ b/src/components/RightSidebar/RightSidebar.vue
@@ -193,12 +193,13 @@ export default {
 		const throttleHandleWheelEvent = (event) => {
 			if (!throttleTimeout) {
 				handleWheelEvent(event)
-				throttleTimeout = setTimeout(() => {
-					// do something
-					clearTimeout(throttleTimeout)
-					throttleTimeout = null
-				}, 300 /* var(--animation-slow) */)
 			}
+			clearTimeout(throttleTimeout)
+			throttleTimeout = null
+			throttleTimeout = setTimeout(() => {
+				clearTimeout(throttleTimeout)
+				throttleTimeout = null
+			}, 300 /* var(--animation-slow) */)
 		}
 
 		useEventListener(sidebar, 'wheel', throttleHandleWheelEvent, { capture: true })
--- a/src/components/ConversationIcon.vue
+++ b/src/components/ConversationIcon.vue
@@ -68,6 +68,7 @@ import { useIsDarkTheme } from '@nextcloud/vue/composables/useIsDarkTheme'
 import { AVATAR, CONVERSATION } from '../constants.ts'
 import { getConversationAvatarOcsUrl } from '../services/avatarService.ts'
 import { hasTalkFeature } from '../services/CapabilitiesManager.ts'
+import { getFallbackIconClass } from '../utils/conversation.ts'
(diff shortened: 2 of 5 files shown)


nickvergessen (changes requested):
PR title is not mergable

ShGKme:
`useNow` works over `useIntervalFn`, which is a basic `setInterval`.

If we want to use it in more places, we can wrap it into `createSharedComposable`. Then we'll have it only once for every place, and all the places will be updated simultaneously.

I'd propose to create a new component to display local updatable time. Currently, this computed is recalculated every 10 seconds, which (especially in Vue 2) may trigger the entire sidebar re-rendering. If it is only used in a new component, then only a small component is rerendered.

And, if we only render minutes, we don't need to recalculate it every 10 seconds. We only need it in about 60 seconds. We can have a 60s interval, but start it with a delay, so the 60s interval is triggered every minute change. With a check that it is not too far from a minute due to a cumulative error. If this is not too much optimization.

Antreesy:
> I'd propose to create a new component to display local updatable time

Could be done. But atm it's currently isolated in RightSidebarContent, which is an image + several lines of text, so already not a lot to re-render.

If you think it's blocking for the rest to be merged, I can split it into separate PR

src/components/RightSidebar/RightSidebar.vue
```diff
@@ -193,12 +193,13 @@ export default {
...
-					throttleTimeout = null
-				}, 300 /* var(--animation-slow) */)
 			}
+			clearTimeout(throttleTimeout)
+			throttleTimeout = null
+			throttleTimeout = setTimeout(() => {
+				clearTimeout(throttleTimeout)
+				throttleTimeout = null
+			}, 300 /* var(--animation-slow) */)
 		}
```
DorraJaouad, line 203: Throttling is incorrect here? as you can keep firing events within 300ms and it won't be called until you stop for 300ms
```js
		const throttleHandleWheelEvent = (event) => {
			if (!throttleTimeout) {
				handleWheelEvent(event)
				throttleTimeout = setTimeout(() => {
					clearTimeout(throttleTimeout)
					throttleTimeout = null
				}, 300 /* var(--animation-slow) */)
			}
		}
```

Antreesy: > it won't be called until you stop for 300ms

This was a desired behaviour. Try to scroll with touchpad and see how long events will be fired.
Maybe 300ms is a bit much in this case

DorraJaouad: I thought you want to throttle on one action per 300ms not stop it excessively until user stops scrolling

DorraJaouad (approved):
Everything else is good

Issue #8820, opened 2023-02-22
Allow "Secret" chats


PasqualePerilli:
**Is your feature request related to a problem? Please describe.**
It would be nice to create "secret" chats with specific users so that once a message is received, the notification doesn't preview the actual sender and content of the message.
This is particularly useful when sharing the computer screen or on mobile too (to avoid people looking at the message preview with the phone still locked).
This increases privacy and makes it so that for specific chats there must be an intentional action to actually look at the message of the secret chat.

**Describe the solution you'd like**
The "secret" chat could have a lock icon next to the person name to clearly indicate the different notification settings.
When a message is received from a "secret" chat, the notification doesn't preview the message content (neither on mobile nor on web) nor the sender.
The notification would just say something to the like: "New message received" and the sender would be displayed as "Private Conversation".

**Describe alternatives you've considered**
None

**Additional context**
This is a feature that would increase the privacy, especially in situations where one is sharing the screen. 
On the mobile front, whether the screen is active or not, for privacy reasons it would be great to hide the preview of the message for the secret chat so that the content is not displayed unless there is an intentional action to do so.
Other messaging applications such as Telegram already have this.

nickvergessen:
Regarding screensharing, most people in my enviroment use the "DND" option of the user status (which stops push notifications and desktop popup notifications)

agibson-fl:
It would be nice to get the notification but have the content obscured.  Probably good to have an option somewhere to hide the content and/or hide the sender in the notification. Signal allows you to hide the content of a notification.

PR #15053, opened 2025-05-09
feat(conversations): Add "sensitive conversations" that hide the last message

SystemKeeper:
## 🛠️ API Checklist

* Fixes https://github.com/nextcloud/spreed/issues/8820 (?)

### 🏁 Checklist

### DIFF
--- a/lib/Dashboard/TalkWidget.php
+++ b/lib/Dashboard/TalkWidget.php
@@ -265,9 +265,12 @@ class TalkWidget implements IAPIWidget, IIconWidget, IButtonWidget, IOptionWidge
 
 	protected function prepareRoom(Room $room, string $userId): WidgetItem {
 		$participant = $this->participantService->getParticipant($room, $userId);
+		$attendee = $participant->getAttendee();
 		$subtitle = '';
 
-		if ($room->getLastMessageId() && $room->isFederatedConversation()) {
+		if ($attendee->isSensitive()) {
+			// Don't leak sensitive last messages on dashboard
+		} elseif ($room->getLastMessageId() && $room->isFederatedConversation()) {
 			try {
 				$cachedMessage = $this->pcmService->findByRemote(
 					$room->getRemoteServer(),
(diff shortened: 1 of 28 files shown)


lib/Notification/Notifier.php
```diff
@@ -625,6 +625,10 @@ protected function parseChatMessage(INotification $notification, Room $room, Par
 			}
 		}
 
+		if ($participant->getAttendee()->isSensitive()) {
+			$message->setMessage($l->t('New message'), []);
```
nickvergessen, line 629: Not sure about this, but maybe we just skip the message part?

nickvergessen (commented):
Otherwise very nice and fully covered 🙇

lib/Dashboard/TalkWidget.php
```diff
@@ -286,6 +286,11 @@ protected function prepareRoom(Room $room, string $userId): WidgetItem {
 		}
 
 		$attendee = $participant->getAttendee();
+
+		if ($attendee->isSensitive()) {
+			$subtitle = '';
+		}
```
nickvergessen, line 292: move before previous if so we save to load, generate and parse the message in first place?

lib/Migration/Version21002Date20250509153510.php
```diff
@@ -0,0 +1,41 @@
...
+
+namespace OCA\Talk\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version21002Date20250509153510 extends SimpleMigrationStep {
```
nickvergessen, line 18:
```suggestion
class Version21001Date20250509153510 extends SimpleMigrationStep {
```

gonna make it part of 21.1 the version is XYYY when Talk is X.Y.Z

SystemKeeper:
- [ ] Check if it’s better to have subject and message on push set

PR #15085, opened 2025-05-14
fix: throttle participants request instead of debouncing

Antreesy:
### ☑️ Resolves

* I was wrong in  #11064 =P
	* Original solution was introduced in 29936c357a3953e995dea091a384e55dccb5785c
	* Previous request still cancels the parallel one, but each debounce trigger pushed the request invocation.
	* E.g. if there is a signaling message every 2 seconds (with update interval of 3 seconds), we will never request the list
	* At the start of the large calls, that explains , if you're not in the call yet, why participants loading is taking long (but maybe that also spreads the server load from when everyone joins at the same time)
	* After joining (or without a call), interval is 3 seconds, so it's harder to notice.
* New approach:
  * Do not debounce requests by 3/15 seconds, but delay the first invocation by 3/15 seconds, and ignore consequent ones happening within that interval

🏚️ Before | 🏡 After
-- | --
2025-05-14_19h29_12 | 2025-05-14_19h26_13

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/composables/useGetParticipants.js
+++ b/src/composables/useGetParticipants.js
@@ -2,7 +2,6 @@
  * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
-import debounce from 'debounce'
 import { ref, nextTick, computed, watch, onBeforeUnmount, onMounted } from 'vue'
 
 import { subscribe, unsubscribe } from '@nextcloud/event-bus'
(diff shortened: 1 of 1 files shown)


src/composables/useGetParticipants.js
```diff
@@ -85,17 +86,31 @@ export function useGetParticipants(isActive = ref(true), isTopBar = true) {
...
-		cancelableGetParticipants, 3000)
-	const debounceSlowUpdateParticipants = debounce(
-		cancelableGetParticipants, 15000)
+	const debounceFastUpdateParticipants = () => {
+		if (debounceFastUpdateParticipantsTimeout) {
+			// Ignore consequent requests, but no longer than defined timeout
+			return
+		}
+		debounceFastUpdateParticipantsTimeout = setTimeout(cancelableGetParticipants, 3_000)
+	}
```
ShGKme, line 106: It is not a `debounce`, it is called `throttle` 👀

Antreesy: `throttle` has a leading function invoke :shrug:
On the other hand, `npm debounce` does not have maxWait time, and `vueuse useDebounceFn` does not have a cancel option

So how do I name it then?) maybe defer(Fast|Slow)Update?

ShGKme (approved):
Nice catch

DorraJaouad (approved):
Tested

PR #15086, opened 2025-05-14
chore(desktop): do not skip check for https

ShGKme:
### ☑️ Resolves

* Not needed after https://github.com/nextcloud/talk-desktop/pull/1261


Antreesy (approved):
I guess that Desktop client on http is going to fail way before it reaches that code =)

ShGKme:
Servers on http are not supported at all at the moment, blocked on the login.

We could add support on the desktop client, but many important features are not available in the web browser anyway.

PR #15107, opened 2025-05-16
fix(instant-meetings): Fix missing retention duration from instant meetings

nickvergessen:
## 🛠️ API Checklist
- [x] Make sure non-local configs are documented as such
- [x] Mark instant-meetings as local config
- [x] Make sure the frontend uses it

### 🏁 Checklist


Antreesy (approved):
Tested 🙇🏽 
image

Issue #15096, opened 2025-05-15
"upcoming meetings" duplicated in 2 parts of the UI


nickvergessen:
> I wouldn't name it "upcoming meetings" as it will be confusing to infer with meetings list in Topbar? should it be "meetings in common" or sthg similar ? @nimishavijay  @nickvergessen 
> 
> Follow-up, we need to find a better wording or explanation of the roles of both events lists.

_Originally posted by @DorraJaouad in https://github.com/nextcloud/spreed/pull/15039#discussion_r2091079352_

Antreesy:
Additional follow-ups:

- [x] reactive 'Local time" - #15109 
- [ ] show info in conversation info (instead of 'Basic Info')
- [ ] polish scroll behaviour
  - [x] p1. https://github.com/nextcloud/spreed/pull/15052#discussion_r2091129753
  - [x] p2. disable scroll up (on whole component) to expand, replace with click event
  - [ ] p3. https://w3c.github.io/uievents/#cancelability-of-wheel-events
- [ ] rework shared items
- [ ] Out of office message expanded always
  - [ ] Additional change: Show only the OOO message title above the message input field with a chevron-right icon
  - [ ] Clicking on it would open the sidebar instead of expand the message
- [ ] Nice-to-have: Upcoming shared meetings with datetime and relevant group chat (if any). Possibly a follow-up
- [ ] Shared media, files, etc (all 3 + see more)
- [ ] Recent group conversations (3 with most recent activity)
- [ ] Nice-to-have: If there are no shared items, show some action items like `Create new text file` `Upload file or image` `Create a new poll` etc (not in screenshot)

Posted in https://github.com/nextcloud/spreed/issues/14411#issuecomment-2690347927

nimishavijay:
Good point. We could either do "Upcoming private meetings" in the top bar, or "Meetings with you and [name]" in the sidebar.

Rn it seems a bit confusing that these 2 lists of meetings show different things. Tbh I'm not sure if we need the upcoming meetings on the top, in the future we can just combine the 2 and show it only in the sidebar and just go with "Upcoming meetings" there. (For group conversations also)

nimishavijay:
Works for me! 
Could we already also show the 1on1s in the sidebar? Since they are technically also "Meetings with you"?

Antreesy:
> Could we already also show the 1on1s in the sidebar

[...]

PR #15109, opened 2025-05-16
fix(RightSidebar): make local time reactive

Antreesy:
### ☑️ Resolves

* Ref #15096
  * make time in Right SIdebar update every minute (align with system clock)
  * introduce shared composable for Date() object
  * introduce shared component for showing the date

It works

### 🚧 Tasks

- [x] Cumulative error? check if has real effect (should not be noticable without seconds shown)
  - [x] Didn't notice on tab retrieved from background (after 5-7 minutes)

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [x] Talk Desktop

### DIFF
--- a/src/composables/useCurrentTime.ts
+++ b/src/composables/useCurrentTime.ts
@@ -1,0 +1,54 @@
+/*
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import { createSharedComposable } from '@vueuse/core'
+import { onUnmounted, readonly, ref } from 'vue'
+import type { DeepReadonly, Ref } from 'vue'
+
+/**
+ * Composable to get current time (as Date object)
+ * @param [precision] - precision in milliseconds (defaults to one minute)
+ * @return Date reactive object with current time
+ */
+export function useCurrentTimeComposable(precision: number = 60_000): DeepReadonly<Ref<Date>> {
+	let timeout: ReturnType<typeof setTimeout>
+
+	const date = ref(new Date())
+
+	requestUpdate()
+
+	/**
+	 * Called for shared composable when all subscribers are unmounted (onScopeDispose)
+	 */
+	onUnmounted(() => {
+		clearTimeout(timeout)
+	})
+
+	/**
+	 * Recursively request an update of the current time (with compensation of microtask queue inaccuracy)
+	 */
+	function requestUpdate() {
+		date.value = new Date()
+		timeout = setTimeout(() => {
+			requestUpdate()
+		}, precision - date.value.valueOf() % precision)
+	}
+
+	return readonly(date)
+}
+
+/**
+ * Composable to get current time (as Date object)
+ * @return Date reactive object with current time (with 60 seconds precision)
+ */
+function useCurrentTimeMinuteComposable(): DeepReadonly<Ref<Date>> {
+	return useCurrentTimeComposable(60_000)
+}
+
+/**
+ * Shared composable to get current time (as Date object)
+ * @return Date reactive object with current time (with 60 seconds precision)
+ */
+export const useCurrentTime = createSharedComposable(useCurrentTimeMinuteComposable)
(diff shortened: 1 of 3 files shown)


src/composables/useCurrentTime.ts
```diff
@@ -0,0 +1,47 @@
...
+import { onUnmounted, readonly, ref } from 'vue'
+import type { DeepReadonly, Ref } from 'vue'
+
+/**
+ * Composable to get current time (as Date object)
+ * @param [precision] - precision in milliseconds (defaults to one minute)
+ * @return Date reactive object with current time
+ */
+export function useCurrentTimeComposable(precision: number = 60_000): DeepReadonly<Ref<Date>> {
+	let timeout: NodeJS.Timeout = null
```
ShGKme, line 16: `null` is as `object`

For this variable we can just keep it undefined

```suggestion
	let timeout: NodeJS.Timeout
```

ShGKme: Also `NodeJS.Timeout` is not a correct type for a web-browser. But probably should be fixed on tsconfig level.

src/composables/useCurrentTime.ts
```diff
@@ -0,0 +1,47 @@
...
+	return readonly(date)
+}
+
+/**
+ * Composable to get current time (as Date object)
+ * @return Date reactive object with current time (with 60 seconds precision)
+ */
+function useCurrentTimeMinuteComposable(): DeepReadonly<Ref<Date>> {
+	return useCurrentTimeComposable(60_000)
+}
```
ShGKme, line 45: Why do we need this function?

src/composables/useCurrentTime.ts
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Composable to get current time (as Date object)
+ * @param [precision] - precision in milliseconds (defaults to one minute)
+ * @return Date reactive object with current time
+ */
+export function useCurrentTimeComposable(precision: number = 60_000): DeepReadonly<Ref<Date>> {
+	let timeout: NodeJS.Timeout = null
+	let interval: NodeJS.Timeout = null
+
+	const date = ref<Date>(new Date())
```
ShGKme, line 19: Type is inferred
```suggestion
	const date = ref(new Date())
```

src/composables/useCurrentTime.ts
```diff
@@ -0,0 +1,47 @@
...
+
+	timeout = setTimeout(() => {
+		date.value = new Date()
+		interval = setInterval(() => {
+			date.value = new Date()
+		}, precision)
+	}, precision - date.value % precision)
```
ShGKme, line 26: `precision - date.value % precision` works only for a short time after composable was used, but will break after some time due to cumulative error.

src/composables/useCurrentTime.ts
Antreesy: To not be able to call shared composable with arguments. That way we know all subscribers would be updated once a minute, and not depending on who called first and with what

src/composables/useCurrentTime.ts
ShGKme: Interval is started on `X hours : Y min : 00 sec`. And it calls the function every N seconds, where N > 60.

If it would be `N === 60`, then the minute would update every actual minute change.

But `N > 60`. So we have updates on:
- 12:01:00
- 12:02:01
- 12:03:05
- 12:04:15
- ...
- 15:20:45
- 15:21:50
- 15:23:10

src/composables/useCurrentTime.ts
Antreesy: Yes, it is possible in component to call:
`const currentTime = useCurrentTime(1000)`
and update it every second

src/composables/useCurrentTime.ts
ShGKme: Solution - only use `setTimeout`

Antreesy: tested with 5-seconds precision
```js
console.log('requestUpdate', date.value.valueOf(), precision - date.value.valueOf() % precision)
<timestamp_seconds_milliseconds> <till next update>
 1747402_705_740 4260
 1747402_710_742 4258
 1747402_715_731 4269
 -- here tab is suspended --
 1747402_734_730 270
 1747402_794_733 267
 1747402_854_727 273
 -- back to normal --
 1747402_900_888 4112
 1747402_905_009 4991
 1747402_910_011 4989
```

ShGKme (approved):

PR #15112, opened 2025-05-16
fix(TalkDashboard):  support mobile view and correct minor styling issues

DorraJaouad:

🏚️ Before | 🏡 After
-- | --
image | image
image | image
image | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/Dashboard/EventCard.vue
+++ b/src/components/Dashboard/EventCard.vue
@@ -66,8 +66,11 @@ const eventDateLabel = computed(() => {
 			nextWeek: 'dddd',
 			sameElse: 'dddd'
 		})
-	// FIXME should be a translated string
-	return `${dateString} ${startDateString} - ${endDateString}`
+	return t('spreed', '{dateString} {startDateString} - {endDateString}', {
+		dateString,
+		startDateString,
+		endDateString,
+	})
 })
 
 const hasAttachments = computed(() => {
@@ -107,7 +110,7 @@ function handleJoin({ call = false } = {}) {
 			'event-card--highlighted': isToday,
 			'event-card--in-call': hasCall,
 		}">
-		<span class="title">
+		<h4 class="title">
 			<span v-for="calendar in props.eventRoom.calendars"
 				:key="calendar.principalUri"
 				class="calendar-badge"
@@ -115,7 +118,7 @@ function handleJoin({ call = false } = {}) {
 			<span class="title_text">
 				{{ props.eventRoom.eventName }}
 			</span>
-		</span>
+		</h4>
 		<p class="event-card__date secondary_text">
 			{{ eventDateLabel }}
 			<template v-if="hasCall">
@@ -124,7 +127,7 @@ function handleJoin({ call = false } = {}) {
 			</template>
 		</p>
 		<span class="event-card__room secondary_text">
-			<span>
+			<span class="event-card__room-prefix">
 				{{ props.eventRoom.roomType === CONVERSATION.TYPE.ONE_TO_ONE ? t('spreed', 'With') : t('spreed', 'In') }}
 			</span>
 			<NcChip type="tertiary"
@@ -160,7 +163,7 @@ function handleJoin({ call = false } = {}) {
 			<NcButton type="tertiary"
 				@click="handleJoin">
 				<template #icon>
-					<NcIconSvgWrapper :svg="IconTalk" />
+					<NcIconSvgWrapper :svg="IconTalk" :size="20" />
 				</template>
 				{{ t('spreed', 'View conversation') }}
 			</NcButton>
@@ -234,11 +237,12 @@ function handleJoin({ call = false } = {}) {
 		align-items: center;
 		gap: var(--default-grid-baseline);
 
-		& > span {
+		&-prefix {
 			line-height: var(--chip-size);
 			white-space: nowrap;
 			overflow: hidden;
 			text-overflow: ellipsis;
+			flex-shrink: 0;
 		}
 	}
 
@@ -279,12 +283,16 @@ function handleJoin({ call = false } = {}) {
 	align-items: center;
 	padding-inline-start: 6px; // revert negative margin
 	gap: var(--default-grid-baseline);
+	font-size: inherit;
+	margin: 0;
 
 	&_text {
 		font-weight: bold;
 		white-space: nowrap;
 		overflow: hidden;
 		text-overflow: ellipsis;
+		font-size: inherit;
+		margin: 0;
 	}
 }
 
--- a/src/components/Dashboard/EventCard.vue
+++ b/src/components/Dashboard/EventCard.vue
@@ -28,7 +28,7 @@ import { useIsInCall } from '../../composables/useIsInCall.js'
 import { useStore } from '../../composables/useStore.js'
 import { CONVERSATION } from '../../constants.ts'
 import type { DashboardEventRoom } from '../../types/index.ts'
-import { formattedTime } from '../../utils/formattedTime.ts'
+import { formattedTime, ONE_DAY_IN_MS } from '../../utils/formattedTime.ts'
 
 const props = defineProps<{
 	eventRoom: DashboardEventRoom,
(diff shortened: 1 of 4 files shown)


src/components/Dashboard/TalkDashboard.vue
```diff
@@ -220,9 +223,9 @@ function scroll({ direction }: { direction: 'backward' | 'forward' }) {
 				{{ t('spreed', 'Check devices') }}
 			</NcButton>
 		</div>
-		<div class="title">
+		<header class="title">
```
Antreesy, line 226: Maybe h2-h5? Definitely not a header tag should be used, it's a semantic block, like main or footer

ShGKme: Yes. `<header>` is a kind of banner landmark for all the heading elements. It must contain a direct heading element inside.

In this case, it should be `<h3>`:
1. Talk
2. Dashboard
3. Upcoming meetings

src/components/Dashboard/EventCard.vue
```diff
@@ -66,8 +66,11 @@ const eventDateLabel = computed(() => {
 			nextWeek: 'dddd',
 			sameElse: 'dddd'
 		})
-	// FIXME should be a translated string
-	return `${dateString} ${startDateString} - ${endDateString}`
+	return t('spreed', '{dateString} {startDateString} - {endDateString}', {
+		dateString,
+		startDateString,
+		endDateString,
+	})
```
ShGKme, line 73: I feel like translators will have problems understanding this string...

src/components/Dashboard/TalkDashboard.vue
```diff
@@ -167,7 +169,8 @@ function scroll({ direction }: { direction: 'backward' | 'forward' }) {
 		<div class="talk-dashboard__header">
 			{{ t('spreed', 'Hello, {displayName}', { displayName: store.getters.getDisplayName() }, { escape: false }) }}
 		</div>
-		<div class="talk-dashboard__actions">
+		<div class="talk-dashboard__actions"
+			:class="{'talk-dashboard__actions--mobile': isMobile}">
```
ShGKme, line 173: Can we use CSS (`@media`) for styling it?

src/components/Dashboard/EventCard.vue
Antreesy: Maybe we can use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/formatRange

src/components/Dashboard/TalkDashboard.vue
```diff
@@ -406,11 +422,11 @@ function scroll({ direction }: { direction: 'backward' | 'forward' }) {
 		border-radius: var(--border-radius-large);
 
 		&.backward {
-			inset-inline-start: 0;
+			inset-inline-start: -1px;
```
ShGKme, line 425: Could you describe why we need this 1px shift?

src/components/Dashboard/TalkDashboard.vue
```diff
@@ -509,4 +529,17 @@ function scroll({ direction }: { direction: 'backward' | 'forward' }) {
 	gap: 4px;
 	align-items: center;
 }
+
+@container talk-dashboard (max-width: 500px) {
+	@media (max-width: 600px) {
```
ShGKme, line 534: Can we use our default breakpoint size for `@media` and something like "default breakpoint - navigation" for container?

If every adaptive place has its own breakpoint position, everything changers layout at a different moment, making more intermediate states.

Of, if there were specific reasons to use exactly 500/600, let's add a comment, so we know in which case we should change it.

src/components/Dashboard/TalkDashboard.vue
```diff
@@ -233,13 +241,14 @@ function scroll({ direction }: { direction: 'backward' | 'forward' }) {
 					:event-room="eventRoom"
 					class="talk-dashboard__event-card" />
 			</div>
-			<div class="talk-dashboard__event-cards__scroll-indicator">
+			<div id="scroll-arrows-for-event-cards-wrapper"
```
ShGKme, line 244: Is this ID used?

ShGKme:
Why is the scroll event handler added via JS manually and not via `v-on` directive?

src/components/Dashboard/TalkDashboard.vue
DorraJaouad: It is to relate to event cards

ShGKme: I cannot find it...

src/components/Dashboard/TalkDashboard.vue
DorraJaouad: I don't know how to intentionally reproduce it but there is 1 px offset of cards comparing to the arrow button 
image

src/components/Dashboard/TalkDashboard.vue
```diff
@@ -144,16 +144,24 @@ async function startMeeting() {
  * Scrolls the event cards wrapper in the specified direction.
  * @param {string} direction - The direction to scroll ('backward' or 'forward').
  */
-function scroll({ direction }: { direction: 'backward' | 'forward' }) {
+function scrollEventCards({ direction }: { direction: 'backward' | 'forward' }) {
```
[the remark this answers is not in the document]
DorraJaouad: It will :p

src/components/Dashboard/EventCard.vue
```diff
@@ -239,6 +242,7 @@ function handleJoin({ call = false } = {}) {
 			white-space: nowrap;
 			overflow: hidden;
 			text-overflow: ellipsis;
+			flex-shrink: 0;
```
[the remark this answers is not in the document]
DorraJaouad: no need? it's known to always come together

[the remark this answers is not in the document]
Antreesy: It's working while `& > span` has the only match. If by accident NcChip will be changed from div to span (which IMO it should be), your styles will leak there

src/components/Dashboard/TalkDashboard.vue
Antreesy: but there's a gradient behind?

What about using 'tertiary-no-background' and rely only on gradient? red for visbility

image

src/components/Dashboard/TalkDashboard.vue
```diff
@@ -144,16 +142,24 @@ async function startMeeting() {
...
 	if (eventCardsWrapper.value) {
+		const ITEM_WIDTH = 300 + 8 // 300px width + 8px gap
 		let scrollAmount = 0
-		const visibleItems = Math.floor(eventCardsWrapper.value.clientWidth / (300 + 4))
+		const visibleItems = Math.floor(eventCardsWrapper.value.clientWidth / ITEM_WIDTH)
 		if (visibleItems === 0) {
-			// FIXME: mobile view, scroll by 1 item
 			scrollAmount = eventCardsWrapper.value.clientWidth * scrollDirection
+		} else {
+			scrollAmount = visibleItems * ITEM_WIDTH * scrollDirection
```
Antreesy, line 154: Nitpick: it scrolls by 308+-34px, but should be scrolled to the next card border instead. Can't think of a quick solution, we can leave it for the follow-up

src/components/Dashboard/EventCard.vue
```diff
@@ -57,17 +57,42 @@ const eventDateLabel = computed(() => {
...
+	let time
+	if (startDate.toDateString() === endDate.toDateString()) {
+		if (isToday || isTomorrow) {
+			// show the time only
+			const timeRange = Intl.DateTimeFormat(getCanonicalLocale(), {
+				hour: 'numeric',
+				minute: 'numeric',
+			}).formatRange(startDate, endDate)
+
+			// TRANSLATIORS: The time is shown in the format "HH:MM - HH:MM" e.g. "Tomorrow 10:00 - 11:00"
```
Antreesy, line 72:
```suggestion
			// TRANSLATORS: The time is shown in the format "HH:MM - HH:MM" e.g. "Tomorrow 10:00 - 11:00"
```

src/components/Dashboard/EventCard.vue
```diff
@@ -57,17 +57,42 @@ const eventDateLabel = computed(() => {
...
+				: t('spreed', 'Tomorrow {timeRange}', { timeRange })
+		} else {
+			time = Intl.DateTimeFormat(getCanonicalLocale(), {
+				weekday: 'long',
+				hour: 'numeric',
+				minute: 'numeric',
+			}).formatRange(startDate, endDate)
+		}
+	} else {
+		// show the month and the year as well
```
Antreesy, line 84: We should omit year, if it is the same

src/components/Dashboard/EventCard.vue
```diff
@@ -57,17 +57,42 @@ const eventDateLabel = computed(() => {
...
+		if (isToday || isTomorrow) {
+			// show the time only
+			const timeRange = Intl.DateTimeFormat(getCanonicalLocale(), {
+				hour: 'numeric',
+				minute: 'numeric',
+			}).formatRange(startDate, endDate)
+
+			// TRANSLATIORS: The time is shown in the format "HH:MM - HH:MM" e.g. "Tomorrow 10:00 - 11:00"
+			time = isToday
+				? t('spreed', 'Today {timeRange}', { timeRange })
```
Antreesy, line 74: In general, I wouldn't introduce two new strings for this narrow usecase, but something more generic like:
`t('spreed', '{prefix} {dateTime}', { prefix, dateTime })`

See how it's done in d094cf5b21f47aa682d2f1372c8d0a0eaa4cf5b3, we can get 'today' and 'tomorrow' from Intl as well

Don't refactor everything, just make it less harder for translators and easier for us to later move logic to util

src/components/Dashboard/EventCard.vue
ShGKme: https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

DorraJaouad: Thank you

src/components/Dashboard/EventCard.vue
DorraJaouad: It seemed off to me, I usually see the date (DD MM) with the year

src/components/Dashboard/TalkDashboard.vue
```diff
@@ -509,4 +528,14 @@ function scroll({ direction }: { direction: 'backward' | 'forward' }) {
...
+
+// Special breaking point 500px
+@container talk-dashboard (max-width: 500px) {
+	.talk-dashboard__actions {
+		:deep(.button-vue),
+		:deep(.v-popper--theme-dropdown) {
+			width: 100%;
+		}
+	}
+}
```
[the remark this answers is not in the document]
DorraJaouad: Why? and also title should always be left aligned as the content is left aligned

src/components/Dashboard/EventCard.vue
```diff
@@ -69,10 +69,11 @@ const eventDateLabel = computed(() => {
 				minute: 'numeric',
 			}).formatRange(startDate, endDate)
 
-			// TRANSLATIORS: The time is shown in the format "HH:MM - HH:MM" e.g. "Tomorrow 10:00 - 11:00"
-			time = isToday
-				? t('spreed', 'Today {timeRange}', { timeRange })
-				: t('spreed', 'Tomorrow {timeRange}', { timeRange })
+			// TRANSLATORS: e.g. "Tomorrow 10:00 - 11:00"
+			time = t('spreed', '{dayPrefix} {dateTime}', {
+				dayPrefix: isToday ? 'Today' : 'Tomorrow',
```
Antreesy, line 74: `t('spreed', 'Today')`

or use this:
```js
const rtf1 = new Intl.RelativeTimeFormat("en", { numeric: "auto" });
console.log(rtf1.format(0, "day"));
console.log(rtf1.format(1, "day"));
```

```css
	&::first-letter {
		text-transform: capitalize;
	}
```

Antreesy (approved):

PR #15113, opened 2025-05-16
fix: make audio output feature work in calls for Safari

Antreesy:
### ☑️ Resolves

* ~~Safari does not support AudioContext#setSinkId (required for call audio mixer)~~
* ~~UI was considering only Audio#setSinkId~~
* AudioContext destination was actually given to `<audio>` element, which we can modify

🏚️ Before | 🏡 After
-- | --
Screenshot before | to be added

### 🏁 Checklist

  - [ ] Firefox
  - [x] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/utils/webrtc/CallParticipantsAudioPlayer.js
+++ b/src/utils/webrtc/CallParticipantsAudioPlayer.js
@@ -159,11 +159,15 @@ CallParticipantsAudioPlayer.prototype = {
 			return
 		}
 
-		const promises = []
-		for (const audioElement of this._audioElements.values()) {
-			promises.push(this._setAudioElementOutput(deviceId, audioElement))
+		if (this._mixAudio) {
+			await this._setAudioElementOutput(deviceId, this._audioElement)
+		} else {
+			const promises = []
+			for (const audioElement of this._audioElements.values()) {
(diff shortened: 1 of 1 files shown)


src/stores/sounds.js
```diff
@@ -25,7 +26,7 @@ const shouldPlaySounds = hasUserAccount
  * Preferred version is the .ogg, with .flac fallback if .ogg is not supported (Safari)
  */
 const fileExtension = new Audio().canPlayType('audio/ogg') ? '.ogg' : '.flac'
-const isAudioOutputSelectSupported = !!(new Audio().setSinkId)
+const isAudioOutputSelectSupported = !isSafari && !!(new Audio().setSinkId)
```
ShGKme, line 29: Shall we check for AudioContext.setSinkId instead?

Antreesy: That was my first thought
But we don't support `AudioContext.setSinkId` in the code, so atm it's better to go that way

ShGKme: I'm afraid that one day Safari will support it, so `isSafari` check won't work

Ivansss (approved):
Tested on Safari and participants audio plays at the selected output 👍

ShGKme (approved):

PR #15148, opened 2025-05-19
fix: do not leave room session from duplicated session tab

### LINKED ISSUE #15116: Closing tab on duplicated session messages ends call
### How to use GitHub

---

## Steps to reproduce
1. Be in a call
2. Open the same conversation in a second tab
3. See the "Duplicate session" dialog
4. Close the tab again / Press "Leave page"

### Expected behaviour
Call still works

### Actual behaviour
Call is ended in the first tab

## Talk app

**Talk app version:** 21.1. RC2

Antreesy:
### ☑️ Resolves

* Fix #15116
  * `spawnDialog` handles 'yes', 'no' and 'close dialog' cases correctly, but not the page unload
  * If I get it correctly, we are denied to join the room on a second tab and ask user for confirmation. That means:
    * First tab has an active Nextcloud session
    * First tab has connected signaling session
  * If we close tab, 'unload' event triggers signaling kill and ⚠️ leave conversation request, which aborts the call on the first tab

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


src/composables/useSessionIssueHandler.js
```diff
@@ -46,6 +56,31 @@ export function useSessionIssueHandler() {
 		}
 	}
 
+	const handleSessionConflict = (token) => {
+		spawnDialog(ConfirmDialog, {
```
ShGKme: Note: `spawnDialog` now supports Promise API. When a callback is not provided, it returns a promise resolved with the result

src/composables/useSessionIssueHandler.js
```diff
@@ -48,7 +48,7 @@ export function useSessionIssueHandler() {
 				// FIXME: can't use router push as it somehow doesn't clean up
 				// fully and leads the other instance where "Join here" was clicked
 				// to redirect to "not found"
-				window.location = generateUrl(url)
+				window.location.replace(generateUrl(url))
```
ShGKme: Could you describe this change?

src/composables/useSessionIssueHandler.ts
```diff
@@ -4,6 +4,7 @@
  */
 
 import { nextTick, onBeforeMount, onBeforeUnmount, readonly, ref } from 'vue'
+import type { Readonly, Ref } from 'vue'
```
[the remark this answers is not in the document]
ShGKme: `Readonly` is TS's utility type

ShGKme (approved):

src/composables/useSessionIssueHandler.js
Antreesy: 1. TS was complaining at the new value type
2. location.replace won't be saved in history, so user can't go back to the conflicting route
3. More clear from the method, that we want to trigger a page reload

src/composables/useSessionIssueHandler.js
Antreesy: Vue3 only.. I'll leave a FIXME comment

DorraJaouad (approved):
Tested

PR #15181, opened 2025-05-22
test: make Jest fail on console.warn / console.error

Antreesy:
### ☑️ Resolves

* Allows to capture issues at JS code before it got merged
  * Components should be mounted in any test
  * can be avoided by `console.#method = jest.fn()` in place where it's expected

🏚️ Before | 🏡 After
-- | --
Green CI | Red CI

### 🏁 Checklist

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -13,6 +13,7 @@
 						:class="{'hidden-visually': isSearching}"
 						class="talk-home-button"
 						:title="t('spreed', 'Talk home')"
+						:aria-label="t('spreed', 'Talk home')"
 						@click="showTalkDashboard">
 						<template #icon>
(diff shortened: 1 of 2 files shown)


src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -13,6 +13,7 @@
 						:class="{'hidden-visually': isSearching}"
 						class="talk-home-button"
 						:title="t('spreed', 'Talk home')"
+						:aria-label="t('spreed', 'Talk home')"
```
ShGKme, line 16: It is a part of another PR already 👀

ShGKme (approved):

src/components/LeftSidebar/LeftSidebar.vue
Antreesy: It's not conflicting

PR #15206, opened 2025-05-23
fix: add attachment icon and link

DorraJaouad:
### ☑️ Resolves

* Fix:
*  Correct translation in event room location
* Add icon and show if there are other attachments
* Increase height by 25px so it can provide space for the new addition

🏚️ Before | 🏡 After
-- | --
image | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/Dashboard/EventCard.vue
+++ b/src/components/Dashboard/EventCard.vue
@@ -7,8 +7,8 @@
 
 import type { DashboardEventRoom } from '../../types/index.ts'
 
-import { getCanonicalLocale, n, t } from '@nextcloud/l10n'
-import moment from '@nextcloud/moment'
+import { getCanonicalLocale, getLanguage, n, t } from '@nextcloud/l10n'
+import { imagePath } from '@nextcloud/router'
 import usernameToColor from '@nextcloud/vue/functions/usernameToColor'
 import { useNow } from '@vueuse/core'
 import { computed } from 'vue'
@@ -17,7 +17,6 @@ import NcButton from '@nextcloud/vue/components/NcButton'
 import NcChip from '@nextcloud/vue/components/NcChip'
 import NcIconSvgWrapper from '@nextcloud/vue/components/NcIconSvgWrapper'
 import IconCalendarBlank from 'vue-material-design-icons/CalendarBlank.vue'
-import IconTextBox from 'vue-material-design-icons/TextBox.vue'
 import IconVideo from 'vue-material-design-icons/Video.vue'
 import ConversationIcon from '../ConversationIcon.vue'
 import IconTalk from '../../../img/app-dark.svg?raw'
@@ -53,23 +52,22 @@ const eventDateLabel = computed(() => {
 	}
 	const startDate = new Date(props.eventRoom.start * 1000)
 	const endDate = new Date(props.eventRoom.end * 1000)
-	const isToday = startDate.toDateString() === new Date().toDateString()
 	const isTomorrow = startDate.toDateString() === new Date(Date.now() + ONE_DAY_IN_MS).toDateString()
 
 	let time
 	if (startDate.toDateString() === endDate.toDateString()) {
-		if (isToday || isTomorrow) {
+		if (isToday.value || isTomorrow) {
 			// show the time only
 			const timeRange = Intl.DateTimeFormat(getCanonicalLocale(), {
 				hour: 'numeric',
 				minute: 'numeric',
 			}).formatRange(startDate, endDate)
 
-			const relativeFormatter = new Intl.RelativeTimeFormat('en', { numeric: 'auto' })
+			const relativeFormatter = new Intl.RelativeTimeFormat(getLanguage(), { numeric: 'auto' })
 
 			// TRANSLATORS: e.g. "Tomorrow 10:00 - 11:00"
 			time = t('spreed', '{dayPrefix} {dateTime}', {
-				dayPrefix: isToday ? relativeFormatter.format(0, 'day') : relativeFormatter.format(1, 'day'),
+				dayPrefix: isToday.value ? relativeFormatter.format(0, 'day') : relativeFormatter.format(1, 'day'),
 				dateTime: timeRange,
 			})
 		} else {
@@ -93,9 +91,7 @@ const eventDateLabel = computed(() => {
 	return time
 })
 
-const hasAttachments = computed(() => {
-	return Object.keys(props.eventRoom.eventAttachments).length > 0
-})
+const totalAttachments = computed(() => Object.values(props.eventRoom.eventAttachments))
 
 const invitesLabel = computed(() => {
 	const acceptedInvites = props.eventRoom.accepted ? n('spreed', '%n person accepted', '%n people accepted', props.eventRoom.accepted) : ''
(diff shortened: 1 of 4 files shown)


src/components/Dashboard/EventCard.vue
```diff
@@ -114,6 +113,24 @@ const hasCall = computed(() => {
...
 })
 
+const attachmentInfo = computed(() => {
+	if (hasAttachments.value) {
+		const file = Object.entries(props.eventRoom.eventAttachments)[0]?.[1]
+		const totalAttachments = Object.keys(props.eventRoom.eventAttachments).length
+
+		return {
+			icon: OC.MimeType.getIconUrl(file.fmttype) || imagePath('core', 'filetypes/file'),
+			label: file.filename,
```
nickvergessen, line 123: Does something like left-trim exist in JS?
```suggestion
			label: file.filename.ltrim('/'),
```

src/env.d.ts
```diff
@@ -21,6 +21,12 @@ declare global {
 		},
 	}
 
+	const OC : {
+		MimeType: {
+			getIconUrl: (mimetype: string) => string,
```
Antreesy, line 26: According to server code:
```suggestion
			getIconUrl: (mimetype?: string) => string | undefined,
```
Should be fine, since you're providing a file as a fallback ayway

src/components/Dashboard/EventCard.vue
```diff
@@ -276,18 +305,23 @@ function handleJoin({ call = false } = {}) {
...
 
-		& > * {
-			display: inline-block;
-			vertical-align: middle;
+		& > span {
+			white-space: nowrap;
+			overflow: hidden;
+			text-overflow: ellipsis;
+			font-weight: 500;
+			font-size: 0.9em;
```
Antreesy, line 319: `--font-size-small`?

src/components/Dashboard/EventCard.vue
```diff
@@ -114,6 +113,24 @@ const hasCall = computed(() => {
 		&& props.eventRoom.start * 1000 <= (Date.now() - 600_000) // 10 minutes buffer
 })
 
+const attachmentInfo = computed(() => {
+	if (hasAttachments.value) {
+		const file = Object.entries(props.eventRoom.eventAttachments)[0]?.[1]
+		const totalAttachments = Object.keys(props.eventRoom.eventAttachments).length
```
Antreesy, line 119: I think it makes now more sense to replace `hasAttachments` computed with
```js
const totalAttachments = computed(() => Object.values(props.eventRoom.eventAttachments))
```
Should be enough for anything (length, first file, extra label)

src/components/Dashboard/EventCard.vue
```diff
@@ -10,11 +10,10 @@ import { computed } from 'vue'
 import { useRouter } from 'vue-router/composables'
 
 import IconCalendarBlank from 'vue-material-design-icons/CalendarBlank.vue'
-import IconTextBox from 'vue-material-design-icons/TextBox.vue'
 import IconVideo from 'vue-material-design-icons/Video.vue'
 
 import { t, n, getCanonicalLocale } from '@nextcloud/l10n'
-import moment from '@nextcloud/moment'
+import { imagePath } from '@nextcloud/router'
```
Antreesy, line 16: Don't forget to rebase on latest main 👀

src/components/Dashboard/EventCard.vue
```diff
@@ -152,12 +169,11 @@ function handleJoin({ call = false } = {}) {
 			</template>
 		</p>
 		<span class="event-card__room secondary_text">
-			<span class="event-card__room-prefix">
-				{{ props.eventRoom.roomType === CONVERSATION.TYPE.ONE_TO_ONE ? t('spreed', 'With') : t('spreed', 'In') }}
-			</span>
 			<NcChip type="tertiary"
-				no-close
-				:text="props.eventRoom.roomDisplayName">
+				:text="props.eventRoom.roomType === CONVERSATION.TYPE.ONE_TO_ONE
```
Antreesy, line 173: It's  a bit too much for the template. Let's move to computed and add TRANSLATORS comments for this

src/components/Dashboard/EventCard.vue
```diff
@@ -166,12 +182,23 @@ function handleJoin({ call = false } = {}) {
...
-			<!--FIXME-->
-			{{ Object.entries(props.eventRoom.eventAttachments)[0]?.[1]?.filename }}
-		</span>
-		<span class="event-card__invitation-info initial">
+		<template v-if="hasAttachments && attachmentInfo">
+			<a class="event-card__attachment"
+				role="button"
+				:href="attachmentInfo.url"
+				:aria-label="t('spreed', 'View attachment')"
+				:title="t('spreed', 'View attachment')"
```
Antreesy, line 190: I'm not sure aria and title are needed:
- if it's a link, it doesn't need a title, just visual confirmation whether it's clickable or not
- role `button`? it opens a new tab, so `link` is more fitting. but I doubt it's needed for a tag
- extra string for translators no one sees

src/components/Dashboard/EventCard.vue
DorraJaouad: I was assuming it will handled by API? if not, yeah, we can do it here

src/components/Dashboard/EventCard.vue
ShGKme: > * role button? it opens a new tab, so link is more fitting. but I doubt it's needed for a tag

👍

ShGKme: > * extra string for translators no one sees

No one sees but screen readers users hear.

However, this element has a label already — its text content `attachmentInfo.label`.
It is not valid to have a real label and `aria-label` at the same time, especially different.

There should be no `aria-label`

src/components/Dashboard/EventCard.vue
```diff
@@ -109,6 +105,29 @@ const hasCall = computed(() => {
...
+	if (totalAttachments.value.length > 0) {
+		const file = totalAttachments.value[0]
+
+		return {
+			icon: OC.MimeType.getIconUrl(file.fmttype) || imagePath('core', 'filetypes/file'),
+			label: file.filename.replace(/^\//, ''),
+			extraLabel: totalAttachments.value.length > 1
+				? n('spreed', 'and %n other attachment', 'and %n other attachments', totalAttachments.value.length - 1)
+				: '',
+			url: file.previewLink,
```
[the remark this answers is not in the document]
Antreesy: href can't be `null`

src/components/Dashboard/EventCard.vue
```diff
@@ -109,6 +105,28 @@ const hasCall = computed(() => {
...
+			: '',
+		url: file.previewLink ?? undefined,
+	}
+})
+
+const roomLabel = computed(() => {
+	return props.eventRoom.roomType === CONVERSATION.TYPE.ONE_TO_ONE
+		? t('spreed', 'With {displayName}', { displayName: props.eventRoom.roomDisplayName }, { escape: false })
+		: t('spreed', 'In {conversation}', { conversation: props.eventRoom.roomDisplayName }, { escape: false })
+})
```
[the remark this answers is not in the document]
Antreesy: And now it's broken =P
image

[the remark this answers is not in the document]
DorraJaouad: Back, and I also disabled sanitization

src/components/Dashboard/EventCard.vue
```diff
@@ -204,7 +229,7 @@ function handleJoin({ call = false } = {}) {
 <style scoped lang="scss">
 .event-card {
 	position: relative;
-	height: 225px;
+	height: 250px;
```
Antreesy, line 232:
```suggestion
	height: 250px;
	background-color: var(--color-main-background);
```
Or it inherits transparent color =P
image

src/components/Dashboard/EventCard.vue
```diff
@@ -296,6 +315,7 @@ function handleJoin({ call = false } = {}) {
 		display: flex;
 		align-items: center;
 		justify-content: space-between;
+		border-radius: var(--border-radius-large);
```
Antreesy, line 318: Should consider outer border radius of event_card
But I'm not sure it's even needed
```suggestion
		border-radius: calc(var(--border-radius-large) - 3px);
```

Antreesy (approved):

PR #15221, opened 2025-05-26
fix: prevent navigation to inaccessible conversations

Antreesy:
### ☑️ Resolves

* Fix broken interface after attempt to navigate to a conversation, which can not be accessed by user
	- if it is not yet in the list - should be tried to fetch
	- in case of 403/404 - navigate to not-found page

🏚️ Before | 🏡 After
-- | --
image | image
App interface is broken | Correct redirect and further work of the app

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -376,7 +376,7 @@ export default {
 			}
 		})
 
-		const beforeRouteChangeListener = (to, from, next) => {
+		const beforeRouteChangeListener = async (to, from, next) => {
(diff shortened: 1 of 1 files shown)


src/App.vue
DorraJaouad, line 1: It looks there are so many checkpoints of `!this.$store.getters.conversation(token)` in different times **after** route change, so we need to scan which ones are still reachable if we are adding the exit shortcut

PR #15252, opened 2025-05-29
fix(groupware): omit duplicate upcoming events

### LINKED ISSUE #14496: Merge "next meeting" from different calendars
I have access to multiple calendars. Each of them has the same weekly call in their calendar.

If I open the "next meeting" drop-down I see the same meeting multiple time:

Image

Would be great if they get merged in one list. Otherwise it can be quite confusing and you risk to miss another meeting between the duplicates.

Antreesy:
### ☑️ Resolves

* Fix #14496 

🏚️ Before | 🏡 After
-- | --
2025-05-29_15h24_20 | 2025-05-29_15h19_40

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/stores/groupware.ts
+++ b/src/stores/groupware.ts
@@ -100,7 +100,12 @@ export const useGroupwareStore = defineStore('groupware', {
 			const location = generateUrl('call/{token}', { token }, { baseURL: getBaseUrl() })
 			try {
 				const response = await getUpcomingEvents(location)
-				Vue.set(this.upcomingEvents, token, response.data.ocs.data.events)
+				const uniqueEvents = response.data.ocs.data.events.filter((event, index, array) => {
+					// Keep only first meeting with the same location and start time
+					return index === array.findIndex((item) => item.start === event.start)
+				})
+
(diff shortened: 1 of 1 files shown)


src/stores/groupware.ts
```diff
@@ -100,7 +100,12 @@ export const useGroupwareStore = defineStore('groupware', {
 			const location = generateUrl('call/{token}', { token }, { baseURL: getBaseUrl() })
 			try {
 				const response = await getUpcomingEvents(location)
-				Vue.set(this.upcomingEvents, token, response.data.ocs.data.events)
+				const uniqueEvents = response.data.ocs.data.events.filter((event, index, array) => {
+					// Keep only first meeting with the same location and start time
+					return index === array.findIndex((item) => item.start === event.start)
+				})
```
ShGKme, line 106: Do we always have only a small number of events here?

Antreesy: Yes, it's limited to 10 by dav endpoint

PR #15265, opened 2025-05-30
fix(capabilities): conditionally show links to calendar

### LINKED ISSUE #15249: Don't link to calendar, if calendar is not enabled/available
### How to use GitHub

---

Even if calendar is not available, we still link to calendar on the dashboard:
Image

but then you end up on a "not found" page:

Image

Antreesy:
### ☑️ Resolves

* Fix #15249

🏚️ Before | 🏡 After
-- | --
2025-05-30_14h42_44 | 2025-05-30_14h42_02

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/lib/TInitialState.php
+++ b/lib/TInitialState.php
@@ -94,6 +94,11 @@ trait TInitialState {
 		);
 
 		$this->initialState->provideInitialState(
+			'calendar_enabled',
+			$appManager->isEnabledForUser('calendar', $user)
+		);
+
+		$this->initialState->provideInitialState(
 			'circles_enabled',
 			$appManager->isEnabledForUser('circles', $user)
 		);
--- a/src/components/ConversationSettings/BasicInfo.vue
+++ b/src/components/ConversationSettings/BasicInfo.vue
@@ -6,7 +6,7 @@
 <template>
 	<Fragment>
 		<!-- eslint-disable-next-line vue/no-v-html -->
-		<p v-if="canFullModerate && isEventConversation" class="app-settings-section__hint" v-html="calendarHint" />
+		<p v-if="isCalendarEnabled && canFullModerate && isEventConversation" class="app-settings-section__hint" v-html="calendarHint" />
 		<h4 class="app-settings-section__subtitle">
 			{{ t('spreed', 'Name') }}
 		</h4>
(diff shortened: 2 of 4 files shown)


DorraJaouad (approved):

src/components/UIShared/CalendarEventSmall.vue
```diff
@@ -21,7 +21,7 @@ const props = defineProps<{
 		<a class="calendar-event__item"
 			:class="{ 'calendar-event__item--thumb': !href }"
 			:href="href"
-			:title="t('spreed', 'Open Calendar')"
+			:title="href ? t('spreed', 'Open Calendar') : name ?? undefined"
 			:tabindex="0"
 			target="_blank">
```
ShGKme: It should not be a focusable interactive element when it actually doesn't work

lib/TInitialState.php
```diff
@@ -93,6 +93,11 @@ protected function publishInitialStateForUser(IUser $user, IRootFolder $rootFold
 			!$this->talkConfig->isNotAllowedToCreateConversations($user)
 		);
 
+		$this->initialState->provideInitialState(
+			'calendar_enabled',
+			$appManager->isEnabledForUser('calendar', $user)
+		);
```
ShGKme: I'm still not a fun of adding a new incorrect usage of the initial state...
Fine with it, if mobile clients don't need it.

Antreesy:
Since API is added now, we can consider it? https://github.com/nextcloud/spreed/pull/15396

nickvergessen:
That is an additional query per user opening the dashboard, sounds pretty bad to me

lib/TInitialState.php
nickvergessen: Capability added in calendar https://github.com/nextcloud/calendar/pull/8009

Antreesy:
FYI: If Calendar is disabled, there are no capabilities served, and fallback to `true` makes zero sense.
So if we backport it, and people won't update Talk together with Calendar, they'll lost these actions. Sounds fine?

nickvergessen:
> So if we backport it, and people won't update Talk together with Calendar, they'll lost these actions. Sounds fine?

Calendar is already released, so fine from my POV. If you update one app but not an other, you need to know what you are doing and can not complain that the interaction between both does not work anymore.

PR #15278, opened 2025-06-02
fix(RightSidebarContent): provide a conversation fallback for guests

Antreesy:
### ☑️ Resolves

* Fix rendering of RightSidebarContent for guests
  * they don't have conversation in the store yet, so need a fallback before it's fetched

🏚️ Before | 🏡 After
-- | --
image | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


ShGKme (commented):
Do we even need to show any content before we got it?

DorraJaouad:
I agree to show a loading placeholder either sentence or an element.

Antreesy:
Agree, but I'll check if it's worth keeping fallbacks =)

For RightSidebar - that would block header from rendering -> a11y issue
For TalkDashboard - you can be removed from conversation -> error in 5 minutes

nickvergessen (approved):

PR #15300, opened 2025-06-04
fix(useActiveSession): block UI from calling if session expired

Antreesy:
### ☑️ Resolves

* Possibly fix some cases, when users experience troubles joining the call

🏚️ Before | 🏡 After
-- | --

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


src/composables/useActiveSession.js
```diff
@@ -3,15 +3,21 @@
...
+import { getTalkConfig, hasTalkFeature } from '../services/CapabilitiesManager.ts'
 import { setSessionState } from '../services/participantsService.js'
+import { messagePleaseTryToReload } from '../utils/talkDesktopUtils.ts'
 import { useDocumentVisibility } from './useDocumentVisibility.ts'
 import { useIsInCall } from './useIsInCall.js'
 import { useStore } from './useStore.js'
 
 const INACTIVE_TIME_MS = 3 * 60 * 1000
+
+const experimentalRecoverSession = (getTalkConfig('local', 'experiments', 'enabled') ?? 0) & CONFIG.EXPERIMENTAL.RECOVER_SESSION
```
DorraJaouad, line 16: 1 & 2 = 0

Antreesy: yes, 1 is for sessionStore list update
we need to set 2 or 3

src/composables/useActiveSession.js
```diff
@@ -74,6 +80,16 @@ export function useActiveSession() {
...
 		} catch (error) {
 			console.error(error)
+			if (error?.response?.status === 404) {
+				// In case of 404 - participant did not have a session, block UI to join call
+				store.dispatch('updateLastJoinedConversationToken', '')
+				if (experimentalRecoverSession) {
+					// Automatically try to join the conversation again
+					store.dispatch('joinConversation', { token: token.value })
+				} else {
+					showError(t('spreed', 'Failed to join the conversation.') + '\n' + messagePleaseTryToReload)
```
nickvergessen: this is not joining, it's making as active

src/composables/useActiveSession.js
```diff
@@ -94,6 +110,16 @@ export function useActiveSession() {
...
```
nickvergessen: this is not joining, it's making as inactive

src/composables/useActiveSession.js
Antreesy: dropping user toast for now

PR #15304, opened 2025-06-06
fix(tokenStore): ease store before migration

Antreesy:
### ☑️ Resolves

* Drop `conversationSettingsToken` logic from the store as redundant
* Handle token variables locally instead, where required

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


src/components/PollViewer/PollManager.vue
```diff
@@ -22,11 +22,11 @@ const showPollEditor = ref(false)
 const showPollDraftHandler = ref(false)
 const container = ref<string | undefined>(undefined)
 
-const token = computed(() => store.getters.getConversationSettingsToken() || store.getters.getToken())
+const pollToken = ref('')
```
ShGKme:
```suggestion
const token = ref('')
```

src/components/ConversationSettings/ConversationSettingsDialog.vue
```diff
@@ -162,21 +164,19 @@ export default {
 
 	setup() {
 		const settingsStore = useSettingsStore()
+		const settingsToken = ref('')
```
ShGKme: The new name adds a feeling that it is something new to me, like we have a conversation token, but also some other kind of tokens.
```suggestion
		const token = ref('')
```

It also allows using shortcuts in Vue 3.

ShGKme: What do you think?

ShGKme (approved):

src/components/ConversationSettings/ConversationSettingsDialog.vue
Antreesy: see PollManager: `pollToken.value = token`.
it wouldn't work with `token.value = token`. So either current approach, or I can rename it in payload

ShGKme: 👀
```js
token.value = payload.token
token.value = event.token
```

ShGKme (approved):

PR #15306, opened 2025-06-06
feat(pinia): initialize pinia globally

Antreesy:
### ☑️ Resolves

* Inspired by https://pinia.vuejs.org/core-concepts/outside-component-usage.html#SSR-Apps
* that way it can be shared and used outside of components
  * sounds utils
  * webrtc
  * e.t.c.

### 🚧 Tasks

- [ ] Verify the concept
- [ ] test

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/init.js
+++ b/src/init.js
@@ -13,12 +13,15 @@ import BrowserStorage from './services/BrowserStorage.js'
 import { EventBus } from './services/EventBus.ts'
 import store from './store/index.js'
 import { useIntegrationsStore } from './stores/integrations.js'
+import pinia from './stores/pinia.ts'
 
 import '@nextcloud/dialogs/style.css'
 
 if (!window.OCA.Talk) {
 	window.OCA.Talk = {}
 }
+
+const integrationsStore = useIntegrationsStore(pinia)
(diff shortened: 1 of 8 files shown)


ShGKme:
> * Verify the concept

Changes in this PR specifically - must have. 

To prevent possible issues from wrong PInia's internal active pinia outside the application context.

In general, it is not good if low-level modules and utilities depend on Pinia stores as high-level Vue UI reactive state, and likely a sign of architectural issues.

For example:
- Not vue-related logic depends on 2 vue libraries (Pinia + Vue), additional places to break during migration. E.g. we should not change signaling code because we migrate from vuex to pinia, or have a risk of breaking it due to reactivity change in Vue 3
- Complexity of reusing such modules
- Circular dependencies

Alternatives:
1. Make sure a specific store actually must be a vue store.
   For example, if it is only used as a global variable with data and functions — it should be a general module
2. Instead of modifying the store from low-level modules and utilities - invert it and handle it on the store side.

PR #15328, opened 2025-06-11
feat: migrate token store to pinia 🍍

Antreesy:
### ☑️ Resolves

* Migrates token store to 🍍 Pinia
* Idea behind splitting `tokenStore` and `useGetToken`:
  * token getter is mostly used in components, and nothing else
  * if we would change/remove tokenStore in favor of simple module, that would not affect changes in all files
  * some tokenStore actions might be later migrated to composable listener (router.beforeEach or so)

### 🚧 Tasks

- [x] Manually test:
  - [x] main
  - [x] sidebars
  - [x] extensions (deck)
  - [x] iinternal signaling
  - [x] external signaling
  - [x] calls

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/RoomSelector.spec.js
+++ b/src/components/RoomSelector.spec.js
@@ -11,6 +11,7 @@ import NcDialog from '@nextcloud/vue/components/NcDialog'
 import ConversationSearchResult from './LeftSidebar/ConversationsList/ConversationSearchResult.vue'
 import RoomSelector from './RoomSelector.vue'
 import { CONVERSATION } from '../constants.ts'
+import { useTokenStore } from '../stores/token.ts'
 import { generateOCSResponse } from '../test-helpers.js'
 
 jest.mock('@nextcloud/axios', () => ({
@@ -35,8 +36,12 @@ const ConversationsSearchListVirtualStub = {
 
 describe('RoomSelector', () => {
 	let conversations
+	let tokenStore
 
 	beforeEach(() => {
+		tokenStore = useTokenStore()
+		tokenStore.token = 'current-token'
+
 		conversations = [{
 			token: 'token-3',
 			displayName: 'zzz',
@@ -80,16 +85,6 @@ describe('RoomSelector', () => {
 			type: CONVERSATION.TYPE.GROUP,
 			objectType: 'file',
 		}]
-
-		global.OCA.Talk = {
-			instance: {
-				$store: {
-					getters: {
-						getToken: jest.fn().mockReturnValue('current-token'),
-					},
-				},
-			},
-		}
 	})
 
 	afterEach(() => {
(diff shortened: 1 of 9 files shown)


src/components/RoomSelector.spec.js
```diff
@@ -82,13 +87,7 @@ describe('RoomSelector', () => {
...
 		global.OCA.Talk = {
-			instance: {
-				$store: {
-					getters: {
-						getToken: jest.fn().mockReturnValue('current-token'),
-					},
-				},
-			},
+			instance: {},
 		}
```
DorraJaouad: unused

src/stores/token.ts
```diff
@@ -0,0 +1,55 @@
...
+	function updateToken(newToken: string) {
+		token.value = newToken
+	}
+
+	/**
+	 * @param newFileId file ID of active conversation
+	 */
+	function updateFileIdForToken(newFileId: string | null) {
+		fileIdForToken.value = newFileId
+	}
```
DorraJaouad: The liaison that forces token passing whenever fileid is updated is now removed, then it's now possible that you can change fileid with no new token which was not possible

src/composables/useGetToken.ts
DorraJaouad, line 1: So, the purpose of reducing the use of stores is not in regards of components but in regards of external elements outside vue app like signaling* modules we have. It's clear that this composable has helped to eliminate the use of store in some components but I don't think it was an issue in the first place. 

Store is still present where it shouldn't be (externally) and cannot be substituted because of sidebar Talk integration outside main app (we can't use router there).

src/composables/useActiveSession.js
```diff
@@ -82,7 +82,7 @@ export function useActiveSession() {
 			console.error(error)
 			if (experimentalRecoverSession && error?.response?.status === 404) {
 				// In case of 404 - participant did not have a session, block UI to join call
-				tokenStore.updateLastJoinedConversationToken('')
+				tokenStore.updateLastJoinedConversationToken(null)
```
DorraJaouad: Why did you change to null?

src/components/RoomSelector.spec.js
Antreesy: test relies on it? If we're inside the Talk app, instance is present, then we can look up current token. Otherwise it might be a deck integration, and we don't have instance there

src/composables/useActiveSession.js
Antreesy: token can be either string or `''` (on dashboard e.g.). If we pass here `''`, computed might return true, which isn't correct, as we're not joined anywhere

src/composables/useGetToken.ts
Antreesy: Agree. 
We should get rid of token store one day (follow-up). Just making steps conveniently for us, with that composable that will greatly reduce amount of files touched

src/stores/token.ts
Antreesy: Possible now, but we're not doing it anyway 🤷

DorraJaouad: But future contribution might not have this liaison in mind

src/composables/useActiveSession.js
DorraJaouad: If computed returns true on '' then computed should be corrected instead of passed value. We are storing here joined token not opened talk

PR #15333, opened 2025-06-13
fix(AdminSettings): show errors in UI when testing websocket connection

Antreesy:
### ☑️ Resolves

* Provides more explicit information for instance administrators about issues while testing Websocket connection

>[!IMPORTANT]
> 🔬 To test:
> - CSP violation:
>   - remove HPB
>   - reload page (CSP rules updated)
>   - add HPB
>   - wait for websocket connection
> - Hello error (easy way for invalid_backend):
>   - open HPB config
>   - disable or comment `allowall`
>   - for [backend-1] set incorrect url

🏚️ Before 
CSP doesn't throw error, we're in infinite loading UI
image
Error from hello message doesn't throw error, we're in OK state
image

🏡 After
CSP is captured, if browser supports it (🦊 in Firefox disabled by default)
image
Hello response contains code and message (see in [signaling repo](https://github.com/strukturag/nextcloud-spreed-signaling/blob/6b04363faf8629cee3f42bebcfc1fa3355aaf8a2/docs/standalone-signaling-api-v1.md#error-codes))
image

### 🏁 Checklist

  - [x] Firefox
  - [ ] Safari

### DIFF
--- a/src/utils/SignalingStandaloneTest.js
+++ b/src/utils/SignalingStandaloneTest.js
@@ -4,6 +4,7 @@
  */
 
 import { generateOcsUrl } from '@nextcloud/router'
+import { isFirefox } from '../utils/browserCheck.ts'
 
 /**
  * This is a simplified version of Signaling prototype (see signaling.js)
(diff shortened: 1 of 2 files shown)


SystemKeeper:
Works nicely in chrome!

CSP does not work in Safari, although it should be supported from reading the docs...
Console logs

image

I am wondering, automatic reloading on CSP error is probably a bad idea, right?

Antreesy:
> automatic reloading on CSP error is probably a bad idea, right?

Unless we want to got into infinite loop of reloading 😜

We can show it as warning instead of error though?

image

SystemKeeper:
> Unless we want to got into infinite loop of reloading 😜

I feared that would be the answer 😅

Antreesy:
> CSP does not work in Safari, although it should be supported

Yup, could not make it with Safari, even with 18.4. Then it'll only be Chrome and Firefox (optional)

nickvergessen (approved):

Antreesy:
Lowered to warningMessage

src/utils/SignalingStandaloneTest.js
```diff
@@ -41,7 +41,30 @@ class StandaloneTest {
...
+					}
+				},
+				{
+					types: ['csp-violation'],
+					buffered: false,
+				},
+			)
+			CSPObserver.observe()
+		} else {
+			// Should be explicitly enabled in Firefox
```
DorraJaouad: Maybe more helpful if you add in the console warn?

PR #15334, opened 2025-06-13
fix: migrate file integration services to TS

Antreesy:
### ☑️ Resolves

* Migrate 3 services to TS
* Unify axios usage
* Fix typo in setCurrentUser

## 🖌️ UI Checklist

No visual changes

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


src/services/filesIntegrationServices.ts
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Gets the public share conversation token for a given share token.
+ *
+ * @param shareToken the token of the share
+ * @param options options
+ */
+const getPublicShareConversationData = async function(shareToken: string, options?: object): getRoomDataByShareTokenResponse {
+	return axios.get(generateOcsUrl('apps/spreed/api/v1/publicshare/{shareToken}', { shareToken }), options)
+}
```
DorraJaouad: Are there actual possible options for this request?

Antreesy: Maybe not for this one) But Axios options are always possible to pass, and we always need them for CancelableRequest, so better to be consistent with every request

src/services/filesIntegrationServices.ts
```diff
@@ -0,0 +1,49 @@
...
+import axios from '@nextcloud/axios'
+import { generateOcsUrl } from '@nextcloud/router'
+
+/**
+ * Gets the conversation token for a given file id
+ *
+ * @param fileId The file id to get the conversation for
+ * @param options options
+ */
+const getFileConversation = async function(fileId: number, options?: object): getRoomDataByFileIdResponse {
```
ShGKme: `object` is not a correct type for an object with any properties. It is a type of an `Object` instance.

For example, you can only use `Object` prototype properties here like `options.hasOwnProperty()` or `options.toString()` but nothing else like `options.abortSignal`.

For an object with any properties, you can use `Record<string, any>`.

But from type `object` and documentation `options options`, it is hard to understand what "options options" means and what actually can be passed here...

Antreesy:
You asked for it =P

src/services/messagesService.ts
```diff
@@ -33,7 +34,7 @@ import { CHAT } from '../constants.ts'
 type ReceiveMessagesPayload = Partial<receiveMessagesParams> & { token: string }
 type GetMessageContextPayload = getMessageContextParams & { token: string, messageId: number }
 type PostNewMessagePayload = Omit<postNewMessageParams, 'replyTo'> & { token: string, parent: ChatMessage }
-type PostNewMessageOptions = Pick<postNewMessageParams, 'silent'> & object
+type PostNewMessageOptions = Pick<postNewMessageParams, 'silent'> & AxiosRequestConfig
```
ShGKme: This looks invalid. Request body data and axios config are never mixed into a single object.

src/services/messagesService.ts
```diff
@@ -154,10 +155,10 @@ const clearConversationHistory = async function(token: string, options?: object)
  * @param param0 The message object that is destructured
  * @param param0.token The conversation token
  * @param param0.id The id of the message to be deleted
- * @param options request options
+ * @param [options] Axios request options
  */
-const deleteMessage = async function({ token, id }: DeleteMessagePayload, options?: object): deleteMessageResponse {
-	return axios.delete(generateOcsUrl('apps/spreed/api/v1/chat/{token}/{id}', { token, id }, options), options)
```
ShGKme, line 160: 🙈

PR #15335, opened 2025-06-13
fix(useGetParticipants): migrate to TS

Antreesy:
### ☑️ Resolves

* Drop some redundant logic (TopBar, isActive tab as prop)
* Move related logic to composable
* Possibly fix extra requests to get participants (?)
* Migrate to TS

No visual changes

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


src/composables/useGetParticipants.ts
```diff
@@ -0,0 +1,264 @@
...
+import { getTalkConfig } from '../services/CapabilitiesManager.ts'
+import { EventBus } from '../services/EventBus.ts'
+import { useActorStore } from '../stores/actor.ts'
+import { useSessionStore } from '../stores/session.ts'
+import { useDocumentVisibility } from './useDocumentVisibility.ts'
+import { useGetToken } from './useGetToken.ts'
+import { useIsInCall } from './useIsInCall.js'
+import { useStore } from './useStore.js'
+
+type UsersUpdatedPayload = (InternalSignalingSession | StandaloneSignalingJoinSession | StandaloneSignalingUpdateSession)[]
```
DorraJaouad: Can we move this to index as we are also using it in session store?

Antreesy:
cc @nickvergessen for fe15d4a1eef77f2f1d7b525634939588d0f51737:

Part of the issue is that we do not show participants videos for some reasons, if users haven't fetched the list yet.

Before: Users opened the page, they're not in the call, starts 15s throttle, join a call, starts another 3s throttle - list will be requested earlier;
Current: Users opened the page, they're not in the call, starts 15s throttle, join a call - still on 15s throttle
With this PR + experimental flag: Users opened the page and fetch the list straight away, following requests will be throttled by 60s

DorraJaouad:
> Current: Users opened the page, they're not in the call, starts 15s throttle, join a call - still on 15s throttle

~~No, once we receive any signaling messages which we do as soon as we join the call, 15s throttle is cancelled and 3s sec is set so total waiting time is 3s + server delay. THIS is true if we receive signaling message instantly upon joining the call.~~
UPD: Well, we do long update not fast on receiving signaling so it's 60s 🙈 

There might be because of the issue with receiving the first participant list changed signaling before receiving join call signaling (it set inCall boolean!). It means throttle participant is called but because inCall is falsy it goes to slow throttle.

src/composables/useGetParticipants.ts
```diff
@@ -0,0 +1,264 @@
...
+			throttleUpdateParticipants()
+		}
+	})
+
+	watch(isModeratorOrUser, (newValue, oldValue) => {
+		if (newValue && !oldValue) {
+			// Fetch participants list if guest was promoted to moderators
+			nextTick(() => throttleUpdateParticipants())
+		}
+	})
```
DorraJaouad, line 231: next tick is not at time of participants tab rendering anymore, so  `isActive` is falsy and fetch is ignored.

src/composables/useGetParticipants.ts
```diff
@@ -0,0 +1,264 @@
...
+			throttleLongUpdate()
+		}
+	}
+
+	/**
+	 * Check if current user permission has been changed from signaling messages
+	 */
+	async function checkCurrentUserPermissions([users]: [StandaloneSignalingUpdateSession[]]) {
+		// TODO: move logic to sessionStore once experimental flag is dropped
+		const currentUser = users.find((user) => user.userId === actorStore.userId)
```
DorraJaouad: It will be faster to update guest permissions based on signaling messages rather than waiting for room fetch.
So instead, filter by `actorId`.

src/composables/useGetParticipants.ts
```diff
@@ -0,0 +1,264 @@
...
+
+	/**
+	 * Trigger participants list update upon joining
+	 */
+	async function onJoinedConversation() {
+		if (isOneToOneConversation.value) {
+			await cancelableGetParticipants()
+		} else if (experimentalUpdateParticipants) {
+			// Make a first request immediately to get the list
+			await cancelableGetParticipants()
```
DorraJaouad, line 151: Ok, sounds plausible for the overall joining experience (you see less loading in participants tab). However, we still have leaving call as a critical moment for live update as we are seeing participants tab AND signaling message only gives session id (not much information to use).

src/composables/useGetParticipants.ts
```diff
@@ -0,0 +1,264 @@
...
+		}
+	}
+
+	/**
+	 * Patch participants list from signaling messages (left)
+	 */
+	function handleUsersLeft([sessionIds]: [StandaloneSignalingLeaveSession[]]) {
+		sessionStore.updateSessionsLeft(token.value, sessionIds)
+		throttleLongUpdate()
+	}
```
DorraJaouad, line 117: This can be received for own leave or others
```suggestion
	async function handleUsersLeft([sessionIds]: [StandaloneSignalingLeaveSession[]]) {
		const isOwnSessionLeft = actorStore.sessionId && sessionIds.includes(actorStore.sessionId)
		sessionStore.updateSessionsLeft(token.value, sessionIds)
		if (isInCall.value && isOwnSessionLeft) {
			await cancelableGetParticipants()
		} else {
			throttleLongUpdate()
		}
	}
```

src/composables/useGetParticipants.ts
```diff
@@ -0,0 +1,264 @@
...
+
+	/**
+	 * Patch participants list from signaling messages (end call for everyone)
+	 */
+	function handleUsersDisconnected() {
+		sessionStore.updateParticipantsDisconnectedFromStandaloneSignaling(token.value)
+		throttleLongUpdate()
+	}
```
DorraJaouad, line 125:
```suggestion
	async function handleUsersDisconnected() {
		sessionStore.updateParticipantsDisconnectedFromStandaloneSignaling(token.value)
		if (isInCall.value) {
			await cancelableGetParticipants()
		} else {
			throttleLongUpdate()
		}
	}
```

src/composables/useGetParticipants.ts
```diff
@@ -0,0 +1,264 @@
...
+	const isInCall = useIsInCall()
+	const isDocumentVisible = useDocumentVisibility()
+
+	const isActive = computed(() => activeTab.value === 'participants')
+	const token = useGetToken()
+	const conversation = computed<Conversation | undefined>(() => store.getters.conversation(token.value))
+	const isInLobby = computed<boolean>(() => store.getters.isInLobby)
+	const isModeratorOrUser = computed<boolean>(() => store.getters.isModeratorOrUser)
+	const isOneToOneConversation = computed(() => conversation.value?.type === CONVERSATION.TYPE.ONE_TO_ONE
+		|| conversation.value?.type === CONVERSATION.TYPE.ONE_TO_ONE_FORMER)
```
DorraJaouad, line 52: Why do we need an instant fetch on a former one to one?

src/composables/useGetParticipants.ts
Antreesy: Users left are for closing session with current room, when you navigate away. 
Why do you need to fetch participants list, if you're in the call and you're leaving this conversation? 👀

src/composables/useGetParticipants.ts
Antreesy: Cause we considered it 1-1 in all other places

DorraJaouad: Participant doesn't exist anymore, there is no need to include here

Antreesy: I'd skip it for now, as it causes no real pressure on the instance and might be removed together with the flag

src/composables/useGetParticipants.ts
Antreesy: Good catch here

src/composables/useGetParticipants.ts
Antreesy: Is it happening for you? For me, guest moderator now gets this update from `checkCurrentUserPermissions`, so I would even remove this watcher

NVM, found the bug

Antreesy:
> Part of the issue is that we do not show participants videos for some reasons, if users haven't fetched the list yet.

The reason: 36391053c9b25c136f4229a9c264011dbb5eaa8b

PR #15360, opened 2025-06-16
fix(l10n): plurals grammar fix for Chat manager

roliverio:
### ☑️ Resolves

Fixes plurals grammar for ChatManager

@rakekniven 👀

### DIFF
--- a/tests/integration/features/chat-2/mentions.feature
+++ b/tests/integration/features/chat-2/mentions.feature
@@ -61,7 +61,7 @@ Feature: chat/mentions
       | roomName | room |
     Then user "participant1" gets the following candidate mentions in room "group room" for "" with 200
       | id           | label                    | source | mentionId    | details           |
-      | all          | room                     | calls  | all          | All 1 participant |
+      | all          | room                     | calls  | all          | 1 participant |
 
   Scenario: get mentions in a group room
     When user "participant1" creates room "group room" (v4)


rakekniven:
@roliverio You need to adapt the tests as well

https://github.com/nextcloud/spreed/blob/475cd465f7cfaecb502af24246d054b8df8102af/tests/php/Chat/ChatManagerTest.php#L738

nickvergessen:
- Hi @rakekniven this app is very well maintained, therefore it's really not needed to merge things without the teams approval or chance to comment on it.
I reverted the PR in #15361 
- Changing a source string in master/main only is not going to remove it from the translations, so we need to clarify what to do and if we would want to backport things, etc.
- The other thing is, while it is okay for English to skip the "all" when it's just 1, it is much better for languages like Slovenian that have plural forms for numbers ending with 1, 2, 3+4 and 5-0 or other non singular-plural languages to have a proper string on the "1" case as well, or am I misjudging this?

rakekniven:
Hey @nickvergessen ,

I don't quite understand the message, because I have already merged changes in the past after creating a review?
I'll keep my hands off it here then.

There is a discussion about the string on Transifex.

In German and English, the grammar of the singular makes no sense.
The plural form and the placeholder are still present.
Therefore, a translation into e.g. Slovenian should be possible without any problems.
Or am I making a mistake here?

Antreesy:
IMO, both singular-plural forms should be consistent with each other. In this case, `All` prefix is stressing on 'literally everyone'. If there is a confusion among translators, then it requires TRANSLATORS comment from our side, explaining it better.

You can always reach out to us at public channel or Transifex directly, so we can fix it ourselves; we're just concerned, that it's pushed, merged, and already sent to Transifex (and daily instances deployed from master branches) overnight without a chance for us to react.

P.S. Translation shouldn't be a problem, I also dropped this prefix for russian translation, as it brings no additional meaning for singular form.
But I am aware as a developer, that end user will likely never see this string (unless they want to mention everyone in Note to self). Other translators are not; and they should have an explanation too.

nickvergessen:
> I don't quite understand the message, because I have already merged changes in the past after creating a review?
> I'll keep my hands off it here then.

Reviewing is fine and appriciated, we'd just like to have a chance to comment before a merge.

> In German and English, the grammar of the singular makes no sense.
The plural form and the placeholder are still present.

That is right, but my problem is that translators now see:
grafik

How many translators will check the English plural again when switching through the list of available plurals in their language to see that "other" has something else then just `%n participants` and how should they decide if they need to add `All` or not.
We can also hardcode the case of 1 into a different non-plural string `1 participant` or skip the sub line when it's only 1 (the current) user.

Just as a record, we are talking about this here:
grafik

rakekniven:
> Reviewing is fine and appriciated, we'd just like to have a chance to comment before a merge.

Ok, I'll take that to heart

> How many translators will check the English plural again when switching through the list of available plurals in their language to see that "other" has something else then just %n participants

The UI of Transifex will not allow you to save a string if one plural form is missing. Just double checked that. 

The image you show will be fixed using the PR from olivero, or

Singular: 1 participant
Plural: All 3 participants

But fore sure we can remove the prefix "All".

nickvergessen:
We will remove the subline in the case of 1 and add a translator hint pointing that out.

PR #15363, opened 2025-06-17
feat(web): always render code blocks LTR

### LINKED ISSUE #15362: Code blocks should always be LTR regardless of the message text direction
## Feature request

**Which Nextcloud Version are you currently using:**
Nextcloud AIO (Hub 10, 31.0.5)

It has always been the case that a message written in RTL language (Arabic) would have the code block rendered in RTL as well. Along with Nextcloud 10 and complete RTL support, I would have assumed this case would have been considered and code blocks would have been rendered in LTR always.

Code blocks should always render in LTR.

N/A

**Additional context**
Example of how code blocks are rendered in RTL:
Image

mahmoudajawad:
### ☑️ Resolves

* Fix #15362

🏚️ Before | 🏡 After
-- | --
image | image

### 🚧 Tasks

- [ ] To test locally built

### 🏁 Checklist

  - [X] Firefox
  - [ ] Safari
  - [ ] Talk Desktop
  - [X] Not risky to browser differences / client
- [X] ⛑️ Tests are included or not possible

### DIFF
--- a/src/components/MessagesList/MessagesGroup/Message/MessagePart/MessageBody.vue
+++ b/src/components/MessagesList/MessagesGroup/Message/MessagePart/MessageBody.vue
@@ -613,6 +613,11 @@ export default {
 	direction: inherit;
 }
 
+// Always render code blocks LTR
+:deep(.rich-text--wrapper) pre {
+	direction: ltr;
+}
+
 // Hardcode to restrict size of message images for the chat
 :deep(.widget-default--image) {
 	max-width: 240px;


Antreesy (approved):
Tested, thanks a lot!

nickvergessen:
Looks like it should be upstreamed?
Can you take care @Antreesy and then revert here after the bump?

Antreesy:
> Looks like it should be upstreamed?

I wasn't sure. Library does not control direction at all, leaving it to the apps; would it be correct to force it for everyone? As it can be only RTL text in there, with no code, just for styling purposes

nickvergessen:
> Library does not control direction at all, leaving it to the apps; would it be correct to force it for everyone? 

Well currently everyone using the component with markdown would now need to add that line, which feels unnecessary. I would argue if someone wants to have it always in reading direction, they can overwrite it. But for announcements, collectives, … where it's supposed to be alwas code, it should always be LTR?

mahmoudajawad:
Thanks, @Antreesy!

Issue #15313, opened 2025-06-06
🧵 Threads overview


nickvergessen:
- [ ] Community issue https://github.com/nextcloud/spreed/issues/1469

---

- [x] API https://github.com/nextcloud/spreed/issues/9679
- [x] Frontend https://github.com/nextcloud/spreed/issues/9680
- [x] iOS https://github.com/nextcloud/talk-ios/issues/1266
- [x] Android https://github.com/nextcloud/talk-android/issues/3074

nickvergessen:
## V1 draft during the team week

- 2 Modes due to different usecases and expectations (user level)
  - Regular/Current chat (would be used by e.g. by engineers and engineering TLs, that need to monitor all the messages in their conversations, public, helpdesk, …)
    - Each message that has a parent or child has an option to jump to a "Thread view" showing only those messages
    - Thread view should also be accessible without replies to allow subscribing in case someone replies
    - Inside the thread view chat API will automatically reply to the thread root if no other message was quoted
  - Thread view (would be used e.g. by managers, "visitors from other teams" as they care about their threads/questions, and less about other topics being discussed)
    - Hide all messages that have a parent / threadId
    - Show subline with numbers of replies (bonus points for unread dot, more points for unread   counter)
    - Show a single line "system message" when you were replied/mentioned in a thread for easier "becoming aware" and have an "entry lever" after navigating to a conversation with a "blue dot" / unread mention
- Chat endpoints gain an additional threadId parameter
- Posting stays untouched, threadId will be generated from the replyTo's threadId
- Notifications: Option to "subscribe to a thread" (new DB table)
- Search: option to filter by thread id
- Frontend: Split message store into "chat view store" + "message store" to mimik current behaviour and add a "thread view store" for the threaded version
- Frontend ~7 weeks, iOS ~7 weeks, Android more, API less

## V2

After discussing some "details" we realized that this is not going to scale so we have to reduce the functionality to something that "most" other messengers also do:

- Threads have to be created explicitly 🔃 changed
  - Same as slack and others now, could even split permissions into "Reply only" vs. "Reply and Create message/thread" to allow "Channels" like Teams has them. 🔃 changed
  - Previously we thought every Message without a parent could be a Thread, but that would mean the "Thread view" would need a read marker for each of them as they are all threads.
[...]

PR #15369, opened 2025-06-17
feat(chat): Simplify the mark as unread

nickvergessen:
## 🛠️ API Checklist
- Ref #15313 

### :construction: Tasks
- [ ] Marking a specific message unread can no longer be supported 🚷 (please use reminder instead?)
  - [ ] API will stay in-tact to not break clients that did not yet implement threads, but clients that support threads should not call this anymore as the behaviour for threads that are in the rage of now unread messages can not be restored easily
- [x] Marking as unread can no longer be supporting the count 🚷 (like on whatsapp, telegram, teams, slack, …)
- [ ] When reading the last message of conversation we can: 🔃 changed
  - [ ] Remove all thread-read-markers (that don't have a non-default notification setting)
  - [x] When marking conversation as unread again, we store an "out-of-band" marker that simply adds the dot
  - [x] When a further message happens marking the chat unread again, the number is simply shown again

### 🏁 Checklist

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -18,7 +18,7 @@
 * 🌉 **Sync with other chat solutions** With [Matterbridge](https://github.com/42wim/matterbridge/) being integrated in Talk, you can easily sync a lot of other chat solutions to Nextcloud Talk and vice-versa.
 ]]></description>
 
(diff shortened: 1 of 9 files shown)


nickvergessen:
Closing for now

PR #15375, opened 2025-06-19
fix(TalkDashboard): make home button a link for a11y and UX and add refresh function 

### LINKED ISSUE #15321: Make home button primary when in home view
Just like with other navigation elelments, I think this button should become primary when the home view is active.

Image

DorraJaouad:
### ☑️ Resolves

* Fix #15321

When current page is dashboard.

🏚️ Before | 🏡 After
-- | --
image | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/Dashboard/TalkDashboard.vue
+++ b/src/components/Dashboard/TalkDashboard.vue
@@ -56,33 +56,44 @@ const upcomingReminders = computed(() => dashboardStore.upcomingReminders || [])
 const eventsInitialised = computed(() => dashboardStore.eventRoomsInitialised)
 const remindersInitialised = computed(() => dashboardStore.upcomingRemindersInitialised)
 const conversationName = ref('')
-let actualiseDataInterval: ReturnType<typeof setInterval> | null = null
+let actualizeDataInterval: ReturnType<typeof setInterval> | null = null
 
 // Data fetching handlers
 
 /**
  * Fetches all necessary data for the dashboard.
  */
-async function actualiseData() {
+async function actualizeData() {
 	await Promise.all([
 		dashboardStore.fetchDashboardEventRooms(),
 		dashboardStore.fetchUpcomingReminders(),
 	])
 }
 
-onMounted(() => {
-	actualiseData()
-	actualiseDataInterval = setInterval(actualiseData, 300_000)
-})
+/**
+ * Initializes the data fetching interval and fetches initial data.
+ */
+function initActualizeData() {
+	if (actualizeDataInterval) {
+		clearInterval(actualizeDataInterval)
+	}
+	actualizeData()
+	actualizeDataInterval = setInterval(actualizeData, 300_000)
+}
+
+initActualizeData()
+EventBus.on('refresh-talk-dashboard', initActualizeData)
 
 onBeforeUnmount(() => {
-	if (actualiseDataInterval) {
-		clearInterval(actualiseDataInterval)
+	if (actualizeDataInterval) {
+		clearInterval(actualizeDataInterval)
 	}
 
 	if (eventCardsWrapper?.value) {
 		resizeObserver.disconnect()
 	}
+
+	EventBus.off('refresh-talk-dashboard', initActualizeData)
 })
 
 watch(eventCardsWrapper, (newValue) => {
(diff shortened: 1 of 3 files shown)


src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -508,6 +509,16 @@ export default {
...
 		},
+
+		isInDashboard() {
+			return this.$route.name === 'root'
+		},
+
+		dashboardButtonLabel() {
+			return this.isInDashboard
+				? t('spreed', 'Reload dashboard')
+				: t('spreed', 'Talk home')
```
ShGKme: It is confusing that sometimes this button is for `home`, but sometimes for `dashboard`.

src/services/EventBus.ts
```diff
@@ -36,6 +36,7 @@ export type Events = {
 	'poll-editor-open': { token: string, id: number | null, fromDrafts: boolean, action?: string, selector?: string }
 	'refresh-peer-list': void
 	'retry-message': number
+	'refresh-talk-dashboard': void
```
ShGKme, line 38: 🥲

src/components/LeftSidebar/LeftSidebar.vue
DorraJaouad: home only on the surface, dashboard BTS :p

DorraJaouad: maybe `Reload Talk home` is more consistent here

ShGKme: The page itself has no accessible name (the heading is "Hello, displayName").
I'd use dashboard as this is the name used in the blog and in the changelog.

But the important part is to keep it consistent on a single button. Otherwise for a screen reader it is:
1. "Link, Talk home, Talk home" (some screen-readers doesn't handle title duplicating aria-label)
2. *click*
3. "Link, Reload dashboard, current page" (it was for home, and not it is for some dashboard without moving to a new location).

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -9,12 +9,13 @@
...
+						:variant="isInDashboard ? 'primary' : 'tertiary'"
 						:class="{ 'hidden-visually': isSearching }"
 						class="talk-home-button"
-						:title="t('spreed', 'Talk home')"
-						:aria-label="t('spreed', 'Talk home')"
-						@click="showTalkDashboard">
+						:title="dashboardButtonLabel"
+						:aria-label="dashboardButtonLabel"
+						:to="{ name: 'root' }"
+						@click="refreshTalkDashboard">
```
ShGKme, line 18: To fix in `nextlcoud/vue`: `aria-current` is missing.

src/components/Dashboard/TalkDashboard.vue
```diff
@@ -73,6 +73,8 @@ async function actualiseData() {
 onMounted(() => {
 	actualiseData()
 	actualiseDataInterval = setInterval(actualiseData, 300_000)
+
+	EventBus.on('refresh-talk-dashboard', actualiseData)
 })
```
Antreesy: I'd restart a timer here in that case. Also could be called earlier without onMounted hook?
```suggestion
initActualizeData()
EventBus.on('refresh-talk-dashboard', initActualizeData)

function initActualizeData() {
	if (actualiseDataInterval) {
		clearInterval(actualiseDataInterval)
	}
	actualiseData()
	actualiseDataInterval = setInterval(actualiseData, 300_000)
}
```

Antreesy (commented):
Otherwise fine

src/components/LeftSidebar/LeftSidebar.vue
Antreesy: IIRC we should not use it to not conflict with Nextcloud Dashboard. So `reload Talk home` sounds valid here from my pov

src/components/Dashboard/TalkDashboard.vue
ShGKme: In this case it should be `setTimeout` (it is restarted on each iteration anyway)

Antreesy: Unless you're staying on home page longer than 5 minutes, then `actualiseData()` should set it again

Antreesy (approved):
What was wrong with `actualiSe`?)

DorraJaouad:
> What was wrong with `actualiSe`?)

I am used to `actualize` and we are mainly using american english so..

PR #15376, opened 2025-06-19
perf(frontend): lazy load `libphonenumber-js`

ShGKme:
### ☑️ Resolves

* Ref https://github.com/nextcloud/spreed/issues/9406
* `libphonenumber-js` is quite a large library
  * 2 used functions alone: `121 kB` after tree-shaking [bundlejs](https://bundlejs.com/?q=libphonenumber-js%401.12.9&treeshake=%5B%7B+parsePhoneNumberFromString%2CvalidatePhoneNumberLength+%7D%5D)
* `120kb` is not a big deal for our large bundle, but:
  * It is only used in a rare case - on a phone number search (not supported in many instances)
  * Many small libraries make a larger part later
* Added `useAsyncInit` helper to be used in other similar cases

Entity | 🏚️ Before | 🏡 After
-- | -- | --
`/js` | `50_384 kb` | `49_900 kb` (🔻1%)
`talk-main.js` | `7_549_282 b` | `7_425_859 b` (🔻1.6%)

image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/composables/useLibphonenumber.ts
+++ b/src/composables/useLibphonenumber.ts
@@ -1,0 +1,33 @@
+/*
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import { useAsyncInit } from './useAsyncInit.ts'
+
+/**
+ * Dynamically import libphonenumber-js module
+ */
+async function loadLibphonenumber() {
+	// Destructure immediately to help bundlers to tree-shake unused parts
+	const { parsePhoneNumberFromString, validatePhoneNumberLength } = await import('libphonenumber-js')
+	return {
+		parsePhoneNumberFromString,
+		validatePhoneNumberLength,
+	}
+}
+
+/**
+ * Use libphonenumber-js asynchronously loaded module
+ */
+export function useLibphonenumber() {
+	const {
+		isReady: isLibphonenumberReady,
+		result: libphonenumber,
+	} = useAsyncInit(loadLibphonenumber, true)
+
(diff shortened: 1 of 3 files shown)


ShGKme:
I forgot to test after refactoring

ShGKme:
> I forgot to test after refactoring

I forgot to immediately init the init.

ShGKme:
And I forgot the SPDX header 👀

src/composables/useLibphonenumber.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+/**
+ * Use libphonenumber-js asynchronously loaded module
+ */
+export function useLibphonenumber() {
+	const {
+		isReady: isLibphonenumberReady,
+		result: libphonenumber,
+	} = useAsyncInit(loadLibphonenumber, true)
```
Antreesy, line 27: So it will be triggered with component setup, not with the actual function call? Not a big deal, just to understand the concept

ShGKme: It is defined by the second param (`immediate`).

In this case - yes, with `useAsyncInit(..., true)` it will be called immediately on component creations. Because this component needs it immediately when it is used (it is already a phone number search result).

Alternative here would be to use `defineAsyncComponent`, but then it needs a loading placeholder for an entire component, and a wrapper to make sure it is used this way in all 3 places.

I found a composable inside the componwent a simpler solution.

In comparison, in [this PR](https://github.com/nextcloud/spreed/pull/15378) the same composable is used without `immediate` to init media recorder when a user clicks on the button, not when it is rendered and never used in fact.

PR #15377, opened 2025-06-19
perf(frontend): replace `lodash` with `es-toolkit`

ShGKme:
### ☑️ Resolves

* Ref: https://github.com/nextcloud/spreed/issues/9406
* `Lodash` is a very old library with low performance and sometimes issues on tree-shaking
* Many functions from `lodash` are not needed in modern JS
* In runtime: replaced with our own utils
* In tests: replaced with `es-toolkit`, a good well-known replacement
  * It also has some useful utils we can use
  * Alternatives:
    * Individual libraries, like `klona` and `dequel` for deep cloning/comparison (even smaller)

Entity | 🏚️ Before | 🏡 After
-- | -- | --
`/js` | `164_512 kb` | `160_216 kb` (🔻2.4%)
`talk-main.js` | `6_072_269 b` | `5_983_249 b` (🔻1.5%)

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/test-helpers.js
+++ b/src/test-helpers.js
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-import { cloneDeep } from 'lodash'
+import { cloneDeep } from 'es-toolkit'
 import NcActionButton from '@nextcloud/vue/components/NcActionButton'
 import NcActionText from '@nextcloud/vue/components/NcActionText'
 import NcButton from '@nextcloud/vue/components/NcButton'
--- a/src/utils/e2ee/encryption.js
+++ b/src/utils/e2ee/encryption.js
@@ -6,10 +6,10 @@
 import Olm from '@matrix-org/olm'
 import base64js from 'base64-js'
 import debounce from 'debounce'
-import { isEqual } from 'lodash'
 import { v4 as uuidv4 } from 'uuid'
 import { getTalkConfig, hasTalkFeature } from '../../services/CapabilitiesManager.ts'
 import Signaling from '../signaling.js'
+import { isEqualJson } from '../utils.ts'
 import Peer from '../webrtc/simplewebrtc/peer.js'
 import SimpleWebRTC from '../webrtc/simplewebrtc/simplewebrtc.js'
 import { importKey, ratchet } from './crypto-utils.js'
@@ -501,7 +501,7 @@ class Encryption {
 			const key = base64js.toByteArray(decoded.key)
 			const index = decoded.index
 
-			if (!isEqual(sessionData.lastKey, key)) {
+			if (!isEqualJson(sessionData.lastKey, key)) {
 				sessionData.lastKey = key
 				console.debug('Key updated', sessionId, index, decoded.key)
 				this.context.setKey(sessionId, key, index)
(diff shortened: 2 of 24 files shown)


Antreesy:
I was thinking, that for our use case even 'cloneDeep' - > 'structuredClone' is enough, as we're only cloning data objects?

ShGKme:
> I was thinking, that for our use case even 'cloneDeep' - > 'structuredClone' is enough, as we're only cloning data objects?

`structuredClone` is actually very powerful.

For data objects `JSON.parse(JSON.stringify())` is enough and superfast.

But I didn't check if it works in all the cases.

Nice catch!

ShGKme:
Updated:
- Use local `cloneDeepJson` and `isEqualJson` utils instead of external libraries
- Use `es-toolkit` only for development (tests)

src/utils/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+	return JSON.parse(JSON.stringify(obj)) as T
+}
+
+/**
+ * Compare two plain JS objects deeply via JSON serialization
+ *
+ * @param a - First object
+ * @param b - Second object
+ */
+export function isEqualJson<T>(a: T, b: T): boolean {
```
Antreesy: Is it correct to assume, that `a` is the same type as `b`? I would say T should be at least unknown by default

src/utils/e2ee/encryption.js
```diff
@@ -500,7 +500,7 @@ class Encryption {
 			const key = base64js.toByteArray(decoded.key)
 			const index = decoded.index
 
-			if (!isEqual(sessionData.lastKey, key)) {
+			if (!isEqualJson(sessionData.lastKey, key)) {
```
Antreesy, line 504: Wasn't sure, but tested. Keys are `Uint8Array`, and `base64js.toByteArray` returns `Array`. So should be fine to use it here?

image

src/test-helpers.js
```diff
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-import { cloneDeep } from 'lodash'
+import { cloneDeep } from 'es-toolkit'
```
Antreesy, line 6: This one is JSON payload

src/utils/e2ee/encryption.js
ShGKme: Yes. It wouldn't work with `cloneDeepJson`, but works with comparison.

src/test-helpers.js
ShGKme: Yes, but it is only used in tests.

Some other places in tests require `cloneDeep` anyway.

Alternative here is `structuredClone`, but then it requires `toRaw` in Vue 3 for reactive objects.

src/utils/utils.ts
ShGKme: Yes, better to make both `unknown`

src/test-helpers.js
Antreesy: Works with RsPack and Vitest, so should be fine

Antreesy:
Rebased

Antreesy:
Rebased again, switched to `structuredClone` (a bit more safe in cloning object, and we're using it in two places on demand, so not critical)

Antreesy (approved):

PR #15378, opened 2025-06-19
perf(frontend): lazy load and init `extendable-media-recorder`

ShGKme:
### ☑️ Resolves

* Ref https://github.com/nextcloud/spreed/issues/9406
* `extendable-media-recorder` is quite a large library
  * 3 used functions alone: `152 kB` after tree-shaking [bundlejs](https://bundlejs.com/?q=extendable-media-recorder%409.2.27%2Cextendable-media-recorder-wav-encoder%407.0.129&treeshake=%5B%7B+register%2CMediaRecorder+%7D%5D%2C%5B%7B+connect+%7D%5D)
  * Also, it requires additional initialization
* `152 kb` is not a big deal for our large bundle, but:
  * It is only used with voice messages. Many users don't even use this feature
  * No need to load and init it in advance
  * Many small libraries make a larger part later

Entity | 🏚️ Before | 🏡 After
-- | -- | --
`/js` | `50_384 kb` | `49_740 kb` (🔻1.3%)
`talk-main.js` | `7_549_282 b` | `7_391_420 b` (🔻2%)

image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


src/components/NewMessage/NewMessageAudioRecorder.vue
```diff
@@ -9,8 +9,9 @@
 			v-if="!isRecording"
 			:title="startRecordingTitle"
 			:aria-label="startRecordingTitle"
+			:aria-description="isMediaRecorderLoading && t('spreed', 'Loading …')"
```
Antreesy: that would be an empty attribute, if `false`, I guess

Antreesy (approved):

src/components/NewMessage/NewMessageAudioRecorder.vue
ShGKme: If was changed in Vue 3
https://v3-migration.vuejs.org/breaking-changes/attribute-coercion.html#_3-x-syntax

ShGKme: But maybe here we need explicit undefined/null value instead

ShGKme: Replaced with
```
:aria-description="isMediaRecorderLoading ? t('spreed', 'Loading …') : undefined"
```

Issue #14997, opened 2025-04-30
Extract CSS from bundle


ShGKme:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Feature request

**Is your feature request related to a problem? Please describe.**

JS Bundle size is quite large, and it includes CSS.
It increases JS load time.

**Describe the solution you'd like**

Extract CSS from a JS bundle.

RubensRainelli:
Moving the CSS outside would even give benefits during compression, allowing to reduce the total load time.

PR #15379, opened 2025-06-19
perf(frontend): extract CSS from JS

ShGKme:
### ☑️ Resolves

* Ref: https://github.com/nextcloud/spreed/issues/9406
* Fix: https://github.com/nextcloud/spreed/issues/14997
* Also, compile all the styles using bundler

The current implementation doesn't use shared chunks:
- Every entrypoint has its own complete CSS, even ~80% is the same
- Webpack can handle it to move all the shared CSS into a new shared chunk (or many chunks)
- However, unlike JS, such shared CSS chunks must be loaded manually, with is complicated with PHP templates...

Entity | 🏚️ Before | 🏡 After
-- | -- | --
`/js` | `50_384 kb` | `48_052 kb` (🔻4.6%)
`talk-main.js` | `7_052_979 b` | `7_425_859 b` (🔻6.6%)
`/css` | `32 kb` | `2_124 kb` (🔺650%)

image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


nickvergessen:
In the past we migrated to this, as the second request (per app) adds much more problems to the server at the moment then stripping the kbs is saving

ShGKme:
Rebased onto `main`, no changes were required.

PR #15384, opened 2025-06-20
test: do not mock session storage globally

Antreesy:
Signed-off-by: Maksim Sukharev <antreesy.web@gmail.com>### ☑️ Resolves

* Splitted from #15349
* Drop non-compatible with Jest 30 libraries

### 🏁 Checklist


Antreesy:
> Doesn't it work without mocking the local storage?

It does, good catch

PR #15400, opened 2025-06-23
feat(threads): Add database structure and models

nickvergessen:
## 🛠️ API Checklist

- Ref #9679

### 🚧 Tasks

- [x] DB schema and models

### 🏁 Checklist

### DIFF
--- a/openapi-backend-sipbridge.json
+++ b/openapi-backend-sipbridge.json
@@ -428,6 +428,13 @@
                             },
                             "silent": {
                                 "type": "boolean"
+                            },
+                            "threadId": {
+                                "type": "integer",
+                                "format": "int64"
+                            },
+                            "isThread": {
+                                "type": "boolean"
                             }
                         }
                     }
--- a/openapi-federation.json
+++ b/openapi-federation.json
@@ -428,6 +428,13 @@
                             },
                             "silent": {
                                 "type": "boolean"
+                            },
+                            "threadId": {
+                                "type": "integer",
+                                "format": "int64"
+                            },
+                            "isThread": {
+                                "type": "boolean"
                             }
                         }
                     }
--- a/openapi-full.json
+++ b/openapi-full.json
@@ -628,6 +628,13 @@
                             },
                             "silent": {
                                 "type": "boolean"
+                            },
+                            "threadId": {
+                                "type": "integer",
+                                "format": "int64"
+                            },
+                            "isThread": {
+                                "type": "boolean"
                             }
                         }
                     }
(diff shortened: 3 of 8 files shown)


lib/Controller/ThreadController.php
```diff
@@ -0,0 +1,173 @@
...
+			$this->timeFactory->getDateTime(),
+			false,
+			null,
+			$comment,
+			true,
+			true
+		);
+
+		$this->threadService->addAttendeeToThread($this->participant->getAttendee(), $thread);
+		return new DataResponse($thread->jsonSerialize(), Http::STATUS_OK);
```
miaulalala, line 171: **nit**
```suggestion
		} catch (DoesNotExistException) {
			$thread = $this->threadService->createThread($this->room, $threadId);
			$this->chatManager->addSystemMessage(
				$this->room,
				$this->participant->getAttendee()->getActorType(),
				$this->participant->getAttendee()->getActorId(),
				json_encode(['message' => 'thread_created', 'parameters' => ['thread' => $threadId]]),
				$this->timeFactory->getDateTime(),
				false,
				null,
				$comment,
				true,
				true
			);
		}
		$this->threadService->addAttendeeToThread($this->participant->getAttendee(), $thread);
		return new DataResponse($thread->jsonSerialize(), Http::STATUS_OK);
```

lib/Model/Message.php
```diff
@@ -194,8 +194,11 @@ public function toArray(string $format): array {
 			$reactions = new \stdClass();
 		}
 
+		$id = (int)$this->getComment()->getId();
+		$threadId = (int)$this->getComment()->getTopmostParentId() ?: $id;
```
miaulalala, line 198: Will the `(int)` conversion convert `null` to `0` here?

lib/Model/Thread.php
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	/**
+	 * @return TalkThread
+	 */
+	#[\Override]
+	public function jsonSerialize(): array {
+		return [
+			'id' => max(1, $this->getId()),
+			'roomId' => max(1, $this->getRoomId()),
```
miaulalala, line 44: Initialised with `0` but max checks on `1` - intentional?

Also, why the `max` on each value if the default is already 0?

lib/Model/ThreadAttendee.php
```diff
@@ -0,0 +1,76 @@
...
+	 * @return TalkThreadAttendee
+	 */
+	#[\Override]
+	public function jsonSerialize(): array {
+		return [
+			'notificationLevel' => min(3, max(0, $this->getNotificationLevel())),
+			'lastReadMessage' => max(0, $this->getLastReadMessage()),
+			'lastMentionMessage' => max(0, $this->getLastMentionMessage()),
+			'lastMentionDirect' => max(0, $this->getLastMentionDirect()),
+			'readPrivacy' => min(1, max(0, $this->getReadPrivacy())),
```
miaulalala, line 73: intentional that some values are stripped when serializing?

lib/Controller/ThreadController.php
nickvergessen: > Do early returns

Does early returns

> Combine paths to not duplicate 2 lines

(╯°□°）╯︵ ┻━┻

In the future more things will be added in the "DoesNotExistException" code path, so I don't think combining makes sense.

lib/Model/Message.php
nickvergessen: there is no null. `getTopmostParentId()` returns a string in "worst case" `'0'` (no null) and it works as intended here: https://3v4l.org/2kFoa

lib/Model/Thread.php
nickvergessen: I don't want to initialize values with a "random" id.
I had to add the max() in the serializes as otherwise Psalm complains on the definitions, or we can not tell a client that this will never be negative.

lib/Model/ThreadAttendee.php
nickvergessen: notification level is limited from 0-3 and read privacy is either 0 or 1
so nothing is "lost"?

Antreesy:
part in `appinfo/routes` is missing, or we don't do it anymore?

ShGKme:
> part in `appinfo/routes` is missing, or we don't do it anymore?

Works via `#[ApiRoute()]`

nickvergessen:
What Grigorii said, if you want to see the routes as a list, try the newly added `occ router:list spreed`

PR #15401, opened 2025-06-23
fix(vue3): provide keys for components, that could be reused

Antreesy:
### ☑️ Resolves

* Fix leaking component data
* passing `key` forces component to rerender (if submenu page was changed)
  * before `close-after-click` could be leaked as `true` to a new component, which should not have it

### 🚧 Tasks

- [ ] Alternative: check it in vue-library instead?

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/LeftSidebar/ConversationsList/Conversation.vue
+++ b/src/components/LeftSidebar/ConversationsList/Conversation.vue
@@ -129,7 +129,9 @@
 				</NcActionButton>
 			</template>
 			<template v-else-if="submenu === 'notifications'">
-				<NcActionButton :aria-label="t('spreed', 'Back')"
+				<NcActionButton
+					key="action-back"
+					:aria-label="t('spreed', 'Back')"
 					@click.stop="submenu = null">
 					<template #icon>
 						<IconArrowLeft class="bidirectional-icon" :size="16" />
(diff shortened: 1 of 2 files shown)


src/components/LeftSidebar/ConversationsList/Conversation.vue
```diff
@@ -129,7 +129,9 @@
 				</NcActionButton>
 			</template>
 			<template v-else-if="submenu === 'notifications'">
-				<NcActionButton :aria-label="t('spreed', 'Back')"
+				<NcActionButton
+					key="action-back"
```
nickvergessen, line 133: does this refresh good enough even when its reused inside a parent that has a different key? Just asking for my understanding

Antreesy: if parent has a different key, it should re-render parent and everything underneath

Antreesy:
> Nearest parent is `NcListItem` (skipping templates) and it has key updated by `token` so it should be enough?

It's not about Virtual scroller, but switching submenus (see above)

src/components/LeftSidebar/ConversationsList/Conversation.vue
ShGKme: > does this refresh good enough even when its reused inside a parent that has a different key? Just asking for my understanding

Key doesn't trigger re-rendring, but re-order/re-mount.

And Vue doesn't support parent remounting without children remounting.

So changing the parent key completely destroys and re-creates children. Not just re-render.

PR #15406, opened 2025-06-24
fix: adjust Jest to Vue3

Antreesy:
### ☑️ Resolves

* Adjust Jest for Vue3 usage

>[!WARNING]
> All tests in `src/components` are disabled
> Reason: Jest + ESM = 💔

### DIFF
--- a/src/store/conversationsStore.spec.js
+++ b/src/store/conversationsStore.spec.js
@@ -146,7 +146,7 @@ describe('conversationsStore', () => {
 			})
 			store.dispatch('addConversation', testConversation)
 
-			expect(store.getters.conversation(testToken)).toBe(testConversation)
+			expect(store.getters.conversation(testToken)).toStrictEqual(testConversation)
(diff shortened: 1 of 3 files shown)


Antreesy:
`Cannot find module '@nextcloud/upload' from 'src/test-setup.js'` 😬

jest.config.js
```diff
@@ -48,16 +49,20 @@ module.exports = {
 	setupFilesAfterEnv: [
 		'<rootDir>/src/test-setup.js',
 	],
-	globalSetup: resolve(__dirname, 'jest.global.setup.js'),
+	globalSetup: './jest.global.setup.js',
```
ShGKme: Without resolve it only works then run from the root

ShGKme (approved):
Thanks!

Antreesy:
Switched from match regex to
```js
	testPathIgnorePatterns: [
		'<rootDir>/src/components',
		'<rootDir>/src/store/fileUploadStore.spec.js',
	],
```

PR #15409, opened 2025-06-25
fix(vue3): fullscreen mode in Vue 3 container 

ShGKme:
### ☑️ Resolves

* Fullscreen layout due to mounting inside `#content` instead of replacing it

🏚️ Before | 🏡 After
-- | --
image | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -692,15 +692,15 @@ body .modal-wrapper * {
 }
 
 // Styles for the app content at fullscreen mode
+:root:has(body.talk-in-fullscreen) {
+	--body-container-margin: 0px !important;
+	--body-container-radius: 0px !important;
+	--header-height: 0px !important;
+}
+
 body.talk-in-fullscreen {
 	#header {
 		display: none !important;
-	}
-	#content-vue {
-		margin: 0;
-		height: 100%;
-		width: 100%;
-		border-radius: 0;
 	}
 }
 


src/App.vue
```diff
@@ -693,9 +693,19 @@ body .modal-wrapper * {
...
+	--body-container-radius: 0 !important;
+	--header-height: 0 !important;
+
 	#header {
 		display: none !important;
 	}
+
+	#content {
+		/* TODO: use --header-height variable on the server */
+		margin-top: 0;
```
DorraJaouad: `margin: 0` as long as we have 
```
#header {
		display: none !important;
	}
```

regardless of `--header-height`

and that is replacement for

```
#content-vue {
	margin: 0;
	border-radius: 0;
}
```

which are not needed anymore as it's the child element now

DorraJaouad (commented):
style 
```
	// Fix fullscreen black bar on top
	&:fullscreen {
		padding-top: 0;

		:deep(.app-sidebar) {
			height: 100vh !important;
		}
	}
```

may not be used

src/App.vue
Antreesy: See https://github.com/nextcloud/server/pull/53697 for a follow-up

ShGKme: Done

Antreesy (approved):

ShGKme:
The problem with an empty space (reproducible only with the default theme) fixed by applying styles to the `:root`

Antreesy (approved):
Tested

PR #15412, opened 2025-06-26
fix(public):  set guest names correctly

Antreesy:
### ☑️ Resolves

* Fix guest name setting on public pages
  * correctly get info from browserstorage, show popups and connect with public user menu
  * fix style issues

🏚️ Before | 🏡 After
-- | --
image | image
image |  image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/GuestWelcomeWindow.vue
+++ b/src/components/GuestWelcomeWindow.vue
@@ -130,7 +130,6 @@ export default {
 <style lang="scss" scoped>
 .modal__content {
 	padding: calc(var(--default-grid-baseline) * 3);
-	background-color: var(--color-main-background);
 }
 
 .conversation-information {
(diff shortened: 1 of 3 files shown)


src/App.vue
```diff
@@ -223,14 +223,15 @@ export default {
...
-	beforeMount() {
+	async beforeMount() {
 		if (!getCurrentUser()) {
 			EventBus.once('joined-conversation', () => {
 				this.fixmeDelayedSetupOfGuestUsers()
 			})
 			EventBus.on('should-refresh-conversations', this.debounceRefreshCurrentConversation)
 		}
 
+		await this.$router.isReady()
```
DorraJaouad: Better to rely on token value passed in `this.fetchSingleConversation(params.token)` in all instances where token is needed befor router is init rather than delaying mounting the app

Antreesy: Will address in separate PR, dropping commit for now

PR #15424, opened 2025-06-27
fix: adjust logic of 'conversation-received' event emitting

Antreesy:
### ☑️ Resolves

* Separated from #15412 and reworked
  * as initial route is '/', and guests are redirected to not-found, we should either wait for router to be ready, or do not use it

🏚️ Before | 🏡 After
-- | --
image | --
2025-06-26_10h23_33 | --

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/LeftSidebar/LeftSidebar.vue
+++ b/src/components/LeftSidebar/LeftSidebar.vue
@@ -258,6 +258,7 @@ import { t } from '@nextcloud/l10n'
 import { useIsMobile } from '@nextcloud/vue/composables/useIsMobile'
 import debounce from 'debounce'
 import { ref } from 'vue'
+import { START_LOCATION } from 'vue-router'
 import NcActionButton from '@nextcloud/vue/components/NcActionButton'
 import NcActionCaption from '@nextcloud/vue/components/NcActionCaption'
 import NcActions from '@nextcloud/vue/components/NcActions'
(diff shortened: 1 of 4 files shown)


src/store/conversationsStore.js
```diff
@@ -963,7 +964,7 @@ const actions = {
 			talkHashStore.clearMaintenanceMode()
 			const response = await fetchConversation(token)
 			talkHashStore.updateTalkVersionHash(response)
-			dispatch('addConversation', response.data.ocs.data)
+			await dispatch('addConversation', response.data.ocs.data)
```
ShGKme: `dispatch` always returns a promise, but `addConversation` action specifically is not async, so this promise is just `Promise.resolve()`.

If we need to wait one `Promise`, let's explicitly use `await Promise.resolve()` with a comment to explain it.

Otherwise let's remove `await` to not make it confusing.

src/components/LeftSidebar/LeftSidebar.vue
```diff
@@ -948,7 +949,7 @@ export default {
 			}
 			if (this.isMobile) {
 				emit('toggle-navigation', {
-					open: false,
+					open: to.name === 'root' && from.name === undefined,
```
ShGKme: https://router.vuejs.org/api/variables/START_LOCATION.html
```suggestion
					open: to.name === 'root' && from === START_LOCATION,
```

ShGKme (approved):

Issue #9680, opened 2023-06-02
Frontend - Message threads


nickvergessen:
* Overview at https://github.com/nextcloud/spreed/issues/1469
* Rough idea is that when a reply message or a message that has a reply is clicked, we filter the chat stream with that "thread" ID (root parent message).
* All newly written messages should be addresses to that root parent as well (needs a new API parameter, so that you can still reply within a thread for quoting and reactions)
* The UI of course needs to see a button to "go back" to the unfiltered list.

sascha08-15:
+1 from me - I would love to see threads (similar to slack), not sure about the last point of the ticket tho (go back / unfiltered).

jooooaah:
+1 whould be helpful in many ways

Jacques-Olivier-Farcy:
We could use it everyday, must have!

max-nextcloud:
UX wise I just used zulip for the first time and found the way they use [topics](https://zulip.com/help/introduction-to-topics) instead of threads quite intuitive and maybe a bit simpler as right now I see no way of deep nested threads which for me is a plus.

szaimen:
> UX wise I just used zulip for the first time and found the way they use [topics](https://zulip.com/help/introduction-to-topics) instead of threads quite intuitive and maybe a bit simpler as right now I see no way of deep nested threads which for me is a plus.

This reminds me of what I proposed in https://github.com/nextcloud/spreed/issues/2448 :)

Antreesy:
Follow-ups:
- [x] Integrations (Files sidebar) support
- [ ] Restructure Shared items -> Details tab
- [x] update thread info from message polling
- [x] Specific message fetching (context + history => thread messages only, do not mix with general chat)
- [x] Parsing of messages to human-readable format
- [x] Focus error when passing to thread tab
- [x] Return state from thread tab
- [ ] Read marker in threaded view is semantically broken
- [x] Patch update threadIn from message polling
- [x] Thread creation action must be embedded with a reply action IMO
- [ ] Return from threaded view in Call mode given that top bar is not usable at that state

PR #15433, opened 2025-06-30
feat(threads): add handling of API endpoints

Antreesy:
### ☑️ Resolves

* Ref #9680

>[!IMPORTANT]
> [wip] frontend is here: `feat/9680/threads-frontend`

🏚️ Main | 🏡 Thread
-- | --
image | image
image | image
image | image
image | image

### 🚧 Tasks
- [ ] UI
  - [ ] 'Load more' logic for thread overviews
  - [ ] Loading state
  - [ ] Replies in correct place

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


src/services/messagesService.ts
```diff
@@ -229,18 +233,62 @@ const setConversationUnread = async function(token: string, options?: AxiosReque
...
+}
+
+/**
+ * Create a new thread for a conversation
+ *
+ * @param token The conversation token
+ * @param messageId The message id of any message belonging to the future thread
+ * @param [options] Axios request options
+ */
+async function makeThreadForConversation(token: string, messageId: number, options?: AxiosRequestConfig): makeThreadResponse {
```
nickvergessen: Not sure you noticed, I chaged that in the last commits as well so it also returns TalkThreadInfo instead of TalkThread only

src/stores/chatExtras.ts
```diff
@@ -0,0 +1,369 @@
...
+	getSingleThreadForConversation,
+	getThreadsForConversation,
+	makeThreadForConversation,
+	summarizeChat,
+} from '../services/messagesService.ts'
+import { parseMentions, parseSpecialSymbols } from '../utils/textParse.ts'
+
+type State = {
+	threads: Record<string, Record<number, ThreadInfo>>
+	threadsOffset: Record<string, number | undefined>
```
DorraJaouad: Why don't we just name it `firstKnownThreadId` ? offset is very vague

src/stores/chatExtras.ts
```diff
@@ -0,0 +1,369 @@
...
+					this.threads[token] = {}
+				}
+
+				const response = await getThreadsForConversation({ token, offsetId: offset ?? this.threadsOffset[token] })
+
+				const threadIds: number[] = []
+				response.data.ocs.data.forEach((threadInfo) => {
+					this.threads[token][threadInfo.thread.id] = threadInfo
+					threadIds.push(threadInfo.thread.id)
+				})
```
DorraJaouad: start looking at the min Id here in this loop and then only handle one-value-argument in updating the first known id (oldest)

src/stores/chatExtras.ts
```diff
@@ -0,0 +1,369 @@
...
+
+	actions: {
+		/**
+		 * Update known threads offset for given conversation
+		 *
+		 * @param token - conversation token
+		 * @param [offsetList] - known id or list of ids
+		 */
+		updateThreadsOffset(token: string, offsetList?: number | number[]) {
+			const normalizedList: number[] = Array.isArray(offsetList) ? offsetList : [offsetList].filter(Boolean)
```
DorraJaouad: So you don't have to do this handling

src/stores/chatExtras.ts
```diff
@@ -0,0 +1,369 @@
...
+			}
+		},
+
+		/**
+		 * Get list of threads for given conversation
+		 *
+		 * @param token - conversation token
+		 * @param [offset] - optional offset for threads list
+		 */
+		async getThreadsList(token: string, offset?: number) {
```
DorraJaouad: I don't think we will ever use offset other than the one present in store...

src/stores/chatExtras.ts
```diff
@@ -0,0 +1,369 @@
...
+			}
+		},
+
+		/**
+		 * Get a thread in given conversation
+		 *
+		 * @param token - conversation token
+		 * @param threadId - thread id to fetch
+		 */
+		async getSingleThread(token: string, threadId: number) {
```
DorraJaouad: `fetchSingleThread` would make more sense here because you are not getting from store even if it is present so basically it can be used to get an updated state of the thread too

src/stores/chatExtras.ts
```diff
@@ -0,0 +1,369 @@
...
+		 * Make reply chain a thread in given conversation
+		 *
+		 * @param token - conversation token
+		 * @param threadId - thread id to fetch
+		 */
+		async makeThread(token: string, threadId: number) {
+			try {
+				if (!this.threads[token]) {
+					this.threads[token] = {}
+				}
```
DorraJaouad: Maybe check for its existence first so we reduce accidental requests from components? unless it will also be used to update a thread which is unlikely

src/services/messagesService.ts
```diff
@@ -229,18 +233,62 @@ const setConversationUnread = async function(token: string, options?: AxiosReque
...
+	return axios.get(generateOcsUrl('apps/spreed/api/v1/chat/{token}/threads/{threadId}', { token, threadId }), options)
+}
+
+/**
+ * Fetch a list of threads for given conversation
+ *
+ * @param data the wrapping object
+ * @param data.token the conversation token
+ * @param [data.limit=25] Number of threads to return
+ * @param [data.offsetId=0] The last thread ID that was known, default 0 starts from the newest
```
DorraJaouad: oldest? last means the newest and in store you are setting min -> oldest, I am confused O.O

src/stores/chatExtras.ts
Antreesy: ~~Also fine by me. But 'Load more' logic is missing for now, so can't tell if it needs any use rather than being offset (and if it  is needed at all)~~

Upd: we need to reset offset to 0 to actualize data, so it's not really needed as first known. I'd replace with getFirstKnownId instead

src/stores/chatExtras.ts
Antreesy: how do you know which is the min one? What if server returns them in descending/ascending order?

I like the idea to gather number[] and then iterate over it once more, than check if we need updating or not in this loop

src/services/messagesService.ts
Antreesy: 0 will return you [150,149,...126]
Then you would need to request with 126 to get [125,124,...100] and so on

src/stores/chatExtras.ts
Antreesy: 'Load more' is the only place to use it with a number.
But we'll likely need to update all threads? and then we would need to fetch them all with 0 (reset offset)

src/stores/chatExtras.ts
DorraJaouad:
``` js
let x = A[0] 
A.forEach(item => {
	if (item < x) {
	x = item
} 
}
```

src/stores/chatExtras.ts
DorraJaouad: No, we can still use the one in store as it is always up to date

src/services/messagesService.ts
DorraJaouad: I understand, but calling as last conveys the last thread id that was created hence the most recent and does no reflect that it is the last item in the array, also we do not know if the server will always maintain that order.

src/constants.ts
```diff
@@ -47,6 +47,10 @@ export const CHAT = {
 	FETCH_NEW: 1,
 } as const
 
+export const THREAD = {
+	FETCH_LIMIT: 2,
```
DorraJaouad: hmmm, depending on the design.. I expected it will be more. Let's keep it for now like this and we discuss it in the UI PR ;) 
UPD: if we are doing patch update with this limit, it shouldn't be this low tbh

src/stores/chatExtras.ts
```diff
@@ -100,64 +101,101 @@ export const useChatExtrasStore = defineStore('chatExtras', {
...
 		/**
-		 * Get a thread in given conversation
+		 * Fetch list of threads from server in given conversation
 		 *
 		 * @param token - conversation token
-		 * @param threadId - thread id to fetch
+		 * @param [offsetId] - optional offset for threads list. 0 to get the newest threads.
 		 */
-		async getSingleThread(token: string, threadId: number) {
+		async fetchThreadsList(token: string, offsetId?: number = 0) {
```
DorraJaouad: Again, it is much simpler if the firstKnownId is integrated in this method. Let the store handle it. If we don't have it, we just default to zero.

src/components/RightSidebar/RightSidebarContent.vue
```diff
@@ -289,6 +295,17 @@ function handleHeaderClick() {
...
+					v-if="!chatExtrasStore.hasNoMoreThreadsToLoad[token]"
+					variant="tertiary"
+					@click="loadMoreThreads">
+					{{ t('spreed', 'Load more threads') }}
+				</NcButton>
+				<NcButton
+					variant="tertiary"
+					@click="chatExtrasStore.updateThreadsList(token)">
+					{{ t('spreed', 'Update list of threads') }}
+				</NcButton>
```
DorraJaouad: image

src/stores/chatExtras.ts
```diff
@@ -100,64 +101,101 @@ export const useChatExtrasStore = defineStore('chatExtras', {
...
+		 * Updates list of threads from server in given conversation
+		 * If offsetId is not provided, we fetch threads starting from newest and need to update all of them
+		 * Otherwise, we only update threads that are older than offsetId
 		 *
 		 * @param token - conversation token
-		 * @param [offset] - optional offset for threads list
+		 * @param [offsetId] - optional offset for threads list. 0 to get the newest threads.
 		 */
-		async getThreadsList(token: string, offset?: number) {
+		async updateThreadsList(token: string, offsetId?: number = 0) {
```
DorraJaouad: It is a nice idea to add patch update but just one tiny note :) : 
```
response.data.ocs.data.forEach((threadInfo) => {
					this.threads[token][threadInfo.thread.id] = threadInfo
				})
```

src/stores/chatExtras.ts
```diff
@@ -100,64 +101,101 @@ export const useChatExtrasStore = defineStore('chatExtras', {
...
+					}
+					if (!maxReceivedId || maxReceivedId < threadInfo.thread.id) {
+						maxReceivedId = threadInfo.thread.id
+					}
+				})
+
+				// Clean up threads that are no longer present in the response
+				Array.from(threadsRequireUpdate).forEach((threadId) => {
+					if (threadId > (minReceivedId ?? 0) && threadId < (maxReceivedId ?? Infinity)) {
+						console.log('remove', threadId)
```
DorraJaouad: Just in case you forget it

src/stores/chatExtras.ts
DorraJaouad: No need for the optional marker because there is a default value

src/constants.ts
Antreesy: ups, that was for debugging. Should be 25

src/components/RightSidebar/RightSidebarContent.vue
Antreesy: It should happen with interval, probably

src/stores/chatExtras.ts
Antreesy: How is it different from current code?

DorraJaouad: You should also use it when fetching ;) just in case

src/components/RightSidebar/RightSidebarContent.vue
DorraJaouad: Logic wise, yes

src/stores/chatExtras.ts
Antreesy: There is fetchThreadsList for unknown items, and updateThreadsList for known, each serve it's own purpose. Or I am not understanding the comment

DorraJaouad: there is a token watcher that fetches on token change even when it has existed before so it serves as an update as well

src/components/RightSidebar/RightSidebarContent.vue
Antreesy: Or we can see, if it's enough to do it from messages polling only (instead of interval)

nimishavijay:
What do you think? Shows more info than before. Using a smaller font size in the first line brings more attention to the latest reply while still providing context. 

It however involves a lot of info: original poster, original poster's avatar, original message, latest poster's name, possibly latest poster's avatar, and latest message. Is requesting all that info possible and would it unreasonably affect the performance? 

image

With avatar: Not super keen on this but tried it anyway

image

Maybe not a topic for right now but I did think about if we should also toggle the view when compact mode is enabled, if so it would look something like this:

image



Other feedback from the screenshots:
- [ ] "Make thread" --> "Start a thread"
- [ ] "See thread" --> "Go to thread" (as it literally feels like you're going to a different conversation) or "Read thread"
- [ ] I see that all the threads are listed in the sidebar outside of the tabs. Is that an intermediate solution before we do either of the solutions in https://github.com/nextcloud/spreed/issues/15313#issuecomment-3019190538 or are those out of scope for now?

nickvergessen:
> "Make thread" --> "Start a thread"

Should be "Create thread" or "Create a thread" the first message could be older and from someone else, so are you really "starting" it?

Antreesy:
You mean, 'Start a thread' for new ones:
image

And 'Create a thread' for already posted:
image

nimishavijay:
From chat with @Antreesy 
- Remove last name of last poster (same as left conversation list)
- add a timestamp 
- unread replies not possible, but total number of replies is possible, so think about how to show unread indicator (like a dot) + total number of replies without making it look like unread counter

src/stores/chatExtras.ts
```diff
@@ -0,0 +1,361 @@
...
+					// Thread does not exist, nothing to update
+					return
+				}
+
+				this.threads[token][threadId] = {
+					thread: payload.thread ?? this.threads[token][threadId].thread,
+					attendee: payload.attendee ?? this.threads[token][threadId].attendee,
+					first: payload.first ?? this.threads[token][threadId].first,
+					last: payload.last ?? this.threads[token][threadId].last,
+				}
```
DorraJaouad: If you created this method to improve update performance, it does nothing as it is a re-assignment.
If you created this method to update the element, it doesn't make sense because you are assuming that the payload might have attributes missing which doesn't happen.

src/stores/chatExtras.ts
```diff
@@ -0,0 +1,361 @@
...
+			}
+		},
+
+		/**
+		 * Fetch list of unknown threads from server in given conversation
+		 * If there are no threads in the store, we start from the newest threads
+		 * Otherwise - from the first known thread id
+		 *
+		 * @param token - conversation token
+		 */
```
DorraJaouad: Outdated

src/services/messagesService.ts
```diff
@@ -229,18 +233,60 @@ const setConversationUnread = async function(token: string, options?: AxiosReque
...
+ * @param [options] Axios request options
+ */
+async function getRecentThreadsForConversation({ token, limit }: { token: string } & getRecentThreadsParams, options?: AxiosRequestConfig): getRecentThreadsResponse {
+	return axios.get(generateOcsUrl('apps/spreed/api/v1/chat/{token}/threads/recent', { token }), {
+		...options,
+		params: {
+			limit,
+		},
+	})
+}
```
DorraJaouad, line 260: so we won't have load more?

Antreesy: No. We'll get the search instead someday

src/stores/chatExtras.ts
Antreesy: I created it to be able to update the thread in fo from message polling (e.g. edited first message, new last message, bump amount of replies / last activity)

DorraJaouad (approved):
Basic handling ready to go

Antreesy:
Moving UI forntend to separate PR: ...

PR #15435, opened 2025-06-30
fix: migrate useMessageInfo and Quote to TS

Antreesy:
### ☑️ Resolves

* prerequisite to further logic/styles changes

No visual changes

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


src/composables/useMessageInfo.ts
```diff
@@ -14,21 +17,45 @@ import { ONE_DAY_IN_MS, ONE_HOUR_IN_MS } from '../utils/formattedTime.ts'
 import { getDisplayNameWithFallback } from '../utils/getDisplayName.ts'
 import { useConversationInfo } from './useConversationInfo.ts'
 
+/**
+ * Type guard for object
+ */
+function isDefined<T>(item: Ref<T | undefined>): item is Ref<T> {
+	return !!item.value
+}
```
ShGKme: Type guard must only check the type. If the checked type is `item is Ref<T>`, then the function must check that item is a Ref, and the ref's value has `T` type.

While this function checkes that ref's value is truthy

src/composables/useMessageInfo.ts
```diff
@@ -14,21 +17,45 @@ import { ONE_DAY_IN_MS, ONE_HOUR_IN_MS } from '../utils/formattedTime.ts'
 import { getDisplayNameWithFallback } from '../utils/getDisplayName.ts'
 import { useConversationInfo } from './useConversationInfo.ts'
 
+/**
+ * Type guard for object
+ */
+function isDefined<T>(item: Ref<T | undefined>): item is Ref<T> {
+	return isRef(item) && item.value !== undefined
+}
```
ShGKme: `isRef(item)` doesn't make sense for a param that is defiend as `Ref` already.

```suggestion
/**
 * Check whether ref's value is not undefined
 */
function isDefinedRef<T>(item: Ref<T | undefined>): item is Ref<T> {
	return item.value !== undefined
}
```

Antreesy: Damn, that's how I did it initially 🙈

ShGKme: 😶 Sorry

Yeah, we don't need to check for `isRef` as it is typed already as `Ref`.

But `!!item.value` was incorrect check for `isRef<T>` because `!!` checked for truthy value, which doesn't help to distingush between `Ref<T>` and `Ref<undefined>` because `Ref<false>` is also not `undefined`.

ShGKme (approved):

PR #15436, opened 2025-06-30
Bugfix/9679/subscribing to threads

nickvergessen:
- Ref #9679 
- [x] Create thread attendee when replying
- [x] Update thread info when replying

### 🏁 Checklist

  - Can not test yet as we don't have an API yet which would return the info if it worked


Antreesy (approved):
Tested from frontend side

nickvergessen:
Accidentally had modified/overwritten the `compareSearchResponse` testing method instead of duplicating it before.

PR #15437, opened 2025-06-30
fix(MediaSettings): set a horizontal layout and merge guest display name setting

### LINKED ISSUE #15244: [RFC] Media settings redesign
### How to use GitHub

---

## Feature request

- [x] Rearrange content to horizontal view

Current | Example
-- | --
Image | Image

- [ ] Rename heading / menu actions ("Devices" / "Mic and camera", e.t.c. - media is too broad)
  - [ ] Vary depending on place of application
- [ ] Split components into Dialog + Content
  - [x] Use Dialog in Talk settings (same as 'Check devices' on Dashboard)
  - [x] Use Content in lobby
    - [x] Combine with name input for guest 
  - [ ] 'You will see yourself mirrored in the call'

DorraJaouad:
### ☑️ Resolves

* Fix #15244

🏚️ Before | 🏡 After
-- | --
image | image
image| image
image |image
max content edge case |  image

### 🚧 Tasks

- [ ] Display name (required) needs more **visual** emphasis? 
- [ ] In addition to the "X" icon, add a close / cancel / go-back button next to the "Start call/Join"?

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


nimishavijay (changes requested):
_Edit: These mockups were part of the bigger change of moving the device checker from a dialog into the content when you start/join a call but are applicable if it stays a dialog as well_


Nice! Was just working on this before threads. My mockups:

**Start call:**

image


**Join call:**

image

**Lobby:**

image



Overall:
- [ ] The tabs for the devices/backgrounds can be regular buttons instead of wide buttons
- [ ] Use a different icon for backgrounds: [background-replace-outline](https://fonts.google.com/icons?icon.set=Material+Symbols&icon.size=24&icon.color=%23222222&selected=Material+Symbols+Outlined:background_replace:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=backgrou) icon
- [x] Always use the wording "Check devices" instead of media settings so we are consistent

Layout:
- [ ] Move the recording consent to the top as people have reported that they miss the checkbox as they have to scroll
- [ ] For the rest of the items, they are always in 2 columns, one with the preview and one with everything else including the start/join call button
  - [ ] Start/join call button is bottom aligned with the preview
  - [ ] I have increased the size of the preview to 500*360 so that it fits all the info on the right of it. The exact size is not important but it is important that everything is aligned nicely to the preview
- [ ] Move the silent call option from a 3 dot menu to a checkbox with the text "Notify all participants about this call"
- [ ] For the name/log in field, we can condense it 
  - [ ] "Display name" --> "Name" (unless it's a legal requirement ?)
  - [ ] Use a person icon on the left so that it is aligned with the rest of the fields on top
  - [ ] Use a floating label 

image


- [ ] in all the screenshots I have posted the "test speaker" button is not present as we discussed to move it into the same row as the speaker dropdown so that it looks like this:

image

DorraJaouad:
Currently, size is static (width: 440px), it's possible to also make it dynamic by following the natural height (fitting content) of the right column and apply it on the preview.

Antreesy:
Regarding log-in prompt for public pages:

- I prompted it when open a page, and I'm blocked from interacting with chat
- This is one of the only things I have in RightSidebar as a guest
- If I'm not logged to instance where I'm able to do it, I should notice some features are missing on the page

Q: do we need to show it one more time before I explicitly decided to join the call? IMO name input is enough
image

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -882,6 +888,12 @@ export default {
 		background: var(--color-main-background);
 		border-radius: var(--border-radius-element, calc(var(--default-clickable-area) / 2));
 		box-shadow: 0 0 var(--default-grid-baseline) var(--color-box-shadow);
+		z-index: 2;
+	}
+
+	&__content {
+		display: flex;
```
Antreesy: I'd prefer this to fine-tune the width, if we know we have two columns here:
```suggestion
		display: grid;
		grid-template-columns: 7fr 5fr;
```

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -882,23 +920,39 @@ export default {
...
 		box-shadow: 0 0 var(--default-grid-baseline) var(--color-box-shadow);
+		z-index: 2;
 	}
 
-	&__call-buttons {
+	&__content {
 		display: flex;
-		z-index: 1;
+		gap: calc(var(--default-grid-baseline) * 2);
 		align-items: center;
```
Antreesy: I think this would give a better result:
```suggestion
		align-items: flex-start;
```

src/PublicShareAuthSidebar.vue
Antreesy, line 1: `<ChatView />` still renders GuestWelcomeWindow behind MediaSettings (and in front of, if you cleared the input

Also check for `conversation.objectType !== CONVERSATION.OBJECT_TYPE.VIDEO_VERIFICATION`

src/components/MediaSettings/MediaSettings.vue
Antreesy: One more argument =)
image

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -4,227 +4,219 @@
...
+								:device-id="audioOutputId"
+								@refresh="updateDevices"
+								@update:device-id="handleAudioOutputIdChange">
+								<template #extra-action>
+									<MediaDevicesSpeakerTest :disabled="!!audioStreamErrorMessage" />
+								</template>
+							</MediaDevicesSelector>
+						</template>
 
+						<template #tab-panel:backgrounds>
```
Antreesy, line 165: Note: with new custom bgs feature, max- height of this should be locked to 3 lines of buttons (default) and be scrollable
Then we are less prone to breaking layout

Antreesy:
Regarding devices selection:
- 'Refresh button' is eating too much space. We can trigger update automatically every time select is opened, or maybe merge it with icon (so it shows reload icon on hover, and refresh list on click)
- 'Test speakers' - same. Also equalizer animation was causing some issues, but I don't recall what. Maybe keep it simple as pause-play button?
image

src/stores/guestName.js
```diff
@@ -91,7 +91,6 @@ export const useGuestNameStore = defineStore('guestName', {
 				await setGuestUserName(token, name)
 
 				setGuestNickname(name || t('spreed', 'Guest'))
-				emit('talk:guest-name:added')
```
Antreesy, line 94: Should be in one more place, could you check?

DorraJaouad: image

src/components/MediaSettings/MediaSettings.vue
DorraJaouad: I'd prefer 4fr 3fr through

src/PublicShareAuthSidebar.vue
DorraJaouad: Nice catch!

nimishavijay:
> Currently, size is static (width: 440px), it's possible to also make it dynamic by following the natural height (fitting content) of the right column and apply it on the preview.

@DorraJaouad  We definitely don't want it to change height as we switch from the device tab to the background tab, or when we open an NcSelect. It should also have a min-height so that it's not too small. Other than those constraints I am fine with either way :) 

> 'Refresh button' is eating too much space. We can trigger update automatically every time select is opened, 

@Antreesy Agreed! If it's possible to trigger refresh everytime you open the NcSelect then this would be perfect. 

> 'Test speakers' - same. Also equalizer animation was causing some issues, but I don't recall what. Maybe keep it simple as pause-play button?

@Antreesy  Agreed that it is taking up a lot of space but not sure if a play button is the best choice (even though I can't really find anything better, the best I could do is this [graphic-eq icon](https://fonts.google.com/icons?icon.set=Material+Symbols&icon.size=24&icon.color=%23222222&icon.query=equal&selected=Material+Symbols+Outlined:graphic_eq:FILL@0;wght@400;GRAD@0;opsz@24) 🤷). Could we shorten it to just "Test"? Now that we are removing the refresh button we have more space in that row anyway.
As for the animation, I think it is important to show as a confirmation that something is playing. So I would say in this case we show a smaller version of the existing animation (squuezed to fit into a 20px icon slot)

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -4,249 +4,240 @@
...
+									{{ videoStreamErrorMessage }}
+								</p>
 							</template>
-							{{ t('spreed', 'The conversation participants will be notified about this call') }}
-						</NcActionButton>
-					</NcActions>
+						</NcPopover>
+					</div>
+				</div>
+				<div class="media-settings__settings">
```
Antreesy, line 139: I remember we discussed to limit VideoBackground height to three rows, to have it consistent

src/components/LobbyScreen.vue
```diff
@@ -5,8 +5,7 @@
 
 <template>
 	<div class="lobby">
-		<GuestWelcomeWindow v-if="isGuestWithoutDisplayName" :token="token" />
-		<div class="lobby emptycontent">
+		<div class="lobby__header">
```
Antreesy, line 8: Lobby content was vertically centered on screen before

src/components/LobbyScreen.vue
```diff
@@ -31,7 +30,7 @@
 					use-extended-markdown />
 			</div>
 		</div>
-		<SetGuestUsername v-if="currentUserIsGuest" class="guest-info" />
+		<MediaSettings :is-dialog="false" />
```
Antreesy, line 33: Also I think we need to have a fixed width for media preview, or correctly wrap it:

Image

src/components/MediaSettings/MediaDevicesSelector.vue
```diff
@@ -7,11 +7,12 @@
...
-import { computed } from 'vue'
+import { computed, h } from 'vue'
+import NcIconSvgWrapper from '@nextcloud/vue/components/NcIconSvgWrapper'
 import NcSelect from '@nextcloud/vue/components/NcSelect'
-import IconMicrophone from 'vue-material-design-icons/Microphone.vue'
-import IconVideo from 'vue-material-design-icons/Video.vue'
-import IconVolumeHigh from 'vue-material-design-icons/VolumeHigh.vue'
+import IconMicrophoneOutline from 'vue-material-design-icons/MicrophoneOutline.vue'
+import IconVideoOutline from 'vue-material-design-icons/VideoOutline.vue'
+import IconVolume from '../../../img/icon-volume-high-outline.svg?raw'
```
Antreesy:
```suggestion
import IconVolumeHighOutline from '../../../img/icon-volume-high-outline.svg?raw'
```

src/components/MediaSettings/MediaDevicesSpeakerTest.vue
```diff
@@ -123,4 +123,8 @@ export default {
 		#{3 * $i}% { height: math.random(40) + 20%; }
 	}
 }
+
+:deep(.button-vue) {
```
Antreesy: We have `.media-devices-speaker-test-button`

And why would we need that? It doesn't work and it breaks the button styles

src/components/MediaSettings/MediaSettingsTabs.vue
```diff
@@ -67,9 +67,7 @@ function isSelected(tabId: string) {
  * @param tabId - New selected tabId
  */
 function handleTabClick(tabId: string) {
-	if (isActive(tabId)) {
-		isOpen.value = !isOpen.value
-	} else {
+	if (!isActive(tabId)) {
```
Antreesy, line 70: Should we completely drop:
-  `isActive`, 
- `isOpen`,
- `BrowserStorage.getItem('audioInputDevicePreferred')`,
- `BrowserStorage.getItem('videoInputDevicePreferred')`,

if we don't close it anymore?

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -885,18 +920,30 @@ export default {
...
-		background-color: var(--color-main-background);
-		padding: 10px 0 20px;
+	}
+
+	&__settings {
+		max-width: 450px; // 1/2 large modal width
+		min-width: 400px;
+		display: flex;
+		flex-direction: column;
+		justify-content: space-between;
```
Antreesy: `.tabs: { margin-bottom: auto }` is enough

```suggestion
```

src/components/MediaSettings/MediaSettings.vue
Antreesy: so no grid? =(

src/PublicShareAuthSidebar.vue
```diff
@@ -132,8 +132,6 @@ export default {
 				this.actorStore.setCurrentUser(currentUser)
 			} else if (guestNickname) {
 				this.actorStore.setDisplayName(guestNickname)
-			} else {
-				subscribe('talk:guest-name:added', this.showGuestMediaSettings)
```
Antreesy, line 136: right, you removed it in both places. `subscribe` and `unsubscribe` aren't used in this file anymore

src/components/SetGuestUsername.vue
```diff
@@ -6,27 +6,32 @@
...
 		<!-- eslint-disable-next-line vue/no-v-html -->
-		<h3 v-html="displayNameLabel" />
+		<h3 v-if="!compact" v-html="displayNameLabel" />
 
-		<NcButton v-if="!isEditingUsername"
+		<NcButton v-if="!isEditingUsername && !compact"
 			@click="toggleEdit">
 			{{ t('spreed', 'Edit display name') }}
 			<template #icon>
 				<Pencil :size="20" />
```
Antreesy:
```suggestion
				<IconPencilOutline :size="20" />
```

src/components/SetGuestUsername.vue
```diff
@@ -6,27 +6,32 @@
...
-			:placeholder="t('spreed', 'Guest')"
-			class="username-form__input"
-			:show-trailing-button="!!guestUserName"
-			trailing-button-icon="arrowEnd"
-			:trailing-button-label="t('spreed', 'Save name')"
-			@trailing-button-click="updateDisplayName"
-			@keydown.enter="updateDisplayName"
-			@keydown.esc="toggleEdit" />
+		<div v-else class="username-form__display-name">
+			<IconAccountOutline class="username-form__display-name-icon" :size="20" />
```
Antreesy, line 20: NcTextField supports it:
```suggestion
<template #icon>
	<IconAccountOutline class="username-form__display-name-icon" :size="20" />
</template>
```

Image

Follow-up: we can discuss it separately, but I think, this large heading, and toggle button to edit can completely go away, and we keep only NcTextField

Antreesy (commented):
Post-call comments:

- [ ] Lobby is not centered
- [ ] Preview might be a bit smaller
- [ ] Shouldn't pop up, when call started

General design note: switch should be used prioritised over checkbox, where possible

src/components/LobbyScreen.vue
DorraJaouad: if we remove max with for lobby screen, centrality issues are all solved 
image
It is not needed in the first place as we should not be bound by the conventional chat length

src/components/SetGuestUsername.vue
DorraJaouad: Design wise, we want these to be similar and aligned afaik
image

src/components/MediaSettings/MediaSettingsTabs.vue
DorraJaouad: first two no because they are still used inside component but the last two, yeah removable because we no longer hide devices

src/components/SetGuestUsername.vue
Antreesy: Right, my bad

DorraJaouad:
> Preview might be a bit smaller

it is the optimal size as it is dependent on the length of the right column , it is fitting the max content length exactly

src/components/LobbyScreen.vue
```diff
@@ -100,13 +100,11 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-@import '../assets/variables';
 @import '../assets/markdown';
 
 .lobby {
 	display: flex;
 	flex-direction: column;
-	max-width: $messages-list-max-width;
```
Antreesy: That was actually a limitation to room description text, not the lobby component in general

src/utils/webrtc/MediaDevicesManager.js
```diff
@@ -302,9 +302,6 @@ MediaDevicesManager.prototype = {
 				this._preferenceAudioInputList = newAudioInputList
 				BrowserStorage.setItem('audioInputPreferences', JSON.stringify(newAudioInputList))
 			}
-			if (!BrowserStorage.getItem('audioInputDevicePreferred')) {
-				BrowserStorage.setItem('audioInputDevicePreferred', true)
```
Antreesy, line 306: Same deletions for 'videoInputDevicePreferred'
And don't forget migration =P

src/components/MediaSettings/VideoBackgroundEditor.vue
```diff
@@ -292,6 +292,8 @@ export default {
 	grid-template-columns: repeat(4, 1fr);
 	gap: calc(var(--default-grid-baseline) * 2);
 	margin-top: calc(var(--default-grid-baseline) * 2);
+	max-height: calc(var(--default-grid-baseline) * 16 * 3 + var(--default-grid-baseline) * 4);
```
Antreesy: That's some advanced math =P

Can we write it like:
```suggestion
	--background-button-height: calc(var(--default-grid-baseline) * 16);
	max-height: calc(var(--background-button-height) * 3 + var(--default-grid-baseline) * 4);
```

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -885,18 +916,28 @@ export default {
...
-		justify-content: center;
-		gap: var(--default-grid-baseline);
-		position: sticky;
-		bottom: 0;
-		background-color: var(--color-main-background);
-		padding: 10px 0 20px;
+		flex-direction: column;
+		height: var(--column-height);
+
+		&--options {
```
Antreesy: This is now always rendered:

Image

Do we need this wrapper at all? I think just a gap would be enough on `&__settings`

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -885,18 +916,28 @@ export default {
 		background: var(--color-main-background);
 		border-radius: var(--border-radius-element, calc(var(--default-clickable-area) / 2));
 		box-shadow: 0 0 var(--default-grid-baseline) var(--color-box-shadow);
+		z-index: 2;
+	}
+
+	&__content {
+		display: grid;
+		grid-template-columns: 4fr 3fr;
```
Antreesy, line 926: gap between columns:

Image

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -4,249 +4,240 @@
...
+						{{ t('spreed', 'The recording might include your voice, video from camera, and screen share. Your consent is required before joining the call.') }}
+					</p>
+					<NcCheckboxRadioSwitch class="checkbox--warning"
+						:model-value="recordingConsentGiven"
+						@update:model-value="setRecordingConsentGiven">
+						{{ t('spreed', 'Give consent to the recording of this call') }}
+					</NcCheckboxRadioSwitch>
+				</template>
+			</NcNoteCard>
+			<div class="media-settings__content" :style="{ 'flex-direction': isMobile ? 'column' : 'row' }">
```
Antreesy: `:class="{ 'media-settings__content--mobile': isMobile }">`

and then in styles `grid-template-columns: 1fr;` and other fixes. column view feels broken
Maybe we should even change NcDialog size prop when mobile

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -855,19 +887,18 @@ export default {
...
 		position: relative;
-		margin: 0 auto calc(var(--default-grid-baseline) * 4);
 		display: flex;
 		align-items: center;
 		justify-content: center;
 		border-radius: var(--border-radius-element, var(--border-radius-large));
 		background-color: var(--color-loading-dark);
-		width: 100%;
 		aspect-ratio: 4/3;
+		height: var(--column-height);
```
Antreesy: in near-mobile width preview takes the whole space, and horizontal scroll appear

max-height looks enough for both columns, they're adjusting themselves a bit

src/components/MediaSettings/MediaSettings.vue
DorraJaouad: intentional

src/components/MediaSettings/MediaSettings.vue
DorraJaouad: It's meant for space betweem tabs and checkboxes but given that tabs has auto margin, in most cases, it should be fine

src/components/MediaSettings/MediaSettings.vue
DorraJaouad: Height is to fix the preview, if it is not set, video stream will affect the size

src/components/LobbyScreen.vue
```diff
@@ -119,29 +106,24 @@ export default {
 .lobby {
 	display: flex;
 	flex-direction: column;
+	margin: 0 auto;
```
Antreesy: Still aligned to the top:

```suggestion
	margin: auto;
```

Image

src/components/SetGuestUsername.vue
```diff
@@ -125,15 +134,24 @@ function toggleEdit() {
 
 <style lang="scss" scoped>
 .username-form {
-	margin-block-end: 12px;
-	margin-inline: auto;
+
+	&__display-name {
+		display: flex;
+		align-items: center;
```
Antreesy: Not centered:

Image

src/components/SetGuestUsername.vue
```diff
@@ -6,27 +6,32 @@
...
-			@keydown.esc="toggleEdit" />
+		<div v-else class="username-form__display-name">
+			<IconAccountOutline class="username-form__display-name-icon" :size="20" />
+			<NcTextField
+				ref="usernameInput"
+				v-model="guestUserName"
+				:placeholder="t('spreed', 'Guest')"
+				class="username-form__input"
+				:label-outside="!compact"
+				:label="compact ? t('spreed', 'Display name (required)') : undefined"
```
Antreesy: Maybe always show it for consistency? I don't mind having it as a guest in Rightsidebar

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -4,249 +4,240 @@
...
-		:label-id="dialogHeaderId"
-		@close="closeModal">
+	<component :is="isDialog ? 'NcModal' : 'div'"
+		v-if="show"
+		:size="isDialog ? 'large' : undefined"
+		:label-id="isDialog ? dialogHeaderId : undefined"
+		@close="close">
 		<div class="media-settings">
-			<h2 :id="dialogHeaderId" class="media-settings__title nc-dialog-alike-header">
-				{{ t('spreed', 'Media settings') }}
```
Antreesy, line 12: In TopBarMenu there is one last string, which we can change to 'Check devices'

src/components/MediaSettings/MediaSettings.vue
```diff
@@ -4,249 +4,240 @@
 -->
 
 <template>
-	<NcModal v-if="modal"
-		:label-id="dialogHeaderId"
-		@close="closeModal">
+	<component :is="isDialog ? 'NcModal' : 'div'"
+		v-if="show"
+		:size="isDialog ? 'large' : undefined"
```
Antreesy, line 9: Make it smaller on mobile view?
Image

Antreesy (approved):
Nothing blocking (apart from modal size in mobile view)

img/icon-volume-high-outline.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" height="24px" viewBox="0 -960 960 960" width="24px" fill="#FFFFFF"><path d="M560-131v-82q90-26 145-100t55-168q0-94-55-168T560-749v-82q124 28 202 125.5T840-481q0 127-78 224.5T560-131ZM120-360v-240h160l200-200v640L280-360H120Zm440 40v-322q47 22 73.5 66t26.5 96q0 51-26.5 94.5T560-320ZM400-606l-86 86H200v80h114l86 86v-252ZM300-480Z"/></svg>
```
Antreesy: Material Icon, not Symbol
```html
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" height="24" width="24" version="1.1">
	<path fill="#fff"
		  d="M3 9v6h4l5 5V4L7 9H3zm7-.17v6.34L7.83 13H5v-2h2.83L10 8.83zM16.5 12c0-1.77-1.02-3.29-2.5-4.03v8.05c1.48-.73 2.5-2.25 2.5-4.02zM14 3.23v2.06c2.89.86 5 3.54 5 6.71s-2.11 5.85-5 6.71v2.06c4.01-.91 7-4.49 7-8.77 0-4.28-2.99-7.86-7-8.77z"/>
</svg>
```

Antreesy (approved):

PR #15438, opened 2025-07-01
fix(Quote): make component compact 

Antreesy:
### ☑️ Resolves

* Reduce component height
* Move 'edit' hint
* Replace file preview with mime icon

🏚️ Before | 🏡 After
-- | --
image | Screenshot after
a | f
a | f
a | f
a | f

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


nimishavijay (changes requested):
Looks reaaally nice! :)
Only change would be the editing indicator, it seems to get lost as it is so far away from the position of the cursor in the beginning. so I would propose the following changes:

- [x] while editing, replace the avatar with the pencil icon 
- [x] Make the `(editing)` non-italic and regular font size
- [ ] Make the right padding the same as for regular replies (looks like it's a bit less now)

Antreesy:
B | A
-- | --
image | image
image | image

> Make the right padding the same as for regular replies (looks like it's a bit less now)

Is it about this padding, on top of input field?
B | A
-- | --
image | image

nickvergessen:
When an image is quoted, it was quite useful to see the mini preview. Is there a chance to improve that one corner case?

DorraJaouad:
OR we keep the icon clickable to open preview as the space is not fitting a visible size.

Antreesy:
OR we ~~do a step back~~ increase a height for fileshare quotes just a bit to fit 32px preview:

src/components/Quote.vue
```diff
@@ -113,10 +90,22 @@ const shortenedQuoteMessage = computed(() => {
 	return simpleQuotedMessage.value.length >= 250 ? simpleQuotedMessage.value.substring(0, 250) + '…' : simpleQuotedMessage.value
 })
 
+/**
+ * Check whether message to quote (parent) existing on server
+ * Otherwise server returns ['id' => (int)$parentId, 'deleted' => true]
+ */
+function isExistingMessage(message: ChatMessage | DeletedParentMessage): message is ChatMessage {
+	return 'messageType' in message
```
DorraJaouad, line 118: what about `!('deleted' in message)`  ?

src/components/Quote.vue
```diff
@@ -191,70 +188,72 @@ function handleQuoteClick() {
...
+	&__avatar {
+		flex-shrink: 0;
+	}
+
+	&__author-info {
+		flex-shrink: 0;
+		font-weight: 600;
+		white-space: nowrap;
+		overflow: hidden;
+		text-overflow: ellipsis;
```
DorraJaouad: It won't work here

src/components/Quote.vue
Antreesy: `deleted: true` is present for deleted messages, that are still in database, thus visible in UI.
We need anything from BaseChatMessage

src/components/Quote.vue
Antreesy: Limited a width for author info

Antreesy:
Some mockups:

One-line replacing icon | Compact two-line
-- | --
image | image

Not sure what to do with mimeicons and simple replies then, don't they look too different?

Shall we try to keep two-line layout (like on the left) and use 13px font-size?

nickvergessen:
I prefer the right side

nimishavijay:
> Is it about this padding, on top of input field?

No, but your after screenshot look better either way. I mean this space on the right of _(editing)_ is too less

image

> Not sure what to do with mimeicons and simple replies then, don't they look too different?

- [x] I prefer the right side too. They do look different but I think for simple messages it is okay. How about we use the 2-line one for photos and mimeicons as well? That way it would be consistent as one UI for simple messages and one UI for any message with an attachment. 
- [x] UPD (from chat): Just an idea: we can remove the avatar for messages with attachments so that everything is aligned (copied directly from telegram)

DorraJaouad (approved):
Looks good, tested

DorraJaouad (commented):
unused type Component

DorraJaouad:
hmm should we set one height for all? plain messages are shorter than all others

Antreesy:
> hmm should we set one height for all? 

2-line with no preview and user avatar looks misaligned, when in two strings. 

> plain messages are shorter than all others

Plain messages also occur more often.

Also change is a matter of moving 3 css lines. Let's see first if we can get used to this.

DorraJaouad:
> > hmm should we set one height for all? 
> 
> 
> 
> 2-line with no preview and user avatar looks misaligned, when in two strings. 
> 
> 
> 
> > plain messages are shorter than all others
> 
> 
> 
> Plain messages also occur more often.
> 
> 
> 
> Also change is a matter of moving 3 css lines. Let's see first if we can get used to this.

Yeah, I am fine with going as-is for now.

PR #15446, opened 2025-07-02
fix(threads): Mark thread attendees as mentioned

nickvergessen:
### 🚧 Tasks

- [x] Mark mentioned attendees as mentioned, ref #9679 

### 🏁 Checklist


lib/Chat/ChatManager.php
```diff
@@ -450,9 +450,11 @@ public function sendMessage(
 			if (!empty($alreadyNotifiedUsers)) {
 				$userIds = array_column($alreadyNotifiedUsers, 'id');
 				$this->participantService->markUsersAsMentioned($chat, Attendee::ACTOR_USERS, $userIds, (int)$comment->getId(), $usersDirectlyMentioned);
+				$this->threadService->markAttendeesAsMentioned($chat, Attendee::ACTOR_USERS, $userIds, (int)$comment->getId(), $usersDirectlyMentioned);
```
miaulalala: I'm thinking this might need a check if a message is part of a thread or not, but I could be wrong.

nickvergessen: Right, fixed

PR #15450, opened 2025-07-03
fix(safari): provide fallback for 'window.requestIdleCallback'

Antreesy:
### ☑️ Resolves

https://caniuse.com/?search=requestidlecallback
* Fix regression introduced for Safari
  * also refactor old migration steps

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/init.js
+++ b/src/init.js
@@ -89,17 +89,18 @@ const migrateDirectLocalStorageToNextcloudBrowserStorage = () => {
 		return
 	}
 
-	const storageKeys = Array.from(Array(localStorage.length), (_, i) => localStorage.key(i)).filter((key) => key.startsWith('audioDisabled_')
-		|| key.startsWith('videoDisabled_')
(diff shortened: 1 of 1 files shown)


src/init.js
```diff
@@ -137,7 +138,13 @@ function cleanOutdatedBrowserStorageKeys() {
...
 }
 
-window.requestIdleCallback(() => {
+if (window.requestIdleCallback) {
+	window.requestIdleCallback(() => {
+		migrateDirectLocalStorageToNextcloudBrowserStorage()
+		cleanOutdatedBrowserStorageKeys()
+	})
+} else {
+	// Fallback for Safari
```
ShGKme, line 147: `setTimeout()` could be a fallback to not block the process during init.

Alternative - store migration version in the LS and check the version.

PR #15460, opened 2025-07-03
fix: remove `CheckTurnCertificate` from background jobs as the class does not exist anymore

szaimen:
Fix the following error message:

```json
[cron] Warning: failed to create instance of background job: OCA\Talk\BackgroundJob\CheckTurnCertificate
	from ? by -- at Jul 3, 2025, 3:15:32 PM
```


SystemKeeper:
Hey, thanks for noticing! I was a bit surprised to see the PR as I was not aware we removed it... but I fear we reference the wrong class and it should actually be https://github.com/nextcloud/spreed/blob/main/lib/BackgroundJob/CheckCertificates.php. Seems I missed the rename in https://github.com/nextcloud/spreed/commit/152ba97d909b2822dffb7fc21b57066c1f3b7d80 🙈 

Wanna adjust your PR? :)

szaimen:
> Hey, thanks for noticing! I was a bit surprised to see the PR as I was not aware we removed it... but I fear we reference the wrong class and it should actually be https://github.com/nextcloud/spreed/blob/main/lib/BackgroundJob/CheckCertificates.php. Seems I missed the rename in [152ba97](https://github.com/nextcloud/spreed/commit/152ba97d909b2822dffb7fc21b57066c1f3b7d80) 🙈

I see

> Wanna adjust your PR? :)

done :)

PR #15481, opened 2025-07-07
fix(setupchecks): Log the error from setupchecks to have more details

nickvergessen:
### 🏁 Checklist


nickvergessen:
Was added in 31 only

Issue #15465, opened 2025-07-04
🔴 "Busy" status and icon refresh for all status


miaulalala:
We're introducing a new status "Busy" that will be integrated with the existing automations and will change some default behaviour.

Changes to automated behaviour:

| User is | Old | New |
|--------|--------|--------|
| In a call | Away | Busy |
| In a meeting / Calendar busy | Away | Busy |

It will be added to the list of choosable status which currently includes "Online", "Away", "Do Not Disturb" and "Invisible".

Additionally, we will offer a new predefined status "Be right back" that will be added to the list of predefined status and should come with a fitting emoji and a default time until reset of 15 minutes.

To Do:

Design @nimishavijay 
- New icons for all status including busy
- Find fitting emoji for new "be right back" predefined status

Backend
- Adjust status automations for "Busy" status
- Add "Be right back" predefined status option
- Add capability for new "Busy" status

Frontend / Clients
- Support new "Busy" status
- Support selecting new "Busy" status
- Replace status icons
- Ensure new "Be right back" status is shown in predefined list

Possible follow ups:

- Change the OOO automation to allow choosing between "Busy" and "Do Not Disturb" status
- Allow user to turn off Calendar automation for "In a meeting"
    - Optionally, allow choosing between "only show me as busy" or "show me as busy and in a meeting"
- Allow to set "Do Not Disturb" status when screensharing in a call to mute notifications

nimishavijay:
Great idea! Some questions from my side:

> New icons for all status including busy

Is there a particular reason behind this or just for a general visual refresh?

> In a call --> Busy 
In a meeting --> Busy 

What do you think of unifying these 2 in the UI as just one "Busy" status message and change the emoji depending on if they are calendar-busy or call-busy? Even though there is a logic behind how it works rn, it could _seem_ pretty arbitrary when someone is in a meeting vs in a call. If we show everything as just "Busy" we don't bring so much prominence to the difference (while still indicating it in the UI subtly), and also avoiding the "How do we know if they are in a meeting or just at the doctors" question.

nickvergessen:
> Is there a particular reason behind this or just for a general visual refresh?

[...]

PR #15484, opened 2025-07-07
feat(userstatus): use busy status for calls

miaulalala:
### ☑️ Resolves

Part of https://github.com/nextcloud/spreed/issues/15465

The "busy" status has existed in the user_status app since NC 28, so I don't think we need to worry about checking if it exists.

---

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/lib/Status/Listener.php
+++ b/lib/Status/Listener.php
@@ -62,7 +62,7 @@ class Listener implements IEventListener {
 
 	protected function setUserStatus(BeforeParticipantModifiedEvent $event): void {
 
-		$status = IUserStatus::AWAY;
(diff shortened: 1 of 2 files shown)


docs/capabilities.md
```diff
@@ -196,3 +196,4 @@
 
 ## 22
 * `threads` - Whether the chat supports threads
+* `busy-status` - Whether the "Busy" user status is settable manually
```
nickvergessen: belongs into server > user_status app/capabilities?

Antreesy (approved):
Image

PR #15494, opened 2025-07-10
feat(threads): Recent threads, no more read marker and add conversation token

nickvergessen:
- [x] Ref #9679 
- [x] We can not find a suitable way for a read marker logic so for now we hide it just like teams, slack, telegram and all others
- [x] Track last activity of threads and change the list to recent threads
- [x] Always return a thread attendee and the room token

### 🏁 Checklist

### DIFF
--- a/lib/Controller/ThreadController.php
+++ b/lib/Controller/ThreadController.php
@@ -130,6 +130,9 @@ class ThreadController extends AEnvironmentAwareOCSController {
 		foreach ($threads as $thread) {
 			$firstMessage = $lastMessage = null;
 			$attendee = $attendees[$thread->getId()] ?? null;
+			if ($attendee === null) {
+				$attendee = ThreadAttendee::createFromParticipant($thread->getId(), $this->participant);
+			}
 
 			$first = $comments[$thread->getId()] ?? null;
 			if ($first !== null) {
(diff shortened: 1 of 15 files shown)


lib/Controller/ThreadController.php
```diff
@@ -51,24 +52,22 @@ public function __construct(
...
 	#[RequireModeratorOrNoLobby]
 	#[RequireParticipant]
-	#[ApiRoute(verb: 'GET', url: '/api/{apiVersion}/chat/{token}/threads', requirements: [
+	#[ApiRoute(verb: 'GET', url: '/api/{apiVersion}/chat/{token}/threads/recent', requirements: [
 		'apiVersion' => '(v1)',
 		'token' => '[a-z0-9]{4,30}',
 	])]
-	public function listThreads(int $limit = 25, int $offsetId = 0): DataResponse {
-		$threads = $this->threadService->findByRoom($this->room, $limit, $offsetId);
+	public function getRecentActiveThreads(int $limit = 50): DataResponse {
```
Antreesy, line 68: Internal Server Error
lastReadMessage is not a valid attribute

nickvergessen: Fixed

Antreesy (approved):
Works

PR #15495, opened 2025-07-10
fix(comments): Stop using deprecated method of comments manager

nickvergessen:
https://github.com/nextcloud/server/blob/9836e9b16484582d309c8437ab46d82e34956941/lib/public/Comments/ICommentsManager.php#L113

Not sure how this slipped 3 years and psalm didn't complain

### 🏁 Checklist


SystemKeeper (approved):
Good question, but nice

PR #15496, opened 2025-07-11
feat(threads) add UI for threads feature

Antreesy:
### ☑️ Resolves

* Ref #9680

Threads overview in RightSidebar > Shared items ⚠️ 

Short list with actions | Full scrollable list (no search yet)
-- | --
image | image
image | --

Chat view
Default | Thread view
-- | --
image | image
image | image

### 🚧 Tasks
- [ ] UI
  - [ ] Details tab
  - [ ] Load of messages, if missing
  - [ ] Files Sidebar threads support
  - [ ] Tests

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/MessagesList/MessagesList.vue
+++ b/src/components/MessagesList/MessagesList.vue
@@ -15,6 +15,16 @@
 		<TransitionWrapper name="fade">
 			<div class="scroller__loading">
 				<NcLoadingIcon v-if="displayMessagesLoader" class="scroller__loading-element" :size="32" />
+				<!-- FIXME return from threaded view during the call -->
+				<NcButton
+					v-else-if="threadId && isInCall"
+					:title="t('spreed', 'Back')"
+					:aria-label="t('spreed', 'Back')"
+					@click="threadId = 0">
+					<template #icon>
+						<IconArrowLeft :size="20" />
+					</template>
+				</NcButton>
 			</div>
 		</TransitionWrapper>
 
@@ -63,8 +73,10 @@ import moment from '@nextcloud/moment'
 import debounce from 'debounce'
 import uniqueId from 'lodash/uniqueId.js'
 import { computed } from 'vue'
+import NcButton from '@nextcloud/vue/components/NcButton'
 import NcEmptyContent from '@nextcloud/vue/components/NcEmptyContent'
 import NcLoadingIcon from '@nextcloud/vue/components/NcLoadingIcon'
+import IconArrowLeft from 'vue-material-design-icons/ArrowLeft.vue'
 import Message from 'vue-material-design-icons/Message.vue'
 import LoadingPlaceholder from '../UIShared/LoadingPlaceholder.vue'
 import TransitionWrapper from '../UIShared/TransitionWrapper.vue'
@@ -84,8 +96,10 @@ const SCROLL_TOLERANCE = 10
 export default {
 	name: 'MessagesList',
 	components: {
+		IconArrowLeft,
 		LoadingPlaceholder,
 		Message,
+		NcButton,
 		NcEmptyContent,
 		NcLoadingIcon,
 		TransitionWrapper,
@@ -675,7 +689,13 @@ export default {
 				if (!this.$store.getters.getFirstKnownMessageId(token)) {
 					try {
 						// Start from message hash or unread marker
-						const startingMessageId = focusMessageId !== null ? focusMessageId : this.conversation.lastReadMessage
+						let startingMessageId = focusMessageId !== null ? focusMessageId : this.conversation.lastReadMessage
+						// Check if thread is initially opened
+						if (this.threadId) {
+							// FIXME temporary get thread messages from the start
+							startingMessageId = this.threadId
+						}
+
 						// First time load, initialize important properties
 						if (!startingMessageId) {
 							throw new Error(`[DEBUG] spreed: context message ID is ${startingMessageId}`)
@@ -1221,10 +1241,9 @@ export default {
 
 		async onRouteChange({ from, to }) {
 			if (from.name === 'conversation' && to.name === 'conversation'
-				&& from.params.token === to.params.token
-				&& from.hash !== to.hash) {
-				// the hash changed, need to focus/highlight another message
-				if (to.hash && to.hash.startsWith('#message_')) {
+				&& from.params.token === to.params.token) {
+				if (to.hash && to.hash.startsWith('#message_') && from.hash !== to.hash) {
+					// the hash changed, need to focus/highlight another message
 					const focusedId = this.getMessageIdFromHash(to.hash)
 					if (this.messagesList.find((m) => m.id === focusedId)) {
 						// need some delay (next tick is too short) to be able to run
@@ -1246,6 +1265,21 @@ export default {
 						})
 						await this.getMessageContext(this.token, focusedId)
 						this.focusMessage(focusedId, true)
+					}
+				} else if (to.query.threadId && from.query.threadId !== to.query.threadId) {
+					// FIXME temporary get thread messages from the start
+					const topMostThreadMessage = this.$store.getters.messagesList(to.params.token).find((message) => message.id === +to.query.threadId)
+					if (!topMostThreadMessage) {
+						// Update environment around context to fill the gaps
+						this.$store.dispatch('setFirstKnownMessageId', {
+							token: this.token,
+							id: +to.query.threadId,
+						})
+						this.$store.dispatch('setLastKnownMessageId', {
+							token: this.token,
+							id: +to.query.threadId,
+						})
+						await this.getMessageContext(this.token, +to.query.threadId)
 					}
 				}
 			}


Antreesy:
<details>
<summary>Details from design team / previous PR discussion</summary>

What do you think? Shows more info than before. Using a smaller font size in the first line brings more attention to the latest reply while still providing context. 

It however involves a lot of info: original poster, original poster's avatar, original message, latest poster's name, possibly latest poster's avatar, and latest message. Is requesting all that info possible and would it unreasonably affect the performance? 

image

With avatar: Not super keen on this but tried it anyway

image

Maybe not a topic for right now but I did think about if we should also toggle the view when compact mode is enabled, if so it would look something like this:

image



Other feedback from the screenshots:
- [ ] "Make thread" --> "Start a thread"
- [ ] "See thread" --> "Go to thread" (as it literally feels like you're going to a different conversation) or "Read thread"
- [ ] I see that all the threads are listed in the sidebar outside of the tabs. Is that an intermediate solution before we do either of the solutions in `https://github.com/nextcloud/spreed/issues/15313#issuecomment-3019190538` or are those out of scope for now?

> "Make thread" --> "Start a thread"

Should be "Create thread" or "Create a thread" the first message could be older and from someone else, so are you really "starting" it?

You mean, 'Start a thread' for new ones:
image

And 'Create a thread' for already posted:
image
- Remove last name of last poster (same as left conversation list)
- add a timestamp 
- unread replies not possible, but total number of replies is possible, so think about how to show unread indicator (like a dot) + total number of replies without making it look like unread counter 
image
image


</details>

src/components/RightSidebar/RightSidebarContent.vue
```diff
@@ -259,15 +275,33 @@ function handleHeaderClick() {
...
+						:thread="thread" />
+				</ul>
+				<NcButton
+					variant="tertiary"
+					wide
+					@click="emit('update:state', 'threads')">
+					<template #icon>
+						<IconDotsHorizontal :size="20" />
+					</template>
+					{{ t('spreed', 'Show all threads') }}
```
nickvergessen: It's not all, but "more"

src/components/RightSidebar/Threads/ThreadItem.vue
```diff
@@ -0,0 +1,160 @@
...
+
+const timeFormat = computed<Intl.DateTimeFormatOptions>(() => {
+	if (new Date().toDateString() === new Date(lastActivity.value).toDateString()) {
+		return { timeStyle: 'short' }
+	}
+	return { dateStyle: 'short' }
+})
+
+/**
+ * FIXME copied from conversation item/composable/quote component, should be shared from utils
```
DorraJaouad, line 70: You are right, we should extract `simpleLastChatMessage` logic from composable and move it to util and use it in all needed spots. We keep follow-up

src/composables/useGetThreadId.ts
```diff
@@ -0,0 +1,27 @@
...
+		return computed<number>({
+			get: () => route.query.threadId ? Number(route.query.threadId) : 0,
+			set: (value: number) => {
+				router.push({ query: { ...route.query, threadId: value !== 0 ? value : undefined } })
+			},
+		})
+	} else {
+		return ref(0)
+	}
+})
```
DorraJaouad, line 27: I don't think it needs to be shared

src/components/MessagesList/MessagesList.vue
```diff
@@ -188,7 +192,14 @@ export default {
...
 		messagesList() {
+			if (!this.threadId) {
+				return this.$store.getters.messagesList(this.token)
+			}
+
 			return this.$store.getters.messagesList(this.token)
+				.filter((message) => {
+					return message.threadId === this.threadId
+				})
 		},
```
DorraJaouad, line 217: Not related to this change of code but how to get out of this thread view :'). I know that the conversation icon will do but I mean it is not obvious from user perspective

DorraJaouad: Image

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
```diff
@@ -139,6 +139,28 @@
...
+
+						<NcActionButton
+							v-else
+							close-after-click
+							@click="chatExtrasStore.createThread(message.token, message.threadId)">
+							<template #icon>
+								<IconForumOutline :size="16" />
+							</template>
+							{{ t('spreed', 'Create a thread') }}
+						</NcActionButton>
```
DorraJaouad: It shouldn't be like this, but like it was in your previous PR. this just creates an empty thread. Threads gets automatically created by the first one who replies as "thread". So the button should be next to reply button

src/components/RightSidebar/RightSidebarContent.vue
```diff
@@ -259,15 +275,33 @@ function handleHeaderClick() {
 						:color="event.color" />
 				</ul>
 			</div>
+			<div v-if="supportThreads && threadsInformation.length"
+				class="content__threads">
+				<NcAppNavigationCaption :name="t('spreed', 'Recent threads')" />
+				<ul class="content__threads-list">
+					<ThreadItem v-for="thread of threadsInformation"
+						:key="`thread_${thread.thread.id}`"
+						:thread="thread" />
```
DorraJaouad: We need to add patch logic to update this list when a reply is added or changed to update the last activity and the numReplies. :(

src/components/RightSidebar/RightSidebarContent.vue
```diff
@@ -259,15 +275,33 @@ function handleHeaderClick() {
 						:color="event.color" />
 				</ul>
 			</div>
+			<div v-if="supportThreads && threadsInformation.length"
+				class="content__threads">
+				<NcAppNavigationCaption :name="t('spreed', 'Recent threads')" />
```
DorraJaouad: Image

It doesn't look right :/

src/composables/useGetThreadId.ts
Antreesy: I think it should:
- with router, we're changing and computing it once for all components
- in sidebar, this is a must, so all components pointing to the same ref

src/components/MessagesList/MessagesList.vue
Antreesy: Maybe a close button next to "subscribe"?

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
Antreesy: We discussed, that it's not user-friendly to have two replies. Will align with design again

src/components/RightSidebar/RightSidebarContent.vue
Antreesy: wip

src/composables/useGetThreadId.ts
DorraJaouad: What should be in sync among all components is the get part which it is already because router is a global state.

src/components/MessagesList/MessagesList.vue
DorraJaouad: Yes, or it seems more like `back` action

src/composables/useGetThreadId.ts
Antreesy: You can't do only a 'get' part shared, and it doesn't make sense. Shared composable is reliable thing here, as we need a single source of truth for all components using it

src/components/MessagesList/MessagesList.vue
nimishavijay: > Not related to this change of code but how to get out of this thread view :'). I know that the conversation icon will do but I mean it is not obvious from user perspective

Original mockups had a back icon on the left of the main conversation button. 

image

Implementation-wise they should be one button with 2 icons, arrow-left and main conversation avatar. (Hover feedback looks ugly with the 2 different border radiuses but it is what it is ¯\_(ツ)_/¯ )
image


Edit: nice-to-have would be an animation that moves the main conversation avatar to the right when to enter a thread

https://github.com/user-attachments/assets/63fff51f-b05f-4e75-be50-f07b1e5262d4

src/components/RightSidebar/RightSidebarContent.vue
nimishavijay: - [x] This would be solved by moving the thread into the shared items tab and renaming the shared items to "Details". 
- [ ] Either way we should move the "Disable lobby" button and "Import email participants" button into the participants tab, where they make the most sense.
- [ ] The disable lobby can be a wide button right below search for participants 
  - [ ] during a call when chat is open this may be hidden under the participants tab but I don't think that will too much of a problem as people will likely see it in the participants tab before they start a call. If really required we can have an icon-only button in the top bar, if we change to bottom controls it leaves a lot of room in the top bar
- [ ] the import button can be an icon-only button on the right of the search field (if sip dial out is enabled, between the search field and the dial button)

nimishavijay:
If it is not possible to show unread markers I would suggest something like 

image


- [x] little reply chip with 16px [reply](https://fonts.google.com/icons?icon.set=Material+Symbols&icon.size=24&icon.color=%23222222&selected=Material+Symbols+Outlined:reply:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=reply) icon and number of replies in `--font-size-small` (i would refrain from going below this font size for a11y reasons, I think we're already pushing it if we use that with maxcontrast) and bg `--primary-element-light`
  - [x] when the thread is selected we can remove the primary-light bg 
image

- [x] show the timestamp in `--font-size-small`

src/components/RightSidebar/RightSidebarContent.vue
Antreesy: It's a conversation information, and should be it's own section (see e.g. mockup):

image

@nimishavijay could you make something nicer of it? And we'll handle it in follow-up

nickvergessen:
> Either way we should move the "Disable lobby" button and "Import email participants" button into the participants tab, where they make the most sense.

This would hide the "Disable lobby" button into a non-active tab, when you started the call already. But maybe when you call while the lobby is active, we should instead show a placeholder participant in the call view? Anyway that sounds like something for later.

src/components/RightSidebar/Threads/ThreadsTab.vue
```diff
@@ -0,0 +1,40 @@
...
+import { useChatExtrasStore } from '../../../stores/chatExtras.ts'
+
+const chatExtrasStore = useChatExtrasStore()
+
+const token = useGetToken()
+const threadsInformation = computed(() => chatExtrasStore.getThreadsList(token.value))
+</script>
+
+<template>
+	<div class="threads-tab">
```
DorraJaouad: Remove the div as it is not necessary?

src/components/Quote.vue
```diff
@@ -50,7 +52,15 @@ const actorInfo = computed(() => [actorDisplayNameWithFallback.value, remoteServ
...
-const component = computed(() => canCancel ? { tag: 'div', link: undefined } : { tag: 'router-link', link: { hash: hash.value } })
+const component = computed(() => (canCancel || !isExistingMessage(message))
+	? { tag: 'div', link: undefined }
+	: {
+			tag: 'router-link',
+			link: {
+				query: { threadId: (threadId.value === message.threadId) ? message.threadId : undefined },
+				hash: hash.value,
+			},
+		})
```
DorraJaouad:
```suggestion
	? { tag: 'div', link: undefined }
	: {
		tag: 'router-link',
		link: {
			query: { threadId: (threadId.value === message.threadId) ? message.threadId : undefined },
			hash: hash.value,
		}
	})
```

src/components/TopBar/TopBar.vue
```diff
@@ -10,17 +10,63 @@
...
-
-		<div class="top-bar__wrapper" :data-theme-dark="isInCall ? true : undefined">
+		<a
+			class="top-bar__icon-wrapper"
+			:class="{ 'top-bar__icon-wrapper--thread': currentThread }"
+			role="button"
+			:title="t('spreed', 'Back')"
+			:aria-label="t('spreed', 'Back')"
+			@click="currentThread ? threadId = 0 : openConversationSettings()">
+			<IconArrowLeft class="top-bar__icon-back" :size="20" />
```
DorraJaouad: bidirectional

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
```diff
@@ -139,6 +139,28 @@
 							{{ t('spreed', 'Download file') }}
 						</NcActionLink>
 					</template>
+					<NcActionSeparator />
+					<template v-if="supportThreads && !threadId">
```
DorraJaouad, line 143: Also if user has right permissions and conversation permissions

src/components/TopBar/TopBar.vue
```diff
@@ -357,12 +447,39 @@ export default {
...
+	}
+
+	.top-bar__icon-back {
+		position: absolute;
+		width: var(--default-clickable-area);
+		height: 100%;
+		top: 0;
+		inset-inline-start: 0;
+	}
+}
```
DorraJaouad: There is no evident assumption on the order of the layers when they are stacked so..

Image

src/components/TopBar/TopBar.vue
```diff
@@ -10,17 +10,63 @@
...
+				:disable-menu="false"
+				show-user-online-status
+				:hide-favorite="false"
+				:hide-call="false" />
+		</a>
+
+		<div
+			v-if="currentThread"
+			class="top-bar__wrapper">
+			<IconChevronRight class="bidirectional-icon" :size="20" />
```
DorraJaouad: Better if it is centered

Image

src/components/TopBar/TopBar.vue
Antreesy: hide it for you

src/components/MessagesList/MessagesGroup/Message/MessageButtonsBar/MessageButtonsBar.vue
Antreesy: not here, but for creation

src/components/TopBar/TopBar.vue
DorraJaouad: Background is still visible

DorraJaouad: image

also

nickvergessen:
Merging to deploy on testing instance

PR #15499, opened 2025-07-11
feat(threads): Support federated conversations

nickvergessen:
### ☑️ Resolves

* Ref #9679 

### 🚧 Tasks

- [x] Support federation
- [ ] Check handling of old servers although the capability should simply not exist there and therefore clients don't promote the threads?

### 🏁 Checklist


nickvergessen:
Need to backport the capbility change and the unit test change

nickvergessen:
API call results in 422 if Host server does not support federation. But since the frontend should hide the thread buttons/options based on capabilities that should be okay.

nickvergessen:
/backport 266d91248405808bb018baefe087d6e5adbae6c5 c8d6f9a6653fe678ed150bc1ca7702932fb353b2 0c2f60c45398368b7dccead6954abd7f0c6c2728 to stable31

Antreesy (approved):

Issue #15473, opened 2025-07-05
🎨 Use outline icons


Antreesy:
**Is your feature request related to a problem? Please describe.**
Update vue-material-design-icons to outline style, where applicable

**Describe the solution you'd like**
See parent commit for tracking script

- [ ] Conversation icons served by server (svg)
- [ ] In-component icons (.vue)
- [ ] ~~mimeicons~~

Antreesy:
Currenty ignored / questioned:
- [x] Poll: doesn't have outline variant
- [x] Volume*: don't have outline variants in vue-mdi library
- [ ] Star (favorite feature): full yellow to make favorite, outline black to unmake?
- [ ] Stop: filled is also fine?
- [x] MenuRight? MenuDown?
- [x] Folder - do not outline if FIles app?
- [x] Creation - do not outline if Assistant app?
- [x] Calendar - do not outline if Calendar app?
- [x] Cog - replace with symbol?

Antreesy:
TODO list (to replace / ignore / discuss):

      5 Found: Account.vue
      1 Found: AccountCancel.vue
      1 Found: AccountCircle.vue
      1 Found: AccountMinus.vue
      3 Found: AccountMultiple.vue
      2 Found: AccountMultiplePlus.vue
      2 Found: AccountOff.vue
      3 Found: AccountPlus.vue
      2 Found: AccountSearch.vue
      9 Found: AlertCircle.vue
      2 Found: AlertOctagon.vue
      2 Found: Bell.vue
      4 Found: BellOff.vue
      3 Found: CalendarBlank.vue
      1 Found: CalendarClock.vue
      1 Found: CalendarRange.vue
      4 Found: Cog.vue
      2 Found: Creation.vue
      1 Found: Crown.vue
     11 Found: Delete.vue
      2 Found: Download.vue
      2 Found: Email.vue
      2 Found: File.vue
      2 Found: FileDownload.vue
      2 Found: FileEdit.vue
      1 Found: FileLock.vue
      4 Found: FileUpload.vue
      1 Found: Filter.vue
      3 Found: Folder.vue
      6 Found: HandBackLeft.vue
      1 Found: Information.vue
      2 Found: Lock.vue
      1 Found: LockOpen.vue
      1 Found: LockOpenVariant.vue
      1 Found: MenuDown.vue
      1 Found: MenuRight.vue
      3 Found: Message.vue
      2 Found: MessageText.vue
      8 Found: Microphone.vue
[...]

PR #15506, opened 2025-07-14
fix: use outline icons in ChatView

Antreesy:
### ☑️ Resolves

* Ref #15473

🏚️ Before | 🏡 After
-- | --
image | image
image | image
image | image
image | image
image | image
image | image
image | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


ShGKme:
I'm unsure with download and upload arrows

nickvergessen:
Same could be said about forward, reply and the edit icons. For now we outline them

Antreesy:
We've got only this alternatives
image image image

ShGKme:
> We've got only this alternatives

Looks much better IMO

ShGKme:
> Same could be said about forward, reply and the edit icons. For now we outline them

I'd keep forward and reply solid as they fall into the exception.

https://fonts.google.com/icons?selected=Material+Icons+Outlined:reply:&icon.query=reply&icon.set=Material+Icons&icon.style=Outlined&icon.size=24&icon.color=%235f6368

Edit is fine IMO

DorraJaouad:
Why not make arrow one line not outlined? sthg like this but curvier
https://pictogrammers.com/library/mdi/icon/arrow-left-top/

ShGKme:
> Why not make arrow one line not outlined? sthg like this but curvier https://pictogrammers.com/library/mdi/icon/arrow-left-top/

I think, this might be a good option, but on the other hand they are different icons and for me feels like a different action in comparison to classic reply/forward

PR #15508, opened 2025-07-14
feat(threads): Allow to subscribe with a different notification level

nickvergessen:
### ☑️ Resolves

* Ref #9679 

### 🚧 Tasks

- [x] Allow to subscribe with a different notification level to threads
    - Value is not used for notifications yet, will be a follow up
    - Integration test will follow when that is implemented
- [x] Improve performance by dropping a SELECT query every time we reply to a thread

### 🏁 Checklist

### DIFF
--- a/lib/Chat/Notifier.php
+++ b/lib/Chat/Notifier.php
@@ -12,9 +12,11 @@ use OCA\Talk\Exceptions\ParticipantNotFoundException;
 use OCA\Talk\Files\Util;
 use OCA\Talk\Model\Attendee;
 use OCA\Talk\Model\Session;
+use OCA\Talk\Model\ThreadAttendee;
 use OCA\Talk\Participant;
 use OCA\Talk\Room;
 use OCA\Talk\Service\ParticipantService;
+use OCA\Talk\Service\ThreadService;
 use OCA\Talk\Webinary;
 use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\Comments\IComment;
@@ -42,6 +44,7 @@ class Notifier {
 		private IUserManager $userManager,
 		private IGroupManager $groupManager,
 		private ParticipantService $participantService,
+		private ThreadService $threadService,
 		private IConfig $config,
 		private ITimeFactory $timeFactory,
 		private Util $util,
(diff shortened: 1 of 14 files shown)


Antreesy:
That's a replacement for `POST chat/{token}/threads/{messageId}`? I see it also covers creation, so don't see how we might use the old endpoint

nickvergessen:
Well exactly my question all the time. Then we can add the level as a parameter to the old one.

Antreesy:
Yes,
```
curl -X 'POST' \
  'https://nextcloud.local/ocs/v2.php/apps/spreed/api/v1/chat/q3t48me6/threads/565/notify' \
  -H 'accept: application/json' \
  -H 'OCS-APIRequest: true' \
  -H 'Content-Type: application/json' \
  -d '{
  "level": 0
}'
```

image

But

```
curl -X 'GET' \
  'https://nextcloud.local/ocs/v2.php/apps/spreed/api/v1/chat/q3t48me6/threads/recent?limit=1' \
  -H 'accept: application/json' \
  -H 'OCS-APIRequest: true'
```
image

lib/Federation/Proxy/TalkV1/Controller/ThreadController.php
```diff
@@ -142,4 +142,49 @@ public function makeThread(Room $room, Participant $participant, int $messageId)
...
+				]);
+			}
+			return new DataResponse($data, $statusCode);
+		}
+
+		/** @var TalkThreadInfo $data */
+		$data = $this->proxy->getOCSData($proxy);
+		$data = $this->userConverter->convertThreadInfo($room, $data);
+
+		return new DataResponse($data, Http::STATUS_OK);
```
miaulalala, line 188: **nit**

```suggestion
		$statusCode = $proxy->getStatusCode();
		if ($statusCode === Http::STATUS_OK) {
			/** @var TalkThreadInfo $data */
			$data = $this->proxy->getOCSData($proxy);
			$data = $this->userConverter->convertThreadInfo($room, $data);

			return new DataResponse($data, Http::STATUS_OK);
		}

		if (!in_array($statusCode, [
			Http::STATUS_BAD_REQUEST,
			Http::STATUS_NOT_FOUND,
		], true)) {
			$statusCode = $this->proxy->logUnexpectedStatusCode(__METHOD__, $statusCode);
			$data = ['error' => 'status'];
		} else {
			/** @var array{error: 'level'|'message'|'top-most'} $data */
			$data = $this->proxy->getOCSData($proxy, [
				Http::STATUS_BAD_REQUEST,
				Http::STATUS_NOT_FOUND,
			]);
		}
		return new DataResponse($data, $statusCode);

```

lib/Chat/Notifier.php
```diff
@@ -258,10 +261,39 @@ public function notifyOtherParticipant(Room $chat, IComment $comment, array $alr
...
 		$participants = $this->participantService->getParticipantsByNotificationLevel($chat, Participant::NOTIFY_ALWAYS);
+		$threadId = (int)$comment->getTopmostParentId();
+		/** @var array<int, ThreadAttendee> $threadAttendees */
+		$threadAttendees = [];
+		if ($threadId !== 0) {
+			$threadAttendees = $this->threadService->findAttendeesForNotificationByThreadId($threadId);
+		}
+
+		// Handle participants that only subscribed with Participant::NOTIFY_ALWAYS to the thread, but not the conversation
+		$threadAttendeeIds = array_keys(array_filter($threadAttendees, static fn (ThreadAttendee $threadAttendee): bool => $threadAttendee->getNotificationLevel() === Participant::NOTIFY_ALWAYS));
```
miaulalala: Saves you a call (unless actor Id is not the right value)

```suggestion
		$threadAttendeeIds = array_filter($threadAttendees, static function(ThreadAttendee $threadAttendee) {
				return $threadAttendee->getNotificationLevel() === Participant::NOTIFY_ALWAYS ? $threadAttendee->getActorId() : false;
			});
```

lib/Federation/Proxy/TalkV1/Controller/ThreadController.php
nickvergessen: The code layout is the same for all proxy methods (first error handling which should only be a array{error: …} and then the positive case, so I'd like to keep that :see_no_evil:

lib/Chat/Notifier.php
nickvergessen: $threadAttendee->getId()

but sounds like it should work, thanks for the hint

PR #15519, opened 2025-07-15
fix(threads): add instant update for threads list

DorraJaouad:
### 🚧 Tasks

- [ ] Message expiration
- [ ] Message deletion
- [ ] Message edited
- [ ] Reply added
- [ ] Conversation deletion
- [ ] Thread creation by another user

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/store/messagesStore.js
+++ b/src/store/messagesStore.js
@@ -595,6 +595,10 @@ const actions = {
 					.forEach((storedMessage) => {
 						context.commit('addMessage', { token, message: Object.assign({}, storedMessage, { isThread: true }) })
 					})
+				// Fetch thread data in case it doesn't exist in the store yet
+				if (!chatExtrasStore.getThread(token, message.threadId) && chatExtrasStore.threads[token] !== undefined) {
+					chatExtrasStore.fetchSingleThread(token, message.threadId)
+				}
 			}
 
 			// Quit processing


src/components/MessagesList/MessagesList.vue
```diff
@@ -371,6 +371,7 @@ export default {
 		 */
 		this.expirationInterval = window.setInterval(() => {
 			this.removeExpiredMessagesFromStore()
+			this.chatExtrasStore.fetchRecentThreadsList(this.token)
```
Antreesy: What do you need any update logic for, if you fetch the list every 30 seconds?)

DorraJaouad: oh shit, I didn't mean that :3, It should placed next to `delete state.messages[token][messageId]`, We can also harden the fetch by checking if the deleted message is present in last or first attributes, but not worth it as it will require looping over threads each time

src/store/messagesStore.js
```diff
@@ -486,6 +487,8 @@ const mutations = {
 			if (state.messages[token][messageId].expirationTimestamp
 				&& timestamp > state.messages[token][messageId].expirationTimestamp) {
 				delete state.messages[token][messageId]
+				const chatExtrasStore = useChatExtrasStore()
+				chatExtrasStore.fetchRecentThreadsList(this.token)
```
Antreesy: Calling it on every removed message is even worse =P

Should at least do something like
```js
		const timestamp = convertToUnix(Date.now())
		const messageIds = Object.keys(state.messages[token])
		let atLeastOneThreadWasChanged = false

		messageIds.forEach((messageId) => {
			if (state.messages[token][messageId].expirationTimestamp
				&& timestamp > state.messages[token][messageId].expirationTimestamp) {
				if (state.messages[token][messageId].isThread) {
					atLeastOneThreadWasChanged = true
				}
				delete state.messages[token][messageId]
			}
		})

		if (atLeastOneThreadWasChanged) {
			const chatExtrasStore = useChatExtrasStore()
			chatExtrasStore.fetchRecentThreadsList(token)
		}
```

src/store/messagesStore.js
```diff
@@ -544,6 +547,8 @@ const actions = {
 	processMessage(context, { token, message }) {
 		const sharedItemsStore = useSharedItemsStore()
 		const actorStore = useActorStore()
+		const chatExtrasStore = useChatExtrasStore()
+		const thread = chatExtrasStore.getThread(token, message.parent.threadId)
```
Antreesy: message can have no parent, and it will fail here. Should be fine to call in two times, if:
1. inside 'system messages' block, as invisible system messages have early return
2. after `commit('addMessage')`

src/store/messagesStore.js
```diff
@@ -574,6 +579,23 @@ const actions = {
...
+						thread: {
+							lastActivity: message.parent.timestamp,
+						},
+					}
+					if (thread.last?.id === message.parent.id) {
+						updatedData.last = message.parent
+					}
+					if (thread.first?.id === message.parent.id) {
+						updatedData.first = message.parent
+					}
```
Antreesy: Could as well do this, thread.thread has only primitives, and don't change chatExtrasStore message for it:
```suggestion
					const updatedData = {
						thread: {
							...thread.thread,
							lastActivity: message.parent.timestamp,
						},
						first: (thread.first?.id === message.parent.id) ? message.parent : undefined,
						last: (thread.last?.id === message.parent.id) ? message.parent : undefined,
					}
```

Type `Partial<ThreadInfo>` won't accept `{ thread: Partial<ThreadInfo['thread']>}`, we'll get an error here when we start migrate this store to ts

src/components/Quote.vue
```diff
@@ -117,6 +117,9 @@ function isExistingMessage(message: ChatMessage | DeletedParentMessage): message
 	return 'messageType' in message
 }
 
+/**
+ * Go to thread if it exists
```
Antreesy, line 121: 🙏🏽

src/stores/chatExtras.ts
```diff
@@ -48,7 +48,10 @@ export const useChatExtrasStore = defineStore('chatExtras', {
 	}),
 
 	getters: {
-		getThread: (state) => (token: string, threadId: number) => {
+		getThread: (state) => (token: string, threadId: number | undefined) => {
```
Antreesy: I don't see why it should be a case

src/store/messagesStore.js
```diff
@@ -647,6 +686,22 @@ const actions = {
 
 		context.commit('addMessage', { token, message })
 
+		const thread = chatExtrasStore.getThread(token, message.parent?.threadId)
```
Antreesy: OK I see. Why not use this instead:

```suggestion
		const thread = chatExtrasStore.getThread(token, message.threadId)
```

src/store/messagesStore.js
```diff
@@ -589,6 +623,10 @@ const actions = {
 					.forEach((storedMessage) => {
 						context.commit('addMessage', { token, message: Object.assign({}, storedMessage, { isThread: true }) })
 					})
+				// Fetch thread data in case it doesn't exist in the store yet
+				if (!chatExtrasStore.getThread(token, message.threadId)) {
+					chatExtrasStore.fetchSingleThread(token, message.threadId)
```
Antreesy, line 600: IMO this should happen only when polling new messages. If I'm loading the chat context or history, I don't need this data probably (exception - when you scroll all the way to the first message in the thread, but then it's just to show number of replies)
Otherwise you created 5 threads, then on page reload it fires 5 requests, plus a request for recent threads, when you open a tab.

Alternative, you have all the data here, just put it in the store and don't fetch:
```js
{
    "thread": {
        "id": message.threadId,
        "roomToken": messageToken,
        "lastMessageId": 0,
        "numReplies": 0,
        "lastActivity": message.timestamp
    },
    "attendee": {
        "notificationLevel": 0
    },
    "first": message.parent,
    "last": null,
}
```

src/stores/chatExtras.ts
```diff
@@ -177,6 +188,20 @@ export const useChatExtrasStore = defineStore('chatExtras', {
 			}
 		},
 
+		clearThreads(token: string, messageId?: number) {
+			if (messageId) {
+				// Clear threads that are older than the given messageId
+				for (const threadId of Object.keys(this.threads[token] || {})) {
+					if (this.threads[token][threadId].thread.lastMessageId < messageId) {
```
Antreesy: You can't have this:
1. message in thread A
2. thread_created
...
3. history_cleared
4. message in thread A

Just remove if threadId < messageId

src/stores/chatExtras.ts
```diff
@@ -166,8 +169,16 @@ export const useChatExtrasStore = defineStore('chatExtras', {
...
+					id: payload.thread?.id ?? this.threads[token][threadId].thread.id,
+					roomToken: payload.thread?.roomToken ?? this.threads[token][threadId].thread.roomToken,
+					lastActivity: payload.thread?.lastActivity ?? this.threads[token][threadId].thread.lastActivity,
+					numReplies: payload.thread?.numReplies ?? this.threads[token][threadId].thread.numReplies,
+					lastMessageId: payload.thread?.lastMessageId ?? this.threads[token][threadId].thread.lastMessageId,
+				}
+
 				this.threads[token][threadId] = {
-					thread: payload.thread ?? this.threads[token][threadId].thread,
+					thread: updatedThreadAttributes,
```
Antreesy: revert?

src/store/messagesStore.js
```diff
@@ -647,6 +686,22 @@ const actions = {
 
 		context.commit('addMessage', { token, message })
 
+		const thread = chatExtrasStore.getThread(token, message.parent?.threadId)
+		// Update threads
+		if (message.isThread && thread) {
+			const updateNumReplies = thread.last?.id < message.id
```
Antreesy: Image

Should probably compare with thread.lastMessageId, since you're updating it. Same value, but it's always a number

src/store/messagesStore.js
```diff
@@ -482,12 +490,22 @@ const mutations = {
...
+				if (state.messages[token][messageId].isThread) {
+					atLeastOneThreadWasChanged = true
+				}
 				delete state.messages[token][messageId]
 			}
 		})
+
+		if (atLeastOneThreadWasChanged) {
+			const chatExtrasStore = useChatExtrasStore()
+			chatExtrasStore.fetchRecentThreadsList(token)
```
Antreesy: You tricked me, this should not happen in store.mutations. It's async.
You can instead try to `return atLeastOneThreadWasChanged` here and work with it in dispatch('removeExpiredMessages')

src/store/messagesStore.js
```diff
@@ -298,6 +299,13 @@ const mutations = {
 		const preparedMessage = !message.parent && storedMessage?.parent
 			? { ...message, parent: storedMessage.parent }
 			: message
+
+		if (preparedMessage.parent && preparedMessage.isThread !== preparedMessage.parent.isThread) {
+			preparedMessage.parent = {
+				...preparedMessage.parent,
+				isThread: preparedMessage.isThread,
+			}
+		}
```
Antreesy: This also works
```suggestion
		if (preparedMessage.parent) {
			preparedMessage.parent.isThread = preparedMessage.isThread
		}
```

src/store/messagesStore.js
Antreesy: Also here's sometimes a race condition between 'pollNewMessages' and 'postNewMessage', and it bumped to +2. Not sure if it's worth to deeply look into

src/store/messagesStore.js
DorraJaouad: > Otherwise you created 5 threads, then on page reload it fires 5 requests, plus a request for recent threads, when you open a tab.

It should not happen bc thread is in store?

Antreesy: It's happening now, because you didn't open shared items tab yet
Worst case - both 5+1 will be fired in parallel

> thread is in the store

I seriously doubt that it's there after page reload 😁

src/stores/chatExtras.ts
DorraJaouad: why?

Antreesy: https://github.com/nextcloud/spreed/pull/15519#discussion_r2208084799

DorraJaouad: oh right

src/store/messagesStore.js
```diff
@@ -589,6 +616,10 @@ const actions = {
 					.forEach((storedMessage) => {
 						context.commit('addMessage', { token, message: Object.assign({}, storedMessage, { isThread: true }) })
 					})
+				// Fetch thread data in case it doesn't exist in the store yet
+				if (!chatExtrasStore.getThread(token, message.threadId) && chatExtrasStore.getThreadsList(token).length > 0) {
```
Antreesy:
```suggestion
				if (!chatExtrasStore.getThread(token, message.threadId) && chatExtrasStore.threads[token] !== undefined) {
```

src/store/messagesStore.js
```diff
@@ -647,6 +679,22 @@ const actions = {
 
 		context.commit('addMessage', { token, message })
 
+		const thread = chatExtrasStore.getThread(token, message.threadId)
+		// Update threads
+		if (message.isThread && thread) {
```
Antreesy:
```suggestion
		// Update threads
		if (message.isThread) {
				const thread = chatExtrasStore.getThread(token, message.threadId)
				if (thread) {
```

src/stores/chatExtras.ts
```diff
@@ -177,6 +177,20 @@ export const useChatExtrasStore = defineStore('chatExtras', {
 			}
 		},
 
+		clearThreads(token: string, messageId?: number) {
+			if (messageId) {
+				// Clear threads that are older than the given messageId
+				for (const threadId of Object.keys(this.threads[token] || {})) {
```
Antreesy:
```suggestion
				for (const threadId of Object.keys(Object(this.threads[token]))) {
```

src/stores/chatExtras.ts
```diff
@@ -177,6 +177,20 @@ export const useChatExtrasStore = defineStore('chatExtras', {
...
+					if (+threadId < messageId) {
+						delete this.threads[token][+threadId]
```
Antreesy, line 191:
```suggestion
						delete this.threads[token][threadId]
```

src/store/messagesStore.js
```diff
@@ -1416,7 +1464,11 @@ const actions = {
 	},
 
 	async removeExpiredMessages(context, { token }) {
-		context.commit('removeExpiredMessages', { token })
+		const updateRequired = context.commit('removeExpiredMessages', { token })
```
Antreesy: this returns undefined
but also makes no sense, as threads are not removed from DB, and server returns them with `first: null` and `last: null`

src/stores/chatExtras.ts
DorraJaouad: ts will complain

src/store/messagesStore.js
```diff
@@ -1416,7 +1445,22 @@ const actions = {
...
+		const timestamp = convertToUnix(Date.now())
+		const messageIds = Object.keys(context.state.messages[token])
+
+		messageIds.forEach((messageId) => {
+			const message = context.state.messages[token][messageId]
+			if (message.expirationTimestamp && timestamp > message.expirationTimestamp) {
+				if (message.isThread) {
+					chatExtrasStore.removeMessageFromThread(token, message.threadId, messageId)
+				}
+				context.commit('deleteMessage', { token, id: messageId })
```
Antreesy: Looks easier?
```suggestion

		context.getters.messagesList(token).forEach((message) => {
			if (message.expirationTimestamp && timestamp > message.expirationTimestamp) {
				if (message.isThread) {
					chatExtrasStore.removeMessageFromThread(token, message.threadId, message.id)
				}
				context.commit('deleteMessage', { token, id: message.id })
```

src/stores/chatExtras.ts
Antreesy: Object keys are always converted back to a string, so easier would be to change store type to:
```
threads: Record<string, Record<number|string, ThreadInfo>>
```

src/stores/chatExtras.ts
```diff
@@ -177,6 +177,49 @@ export const useChatExtrasStore = defineStore('chatExtras', {
 			}
 		},
 
+		/**
+		 * Remove a thread from the store
+		 *
+		 * @param token - conversation token
+		 * @param threadId - thread id to remove
```
Antreesy: Shouldn't be:
```suggestion
		 * @param messageId - message id to remove all preceding threads (remove all, if omitted)
```

src/stores/chatExtras.ts
```diff
@@ -177,6 +177,49 @@ export const useChatExtrasStore = defineStore('chatExtras', {
...
+		 * @param messageId - message id to remove
+		 */
+		removeMessageFromThread(token: string, threadId: number, messageId: number) {
+			if (!this.threads[token]?.[threadId]) {
+				return
+			}
+
+			const thread = this.threads[token][threadId]
+			if (thread.first.id === messageId) {
+				thread.first = null
```
Antreesy, line 215: Put `// @ts-expect-error` for now. Hopefully this can be fixed with https://github.com/nextcloud/openapi-extractor/issues/261

src/stores/chatExtras.ts
DorraJaouad: But also remeber that's a numerical comparison

Antreesy: when we compare ids, they should be casted to numbers, of course.
But keys in Object.keys are always string, and it doesnt matter for object[key] what type is the key: engine will convert it to string

src/components/RightSidebar/Threads/ThreadItem.vue
```diff
@@ -26,9 +27,19 @@ const { thread } = defineProps<{ thread: ThreadInfo }>()
...
+	if (!thread.first) {
+		return
+	}
+	return getDisplayNameWithFallback(thread.first.actorDisplayName, thread.first.actorType, true)
+})
 const lastActivity = computed(() => thread.thread.lastActivity * 1000)
-const name = computed(() => parseToSimpleMessage(thread.first.message, thread.first.messageParameters))
+const name = computed(() => {
+	if (!thread.first) {
+		return t('spreed', 'Thread origin message expired')
```
Antreesy, line 39: I think we already have a string like "message expired or deleted"

Antreesy (approved):
Looks good

src/components/RightSidebar/Threads/ThreadItem.vue
DorraJaouad: Yes, but this is more relevant to the context and specific

PR #15520, opened 2025-07-15
fix(threads): remove gray background on avatar and show load more when there is actually more

DorraJaouad:
### ☑️ Resolves

* Fix : Return state from thread tab and minor fixes

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/RightSidebar/RightSidebar.vue
+++ b/src/components/RightSidebar/RightSidebar.vue
@@ -260,6 +260,7 @@ export default {
 			contactsLoading: false,
 			unreadNotificationHandle: null,
 			contentState: 'default',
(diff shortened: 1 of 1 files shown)


src/components/TopBar/TopBar.vue
```diff
@@ -477,7 +477,6 @@ export default {
 
 .top-bar__icon-wrapper {
 	position: relative;
-	background-color: var(--color-background-dark);
 	border-radius: var(--border-radius-pill);
 	transition-property: width, padding;
```
Antreesy:
```suggestion
	transition-property: width, padding, background-color;
```

PR #15521, opened 2025-07-15
feat(Quote): add a shortcut to threaded view from replies

### LINKED ISSUE #15512: Improve styling of compact reply blocks
Although the idea with making the reply blocks from 2 lines to one was to improve spacing, currently the compact replies seem like they are bit too cramped. With a few font size and spacing adjustments it can be a bit more spaced out. In general we should also be more careful with using a small font size as it is not the most accessible.

Changes I did in the inspector:
- `.quote { padding-block: var(--default-grid-baseline); padding-inline: calc(2*var(--default-grid-baseline)) var(--default-clickable-area); }`
- Added `.quote { margin-bottom: var(--default-grid-baseline) } `
- Removed `.quote { font-size: var(--font-size-small) } `
- `.quote__main-author { gap: var(--default-grid-baseline) } `

| Before| After |
|----|----|
| Image | Image |

Antreesy:
### ☑️ Resolves

* Fix #15512
* Give some space
* Add a shortcut button to thread view from a quote (if this message is already a thread)

⚠️ Screenshots are made against unreleased vue-library (messages do not have margin-bottom)

🏚️ Before | 🏡 After
-- | --
2025-07-15_15h44_34 | 2025-07-15_15h39_58
2025-07-15_15h49_52 | 2025-07-15_15h50_02

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/Quote.vue
+++ b/src/components/Quote.vue
@@ -198,7 +198,6 @@ function handleQuoteClick() {
 		<NcButton v-if="canCancel"
 			class="quote__close"
 			variant="tertiary"
-			size="small"
 			:title="t('spreed', 'Cancel quote')"
 			:aria-label="t('spreed', 'Cancel quote')"
 			@click="handleAbort">
@@ -214,15 +213,15 @@ function handleQuoteClick() {
 
 .quote {
 	position: relative;
-	padding-block: calc(0.5 * var(--default-grid-baseline));
-	padding-inline: calc(1.5 * var(--default-grid-baseline)) var(--default-clickable-area);
+	padding-block: var(--default-grid-baseline);
+	padding-inline: calc(2 * var(--default-grid-baseline)) var(--default-clickable-area);
+	margin-bottom: var(--default-grid-baseline);
 	display: flex;
 	gap: var(--default-grid-baseline);
 	max-width: $messages-text-max-width;
-	min-height: var(--clickable-area-small);
+	min-height: var(--default-clickable-area);
 	border-radius: var(--border-radius-large);
 	border: 2px solid var(--color-border);
-	font-size: var(--font-size-small);
 	color: var(--color-text-maxcontrast);
 	background-color: var(--color-main-background);
 	overflow: hidden;
@@ -280,7 +279,7 @@ function handleQuoteClick() {
 		&-author {
 			display: flex;
 			align-items: center;
-			gap: calc(0.5 * var(--default-grid-baseline));
+			gap: var(--default-grid-baseline);
 
 			&-info {
 				flex-shrink: 0;
--- a/src/components/Quote.vue
+++ b/src/components/Quote.vue
@@ -13,6 +13,7 @@ import { useRoute } from 'vue-router'
 import NcButton from '@nextcloud/vue/components/NcButton'
 import NcLoadingIcon from '@nextcloud/vue/components/NcLoadingIcon'
 import IconClose from 'vue-material-design-icons/Close.vue'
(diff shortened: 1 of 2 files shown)


src/components/Quote.vue
```diff
@@ -303,10 +315,11 @@ function handleQuoteClick() {
 		}
 	}
 
-	&__close {
+	&__button {
 		position: absolute !important;
-		top: 0;
+		top: 50%;
 		inset-inline-end: 0;
+    transform: translateY(-50%);
```
DorraJaouad:
```suggestion
	transform: translateY(-50%);
	inset-inline-end: calc(var(--default-grid-baseline) / 2);
```

also I am not sure if you also intendend this

Image

but I don't mind it :D

src/components/Quote.vue
```diff
@@ -206,6 +207,17 @@ function handleQuoteClick() {
...
+			class="quote__button"
+			variant="tertiary"
+			size="small"
+			:title="t('spreed', 'Go to thread')"
+			:aria-label="t('spreed', 'Go to thread')"
+			@click.stop.prevent="threadId = message.threadId">
+			<template #icon>
+				<IconForumOutline :size="20" />
+			</template>
+		</NcButton>
```
DorraJaouad: Also eliminate temporary message

Antreesy: Why do we have to? I don't think it's problematic to post a message and jump to thread from it, while it's still sending, you'll see it in the thread anyway

DorraJaouad: because our default is not a thread, it is just a message. Making sthg a thread is user action

DorraJaouad: so whenever you send a normal message, you will see a thread icon glitch

Antreesy: It is only `Go to thread`, if thread already exists

DorraJaouad: Yes, but you are making temporary message a default thread, maybe it makes sense to revert that in the first place?

Antreesy: Looks correct to me:

2025-07-15_18h29_21

`threadId` and `isThread` only added if parent message is a thread already

UPD: looks broken inside of a thread, fixing that

DorraJaouad: Ah yes, I was testing in threaded view

src/components/Quote.vue
```diff
@@ -104,6 +105,10 @@ const shortenedQuoteMessage = computed(() => {
 	return simpleQuotedMessage.value.length >= 250 ? simpleQuotedMessage.value.substring(0, 250) + '…' : simpleQuotedMessage.value
 })
 
+const showThreadShortcut = computed(() => {
+	return !threadId.value && isExistingMessage(message) && message.isThread && message.threadId
```
DorraJaouad, line 109:
```suggestion
	return !threadId.value && message.isThread && message.threadId
```

Antreesy:
Modified button styles to make it look fully injected:
image
image

DorraJaouad:
Just a note, maybe it's better to follow the same color tone
image

src/components/Quote.vue
```diff
@@ -112,6 +117,12 @@ function isExistingMessage(message: ChatMessage | DeletedParentMessage): message
 	return 'messageType' in message
 }
 
+function goToThread() {
```
DorraJaouad, line 120: JSDoc alert just in case you missed it

PR #15522, opened 2025-07-15
fix(threads): Mark just created threads as thread

nickvergessen:
### ☑️ Resolves

- Found by @Ivansss 

🏚️ Before | 🏡 After
-- | --
grafik | grafik

### 🏁 Checklist


Antreesy:
Can you also check for missing:

```json
{
    "id": 571,
    "message": "{actor} created a thread",
    "systemMessage": "thread_created",
    "messageType": "system",
    "threadId": 570,
    "isThread": true,
    "parent": {
        "id": 570,
        "threadId": 570
		/* "isThread": true, */
    }
}
```

This is what I would expect to handle on client side (when anyone creates a thread, I'm receiving a system message, I'm not requesting a list of threads)

Antreesy (approved):
Tested

PR #15527, opened 2025-07-16
fix(lobby): update store from server response after lobby settings change

Antreesy:
### ☑️ Resolves

* On top of #15523
* Accepted format for NcDateTimePicker changed in Vue 3 (and for *Native picker it was never the case), see 2nd commit
* Adjusted logic for lobby timer (if we set the time, state is always 1)

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


DorraJaouad (approved):
Tested, works correctly

PR #15530, opened 2025-07-16
fix(MessagesList): extract fetching logic to shared composable

Antreesy:
### ☑️ Resolves

* Extracted without changing logic*
  \*apart from changes due to mounting/unmounting, but that's rather a fix
* polling is no longer cancelled when switching between lobby / chat / call
* some reactivity removed from places where it's not needed (timeout, intervals)

🏚️ Before | 🏡 After
-- | --
cancelled requests in console | all clear

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop


src/views/MainView.vue
```diff
@@ -29,6 +30,9 @@ const actorStore = useActorStore()
 const isInLobby = computed(() => store.getters.isInLobby)
 const connectionFailed = computed(() => store.getters.connectionFailed(props.token))
 
+// Make composable running whenever conversation route is active
+useGetMessages()
+
```
ShGKme: This is a misuse of the shared composable pattern.

The purpose of a shared composable is to have no more than a single instance of some reusable state and logic around it. Meaning if several components are using this composable simultaneously, they are using a single shared instance.

With `useGetMessages` the problem is different. We don't need to prevent a second instance, but we need an instance that always exists in the MainView and available in the components inside the main view. This is why `createSharedComposable` was not enough and you had to make a dump call here without actually using the composable.

What we need here is the context pattern with 2 composables:
1. `useGetMessagesProvider` - does everything `useGetMessages` does and provides it, creating a shared context for the subtree
   - This composable will be used here
2. `useGetMessagesContext` or just `useGetMessages` - injects and returns this context.

I'd define them in a single module (they are supposed to be a pair) and use `Symbol.for()` as the injection key to infer types.

src/composables/useGetMessages.ts
```diff
@@ -0,0 +1,336 @@
...
+const isChatBeginningReachedKey = Symbol.for('isChatBeginningReached') as InjectionKey<ComputedRef<boolean>>
+
+const getMessageContextKey = Symbol.for('getMessageContext') as InjectionKey<(token: string, messageId: number) => Promise<void>>
+const getOldMessagesKey = Symbol.for('getOldMessages') as InjectionKey<(token: string, includeLastKnown: boolean) => Promise<void>>
+
+/**
+ * Check whether caught error is from OCS API
+ */
+function isAxiosErrorResponse(exception: unknown): exception is AxiosError<string> {
+	return typeof exception === 'object' && 'response' in exception!
```
ShGKme: If `exception` is guaranteed to be non-nullable, it should be `NonNullable<unknown>`, not unknown.
If it isn't guaranteed, `!` should not be used (because it may be `null`).

```suggestion
	return exception !== null && typeof exception === 'object' && 'response' in exception
```

src/composables/useGetMessages.ts
```diff
@@ -0,0 +1,336 @@
...
+import { useGetThreadId } from './useGetThreadId.ts'
+import { useGetToken } from './useGetToken.ts'
+
+const loadingOldMessagesKey = Symbol.for('loadingOldMessages') as InjectionKey<Ref<boolean>>
+const isInitialisingMessagesKey = Symbol.for('isInitialisingMessages') as InjectionKey<Ref<boolean>>
+const stopFetchingOldMessagesKey = Symbol.for('stopFetchingOldMessages') as InjectionKey<Ref<boolean>>
+const isChatBeginningReachedKey = Symbol.for('isChatBeginningReached') as InjectionKey<ComputedRef<boolean>>
+
+const getMessageContextKey = Symbol.for('getMessageContext') as InjectionKey<(token: string, messageId: number) => Promise<void>>
+const getOldMessagesKey = Symbol.for('getOldMessages') as InjectionKey<(token: string, includeLastKnown: boolean) => Promise<void>>
```
ShGKme: - Explicitly defining type instead of type assertion with `as`
  - Otherwise there is no type checking
- Adding namespace to the injection key
  - To simplify debugging in dev tools and decrease conflict probability
- Renaming constants to `CONST_CASE`

```suggestion
const LOADING_OLD_MESSAGES_KEY: InjectionKey<Ref<boolean>> = Symbol.for('GetMessages:loadingOldMessages')
const IS_INITIALISING_MESSAGES_KEY: InjectionKey<Ref<boolean>> = Symbol.for('GetMessages:isInitialisingMessages')
const STOP_FETCHING_OLD_MESSAGES_KEY: InjectionKey<Ref<boolean>> = Symbol.for('GetMessages:stopFetchingOldMessages')
const IS_CHAT_BEGINNING_REACHED_KEY: InjectionKey<ComputedRef<boolean>> = Symbol.for('GetMessages:isChatBeginningReached')

const GET_MESSAGE_CONTEXT_KEY: InjectionKey<(token: string, messageId: number) => Promise<void>> = Symbol.for('GetMessages:getMessageContext')
const GET_OLD_MESSAGES_KEY: InjectionKey<(token: string, includeLastKnown: boolean) => Promise<void>> = Symbol.for('GetMessages:getOldMessages')
```

Alternatively, we can provide an entire context as a single object (these values are never provided/injected individually)

```suggestion
type GetMessagesContext = {
  loadingOldMessages: Ref<boolean>
  isInitialisingMessages: Ref<boolean>
  stopFetchingOldMessages: <Ref<boolean>
  isChatBeginningReached: ComputedRef<boolean>
  getMessageContext: (token: string, messageId: number) => Promise<void>
  getOldMessages: (token: string, includeLastKnown: boolean) => Promise<void>
}

const GET_MESSAGES_CONTEXT_KEY: InjectionKey<GetMessagesContext> = Symbol.for('GET_MESSAGES_CONTEXT')
```

src/composables/useGetMessages.ts
```diff
@@ -0,0 +1,336 @@
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
```
```suggestion
 * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
```

src/composables/useGetMessages.ts
Antreesy: `catch(exception)` has type unknown and currently there is no other thing to narrow it down in Typescript:

> Vue: Catch clause variable type annotation must be any or unknown if specified

src/views/MainView.vue
```diff
@@ -29,6 +30,9 @@ const actorStore = useActorStore()
 const isInLobby = computed(() => store.getters.isInLobby)
 const connectionFailed = computed(() => store.getters.connectionFailed(props.token))
 
+// Provide context for MessagesList mounted in different places
+useGetMessagesProvider()
```
DorraJaouad: MessagesList exists only in ChatView but ChatView doesn't exist only in MainView
Image

Antreesy: yes, just noticed 🙈

DorraJaouad:
Ah should it be kept in composables or moved to utils?

PR #15543, opened 2025-07-17
fix(dashboard): don't show cancelled events in upcoming events

miaulalala:
### ☑️ Resolves

* Fix #…

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist


nickvergessen (approved):

lib/Service/CalendarIntegrationService.php
```diff
@@ -107,6 +107,10 @@ public function getDashboardEvents(string $userId): array {
 					continue;
 				}
 
+				if ($event['STATUS'][0] === 'CANCELLED') {
```
nickvergessen, line 110: Warnings logged for this one:
> Undefined array key "STATUS" at apps/spreed/lib/Service/CalendarIntegrationService.php#110

PR #15545, opened 2025-07-18
fix(chat): Fix guests in a fully empty chat

nickvergessen:
### ☑️ Resolves

* Create a chat
* Make it public
* Configure message expiration
* Purge the chat history
* Join as a guest in a private window

Before | After
---|---
See loading animation for ever | See empty content
grafik | grafik

### 🏁 Checklist

### DIFF
--- a/lib/Service/RoomFormatter.php
+++ b/lib/Service/RoomFormatter.php
@@ -418,6 +418,11 @@ class RoomFormatter {
 			}
 		}
 
+		if ($roomData['lastReadMessage'] === 0) {
+			// Guest in a fully expired chat, no history, just loading the chat from beginning for now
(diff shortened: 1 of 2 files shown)


lib/Service/RoomFormatter.php
```diff
@@ -418,6 +418,11 @@ public function formatRoomV4(
 			}
 		}
 
+		if ($roomData['lastReadMessage'] === 0) {
```
Antreesy, line 421: Wouldn't be better to check it later?
```php
if ($currentUser instanceof IUser) {
} else if (/* here */) {
```

nickvergessen: Well in fact any user with 0 will fail. So I thought better be save then sorry?

PR #15559, opened 2025-07-19
fix(useGetMessages): fix getting context around focused message

Antreesy:
### ☑️ Resolves

* Follow-up to #15530
  * Move route change listener to context provider (isolating `getMessageContext()`)
  * Refactor initializing logic (preconditions + first/lastKnown for context + find message)
  * Replace chatIdentifier with token-only tracking,
  * Check preconditions with route change
    * This fix loading initial context for URLs like `{token}/#message_123`

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/composables/useGetMessages.ts
+++ b/src/composables/useGetMessages.ts
@@ -18,7 +18,7 @@ import type {
 import Axios from '@nextcloud/axios'
 import { subscribe, unsubscribe } from '@nextcloud/event-bus'
 import { computed, inject, onBeforeUnmount, provide, ref, watch } from 'vue'
-import { useRoute } from 'vue-router'
+import { START_LOCATION, useRoute, useRouter } from 'vue-router'
 import { useStore } from 'vuex'
 import { CHAT, MESSAGE } from '../constants.ts'
 import { EventBus } from '../services/EventBus.ts'
@@ -53,6 +53,7 @@ let pollingErrorTimeout = 1_000
  */
 export function useGetMessagesProvider() {
 	const store = useStore()
+	const router = useRouter()
 	const route = useRoute()
 
 	const currentToken = useGetToken()
@@ -84,9 +85,19 @@ export function useGetMessagesProvider() {
 			&& ['conversation_created', 'history_cleared'].includes(firstKnownMessage.value.systemMessage))
 	})
 
+	/** Initial check to ensure context is created once route is available */
+	router.isReady().then(() => {
+		if (currentToken.value && isParticipant.value && !isInLobby.value) {
+			handleStartGettingMessagesPreconditions(currentToken.value)
+		}
+	})
+
 	watch(
 		[currentToken, () => isParticipant.value && !isInLobby.value],
 		([newToken, canGetMessages], [oldToken, _oldCanGetMessages]) => {
+			if (route.name === START_LOCATION.name) { // Direct object comparison does not work
+				return
+			}
 			if (oldToken && oldToken !== newToken) {
 				store.dispatch('cancelPollNewMessages', { requestId: oldToken })
 			}
@@ -100,7 +111,6 @@ export function useGetMessagesProvider() {
 			/** Remove expired messages when joining a room */
 			store.dispatch('removeExpiredMessages', { token: newToken })
 		},
-		{ immediate: true },
 	)
 
 	subscribe('networkOffline', handleNetworkOffline)
@@ -191,7 +201,7 @@ export function useGetMessagesProvider() {
 	async function handleStartGettingMessagesPreconditions(token: string) {
 		// prevent sticky mode before we have loaded anything
 		isInitialisingMessages.value = true
-		const focusMessageId = getMessageIdFromHash(route?.hash)
+		const focusMessageId = getMessageIdFromHash(route.hash)
 
 		store.dispatch('setVisualLastReadMessageId', { token, id: conversation.value!.lastReadMessage })
 
--- a/src/composables/useGetMessages.ts
+++ b/src/composables/useGetMessages.ts
@@ -90,7 +90,11 @@ export function useGetMessagesProvider() {
 		if (oldValue) {
 			store.dispatch('cancelPollNewMessages', { requestId: oldValue })
 		}
-		handleStartGettingMessagesPreconditions(currentToken.value)
+		if (currentToken.value && isParticipant.value && !isInLobby.value) {
+			handleStartGettingMessagesPreconditions(currentToken.value)
+		} else {
+			store.dispatch('cancelPollNewMessages', { requestId: newValue })
+		}
 
 		/** Remove expired messages when joining a room */
 		store.dispatch('removeExpiredMessages', { token: currentToken.value })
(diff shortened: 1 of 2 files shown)


src/composables/useGetMessages.ts
```diff
@@ -85,18 +84,18 @@ export function useGetMessagesProvider() {
 			&& ['conversation_created', 'history_cleared'].includes(firstKnownMessage.value.systemMessage))
 	})
 
-	watch(chatIdentifier, (newValue, oldValue) => {
-		if (oldValue) {
-			store.dispatch('cancelPollNewMessages', { requestId: oldValue })
+	watch(() => isParticipant.value && !isInLobby.value, (shouldStartGettingMessagesPreconditions: boolean) => {
+		if (currentToken.value && shouldStartGettingMessagesPreconditions) {
+			const focusMessageId = route?.hash?.startsWith('#message_') ? parseInt(route.hash.slice(9), 10) : null
```
DorraJaouad: Maybe better in computed or a method as it is used more than one time

src/composables/useGetMessages.ts
```diff
@@ -185,13 +236,19 @@ export function useGetMessagesProvider() {
 	 * @param messageId messageId
 	 */
 	async function getMessageContext(token: string, messageId: number) {
-		// Make the request
 		loadingOldMessages.value = true
 		try {
 			debugTimer.start(`${token} | get context`)
+			// Update environment around context
+			await store.dispatch('setFirstKnownMessageId', { token, id: messageId })
+			await store.dispatch('setLastKnownMessageId', { token, id: messageId })
```
DorraJaouad: why async?

src/composables/useGetMessages.ts
```diff
@@ -116,66 +116,117 @@ export function useGetMessagesProvider() {
...
+		/** Cancel previously opened chat requests */
+		if (from.name === 'conversation' && from.params.token) {
+			store.dispatch('cancelPollNewMessages', { requestId: from.params.token })
+		}
+
+		if (from.params.token !== to.params.token && isParticipant.value && !isInLobby.value) {
+			handleStartGettingMessagesPreconditions(to.params.token as string, focusMessageId)
+		} else {
+			store.dispatch('cancelPollNewMessages', { requestId: to.params.token })
+		}
```
DorraJaouad: I would say this is redundant as watcher can also handle the cases covered here, we just need to include token value in the watcher

src/composables/useGetMessages.ts
```diff
@@ -116,66 +116,117 @@ export function useGetMessagesProvider() {
...
+	/**
+	 * Handle route changes to initialize chat or thread, and focus given message
+	 */
+	async function onRouteChange({ from, to }: { from: RouteLocation, to: RouteLocation }) {
+		const focusMessageId = to.hash && to.hash.startsWith('#message_') ? parseInt(to.hash.slice(9), 10) : null
+
+		/** Cancel previously opened chat requests */
+		if (from.name === 'conversation' && from.params.token) {
+			store.dispatch('cancelPollNewMessages', { requestId: from.params.token })
+		}
```
DorraJaouad: route change can be in the same conversation

src/composables/useGetMessages.ts
```diff
@@ -116,66 +116,117 @@ export function useGetMessagesProvider() {
...
+
+		if (from.params.token !== to.params.token && isParticipant.value && !isInLobby.value) {
+			handleStartGettingMessagesPreconditions(to.params.token as string, focusMessageId)
+		} else {
+			store.dispatch('cancelPollNewMessages', { requestId: to.params.token })
```
DorraJaouad: Also this will be called when navigating in the same conversation bc `from.params.token !== to.params.token` is falsy

src/composables/useGetMessages.ts
```diff
@@ -85,18 +84,28 @@ export function useGetMessagesProvider() {
 			&& ['conversation_created', 'history_cleared'].includes(firstKnownMessage.value.systemMessage))
 	})
 
-	watch(chatIdentifier, (newValue, oldValue) => {
-		if (oldValue) {
-			store.dispatch('cancelPollNewMessages', { requestId: oldValue })
-		}
-		handleStartGettingMessagesPreconditions(currentToken.value)
+	watch(
```
Antreesy, line 95: this should wait for router to be initialized

Antreesy: Done

src/composables/useGetMessages.ts
```diff
@@ -85,18 +85,37 @@ export function useGetMessagesProvider() {
...
-		/** Remove expired messages when joining a room */
-		store.dispatch('removeExpiredMessages', { token: currentToken.value })
-	}, { immediate: true })
+	watch(
+		[currentToken, () => isParticipant.value && !isInLobby.value],
+		([newToken, canGetMessages], [oldToken, _oldCanGetMessages]) => {
+			if (route.name === START_LOCATION.name) { // Direct object comparison does not work
+				return
+			}
+			if (oldToken) {
```
```suggestion
			if (oldToken && oldToken !== newToken) {
```

src/composables/useGetMessages.ts
```diff
@@ -106,76 +125,112 @@ export function useGetMessagesProvider() {
...
+	/**
+	 * Handle route changes to initialize chat or thread, and focus given message
+	 */
+	async function onRouteChange({ from, to }: { from: RouteLocation, to: RouteLocation }) {
+		if (from.name === 'conversation' && to.name === 'conversation'
+			&& from.params.token === to.params.token && typeof to.params.token === 'string') {
+			const focusMessageId = getMessageIdFromHash(to.hash)
+			if (from.hash !== to.hash && focusMessageId !== null) {
+				// the hash changed, need to focus/highlight another message
+				const hasMessageInStore = ('id' in (store.getters.message(to.params.token, focusMessageId) as ChatMessage | Record<string, never>))
```
DorraJaouad: What if we change the return value of not found message to null or undefined to make life easier? 
Otherwise https://www.geeksforgeeks.org/javascript/lodash-_-isempty-method/

src/composables/useGetMessages.ts
```diff
@@ -106,76 +125,112 @@ export function useGetMessagesProvider() {
...
+		if (currentToken.value) {
+			console.debug('Restarting polling of new chat messages')
+			pollNewMessages(currentToken.value)
+		}
+	}
+
+	/**
+	 * Handle route changes to initialize chat or thread, and focus given message
+	 */
+	async function onRouteChange({ from, to }: { from: RouteLocation, to: RouteLocation }) {
```
DorraJaouad, line 165:
```suggestion
	async function onRouteChange({ from, to }: { from: RouteLocation, to: RouteLocation }) {
		if (from.name !== to.name || from.name !== 'conversation' 
		|| from.params.token !== to.params.token || typeof to.params.token !== 'string') {
		return
	}
```

and remove the wrapper if

src/composables/useGetMessages.ts
```diff
@@ -106,76 +125,112 @@ export function useGetMessagesProvider() {
...
+			}
+
+			if (to.query.threadId && from.query.threadId !== to.query.threadId) {
+				// FIXME temporary get thread messages from the start
+				const hasMessageInStore = ('id' in (store.getters.message(to.params.token, to.query.threadId) as ChatMessage | Record<string, never>))
+				if (!hasMessageInStore) {
+					await getMessageContext(to.params.token, +to.query.threadId)
+				}
+			}
+		}
```
DorraJaouad: Unless you will provide different logic to thread context soon, better logic flow is 
```js
		const focusMessageId = getMessageIdFromHash(to.hash)
		const threadIdToCheck = to.query.threadId && from.query.threadId !== to.query.threadId ? +to.query.threadId : null
		
		// Determine which message ID to check and fetch context for
		let messageIdToFetch: number | null = null
		let shouldFocusMessage = false
		
		if (from.hash !== to.hash && focusMessageId !== null) {
			// the hash changed, need to focus/highlight another message
			messageIdToFetch = focusMessageId
			shouldFocusMessage = true
		} else if (threadIdToCheck) {
			// FIXME temporary get thread messages from the start
			messageIdToFetch = threadIdToCheck
		}
		
		if (messageIdToFetch !== null) {
			const hasMessageInStore = ('id' in (store.getters.message(to.params.token, messageIdToFetch) as ChatMessage | Record<string, never>))
			if (!hasMessageInStore) {
				await getMessageContext(to.params.token, messageIdToFetch)
			}
			
			if (shouldFocusMessage) {
				// need some delay (next tick is too short) to be able to run
				// after the browser's native "scroll to anchor" from the hash
				window.setTimeout(() => {
					EventBus.emit('focus-message', focusMessageId!)
				}, 2)
			}
		}
```

src/composables/useGetMessages.ts
Antreesy: Only if it's a follow-up. It's currently in 17 places, and I can't imagine, what might break

And we want to get rid of lodash, not add more of it

src/composables/useGetMessages.ts
Antreesy: I will, thread handling is temporary in this place. Don't think it's viable to mix them now

src/composables/useGetMessages.ts
DorraJaouad: Ah oki noted

PR #15566, opened 2025-07-21
fix(messages): implement chat blocks for message rendering

Antreesy:
### ☑️ Resolves

* Move messages list getter (actually used in components) to a new Pinia store
* Implement chat blocks feature for the later use

No visual changes

### 🚧 Tasks

- [ ] Retain existing behaviour for chat to render messages
- [ ] Prepare for future changes

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/store/messagesStore.js
+++ b/src/store/messagesStore.js
@@ -26,6 +26,7 @@ import {
 } from '../services/messagesService.ts'
 import { useActorStore } from '../stores/actor.ts'
 import { useCallViewStore } from '../stores/callView.ts'
+import { useChatStore } from '../stores/chat.ts'
 import { useChatExtrasStore } from '../stores/chatExtras.ts'
 import { useGuestNameStore } from '../stores/guestName.js'
 import { usePollsStore } from '../stores/polls.ts'
@@ -992,6 +993,10 @@ const actions = {
 		}
 
 		context.commit('loadedMessagesOfConversation', { token })
+		const chatStore = useChatStore()
+		chatStore.processChatBlocks(token, response.data.ocs.data, {
+			mergeBy: +lastKnownMessageId,
+		})
 
 		if (minimumVisible > 0) {
 			debugTimer.tick(`${token} | fetch history`, 'first chunk')
@@ -1087,6 +1092,9 @@ const actions = {
 		}
 
 		context.commit('loadedMessagesOfConversation', { token })
+
+		const chatStore = useChatStore()
+		chatStore.processChatBlocks(token, response.data.ocs.data)
 
 		if (minimumVisible > 0) {
 			debugTimer.tick(`${token} | get context`, 'first chunk')
@@ -1265,6 +1273,11 @@ const actions = {
 		}
 
 		context.commit('loadedMessagesOfConversation', { token })
+
+		const chatStore = useChatStore()
+		chatStore.processChatBlocks(token, response.data.ocs.data, {
+			mergeBy: +lastKnownMessageId,
+		})
 
 		return response
 	},
--- a/src/store/messagesStore.js
+++ b/src/store/messagesStore.js
@@ -510,6 +510,8 @@ const mutations = {
 		newMessagesToRemove.forEach((messageId) => {
 			delete state.messages[token][messageId]
 		})
+		const chatStore = useChatStore()
+		chatStore.removeMessagesFromChatBlocks(token, [...messagesToRemove, ...newMessagesToRemove].map((id) => +id))
 
 		if (state.firstKnown[token] && messagesToRemove.includes(state.firstKnown[token].toString())) {
 			state.firstKnown[token] = +newFirstKnown
@@ -760,6 +762,8 @@ const actions = {
 	 */
 	addTemporaryMessage(context, { token, message }) {
 		context.commit('addTemporaryMessage', { token, message })
+		const chatStore = useChatStore()
+		chatStore.addMessageToChatBlocks(token, message)
 		// Update conversations list order
 		context.dispatch('updateConversationLastActive', token)
 	},
@@ -788,6 +792,9 @@ const actions = {
 	 */
 	removeTemporaryMessageFromStore(context, { token, id }) {
 		context.commit('deleteMessage', { token, id })
+
+		const chatStore = useChatStore()
+		chatStore.removeMessagesFromChatBlocks(token, id)
 	},
 
 	/**
@@ -1338,6 +1345,8 @@ const actions = {
 			// Own message might have been added already by polling, which is more up-to-date (e.g. reactions)
 			if (!context.state.messages[token]?.[response.data.ocs.data.id]) {
 				context.dispatch('processMessage', { token, message: response.data.ocs.data })
+				const chatStore = useChatStore()
+				chatStore.addMessageToChatBlocks(token, response.data.ocs.data)
 			}
 
 			return response
@@ -1462,6 +1471,7 @@ const actions = {
 			return
 		}
 		const chatExtrasStore = useChatExtrasStore()
+		const chatStore = useChatStore()
 		const timestamp = convertToUnix(Date.now())
 
 		context.getters.messagesList(token).forEach((message) => {
@@ -1470,6 +1480,7 @@ const actions = {
 					chatExtrasStore.removeMessageFromThread(token, message.threadId, message.id)
 				}
 				context.commit('deleteMessage', { token, id: message.id })
+				chatStore.removeMessagesFromChatBlocks(token, message.id)
 			}
 		})
 	},
--- a/src/stores/__tests__/chat.spec.js
+++ b/src/stores/__tests__/chat.spec.js
@@ -159,4 +159,95 @@ describe('chatStore', () => {
 			expect(chatStore.chatBlocks[TOKEN]).toEqual([outputSet(chatBlockA), outputSet(chatBlockB), outputSet(chatBlockC)])
 		})
 	})
+
+	describe('add messages', () => {
+		it('creates a new block, if not created yet', () => {
+			// Act
+			chatStore.addMessageToChatBlocks(TOKEN, chatBlockD[0])
+
+			// Assert
+			expect(chatStore.chatBlocks[TOKEN]).toEqual([outputSet(chatBlockD)])
+		})
+
+		it('extends the most recent block', () => {
+			// Arrange
+			chatStore.processChatBlocks(TOKEN, chatBlockA)
+			chatStore.processChatBlocks(TOKEN, chatBlockB)
+
+			// Act
+			chatStore.addMessageToChatBlocks(TOKEN, chatBlockD[0])
+
+			// Assert
+			expect(chatStore.chatBlocks[TOKEN]).toEqual([outputSet(chatBlockD, chatBlockA), outputSet(chatBlockB)])
+		})
+
+		it('does nothing, if message is already present in the most recent block', () => {
+			// Arrange
+			chatStore.processChatBlocks(TOKEN, chatBlockA)
+
+			// Act
+			chatStore.addMessageToChatBlocks(TOKEN, chatBlockA[0])
+
+			// Assert
+			expect(chatStore.chatBlocks[TOKEN]).toEqual([outputSet(chatBlockA)])
+		})
+	})
+
+	describe('remove messages', () => {
+		it('does nothing, if no blocks are created yet', () => {
+			// Act
+			chatStore.removeMessagesFromChatBlocks(TOKEN, chatBlockD[0].id)
+
+			// Assert
+			expect(chatStore.chatBlocks[TOKEN]).toBeUndefined()
+		})
+
+		it('does nothing, if message is not present in existing blocks', () => {
+			// Arrange
+			chatStore.processChatBlocks(TOKEN, chatBlockA)
+
+			// Act
+			chatStore.removeMessagesFromChatBlocks(TOKEN, chatBlockD[0].id)
+
+			// Assert
+			expect(chatStore.chatBlocks[TOKEN]).toEqual([outputSet(chatBlockA)])
+		})
+
+		it('removes a message id from all blocks', () => {
+			// Arrange
+			chatStore.processChatBlocks(TOKEN, chatBlockA)
+
+			// Act
+			chatStore.removeMessagesFromChatBlocks(TOKEN, chatBlockA[0].id)
+
+			// Assert
+			expect(chatStore.chatBlocks[TOKEN]).toEqual([outputSet([chatBlockA[1]])])
+		})
+
+		it('removes a list of message ids and clears up empty blocks', () => {
+			// Arrange
+			chatStore.processChatBlocks(TOKEN, chatBlockA)
+			chatStore.processChatBlocks(TOKEN, chatBlockB)
+
+			// Act
+			chatStore.removeMessagesFromChatBlocks(TOKEN, chatBlockB.map((message) => message.id))
+
+			// Assert
+			expect(chatStore.chatBlocks[TOKEN]).toHaveLength(1)
+			expect(chatStore.chatBlocks[TOKEN]).toEqual([outputSet(chatBlockA)])
+		})
+
+		it('clears up store after removing of all blocks', () => {
+			// Arrange
+			chatStore.processChatBlocks(TOKEN, chatBlockB)
+			chatStore.processChatBlocks(TOKEN, chatBlockA)
+
+			// Act
+			chatStore.removeMessagesFromChatBlocks(TOKEN, chatBlockB.map((message) => message.id))
+			chatStore.removeMessagesFromChatBlocks(TOKEN, chatBlockA.map((message) => message.id))
+
+			// Assert
+			expect(chatStore.chatBlocks[TOKEN]).toBeUndefined()
+		})
+	})
 })
--- a/src/stores/__tests__/chat.spec.js
+++ b/src/stores/__tests__/chat.spec.js
@@ -83,4 +83,80 @@ describe('chatStore', () => {
 			expect(chatStore.getMessagesList('token3')).toEqual([]) // Neither messages nor blocks
 		})
 	})
+
+	describe('process messages chunks', () => {
+		it('creates a new block, if not created yet', () => {
+			// Act
+			chatStore.processChatBlocks(TOKEN, chatBlockA)
+
+			// Assert
+			expect(chatStore.chatBlocks[TOKEN]).toEqual([outputSet(chatBlockA)])
+		})
+
+		it('extends an existing block, if messages overlap', () => {
+			// Arrange
+			chatStore.processChatBlocks(TOKEN, chatBlockA)
+
+			// Act
+			chatStore.processChatBlocks(TOKEN, chatBlockE)
+
+			// Assert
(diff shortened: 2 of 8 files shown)


src/stores/messages.ts
```diff
@@ -0,0 +1,72 @@
...
+	function getMessagesList(token: string): ChatMessage[] {
+		if (!store.state.messagesStore.messages[token] || !messageBlocks[token]?.[0]) {
+			return []
+		}
+
+		return Array.from(messageBlocks[token][0])
+			.sort((a, b) => a - b)
+			.reduce<ChatMessage[]>((acc, id) => {
+			const message = store.state.messagesStore.messages[token][id]
+			if (message) {
```
DorraJaouad: Let's state what cases are when message is not found in store but in ids map? We need to handle them carefully if any

src/store/messagesStore.js
```diff
@@ -963,6 +964,10 @@ const actions = {
 		}
 
 		context.commit('loadedMessagesOfConversation', { token })
+		const messagesStore = useMessagesStore()
+		messagesStore.processMessageBlocks(token, response.data.ocs.data, {
+			mergeBy: +lastKnownMessageId,
+		})
```
DorraJaouad: `mergeBy` is a limited approach IMO, it will only work for `lastKnownMessageId` because we know that's the id after which we need to append more messages, but what if we want to append a block B between two blocks A and C and how can we know it should be merged  to A for example bc there are no more messages in between?

src/stores/messages.ts
```diff
@@ -0,0 +1,72 @@
...
+		// Try to merge by known border id
+		if (options?.mergeBy) {
+			Object.keys(messageBlocks[token]).forEach((key) => {
+				if (messageBlocks[token][key].has(options.mergeBy!)) {
+					newMessageIds.forEach((id: number) => {
+						messageBlocks[token][key].add(id)
+					})
+				}
+			})
+		}
```
DorraJaouad: There is a big assumption here that the id is at the end of the block :P are you sure this is always true?

src/stores/messages.ts
```diff
@@ -0,0 +1,72 @@
...
+			.reduce<ChatMessage[]>((acc, id) => {
+			const message = store.state.messagesStore.messages[token][id]
+			if (message) {
+				acc.push(message)
+			}
+			return acc
+		}, [])
+	}
+
+	function processMessageBlocks(token: string, messages: ChatMessage[], options?: ProcessMessageBlocksOptions): void {
```
DorraJaouad: Where is the case where `mergeBy` is not passed and `!messageBlocks[token]` is not empty?

src/stores/messages.ts
Antreesy: it's always the case for system messages like reactions, edit, delete, e.t.c. Don't see the need to leave a comment for it in the code

src/store/messagesStore.js
Antreesy: Implemented, see `tryMergeExtraBlocks`
We can't know it for sure, until it overlaps

src/stores/messages.ts
Antreesy: See new implementation:
```js
if (options?.mergeBy) {
	newMessageIdsSet.add(options.mergeBy)
}
```
it doesn't matter for the Set, if it's in the end (Set is not sorted anyway). If we have `options.mergeBy`, we know:
- new messages are adjacent to it, so it's fine for new block to include it in set
- we already know mergeBy is from the client last-knowns, so it's already exists in some of the blocks   - there will be guaranteed intersection

src/stores/messages.ts
Antreesy: See `extraChatBlocks` added

src/stores/messages.ts
DorraJaouad: I meant in the comment here, so we don't overlook them

src/stores/chat.ts
```diff
@@ -0,0 +1,162 @@
...
+			chatBlocks[token] = [newMessageIdsSet]
+			return
+		}
+
+		if (options?.mergeBy) {
+			newMessageIdsSet.add(options.mergeBy)
+		}
+
+		if (!chatBlocks[token].some((parentBlock) => tryMergeBlocks(parentBlock, newMessageIdsSet))) {
+			chatBlocks[token].push(newMessageIdsSet)
```
DorraJaouad:
```suggestion
			if (!chatBlocks[token].some((parentBlock) => tryMergeBlocks(parentBlock, newMessageIdsSet))) {
			let i = chatBlocks[token].length - 1
			while (i >= 0 && Math.max(...Array.from(chatBlocks[token][i])) >= Math.max(...Array.from(newMessageIdsSet))) {
				i--
			}
			chatBlocks[token].splice(i + 1, 0, newMessageIdsSet)
		}
```

I remember this logic from high school but feel free to adjust to modern way :') 

If we directly add the item in the right position, we will have the array always sorted and no need to iterate over all of it each time 

~~mergeAndSortChatBlocks(token)~~

src/stores/chat.ts
```diff
@@ -0,0 +1,162 @@
...
+
+		if (!chatBlocks[token]) {
+			// If no blocks exist, create a new one with the first message. First in array will be considered main block
+			chatBlocks[token] = [newMessageIdsSet]
+			return
+		}
+
+		if (options?.mergeBy) {
+			newMessageIdsSet.add(options.mergeBy)
+		}
```
DorraJaouad, line 78: Good one, so the new set gets captured for sure to be merged

src/stores/chat.ts
```diff
@@ -0,0 +1,162 @@
...
+	 * Removes one or more message ids from all chat blocks
+	 */
+	function removeMessagesFromChatBlocks(token: string, messageIds: number | number[]) {
+		if (!chatBlocks[token]) {
+			return
+		}
+
+		const messageIdArray = Array.isArray(messageIds) ? messageIds : [messageIds]
+
+		chatBlocks[token] = chatBlocks[token].filter((block) => block.size > 0)
```
DorraJaouad: You already block keeping empty arrays right below this line, no need to do so here

src/stores/chat.ts
```diff
@@ -0,0 +1,162 @@
...
+			// Sort blocks in descending order by max id. First in array will be considered main block (recent messages)
+			.sort((a, b) => Math.max(...Array.from(b)) - Math.max(...Array.from(a)))
+	}
+
+	/**
+	 * Check, if child block is intersecting with parent, and extend parent in this case
+	 * Returns true if parent was extended, false otherwise
+	 */
+	function tryMergeBlocks(parentBlock: Set<number>, childBlock: Set<number>): boolean {
+		if (checkIfIntersect(parentBlock, childBlock)) {
```
DorraJaouad, line 123:
```suggestion
		if (Array.from(childBlock).some((id) => parentBlock.has(id))) 
```

~~checkIfIntersect~~

DorraJaouad (commented):
Promising

src/stores/chat.ts
ShGKme: Is it different from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/isDisjointFrom ?

Antreesy: It's the same, but Set methods have some issues with node/typescript, haven't figured yet. Possible because of Node 20, but also won't work on Safari < 17

And Array.from().some() iterates the set two times with `O(n+min(n))`, whereas Set method or checkintersect does it with `O(min(n))`

src/stores/chat.ts
Antreesy: If dropping postcheck, this needs adjustments for merging `A + new + B` wherever it's inserted

DorraJaouad: Whenever we try to insert 'new', it will be when merged if needed `tryMergeBlocks(parentBlock, newMessageIdsSet)` or inserted in a sorted way

src/stores/chat.ts
DorraJaouad: Both are `O(n)` we merge `k+n` and `k*n` to `n` afaik

Antreesy: Still a bit more performant to iterate Set, rather than convert Set to Array and iterate Array

src/stores/__tests__/chat.spec.js
```diff
@@ -0,0 +1,290 @@
...
+			const outputMain = new Set(mockMessages.slice(0, 3).map((message) => message.id))
+			const outputExtra = new Set(mockMessages.slice(3).map((message) => message.id))
+			chatStore.processChatBlocks(TOKEN, mockMessages.slice(0, 3)) // main block
+
+			// Act
+			chatStore.processChatBlocks(TOKEN, mockMessages.slice(3))
+
+			// Assert
+			expect(chatStore.chatBlocks[TOKEN]).toHaveLength(2)
+			expect(chatStore.chatBlocks[TOKEN]).toEqual([outputExtra, outputMain])
```
Antreesy: For tests readability
```suggestion
			expect(chatStore.chatBlocks[TOKEN]).toEqual([blockA, blockB])
```

src/stores/chat.ts
ShGKme: > It's the same, but Set methods have some issues with node/typescript, haven't figured yet. Possible because of Node 20, but also won't work on Safari < 17

Other new ES features are also not available in older browsers. That's covered by transpiler and polyfills

Antreesy: Then I will check if fixable for tsc and jest

src/stores/__tests__/chat.spec.js
```diff
@@ -0,0 +1,290 @@
...
+		it('returns empty array if no messages or blocks', () => {
+			// Arrange
+			vuexStore.dispatch('processMessage', { token: 'token1', message: mockMessages[0] })
+			chatStore.processChatBlocks('token2', [mockMessages[0]])
+
+			// Assert
+			expect(chatStore.getMessagesList('token1')).toEqual([])
+			expect(chatStore.getMessagesList('token2')).toEqual([])
+			expect(chatStore.getMessagesList('token3')).toEqual([])
+		})
```
DorraJaouad: Image

*Am I a joke to you?*

src/stores/__tests__/chat.spec.js
```diff
@@ -0,0 +1,290 @@
...
+	describe('get list of messages', () => {
+		it('returns empty array if no messages or blocks', () => {
+			// Arrange
+			vuexStore.dispatch('processMessage', { token: 'token1', message: mockMessages[0] })
+			chatStore.processChatBlocks('token2', [mockMessages[0]])
+
+			// Assert
+			expect(chatStore.getMessagesList('token1')).toEqual([])
+			expect(chatStore.getMessagesList('token2')).toEqual([])
+			expect(chatStore.getMessagesList('token3')).toEqual([])
```
DorraJaouad:
```suggestion
```

src/stores/__tests__/chat.spec.js
```diff
@@ -0,0 +1,290 @@
...
+			delete chatStore.chatBlocks[key]
+		})
+		jest.clearAllMocks()
+	})
+
+	describe('get list of messages', () => {
+		it('returns empty array if no messages or blocks', () => {
+			// Arrange
+			vuexStore.dispatch('processMessage', { token: 'token1', message: mockMessages[0] })
+			chatStore.processChatBlocks('token2', [mockMessages[0]])
```
DorraJaouad:
```suggestion
			chatStore.processChatBlocks('token1', [mockMessages[0]])
```

src/stores/__tests__/chat.spec.js
```diff
@@ -0,0 +1,290 @@
...
+
+			// Assert
+			expect(chatStore.getMessagesList('token1')).toEqual([])
```
DorraJaouad: It worked because it is an array comparison

src/stores/__tests__/chat.spec.js
```diff
@@ -0,0 +1,290 @@
...
+			expect(chatStore.getMessagesList('token2')).toEqual([])
+			expect(chatStore.getMessagesList('token3')).toEqual([])
+		})
+
+		it('returns array of messages from chatStore', () => {
```
DorraJaouad:
```suggestion
		it('returns array of messages only if they exist in chatStore', () => {
```

src/stores/__tests__/chat.spec.js
```diff
@@ -0,0 +1,290 @@
...
+
+			// Act
+			chatStore.processChatBlocks(TOKEN, mockMessages.slice(3))
+
+			// Assert
+			expect(chatStore.chatBlocks[TOKEN]).toHaveLength(1)
+			expect(chatStore.chatBlocks[TOKEN]).toEqual([output])
+		})
+
+		it('extends and existing block, if messages adjacent', () => {
```
DorraJaouad:
```suggestion
		it('extends an existing block, if messages are adjacent', () => {
```

src/stores/__tests__/chat.spec.js
```diff
@@ -0,0 +1,290 @@
...
+			// Arrange
+			const output = new Set(mockMessages.map((message) => message.id))
+			// Act
+			chatStore.processChatBlocks(TOKEN, mockMessages)
+
+			// Assert
+			expect(chatStore.chatBlocks[TOKEN]).toEqual([output])
+		})
+
+		it('extends and existing block, if messages overlap', () => {
```
DorraJaouad:
```suggestion
		it('extends an existing block, if messages overlap', () => {
```

src/stores/__tests__/chat.spec.js
```diff
@@ -0,0 +1,290 @@
...
+			expect(chatStore.chatBlocks[TOKEN]).toEqual([output])
+		})
+
+		it('extends and existing block, if messages adjacent', () => {
+			// Arrange
+			const output = new Set(mockMessages.map((message) => message.id))
+			chatStore.processChatBlocks(TOKEN, mockMessages.slice(0, 3))
+
+			// Act
+			chatStore.processChatBlocks(TOKEN, mockMessages.slice(3), { mergeBy: mockMessages[3 - 1].id })
```
DorraJaouad:
```suggestion
			chatStore.processChatBlocks(TOKEN, mockMessages.slice(3), { mergeBy: mockMessages[2].id })
```

no need

src/stores/__tests__/chat.spec.js
```diff
@@ -0,0 +1,290 @@
...
+			chatStore.processChatBlocks(TOKEN, mockMessages.slice(0, 2))
+			chatStore.processChatBlocks(TOKEN, mockMessages.slice(2, 4)) // main block
+
+			// Act
+			chatStore.processChatBlocks(TOKEN, mockMessages.slice(4))
+
+			// Assert
+			expect(chatStore.chatBlocks[TOKEN]).toHaveLength(3)
+			expect(chatStore.chatBlocks[TOKEN]).toEqual([outputMain, outputExtra1, outputExtra2])
+		})
```
DorraJaouad: You tested these very well in above tests, it is redundant

src/stores/__tests__/chat.spec.js
```diff
@@ -0,0 +1,290 @@
...
+			chatStore.processChatBlocks(TOKEN, mockMessages.slice(0, 1))
+			chatStore.processChatBlocks(TOKEN, mockMessages.slice(1, 3))
+
+			// Act
+			chatStore.processChatBlocks(TOKEN, mockMessages.slice(2, 5))
+
+			// Assert
+			expect(chatStore.chatBlocks[TOKEN]).toHaveLength(2)
+			expect(chatStore.chatBlocks[TOKEN]).toEqual([outputMain, outputExtra])
+		})
```
DorraJaouad: I'd keep this only if you test merge of 3 blocks :3

src/stores/__tests__/chat.spec.js
```diff
@@ -0,0 +1,290 @@
...
+			// Act
+			chatStore.addMessageToChatBlocks(TOKEN, mockMessages[2])
+
+			// Assert
+			expect(chatStore.chatBlocks[TOKEN]).toEqual([output])
+		})
+	})
+
+	describe('remove messages', () => {
+		it('does nothing, if main block is not created yet', () => {
```
DorraJaouad: Please do not forget rewording in sentences too

src/stores/__tests__/chat.spec.js
```diff
@@ -0,0 +1,290 @@
...
+		it('does nothing, if message is not present in existing block', () => {
+			// Arrange
+			chatStore.processChatBlocks(TOKEN, mockMessages.slice(0, 3))
+
+			// Act
+			chatStore.removeMessagesFromChatBlocks(TOKEN, mockMessages[4].id)
+
+			// Assert
+			expect(chatStore.chatBlocks[TOKEN][0].has(mockMessages[4].id)).toBeFalsy()
+		})
```
DorraJaouad: You need to show that block before and after is the same not the existence of the message id

src/stores/__tests__/chat.spec.js
```diff
@@ -0,0 +1,290 @@
...
+			chatStore.processChatBlocks(TOKEN, mockMessages.slice(2, 4)) // main block
+			expect(chatStore.chatBlocks[TOKEN][0].has(mockMessages[3].id)).toBeTruthy()
+			expect(chatStore.chatBlocks[TOKEN][1].has(mockMessages[3].id)).toBeFalsy()
+
+			// Act
+			chatStore.removeMessagesFromChatBlocks(TOKEN, mockMessages[3].id)
+
+			// Assert
+			expect(chatStore.chatBlocks[TOKEN][0].has(mockMessages[3].id)).toBeFalsy()
+		})
```
DorraJaouad: To be removed

src/stores/__tests__/chat.spec.js
```diff
@@ -0,0 +1,290 @@
...
+
+		it('clear up empty blocks after removing', () => {
+			// Arrange
+			chatStore.processChatBlocks(TOKEN, mockMessages.slice(0, 2))
+			chatStore.processChatBlocks(TOKEN, mockMessages.slice(2, 4)) // main block
+			expect(chatStore.chatBlocks[TOKEN][0]).toBeDefined()
```
DorraJaouad:
```suggestion
			expect(chatStore.chatBlocks[TOKEN].length).toEqual(2)
```

src/stores/__tests__/chat.spec.js
Antreesy: It's not in the messagesStore under token2, so I wouldn't expect it to appear in chatStore

src/stores/__tests__/chat.spec.js
Antreesy: Test is named 'returns empty array' for a reason

src/stores/__tests__/chat.spec.js
Antreesy: It works as expected

src/stores/__tests__/chat.spec.js
Antreesy: Was done for main+extras, will try to merge some tests without losing coverage

DorraJaouad: But we don't have main and extras anymore

DorraJaouad (approved):
Good to go 🦅

PR #15568, opened 2025-07-23
fix(sharing): Use insertIfNotExists to insert share exceptions to red…

nickvergessen:
…uce concurrency issues

### 🏁 Checklist

### DIFF
--- a/lib/Share/RoomShareProvider.php
+++ b/lib/Share/RoomShareProvider.php
@@ -353,45 +353,31 @@ class RoomShareProvider implements IShareProvider {
 		$this->cleanSharesByIdCache();
 
 		// Check if there is a userroom share
-		$qb = $this->dbConnection->getQueryBuilder();
-		$stmt = $qb->select(['id', 'permissions'])
-			->from('share')
-			->where($qb->expr()->eq('share_type', $qb->createNamedParameter(self::SHARE_TYPE_USERROOM)))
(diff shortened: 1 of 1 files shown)


lib/Share/RoomShareProvider.php
```diff
@@ -353,45 +353,31 @@ public function deleteFromSelf(IShare $share, $recipient): void {
...
+			'item_type' => $share->getNodeType(),
+			'item_source' => $share->getNodeId(),
+			'file_source' => $share->getNodeId(),
+			'file_target' => $share->getTarget(),
+			'permissions' => 0,
+			'stime' => $share->getShareTime()->getTimestamp(),
+		], ['share_type', 'share_with', 'parent']);
+		$this->dbConnection->commit();
+
+		if ($inserted === 0) {
```
ChristophWurst, line 372: Unfortunately wrapping insertIfNotExists in a transaction does not give more atomicity

https://github.com/nextcloud/server/pull/54041#issuecomment-3102721273

nickvergessen: While that is sad, the reduced roundtrip time should help
And the change here in `deleteFromSelf` which updates ALL share_type+with+parent items instead of the first one found by ID should solve the issue we are facing at the moment:
- After using "Unshare from self" and reloading the page, the share is visible again.
- THat should be fixed, as now all items have permissions = 0 and are not mounted

miaulalala (approved):

ChristophWurst (approved):

PR #15569, opened 2025-07-23
fix: respect presence of router in Talk integrations

Antreesy:
### ☑️ Resolves

* Fix regression from #15559

🏚️ Before | 🏡 After
-- | --
image | All clean
image | All clean

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop
  - [x] Integrations with Files sidebar and other apps ⚠️ 🆕 


src/composables/useGetMessages.ts
```diff
@@ -86,7 +86,7 @@ export function useGetMessagesProvider() {
 	})
 
 	/** Initial check to ensure context is created once route is available */
-	router.isReady().then(() => {
+	router?.isReady().then(() => {
```
ShGKme: Does it mean `handleStartGettingMessagesPreconditions` is not called without router and it is intended?

ShGKme:
What about using in-memory router for integrations?

src/composables/useGetMessages.ts
Antreesy: It will be triggered with a watcher, as before. I checked that it doesn't require immediate flag, but maybe we can put it for integrations, if it's safer

Antreesy:
> What about using in-memory router for integrations?

Would be nice to have it. Then we can unify handling of hash and queries for context and threads there.

But it didn't  work before, so it's not related to this PR. I will open an issue

src/composables/useGetMessages.ts
Antreesy: UPD: All good here, integrations need to fetch a conversation first, that will guarantee identifying values of `token`, `isParticipant`, `isInLobby`, which are blockers for receiving a messages context

image

immediate is not required, route is not required either

PR #15574, opened 2025-07-24
fix: use memory router for Talk integration

### LINKED ISSUE #15573: RFC: Using in-memory router for Talk integrations
> What about using in-memory router for integrations?

_Originally posted by @ShGKme in https://github.com/nextcloud/spreed/issues/15569#issuecomment-3109800214_

Currently Talk integrations (in Files Sidebar) do not have Vue-Router initializing, which blocks utilizing following features:
- navigating to quoted message
  - getting context for older messages (in the same chat)
- navigating to/from thread
  - currently workaround is implemented in useGetThreadId
- anything else..?

Ideas:
1. Implement cropped router (only with hash, query, params). Available route is only one - 'conversation'. Navigation (push, replace) only allowed with hash and query. Don't affect page URL
  1.1. use `createMemoryHistory()` for it: https://router.vuejs.org/guide/migration/#New-history-option-to-replace-mode

2. If it's only two features, realize both via separate composables (that rely on route / memory value)
  2.1 Not future-proof

Antreesy:
### ☑️ Resolves

* Fix #15573
  * testing with Files Sidebar (from files, public, public auth)
  * integration route is matching message actual URL

🏚️ Before | 🏡 After
-- | --
image | All clean
image | All clean

🏡 After

https://github.com/user-attachments/assets/916600a3-339c-4935-9e42-07a4663cee7b

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop
  - [x] Integrations with Files sidebar and other apps 🆕 

### DIFF
--- a/src/PublicShareAuthSidebar.vue
+++ b/src/PublicShareAuthSidebar.vue
@@ -14,7 +14,7 @@
 				<TopBar is-in-call is-sidebar />
 				<CallView :token="token" is-sidebar />
 				<InternalSignalingHint />
-				<ChatView is-sidebar />
+				<RouterView />
 				<PollManager />
 				<PollViewer />
 				<MediaSettings v-model:recording-consent-given="recordingConsentGiven" />
@@ -29,7 +29,6 @@ import { emit } from '@nextcloud/event-bus'
 import { loadState } from '@nextcloud/initial-state'
 import { t } from '@nextcloud/l10n'
 import CallView from './components/CallView/CallView.vue'
-import ChatView from './components/ChatView.vue'
 import MediaSettings from './components/MediaSettings/MediaSettings.vue'
 import PollManager from './components/PollViewer/PollManager.vue'
 import PollViewer from './components/PollViewer/PollViewer.vue'
@@ -56,7 +55,6 @@ export default {
 	components: {
 		InternalSignalingHint,
 		CallView,
-		ChatView,
 		MediaSettings,
 		PollManager,
 		PollViewer,
@@ -136,6 +134,7 @@ export default {
 				this.actorStore.setDisplayName(guestNickname)
 			}
 
+			await this.$router.push({ name: 'conversation', params: { token: this.token } })
 			await this.$store.dispatch('joinConversation', { token: this.token })
 
 			// Add guest name to the store, only possible after joining the conversation
--- a/src/router/router.ts
+++ b/src/router/router.ts
@@ -3,8 +3,14 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
+import type { RouteRecordRaw } from 'vue-router'
+
 import { generateUrl, getRootUrl } from '@nextcloud/router'
-import { createRouter, createWebHashHistory, createWebHistory } from 'vue-router'
+import {
+	createRouter,
+	createWebHashHistory,
+	createWebHistory,
+} from 'vue-router'
 import CallView from '../components/CallView/CallView.vue'
 import ForbiddenView from '../views/ForbiddenView.vue'
 import MainView from '../views/MainView.vue'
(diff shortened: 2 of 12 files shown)


src/views/FilesSidebarChatView.vue
```diff
@@ -8,7 +8,7 @@
 		<InternalSignalingHint />
 		<CallButton v-if="!isInCall" class="talk-tab__call-button" />
 		<CallFailedDialog v-if="connectionFailed" :token="token" />
-		<ChatView class="talk-tab__chat-view" is-sidebar />
+		<router-view class="talk-tab__chat-view" />
```
ShGKme:
```suggestion
		<RouterView class="talk-tab__chat-view" />
```

src/router/router.ts
```diff
@@ -27,48 +34,66 @@ function generateTalkWebBasePath(): string {
...
+		})
+	}
+
+	return createRouter({
+		// On desktop (Electron) app is open via file:// protocol - History API is not available and no base path
+		history: !IS_DESKTOP ? createWebHistory(generateTalkWebBasePath()) : createWebHashHistory(''),
+		linkActiveClass: 'active',
+		routes,
+	})
+}
```
ShGKme: With a generic function this module always uses all the imported components, which is likely to result in bundling all the views.

Views like `NotFoundView` are not a problem, but `WelcomeView` with the dashboard, recording might increase bundle more significantly.

Instead, let's provide two separated functions: `createTalkRouter` for the main app and `createTalkIntegrationRouter` or `createTalkMemoryRouter`. And define the routes inside functions, not in a module-scope. It would also make functions more explicit with routes inside.

src/PublicShareAuthSidebar.vue
```diff
@@ -65,6 +64,13 @@ export default {
 	},
 
 	setup() {
+		const router = useRouter()
+		router.beforeEach((to, from) => {
+			if (from.name === 'conversation' && to.name === 'conversation' && from.params.token === to.params.token) {
+				EventBus.emit('route-change', { from, to })
+			}
+		})
```
ShGKme: Let's do it in the router setup to have it only once.

ShGKme: But the condition looks suspicious. It is supposed to emit an event only on route query/hash change?

Antreesy: Yes, it's an exerpt from App router listener: https://github.com/nextcloud/spreed/blob/2d07925cc17cd96486b0cca997385fde064ff23a/src/App.vue#L401-L405

And it's listened by useGetMessageProvider:
https://github.com/nextcloud/spreed/blob/ef8530b41b2527170e4958110704f4e7b6500614/src/composables/useGetMessages.ts#L164-L167

I'd love to move it to router setup, but atm it heavily relies on some App component variables.

And we currently need only this for chat fetching; and in sidebar it's always the same conversation/token

Antreesy: Or you mean this?

```js
export function createMemoryRouter() {
	const router = createRouter({
		history: createMemoryHistory(generateTalkWebBasePath()),
		routes,
	})
	
	router.beforeEach((to, from) => {
		if (from.name === 'conversation' && to.name === 'conversation' && from.params.token === to.params.token) {
			EventBus.emit('route-change', { from, to })
		}
	})
	
	return router
}
```

DorraJaouad: Well, route path is only conversation's and we only have one conversation open at a time, so that condition is always truthy?

Antreesy: In general, so far in sidebar it would be only used by useGetMessages, which has the same check (for now).

But for other cases, we have in MessageButtonsBar:
```
this.$router.push({ name: 'conversation', params: { token: conversation.token } })
```
which might reroute to different conversation. It would be safe to keep the check for cases like this.

src/router/router.ts
```diff
@@ -70,5 +77,39 @@ export default createRouter({
...
+
+	const router = createRouter({
+		history: createMemoryHistory(generateTalkWebBasePath()),
+		routes,
+	})
+
+	router.beforeEach((to, from) => {
+		if (from.name === 'conversation' && to.name === 'conversation' && from.params.token === to.params.token) {
+			EventBus.emit('route-change', { from, to })
+		}
```
Antreesy, line 117: This is missing to block navigation to other routes/conversations
https://router.vuejs.org/guide/advanced/navigation-guards#Global-Before-Guards

```suggestion
		} else {
			// explicitly return false to cancel the navigation
			return false
		}
```

PR #15582, opened 2025-07-25
fix(thread): replace createThread with setThreadNotificationLevel

Antreesy:
### ☑️ Resolves

* Follow-up to #15579

A | B | C | D
-- | -- | -- | --
image | image | image | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/TopBar/TopBar.vue
+++ b/src/components/TopBar/TopBar.vue
@@ -56,16 +56,27 @@
 				</div>
 			</span>
 
-			<NcButton
-				:aria-label="threadNotificationLabel"
-				:title="threadNotificationLabel"
-				:primary="!!threadNotification"
-				@click="() => {}">
+			<NcActions
+				:aria-label="t('spreed', 'Subscribe to thread')"
+				:title="t('spreed', 'Subscribe to thread')"
+				:variant="threadNotificationVariant">
 				<template #icon>
-					<IconBellOffOutline v-if="threadNotification" :size="20" />
-					<IconBellOutline v-else :size="20" />
+					<component :is="notificationLevelIcons[threadNotification]" :size="20" />
 				</template>
-			</NcButton>
+				<NcActionButton v-for="level in notificationLevels"
+					:key="level.value"
+					:model-value="threadNotification.toString()"
+					:value="level.value.toString()"
+					:description="level.description"
+					type="radio"
+					close-after-click
+					@click="chatExtrasStore.setThreadNotificationLevel(token, threadId, level.value)">
+					<template #icon>
+						<component :is="notificationLevelIcons[level.value]" :size="16" />
+					</template>
+					{{ level.label }}
+				</NcActionButton>
+			</NcActions>
 		</div>
 
 		<div v-else
@@ -164,6 +175,8 @@
 <script>
 import { emit } from '@nextcloud/event-bus'
 import { n, t } from '@nextcloud/l10n'
+import NcActionButton from '@nextcloud/vue/components/NcActionButton'
+import NcActions from '@nextcloud/vue/components/NcActions'
 import NcButton from '@nextcloud/vue/components/NcButton'
 import NcPopover from '@nextcloud/vue/components/NcPopover'
 import NcRichText from '@nextcloud/vue/components/NcRichText'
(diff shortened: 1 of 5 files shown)


src/components/TopBar/TopBar.vue
```diff
@@ -60,7 +60,7 @@
 				:aria-label="threadNotificationLabel"
 				:title="threadNotificationLabel"
 				:primary="!!threadNotification"
-				@click="() => {}">
+				@click="chatExtrasStore.setThreadNotificationLevel(token, threadId, threadNotification ? 0 : 1)">
```
nickvergessen: We should allow all 4 notification levels?
- Default (basically reset to conversation level)
- Always
- Mention
- Never

?

And if we say "that's too much", we should toggle between 1 and 3 (always and never), and not always and default (which could be the same, as many people asked for setting the group default also to always)

Antreesy: We should, that's just a plug for now

Antreesy: Have this plenty of options:

image

Since we need to cover:
- Default (bell-outline)
- Always (bell-ring / bell-alert) ❓ 
- Mentions (bell-cog / @ symbol as in other places) ❓ 
- Never (bell-off-outline)

nickvergessen: grafik

Should align with the conversation settings icon?

Antreesy: Yeah, but these make sense inside of the menu only, and it's missing 'default'

On the outside, we should see some variations of a bell icon

nickvergessen: Always / All messages could use Chat or Comment, or the Reply icon?
grafik

Antreesy: default | other | never
-- | -- | --
image | image | image

ShGKme: For `@-mentions` we can use a bell with `@` sign on the bottom right
(To have direct association with @ mentions)

ShGKme: For "All messages" I'd use just bell-outline

ShGKme: And for "Follow conversation settings" - bell-cog-outline (related to settings)

Antreesy: >  a bell with `@` sign on the bottom right

If such icon would exist

nimishavijay: > If such icon would exist

It doesn't. What do you think about differentiating by color? So 
- follow conversation settings: tertiary with bell
- All messages: secondary with bell-ring
- @-mentions: secondary with bell
- Off: tertiary with bell-off

This brings attention to the icon if it's set to all or mentions, and not if it is default or off, what do you think?

Also unrelated, would it be possible to show what the conversation setting is with a subline?

```
🔇 Off                          ✔️
   from conversation settings
---------------------------------------
🔊 All messages
👤 @-mentions only
🔇 Off
```

ShGKme: > If such icon would exist

We can always stack one icon on top of another 👀

ShGKme: image

src/components/TopBar/TopBar.vue
```diff
@@ -56,16 +56,27 @@
...
 
-			<NcButton
-				:aria-label="threadNotificationLabel"
-				:title="threadNotificationLabel"
-				:primary="!!threadNotification"
-				@click="() => {}">
+			<NcActions
+				:aria-label="t('spreed', 'Subscribe to thread')"
+				:title="t('spreed', 'Subscribe to thread')"
+				:variant="threadNotificationVariant">
```
ShGKme, line 62: Doesn't it miss a capability check?

Antreesy: It's inside of `<div v-if="currentThread" />`, so I don't think it's needed

src/components/TopBar/TopBar.vue
jancborchardt: (Replying directly as @nimishavijay  is out today :)
We don’t need to have icons which all include the notification bell, as they would also be more difficult to tell apart.

Is there a specific issue about the current icons that needs to be solved? If we want to switch from the "person" icon to an "@" that’s fine too, however it repeats the first character of the text right next to it.

Antreesy: > We don’t need to have icons which all include the notification bell, as they would also be more difficult to tell apart.
> Is there a specific issue about the current icons that needs to be solved?

My take is that outside icon should always depict it's a notification, and thus include a bell icon. It feels off to have '@' outside, I as user wouldn't understand, what's this about
image

If we don't implement custom icons for 'missing' types of bells, then color coding should be enough

PR #15584, opened 2025-07-25
feat(threads): Allow to set thread title (frontend)

Antreesy:
### ☑️ Resolves

* Ref #9680

🏚️ Before | 🏡 After
-- | --
action | 2025-07-29_11h49_17 2025-07-29_11h49_51
input | image image
file share | image

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/components/NewMessage/NewMessage.vue
+++ b/src/components/NewMessage/NewMessage.vue
@@ -85,6 +85,8 @@
 				<NcRichContenteditable ref="richContenteditable"
 					:key="container"
 					v-model="text"
+					:class="{ 'new-message-form__input-rich--required': errorTitle }"
+					:title="errorTitle"
 					:auto-complete="autoComplete"
 					:disabled="disabled"
 					:user-data="userData"
@@ -324,6 +326,7 @@ export default {
 	data() {
 		return {
 			text: '',
+			errorTitle: '',
 			silentChat: false,
 			// True when the audio recorder component is recording
 			isRecordingAudio: false,
@@ -525,6 +528,7 @@ export default {
 		},
 
 		text(newValue) {
+			this.errorTitle = ''
 			if (this.currentUploadId && !this.upload) {
 				return
 			} else if (this.dialog && this.broadcast) {
@@ -701,6 +705,10 @@ export default {
 
 			if (this.hasText) {
 				this.text = parseSpecialSymbols(this.text)
+			} else if (this.threadCreating && !this.hasText) {
+				// TRANSLATORS Error indicator: do not allow to create a thread without a message text
+				this.errorTitle = t('spreed', 'Message text is required')
+				return
 			}
 
 			// Clear input content from store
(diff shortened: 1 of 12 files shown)


src/components/NewMessage/NewMessage.vue
```diff
@@ -466,6 +486,29 @@ export default {
...
+		},
+
+		threadTitle: {
+			get() {
+				return this.chatExtrasStore.getThreadTitle(this.token)
+			},
+
+			set(value) {
+				this.chatExtrasStore.setThreadTitle(this.token, value)
+			},
```
DorraJaouad, line 506: Image

src/components/NewMessage/NewMessage.vue
```diff
@@ -655,37 +710,45 @@ export default {
 			this.debouncedUpdateChatInput.clear()
 			this.chatExtrasStore.removeChatInput(this.token)
 
-			if (this.hasText || (this.dialog && this.upload)) {
+			if (this.hasText || this.threadTitle || (this.dialog && this.upload)) {
+				// Substitute message text / thread title, if one is missing
+				const message = this.text.trim() || this.threadTitle.trim()
+				const threadTitle = this.threadCreating
+					? this.threadTitle.trim() || this.text.trim()
+					: undefined
```
DorraJaouad: IMO, we should not accept thread title without message. If someone uses that field, he accepts the structure and has to add  message too.

src/components/NewMessage/NewMessage.vue
```diff
@@ -359,6 +377,8 @@ export default {
 		sendMessageLabel() {
 			if (this.silentChat) {
 				return t('spreed', 'Send message silently')
+			} else if (this.threadCreating) {
+				return t('spreed', 'Create a thread')
 			} else {
 				return t('spreed', 'Send message')
 			}
```
DorraJaouad: What about when silent is toggled on? `t('spreed', 'Create a thread silently')`

src/components/NewMessage/NewMessage.vue
```diff
@@ -72,6 +74,14 @@
 				<NcNoteCard v-if="showMentionEditHint"
 					type="warning"
 					:text="t('spreed', 'Adding a mention will only notify users who did not read the message.')" />
+				<NcTextField
```
DorraJaouad, line 77: If this is present, I suggest to make new message input longer in height.

src/components/NewMessage/NewMessageAttachments.vue
```diff
@@ -52,6 +52,16 @@
...
 
+		<NcActionButton
+			v-if="canCreateThread"
+			close-after-click
+			@click="$emit('createThread', true)">
+			<template #icon>
+				<IconForumOutline :size="16" />
+			</template>
+			{{ t('spreed', 'Create a thread') }}
+		</NcActionButton>
```
DorraJaouad, line 63: I don't suggest to put it here, I was expecting it in 3 dots button where silent option is and called 'Send as a thread'

src/components/NewMessage/NewMessage.vue
```diff
@@ -466,6 +486,29 @@ export default {
 		canEditMessage() {
 			return hasTalkFeature(this.token, 'edit-messages')
 		},
+
+		supportThreads() {
+			return hasTalkFeature(this.token, 'threads')
+		},
```
DorraJaouad, line 492: Why did we return to putting it in computed instead of constant? I missed that episode :3

src/components/NewMessage/NewMessage.vue
Antreesy: I don't see the reason behind it. There is already a placeholder in place

src/components/NewMessage/NewMessageAttachments.vue
Antreesy: It's not a message modificator, but another entity. It will not look like a normal message in chat either. 

We already discussed to put it in attachments menu

src/components/NewMessage/NewMessage.vue
Antreesy: If feature is supported in federations and not locally only, it should be computed from remote capabilities

src/components/NewMessage/NewMessage.vue
DorraJaouad: Visual reasons

src/components/NewMessage/NewMessage.vue
Antreesy: Reported upstream

DorraJaouad:
@Antreesy Do you think we should show the title in bold and not hide it? It feels weird to have it not shown in message. 

I would tweak MessageBody to show the title too in different computed from the message text and also when editing, you will have title and message separately.

src/components/NewMessage/NewMessage.vue
DorraJaouad: Maybe place the emoji picker in top corner? `new-message-form__emoji-picker`

src/components/NewMessage/NewMessage.vue
```diff
@@ -1000,6 +1075,23 @@ export default {
...
+			border: 2px solid var(--color-border-maxcontrast);
+		}
+
+		& + :deep(.rich-contenteditable > .rich-contenteditable__input) {
+			min-height: calc(2lh + 2 * var(--contenteditable-block-offset) + 4px);
+		}
+
+		&--required + :deep(.rich-contenteditable > .rich-contenteditable__input) {
+			border-color: var(--color-error) !important;
+		}
```
DorraJaouad: No need if the button is disabled. An alternative is to keep the button enabled and show error border if it is pressed while the message input is empty

DorraJaouad (approved):

PR #15596, opened 2025-07-28
fix(chatStore): add search, cleanup, getFirst / getLast methods

Antreesy:
### ☑️ Resolves

* Follow-up to #15566 
* Pre-requisite to #10084

No visual changes

### 🏁 Checklist

  - [ ] Firefox
  - [ ] Safari
  - [ ] Talk Desktop

### DIFF
--- a/src/stores/chat.ts
+++ b/src/stores/chat.ts
@@ -97,6 +97,64 @@ export const useChatStore = defineStore('chat', () => {
 	}
 
 	/**
+	 * Returns first known message id, belonging to current context. Defaults to given messageId
+	 */
+	function getFirstKnownId(
+		token: string,
+		{ messageId = 0, threadId = 0 }: GetMessagesListOptions = { messageId: 0, threadId: 0 },
+	): number {
+		if (!chatBlocks[token]) {
+			return messageId
+		}
+
+		if (threadId) {
+			// If topmost message of thread is in the store, return its id
+			if (hasMessage(token, { messageId: threadId, threadId })) {
+				return threadId
+			}
+			// FIXME temporary check all messages for given thread from all chat blocks
+			return Math.min(...prepareMessagesList(token, new Set(Array.from(chatBlocks[token].flatMap((set) => Array.from(set)))))
+				.filter((message) => {
+					return message.threadId === threadId
+				}).map((message) => message.id))
+		}
+
+		if (messageId <= 0) {
+			return Math.min(...chatBlocks[token][0])
+		}
+
+		const contextBlock = chatBlocks[token].find((set) => set.has(messageId))
+		return contextBlock ? Math.min(...contextBlock) : Math.min(...chatBlocks[token][0])
+	}
+
+	/**
+	 * Returns last known message id, belonging to current context. Defaults to given messageId
+	 */
+	function getLastKnownId(
+		token: string,
+		{ messageId = 0, threadId = 0 }: GetMessagesListOptions = { messageId: 0, threadId: 0 },
+	): number {
+		if (!chatBlocks[token]) {
+			return messageId
+		}
+
+		if (threadId) {
+			// FIXME temporary check all messages for given thread from all chat blocks
+			return Math.max(...prepareMessagesList(token, new Set(Array.from(chatBlocks[token].flatMap((set) => Array.from(set)))))
+				.filter((message) => {
+					return message.threadId === threadId
+				}).map((message) => message.id))
+		}
+
+		if (messageId <= 0) {
+			return Math.max(...chatBlocks[token][0])
+		}
+
+		const contextBlock = chatBlocks[token].find((set) => set.has(messageId))
+		return contextBlock ? Math.max(...contextBlock) : Math.max(...chatBlocks[token][0])
+	}
+
+	/**
 	 * Populate chat blocks from given arrays of messages
 	 * If blocks already exist, try to extend them
 	 */
@@ -241,6 +299,8 @@ export const useChatStore = defineStore('chat', () => {
 
 		getMessagesList,
 		hasMessage,
+		getFirstKnownId,
+		getLastKnownId,
 		processChatBlocks,
 		addMessageToChatBlocks,
 		removeMessagesFromChatBlocks,
--- a/src/stores/chat.ts
+++ b/src/stores/chat.ts
@@ -12,6 +12,13 @@ import { defineStore } from 'pinia'
 import { reactive } from 'vue'
 import { useStore } from 'vuex'
 
+type GetMessagesListOptions = {
+	/** if given, look for Set that has it */
+	messageId?: number
+	/** if given, look for thread Set */
+	threadId?: number
+}
+
 type ProcessChatBlocksOptions = {
 	/** if given, look for Set that has it */
 	mergeBy?: number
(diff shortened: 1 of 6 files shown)


src/stores/chat.ts
```diff
@@ -161,16 +262,67 @@ export const useChatStore = defineStore('chat', () => {
...
+		if (deleteIndex === -1) {
+			// Not found, nothing to delete
+			return
+		} else if (deleteIndex === 0) {
+			// If first block is to be deleted, remove all blocks
+			purgeChatStore(token)
+			return
+		} else {
+			// Remove all blocks with max id less than given id
+			chatBlocks[token] = chatBlocks[token].slice(0, deleteIndex)
```
DorraJaouad, line 280: slice end is exclusive
```suggestion
			chatBlocks[token] = chatBlocks[token].slice(0, deleteIndex + 1)
```

also you need to include the block that contains `idToDelete`

```suggestion
			chatBlocks[token] = chatBlocks[token].slice(0, deleteIndex + 2)
```

But you can change find index comparison to inclusive `<=` and you have to change the case when first block is to be deleted, so I recommend the suggestion :P

src/stores/chat.ts
```diff
@@ -161,16 +262,67 @@ export const useChatStore = defineStore('chat', () => {
...
+			const lastBlock = chatBlocks[token].at(-1)!
+			for (const id of lastBlock) {
+				if (id < idToDelete) {
+					lastBlock.delete(id)
+				}
+			}
+			if (chatBlocks[token].length === 1) {
+				firstKnown[token] = Math.min(...chatBlocks[token][0])
+				lastKnown[token] = Math.max(...chatBlocks[token][0])
+			}
```
DorraJaouad: I am not sure if we forced order inside a block?

src/stores/chat.ts
```diff
@@ -59,17 +79,92 @@ export const useChatStore = defineStore('chat', () => {
...
+			return messageId
+		}
+
+		if (threadId) {
+			// FIXME temporary check all messages for given thread from all chat blocks
+			return Math.min(...prepareMessagesList(token, new Set(Array.from(chatBlocks[token].flatMap((set) => Array.from(set)))))
+				.filter((message) => {
+					return message.threadId === threadId
+				}).map((message) => message.id))
+		}
```
DorraJaouad: If message that has id equal to threadId then that's the first known. Otherwise, you missed substracting the last element and you are returning an array now

src/stores/chat.ts
```diff
@@ -59,17 +79,92 @@ export const useChatStore = defineStore('chat', () => {
...
+
+		if (messageId <= 0) {
```
DorraJaouad, line 122: Why less then 0?

src/stores/chat.ts
```diff
@@ -78,6 +173,8 @@ export const useChatStore = defineStore('chat', () => {
 		}
 
 		chatBlocks[token] = mergeAndSortChatBlocks(chatBlocks[token], newMessageIdsSet)
+		firstKnown[token] = Math.min(...chatBlocks[token][0])
+		lastKnown[token] = Math.max(...chatBlocks[token][0])
```
DorraJaouad: It doesn't seem correct to make an update here, not in general chat because current processed block can be older and has first known and nor in threads as it will affect its context

src/stores/chat.ts
```diff
@@ -161,16 +262,67 @@ export const useChatStore = defineStore('chat', () => {
 		}, [])
 
 		if (chatBlocks[token].length === 0) {
-			delete chatBlocks[token]
+			purgeChatStore(token)
+		} else {
+			firstKnown[token] = Math.min(...chatBlocks[token][0])
+			lastKnown[token] = Math.max(...chatBlocks[token][0])
```
DorraJaouad: Same here, I would update only if the removed message id is one these two

src/stores/chat.ts
DorraJaouad: Also, we can do this regardless of length, last known is correct like that and first known is definitely `idToDelete`.

Again, as mentioned previously, this may affect thread context

Antreesy: We don't and we can't

src/stores/chat.ts
Antreesy: > If message that has id equal to threadId then that's the first known.

Ok, for firstKnown it makes sense. But I should have it in the store anyway

src/stores/chat.ts
Antreesy: CHAT_BEGIN_ID: -2,
CHAT_MIGRATION_ID: -1,

src/stores/chat.ts
Antreesy: We are updating only first block here always. if it's older and separate, we shouldn't care

src/stores/chat.ts
Antreesy: > also you need to include the block that contains `idToDelete`

No, because idToDelete can be not the last, and block will be only partially cleared? It is handled as `lastBlock` below

src/stores/chat.ts
DorraJaouad: then why are you only checking first item?

src/stores/chat.ts
DorraJaouad: idToDelete can be in the middle 

```js	
	chatBlocks[token] = chatBlocks[token].slice(0, deleteIndex)
			const lastBlock = chatBlocks[token].at(-1)!
			for (const id of lastBlock) {
				if (id < idToDelete) {
					lastBlock.delete(id)
				}
			}
```

This lastBlock is the block that doesn't inlcude idToDelete

src/stores/chat.ts
DorraJaouad: > But I should have it in the store anyway

You mean firstKnown will be threadId without this part of code? how?

src/stores/chat.ts
DorraJaouad: > We are updating only first block here always. 

No we are not

src/stores/chat.ts
Antreesy: > const deleteIndex = chatBlocks[token].findIndex((block) => Math.max(...block) < idToDelete)

for lastBlock this would be false, thus it's deleted. It either includes idToDelete or not, but in this case Math.min is larger and won't be deleted

src/stores/chat.ts
Antreesy: `inside a block` - it's a Set, it can't be ordered
`chatBlocks` - it's an array of Sets, where first item [0] is the most recent, and we're only taking it in consideration

src/stores/chat.ts
DorraJaouad: Checked, all good

src/stores/chat.ts
Antreesy: If I have only second part of the thread and want to scroll up to get the rest, I need to know firstKnown of that I already have in the store to start fetch from it

src/stores/chat.ts
Antreesy: chatBlocks[token] is actual at this point:
`chatBlocks[token] = mergeAndSortChatBlocks(chatBlocks[token], newMessageIdsSet)`

but we can't tell, if it was modified during the process. That would require storing previous state of chatBlocks[token][0], and comparing to the new one - easier to just look for min/max values one extra time

src/stores/chat.ts
Antreesy: thread context is ignored until follow-up

src/stores/chat.ts
DorraJaouad: But it's wrong, you are using this method in 3 points:
1) Poll new messages
2) get context
3) Fetch messages

For 1., this way is fully accurate but for 2 if it is focused message id context, this is completely wrong and for 3 it is only changing of those two by expanding it away. and no, it is not a corner case

DorraJaouad: My suggestion: you need to handle these two inside merge and sort. 
1. We first evaluate the current last and first known to catch in which block they are.
2. If their block is merged we update with new version of it
3. If not, we do nothing

Antreesy: Drop in favor of computations in-place

DorraJaouad (approved):
Fine for now

PR #15600, opened 2025-07-29
fix(threads): fix creating threads from file shares

Antreesy:
### ☑️ Resolves

* Fixup to #15579
* Not sure if it's the right way, but there are a lot of checks like this 🤷🏽 

### 🚧 Tasks

- [ ] Check if it's working

### 🏁 Checklist


nickvergessen (approved):
Yeah sorry, these strings that are numbers always mess me up. getParentId vs. getTopmostParentId makes no difference here, but the `''` vs. `'0'` does.

PR #15604, opened 2025-07-30
fix(threads): Add the thread title to the system message

nickvergessen:
## 🛠️ API Checklist
- Should allow clients to update the thread list from pulling the chat
  - Title is a parameter
  - Last activity the timestamp of the system message
  - num replies 0
  - last message null
  - first message the parent

### 🏁 Checklist

### DIFF
--- a/lib/Chat/SystemMessage/Listener.php
+++ b/lib/Chat/SystemMessage/Listener.php
@@ -412,7 +412,7 @@ class Listener implements IEventListener {
 			$this->sendSystemMessage(
 				$room,
 				'thread_created',
-				['thread' => (int)$comment->getId()],
+				['thread' => (int)$comment->getId(), 'title' => $threadTitle],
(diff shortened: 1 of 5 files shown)


nickvergessen:
- Red CI fixed here https://github.com/nextcloud/spreed/pull/15605

lib/Chat/SystemMessage/Listener.php
```diff
@@ -412,7 +412,7 @@ protected function fixMimeTypeOfVoiceMessage(ShareCreatedEvent|BeforeDuplicateSh
 			$this->sendSystemMessage(
 				$room,
 				'thread_created',
-				['thread' => (int)$comment->getId()],
+				['thread' => (int)$comment->getId(), 'title' => $threadTitle],
```
Antreesy, line 427: This system message is missing the parent

PR #15608, opened 2025-07-31
feat(threads): Allow to rename a thread

nickvergessen:
## 🛠️ API Checklist
- Ref #9679
- Allow to rename a thread
- Add "you" version of rename and create system message

### 🏁 Checklist

### DIFF
--- a/lib/Controller/ThreadController.php
+++ b/lib/Controller/ThreadController.php
@@ -28,6 +28,7 @@ use OCP\AppFramework\Http\Attribute\PublicPage;
 use OCP\AppFramework\Http\Attribute\RequestHeader;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\Comments\NotFoundException;
 use OCP\EventDispatcher\IEventDispatcher;
 use OCP\IL10N;
 use OCP\IRequest;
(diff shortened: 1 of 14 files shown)


lib/Controller/ThreadController.php
```diff
@@ -128,6 +132,83 @@ public function getThread(int $threadId): DataResponse {
...
+	#[RequireModeratorOrNoLobby]
+	#[RequireParticipant]
+	#[RequestHeader(name: 'x-nextcloud-federation', description: 'Set to 1 when the request is performed by another Nextcloud Server to indicate a federation request', indirect: true)]
+	#[ApiRoute(verb: 'PUT', url: '/api/{apiVersion}/chat/{token}/threads/{threadId}', requirements: [
+		'apiVersion' => '(v1)',
+		'token' => '[a-z0-9]{4,30}',
+		'threadId' => '[0-9]+',
+	])]
+	public function renameThread(int $threadId, string $threadTitle): DataResponse {
+		$threadTitle = trim($threadTitle);
```
miaulalala, line 157: Should there be a length limit on the thread title?

nickvergessen: yup should do so, also on posting to avoid DB length issues,
or should we trim?

miaulalala: I'd vote for throwing the same exception as for an empty string in the db code, and then you could trim with an ellipsis in the controller unless the exception should be bubbled up.

nickvergessen: I keep the trim in the service now, so I don't have to do it in many places :see_no_evil:

SystemKeeper: > I keep the trim in the service now, so I don't have to do it in many places 🙈

But there's only truncation right now, no trimming? For rooms we have a trim before checking for empty string: https://github.com/nextcloud/spreed/blob/d77e6f4994931ffe9eb7caa6f8ebbc4c1accd4f5/lib/Service/RoomService.php#L157-L160

